PR #1, opened 2015-06-08
Revised Get-SQLDefaultPaths function

dougjjj:
I had to remove the split-path($filepath), on systems with mount points and shallow directory structures it was reverting to folders which database engine lacked permissions (for instance if the data directory is d:\data it would change to d:\ and would fail due to lack of permissions).  Also revised the switch which was replacing "data" with "mdf" and "log" with "ldf" and then bypassing the conditional test.


ctrlbold:
Woo! Thanks for the fix.

PR #5, opened 2016-01-27
Added DisableJobsOnDestination flag to Copy-SqlJobServer

thomasalley:
I ran into a scenario where I wanted to copy all of the jobs from server A to server B, but those copied jobs needed to be immediately disabled on server B. Dropped a flag into Copy-SqlJobServer to support this by disabling the job in the SMO object before scripting out for the copy. Figured other people working with Availability Groups may want the same functionality.

I also ran into a situation where job creation on the destination was failing due to the Job's starting database not existing on the server. Since the exception had 'exist' somewhere in there, the command misreported that the job had already been migrated. Tweaked that check and added a Write-Error to get some useful output from this case.


ctrlbold:
Awesome, thanks, Thomas!

PR #14, opened 2016-03-09
More copy options and support of -WhatIf, -Confirm

sqlslinger:
I added support for crucial objects that appear in system databases,
such as users, roles, database role memberships, defaults, rules,
synonyms, etc.

I also added support for -WhatIf and -Confirm options, which I think are
important here.


ctrlbold:
Frikken awesome! Yes. Thank you.

PR #15, opened 2016-03-18
Detects server version before CMS migration 

dhrubomoy:
If the source or destination server's SQL Server version is less than 2008 (versionMajor less than 10) then Copy-SqlCentralManagementServer will quit.


ctrlbold:
Awesome, thank you so much Das!

PR #20, opened 2016-05-10
Addition of CopyAllUserDefinedFunctions property set

thesqlpimp:
This function is missing the property set to transfer functions across. I tested with this being added in and the functions came across fine.

### DIFF
--- a/Functions/Copy-SqlSysDbUserObjects.ps1
+++ b/Functions/Copy-SqlSysDbUserObjects.ps1
@@ -70,6 +70,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 		$transfer.CopyAllUserDefinedDataTypes = $true
 		$transfer.CopyAllUserDefinedTableTypes = $true
 		$transfer.CopyAllUserDefinedTypes = $true
+		$transfer.CopyAllUserDefinedFunctions = $true
 		$transfer.CopyAllUsers = $true
 		$transfer.PreserveDbo = $true
 		$transfer.Options.AllowSystemObjects = $false


ctrlbold:
Fantastic! Thank you. Btw, keep an eye out on https://dbatools.io. I'm releasing a new version this week.

thesqlpimp:
Good afternoon Chrissy.

I'm giving this a good workout today and have another change. Before I submit it I thought I thought I'd check if you had had any similar issue. In the Copy-SqlDatabaseMail function, I hit errors with the Write-Verbose statements. I tried it on an instance with eight mail profiles and four accounts, which resulted in errors as the Write-Verbose statements don't like arrays, either of System.Object or String Collection. Easy to fix with a quick check and -join logic if an array is returned from the SMO script functions. I'm happy to send you the changed function prior to submitting it if you want to test it first? I have tested it with the changes and it's fine as, like I say, the only functional change is to the logic around what is written to the Verbose stream.

I believe that you met my colleague Mark at the PowerShell conference in Germany and we have done a lot of similar PowerShell work. We are already in the process of putting your tools to work!

If you're feeling brave then I think we can contribute to your tools in a good way. Mark said that he mentioned the way we mirror the databases then fail-over for almost no downtime. I'm happy to put in the functions to do this if you think that would be cool? 

In reality what we normally do on an instance migration is as follows:

Prepare the destination:

1. - Synch logins between source and destination
2. - Create the databases on the destination as schema-only (via SMO, just schema no data, I obviously have the code for this and this could just be another path through Migrate Instance in addition to detach and restore)
3. - Now that the database shells are on the destination we can sync linked servers, jobs, etc.

Mirror and Fail-over:

- Mirror the database from source to destination (async)
- Just prior to fail-over increase the mirroring to high-safety mode (sync)
- Fail-over to the source instance
- Set the jobs enabled on the destination that were enabled on the source instance
- Break the mirroring

In actuality it's a three-stage process: 

1. Prepare the destination instance (everything synchronised but no actual data moved)
2. Mirror the databases from the source onto there
3. Do a controlled fully synchronised fail-over 

We also move replication at the same time but let's not get too giddy, plenty of time for that later.

I think a good starter would be the schema-only path through Migrate Instance. This way, people could test the whole thing without having to do any backups or detaching. This would then lay the foundation for putting the mirroring code in which would be very cool.

Let me know what you think and I'll get cracking. Oh and let me know if you want me to submit the Copy-SqlDatabaseMail changes.

James

ctrlbold:
Hey James,
This is awesome, thank you! I'll take a look after this webinar. I also
sent you an invite to the Trello board https://dbatools.io/trello because
you'd be awesome on there!

ctrlbold:
Ahh, sorry, long week. I would *love* if you would contribute. I am releasing a 0.8 on Monday and have added a bunch of stuff. Please ensure you sync before attempting a pull request. I'm thinking, though, that Mirroring would be a new Function called Invoke-SqlDatabaseMirror or should it just be in Copy-SqlDatabase with a -Mirror param? Also, feel free to hit me up on Gchat if you're there -- clemaire@gmail.com

ctrlbold:
Also, if you don't have the code written already, someone sent me this his additions to a really old version of Start-SqlMigration that I was going to pick apart. [Code attached here](https://trello.com/c/Z62lTVAz/74-migration-via-mirroring)

PR #22, opened 2016-06-02
Update Reset-SqlAdmin.ps1

wabby80:
DBAs or whoever who is using this script should be informed that this will restart the instance and that resetting a user in production might give consequences.

### DIFF
--- a/functions/Reset-SqlAdmin.ps1
+++ b/functions/Reset-SqlAdmin.ps1
@@ -142,7 +142,18 @@ Internal function.
 	
 	PROCESS
 	{
-		
+	        # Before we continue, we need confirmation.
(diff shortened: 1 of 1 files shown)


dsolodow:
What about using SupportShouldProcess and 'ConfirmImpact = High'?
This would allow WhatIf support as well as being more in keeping with typical cmdlet behavior?

wabby80:
Good point, I've never used SupportShouldProcess, but it would make more sense using it I guess. The most important part here is that I want to inform the enduser that this will affect the whole instance. 
I'll rewrite it using this instead.

ctrlbold:
(Sorry for the delay, I've been on vacation.) 

Nice work! I'm going to merge the pull request then add in -Force support. Generally, people use -Confirm:$false to just force no prompting but I aim for the tool to be a bit of an intro to people and that may be too weird of syntax.

PR #31, opened 2016-07-05
Add begin block and server connection code

dansqldba:
Added in missing server connection code, missing begin block. Moved code
better suited to begin block from top of the process block.


ctrlbold:
Awesome! Thanks for catching that.

ctrlbold:
LOL my bad, dude

PR #35, opened 2016-07-08
Added logic for Simple Recovery and No full backup

js0505:
Also implemented @ctrlb code review changes.


ctrlbold:
damn, i thought so, conflicts with claudio's commit which is a ton of new code. Please rewrite with his changes as a base :|

PR #73, opened 2016-07-22
Fixes an issue with multiple disks being returned from a machine Iden…

kilasuit:
…tified in the PowerShell Slack channel via boudreaux

### DIFF
--- a/functions/Get-DiskSpace.ps1
+++ b/functions/Get-DiskSpace.ps1
@@ -111,7 +111,7 @@ Returns a custom object filled with information for server1, server2 and server3
             
             if ($data.Count -gt 1) 
             {
-                $data.GetEnumerator() | Foreach {  $null = $collection.Add($_) }
+                $data.GetEnumerator() | ForEach-Object {  $null = $collection.Add($_) }
             }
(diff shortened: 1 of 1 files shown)


kilasuit:
Just to give the full background,

Without this change the original output when run in the console looks ok but if you were to do this 

```powershell
$servers = $env:computername,'192.168.0.10'
$data = New-Object System.Collections.ArrayList
foreach ($server in $servers) {
    $drive = (Get-DiskSpace -ComputerName $server)
    $null = $data.add($Drive)
    }
```
then the $data variable doesn't end up holding an Object per disk but 1 per Server which when you attempt to pipe it to Select-Object or Sort-Object means that it can't deal with the underlying object properties and can occasionally cause an error especially if you pipe it to Select-Object -ExpandProperty Name

So this fix will check if there is more than 1 disk object that is returned by the Get-AllDiskSpace inner function (I'm not a fan of doing this btw but thats another story) then it will correctly add all of the individual disk objects to the final output from the Get-DiskSpace.

ctrlbold:
tks, baw!

PR #86, opened 2016-07-29
added check for C:\

ctrlbold:


dansqldba:
A while ago on Slack you said you were replacing the C: check with $env:SystemDrive but I don't see that in this commit?

ctrlbold:
Correct, it's just on my todo list on Trello. To come later.

ctrlbold:
bad file name. resubmitting with proper names and not mixing files. So Test-Allocation will be in another one.

PR #88, opened 2016-07-30
Removed line breaks; fixed some spelling errors

doculu:


dansqldba:
Only thing I'm not sure about here is the Unicode character added to Expand-SqlTLogResponsibly. The original file looks to be saved as ANSI, whereas most of the rest of the modules files are UTF-8.

PR #100, opened 2016-08-04
implemented claudio's change

ctrlbold:


dansqldba:
I can't get this to work, on servers where I've confirmed the tempdb is on C, this is always showing as IsBestPractice = true and current/recommended are false for the location check. I've tried this on the released version of the cmdlet and get the same results.

dansqldba:
Looking for a volunteer to test this cmdlet, as it works fine for @ctrlbold but is broken for me!

dansqldba:
CK and @MikeFal were on the case and Mike worked out what was going on with this. Something to do with no array being returned if your tempdb only had one file and log file on the C drive. As much as it pains me to close a pull request from you on your own project (lol) that seems like the more sensible option, as Mike has included a fix for testing in PR #107 .

ctrlbold:
Lol! This is exactly what I wanted! You guys are beasts 😈😈😈

PR #104, opened 2016-08-08
Initial Commit

ConstantineK:
Initial Commit for Test-DbaDiskAlignment, hopefully looking good!

### DIFF
--- a/functions/Test-DbaDiskAlignment.ps1
+++ b/functions/Test-DbaDiskAlignment.ps1
@@ -1,0 +1,359 @@
+Function Test-DbaDiskAlignment
+{
+<#
+.SYNOPSIS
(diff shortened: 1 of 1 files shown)


dansqldba:
There's one thing which keeps confusing me about this cmdlet. You refer to "Allocation" and not being able to determine this information accurately, but that isn't the case. You can't determine the "stripe unit size" and that's what you're testing for, right? The output also refers to "TestingAllocationUnitSizeKB" but should that be "TestingStripeUnitSizeKB" instead? Give me a shout on Slack if you want to educate me, I may just be misunderstanding this cmdlet completely!

ConstantineK:
Closing and making new PR as I have no idea if I can use a new commit as the basis for the PR in this interface.

PR #116, opened 2016-08-09
Implement Exclude into Copy-SqlJob

wsmelton:
Added "-Exclude" parameter into the Copy-SqlJob cmdlet.


dansqldba:
@ctrlbold and @wsmelton looking into the issues with tabs/spaces resulting in hundreds of modified lines, likely to be reformatted and committed again soon. Exclude functionality tested.

PR #122, opened 2016-08-11
New function! Get-SqlClusterActiveNode :fireworks:

ctrlbold:


dansqldba:
Great cmdlet, extremely useful! Tested with Windows Server 2012 R2/SQL 2012 cluster, Windows Server 2008 R2/SQL 2008 (none R2) cluster. These are both two node clusters, so might need some more testing by people with clusters with many more nodes!

PR #129, opened 2016-08-13
updated param document. should be plural (tho both work)

ctrlbold:


ctrlbold:
oh oops, i did two unrelated PRs in one branch. my bad.

PR #131, opened 2016-08-14
added support for multiple servers

ctrlbold:
PS C:\github\dbatools> $servers = Get-SqlRegisteredServerName sqlserver
PS C:\github\dbatools> Install-SqlWhoIsActive -SqlServer $servers
-Database master
Finished installing/updating sp_WhoIsActive in master on sql2005
Finished installing/updating sp_WhoIsActive in master on
sql2005express\sqlexpress
Finished installing/updating sp_WhoIsActive in master on sql2008
Finished installing/updating sp_WhoIsActive in master on sql2008\sql2k8
Finished installing/updating sp_WhoIsActive in master on sql2012

### DIFF
--- a/functions/Install-SqlWhoIsActive.ps1
+++ b/functions/Install-SqlWhoIsActive.ps1
@@ -54,29 +54,28 @@ Install-SqlWhoIsActive -SqlServer sqlserver2014a -SqlCredential $cred
 
 Pops up a dialog box asking which database on sqlserver2014a you want to install the proc to. Logs into SQL Server using SQL Authentication.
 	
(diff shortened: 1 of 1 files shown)


dansqldba:
Works a treat, unless one of the SQL instances is offline or has connectivity issues etc. I wrapped a bit of Test-SqlConnection around it, worked fine, but should some of that logic live within the cmdlet? Depends how easy you're trying to make it for people. Additionally, the install fails on SQL 2000, so might be worth adding a check to make sure its 2005 and above before trying.

PR #134, opened 2016-08-15
Added Set- and Test-DbaJobOwner functions.

MikeFal:


dansqldba:
Appears to have support for -Detailed parameter in Test-DbaJobOwner but this isn't a valid parameter. Set-DbaJobOwner passes validation for Windows groups, but they can't be job owner, suggest additional validation checks for Windows groups.

PR #158, opened 2016-08-18
Updates to DbaDatabaseOwner functions

MikeFal:
Added logic to DbaDatabaseOwner functions to exclude any databases that are not online to address issue https://github.com/ctrlbold/dbatools/issues/156.
Added check in Set-DbaDatabase owner for databases in ReadOnly mode. If not updateable, a warning will be thrown and database skipped.

### DIFF
--- a/functions/Set-DbaDatabaseOwner.ps1
+++ b/functions/Set-DbaDatabaseOwner.ps1
@@ -132,12 +132,21 @@ Sets database owner to 'sa' on the db1 and db2 databases if their current owner 
 					{
 						Write-Output "Setting database owner for $dbname to $TargetLogin on $servername"
 						# Set database owner to $TargetLogin (default 'sa')
-                        # Database is ReadOnly/Not Updateable, print warning and skip.
+						# Ownership validations checks
+                        
+						#Database is online and accessible 
                         if($db.Status -ne 'Normal'){
-                            Write-Warning "$dbname on $servername is in a  $($db.Sate) state and can not be altered. It will be skipped."
-                        } elseif ($db.IsUpdateable -eq $false) {
+                            Write-Warning "$dbname on $servername is in a  $($db.Sate) state and can not be altered. It will be skipped."						 
+                        } 
+						#Database is updatable, not read-only
+						elseif ($db.IsUpdateable -eq $false) {
 							Write-Warning "$dbname on $servername is not in an updateable state and can not be altered. It will be skipped."
-						} else {
+						} 
+						#Is the login mapped as a user? Logins already mapped in the database can not be the owner
+						elseif ($db.Users.name -contains $TargetLogin) {
+							Write-Warning "$dbname on $servername has $TargetLogin as a mapped user. Mapped users can not be database owners."
+						}
+						else {
                             $db.SetOwner($TargetLogin)
                         }
 					}
--- a/functions/Set-DbaDatabaseOwner.ps1
+++ b/functions/Set-DbaDatabaseOwner.ps1
@@ -107,9 +107,9 @@ Sets database owner to 'sa' on the db1 and db2 databases if their current owner 
 			#Otherwise, use all databases on the instance where owner not equal to -TargetLogin
 			Write-Verbose "Gathering databases to update"
 			
-            #use online/available dbs where owner and target login do not match
+            #use where owner and target login do not match
             #exclude system dbs
-            $dbs = $server.Databases | Where-Object {$_.Status -eq 'Normal' -and $_.Owner -ne $TargetLogin -and @('master', 'model', 'msdb', 'tempdb', 'distribution') -notcontains $_.Name}
+            $dbs = $server.Databases | Where-Object {$_.Owner -ne $TargetLogin -and @('master', 'model', 'msdb', 'tempdb', 'distribution') -notcontains $_.Name}
 
             #filter collection based on -Databases/-Exclude parameters
 			if ($Databases.Length -gt 0)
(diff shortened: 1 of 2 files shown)


functions/Set-DbaDatabaseOwner.ps1
```diff
@@ -107,24 +107,21 @@ Sets database owner to 'sa' on the db1 and db2 databases if their current owner
 			#Otherwise, use all databases on the instance where owner not equal to -TargetLogin
 			Write-Verbose "Gathering databases to update"
 			
+            #use online/available dbs where owner and target login do not match
+            #exclude system dbs
+            $dbs = $server.Databases | Where-Object {$_.Status -eq 'Normal' -and $_.Owner -ne $TargetLogin -and @('master', 'model', 'msdb', 'tempdb', 'distribution') -notcontains $_.Name}
```
dansqldba: As you're filtering out all databases which aren't in a "normal" state, the user is never warned about the state because it's never attempted. For example, if you try to update a single database that's in read only, the cmdlet just says "Nothing to do!" and no warnings/errors are presented.

functions/Test-DbaDatabaseOwner.ps1
```diff
@@ -109,14 +109,13 @@ that TargetLogin must be a valid security principal that exists on the target se
 					Continue
 				}
 			}
-			
+			#use online/available dbs
+            $dbs = $server.Databases | Where-Object {$_.Status -eq 'Normal'}
```
dansqldba: This is also filtering out databases which are not in a normal state, not sure myself whether this is desirable. As a user, I might expect to see those databases when I do -Detailed, for example?

dansqldba:
Hi Mike, will you be including a check for existing mapped logins for the new owner causing the cmdlet to fail, described in #158, or will that be a separate PR?

MikeFal:
I'm not clear what you mean by a check for existing mapped logins. I don't see an open issue 158 and this is PR 158. Can you clarify?

functions/Set-DbaDatabaseOwner.ps1
MikeFal: So would we prefer a warning/error if the database can't be updated instead of filtering out offline/restoring/suspect databases?

ctrlbold: Yeah, I'd agree. I just realized I do the same thing with Copy-SqlDatabase. I warn them that it's skipped.

In addition, if something throws an error I just Write-Exception then Continue. I don't throw any stopping exceptions.

dansqldba:
Sorry Mike getting myself in a twist, issue is #156. If you try to set the owner as a login which is already mapped to the database, it throws an error. To reproduce, create a sql login, map it to the database, then try to set that login as owner.

PS C:\WINDOWS\system32> Set-DbaDatabaseOwner -SqlServer testserver -Databases dbareports -TargetLogin testlogin
Setting database owner for dbareports to testlogin on testserver
WARNING: See error log C:\Users\DanielAlexander\Documents\dbatools-exceptions.txt for more details.
Exception calling "SetOwner" with "1" argument(s): "Set owner failed for Database 'dbareports'. "
At C:\Users\DanielAlexander\My Documents\WindowsPowerShell\Modules\dbatools\functions\Set-DbaDatabaseOwner.ps1:139 char:29
+                             $db.SetOwner($TargetLogin)
+                             ~~~~~~~~~~~~~~~~~~~~~~~~~~
    + FullyQualifiedErrorId : FailedOperationException

functions/Test-DbaDatabaseOwner.ps1
ctrlbold: Ditto with the Copy-Database thing where it says if its skipped. Sorry, Mike asked me yesterday if we'd like to implement an internal get-cleandatabase sort of script and I excitedly said yes not thinking that it's better to see what is skipped and perhaps why.

functions/Set-DbaDatabaseOwner.ps1
MikeFal: Ok, just to be clear, is the problem with the current version is that it throws a stopping exception when it encounters a db it can't change the owner on?

MikeFal:
Added updates to the PR:

- Revoked check for Normal status in Test-DbaDatabaseOwner. Now the output result shows database state for additional info.
- Altered Set-DbaDatabaseOwner to include databases not in normal state, but now it will throw a warning message with the database state if trying to alter a database not in normal state.

Let me know if this handles the issues/concerns. I'll look at the mapped user issue next. Depending on timing I could get it in to this PR, but don't let it stop you from merging this if I don't include it.

functions/Test-DbaDatabaseOwner.ps1
```diff
@@ -131,6 +130,7 @@ that TargetLogin must be a valid security principal that exists on the target se
 				$row = [ordered]@{
 					Server = $server.Name
 					Database = $db.Name
+					DBState = $db.State
```
dansqldba: State is always "Existing" for me, should this be "Status" instead?

MikeFal: Shoot, yes. Thanks. I'll get that fixed.

functions/Set-DbaDatabaseOwner.ps1
```diff
@@ -135,13 +132,21 @@ Sets database owner to 'sa' on the db1 and db2 databases if their current owner
 					{
 						Write-Output "Setting database owner for $dbname to $TargetLogin on $servername"
 						# Set database owner to $TargetLogin (default 'sa')
-						$db.SetOwner($TargetLogin)
+                        # Database is ReadOnly/Not Updateable, print warning and skip.
+                        if($db.Status -ne 'Normal'){
+                            Write-Warning "$dbname on $servername is in a  $($db.Sate) state and can not be altered. It will be skipped."
```
dansqldba: Typo in the warning, also might be Status instead of State?

dansqldba:
Leaving issue #156 open as there's a bit more discussion in there, but a bunch of checks discussed so far are now added.

PR #159, opened 2016-08-20
Bugfix resetsqlsa localhost

ctrlbold:


ctrlbold:
Gotta approve quick cuz someone is doing a presentation in 50 mins. I thought it worked across the board on localhost but it seems not. Made changes to not use -Computername . when using ., $env;ComputerName or localhost. Just straight Get-Service

PR #163, opened 2016-08-24
initial commit of Test-DbaDatabaseCollation

ctrlbold:


dansqldba:
I don't think you need to restrict the output if dealing with only one database here. One test I performed was to compare the collation across two SCOM servers and the OperationsManager database. As it's only one database, I only see Server, IsEqual. Collation is important to SCOM so when I specify -Detailed, I'm hoping to see the collation mentioned, even if its just one database. I appreciate this is probably a very niche test, I just think detailed should always mean detailed ;)

ctrlbold:
Totally true, that was not the intent. Fixing it now :)

PR #165, opened 2016-08-25
initial commit of Test-DbaServerName and Repair-DbaServerName

ctrlbold:
I tested replication (and only to the degree that I know how it works) but not mirror breakage.

### DIFF
--- a/functions/Repair-DbaServerName.ps1
+++ b/functions/Repair-DbaServerName.ps1
@@ -97,11 +97,11 @@ Skips some prompts/confirms but not all of them.
 				{
 					# If we ever decide with a -Force to support a cluster name change
 					# We would compare $server.NetName, and never ComputerNamePhysicalNetBIOS
-					throw "$servername is a cluster. Microsoft does not support renaming clusters. Not messing with that."
+					throw "$servername is a cluster. Microsoft does not support renaming clusters."
 				}
 				else
 				{
-					Write-Warning "$servername is a cluster. Microsoft does not support renaming clusters. Not messing with that."
+					Write-Warning "$servername is a cluster. Microsoft does not support renaming clusters."
 					Continue
 				}
 			}
--- a/functions/Repair-DbaServerName.ps1
+++ b/functions/Repair-DbaServerName.ps1
@@ -97,11 +97,11 @@ Skips some prompts/confirms but not all of them.
 				{
 					# If we ever decide with a -Force to support a cluster name change
 					# We would compare $server.NetName, and never ComputerNamePhysicalNetBIOS
-					throw "$servername is a cluster. Not messing with that."
+					throw "$servername is a cluster. Microsoft does not support renaming clusters. Not messing with that."
 				}
 				else
 				{
-					Write-Warning "$servername is a cluster. Not messing with that."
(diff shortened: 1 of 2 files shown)


ctrlbold:
I don't believe this supports named instances. Test may not either. It needs to. It can't change the \named  name, but can change the server name. This should be done before it's finished (and has to be done before my presentation in early oct. personal goal.)

ctrlbold:
Oh snap, Test-DbaServerName does support named instances! Cool, onto testing Repair-DbaServerName with an instance.

ctrlbold:
Repair support named instances, too. I'm dumm for thinking I was dumm. Testing now.

functions/Repair-DbaServerName.ps1
```diff
@@ -0,0 +1,316 @@
...
+							$options = [System.Management.Automation.Host.ChoiceDescription[]]($yes, $no)
+							$result = $host.ui.PromptForChoice($title, $message, $options, 1)
+							
+							if ($result -eq 1)
+							{
+								Write-Output "Okay, moving on."
+							}
+							else
+							{
+								Write-Output "Removing Mirroring"
```
dansqldba: Write-Warning more appropriate? Shouldn't this support WhatIf?

ctrlbold: Yes indeed. Fixed.

functions/Test-DbaServerName.ps1
```diff
@@ -0,0 +1,195 @@
...
+				$mirroreddb = $server.Databases | Where-Object { $_.IsMirroringEnabled -eq $true }
+				
+				if ($mirroreddb.count -gt 0)
+				{
+					$dbs = $mirroreddb.name -join ", "
+					$reasons += "Databases are being mirrored: $dbs"
+				}
+				
+				# check for replication
+				$sql = "select name from sys.databases where is_published = 1 or is_subscribed =1 or is_distributor = 1"
```
dansqldba: This T-SQL fails on SQL 2000, throws an error on screen. Had to add my own Write-Verbose to find out what was causing the error.

ctrlbold: Good call. Man, I dont even wanna support SQL 2000. Ill make it skip.

functions/Test-DbaServerName.ps1
```diff
@@ -0,0 +1,195 @@
...
+				}
+				catch
+				{
+					if ($NoWarnings -eq $false)
+					{
+						Write-Warning "Can't contact $($server.ComputerNamePhysicalNetBIOS) using Get-Service. This means the script will not be able to automatically restart SQL services."
+					}
+				}
+				
+				if ($rs.count -gt 0)
```
dansqldba: Something funky is going on at this line which causes bogus warnings about needing to restart SSRS, even on servers where SSRS is not installed.

dansqldba: Something funky is going on at this line which causes bogus warnings about needing to restart SSRS, even on servers where SSRS is not installed.

dansqldba:
Making $rs null before checking for SSRS services fixed the issues I was having with servers showing they had SSRS when they didn't. Yay!

PR #173, opened 2016-09-02
initial commit of Install-FirstResponderKit

Splaxi:
Didn't include it in the psd1 file.


dansqldba:
Pending permission from Brent Ozar Unlimited, all clear from @ctrlbold.

PR #174, opened 2016-09-06
initial commit of Test-DbaDatabaseBestPractice

Splaxi:
Did not include the function in the psd1


functions/Test-DbaDatabseBestPractice
```diff
@@ -0,0 +1,305 @@
...
+	
+#>
+	[CmdletBinding(SupportsShouldProcess = $true)]
+	Param (
+		[parameter(Mandatory = $true, ValueFromPipeline = $true)]
+		[Alias("ServerInstance", "SqlInstance")]
+		[string[]]$SqlServer,
+		[object]$SqlCredential,
+		[switch]$Detailed,
+		[switch]$IncludeSystemDbs #added this, saw you all talking about it
```
dansqldba: This shows up as help, there's no dedicated .PARAMETER help description for this switch, consider adding.

functions/Test-DbaDatabseBestPractice
```diff
@@ -0,0 +1,305 @@
...
+.DESCRIPTION
+Returns Database name and $true or $false by default for one server. Returns Server name, Database name and IsBestPractice for more than one server.
+	
+Specify -Detailed for details.
+	
+References:
+https://technet.microsoft.com/en-us/library/dd758814(v=sql.100).aspx - "The performance question here is usually not one of correlation per the formula, but whether the cluster size ..has been explicitly defined at 64 KB, which is a best practice for SQL Server."
+http://tk.azurewebsites.net/2012/08/
+	
+.PARAMETER ComputerName
```
dansqldba: This should probably be .PARAMETER SqlServer ?

functions/Test-DbaDatabseBestPractice
```diff
@@ -0,0 +1,305 @@
+Function Test-DbaDatabaseBestPractice
```
dansqldba: There's a typo in the file name! Cmdlet name is right though.

functions/Test-DbaDatabseBestPractice
```diff
@@ -0,0 +1,305 @@
...
+		}
+
+		$collection = New-Object System.Collections.ArrayList
+	}
+	
+	PROCESS
+	{
+        $processed = New-Object System.Collections.ArrayList
+
+		foreach ($server in $sqlserver)
```
dansqldba: Doesn't accept multiple servers at the command line (creates a single server name from the whole list and tries to connect, failing each time. For example, server1,server2,server3 becomes server1 server2 server3 and results in the following error being displayed three times:

WARNING: Can't connect to server1 server2 server3 or access denied. Skipping.

functions/Test-DbaDatabseBestPractice
```diff
@@ -0,0 +1,305 @@
...
+					AutoGrowthDetails = $autogrowthbps
+				}
+				$null = $collection.Add($object)
+			}
+		}
+	}
+	
+	END
+	{
+		if ($Detailed -eq $true)
```
dansqldba: The detailed output results in file autogrowth details causing other stuff to be chopped off the screen, only visible if you output to grid etc. Consider setting the order of the output to Server, Database, IsBestPractice[n] and putting filegrowth detail at the end. This is similar to the order mentioned in the description.

functions/Test-DbaDatabseBestPractice
```diff
@@ -0,0 +1,305 @@
...
+
+To return true or false for databases db1 and db2 not being compliant with the best practice
+
+.EXAMPLE   
+Test-DbaDatabaseBestPractice -SqlServer sqlserver2014 -Detailed
+	
+To return detailed information about databases and their best practice compliance 
+	
+#>
+	[CmdletBinding(SupportsShouldProcess = $true)]
```
wsmelton: Since you are not making changes in this cmdlet, you can remove the SupportsShouldProcess. This is only needed I believe to support "-WhatIf" and "-Confirm", which is not needed for commands that are just reading and outputting information.

functions/Test-DbaDatabseBestPractice
```diff
@@ -0,0 +1,305 @@
...
+
+			Write-Verbose "Attempting to connect to $sqlserver"
+			
+			try
+			{
+				$server = Connect-SqlServer -SqlServer $sqlserver -SqlCredential $SqlCredential
+			}
+			catch
+			{
+				Write-Warning "Can't connect to $sqlserver or access denied. Skipping."
```
wsmelton: You don't need this this try/catch because `Connect-SqlServer` includes try/catch. On another note you should provide the exception output `Write-Exception` for try/catch blocks to output the actual reason why it failed or error was caught.

functions/Test-DbaDatabseBestPractice
```diff
@@ -0,0 +1,305 @@
...
+                    {
+                        $localtemp = $properties."$prop";
+                        $IsBestPractice = $localtemp -and $IsBestPractice;
+                    }
+
+					$newcollection += [PSCustomObject]@{
+						Server = $server
+						Database = $dbInner.Database
+						IsBestPractice = $IsBestPractice
+					}
```
wsmelton: I would ensure you are being consistent in your code. In places you are using `New-Object PSObject -Property @{}` to build a collection, where in these lines you utilize `[PSCustomerObject]`. The PSCustomeObject is preferred (by me at least).

functions/Test-DbaDatabseBestPractice
```diff
@@ -0,0 +1,305 @@
...
+			$List = $newcollection | Select-Object Server -Unique
+			
+			if ($List.Server.Count -eq 1)
+			{
+				return $newcollection | Sort-Object Database, SortOrder | Select-Object Database, IsBestPractice
+			}
+			else
+			{
+				return $newcollection | Sort-Object Server, Database, SortOrder 
+			}
```
wsmelton: This might be up for discussion on how we want to standardize outputting data. I know some have used a datatable to put the information so it can be more easily sent to another source (database for history, etc.). As well the manner used here shows the database at the bottom of the list, would think it is best to have the database name show first.

wsmelton: Main reason for the review here is when you output the AutoGrowth details, folks may not know exactly how to expand those details where they are readable.  Reason we may just need to discuss making a standard, or even building a format file to use with the module. https://msdn.microsoft.com/en-us/powershell/scripting/getting-started/cookbooks/using-format-commands-to-change-output-view

functions/Test-DbaDatabseBestPractice
dansqldba: I believe this is part of the template and likely found all over the module.

wsmelton: It is not part of the [templates linked on dbatools.io guideline page](https://dbatools.io/join-us/guidelines/), unless there are templates found elsewhere?

dansqldba: I was referring to https://github.com/sqlcollaborative/dbatools-templates/blob/master/Test-SqlBestPractice.ps1 so I think those links on the dbatools.io site might be out of date. Have made @ctrlbold aware.

wsmelton: Either way, the current code is masking the actual error. Don't load SMO, try to run the cmdlet and the same message is output. You should not mask the errors unless you are going to check for every possible cause so you can output a clear reason.

dansqldba: Feel free to log a new issue if you feel this is a problem. It's in the templates created by the project owner and @Splaxi is likely to have used those templates. This will not block the release of this cmdlet.

functions/Test-DbaDatabseBestPractice
```diff
@@ -0,0 +1,305 @@
...
+				if ($falsecount.count -gt 0)
+				{
+					$IsBestPracticeAutoGrowth = $false
+				}
+				else
+				{
+					$IsBestPracticeAutoGrowth = $true
+				}
+				
+				$object = New-Object PSObject -Property @{
```
ctrlbold: Please use [pscustomobjet] across the board instead. See [PowerShell v3 – Creating Objects With pscustomobject – it’s fast!](http://www.jonathanmedd.net/2011/09/powershell-v3-creating-objects-with-pscustomobject-its-fast.html) for performance comparisons.

functions/Test-DbaDatabseBestPractice
ctrlbold: Agreed, and its fastest by far.

Splaxi:
We are going to rewrite stuff and split it into ATOMS! Closing this.

PR #208, opened 2016-09-18
Get-DbaRestoreHistory - fixed output, updated docs

ctrlbold:


dansqldba:
Default output much improved. 

Verbose output shows the SQL used, which is quite overwhelming when dealing with multiple servers, perhaps should be Write-Debug output instead? I had to change it during testing to assess performance:

Write-Verbose "Retrieving information from $server"

Performance is still an issue for this cmdlet for me, although it works fine with multiple servers. If I try the CMS example I have problems. I've opened issue #212 to look at the performance again and do some more testing.

PR #210, opened 2016-09-19
Get-DbaTcpPort - Fixes

ctrlbold:


dansqldba:
The verbose output does not mention the instance name the cmdlet is dealing with. This results in output similar to the following:

VERBOSE: Local connect worked
VERBOSE: Parsing information for 6 IP addresses
VERBOSE: Local connect worked
VERBOSE: Parsing information for 10 IP addresses
VERBOSE: Local connect worked
VERBOSE: Parsing information for 6 IP addresses

ctrlbold:
Done. There are times when it wont say "Parsing X IP addresses" and I'm okay with that cuz I'm too lazy to figure out why, but it seems like verbose output is getting eaten by the remote server.

ctrlbold:
Thanks! Please do not delete this branch. It has a child branch for Set and I don't know if it's dependent.

Issue #216, opened 2016-09-21
Install-SqlWhoIsActive - Database 'Downloading sp_WhoIsActive zip file, unzipping and installing. master' does not exist


dwierenga:
When running a SQL 2012 migration from one clustered instance to another, I got this message in the `dbatools-exceptions.txt` logfile.  It does not seem fatal, but it does seem like a very odd message.

```
9/21/2016 12:05:21 PM
System.Management.Automation.MethodInvocationException: Exception calling "ChangeDatabase" with "1" argument(s): "Database 'Downloading sp_WhoIsActive zip file, unzipping and installing. master' does not exist. Make sure that the name is entered correctly." ---> System.Data.SqlClient.SqlException: Database 'Downloading sp_WhoIsActive zip file, unzipping and installing. master' does not exist. Make sure that the name is entered correctly.
   at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, Boolean& dataReady)
   at System.Data.SqlClient.TdsParser.Run(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj)
   at System.Data.SqlClient.SqlInternalConnectionTds.ChangeDatabaseInternal(String database)
   at System.Data.SqlClient.SqlInternalConnection.ChangeDatabase(String database)
   at System.Data.SqlClient.SqlConnection.ChangeDatabase(String database)
   at CallSite.Target(Closure , CallSite , Object , Object )
   --- End of inner exception stack trace ---
   at System.Management.Automation.ExceptionHandlingOps.CheckActionPreference(FunctionContext funcContext, Exception exception)
   at System.Management.Automation.Interpreter.ActionCallInstruction`2.Run(InterpretedFrame frame)
   at System.Management.Automation.Interpreter.EnterTryCatchFinallyInstruction.Run(InterpretedFrame frame)
[...]

PR #218, opened 2016-09-22
Install-SqlWhoIsActive, Show-SqlWhoIsActive - Bug fixes

ctrlbold:
To test, ensure that who_is_active_v11_11.sql does not exist in $env:temp. Also remove the sproc from the destination SQL Server if it exists.
Run

Show-SqlWhoIsActive -SqlServer sql2005 -ShowSystemSpids

Select the destination database. It should auto install and an OGV should pop up with the info.


dansqldba:
Fixes #216

PR #220, opened 2016-09-22
updated version, added command names

ctrlbold:


ctrlbold:
That's more like it 💃

PR #221, opened 2016-09-22
fix spelling

ctrlbold:


ctrlbold:
Mm where did my comment go? Anyway, I can't spell compatibility, so I fixed it. Will push through, did all the tests.

PR #223, opened 2016-09-26
Update-SqlPermissions, Sync-SqlLoginPermissions - Fixes for inaccessible databases

### LINKED ISSUE #222: Copy-SqlLogin and Database in Availability Group
Testing in lab to implement dbatools and desire was to only copy over logins that were new, so no use of the `-Force` parameter.

I have not looked in detail but there is shared functions erring when trying to create a login that was added to a primary replica on secondaries. Just based on errors it is from the code trying to apply the permissions in the database, which is not going to be allowed if the database is involved in an AG. First guess there just needs to be a check to make sure the database is accessible, like `IsAccessible` or something before trying to apply the permission on the destination.

In the end the login is created, it just errors trying to apply permissions. In the case of an AG, the permissions will be picked up once a failover to that secondary occurs because the login is copied with the same SID.

Exception output log:
[dbatools-exceptions.txt](https://github.com/sqlcollaborative/dbatools/files/491493/dbatools-exceptions.txt)

wsmelton:
Adjust sharedfunctions and update-sqlpermissions to account for databases not being accessible, but not causing issues during migrations.

Fixes #222

### DIFF
--- a/functions/SharedFunctions.ps1
+++ b/functions/SharedFunctions.ps1
@@ -1500,7 +1500,7 @@ Function Update-SqlPermissions
 		{
 			if (!$destdb.IsAccessible)
 			{
-				Write-Output "Database is not accessible. Skipping"
+				Write-Output "Database [$($destdb.Name)] is not accessible. Skipping"
 				Continue
 			}
 			if ($destdb.users[$dbusername] -eq $null)
--- a/functions/SharedFunctions.ps1
+++ b/functions/SharedFunctions.ps1
@@ -1498,6 +1498,11 @@ Function Update-SqlPermissions
 		
 		if ($destdb -ne $null)
 		{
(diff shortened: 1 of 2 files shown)


functions/SharedFunctions.ps1
```diff
@@ -1498,6 +1498,11 @@ Function Update-SqlPermissions
 		
 		if ($destdb -ne $null)
 		{
+			if (!$destdb.IsAccessible)
+			{
+				Write-Output "Database is not accessible. Skipping"
```
schwartzmx: thoughts on outputting which database?
`Write-Output "Database $destdb is not accessible. Skipping"`

dansqldba: Agreed, please add the name of the database to the output.

PR #232, opened 2016-09-28
Update Set-SQLTempDBConfiguration

### LINKED ISSUE #231: Set-SqlTempDbConfiguration compares total size of current tempdb to divided size of new db configuration instead of comparing total vs total
1. In [Set-SqlTempDbConfiguration](https://github.com/sqlcollaborative/dbatools/blob/master/functions/Set-SqlTempDBConfiguration.ps1) we use `$datafilesizemb` to determine what the final size for your new tempdb configuration should be.
2. We then determine that `$dataFilesizeSingleMB = $([Math]::Floor($datafilesizemb/$datafilecount))` (so that a single file size for our tempdb split will be a fraction of the total space we would like to use)
3. And finally, `$toobigcount = $server.Databases['tempdb'].ExecuteWithResults("SELECT count(1) as FileCount FROM sys.database_files WHERE size/128 > $dataFilesizeSingleMB AND type = 0").Tables[0].FileCount` (which asks, are there any files in the current tempdb configuration which are larger than your single divided file?)

I can see why we might want to avoid shrinking tempdb files but in the case of moving from one large file to several smaller files this would throw an error.

ConstantineK:
Example of possible changes to resolve #231 but I don't know if this is the final way we want to accomplish this. 

Others mentioned things like an -Expert flag or -Force, all this does check and is stop you if you would be altering your tempdb files to a filesize smaller than what is in use(instead of comparing max file sizes.)


functions/Set-SqlTempDBConfiguration.ps1
```diff
@@ -172,24 +172,17 @@ Returns PSObject representing tempdb configuration.
...
 		{
-			throw "Current tempdb not suitable to be reconfigured. The current tempdb is larger than the calculated configuration."
-		}
-		
-		$equalcount = $server.Databases['tempdb'].ExecuteWithResults("SELECT count(1) as FileCount FROM sys.database_files WHERE size/128 = $dataFilesizeSingleMB AND type = 0").Tables[0].FileCount
-		
-		if ($equalcount -gt 0)
-		{
-			throw "Current tempdb not suitable to be reconfigured. The current tempdb is the same size as the specified DataFileSizeMB."
+			throw "Current tempdb not suitable to be reconfigured. The current tempdb has at least one file which has greater than or equal to the amount of free space requested in all files."
```
ClaudioESSilva: The message is almost perfect, you say the cause but I think we need to give the solution. Here should be something like restart instance, shrink/grow files to the same size and it's ok.

ConstantineK:
Do you think "Consider rebooting your SQL Server to flush all space or quiescing transactions to free up enough space to allow this operation to complete." after the error message is too wordy?

ClaudioESSilva:
I think we must be more pratical. 
"Please verify if you can shrink the file(s) in this situation.
If you don't feel confortable with it, we suggest you restart you instance (maintenance window) and do the configuration after."

I'm thinking if I would like to see the -Force parameter used as receipt.
Will be the message before with command to run:
"We suggest you restart you instance (maintenance window), and run the following command to complete this action:
Set-SQLTempDBConfiguration -SqlServer <server> -Force"

We need to hear more opinions from our team mates :)

functions/Set-SqlTempDBConfiguration.ps1
```diff
@@ -172,24 +172,17 @@ Returns PSObject representing tempdb configuration.
...
 		{
-			throw "Current tempdb not suitable to be reconfigured. The current tempdb is larger than the calculated configuration."
-		}
-		
-		$equalcount = $server.Databases['tempdb'].ExecuteWithResults("SELECT count(1) as FileCount FROM sys.database_files WHERE size/128 = $dataFilesizeSingleMB AND type = 0").Tables[0].FileCount
-		
-		if ($equalcount -gt 0)
-		{
-			throw "Current tempdb not suitable to be reconfigured. The current tempdb is the same size as the specified DataFileSizeMB."
+			throw "Current tempdb not suitable to be reconfigured. Please verify if you can shrink the file(s) in this situation. If you don't feel confortable with it, we suggest you restart you instance (maintenance window) and perfom the configuration after."
```
schwartzmx: >Current tempdb not suitable to be reconfigured. Please verify if you can shrink the file(s) in this situation. If you don't feel *comfortable* with it, we suggest you restart *your* instance (maintenance window) and perfom the configuration after.

Otherwise, looks good to me.  I think the wording is appropriate 👍

wsmelton: I would not make it overly complicated, keep the message short and to the point. In regards to free space size though, it is easy enough to check if there are active transactions against tempdb, maybe include that in the output below the message. You may also reference this to get tempdb to shrink: https://www.brentozar.com/archive/2016/02/when-shrinking-tempdb-just-wont-shrink/

> tempdb is found in a state that does not allow shrinking. Verify active transactions or if a restart can be performed.

ConstantineK:
I have hit a bit of a blocker on this and would appreciate some advice on how to format output so that the correct message is communicated.

I changed the wording immediately, but I am trying to fulfill returning info about running transactions and space (and an error) wondering if there is a better way to display that or a more "idiomatic" way. 

https://gist.github.com/ConstantineK/148680811c210ae5c3e5739f8d3aaba5 has my example which (unfortunately) mixes write-warning, write-host, and throw; there must be a better way to inform the user but most of the resources I have come across dont really address this problem. 

The current script is written to just throw any errors, @Splaxi mentioned maybe just writing a custom object which reports at the end of the script, but it is a terminating error and I want to make sure the user is "properly informed".

ctrlbold:
@ConstantineK What about a custom object with the Notes field? We discussed this addition to other commands today. Unsure how to convey a terminating error, however.

ConstantineK:
I am just having an issue figuring out the best way to accomplish that without a much more major rewrite of the code; if the entire thing returned a custom object with success/failure it would work, but that would involve removing all the existing throws and changing a bit more than I was hoping to with this PR.

Would we be happy with just something like "You have x space taken and x running transactions in tempdb" + existing verbage/article reference?

ConstantineK:
Added changes so that this fixes #244 as well.

dansqldba:
@ctrlbold and @ConstantineK where are we at with this? I seem to have assigned it for testing but I'm not sure if you guys are done here?

ConstantineK:
Done with the first two fixes, just added a third, thinking about the auto-calculations now.

ctrlbold:
approved :D

PR #233, opened 2016-09-29
Get-DbaDiskSpace - Fix #190, added bytes as unit, filesystem, fragmentation and -Detailed

### LINKED ISSUE #190: Get-DbaDiskSpace - Output format issue
Get-DbaDiskSpace - Output format might give an error like "Cannot convert value "1.010.880,00" to type "System.Int64"" when try to set the value to an int/long variable. It is not really a bug is an improvement that can be made to the digit symbol group not be returned.

For example: 1.010.880,00 should be returned to: 1010880,00

Here I'm trying to set the value for an [long]$variable the FreeInKB from the Get-DbaDiskSpace.

Obviously I can replace the '.' or ',' for '' and work, but I think once this command can and should be reused should be more generic.
Thoughts?

Jagoop:
Added bytes as a unit
Removed thousand separator
Added Filesystem and fragmentation info
New -Detailed switch which also include previously left out volumes like \\\\?\\*


ctrlbold:
Ahh, these fixes and additions were so necessary. And now I know that "{0:n2}" -f  acts like Format-Table. Thanks, Jakob! These changes look great.

functions/Get-DbaDiskSpace.ps1
```diff
@@ -137,57 +175,57 @@ Returns a custom object filled with information for server1, server2 and server3
...
+						Label = $disk.Label
+						"SizeIn$unit" = $total
+						"FreeIn$unit" = $free
+						PercentFree = $percentfree
+						BlockSize = $disk.BlockSize
+                    }
+
+                    if ($CheckForSql -or $Detailed)
+                    {
+                        Add-Member -InputObject $diskinfo -MemberType Noteproperty IsSqlDisk -value $sqldisk
```
dansqldba: This column is added for detailed but only populated if -CheckForSql is included, is this intentional? Your final example suggests the -CheckForSql switch is not required for it to display a value in the -Detailed output.

functions/Get-DbaDiskSpace.ps1
```diff
@@ -137,57 +175,57 @@ Returns a custom object filled with information for server1, server2 and server3
...
+                    }
+
+                    if ($FileSystem -or $Detailed)
+                    {
+                        Add-Member -InputObject $diskinfo -MemberType Noteproperty FileSystem -value $disk.FileSystem
+                    }
+
+                    if ($CheckFragmentation)
+                    {
+                        Add-Member -InputObject $diskinfo -MemberType Noteproperty FileSystem -value $disk.FileSystem
```
dansqldba: Copy paste is not your friend today? :)

This line causes an error, as you're trying to add a member which already exists from the previous line. Should this be something like:

Add-Member -InputObject $diskinfo -MemberType Noteproperty Fragmentation -value $disk.FilePercentFragmentation

functions/Get-DbaDiskSpace.ps1
Jagoop: That is a bug. I will fix it.

Jagoop:
JUbiiiii!!!!

PR #234, opened 2016-09-29
initial commit of Get-DbaLastGoodCheckDb

Jagoop:
First attempt - feedback more than welcome!


functions/Get-DbaLastGoodCheckDb.ps1
```diff
@@ -0,0 +1,183 @@
+﻿Function Get-DbaLastGoodCheckDb
```
dansqldba: What makes it good or bad? The code suggests you're just looking for the datetime value of the last checkdb? Wouldn't that just be Get-DbaLastCheckDb?

functions/Get-DbaLastGoodCheckDb.ps1
```diff
@@ -0,0 +1,183 @@
...
+			{
+				$dbs = $dbs | Where-Object { $exclude -notcontains $_.Name }
+			}
+
+			foreach ($db in $dbs)
+			{
+				Write-Verbose "Processing $($db.name) on $servername"
+
+				$sql = "DBCC DBINFO ([$($db.name)]) WITH TABLERESULTS"
+				Write-Verbose "T-SQL: $sql"
```
dansqldba: Please use Write-Debug for outputting T-SQL

functions/Get-DbaLastGoodCheckDb.ps1
```diff
@@ -0,0 +1,183 @@
...
+.NOTES
+Copyright (C) 2016 Jakob Bindslet (jakob@bindslet.dk)
+
+This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+
+This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+.LINK
```
dansqldba: Maybe some links about CheckDB?

functions/Get-DbaLastGoodCheckDb.ps1
```diff
@@ -0,0 +1,183 @@
...
+<#
+.SYNOPSIS
+Get date/time for last known good DBCC CHECKDB
+
+.DESCRIPTION
+Retrieves and compares the date/time for the last known good DBCC CHECKDB, as well as the creation date/time for the database.
+
+Please note that this script uses the DBCC DBINFO() WITH TABLERESULTS. DBCC DBINFO has several known weak point, such as:
+ - DBCC DBINFO is an undocumented feature/command.
+ - The LastKnowGood timestamp is resat when a DBCC CHECKFILEGROUP is performed.
```
dansqldba: Typo

functions/Get-DbaLastGoodCheckDb.ps1
```diff
@@ -0,0 +1,183 @@
...
+
+			foreach ($db in $dbs)
+			{
+				Write-Verbose "Processing $($db.name) on $servername"
+
+				$sql = "DBCC DBINFO ([$($db.name)]) WITH TABLERESULTS"
+				Write-Verbose "T-SQL: $sql"
+
+				$resultTable = $db.ExecuteWithResults($sql).Tables[0]
+				[datetime]$lastKnownGood = ($resultTable | Where-Object Field -eq 'dbi_dbccLastKnownGood').Value
```
dansqldba: This has issues for me on some servers. It seems the field dbi_dbccLastKnownGood exists twice on some databases on some of my test servers. Is that normal? Either way, it causes the cmdlet to have issues assigning a value to this variable then causes other errors further down!

dansqldba:
Here is an example of the error when trying to add two values to [datetime]$lastKnownGood although it doesn't really tell you much...

Cannot convert the "System.Object[]" value of type "System.Object[]" to type "System.DateTime".
At C:\users\DanielAlexander\My Documents\WindowsPowerShell\Modules\dbatools\functions\Get-DbaLastGoodCheckDb.ps1:132 char:5
+ ...             [datetime]$lastKnownGood = ($resultTable | Where-Object F ...
    + CategoryInfo          : MetadataError: (:) [], ArgumentTransformationMetadataException
    + FullyQualifiedErrorId : RuntimeException

functions/Get-DbaLastGoodCheckDb.ps1
RGibsonNGPVAN: If you need both values: 
[DateTime[]]$LastKnownGood = ($resultTable | Where-Object Field -eq 'dbi_dbccLastKnownGood').Value

functions/Get-DbaLastGoodCheckDb.ps1
Jagoop: Good suggestion - I've added a few.

functions/Get-DbaLastGoodCheckDb.ps1
Jagoop: Done

functions/Get-DbaLastGoodCheckDb.ps1
Jagoop: This is very interesting - I ran my script against a quite a few databases, without getting into this dbi_dbccLastKnownGood x2 issue.
Was the DBs in question of some special kind - like upgraded from a very old SQL version, from Sharepoint or similar?

dansqldba:
Further testing reveals the field you're reading to get the information doesn't exist in SQL 2000, so you'll either need to skip that version, or account for the field not existing.

My advice would be to code for both, that way you're informing the user why it was skipped (there's examples in other cmdlets), but also trapping any other errors that may come up, rather than spitting out a red wall of text about time spans :+1:

functions/Get-DbaLastGoodCheckDb.ps1
dansqldba: I have no idea what the cause of this is, but you're right to highlight instance upgrades. Instances which I know to be brand new, never upgraded, are not exhibiting this behaviour. Other instances which are likely to have been in place upgraded are showing the value twice.

If I get time I might try and confirm this with a new VM.

dansqldba: I have no idea what the cause of this is, but you're right to highlight instance upgrades. Instances which I know to be brand new, never upgraded, are not exhibiting this behaviour. Other instances which are likely to have been in place upgraded are showing the value twice.

If I get time I might try and confirm this with a new VM.

dansqldba:
@Jagoop: The warning output does not mention the instance name, for example: 

WARNING: Get-DbaLastGoodCheckDb is only supported on SQL Server 2005 and above. Skipping Instance.
WARNING: The database DBName is not accessible. Skipping database.

Which instance was skipped? On which instance was the database DBName skipped?

dansqldba:
Hey @Jagoop we're not sure why this PR is now showing a merge conflict. As the branch is a fork in your own repo, we're also not sure we can do anything about it, so we might have to bump this into November unless you can clean it up today at some point, unless someone else can come to the rescue?

PR #242, opened 2016-10-06
Fix dbdetachedinfo

ctrlbold:
changed name, fixed for sander's commands, so steamrolling.


ctrlbold:
- [ ]  Tested on SQL 2000 and SQL 2016
- [ ]  Returns only an object
- [ ]  Throws error if the SQL Server can't access the path
- [ ]  Throws error if the db file is still attached and can't be read

PR #251, opened 2016-10-09
initial commit of Get-DbaAvailabilityGroup

wsmelton:
Initial commit


ctrlbold:
@ClaudioESSilva do you have an environment to support testing for this command? If so, would you be available to do so this week?

Here is the QA checklist

- [ ] Works remotely
- [ ] Works locally

ClaudioESSilva:
@ctrlbold I have and I am.

functions/Get-DbaAvailabilityGroup.ps1
```diff
@@ -0,0 +1,151 @@
...
+	{
+		
+		$server = Connect-SqlServer -SqlServer $sqlserver -SqlCredential $SqlCredential
+	}
+	
+	PROCESS
+	{
+		if (!$server.IsHadrEnabled)
+		{
+			Return "Availability Group is not configured."
```
ClaudioESSilva: Runned against SQL Server lower than 2012 (when AG feature appears) and the message is "Availability Group is not configured."

I suggest to validate server major version and if lower than 11 then say something like "Availability Groups is available only on SQL Server 2012 (v11) or higher versions. Server '$server' has version $($server.Major). Skipping." 

We can validate the edition, too. AG is only available on Enterprise Edition so if running on, for example, Standard Edition should say "Availability Groups feature is only available on Entrepise edition. Server '$Server' is running '$server.Edition'. Skipping."

functions/Get-DbaAvailabilityGroup.ps1
```diff
@@ -0,0 +1,151 @@
...
+#>
+	[CmdletBinding(SupportsShouldProcess = $true)]
+	Param (
+		[parameter(Mandatory = $true, ValueFromPipeline = $true)]
+		[Alias("ServerInstance", "SqlInstance")]
+		[object]$SqlServer,
+		[object]$SqlCredential,
+        [switch]$Detailed,
+        [string]$AvailabilityGroupName,
+		[switch]$IsPrimary
```
ClaudioESSilva: If this parameter needs the $AvailabilityGroupName and if we don't use dynamic parameters, we should set a PARAMETERSET which enforce the mandatory of both parameters.
If we will use dynamic parameter maybe we should validate this on the beginning of the Process instead after get all the info.

functions/Get-DbaAvailabilityGroup.ps1
```diff
@@ -0,0 +1,151 @@
...
+		[parameter(Mandatory = $true, ValueFromPipeline = $true)]
+		[Alias("ServerInstance", "SqlInstance")]
+		[object]$SqlServer,
+		[object]$SqlCredential,
+        [switch]$Detailed,
+        [string]$AvailabilityGroupName,
+		[switch]$IsPrimary        
+	)
+	
+#	DynamicParam { if ($sqlserver) { return Get-ParamSqlDatabases -SqlServer $sqlserver -SqlCredential $SqlCredential } }
```
wsmelton: Waiting to know what dynamic parameters are changing to... I don't really think this needs or should have dynamic parameters.

functions/Get-DbaAvailabilityGroup.ps1
wsmelton: Yeah, being that Connect-SqlServer was going to be updated to have the version flags, figured wait until that is done. Instead of duplicating code.

functions/Get-DbaAvailabilityGroup.ps1
```diff
@@ -0,0 +1,151 @@
...
+Returns true/false if the server, sqlserver2014a, is the primary replica for AG-a Availability Group
+#>
+	[CmdletBinding(SupportsShouldProcess = $true)]
+	Param (
+		[parameter(Mandatory = $true, ValueFromPipeline = $true)]
+		[Alias("ServerInstance", "SqlInstance")]
+		[object]$SqlServer,
+		[object]$SqlCredential,
+        [switch]$Detailed,
+        [string]$AvailabilityGroupName,
```
wsmelton: Until fix is in place where dynamic parameters actually show up in help, I don't use them. Considering this would likely be a parameter that will/should get used I don't care to have it hidden.

ctrlbold:
Sorry for the delay @wsmelton, gotta build a new AG cluster to test. But at least I'll finally have AGs in the lab again 💃

functions/Get-DbaAvailabilityGroup.ps1
ctrlbold: Fix is coming this month, but I would prefer that we try to be as standardized as possible and always try to support DynamicParams when possible.

functions/Get-DbaAvailabilityGroup.ps1
ctrlbold: Would like this to be standard as well. It helps because I can do mass finds and replaces.

PR #252, opened 2016-10-09
New module to collect metadata from CSV files.

fr3ky:
Proper help, examples need to be written.
Verbose output sort of completed.
Few debug checkpoints available.
Includes some level of error handling.
E.g:
image

returns PS object with vital metadata which then can be parsed.
E.g.:
image

### DIFF
--- a/functions/Get-CSVMetadata.ps1
+++ b/functions/Get-CSVMetadata.ps1
@@ -27,13 +27,17 @@ function Get-CSVMetaData
      $HeadersInRow = 1,
      [string]$Delimiter = ",",
      [bool]$HasFieldsEnclosedInQuotes = $False,
-     [switch]$noreturn
+     [switch]$DataTypes,
+     [switch]$TableOutput,
+     [switch]$Speedtest
     )
 
     # Load the basics
 	[void][Reflection.Assembly]::LoadWithPartialName("System.Data")
 	[void][Reflection.Assembly]::LoadWithPartialName("Microsoft.VisualBasic")
 	[void][Reflection.Assembly]::LoadWithPartialName("System.Windows.Forms")
+    Write-Verbose "[stopwatch] start"
+    $sw = [Diagnostics.Stopwatch]::StartNew()
 
     if (!$csv) 
     {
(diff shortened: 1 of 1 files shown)


functions/Get-CSVMetadata.ps1
```diff
@@ -0,0 +1,283 @@
+﻿function Get-UserPrompt {
```
ctrlbold: Is Get-CsvMetadata intended to be a public function or an internal one? If public, please go through Trello first when proposing new commands, as they must get prior approval. If private, please add to SharedFunctions.ps1 (which will soon be moved to a new folder and broken down by command.)

This seems like a pretty cool command overall, can you give details about how it will be used? If we ultimately agree on its use, please rename to Parse-DbaCsvMetadata and change into an advanced command with documentation, a BEGIN (this is where internal functions go), PROCESS and END. You can see this in the [dbatools-templates](https://github.com/sqlcollaborative/dbatools-templates) repo. 

Sorry this wasn't clarified before, I plan to work on the onboarding documentation in the wiki this week.

functions/Get-CSVMetadata.ps1
```diff
@@ -0,0 +1,283 @@
...
+            #  [int]       32-bit signed integer
+            #  [long]      64-bit signed integer
+            #  [bool]      Boolean True/False value
+            #  [decimal]   A 128-bit decimal value
+            #  [single]    Single-precision 32-bit floating point number
+            #  [double]    Double-precision 64-bit floating point number
+            #  [DateTime]  Date and Time
+            #  [string]    Fixed-length string of Unicode characters
+    
+            $hash = New-Object System.Object
```
ctrlbold: Please use [pscustomobject] instead. It's [beautiful and faster](http://www.jonathanmedd.net/2011/09/powershell-v3-creating-objects-with-pscustomobject-its-fast.html)

    [PSCustomObject]@{
        HeaderName = $HeaderName
        Column = $Column
        SampleLength = $SampleLength 
    }

Piping should be avoided in this case because it's slower

functions/Get-CSVMetadata.ps1
```diff
@@ -0,0 +1,283 @@
...
+            #  [decimal]   A 128-bit decimal value
+            #  [single]    Single-precision 32-bit floating point number
+            #  [double]    Double-precision 64-bit floating point number
+            #  [DateTime]  Date and Time
+            #  [string]    Fixed-length string of Unicode characters
+    
+            $hash = New-Object System.Object    
+            $hash | Add-Member -type NoteProperty -name HeaderName -value $HeaderName
+            $hash | Add-Member -type NoteProperty -name SampleValue -value $Column
+            $hash | Add-Member -type NoteProperty -name SampleLenght -value $Column.Length
```
ctrlbold: Typo - should be SampleLength

fr3ky:
Thank you for your comments! I have added in most of the bits. I send it as a new pull request because I messed up my copy of the repo and had to delete it. This request seem to have become an orphan as the result... :/ sry...

PR #259, opened 2016-10-13
Get-DbaDatabaseFreeSpace added columns

PowerDBAKlaas:
added growth, growtype, spaceuntilmaxsizeMB, AutoGrowthPossibleMB, UnusableSpaceMB to the SQL query and to the output object

### DIFF
--- a/functions/Get-DbaDatabaseFreespace.ps1
+++ b/functions/Get-DbaDatabaseFreespace.ps1
@@ -92,7 +92,7 @@ Returns database files and free space information for the db1 and db2 on localho
 				    ,CAST((f.size/128.0) AS DECIMAL(15,2)) AS [FileSizeMB]
 				    ,CAST((FILEPROPERTY(f.name, 'SpaceUsed')/(f.size/1.0)) * 100 as DECIMAL(15,2)) as [PercentUsed]
 					,CAST((f.growth/128.0) AS DECIMAL(15,2)) AS [GrowthMB]
-					,CASE is_percent_growth WHEN 1 THEN 'pct' WHEN 0 THEN 'KB' ELSE 'Unknown' END AS [GrowthType]
+					,CASE is_percent_growth WHEN 1 THEN 'pct' WHEN 0 THEN 'MB' ELSE 'Unknown' END AS [GrowthType]
 					,CAST((f.max_size/128.0) AS DECIMAL(15,2)) AS [MaxSizeMB]
 					,CAST((f.size/128.0) AS DECIMAL(15,2)) - CAST(CAST(FILEPROPERTY(f.name, 'SpaceUsed') AS int)/128.0 AS DECIMAL(15,2)) AS [SpaceBeforeAutoGrow]
 					,CASE f.max_size	WHEN (-1)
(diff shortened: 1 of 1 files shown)


PowerDBAKlaas:
- I'm not sure about the column names. suggestions are welcome.
- I had a PS script using SMO to collect this info, which I translated into t-Sql now. I don't know which method is the most efficient. Maybe I 'll compare them later on.
- All objects are held and in the end the collection is returned. PS best practices say explicitly not to do that. Is there a reason this way was chosen?

functions/Get-DbaDatabaseFreespace.ps1
```diff
@@ -168,12 +206,16 @@ Returns database files and free space information for the db1 and db2 on localho
 				'UsedSpaceMB' = $row.UsedSpaceMB;`
 				'FreeSpaceMB' = $row.FreeSpaceMB;`
 				'FileSizeMB' = $row.FileSizeMB;`
-				'PercentUsed' = $row.PercentUSed
+				'PercentUsed' = $row.PercentUSed;`
+				'AutoGrowth' = $row.GrowthMB;`
+				'AutoGrowType' = $row.GrowthType;`
```
dansqldba: This is displaying as "KB" when it should probably be "MB"

dansqldba: Latest commit 50213e8 seems to have reverted GrowthType back to KB instead of MB?

functions/Get-DbaDatabaseFreespace.ps1
```diff
@@ -91,6 +91,52 @@ Returns database files and free space information for the db1 and db2 on localho
...
+												THEN	CASE f.max_size
+														WHEN (-1)
+														THEN CONVERT([int],f.Size*power((1)+CONVERT([float],f.Growth)/(100),CONVERT([int],log10(CONVERT([float],(2147483648.))/CONVERT([float],f.Size))/log10((1)+CONVERT([float],f.Growth)/(100)))))
+														ELSE CONVERT([int],f.Size*power((1)+CONVERT([float],f.Growth)/(100),CONVERT([int],log10(CONVERT([float],f.Max_Size)/CONVERT([float],f.Size))/log10((1)+CONVERT([float],f.Growth)/(100)))))
+														END
+														ELSE (0)
+														END ))/128.0 AS DECIMAL(15,2))
+									END AS [UnusableSpaceMB]
+ 
 				FROM sys.database_files AS f WITH (NOLOCK)
```
dansqldba: Support for SQL 2000 is broken by your proposed changes in this cmdlet.

Msg 208, Level 16, State 1, Line 1
Invalid object name 'sys.database_files'.
Msg 208, Level 16, State 1, Line 1
Invalid object name 'sys.filegroups'.

Splaxi: SELECT * FROM dbo.sysfilegroups
SELECT * FROM dbo.sysfiles
This should work on SQL 2000

dansqldba: Seems I'm full of it and this cmdlet never worked on SQL 2000 at all. Somehow this was missed when the cmdlet was first created. The changes made by @PowerDBAKlaas weren't at fault here. Sincere apologies to you!

PR #263, opened 2016-10-13
initial commit of Find-DbaOrphanedFile

sanderstad:
Function to find any database related file in the SQL Server directories which is not part of the active databases in the instance


functions/Find-DbaOrphanedFile.ps1
```diff
@@ -0,0 +1,231 @@
...
+                $SourceDirectory += $sourceserver.Settings.DefaultFile
+            }
+
+            # Check if the default log path is set
+	        if (($sourceserver.Settings.DefaultLog).Length -eq 0) {
+	            $SourceDirectory = $sourceserver.Information.MasterDBLogPath
+	        }
+            else
+            {
+                $SourceDirectory += $sourceserver.Settings.DefaultLog
```
ctrlbold: There is an internal command in SharedFunctions.ps1 to get this because it's not always set properly in SMO. Please see Get-SqlDefaultPaths, which tries 3 different ways to get defaults.

functions/Find-DbaOrphanedFile.ps1
```diff
@@ -0,0 +1,231 @@
...
+                
+            }
+
+            # If there are files found
+            if($diskfiles.count -ge 1)
+            {
+                switch($method)
+                {
+                    'local' { $dbfiles = $databasefiles.Tables[0].filename }
+                    'remote' { $dbfiles = $databasefiles.Tables[0].filename | ? {$_ -match ":"} | % {$_ -replace ':','$'} | % {"\\$servername\" + $_} }
```
ctrlbold: gorgeous!

wsmelton: We should not be using aliases for PowerShell commands right? So `?` should be `where` or `where-object`, then `%` should be foreach`, etc.

ctrlbold: Oh snap, brain interpreted it as full commands. Correct, we do not use aliases/shortcuts, and in addition, we also cap. So this would be Where-Object and ForEach-Object

functions/Find-DbaOrphanedFile.ps1
```diff
@@ -0,0 +1,231 @@
...
+	# This is a sample. Please continue to use aliases for discoverability. Also keep the [object] type for sqlserver.
+	[CmdletBinding(SupportsShouldProcess = $true)]
+	Param (
+		[parameter(Mandatory = $true, ValueFromPipeline = $true)]
+		[Alias("ServerInstance", "SqlInstance")]
+		[object]$SqlServer,
+        [parameter(Mandatory = $true, ValueFromPipeline = $true)]
+		[object]$SqlCredential,
+        [parameter(Mandatory = $false, ValueFromPipeline = $true)]
+        [string[]]$SourceDirectory
```
sanderstad: I will rename the SourceDirectory param to Path, no problem.
Yes, this path will be in addition to the defaults in some cases we have databases in other directories or drives that are not the default in SQL Server.

functions/Find-DbaOrphanedFile.ps1
```diff
@@ -0,0 +1,231 @@
...
+	)
+	
+    DynamicParam { if ($SqlServer) { return Get-ParamSqlDatabases -SqlServer $SqlServer -SqlCredential $SqlCredential } }
```
sanderstad: Should have removed that, I did use it in an earlier version

functions/Find-DbaOrphanedFile.ps1
sanderstad: Replaced the smo version with the shared function. I still needed to add the rootdir though because that was not returned by the shared function.

functions/Find-DbaOrphanedFile.ps1
sanderstad: Small change, will make it so

functions/Find-DbaOrphanedFile.ps1
```diff
@@ -0,0 +1,225 @@
...
+        {
+            $servername = $SqlServer.Split('\')[0]
+        }
+        else
+        {
+            $servername = $SqlServer
+        }
+
+        # Check if the servername is the local server
+        if(($env:computername -eq $servername) -or ($env:computername -eq '127.0.0.1'))
```
ctrlbold: $env:computername will never equal 127.0.0.1, it should be $servername -eq '127.0.0.1'

functions/Find-DbaOrphanedFile.ps1
```diff
@@ -0,0 +1,225 @@
...
+                {
+                    'local' { $dbfiles = $databasefiles.Tables[0].filename }
+                    'remote' { $dbfiles = $databasefiles.Tables[0].filename | Where-Object {$_ -match ":"} | ForEach-Object {$_ -replace ':','$'} | ForEach-Object {"\\$servername\" + $_} }
+                }
+
+                # Compare the two lists and save the items that are not in the database file list 
+                $orphanedfiles = (Compare-Object -ReferenceObject ($dbfiles) -DifferenceObject $diskfiles.FullName).InputObject
+            }
+                
+            return $orphanedfiles
```
ctrlbold: Works! But please return two columns LocalPath and RemotePath, filled with local (M:\etc) and remote (\\server\m$\etc)

ctrlbold:
I will take this one for QA. 

@sanderstad, I was dreaming last night and realized that $server.EnumDirectories() would keep this 100% SMO with no external access requirements. Please use this method instead of Get-ChildItem with the UNC path. If you do not have time, let me know and I can rush it in.  This will also mean that you can remove the local admin requirement.

- [ ] Works remotely
- [ ] Works locally

functions/Find-DbaOrphanedFile.ps1
```diff
@@ -0,0 +1,225 @@
...
+{
+<#
+.SYNOPSIS 
+Find-DbaOrphanedFile moves/removes orphaned database files
+
+.DESCRIPTION
+Get all the database files for all the database for the instance
+Get the various directories of the instance and get all the present database files.
+Compare which the two lists to see if there are any orphaned files and return the list
+
```
wsmelton: Would reword this to be "find" instead of "get", to just match the verb used in naming the command. 😉 

Would also recommend formatting this in 2-3 sentence paragraph, just looks like bullet list. Description does not have to explain exactly what is done, more of what problem it solves.

functions/Find-DbaOrphanedFile.ps1
```diff
@@ -0,0 +1,225 @@
+﻿
+Function Find-DbaOrphanedFile
+{
+<#
+.SYNOPSIS 
+Find-DbaOrphanedFile moves/removes orphaned database files
```
wsmelton: I don't see any code that shows you do any "moves/removes" for orphaned files found. Should be reworded.

sanderstad: The reason for that is because it only finds the orphaned files. You could easily pipe the returned values to Remove-Item or Move-Item to execute those actions. I' not sure we're still willing to create the move-dbaorphanedfile and remove-dbaorphanedfile functions.

wsmelton: The synopsis should state what the function does, not necessarily what you can do with the output. You might put those as examples with using `Remove-Item` or `Move-Item`. It implies that the function will do those actions, which is a bit misleading.

sanderstad: I agree, I'll change the synopsis.

ctrlbold:
@sanderstad you still need me to finish up the code? We'll probably release by Oct 27.

sanderstad:
@ctrlbold If you could help with that would really help me out. I'll probably not make it to changes to the code in time

PR #267, opened 2016-10-17
SharedFunctions - Move all to individual files

ctrlbold:
Moved SharedFunctions functions to \internal


ctrlbold:
How to test

- [ ] Compare functions found in the old SharedFuntions.ps1 to the functions found in the new directory; \internal, ensuring all of them are there
- [ ] Compare that the function names are the same inside the file as the filename itself
- [ ] See that two functions have been moved to public - Test-SqlConnection and Update-dbatools
- Run a couple commands that require a bunch of internal commands such as Copy-SqlDatabase and Sync-SqlLoginPermissions

functions/Test-SqlConnection.ps1
```diff
@@ -0,0 +1,222 @@
+Function Test-SqlConnectionTest-SqlConnection
```
dansqldba: This doesn't look right.

dansqldba:
@ctrlbold has asked for this to go through before anything else. Need to see if any other PR's are making changes to ShareFunctions.

Quick look through the code only one things looks suspect.

functions/Test-SqlConnection.ps1
ctrlbold: Fixed, thanks

PR #278, opened 2016-10-19
Copy-SqlDatabase - Fixed Check for ROWS or LOG

dczarny:
The check for a SQL Server 2000 to see if a datafile was ROWS or LOG had
a bug if a database had more than one filegroup associated with it.


ClaudioESSilva:
Approved :+1:

PR #285, opened 2016-10-26
Bugfix login - moved kill closer to drop

ctrlbold:


ClaudioESSilva:
My suggestion is Disable the destination login before kill all processes.
That way we ensure that no new connections can be done. Then we can kill the existing sessions followed by the Drop.

Just add the `$login.Disable()` line of code before the KillProcess
```
$login.Disable()
$destserver.EnumProcesses() | Where-Object { $_.Login -eq $username } | ForEach-Object { $destserver.KillProcess($_.spid) }
$login.Drop()
```

have test it and works great :smile:

ctrlbold:
added Disable thanks 💃

PR #287, opened 2016-10-27
Test-SqlTempDBConfiguration - Improve output - Fixes #269

### LINKED ISSUE #269: Test-SqlTempDbConfiguration - Doesn't give feedback when everything is OK
Was showing the command to a collegue, after he apply all the configs he runs the Test command again and he was like "is this OK?" and then "Cláudio come here please" :smile: 

Unless we use the `-Detailed` switch, if everything is configured right the execution don't return anything.
I was expecting to receive a message like "Everything looks good." or "Nothing to point. If you want see more information please use -Detailed switch" something like that.

btw we need to update this command page on dbatools site to match the screenshots. Actually no warning is issued and without Format-Table the output is not the one shown on the image.

ctrlbold:
Tests

- [x]  Test if it returns "All tests passed! tempdb is properly optimized." when all best practices were followed
- [x] Test if it returns the proper values for -Detailed
- [x] Test if it returns failed best practice if some exist

I already tested all of them and it worked. This should be an easy merge.

### DIFF
--- a/dbatools.psd1
+++ b/dbatools.psd1
@@ -153,7 +153,8 @@
 		'Get-DbaLastGoodCheckDb',
 		'Test-DbaDatabaseBestPractice',
 		'Remove-DbaBackup',
-		'Find-DbaOrphanedFile'
+		'Find-DbaOrphanedFile',
+		'Test-DbaFullRecoveryModel'
 	)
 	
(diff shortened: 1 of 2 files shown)


dansqldba:
Fixed #269 but don't see any link here for that issue...

PR #301, opened 2016-11-04
inital commit of Find-DbaAgentJob

Staggerlee011:
Looking for input, on if I should add more filter options, or suggested changes.


functions/Find-DbaAgentJob.ps1
```diff
@@ -0,0 +1,137 @@
...
+	Param (
+		[parameter(Position = 0, Mandatory = $true, ValueFromPipeline = $True)]
+		[Alias("ServerInstance", "SqlInstance", "SqlServers")]
+		[string[]]$SqlServer,
+		[System.Management.Automation.PSCredential]$SqlCredential,
+		[int]$LastUsed,
+        [switch]$Disabled,
+        [switch]$NoSchedule,
+        [switch]$EmailNotification,
+        [string[]]$Filter,
```
ctrlbold: Looking at how $Filter is used, this should be $Name instead

functions/Find-DbaAgentJob.ps1
```diff
@@ -0,0 +1,137 @@
...
+			    {
+				    Write-Verbose "Failed to connect to: $instance"
+				    continue
+			    }
+                
+                
+                IF ($Filter)
+                    {
+                        Write-Verbose "Applying filter/s"
+                        $jobs | Where-Object {$Filter -notcontains $_.name}
```
ctrlbold: Filters are not to filter out, they are more like Where's on the "left" side. Check out [Windows PowerShell Filter Left, Format Right](https://technet.microsoft.com/en-us/library/2009.09.windowspowershell.aspx)

I've never really researched how to do a proper PowerShell filter. We may just have no filter for v1.

functions/Find-DbaAgentJob.ps1
```diff
@@ -0,0 +1,137 @@
...
+        [string[]]$Filter,
+        [switch]$Detailed
+	)
+	BEGIN
+	    {
+            $output = @()
+	    }
+	PROCESS
+	    {
+		    FOREACH ($instance in $SqlServer)
```
ctrlbold: please change this (and subsequent occurrences) to foreach ($servername in $sqlserver) until we decide on standards. i actually like $instance a lot and will propose it as the variable we use.

functions/Find-DbaAgentJob.ps1
```diff
@@ -0,0 +1,137 @@
...
+.PARAMETER Disabled
+Find all jobs that are disabled
+	
+.PARAMETER NoSchedule
+Find all jobs with schedule set to it
+	
+.PARAMETER EmailNotification
+Find all jobs without email notification configured
+
+.PARAMETER Filter
```
ctrlbold: This would be an Exclude

functions/Find-DbaAgentJob.ps1
PowerDBAKlaas: Isn't ther an assignment missing here?
$jobs = ...

ctrlbold: good catch!

functions/Find-DbaAgentJob.ps1
```diff
@@ -0,0 +1,137 @@
...
+                        Write-Verbose "Finding job/s that have no email operator defined"
+                        $output += $jobs | Where-Object {$_.OperatorToEmail -eq "" }
+                    }
+            }
+        }
+    END
+        {
+            IF ($Detailed -eq $true)
+		        {
+                    return ($output | Select-Object name, LastRunDate, IsEnabled, HasSchedule, OperatorToEmail -Unique)
```
PowerDBAKlaas: Shouldn't the $servername ( or better: $instance) be included in the output?

functions/Find-DbaAgentJob.ps1
PowerDBAKlaas: Actually we need both. A function like Get-DbaPowerplan works on a computer (host), while this one needs SQL instance(s). Therefore I find it important to use both where they're appropriate. Since 'server' is the most ambiguous, I prefer $computername for computers and $instance(name) for instances.

PowerDBAKlaas:
If both -disabled and -noschedule are applied, the ouptut contains jobs that are disabled OR have no schedule, where I would expect only jobs that pass both filters.

Jobs can be added multiple times to output now and in the end duplicates are removed.

If a $selection could be built by adding 'where' clauses, the selection can be limited to jobs that pass all filters, are selected in one go, and every job will be selected only once from the beginning.

dansqldba:
Inserting QA checklist...

functions/Find-DbaAgentJob.ps1
```diff
@@ -192,8 +196,35 @@ Queries CMS server to return all SQL instances in the Production folder and then
...
+                                                $DynString += ' | Where-Object { $Owner -contains $_.OwnerLoginName '
+                                            }
+                                        ELSEIF ($filter -eq 1)
+                                            {
+                                                $DynString += '-and $Owner -contains $_.OwnerLoginName '
+                                            }
+                                    }
                             }
 
                         IF (!($DynString -eq '$output = $server.JobServer.jobs'))
```
PowerDBAKlaas: this means "when we built a string that's more complex than return all jobs", right?
IF ( $DynString -ne '$output = $server.JobServer.jobs') is a bit more direct

functions/Find-DbaAgentJob.ps1
```diff
@@ -246,7 +277,21 @@ Queries CMS server to return all SQL instances in the Production folder and then
...
                             }
-
+                        IF ($Owner)
+                            {
+                                Write-Verbose "Finding job/s with owner critera"
+                                IF ($Owner -match "-")
+                                    {
+                                        $OwnerMatch = $Owner -replace "-", ""
+                                        Write-Verbose "Checking for jobs that NOT owned by: $OwnerMatch"
+                                        $output += $server.JobServer.jobs | Where-Object { $OwnerMatch -notcontains $_.OwnerLoginName }
```
PowerDBAKlaas: I think that, instead of reassigning $jobs and $output so many times, building a where clause string ( as above with $dynstring) and only getting the $output once in the end, could improve the performance

functions/Find-DbaAgentJob.ps1
```diff
@@ -258,7 +303,7 @@ Queries CMS server to return all SQL instances in the Production folder and then
         {
             IF ($Detailed -eq $true)
```
PowerDBAKlaas: move this block to PROCESS
objects should be released sooner

niphlod:
@Staggerlee011 : I'd add:
- a "status" filter to report only for a specific status (i.e. give me all jobs that failed)... this could also probably be handled in a custom PS filtering the resultset, but in my env there are servers firing thousands jobs per hour and being able to filter from source would be great
- the only other thing I can think of is that lots of "check/alarm" software report for job executions that ran longer than the 20%-30% of the mean of the last executions. Takes a little math but can be added afterwards.

Also, pertaining the "temporal filter" (is really LastUsed the best name here? LastRunTime maybe...) it'd be better to accept an integer for days and the usual notation (i.e. 1m, 1w, etc) to represent a timespan. To that end I'd suggest in releases down the line to extrapolate the RetentionPeriod calculations from "remove-dbadatabasebackup", put it in an utility function and use it for all timespan-related params.

Staggerlee011:
going to re-write this and make it a bit cleaner (remove the option to combine filters and just make it default), follow the new style designs ((outputs shouldn't wait till end {} now etc). is there a point where we are creating to many switches though and should do filtering another way? at the moment there is 

[string[]]$SqlServer,
[int]$LastUsed,
[switch]$Disabled,
[switch]$NoSchedule,
[switch]$EmailNotification,
[string[]]$Category,
[string]$Owner,
[string[]]$Exclude,
[string[]]$Name,
[string]$Database
[switch]$Detailed

dansqldba:
Are you going to have time to do your re-write before the release deadline, or should we bump this/close the PR?

ctrlbold:
Accepting and then making some changes that didn't seem to work right off. Will do well for v1.0.

Issue #240, opened 2016-10-05
Set-SqlMaxMemory - Wrong variable $servername


ClaudioESSilva:
On line 89 (https://github.com/sqlcollaborative/dbatools/blob/development/functions/Set-SqlMaxMemory.ps1#L89) inside if block for testing sysadmin permissions we are using `$servername` but this variable does not exists on the script.
Should be `$sqlserver` instead.

ConstantineK:
This is fixed in #304

PR #304, opened 2016-11-05
Set-SqlMaxMemory - Bugfix for wrong variable $servername #240

SirCaptainMitch:
Bug fix for issue #240  

line 89 variable change to $sqlserver so that it will use the real variable name.


ctrlbold:
Awesome,thanks @SirCaptainMitch !

PR #307, opened 2016-11-10
Find-DbaOrphanedFile - bug fixes and refactoring

### LINKED ISSUE #300: Find-DbaOrphanedFile bug
From two users:

1. Hekaton directories were appearing
2. This issue:

I found a bug in the Find-DBaOrphanedFiles command. if you have any filestream filegroup, this would include in-memory tables, then this command flags those files as orphans.

A where clause could be added to the sql that gets all files to only get types 0 and 1. This would stop the  command from incorrectly flagging these as orphans.

line 100: 	$sql = "SELECT Physical_Name AS filename FROM sys.master_files mf INNER JOIN  sys.databases db ON db.database_id = mf.database_id where type in (0,1)"

alternatively, directories can be added to the filesondisk results. this would be preferred as it would also check for these valid file but would unfortunately find any other directories in the paths. (this edit attached).

[bk-Find-DbaOrphanedFile.txt](https://github.com/sqlcollaborative/dbatools/files/570154/bk-Find-DbaOrphanedFile.txt)

Please resolve in November

ConstantineK:
Contributing changes to fix and close #300 
This may need some additional review due to not exactly following what was stated in the issue as the resolution, instead doing a slightly more wide ranging refactoring.

It now iterates over the folders to build a SQL string to insert multiple times, joins the names in a temp table, and returns all the values as null which do not match.

### DIFF
--- a/functions/Find-DbaOrphanedFile.ps1
+++ b/functions/Find-DbaOrphanedFile.ps1
@@ -39,6 +39,8 @@ This program is free software: you can redistribute it and/or modify it under th
 This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 You should have received a copy of the GNU General Public License along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+Thanks to Paul Randal's notes on FILESTREAM which can be found at http://www.sqlskills.com/blogs/paul/filestream-directory-structure/
+
 .LINK
 https://dbatools.io/Find-DbaOrphanedFile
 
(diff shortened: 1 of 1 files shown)


ConstantineK:
Also shout out to @gbargsley for testing this to test if it resolved his issue with duplicate file names returning.

niphlod:
hi @ConstantineK, I think that there's an issue with the new implementation...

as you join on

mf.[current_database_files] = e.fs_filename

it can happen that any orphan file that is named as one of the ones resulting from
```
     SELECT REVERSE(SUBSTRING(REVERSE(filename), 0, CHARINDEX('\', REVERSE(filename)))) AS [current_database_files]
      FROM sys.sysaltfiles AS m
```
gets not checked. 
I noticed it because I have two distinct dirs for data and logs (e.g. C:\SQLData, C:\SQLLog) and had a orphaned ldf in the data dir (e.g. C:\SQLData\mydb.ldf)
Given ldf was named exactly as the correct one in the log dir (which is C:\SQLLog\mydb.ldf) , it resulted in not being an orphan.
I think we should compare the result of #enum normalizing the column "filename" it to lowercase AND a pass of [IO.Path]::GetFullPath() (that should take care of funny entries like C:\SQLData\\mydb.ldf, notice the double backslash), and do the same for filename in sys.sysaltfiles, and then compare the two.

What do you think ?

ctrlbold:
Hey @niphlod, you have any code to propose while @ConstantineK is sleeping? ;D I'd like to get a release out today and if you have some code to paste (or a formal PR), we can slip it in.

niphlod:
Hum, I can get to work on it but dunno if it'll be ready in an hour ^_^

niphlod:
there's also another bug because in the new Format-Comparison a proper [Microsoft.SqlServer.Management.Smo.SqlSmoObject] isn't built . guess that waiting tomorrow is not an option ??

ctrlbold:
This is one of my favorite commands -- I'd like to get it right. I can wait until tomorrow 👍

niphlod:
mine too but needs a little bit more ❤️

ConstantineK:
Hey @niphlod, 

I would love to dive into your request, but I am a little confused on two points:

Format-comparison doesnt return a SqlSmoObject, and I was wondering why it should? 

I dont fully ken your example about the sysaltfiles, that reverse method should catch the first slash and then not have an issue there with paths with double slashes, while I do agree that making them lowercase might help, the join itself is case insensitive.

You mentioned that "it can happen that any orphan file that is named as one of the ones resulting from" (query) "are not checked"
This is by design, a file in your sysaltfiles would not be an orphan, it definitely would be one of your database files.

I will be in and out throughout the day, but I would love to understand more of this problem so I can solve it for us :)

niphlod:
hi @ConstantineK , I'll try to rephrase it better

1. in format-comparison, as $smoserver isn't passed, the logic querying either sysaltfiles or master_files doesn't get triggered
2. when you do the left join, it's done on the filename part, not the complete path. This opens the possibility to NOT report an orphan C:\SQLData\mydb.ldf file IF a valid file is, e.g., C:\SQLLog\mydb.ldf because the comparison is done on  "mydb.ldf"

I'm doing a PR ASAP that should fix and streamline the processing of other subtleties, hoping that #300, which I really can't replicate, is still fixed.

ConstantineK:
Thank you, that makes sense.

I have pushed a change to fix the issue with the first problem, but the subtlety of the second one is difficult and a good point and I often see duplicate filenames in different folders.

niphlod:
@ConstantineK : there's https://github.com/sqlcollaborative/dbatools/pull/321 with what I "think" it should do.

@ctrlbold : there's a blunt "FIXME" as right now (and also in this PR) a parameter is missing and an entire piece of code gets skipped, that being said it's a lot better than the current version. Figuring out what to do with FT catalogs and FS dirs can be left for the new-year release ^_^

PR #308, opened 2016-11-10
initial commit of Get-DbaPermission

PowerDBAKlaas:
A new function to list all permissions on server and database level.
Every output object stands for a permission: a state of a permissiontype granted on a securable to a principal.

Used Get-DbaLastgoodCheckDB as template.
Besides the common parameters sqlserver, exclude, databases and credential,
I also kept the -Detailed switch,
and used a -IncludeServerLevel switch to include the permissions on server level,
and a -NoSystemObjects to exclude securables that are system objects

### DIFF
--- a/functions/Get-DbaPermission.ps1
+++ b/functions/Get-DbaPermission.ps1
@@ -1,4 +1,4 @@
-﻿Function Get-DbaPermissions
+﻿Function Get-DbaPermission
 {
 <#
 .SYNOPSIS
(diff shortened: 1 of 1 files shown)


dansqldba:
@PowerDBAKlaas did this start out on Trello, there's no trace of it? Has it been approved by @ctrlbold?

PowerDBAKlaas:
@dansqldba I tried adding a card on Trello, but I don't have that option. Then I saw this:
image
So I thought creating a pull request was acceptable.

PowerDBAKlaas:
image

dansqldba:
I think the preference is for _new stuff_ to go through Trello first, the "development participation" sends you to the Trello board. From there your idea can be discussed and an outline for the code agreed before development begins :)

ctrlbold:
- [ ] Works remotely
- [ ] Works locally

PR #314, opened 2016-11-17
Copy-SqlCredential - fixed words

ctrlbold:


ctrlbold:
- [ ] Works remotely
- [ ] Works locally

PR #318, opened 2016-11-18
Get-DbaTcpPort - Added more info about SqlWmi exceptions

ctrlbold:
Added additional exception information for failed SqlWmi connections.


ctrlbold:
Output now looks like this:

image

ctrlbold:
Thanks for the review, bud!

PR #320, opened 2016-11-20
New command Get-DbaStartupParameter

ctrlbold:
Why isn't my template showing up like it did for @ConstantineK? 👎  

- [x] Works remotely
- [x] Works locally

### DIFF
--- a/functions/Get-DbaStartupParameter.ps1
+++ b/functions/Get-DbaStartupParameter.ps1
@@ -5,16 +5,17 @@
 Displays SQL Server Startup Parameters
 
 .DESCRIPTION
-Displays a detailed list of SQL Server Startup Parameters. 
-	
+Displays values for a detailed list of SQL Server Startup Parameters including Master Data Path, Master Log path, ErrorLog, Trace Flags, Parameter String and much more.
+
+See https://msdn.microsoft.com/en-us/library/ms190737.aspx for more information.
 .PARAMETER SqlServer
 The SQL Server that you're connecting to.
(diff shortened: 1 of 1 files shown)


ctrlbold:
@js0505 this is for you!

ctrlbold:
Output looks like this

image

functions/Get-DbaStartupParameter.ps1
```diff
@@ -0,0 +1,163 @@
...
+					$displayname = $args[1]
+					
+					$wmisvc = $wmi.Services | Where-Object { $_.DisplayName -eq $displayname }
+					
+					$params = $wmisvc.StartupParameters -split ';'
+					
+					$masterdata = $params | Where-Object { $_.StartsWith('-d') }
+					$masterlog = $params | Where-Object { $_.StartsWith('-l') }
+					$errorlog = $params | Where-Object { $_.StartsWith('-e') }
+					$traceflags = $params | Where-Object { $_.StartsWith('-T') }
```
Splaxi: We should look for the lowercase trace flags that might have been set by accident or during a support case with Microsoft and then left on. 

msdn.microsoft.com/en-us/library/ms190737.aspx

_** Important ** When specifying a trace flag with the -T option, use an uppercase "T" to pass the trace flag number. A lowercase "t" is accepted by SQL Server, but this sets other internal trace flags that are required only by SQL Server support engineers. (Parameters specified in the Control Panel startup window are not read.)_

PR #321, opened 2016-11-21
Updated Find-DbaOrphanedFile

### LINKED ISSUE #300: Find-DbaOrphanedFile bug

niphlod:
Fixes #300  

 - comparison for the complete filepath
 - lowercase comparison
 - path normalization via [IO.Path]::GetFullPath($_)

- [x]  Powershell 3
- [x]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [x]  Windows 10
- [ ]  Azure Database

fixing some subtleties on top of the mods made on https://github.com/sqlcollaborative/dbatools/pull/307

BTW: dunno why it's there but I left a FIXME just to point out something that should be fixed in the following releases

### DIFF
--- a/functions/Find-DbaOrphanedFile.ps1
+++ b/functions/Find-DbaOrphanedFile.ps1
@@ -93,7 +93,7 @@ Finds the orphaned ending with ".fsf" and ".mld" in addition to the default file
 			)	
 			# use sysaltfiles in lower versions
 			
-			$q1 = "CREATE TABLE #enum ( id int IDENTITY, fs_filename nvarchar(512), depth int, is_file int, parent nvarchar(512) ); DECLARE @dir nvarchar(512);"
+			$q1 = "CREATE TABLE #enum ( id int IDENTITY, fs_filename varchar(512), depth int, is_file int, parent nvarchar(512) ); DECLARE @dir nvarchar(512);"
 			$q2 = @"
 				SET @dir = 'dirname';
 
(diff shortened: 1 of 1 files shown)


ctrlbold:
Hey @niphlod, @ConstantineK and i were thinking that we'd love to have you in #dbatools on Slack. Would you wanna join the party there, too? https://dbatools.io/slack

niphlod:
fortunately for me, unfortunately for dbatools, I range from 2008 to 2016 and tested just there ;-(

PR #326, opened 2016-11-22
initial commit of Get-DbaMemoryUsage

PowerDBAKlaas:
- [ ]  PowerShell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally

### DIFF
--- a/functions/Get-DbaMemoryUsage.ps1
+++ b/functions/Get-DbaMemoryUsage.ps1
@@ -2,18 +2,18 @@
 {
 <#
 .SYNOPSIS
-Get amount of memory in use by SQL Server components
+Get amount of memory in use by *all* SQL Server components and instances
 
 .DESCRIPTION
-Retrieves the amount of memory per performance counter
-Default output includes columns Server, counter instance, counter, number of pages, memory in KB, memory in MB
+Retrieves the amount of memory per performance counter. Default output includes columns Server, counter instance, counter, number of pages, memory in KB, memory in MB
 SSAS and SSIS are included.
+	
 SSRS does not have memory counters, only memory shrinks and memory pressure state.
 
 This function requires local admin role on the targeted computers.
 
 .PARAMETER ComputerName
-The SQL Server that you're connecting to.
+The Windows Server that you are connecting to. Note that this will return all instances, but Out-GridView makes it easy to filter to specific instances.
 
 .PARAMETER Credential
 Credential object used to connect to the SQL Server as a different user
--- a/functions/Get-DbaMemoryUsage.ps1
+++ b/functions/Get-DbaMemoryUsage.ps1
@@ -7,6 +7,9 @@ Get amount of memory in use by SQL Server components
 .DESCRIPTION
 Retrieves the amount of memory per performance counter
 Default output includes columns Server, counter instance, counter, number of pages, memory in KB, memory in MB
+SSAS and SSIS are included.
+SSRS does not have memory counters, only memory shrinks and memory pressure state.
+
 This function requires local admin role on the targeted computers.
 
 .PARAMETER ComputerName
@@ -26,6 +29,8 @@ Copyright (C) 2016 Chrissy LeMaire
 This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
 This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
 You should have received a copy of the GNU General Public License along with this program. If not, see http://www.gnu.org/licenses/.
+
+SSIS Counters: https://msdn.microsoft.com/en-us/library/ms137622.aspx
 
 .LINK
  https://dbatools.io/Get-DbaMemoryUsage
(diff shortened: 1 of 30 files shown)


functions/Get-DbaMemoryUsage.ps1
```diff
@@ -0,0 +1,137 @@
...
+
+    }
+
+	PROCESS
+	{
+        foreach ($servername in $ComputerName)
+        {
+            if ($servername -match '\\')
+			{
+				$servername = $servername.Split('\\')[0]
```
ctrlbold: They may throw an aliased server name. Please use the newly created internal command Resolve-IpAddress to get the IP. You can throw a few things at it: a string, an smo object or even a string with a named instance.

functions/Get-DbaMemoryUsage.ps1
```diff
@@ -0,0 +1,137 @@
...
+			}
+            Write-Verbose "Connecting to $servername"
+			if ( test-connection -ComputerName $servername -Quiet -count 1)
+            {
+                $availablecounters = (get-counter -ComputerName $servername -ListSet '*sql*:Memory Manager*').paths
```
ctrlbold: Please use upper case for all commands, including Test-Connection

functions/Get-DbaMemoryUsage.ps1
```diff
@@ -0,0 +1,137 @@
...
+				                MemKB = $_.cookedvalue
+				                MemMB = $_.cookedvalue / 1024
+                                }
+                            }
+
+
+                $availablecounters = (get-counter -ComputerName $servername -ListSet '*sql*:Plan Cache*' ).paths
+                (get-counter -ComputerName $servername -Counter $availablecounters).countersamples |
+                    Where-Object {$_.Path -match $Plancounters} |
+                    foreach { [PSCustomObject]@{
```
ctrlbold: love 💃

functions/Get-DbaMemoryUsage.ps1
```diff
@@ -0,0 +1,137 @@
...
+            }
+			else
+			{
+				Write-Warning "Can't connect to $servername. Moving on."
+				Continue
+			}
+        }
+    }
+    END
+    {}
```
ctrlbold: You can remove this part

functions/Get-DbaMemoryUsage.ps1
```diff
@@ -0,0 +1,137 @@
...
+                                Counter = $_.Path.split("\")[-1]
+					            Pages = $_.cookedvalue
+					            MemKB = $_.cookedvalue * 8192 / 1024.0
+					            MemMB = $_.cookedvalue * 8192 /1048576.0
+                                }
+                            }
+            }
+			else
+			{
+				Write-Warning "Can't connect to $servername. Moving on."
```
ctrlbold: Is there a way for us to catch this earlier?

PowerDBAKlaas: I don't think so: the test-connection is inside the if clause. If this returns nothing, the if {} block is skipped and the next step is the else {} block.

functions/Get-DbaMemoryUsage.ps1
PowerDBAKlaas: done

functions/Get-DbaMemoryUsage.ps1
PowerDBAKlaas: it's neat to leave it in, according to me and Don Jones :-)

ctrlbold:
Hey @PowerDBAKlaas, can you make the change to the IP/Hostname and I can test and merge? So many PR's have been stuck.

ctrlbold:
Also I'm not sure why but there are too many files in this PR, including psd1

PowerDBAKlaas:
I've seen those files, and I don't know what they have to do with this PR, or how they ended up in my branch. Can a Github specialist fix this?

How should I handle the hostname issue right now? use resolve-ipaddress? or Get-Connection? I haven't come to write a better version of the internal function yet.

ConstantineK:
Sorry that I lost some of your commits, but it looked like the files had been changed in a previous version and was having trouble merging, so I did the following:

- Renamed the existing branch (git branch -m old_branch new_branch).
- Created a new branch from development (git checkout development; git checkout -b New-Get-DbaMemoryUsage), remember that case counts!
- Checked out the code we were interested in (git checkout old_branch filepath) to add it to our existing branch.
- Committed and force pushed the code to this branch (git commit -m "message"; git push origin New-Get-Dba-MemoryUsage --force)

ctrlbold:
oh man, that 1 files changed is gorgeous! amazing work @ConstantineK !

ctrlbold:
This returns memory usage for all instances on a server. When using it, I'd assume it would be per SQL Server.

PowerDBAKlaas:
@ctrlbold using Resolve-DbaNetworkName for the first time. 😃

PowerDBAKlaas:
@ctrlbold can you show the output you get, because I get a counter per sqlinstance (as long as the counters are present on the computer, they can easily go missing 😢 )

ctrlbold:
Yes, here, sorry :(

image

PowerDBAKlaas:
and what would you like to be different?

ctrlbold:
My initial impression is that it was for _one_ SQL Server instance. So the output should only be related to the SQL Instance that was passed. Otherwise, this should be a ComputerName only parameter. 

Even if that is the case, the output should be parsed and an extra column should be added that has the instance name. In this case, sqlserver seems like MSSQLSERVER/default and mssql$instancename seems like the named instances.

Ultimately, it'd be awesome to have -- if they pass ComputerName, then all instances, but if they pass SqlInstance, then the specific instance.

PowerDBAKlaas:
it is ComputerName, but
as you asked to "handle anything they can throw at us", I accept sqlinstances and get the computername out of it, so even 'realcomputer\nonexistingInstance' will work.
All this is handled by the praised Resolve-DbaNetworkName. 😃 
I removed the 'sqlinstance' and 'sqlserver' aliases.

I put the sqlinstance in a separate property.

This is not the ultimate version yet, but acceptable as an early release?

PR #336, opened 2016-11-25
New commit Read-DbaBackupHeader

ctrlbold:
needs dynamic params and other internal files approved first, then a sync

- [x] Works remotely
- [x] Works locally
- [x] Supports multiple files being piped in
- [x] No un-handled errors which stop the command working with multiple files

### DIFF
--- a/functions/Read-DbaBackupHeader.ps1
+++ b/functions/Read-DbaBackupHeader.ps1
@@ -111,11 +111,11 @@ Similar to running Read-DbaBackupHeader -SqlServer sql2016 -Path "C:\temp\myfile
 			{
 				if (!(Test-SqlPath -SqlServer $server -Path $file))
 				{
-					Write-Warning "File does not exist or access denied"
+					Write-Warning "File does not exist or access denied. The SQL Server service account may not have access to the source directory."
 				}
 				else
 				{
-					Write-Warning "File list could not be determined. This is likely due to the file not existing, the backup version being incompatible or unsupported, connectivity issues or tiemouts with the SQL Server, or the SQL Server service account does not have access to the file share."
+					Write-Warning "File list could not be determined. This is likely due to the file not existing, the backup version being incompatible or unsupported, connectivity issues or tiemouts with the SQL Server, or the SQL Server service account does not have access to the source directory."
 				}
 				
 				Write-Exception $_
--- a/functions/Read-DbaBackupHeader.ps1
+++ b/functions/Read-DbaBackupHeader.ps1
@@ -133,7 +133,7 @@ Also returns detailed information about each of the datafiles contained in the b
 			
 			if ($Simple)
 			{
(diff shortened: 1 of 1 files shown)


functions/Read-DbaBackupHeader.ps1
```diff
@@ -0,0 +1,153 @@
...
+			catch
+			{
+				if (!(Test-SqlPath -SqlServer $server -Path $file))
+				{
+					Write-Warning "File does not exist or access denied"
+				}
+				else
+				{
+					
+					Write-Warning "File list could not be determined. This is likely due to the file not existing, the backup version being incompatible or unsupported, connectivity issues or tiemouts with the SQL Server, or the SQL Server service account does not have access to the file share."
```
PowerDBAKlaas: 'file not existing' is supposed to be a reason in the if-block
can it also be the reason for the else-block?
both test-sqlpath -eq $true and $false are possible for a non-existing file?

functions/Read-DbaBackupHeader.ps1
```diff
@@ -0,0 +1,153 @@
...
+
+.EXAMPLE
+Read-DbaBackupHeader -SqlServer sql2016 -Path \\nas\sql\backups\mydb\mydb.bak, \\nas\sql\backups\otherdb\otherdb.bak
+
+Logs into sql2016 and reads two backup files - mydb.bak and otherdb.bak. The SQL Server service account must have rights to read this file.
+	
+.EXAMPLE
+Read-DbaBackupHeader -SqlServer . -Path C:\temp\myfile.bak -Simple
+	
+Logs into the local worksation and shows simplified output about C:\temp\myfile.bak. The SQL Server service account must have rights to read this file.
```
PowerDBAKlaas: typo: worksation -> workstation (or computer)

functions/Read-DbaBackupHeader.ps1
```diff
@@ -0,0 +1,153 @@
...
+.PARAMETER FileList
+Returns detailed information about the files within the backup	
+
+.NOTES 
+dbatools PowerShell module (https://dbatools.io, clemaire@gmail.com)
+Copyright (C) 2016 Chrissy LeMaire
+
+This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+
+This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
```
ctrlbold: Then it makes it look weird in Get-Help. Allowing PS to word wrap instead of us (because people likely wouldn't stop at 80chars) ensures it'll be pretty when using Get-Help

functions/Read-DbaBackupHeader.ps1
ctrlbold: Yes, it can be both, depending on permissions.

PR #351, opened 2016-11-28
Find-DbaOrphanedFile - speed improvements

niphlod:
Fixes:
- a subtle issue with format-path
- ft catalogs in <= 2005 (please test it)
- tempdev*.ndf wrong reporting

 - speedup (in my enterprise, a cutoff of an order of magnitude)

- [x]  Powershell 3
- [x]  Windows 7
- [x]  SQL Server 2000

- [ ]  SQL Server vNext
- [x]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally

### DIFF
--- a/functions/Find-DbaOrphanedFile.ps1
+++ b/functions/Find-DbaOrphanedFile.ps1
@@ -140,19 +140,18 @@ Finds the orphaned ending with ".fsf" and ".mld" in addition to the default file
 			# Add support for Full Text Catalogs in Sql Server 2005 and below
 			if ($server.VersionMajor -lt 10)
 			{
-				$databaselist = $smoserver.Databases | select name
+				$databaselist = $smoserver.Databases | select Name, IsFullTextEnabled
 				foreach ($db in $databaselist)
 				{
-					# Add support for Full Text Catalogs in Sql Server 2005 and below
-					if ($server.VersionMajor -lt 10)
-					{
-						$database = $db.name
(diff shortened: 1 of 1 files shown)


wsmelton:
Tested this on a SQL Server 2005 instance and it fails for some system databases and a few user databases.

Shows this message in exception:
> Full-Text Search is not enabled for the current database. Use sp_fulltext_database to enable Full-Text Search. The functionality to disable and enable full-text search for a database is deprecated. Please change your application.

It fails against master and tempdb, and then out of 230+ databases failed against 3 user database. I'm still trying to tell what could be different on those databases.

wsmelton:
Also confirm there is no speed improvements in this version, at least server I ran it on took 6 seconds longer to run. Original version shown first, and then the version in this PR.

image

niphlod:
ok, the fact that it runs slower is because on 2005 it now checks correctly FT catalogs, so it's 1 additional query per database (before they were skipped entirely). 
I'll send a patch in the next few hours to try to avoid running that query if FT is not enabled on the database, is not catching the error will be the last resort.
thanks for testing

niphlod:
@wsmelton : it seems like an easy fix. can you try the last commit ?

niphlod:
@wsmelton : did you try it ? (un)fortunaly, I have no 2005 around

wsmelton:
Sorry, just tested and all good.

ctrlbold:
sorry, i had too haha. worked well on all versions.

niphlod:
NP, just glad it's fixed.

PR #352, opened 2016-11-29
Copy-SqlLogin - Add warning for SQL logins when mixed mode not enabled on destination

### LINKED ISSUE #177: Copy-SqlLogin should warn or not copy SQL Logins if the destination doesn't have mixed mode enabled
Question raised during presentation at SQL Saturday, this cmdlet should probably warn or just not copy over SQL Logins if mixed mode authentication is not enabled on the target.

SirCaptainMitch:
Fixes #177 

 - Add warning to Copy-SqlLogin when the server is not set to mixed mode. 

How to test this code: 
- [x]  Attempt to migrate SQL Logins from Source server to a Dest Server without Mixed Mode enabled. Should write warning to console. "$Destination does not have Mixed Mode enabled. Enable this after the migration."

Tests for tester:

- [x] Works remotely
- [x] Works locally

### DIFF
--- a/functions/Copy-SqlLogin.ps1
+++ b/functions/Copy-SqlLogin.ps1
@@ -154,7 +154,7 @@ Limitations: Does not support Application Roles yet
 				
 				if($destserver.LoginMode -ne "Mixed")
 				{ 
-					Write-Output "$Destination does not have Mixed Mode enabled. Enable this after the migration."
+					Write-Warning "$Destination does not have Mixed Mode enabled. Enable this after the migration."
 					continue 
 				}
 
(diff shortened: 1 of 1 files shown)


functions/Copy-SqlLogin.ps1
```diff
@@ -152,6 +152,12 @@ Limitations: Does not support Application Roles yet
 					continue
 				}
 				
+				if($destserver.LoginMode -ne "Mixed")
+				{ 
+					Write-Output "$Destination does not have Mixed Mode enabled. Enable this after the migration."
```
dansqldba: This is probably open for debate, but I think this should be Write-Warning, to be consistent with the rest of the output.

mshamann: >Login never appears to be migrated anyway 

This seems to be a bug. I don't have an environment to test that on. I can fix the other stuff however.

dansqldba: I have opened a new issue for this #358.

PR #362, opened 2016-12-01
Initial commit of Get-Connection

PowerDBAKlaas:
Tests for tester:
- [ ] Works remotely
- [ ] Works locally

…nd computername(s)


internal/Get-Connection.ps1
```diff
@@ -0,0 +1,115 @@
+﻿Function Get-Connection
```
ctrlbold: I really like this command and how many failbacks it has. I'm happy to remove Resolve-IpAddress but Get-Connection isn't right. This does a lot of resolutions -- I think it should be Resolve-NetworkInfo (Similar to Microsoft's Resolve-DnsName)

Also, what do you think about getting the Kerberos/SPN name/entry too? I'd need that in another command.

internal/Get-Connection.ps1
```diff
@@ -0,0 +1,115 @@
...
+.SYNOPSIS
+Get a connection object from a given computer
+
+.DESCRIPTION
+Retrieves the IPAddress, ComputerName from one computer.
+The object can be used to take action against its name or IPAddress.
+
+First ICMP is used to test the connection, and get the connected IPAddress.
+
+If your local Powershell version is not higher than 2, WMI is tried to get the computername.
```
ctrlbold: If your local Powershell version is not higher than 2, 

The module is forbidden to run on versions lower than 3, as specified in the psd1. But please keep as many failbacks as possible.

PowerDBAKlaas: haa, I thought it was 2
3 is much better! good, good, good

internal/Get-Connection.ps1
PowerDBAKlaas: I named it like this because it will replace Test-Connection in most functions. I thought that would read easy, and make clear this is the step where we test the connection and collect the IP and names to work with. Whether we resolve or grow or hammer the result in it's shape is not important to the user.
But you're right it's not really the connection, it's indeed only the info.

@RamblingCookieMonster wrote this: https://gallery.technet.microsoft.com/scriptcenter/Get-SPN-Get-Service-3bd5524a

Should this be one function? Maybe if 1 SPN for a computer is selected. Would you always need the Host or RestrictedKerberosHost? or the WSMAN?
I think we should keep it separate if you could need multiple SPN's, or if we need another function anyway to get computerinfo from AD. 

Maybe it can go in Test-SqlConnection too?

ctrlbold: Agreed and agreed 👍

PowerDBAKlaas: Should we name this one Test-ComputerConnection?
- it does test both ping and CIM / WMI
- it's similar to Test-SqlConnection but at host level
- this doesn't raise te expectation of having a connection object

ctrlbold: I don't think so, no. We aren't reporting if it the connection is up or down like with Test-Connection or even Test-SqlConnection. We are simply reporting back lookups.

PowerDBAKlaas: sorry, I mixed some functions up:
It's similar to Connect-SqlServer
So, would Connect-Computer be more consisent?

Test-Sqlconnection is with the more extensive tests and info, which we're also working on.

ctrlbold: There are several ways to resolve an IP address and hostname, including using .NET objects. We should not should rely on underlying methods to determine what a function name should be. 

Initially, I thought this was going to simply resolve primary IP and Hostname/DNS. If it is used to determine how to connect to a server, then it should be Test. Though this is sounding like two commands, no?

ctrlbold: Happens all the time to me! I do like all of your ideas about Test-SqlConnection. I forgot, would Kerberos go in there? 

Yeah  I agree that this command would be complete and would also like the Test-Connection in the first one -- I find it to be fastest, but that may be my bias. You?

PowerDBAKlaas: In-Test-sqlconnection I would bundle all there is to know about settings and requirements for a connection. So, client side info can be OS, PSVersion, elevated rights, smo installed,.... and then server side, and indeed domain and network items if they are of concern. I would include Kerberos info too.

I don't understand what you mean by the second remark. You prefer separate internal Test- functions or one containing both ping and CIM?

PR #415, opened 2016-12-11
Remove-SqlOrphanUser - enhancements

ClaudioESSilva:
Check for user as schema owner

### DIFF
--- a/functions/Remove-SqlOrphanUser.ps1
+++ b/functions/Remove-SqlOrphanUser.ps1
@@ -174,8 +174,8 @@ Will remove from all databases the user OrphanUser EVEN if exists their matching
 
                             if ($Users.Count -eq 0)
                             {
-                                #the fourth validation will remove from list sql users without login. The rule here is Sid with length higher than 16 bytes, with NoLogin
-                                $Users = $db.Users | Where-Object {$_.Login -eq "" -and ($_.ID -gt 4) -and ($_.Sid.Length -gt 16 -and $_.LoginType -eq [Microsoft.SqlServer.Management.Smo.LoginType]::SqlLogin -and $_.id -gt 4) -eq $false}
+                                #the third validation will remove from list sql users without login. The rule here is Sid with length higher than 16
+                                $Users = $db.Users | Where-Object {$_.Login -eq "" -and ($_.ID -gt 4) -and (($_.Sid.Length -gt 16 -and $_.LoginType -eq [Microsoft.SqlServer.Management.Smo.LoginType]::SqlLogin) -eq $false)}
                             }
                             else
                             {
@@ -186,8 +186,8 @@ Will remove from all databases the user OrphanUser EVEN if exists their matching
 		                        }
                                 else
                                 {
-                                    #the fourth validation will remove from list sql users without login. The rule here is Sid with length higher than 16 bytes, with NoLogin
-                                    $Users = $db.Users | Where-Object {$_.Login -eq "" -and ($_.ID -gt 4) -and ($Users -contains $_.Name) -and (($_.Sid.Length -gt 16 -and $_.LoginType -eq [Microsoft.SqlServer.Management.Smo.LoginType]::SqlLogin -and $_.id -gt 4) -eq $false)}
+                                    #the fourth validation will remove from list sql users without login. The rule here is Sid with length higher than 16
+                                    $Users = $db.Users | Where-Object {$_.Login -eq "" -and ($_.ID -gt 4) -and ($Users -contains $_.Name) -and (($_.Sid.Length -gt 16 -and $_.LoginType -eq [Microsoft.SqlServer.Management.Smo.LoginType]::SqlLogin) -eq $false)}
                                 }
                             }
                         }
--- a/functions/Remove-SqlOrphanUser.ps1
+++ b/functions/Remove-SqlOrphanUser.ps1
@@ -174,7 +174,7 @@ Will remove from all databases the user OrphanUser EVEN if exists their matching
 
                             if ($Users.Count -eq 0)
                             {
-                                $Users = $db.Users | Where {$_.Login -eq "" -and ("dbo","guest","sys","INFORMATION_SCHEMA" -notcontains $_.Name)}
+                                $Users = $db.Users | Where {$_.Login -eq "" -and ("dbo","guest","sys","INFORMATION_SCHEMA" -notcontains $_.Name) -and ($_.UserType -in ([Microsoft.SqlServer.Management.Smo.UserType]::SqlLogin, [Microsoft.SqlServer.Management.Smo.UserType]::SqlUser))}
                             }
                             else
                             {
@@ -185,7 +185,7 @@ Will remove from all databases the user OrphanUser EVEN if exists their matching
 		                        }
                                 else
                                 {
-                                    $Users = $db.Users | Where {$_.Login -eq "" -and ($Users -contains $_.Name)}
+                                    $Users = $db.Users | Where {$_.Login -eq "" -and ($Users -contains $_.Name) -and ($_.UserType -in ([Microsoft.SqlServer.Management.Smo.UserType]::SqlLogin, [Microsoft.SqlServer.Management.Smo.UserType]::SqlUser))}
                                 }
                             }
                         }
--- a/functions/Remove-SqlOrphanUser.ps1
+++ b/functions/Remove-SqlOrphanUser.ps1
@@ -27,6 +27,12 @@ List of users to remove
 If exists any schema which owner is the User, this will force the change of the owner to 'dbo'.
 If exists a login to map the drop will not be performed unless you specify this parameter.
 
+.PARAMETER WhatIf
+Shows what would happen if the command were to run. No actions are actually performed. 			
+
+.PARAMETER Confirm 		
+Prompts you for confirmation before executing any changing operations within the command. 
+
 .NOTES
 Original Author: Cláudio Silva (@ClaudioESSilva)
 dbatools PowerShell module (https://dbatools.io, clemaire@gmail.com)
--- a/functions/Remove-SqlOrphanUser.ps1
+++ b/functions/Remove-SqlOrphanUser.ps1
@@ -10,7 +10,7 @@ If user is the owner of the schema with the same name and if if the schema does 
 If user owns more than one schema, the owner of the schemas that does not have the same name as the user, will be changed to 'dbo'. If schemas have underlying objects, you must specify the -Force parameter so the user can be dropped.
 If exists a login to map the drop will not be performed unless you specify the -Force parameter (only when calling from Repair-SqlOrphanUser.
 
-.PARAMETER SqlInstance
+.PARAMETER SqlServer
 The SQL Server instance.
 
 .PARAMETER SqlCredential
@@ -55,27 +55,27 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 https://dbatools.io/Remove-SqlOrphanUser
 
 .EXAMPLE
-Remove-SqlOrphanUser -SqlInstance sql2005
+Remove-SqlOrphanUser -SqlServer sql2005
 
 Will find and drop all orphan users without matching login of all databases present on server 'sql2005'
 
 .EXAMPLE
-Remove-SqlOrphanUser -SqlInstance sqlserver2014a -SqlCredential $cred
+Remove-SqlOrphanUser -SqlServer sqlserver2014a -SqlCredential $cred
 
 Will find and drop all orphan users without matching login of all databases present on server 'sqlserver2014a'. Will be verified using SQL credentials.
 
 .EXAMPLE
-Remove-SqlOrphanUser -SqlInstance sqlserver2014a -Databases db1, db2 -Force
+Remove-SqlOrphanUser -SqlServer sqlserver2014a -Databases db1, db2 -Force
 
 Will find all and drop orphan users even if exists their matching login on both db1 and db2 databases
 
 .EXAMPLE
-Remove-SqlOrphanUser -SqlInstance sqlserver2014a -Users OrphanUser
+Remove-SqlOrphanUser -SqlServer sqlserver2014a -Users OrphanUser
 
 Will remove from all databases the user OrphanUser only if not have their matching login
 
 .EXAMPLE
-Remove-SqlOrphanUser -SqlInstance sqlserver2014a -Users OrphanUser -Force
+Remove-SqlOrphanUser -SqlServer sqlserver2014a -Users OrphanUser -Force
 
 Will remove from all databases the user OrphanUser EVEN if exists their matching login. First will change any schema that it owns to 'dbo'.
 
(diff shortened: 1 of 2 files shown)


functions/Remove-SqlOrphanUser.ps1
```diff
@@ -181,49 +184,122 @@ Will remove from all databases the user OrphanUser EVEN if exists their matching
...
+
+                                    foreach ($sch in $Schemas)
+                                    {
+                                        $NumberObjects = ($db.EnumObjects() | Where-Object {$_.Schema -eq $sch.Name} | Measure-Object).Count
+
+                                        if ($NumberObjects -gt 0)
+                                        {
+                                            if ($Force)
+                                            {
+                                                Write-Output "Parameter -Force was used! The schema '$($sch.Name)' have $NumberObjects underlying objects. We will change schema owner to 'dbo' and drop the user."
```
ConstantineK: Do we want to use write-warning here?

ClaudioESSilva: On other commands we have Verbose. Changed for that. We need to talk about it and define a standard for when $force is specified.

functions/Remove-SqlOrphanUser.ps1
```diff
@@ -181,49 +184,122 @@ Will remove from all databases the user OrphanUser EVEN if exists their matching
...
-                            
-                            if ($server.versionMajor -gt 8)
-                            {
-                                $query = "DROP USER " + $User
-                            }
-                            else
-                            {
-                                $query = "EXEC master.dbo.sp_droplogin @loginame = N'$User'"
-                            }
+                            $SkippUser = $false
```
ClaudioESSilva: Corrected

ClaudioESSilva:
We don't want to drop objects because they can be in use. So the far we will go is change the owner to 'dbo' and then we can drop the user.

Regarding the auhorization, we are just replace the owner. No objects will be renamed.

ctrlbold:
hey @ConstantineK, did his commit address your concerns?

ConstantineK:
Everything but my second comment about listing the objects to the user. 
We talked a little bit about it in slack and as far as I knew there would be some additional changes he was working on to return a list of schemas that had their owner changed to dbo so that it could be piped out or used to perform further work (at the very least having a history of things changed) instead of going out and asking "which schemas are owned by dbo" and having to compare it with "which schemas do I expect to have owned by dbo"

ClaudioESSilva:
@ConstantineK Thanks on the input!
I created a PSCustomObject with the Action, Before and After schema owner.
image

@PowerDBAKlaas we have a question here for you 😄 
I kept the `Write-Output` along with the PSCustomObject. I have tested on my machine (PS v5.1) and the pipeline works just great. do you think we can leave this that way?
image

Otherwise we need to move `Write-Output` to `Write-Verbose`.. but that way, when we don't need to deal with schemas the screen will be empty..no output.. no feedback. And..I don't like that.

Please share your thoughts 😃

PowerDBAKlaas:
sorry, @ClaudioESSilva we can't output this way. You say the pipeline works great, have you tested piping to Out-Gridview, Out-Datatable or .csv?
I can't do any testing myself now, but I think most of those will not produce the desired results.

Let's say a user wants to log the outcome of Remove-SqlOrphanUser, either to a csv or into a table. We need one consistent output type to make that possible. In this case it's the PSCustomObject. If we add strings with messages to the output stream, those too will be put into the pipeline.
To separate the messages from the output objects, we have to put them in an alternative stream; 'Error', 'Warning', 'Verbose' or 'Debug', or write directly to the host. Streams are preferred.
Write-Output adds to the 'Output' (or 'Succes' or 'Standard') stream, so it could only be used to output the objects. That's what happens implicitly anyway.
Write-Output would only be an acceptable solution if there would be no other output than strings.

So, in this function, all Write-Output needs to be replaced with Write-Verbose (for optional info) or Write-Warning or Write-Error for those messages we want to show even when the user didn't explicitly asked for it. Because I do agree with you that it's neat to show at least something to the user, even if nothing was done.

@juneb explains:
https://blogs.technet.microsoft.com/heyscriptingguy/2014/03/30/understanding-streams-redirection-and-write-host-in-powershell/

functions/Remove-SqlOrphanUser.ps1
```diff
@@ -181,49 +186,150 @@ Will remove from all databases the user OrphanUser EVEN if exists their matching
...
+                                        {
+                                            if ($Force)
+                                            {
+                                                Write-Verbose "Parameter -Force was used! The schema '$($sch.Name)' have $NumberObjects underlying objects. We will change schema owner to 'dbo' and drop the user."
+
+                                                if ($Pscmdlet.ShouldProcess($db.Name, "Changing schema '$($sch.Name)' owner to 'dbo'. -Force used."))
+				                                {
+                                                    $AlterSchemaOwner += "ALTER AUTHORIZATION ON SCHEMA::[$($sch.Name)] TO [dbo]`r`n"
+
+                                                    $results += [pscustomobject]@{
```
PowerDBAKlaas: If no further manipulation of the object is needed, there is no reason to collect it in $results,
just put it in the output right away

functions/Remove-SqlOrphanUser.ps1
```diff
@@ -181,49 +186,150 @@ Will remove from all databases the user OrphanUser EVEN if exists their matching
...
+                                        {
+                                            if ($sch.Name -eq $User.Name)
+                                            {
+                                                Write-Verbose "The schema '$($sch.Name)' have the same name as user '$($User.Name)'. Schema will be dropped."
+
+                                                if ($Pscmdlet.ShouldProcess($db.Name, "Dropping schema '$($sch.Name)'."))
+                                                {
+                                                    $DropSchema += "DROP SCHEMA [$($sch.Name)]"
+
+                                                    $results += [pscustomobject]@{
```
PowerDBAKlaas: If no further manipulation of the object is needed, there is no reason to collect it in $results,
just put it in the output right away

functions/Remove-SqlOrphanUser.ps1
```diff
@@ -181,49 +186,150 @@ Will remove from all databases the user OrphanUser EVEN if exists their matching
...
+                                            }
+                                            else
+                                            {
+                                                Write-Warning "Schema '$($sch.Name)' ownership will be changed to 'dbo' so the user can be dropped."
+
+                                                if ($Pscmdlet.ShouldProcess($db.Name, "Changing schema '$($sch.Name)' owner to 'dbo'."))
+                                                {
+                                                    $AlterSchemaOwner += "ALTER AUTHORIZATION ON SCHEMA::[$($sch.Name)] TO [dbo]`r`n"
+
+                                                     $results += [pscustomobject]@{
```
PowerDBAKlaas: If no further manipulation of the object is needed, there is no reason to collect it in $results,
just put it in the output right away

functions/Remove-SqlOrphanUser.ps1
```diff
@@ -257,6 +363,7 @@ Will remove from all databases the user OrphanUser EVEN if exists their matching
 		if ($StackSource -ne "Repair-SqlOrphanUser")
         {
            Write-Output "Total Elapsed time: $totaltime"
+           return $results
```
PowerDBAKlaas: if all objects are streamed into the pipeline directly, no return $results is needed

functions/Remove-SqlOrphanUser.ps1
```diff
@@ -246,7 +352,7 @@ Will remove from all databases the user OrphanUser EVEN if exists their matching
             Write-Output "There are no databases to analyse."
```
PowerDBAKlaas: I would avoid using throw, and rather catch errors and define the output ourselves, like with Write-Error

functions/Remove-SqlOrphanUser.ps1
```diff
@@ -52,30 +55,30 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
...
+.EXAMPLE
 Remove-SqlOrphanUser -SqlServer sqlserver2014a -Users OrphanUser -Force
 
-Will remove from all databases the user OrphanUser EVEN if exists their matching login.
-	
+Will remove from all databases the user OrphanUser EVEN if exists their matching login. First will change any schema that it owns to 'dbo'.
+
 #>
 	[CmdletBinding(SupportsShouldProcess = $true)]
 	Param (
```
PowerDBAKlaas: please change parameter name to 'SqlInstance' and make 'SqlServer' an alias

functions/Remove-SqlOrphanUser.ps1
```diff
@@ -89,19 +92,19 @@ Will remove from all databases the user OrphanUser EVEN if exists their matching
 	)
 
     DynamicParam { if ($SqlServer) { return Get-ParamSqlDatabases -SqlServer $SqlServer -SqlCredential $SqlCredential } }
-	
+
 	BEGIN
 	{
         Write-Output "Attempting to connect to Sql Server.."
```
PowerDBAKlaas: Write-Output will mix with the PSCustomobject.
The message strings need to be written to an alternative stream.

functions/Remove-SqlOrphanUser.ps1
```diff
@@ -6,31 +6,34 @@ Drop orphan users with no existing login to map
...
 .PARAMETER Users
 List of users to remove
 
 .PARAMETER Force
+If exists any schema which owner is the User, this will force the change of the owner to 'dbo'.
 If exists a login to map the drop will not be performed unless you specify this parameter.
 
-.PARAMETER WhatIf 
-Shows what would happen if the command were to run. No actions are actually performed. 
+.PARAMETER WhatIf
```
PowerDBAKlaas: I think comment based help for the common parameters is auto-generated.

functions/Remove-SqlOrphanUser.ps1
```diff
@@ -89,19 +92,19 @@ Will remove from all databases the user OrphanUser EVEN if exists their matching
...
-	
+
 	BEGIN
 	{
         Write-Output "Attempting to connect to Sql Server.."
 		$server = Connect-SqlServer -SqlServer $SqlServer -SqlCredential $SqlCredential
 	}
-	
+
 	PROCESS
```
PowerDBAKlaas: multiple SqlInstances are allowed but there seems to be no foreach in PROCESS{} ?
This way processing multiple instances can only be processed when piped in with | , but not when given as parameter values with -SqlInstance.
Since connecting to $sqlserver is done in BEGIN{}, that also will only be executed once and not for every instance.
I believe the function will fail to process multiple instances anyway. Has this been tested?

functions/Remove-SqlOrphanUser.ps1
ClaudioESSilva: For now just keep it. Pester fails when help for this parameters does not exists

functions/Remove-SqlOrphanUser.ps1
ClaudioESSilva: Changed

functions/Remove-SqlOrphanUser.ps1
ClaudioESSilva: Changed to Write-Verbose

functions/Remove-SqlOrphanUser.ps1
ClaudioESSilva: Now supports multiple instances for real 😳

functions/Remove-SqlOrphanUser.ps1
ClaudioESSilva: Removed variable

functions/Remove-SqlOrphanUser.ps1
ClaudioESSilva: Removed variable

functions/Remove-SqlOrphanUser.ps1
ClaudioESSilva: Removed variable

functions/Remove-SqlOrphanUser.ps1
ClaudioESSilva: Using Write-Error instead

functions/Remove-SqlOrphanUser.ps1
ClaudioESSilva: Removed.

ctrlbold:
We just got a report that "remove-orphaneduser removes logins that is created with nologin" can we get that resolved in this PR or should that be another one?

functions/Remove-SqlOrphanUser.ps1
```diff
@@ -52,201 +55,314 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 https://dbatools.io/Remove-SqlOrphanUser
 
 .EXAMPLE
-Remove-SqlOrphanUser -SqlServer sql2005 
+Remove-SqlOrphanUser -SqlInstance sql2005
```
ctrlbold: Please continue to use -SqlServer in examples until we move to 1.0. We have a lot of screenshots on the website that I will not have time to update until there's a dedicated time to do so.

ClaudioESSilva: @ctrlbold  done!

ctrlbold:
Hey @ClaudioESSilva -- is this ready to test? Do you have some sample test proposals I can run?

ClaudioESSilva:
@ctrlbold Will be after my following commit! Then I will add script to test it.

ClaudioESSilva:
After talking with @Staggerlee011 I have done some more tests and I found that the last commit does not address the issue.

This happens because, as you can see on the following image, an real orphan user and a UserWithoutLogin have the same properties. The only difference is the SID length.
image

With this extra validation we can exclude all users without login that in matter of fact are not orphan users.

ClaudioESSilva:
@ctrlbold to test use the T-SQL script (at the end) to generate:
 - Database named 'dbOrphanUsers'
 - 2 logins
   - UserOrphan
   - UserNOTOrphan
 - 3 users (one without login)
   - UserOrphan
   - UserNOTOrphan
   - UserWithoutLogin

You can use the following code to view the same table I have shown before on this thread:
```
$srv  = Connect-DbaSqlServer -SqlServer sql2008
$srv.Databases["dbOrphanUsers"].Users | Select-Object Name, LoginType, UserType, State, HasDBAccess, AuthenticationType, Login, @{ Name="SIDLength"; Expression={$_.Sid.Length}} | Format-Table
```

Then, run the command on this database and **ONLY** UserOrphan should be removed.
`Remove-SqlOrphanUser -SqlServer sql2008 -Databases dbOrphanUsers `

**T-SQL script**
````
USE [master]
GO

CREATE DATABASE dbOrphanUsers
GO


CREATE LOGIN [UserOrphan] WITH PASSWORD=N'123', DEFAULT_DATABASE=[master], DEFAULT_LANGUAGE=[us_english], CHECK_EXPIRATION=OFF, CHECK_POLICY=OFF
GO

USE [dbOrphanUsers]
GO
CREATE USER [UserOrphan] FOR LOGIN [UserOrphan]
GO



/* Drop login so user TestOrphan1 will be orphan */
USE [master]
GO
DROP LOGIN [UserOrphan]
GO


CREATE LOGIN [UserNOTOrphan] WITH PASSWORD=N'123', DEFAULT_DATABASE=[master], DEFAULT_LANGUAGE=[us_english], CHECK_EXPIRATION=OFF, CHECK_POLICY=OFF
GO

/* This is a user that will not be orphan */
USE [dbOrphanUsers]
GO
CREATE USER [UserNOTOrphan] FOR LOGIN [UserNOTOrphan]
GO

/* Create a user without login. NOTE: This is **NOT** a orphan user */
USE [dbOrphanUsers]
GO
CREATE USER [UserWithoutLogin] WITHOUT LOGIN WITH DEFAULT_SCHEMA=[dbo]
GO
````

ctrlbold:
That was a really thorough test, thanks so much for that, @ClaudioESSilva! I tested and it worked -- but I also changed 2 verboses to outputs until we go with objects in the future. For a few times there, I didn't even realize the command ran properly because i was expecting output.

PR #436, opened 2016-12-16
initial commit of New-DbaDatabaseSnapshot

niphlod:
How to test this code: 
- [ ] start testing with one/two instances and one/two databases, to see if any of the core method fail or warns
- [ ] carefully enlarge the set
- [ ] the most error-prone part for the code is figuring out the entire filestructure. Try throwing at it your most intricated structures
- [ ] test if a database with FileStream gets partially snapshotted (with -Force)

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally

### DIFF
--- a/functions/New-DbaDatabaseSnapshot.ps1
+++ b/functions/New-DbaDatabaseSnapshot.ps1
@@ -51,7 +51,7 @@ You should have received a copy of the GNU General Public License along with thi
 .EXAMPLE
 New-DbaDatabaseSnapshot -SqlServer sqlserver2014a -Database HR, Accounting
 
-Creates snapshot all supported databases, returning a custom object displaying Server, Database, DatabaseCreated, SnapshotOf, SizeMB, DatabaseCreated, IsReadCommittedSnapshotOn, SnapshotIsolationState
+Creates snapshot for HR and Accounting, returning a custom object displaying Server, Database, DatabaseCreated, SnapshotOf, SizeMB, DatabaseCreated, Status, Notes
 
 .EXAMPLE
 New-DbaDatabaseSnapshot -SqlServer sqlserver2014a -Databases HR -Name '_snap'
@@ -225,14 +225,16 @@ Creates snapshots for HR and Accounting databases, storing files under the F:\sn
 						{
 							Continue
 						}
-						foreach($file in $fg.Files) {
+						foreach($file in $fg.Files)
+						{
 							$counter += 1
 							# fixed extension is hardcoded as "ss", which seems a "de-facto" standard
 							$fname = [io.path]::ChangeExtension($file.Filename, "ss")
 							$fname = Join-Path (Split-Path $fname -Parent) ("{0}_{1}" -f $DefaultSuffix, (Split-Path $fname -Leaf))
 
 							# change path if specified
-							if($FilePath.Length -gt 0) {
+							if($FilePath.Length -gt 0)
+							{
 								$basename = Split-Path $fname -Leaf
 								# we need to avoid cases where basename is the same for multiple FG
 								$basename = '{0:0000}_{1}' -f $counter, $basename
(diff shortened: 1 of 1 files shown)


niphlod:
found a bug. db with filestream fail. need to check it before starting the snap process and require -Force to create a "partial" snapshot leaving FS out of it.

niphlod:
fixed the case where a Filestream is in the database

ctrlbold:
Unfortunately, on .create(), I'm getting: Failed to update database "Application_Registry_Service_DB_cde90313b64848408b625cb9b083eaf3_20161219_120245" because the database is read-only.

niphlod:
ok, we need to change the approach. 
The underlying problem is that SMO is somewhat bugged and *can* issue multiple statements, but we can't anticipate beforehand if all of them are plain wrong or just "additional details" we can't foresee. 
Given though the first is always the one actually creating the snapshot and all of the following are "post-creation" statements, I'd go for a double try{}catch{} (the first with what SMO suggests, if it fails just the first statement it generates) and we'll track the "partial" execution with a "status" column. In the "notes" section, if it's a known case, we'll add progressively what we find out (e.g. Filestream Groups, Read-Only database, etc)
On top of that, to be able to receive user inputs, I'll add a Write-Debug with the generated statement so we can figure out ourselves without requiring the user to trace the error at the backend what is going on under the hood of SMO.

niphlod:
@ctrlbold : could you try again please ?

ctrlbold:
PS C:\Users\ctrlb.BASE> New-DbaDatabaseSnapshot -SqlInstance sqlserver -Databases Application_Registry_Service_DB_cde90313b64848408b625cb9b083eaf3


Server          : sqlserver
Database        : Application_Registry_Service_DB_cde90313b64848408b625cb9b083eaf3_20161219_215033
SnapshotOf      : Application_Registry_Service_DB_cde90313b64848408b625cb9b083eaf3
SizeMB          : 9.75
DatabaseCreated : 12/19/2016 12:50:34 PM
Status          : Partial
Notes           :

Unsure what Partial means? otherwise, works!

niphlod:
dang. Partial means something along the line errored. But if that is a ReadOnly db, it should have been accompanied by a "proper" Notes column.

ctrlbold:
notes are empty D: want access to my lab?

niphlod:
please try one more time, if it doesn't work I'll need to review the logic in a proper 2016 instance

ctrlbold:
So sorry, still partial and did a copy / paste of new code. this is 2016

niphlod:
at least with -Debug you can now inspect what SMO is trying to do ^_^ . lemme fire up a remote connection to a 2016 instance and report back

niphlod:
image
 this is what I get on mine, which is 13.0.4001.0, for a snapshot on a read-only database

ctrlbold:
Hey @niphlod - this works beautifully in 2005-2014, but fails in both 2016 and vnext. Can you put in a check -- if version is < 2016 use the whole statement, but if more then just use the first statement? or alternatively, check for a -match of "ALTER DATABASE SCOPED CONFIGURATION" and remove that array item from the resulting script.

niphlod:
It should still create the snap...or doesn't it?
The whole point is having feedback if something doesn't go as SMO plans,
but the snap is created nonetheless, and work around before issuing the
creation if it fails even to create the snap.

Il 24 dic 2016 01:23, "Chrissy LeMaire" <notifications@github.com> ha
scritto:

ctrlbold:
Yes, it works, but I'm left with the impression that it didn't. And it's not being honest -- the snapshot was created just fine but SMO has a bug where it tries to do something it shouldn't. I'm going to merge then modify.

PR #438, opened 2016-12-16
initial commit of Find-DbaDatabase

Staggerlee011:
v1, sure theres many code and layout issues.. functionality input welcome to :)

### DIFF
--- a/functions/Find-DbaDatabase.ps1
+++ b/functions/Find-DbaDatabase.ps1
@@ -1,37 +1,200 @@
-﻿FUNCTION Find-DbaDatabase
+﻿function Find-DbaDatabase
 {
+<#
+.SYNOPSIS
+Find database/s on multiple servers that match critea you input
+
+.DESCRIPTION
+Allows you to search Sql Instances for database that have either the same name or service broker guid. You can use the like operator for database name, but this cannot be used in the service broker search.
+There a several reasons for the service broker guid not matching on a restored database primarily using alter database new broker. or turn off broker to return a guid of 0000-0000-0000-0000. 
+
+.PARAMETER SqlServer
+The SQL Server that you're connecting to.
+
+.PARAMETER Credential
+Credential object used to connect to the SQL Server as a different user
+
+.PARAMETER Type
+What sort of search would like to complete. Either Service Broker GUID or Database name. 
+
+.PARAMETER Name
+Value that is searched for
+
+.PARAMETER Like
+Allows you to search database name using *<NAME>*
+
+.PARAMETER Detailed
+Output a more detailed view showing ComputerName, SqlInstance, Database, ServiceBrokerGUID, Tables, StoredProcedures,Views and ExtendedProperties to see they closely match to help find related databases.
+
+.NOTES
+Author: Stephen Bennett: https://sqlnotesfromtheunderground.wordpress.com/
+
+dbatools PowerShell module (https://dbatools.io)
+Copyright (C) 2016 Chrissy LeMaire
+This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+You should have received a copy of the GNU General Public License along with this program. If not, see http://www.gnu.org/licenses/.
+
+.LINK
+ https://dbatools.io/Find-DbaDatabase
+
+.EXAMPLE
+Find-DbaDatabase -SqlInstance "DEV01", "DEV02", "UAT01", "UAT02", "PROD01", "PROD02" -Type Database -Name TestDB -Detailed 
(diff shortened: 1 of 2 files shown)


niphlod:
hum. Sure this will end up with "please add a parameter to check for 'y' being equal to 'z' " ...

Can we just offer a -Name and a -Filter and pass along the real db oject so it can be filtered with any property OR with a proper Where down the pipeline ?

Staggerlee011:
i like it! 

only quesiton i have name set for the databasename, should i change that to Database (or DatabaseName) 
and would the name be FilterName or Name?

Staggerlee011:
hmm actually, are you saying do a search for say Where $_.<filter> -eq $FilterValue ? instead of the current that is trying to find copies of a database via name or service broker guid?  (so all are based on Where $_.Name -eq $Database). 

I could change the Type param to be Database, Service_Broker_GUID, Filter) ? that way you could search for all databases that are online or all database with autoclose off

niphlod:
actually dunno the specifics. 
I'm just trying to anticipate what users would ask for a general "find my database". 
I for sure go for an explicit parameter to restrict the name name, but how 'bout "I want to see every database with autoshrink, readonly, etc etc etc" ? 
The -Filter I was thinking about is really a string repr of a Where (just to cut off the number of elements passed down the pipe). 
ATM I can't think about a structure in Powershell that gives some flexibility, so maybe just a string that gets invoked ?

functions/Find-DbaDatabase.ps1
```diff
@@ -0,0 +1,200 @@
...
+
+.EXAMPLE
+Find-DbaDatabase -SqlInstance "DEV01", "DEV02", "UAT01", "UAT02", "PROD01", "PROD02" -Type Service_Broker_GUID -Name 25b64fef-faeb-495a-9898-f25a782835f5 -Detailed 
+Returns all database from the SqlInstances that have the same Service Broker GUID with a deeatiled output
+
+#>
+	[CmdletBinding()]
+	param (
+		[Parameter(Mandatory = $true, ValueFromPipeline = $true)]
+		[Alias('cn','host','computer','server')]
```
ctrlbold: these are the aliases for windows centric commands. Please use the ones for SQL Server

[Alias("ServerInstance", "SqlServer", "Instance")]
		[string[]]$SqlInstance,

functions/Find-DbaDatabase.ps1
```diff
@@ -0,0 +1,200 @@
...
+        {
+            try
+	        {
+	            Write-Verbose "Connecting to $Instance"
+                $srv = Connect-SqlServer -SqlServer $Instance -SqlCredential $sqlcredential
+	        }
+	        catch
+	        {
+	            Write-Warning "Failed to connect to: $srv"
+                break
```
ctrlbold: mmmmm should this be break or continue? I need to write that blog post about these behaviors.

functions/Find-DbaDatabase.ps1
```diff
@@ -0,0 +1,200 @@
...
+        [switch]$Detailed
+	)
+    process
+    {
+        foreach ($Instance in $SqlInstance)
+        {
+            try
+	        {
+	            Write-Verbose "Connecting to $Instance"
+                $srv = Connect-SqlServer -SqlServer $Instance -SqlCredential $sqlcredential
```
ctrlbold: please always use $server

functions/Find-DbaDatabase.ps1
```diff
@@ -0,0 +1,200 @@
...
+            {
+                $computername = $srv.Name
+                $inst =  ""
+            }
+
+        if ($Type -eq 'Database')
+        {
+            if ($Like)
+            {
+                $match = $srv.Databases | Where-Object {$_.Name -like "*$name*"}
```
ctrlbold: Should we auto detect * and % instead of making them specify Like? And just get rid of this lil block, @PowerDBAKlaas ?

functions/Find-DbaDatabase.ps1
```diff
@@ -0,0 +1,200 @@
...
+		                       Value = $db.ExtendedProperties[$xp.Name].Value
+                                }
+                            $outep += $extdetails
+                        }
+                    }
+                    else
+                    {
+                        $outep = ""
+                    }
+                    $out = [PSCustomObject]@{
```
ctrlbold: No need to assign the variable since it needs to manipulation after (like with Select-DefaultItem

Just send [PSCustomObject]@{ right now to the pipeline

functions/Find-DbaDatabase.ps1
```diff
@@ -0,0 +1,200 @@
...
+#>
+	[CmdletBinding()]
+	param (
+		[Parameter(Mandatory = $true, ValueFromPipeline = $true)]
+		[Alias('cn','host','computer','server')]
+		[string[]]$SqlInstance,
+		[System.Management.Automation.PSCredential]$SqlCredential,
+        [parameter(Mandatory = $true)]
+        [ValidateSet('Database','Service_Broker_GUID')]
+        [string]$Type,
```
ctrlbold: This should have a default value or Database and not be required, no?

functions/Find-DbaDatabase.ps1
PowerDBAKlaas: I think that has to be "Failed to connect to $instance" as $srv will be empty when it fails, no?

PR #441, opened 2016-12-17
Initial commit of Get-DbaTrigger

PowerDBAKlaas:
How to test this code: 

- [ ]  PowerShell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally

### DIFF
--- a/functions/Get-DbaTrigger.ps1
+++ b/functions/Get-DbaTrigger.ps1
@@ -1,0 +1,93 @@
+﻿Function Get-DbaTrigger
+{
+<#
+.SYNOPSIS
+Get all existing triggers on one or more SQL instances.
+
+.DESCRIPTION
+Get all existing triggers on one or more SQL instances.
+
+Default output includes columns ComputerName, SqlInstance, Database, TriggerName, IsEnabled and DateLastMofied.
+
+.PARAMETER SqlInstance
+The SQL Instance that you're connecting to.
+
+.PARAMETER Credential
+Credential object used to connect to the SQL Server as a different user.
+
+.NOTES
+Author: Klaas Vandenberghe ( @PowerDBAKlaas )
+
+dbatools PowerShell module (https://dbatools.io)
+Copyright (C) 2016 Chrissy LeMaire
+This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+You should have received a copy of the GNU General Public License along with this program. If not, see http://www.gnu.org/licenses/.
+
+.LINK
+ https://dbatools.io/Get-DbaTrigger
+
+.EXAMPLE
+Get-DbaTrigger -SqlInstance ComputerA\sql987
+
+Returns a custom object displaying ComputerName, SqlInstance, Database, TriggerName, IsEnabled and DateLastMofied.
+
+.EXAMPLE
+Get-DbaTrigger -SqlInstance 'ComputerA\sql987','ComputerB'
+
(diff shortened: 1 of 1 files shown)


PowerDBAKlaas:
Dilemma:
output objects are serverddltrigger and / or databaseddltrigger, but contain the same properties
this is not neat, but the fastest way
Should I leave this as is or convert to PSCustomobject anyway, even if that has a performance hit?
@ctrlbold @wsmelton ?

ctrlbold:
Yeah, I think the PSCustomObject will be better and only add a couple milliseconds

PowerDBAKlaas:
New Dilemma:
Inspired by the discussion on Test-DbaValidLogin,
I used 'Enabled ' and 'Disabled' as values for [Status],
instead of True/False for [IsEnabled].

The latter is the original, so, no manipulation required, and most space efficient,
yet the more verbose is clearer and keeps its meaning without context (headers).
Also because MS pairs IsEnabled with IsDisabled columns here and there, I feel showing a readable value is better.

It's very much like the OLTP vs DWH logic; there's no right or wrong but each has its purpose.

Opinions?

functions/Get-DbaTrigger.ps1
```diff
@@ -0,0 +1,117 @@
...
+.LINK
+ https://dbatools.io/Get-DbaTrigger
+
+.EXAMPLE
+Get-DbaTrigger -SqlInstance ComputerA\sql987
+
+Returns a custom object displaying ComputerName, SqlInstance, Database, TriggerName, IsEnabled and DateLastMofied.
+
+.EXAMPLE
+Get-DbaTrigger -SqlInstance 'ComputerA\sql987','ComputerB'
```
ctrlbold: Please keep using SqlServer in examples until 1.0. It will be easy to do mass replaces.

functions/Get-DbaTrigger.ps1
```diff
@@ -0,0 +1,117 @@
...
+			    continue
+		        }
+
+            Write-Verbose "Getting Server Level Triggers on $Instance"
+            $server.Triggers | 
+            ForEach-Object {
+                    [PSCustomObject]@{
+                            ComputerName     = $server.NetName
+                            SqlInstance      = $server.ServiceName
+                            TriggerLevel     = "ServerLevel"
```
ctrlbold: Level is repeated. It will suffice to say TriggerLevel Server or Database instead of TriggerLevel ServerLevel, DatabaseLevel

functions/Get-DbaTrigger.ps1
```diff
@@ -0,0 +1,117 @@
...
+                            Database         = ""
```
ctrlbold: Please use $null instead of "". It's prettier.

ctrlbold:
I see what your'e saying about Status vs IsEnabled. Putting it up for a vote. For now, that works.

functions/Get-DbaTrigger.ps1
PowerDBAKlaas: True
Done

functions/Get-DbaTrigger.ps1
PowerDBAKlaas: Are you sure we can automate that?
[SqlServer] will have to be replaced by [ComputerName] in one function and by [SqlInstance] in an other.
If we change back to 'sqlserver' everywhere now, and it turns out we'll have to review them one by one in the end, I feel we're taking two steps backward.

functions/Get-DbaTrigger.ps1
PowerDBAKlaas: Oops I did it again

Corrected

functions/Get-DbaTrigger.ps1
ctrlbold: Yes, I'll be using Notepad++ and will have to do this for all other commands. This will be just one of over one hundred, and I'd prefer to keep changes minimal until we're introduce the mass breaking changes.

PR #444, opened 2016-12-17
Initial commit for Restore-DbaFromDatabaseSnapshot

niphlod:
How to test this code: 
- [ ] start testing with one/two instances and one/two databases, to see if any of the core method fail or warns
- [ ] carefully enlarge the set
- [ ] the most error-prone part for the code is figuring out the steps that need to be made to restore to a particular snapshot (as you can't have more than one snap for db if you want to restore from them). Try carefully with the -WhatIf (with or without -Force) to inspect if the logic is correct

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [x]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally

### DIFF
--- a/functions/Restore-DbaFromDatabaseSnapshot.ps1
+++ b/functions/Restore-DbaFromDatabaseSnapshot.ps1
@@ -17,9 +17,11 @@ Credential object used to connect to the SQL Server as a different user
 
 .PARAMETER Databases
 Restores from the last snapshot databases with this names only
+NB: you can pass either Databases or Snapshots
 
 .PARAMETER Snapshots
-Restores databases from snapshots with this names only. NB: it overrides the -Databases parameter
+Restores databases from snapshots with this names only
+NB: you can pass either Databases or Snapshots
 
 .PARAMETER WhatIf
 Shows what would happen if the command were to run. No actions are actually performed.
@@ -41,7 +43,7 @@ This program is distributed in the hope that it will be useful, but WITHOUT ANY 
 You should have received a copy of the GNU General Public License along with this program. If not, see http://www.gnu.org/licenses/.
 
 .LINK
- https://dbatools.io/Restore-DbaDatabaseSnapshot
+ https://dbatools.io/Restore-DbaFromDatabaseSnapshot
 
 .EXAMPLE
 Restore-DbaFromDatabaseSnapshot -SqlServer sqlserver2014a
@@ -118,16 +120,19 @@ Restores databases from snapshots named HR_snap_20161201 and Accounting_snap_201
 			if ($snapshots.count -eq 0 -and $databases.count -eq 0)
 			{
 				# Restore all databases from the latest snapshot
+				Write-Verbose "Selected all databases"
 				$dbs = $all_dbs | Where-Object IsDatabaseSnapshot -eq $true
 			}
 			elseif ($databases.count -gt 0)
 			{
 				# Restore only these databases from their latest snapshot
+				Write-Verbose "Selected only databases"
 				$dbs = $all_dbs | Where-Object { $databases -contains $_.DatabaseSnapshotBaseName }
 			}
 			elseif ($snapshots.count -gt 0)
 			{
 				# Restore databases from these snapshots
+				Write-Verbose "Selected only snapshots"
 				$dbs = $all_dbs | Where-Object { $snapshots -contains $_.Name }
 				$base_databases = $dbs | Select-Object -ExpandProperty DatabaseSnapshotBaseName | Get-Unique
 				if($base_databases.count -ne $snapshots.count)
@@ -188,7 +193,7 @@ Restores databases from snapshots named HR_snap_20161201 and Accounting_snap_201
 			{
                 # Check if there are FS, because then a restore is not possible
                 $all_FS = $server.Databases[$op['to']].FileGroups | Where-Object FileGroupType -eq 'FileStreamDataFileGroup'
-                if($all_FS.Count -gt 0) 
+                if($all_FS.Count -gt 0)
                 {
                     Write-Warning "Database '$($op['to'])' has FileStream group(s). You cannot restore from snapshots"
                     [PSCustomObject]@{
--- a/functions/Restore-DbaFromDatabaseSnapshot.ps1
+++ b/functions/Restore-DbaFromDatabaseSnapshot.ps1
@@ -1,0 +1,291 @@
+﻿Function Restore-DbaFromDatabaseSnapshot
+{
+<#
+.SYNOPSIS
+Restores databases from snapshots
+
+.DESCRIPTION
+Restores the database from the snapshot, discarding every modification made to the database
+NB: Restoring to a snapshot will result in every other snapshot of the same database to be dropped
+It also fixes some long-standing bugs in SQL Server when restoring from snapshots
+
+.PARAMETER SqlInstance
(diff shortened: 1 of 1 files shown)


functions/Restore-DbaFromDatabaseSnapshot.ps1
```diff
@@ -0,0 +1,291 @@
...
+The SQL Server that you're connecting to
+
+.PARAMETER Credential
+Credential object used to connect to the SQL Server as a different user
+
+.PARAMETER Databases
+Restores from the last snapshot databases with this names only
+
+.PARAMETER Snapshots
+Restores databases from snapshots with this names only. NB: it overrides the -Databases parameter
```
PowerDBAKlaas: Here it says snapshots parameter overrides databases but in the code there's a termination when both exist.

functions/Restore-DbaFromDatabaseSnapshot.ps1
```diff
@@ -0,0 +1,291 @@
...
+Author: niphlod
+
+dbatools PowerShell module (https://dbatools.io)
+Copyright (C) 2016 Chrissy LeMaire
+This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+You should have received a copy of the GNU General Public License along with this program. If not, see http://www.gnu.org/licenses/.
+
+.LINK
+ https://dbatools.io/Restore-DbaDatabaseSnapshot
```
PowerDBAKlaas: The name of the link is not the same as the function name.

functions/Restore-DbaFromDatabaseSnapshot.ps1
```diff
@@ -0,0 +1,291 @@
...
+			}
+
+			$all_dbs = $server.Databases
+
+			# vault to hold all programmed operations from --> to
+			$operations = @()
+
+			if ($snapshots.count -eq 0 -and $databases.count -eq 0)
+			{
+				# Restore all databases from the latest snapshot
```
PowerDBAKlaas: I see a lot of opportunities for Write-Verbose, here and further in the script.

functions/Restore-DbaFromDatabaseSnapshot.ps1
```diff
@@ -0,0 +1,291 @@
...
+
+			foreach($op in $operations)
+			{
+                # Check if there are FS, because then a restore is not possible
+                $all_FS = $server.Databases[$op['to']].FileGroups | Where-Object FileGroupType -eq 'FileStreamDataFileGroup'
+                if($all_FS.Count -gt 0) 
+                {
+                    Write-Warning "Database '$($op['to'])' has FileStream group(s). You cannot restore from snapshots"
+                    [PSCustomObject]@{
+						Server   = $Server.Name
```
PowerDBAKlaas: ComputerName = $server.NetName
SqlInstance = $server.ServiceName

niphlod:
some concerns have been addressed. 
I'd refrain to use Write-Verbose "everywhere", seeing it as an "INFO" channel given Write-Host is out of the options and Write-Output should only be used (correctly) for passing down the pipeline objects rather than a "debug" one. 
At this point maybe the need for "ultra-verbosity" can be better addressed by Write-Debug?

Also, on the object output format, I tried to comply with the other two commands are available yet for managing snapshot. 
This is a proper "styleguide" issue (and I'll comply as soon as something is written about it).

My 2 cents: I don't see meaningful knowing what computername I was connected when I created snapshot (or, really, anything that has to do with databases only, which I view as a service offering rather than a database hosted in some infrastructure)
SqlInstance may be a better name for $server.Name (so it's consistent with the parameter) but a column filled with either "MSSQLSERVER" or "namedinstance" imho just takes space for nothing ...

PowerDBAKlaas:
The returned object may be consumed according to different wishes of the user.
One of the most likely is they want to redirect to a file or table as a means of logging or for auditing. Seen both instance name and databasename can be the same over multiple computers, the only way to uniquely identify the database concerned, is in the hierarchy ComputerName > SqlInstance > Database. Those properties are indispensable.
Or they may be working on multiple computers simultaneously. Seeing the computername can prevent serious mistakes.

niphlod:
There must be something I'm missing... I'm assuming that $server.name is always the string that matches the EXPLICIT parameter they passed in. How can it be of use to return not that parameter but two different properties that don't match the ones passed in ?

If instead we're talking about outputting "more" properties (i.e. "in addition to" $sqlinstance, opposed to "instead of $sqlinstance" ), ok, probably I'll add them as hidden.

I'd still have something to say about a a usecase where you issue a command that DOESN'T interact in any way with the underlying OS returning what host it was run into. In the cloud era and PaaS it's even more unhelpful. If anyone creates a database on a mssql instance (think, failover) what good does it make for auditing to know the current node it was on to ?

PowerDBAKlaas:
ha, yes, 'in addition to' is better

a very good point about cases of HA, it may require additional logic,
but 'the age of clouds and PaaS' don't arrive simultaneously in every company; it does not make all 'ancient' setups dissapear.

the returned properties are not needed for the function itself, but to make a flexible and consistent user experience, like for redirecting or logging

niphlod:
hidden properties will be added, both here and in other *snapshot* functions, so they'll get consistent.
Just to avoid to code something wrong, is it agreed the following mapping:

- SqlInstance : $server.Name
- InstanceName: $server.ServiceName
- ComputerName : $server.NetName 

?

PowerDBAKlaas:
That sounds like the ideal solution!

I'm going to rewrite my functions to comply to those namings.

Thanks!

ctrlbold:
I like all of those 3 names and would like them in all functions, but usually the 2 will be hidden. Working on Select-DefaultItems to support Exclude to make this all easier :D

niphlod:
Can't we just have the objects with all of three and select-default only
sqlinstance for commands that have that parameter and/or the other for the
ones that interact with the OS? Imho all of the three aren't usable if e.g.
no mssql is involved

Il 22 dic 2016 14:43, "Chrissy LeMaire" <notifications@github.com> ha
scritto:

ctrlbold:
yah totally, sorry, was shortcutting. ComputerName when it's a Windows function, and all 3 with 2 hidden when it's a SQL Server function.

niphlod:
@ctrlbold : did we pin for styleguide the 

> 
> SqlInstance : $server.Name
> InstanceName: $server.ServiceName
> ComputerName : $server.NetName

part ?

@dansqldba : given I should rewrite the output of all *databasesnapshot* commands, is it better to raise 4 PRs for the same goal or just one with the 4 modifications ?

ctrlbold:
I'm going to be changing all commands to 

> SqlInstance : $server.Name
> InstanceName: $server.ServiceName
> ComputerName : $server.NetName

in one giant or a few giant PRs, coming very very soon.

PowerDBAKlaas:
@ctrlb will fix the output properties ComputerName, SqlInstance and InstanceName.

PR #445, opened 2016-12-18
Initial commit of Get-DbaXEventsSession

PowerDBAKlaas:
Tests for tester:
- [ ] Works remotely
- [ ] Works locally

### DIFF
--- a/functions/Get-DbaXEventsSession.ps1
+++ b/functions/Get-DbaXEventsSession.ps1
@@ -58,7 +58,7 @@ Returns a custom object with ComputerName, SQLInstance, Session, StartTime, Stat
 	{
 		if ([System.Reflection.Assembly]::LoadWithPartialName("Microsoft.SqlServer.Management.XEvent") -eq $null)
 		{
-			throw "SMO version is too old. To collect Extended Events, you must have Shared Management Objects 2008 R2 or higher installed."
+			throw "SMO version is too old. To collect Extended Events, you must have SQL Server Management Studio 2012 or higher installed."
 		}
 		$sessions = $psboundparameters.Sessions
 	}
--- a/functions/Get-DbaXEventsSession.ps1
+++ b/functions/Get-DbaXEventsSession.ps1
@@ -4,11 +4,7 @@
 Get a list of Extended Events Sessions
 
 .DESCRIPTION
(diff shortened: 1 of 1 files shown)


functions/Get-DbaXEventsSession.ps1
```diff
@@ -0,0 +1,104 @@
...
+                Write-Warning " Failed to connect to $instance ."
+                continue
+            }
+            if($server.versionmajor -ge '11')
+            {
+                Write-Verbose "Getting XEvents Sessions on $instance ."
+                if ($instance -notlike '*\*') { $instance = "$instance\DEFAULT" }
+                try
+                {
+                    Get-ChildItem SQLSERVER:\XEvent\$instance\sessions -ErrorAction SilentlyContinue -WarningAction SilentlyContinue |
```
ctrlbold: This is a reliance on SQLPS and the Provider which we do not support. Please see  Copy-SqlExtendedEvent for the SMO code and connect example. It's pretty intense, but easy to copy/paste.

SMO is loaded up in the psm1 (Microsoft.SqlServer.Management.XEvent)

PowerDBAKlaas:
- requirements in Copy-SqlExtendedEvent say major version -ge 10, and XEvents should be there from 2008 on, right? But the function failed on my SQL2008R2 instances. @MikeFal also used -ge 11. Can anyone shed some light on this?

- I copied the warning on 'version too old', but it used to say 'you need SSMS 2008R2 or newer installed'. I think we need another wording, but I can't formulate it right. Can anyone help on this, please?
Technically, requirements are: 1. CLR Types, 2. SMO. Both could be installed via client SDK on SQL install, or via the SQL Feature Pack. That's probably not the message we want to display?

ctrlbold:
hey @PowerDBAKlaas, I had searched for sql 2008 extended events and saw a title of a blog post that was misleading -- you're right, it's 11 and not 10 or even 10.5. I'll update.

While technically, it is possibly true that you may need only SMO and CLR types (and the english sucks on that message), it will be hell supporting people who only install those two. I used to say only SMO and CLR but recently changed it to SSMS because so much of dbatools doesn't work without all of the the libraries SSMS provides. I'm hoping that Microsoft releases some kind of package with SqlServer module that we can point to that.

PR #451, opened 2016-12-20
Initial commit of Get-DbaSqlService

PowerDBAKlaas:
How to test this code: 

- [ ]  PowerShell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally


ctrlbold:
My understanding is that you do not want any ugly red throws -- only pretty warnings.

image

Is it possible to have Get-Service as a failback? That would help this work on even Windows 2003.

PowerDBAKlaas:
get-Service may be a possibility, but requirements are different and the filtering is complexer.
I will look into it.

functions/Get-DbaSqlService.ps1
```diff
@@ -0,0 +1,87 @@
...
+                                Sort-Object Name -Descending | Select-Object -First 1
+                    Write-Verbose "Getting Cim class SqlService in Namespace $($namespace.Name) on $Computer"
+                    try
+                    {
+                        Get-CimInstance -ComputerName $Computer -Namespace $("root\Microsoft\SQLServer\" + $namespace.Name) -Query "SELECT * FROM SqlService" -ErrorAction SilentlyContinue |
+                        Select-Object @{l='ComputerName';e={$_.HostName}}, ServiceName, DisplayName, StartName,
+                         @{l='TypeDescr';e={switch($_.SQLServiceType){1 {'Database Engine'} 2 {'SQL Agent'} 3 {'Full Text Search'} 4 {'SSIS'} 5 {'SSAS'} 6 {'SSRS'} 7 {'SQL Browser'} 8 {'Unknown'} 9 {'FullTextFilter Daemon Launcher'}}}},
+                         @{l='status';e={switch($_.state){ 1 {'Stopped'} 2 {'Start Pending'}  3 {'Stop Pending' } 4 {'Running'}}}},
+                         @{l='startmodus';e={switch($_.startmode){ 1 {'unknown'} 2 {'Automatic'}  3 {'Manual' } 4 {'Disabled'}}}}
```
niphlod: I'd go for "Status" instead of "status" (capital "s") and "Startup" instead of "startmodus", which is more common "naming" in monitoring-services world, or "StartType", which complies with the naming returned by get-service (it's services this command is returning, after all)

PowerDBAKlaas: oh yes, true
those names are still those from my own Dutch version.

I looked up the difference between 'State' and 'Status', seems you need a degree in Language Sciences to understand that.

I find 'StartMode' in the CIM class SqlService, 'ServiceStartMode and 'StartType' in the Object returned by Get-Service, 'Startup Type' in the Services MMC, 'StartMode' in Win32_Service.

I also renamed 'TypeDescr' to 'SqlServiceType'.

It's probably usually best to use the original property name when it's clear and concise enough, and only choose something else if it's a real improvement,
or if it's more consistent with our project wide choices, like ComputerName for HostName.

PowerDBAKlaas:
as for a fail back method, I'll also research the Cim-session method.

ctrlbold:
Unfortunately, this command has too high of a rate of failure. It works beautifully on my sql2012+ machines (which are probably running Windows 2012 server) and I really do love all of the output. But it did not work for me for 2000-2008. 

Please consider also using the internal command Invoke-MangementComputerCommand which uses SQLWMI and ScriptBlocks. You can see it in use in Get-SqlStartupParameter (or is it DBA?). I'd also venture to guess it would be easier just making the SELECT an object even if it's slightly less efficient (and maybe not -- that select is intense.)

ctrlbold:
Closing but not deleting branch

wsmelton:
I would not say the high rate of failure is bad, it is consider what OS is supported. CIM commands don't exist on anything less than Window Server 2012/Windows 8. So any OS under that would have to utilize WMI.

PowerDBAKlaas:
I finally found a Windows 2003 server with SQL 2005 on it, and PS V2.

CIM cmdlets are delivered with PS V3 and are only needed on the client.

Get-DbaSqlService -ComputerName Win2003 works perfect here.
Can you find out what step exactly fails?
The only failure I get is from servers on which I am not local admin.

ConstantineK:
Reopening as this is now in a state ready for review and submitting.

ctrlbold:
I'm iffy on prepending everything with Sql. I think we should go the other way around and be like Get-DbaWinSomething if it's Windows. Get-DbaProcess follows this approach, as does Get-DbaMemoryUsage. This command should be renamed to Get-DbaService. 

Can you return a hidden object that can be piped to Stop, Start and Restart Service (not sure if that's possible)? I also noticed that you had some filters up in there -- `SQLServiceType = switch($_.SQLServiceType){1 {'Database Engine'} 2 {'SQL Agent'} 3 {'Full Text Search'} 4 {'SSIS'} 5 {'SSAS'} 6 {'SSRS'} 7 {'SQL Browser'} 8 {'Unknown'} 9 {'FullTextFilter Daemon Launcher'}}`

First, can you rename that column to SqlServiceType and second, can you make that a tabbable/validated param called -Type so that people can get like, Only database engine, Only Agent, etc?

Is there any benefit to using CIM over Get-Service -ComputerName and -Credential?

ctrlbold:
Also, thanks again! I'm extra excited about this command and will be using it a lot myself. It will help demystify the SQL Services that are available on remote machines. 💯

PowerDBAKlaas:
image
This is Service.


image
This is SqlService.

image
It returns the SqlService Class.

I'll have to research if this can be reused to start/stop or set configuration.
We know sql services should only be altered via Configuration manager, because it does more than when altered via Services. So I do want to implement that if it is possible.

The column's name is SqlServiceType. What exactly do you want to be renamed? ❓ 

I'm gonna start with the -Type filter. 😄

ctrlbold:
Ah ha! Makes sense.

`ServiceType` works. I think we should give ourselves some leeway and use the column names as a guide but not a definite. Like PSComputerName we turn into ComputerName. Alternatively, if you want to have both, you can `Select-DefaultView -Property SqlServiceType as ServiceType, etc, etc, etc`. And it'll just add a ServiceType alias to SqlServiceType then hide it.

Can we add a -WindowsService param in v2.0 and use it to get all services using Get-Service? That one is actually way more reliable for restarting if i recall. SQLWMI reports back too quickly (though I could be wrong. One of them I had to accommodate to wait and confirm)

I look fwd to the Type filter!

ctrlbold:
hey @PowerDBAKlaas - I know you wanted to get this in but Type is not there and the column has not been renamed. Did you hit sync with your commits?

Also, I ran `Get-DbaSqlService -ComputerName sql2000 -Verbose` and didn't get any errors but also didn't get any services. What are the minimum reqs?

ClaudioESSilva:
@PowerDBAKlaas I have run it again and it gives an error but it seems legit since is not configured. (based on all attempts)

image

I think is missing a TRY/CATCH to handle the error and output a custom error message:
image

ctrlbold:
beautiful! everything has been addressed. works with pipes. works with localhost. works with 100% of my servers and gives a warning for sql 2000 namespaces. Love the command, thank you, @PowerDBAKlaas!

PR #470, opened 2016-12-23
Add rename functionality to Copy-SqlLogin and add new function Update-SQlLoginName

### LINKED ISSUE #410: Copy-SqlLogin - allow different name on destination
**Do you want to request a *feature* or report a *bug*?**
Feature

**What is the current behavior?**
Copies same name between source and destination

**If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem.**

**What is the expected behavior?**

**Which versions of Powershell and which OS are affected by this issue? Did this work in previous versions of our scripts?**

ctrlb [8:36 AM]  
i can see it, yeah, sorta like how Copy-SqlDatabase should do the same. It'd be a new param that accepts a hashtable

mshamann:
Fixes #410 

Tested on windows 10, powershell 5 and SQL Server 2008R2

 - Adds a hashtable parameter to the copy-sqllogin function to rename multiple users.
 - Adds Update-SqlLoginName function that is called in copy-sqlloging, and for user as stand alone 

How to test this code: 
- [ ] run copy-sqllogin with rename for windows ( where SID matches ) and an SQL Server login 
- [ ] run Update-SQlLoginName to change a user on a server 

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally


dbatools.psd1
```diff
@@ -171,7 +171,8 @@
 		'Resolve-DbaNetworkName',
 		'Test-DbaValidLogin',
 		'Get-DbaMemoryUsage',
-		'Export-DbaAvailabilityGroup'
+		'Export-DbaAvailabilityGroup',
+		'Update-SqlLoginName'
```
ctrlbold: Please remove the .psd1 change from the PR. .psd1's can only be made by me since they always create conflicts.

mshamann:
Done

functions/Copy-SqlLogin.ps1
```diff
@@ -356,6 +366,15 @@ Limitations: Does not support Application Roles yet
 				{
 					Update-SqlPermissions -sourceserver $sourceserver -sourcelogin $sourcelogin -destserver $destserver -destlogin $destlogin
 				}
+				if ($rename.Keys -contains $username) {
```
ctrlbold: Very nice! You really simplified this task. Can you do the same for Databases? ;) mmm that one would be harder, though because of the files inside, boo.

Please rename Update-SqlLoginName to Update-DbaLoginName -- both within the file and within the filename. 

Can you please research the variable name that people usually use for a rename hash? I'm guessing it would be LoginRenameHash? Something more specific than Hashtable.

Otherwise, dope AF! Thank you 🥇

functions/Update-SqlLoginName.ps1
```diff
@@ -0,0 +1,113 @@
...
+.PARAMETER SqlInstance 
+Allows you to login to servers using SQL Logins as opposed to Windows Auth/Integrated/Trusted. To use:
+
+$scred = Get-Credential, then pass $scred object to the -SourceSqlCredential parameter. 
+
+Windows Authentication will be used if DestinationSqlCredential is not specified. SQL Server does not accept Windows credentials being passed as credentials. 	
+To connect as a different Windows user, run PowerShell as that user.
+
+.PARAMETER Username 
+The current username on the server
```
ctrlbold: Should be Login and NewLogin ?

functions/Update-SqlLoginName.ps1
```diff
@@ -0,0 +1,113 @@
...
+	param (
+		[parameter(Mandatory = $true)]
+		[object]$SqlInstance,
+		[System.Management.Automation.PSCredential]$SqlInstanceCredential,
+		[parameter(Mandatory = $true)]
+		[String]$UserName, 
+		[parameter(Mandatory = $true)]
+		[String]$NewUserName
+	)
+	DynamicParam { if ($SqlInstance) { return (Get-ParamSqlLogins -SqlServer $SqlInstance -SqlCredential $SqlInstanceCredential) } }
```
ctrlbold: I'd remove this for now -- it's an auto completer but it accepts multiples which wont work for you. please remind me to create a login dynamic param.

functions/Update-SqlLoginName.ps1
```diff
@@ -0,0 +1,113 @@
...
+	{	
+		$sourceserver = Connect-SqlServer -SqlServer $SqlInstance -SqlCredential $SqlInstanceCredential					
+		$Databases = $sourceserver.Databases
+		$currentUser = $sourceserver.Logins[$UserName]
+		
+	}
+	PROCESS
+	{
+		Write-Output "Changing Login name from $userName to $NewUserName"		
+		try {
```
ctrlbold: This is really good stuff, but -WhatIf would fail

Please check out other commands to see how to support whatif. It's basically wrapping an If around some try catches

If ($Pscmdlet.ShouldProcess(wherverver_youre_executing, A_message_about_what_you're_doing) { try...catch..etc. }

mshamann:
Updated based on @ctrlbold suggestions. 

1. Renamed new function to Rename-DbaLogin 
2. Support for -WhatIf Switch parameter 
3. Renamed Parameters to be more verbose and accurate. 
4. Added/Updated examples.

ctrlbold:
It looks likely this should be approved in next release closer to 2/1 since it will take a bit more testing since it's somewhat complex. Thanks so much for all of the effort 👍

ClaudioESSilva:
Hi @mshamann  I have some questions here.
I have done some tests but I don't have sure if I agree with the result.

Let me try to explain:
Rename is done after copy, which means if already have a login with the original name in the destination instance this will not be copy unless we specify the `-Force` . This could bring some issues, if I use `-Force` and use the new parameter to rename from 'Test' to 'TestRenamed' it will run without error but it does what you, as user, are expecting? Not for me.
1 - User 'Test' user was dropped only user 'TestRenamed' remains
2 - If I try to copy user Test so I can keep both will give an error because SID is already in use (SID is in use by TestRenamed)
3 - If I try to use the -Force parameter will do nothing because user with name Test does not exists.

In this scenario how can I have both users?
Should we, in this scenario, create a new user (that will generate new SID)  and give it the new name?

Of course, this could happen if we use actual Copy-SqlLogin and rename it after, then go back and try to copy it again. The SID error will occurs. 

We want to address this? Or am I complicating? 😳 
Thoughts? @ctrlbold @ConstantineK @PowerDBAKlaas @wsmelton @niphlod

PowerDBAKlaas:
@ClaudioESSilva the current behaviour is indeed not correct.
The solution is not so easy, since there may be different scenario's.
Creating the new user seems mostly the right choice.
When we're starting with one login on origin server, and it becomes two users on destination server, only one can have the SID. Does a user want the login with the same name to have it or the login that is a copy from the origin? I don't know that.

SirCaptainMitch:
@ClaudioESSilva @PowerDBAKlaas  

I guess when I wrote it the idea of ending up with two logins at the end had not occurred to me. Which is why I wrote it this way, pull down the user the same way and then simply rename the login. Because in my head a rename parameter is simply that. Rename the object you specify, not to duplicate and leave the old one. 

To end up with two logins might be easy for the SQL Login, but not so for the Windows, given the SID problem. 

So we would have to Create a new login in the destination, then sync the permissions from original login to the new login. 

I am not sure how we would handle the Windows side of things in this case.

ctrlbold:
That seems like an edge case that we can support in the next release. Approving for now 👍

ctrlbold:
Nice work @SirCaptainMitch! <3

PR #479, opened 2016-12-27
function does not exit anymore if there is no ICMP reply

PowerDBAKlaas:
Fixes # 

 - because ICMP may be blocked by Firewall while WSMan is not, we now continue even if there is no ICMP reply
 - 
 - 

How to test this code: 
- [ ] 
- [ ] 

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally


ctrlbold:
worked with 100% of my servers, old and new <3

PR #480, opened 2016-12-27
added fixes to New-DbaDatabaseSnapshot

ctrlbold:


niphlod:
I'd go for a different route rather than blindly executing a part of the smo script on >= 2016 . 
Just stick with what we (dbas) know: if snap.create() successfully creates a snap, even if it errors out ("SMO" and "atomicity" here kinda sucks, but we can leverage it), leave it as "success" and no notes (until someone reports that we SHOULD put some notes). 
The only "partial" will then be for databases with FileStream filegroups.

ctrlbold:
K, can you fix? You have permissions to that branch and can clone the repo

niphlod:
should work out fine this way.

niphlod:
BTW: select-defaultfield doesn't seem to support pipelining (i.e. @{} | select-defaultField ) at least in this branch........ something cooking in other ones ?

PR #481, opened 2016-12-27
Copy-SqlDatabase cleanup

ctrlbold:
added file deletes, set default to 3 files, validated range

### DIFF
--- a/functions/Copy-SqlDatabase.ps1
+++ b/functions/Copy-SqlDatabase.ps1
@@ -1,6 +1,6 @@
 Function Copy-SqlDatabase
 {
-<# 
+<#
 .SYNOPSIS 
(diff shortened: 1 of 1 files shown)


functions/Copy-SqlDatabase.ps1
```diff
@@ -87,6 +87,9 @@ Prompts you for confirmation before executing any changing operations within the
 .PARAMETER DbPipeline
 Takes dbobject from pipeline
 
+.PARAMETER NumberFiles
+Number of files to split the backup. Default is 1.
```
alonsoh: This will need to be changed to 3 instead of 1.

ctrlbold: Oh, thanks!

PR #483, opened 2016-12-27
Resolve-DbaServerName - added last failsafe

ctrlbold:


functions/Resolve-DbaNetworkName.ps1
```diff
@@ -114,13 +114,24 @@ Returns a custom object displaying InputName, ComputerName, IPAddress, DNSHostNa
...
+				}
+				
+				if (!$conn)
+				{
+					Write-Verbose "Cim and Wmi both failed. Defaulting to .NET"
+					$fqdn = ([System.Net.Dns]::GetHostEntry($Computer)).HostName
+					$hostname = $fqdn.Split(".")[0]
+					
+					$conn = [PSCustomObject]@{
+						Name = $Computer
```
PowerDBAKlaas: nice addition

but this object needs to have the same properties as the other one.

ctrlbold: It just builds $conn then lets the final object use its properties

PR #499, opened 2017-01-03
fix for restore-dbafromsnapshot

ctrlbold:
fixes errors when using -snapshots that were caused by it deleting all snapshots, including the snapshot being restored. @niphlod please approve/merge.


niphlod:
LGTM

PR #504, opened 2017-01-06
Created Test-DbaOptimizeAdHoc check

scinobe:
Tests for tester:
- [x] Works remotely
- [x] Works locally


ctrlbold:
Everything looks great. The only thing that will likely change in the final are the column names and the message (ill remove the quotes around the 1).

ctrlbold:
image

I think we should probably add an IsBestPractice like a few of the others? we need to discuss this in style.

Issue #502, opened 2017-01-06
Set-DbaDatabaseOwner uses wrong variable - Can't use SQL authentication


ClaudioESSilva:
Is this a bug:

- Powershell Version: 4.0
- OS Version: 2012 R2 Standard

**Steps to Reproduce**
`$cred = Get-Credential
Set-DbaDatabaseOwner -SqlServer serverName -SqlCredential $cred -Databases dbName`

**Action Results**
Returns the following error
![image](https://cloud.githubusercontent.com/assets/19521315/21726747/ec7eb312-d435-11e6-984b-b4e7825e1a10.png)

**Expected Results**
Connection to instance could be made and -Databases parameter could be used.

**Additional comments**
The error in code is here:  https://github.com/sqlcollaborative/dbatools/blob/master/functions/Set-DbaDatabaseOwner.ps1#L75 should be `$SqlCredential` instead of `$SourceSqlCredential`

I have changed that locally and it worked as expected.

ClaudioESSilva:
Fixed on PR: https://github.com/sqlcollaborative/dbatools/pull/505

SirCaptainMitch:
Closing, this was merged.

PR #505, opened 2017-01-06
Fix set dbadatabaseowner

ClaudioESSilva:
Fixes the issue #502


ClaudioESSilva:
Please see the issue https://github.com/sqlcollaborative/dbatools/issues/502
Use that example to test it 😄

PR #506, opened 2017-01-07
updated spwhoisactive version

ctrlbold:


ctrlbold:
Very small change. Tested on 2 servers and works.

PR #507, opened 2017-01-07
Get-DbaConfiguration - reads sys.configuration data using SMO

sirsql:
Fixes # 

 - New function that uses SMO to read the sp_configure/sys.configurations data

How to test this code: 
- [ ] Get-DbaConfiguration -SqlServer localhost
- [ ] 

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally

### DIFF
--- a/functions/Get-DbaConfiguration.ps1
+++ b/functions/Get-DbaConfiguration.ps1
@@ -1,0 +1,108 @@
+﻿FUNCTION Get-DbaConfiguration
+{
+<#
+.SYNOPSIS
+Returns all server level system configuration (sys.configuration/sp_configure) information
+
+.DESCRIPTION
+This function returns server level system configuration (sys.configuration/sp_configure) information. The information is gathered through SMO Configuration.Properties and be returned either in detailed or standard format
+
+.PARAMETER SqlServer
+SQLServer name or SMO object representing the SQL Server to connect to. This can be a
+collection and recieve pipeline input
+
+.PARAMETER SqlCredential
+PSCredential object to connect as. If not specified, current Windows login will be used.
+
+.PARAMETER Detailed
+Returns more information about the configuration settings than standard
+
+.NOTES 
+Original Author: Nic Cain, https://sirsql.net/
+dbatools PowerShell module (https://dbatools.io, clemaire@gmail.com)
+Copyright (C) 2016 Chrissy LeMaire
+This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+You should have received a copy of the GNU General Public License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.	
+
+.LINK
+https://dbatools.io/Get-DbaConfiguration
+
+.EXAMPLE
+Get-DbaConfiguration -SqlServer localhost
+Returns server level configuration data on the localhost (ServerName, ConfigName, DisplayName, ConfiguredValue, CurrentlyRunningValue)
+
+.EXAMPLE
+Get-DbaConfiguration -SqlServer localhost -Detailed
+Returns detailed information on server level configuration data on the localhost (ServerName, ConfigName, DisplayName, Description, IsAdvanced, IsDynamic, MinValue, MaxValue, ConfiguredValue, CurrentlyRunningValue)
+
+.EXAMPLE
+'localhost','localhost\namedinstance' | Get-DbaConfiguration
+Returns system configuration information on multiple instances piped into the function
+
+#>
+	[CmdletBinding()]
+	Param (
+		[parameter(Position = 0, Mandatory = $true, ValueFromPipeline = $True)]
+		[Alias("ServerInstance", "SqlInstance", "SqlServers")]
+		[string[]]$SqlServer,
+		[System.Management.Automation.PSCredential]$SqlCredential,
+		[switch]$Detailed
+	)
+	BEGIN
+	{
+		$output = @()
+	}
+	PROCESS
+	{
+		FOREACH ($Server in $SqlServer)
+		{
+			TRY
+			{
+				$server = Connect-SqlServer -SqlServer $server -SqlCredential $sqlcredential
+			}
+			CATCH
+			{
+				Write-Verbose "Failed to connect to: $Server"
+				continue
+			}
+			
+			#Get a list of the configuration property parents, and exlude the Parent, Properties values
+			$objList = get-member -InputObject $server.Configuration -MemberType Property -Force | select Name | where { $_.Name -ne "Parent" -and $_.Name -ne "Properties" }
+			
+			#Iterate through the properties to get the configuration settings
+			foreach ($prop in $objList)
+			{
+				$PropInfo = $server.Configuration.$($prop.Name)
+				
+				#Ignores properties that are not valid on this version of SQL
+				if (!([string]::IsNullOrEmpty($PropInfo.RunValue)))
+				{
+					$output += [pscustomobject]@{
+						ServerName = $server.Name
+						ConfigName = $($prop.Name)
+						DisplayName = $PropInfo.DisplayName
+						Description = $PropInfo.Description
+						IsAdvanced = $PropInfo.IsAdvanced
+						IsDynamic = $PropInfo.IsDynamic
+						MinValue = $PropInfo.Minimum
+						MaxValue = $PropInfo.Maximum
+						ConfiguredValue = $PropInfo.ConfigValue
+						CurrentlyRunningValue = $PropInfo.RunValue
+					}
+				}
+			}
+			
+			if ($Detailed -eq $true)
+			{
+				$output | Sort-Object ServerName, DisplayName
+			}
+			else
+			{
+				$output | Sort-Object ServerName, DisplayName | Select-Object ServerName, ConfigName, DisplayName, ConfiguredValue, CurrentlyRunningValue
+			}
+		}
+	}
+}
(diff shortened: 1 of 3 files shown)


functions/Get-DbaConfiguration.ps1
```diff
@@ -0,0 +1,108 @@
+﻿FUNCTION Get-DbaConfiguration
```
PowerDBAKlaas: I feel we need to put "sql" or "instance" somewhere in the name, because the OS and other stuff have configurations too.

functions/Get-DbaConfiguration.ps1
```diff
@@ -0,0 +1,108 @@
...
+						IsDynamic = $PropInfo.IsDynamic
+						MinValue = $PropInfo.Minimum
+						MaxValue = $PropInfo.Maximum
+						ConfiguredValue = $PropInfo.ConfigValue
+						CurrentlyRunningValue = $PropInfo.RunValue
+					}
+				}
+			}
+			
+			if ($Detailed -eq $true)
```
PowerDBAKlaas: I feel when we have 'only' 10 properties, we shouldn't have a -detailed switch.
 This goes for some existing functions too.

functions/Get-DbaConfiguration.ps1
```diff
@@ -0,0 +1,108 @@
...
+			{
+				$output | Sort-Object ServerName, DisplayName
```
PowerDBAKlaas: We must avoid sorting inside a function.
It comes with a performance penalty, 
maybe the user doesn't care about the order, or
worse: he'll pipe again to sort-object

functions/Get-DbaConfiguration.ps1
```diff
@@ -0,0 +1,108 @@
...
+			{
+				Write-Verbose "Failed to connect to: $Server"
+				continue
+			}
+			
+			#Get a list of the configuration property parents, and exlude the Parent, Properties values
+			$objList = get-member -InputObject $server.Configuration -MemberType Property -Force | select Name | where { $_.Name -ne "Parent" -and $_.Name -ne "Properties" }
+			
+			#Iterate through the properties to get the configuration settings
+			foreach ($prop in $objList)
```
PowerDBAKlaas: I don't get the reason for the looping, the if, nor the propinfo.

wouldn't
```
$server.configuration.Properties |
 SELECT @{N="ComputerName";e={$server.NetName}},DisplayName, RunValue,...
```
suffice?

Unless I'm missing something, the next 30 lines are not needed.

functions/Get-DbaConfiguration.ps1
```diff
@@ -0,0 +1,108 @@
...
+	Param (
+		[parameter(Position = 0, Mandatory = $true, ValueFromPipeline = $True)]
+		[Alias("ServerInstance", "SqlInstance", "SqlServers")]
+		[string[]]$SqlServer,
+		[System.Management.Automation.PSCredential]$SqlCredential,
+		[switch]$Detailed
+	)
+	BEGIN
+	{
+		$output = @()
```
PowerDBAKlaas: We shouldn't collect objects, but put them immediately in the pipeline.

functions/Get-DbaConfiguration.ps1
sirsql: Get-DbaInstanceConfiguration be better?
Would indicate that these are instance level configs, rather than the database level configs that exist in 2016+

functions/Get-DbaConfiguration.ps1
sirsql: The looping was needed to handle returning the parent property name (which is how the configuration value is stored in SMO). There is no parent property object in $Server.configuration.properties, so this was the only way to handle that.

functions/Get-DbaConfiguration.ps1
sirsql: Happy to remove.

functions/Get-DbaConfiguration.ps1
sirsql: Will remove the sort.

functions/Get-DbaConfiguration.ps1
sirsql: Are there any examples I could take a look at for this?

functions/Get-DbaConfiguration.ps1
ctrlbold: I based the name off of the proposed name from the official Microsoft board, which was likely based off of SMO. I'm good with keeping it. Alternatively, we can do what I did in a few other commands: Get-DbaSpConfigure 

Copy-SqlSpConfigure
Export-SqlSpConfigure
Import-SqlSpConfigure

Not a huge fan of Instance but have no discernible reason.

functions/Get-DbaConfiguration.ps1
```diff
@@ -0,0 +1,108 @@
...
+		FOREACH ($Server in $SqlServer)
+		{
+			TRY
+			{
+				$server = Connect-SqlServer -SqlServer $server -SqlCredential $sqlcredential
+			}
+			CATCH
+			{
+				Write-Verbose "Failed to connect to: $Server"
+				continue
```
ctrlbold: this should be write-warning

functions/Get-DbaConfiguration.ps1
```diff
@@ -0,0 +1,108 @@
...
+			}
+			
+			#Get a list of the configuration property parents, and exlude the Parent, Properties values
+			$objList = get-member -InputObject $server.Configuration -MemberType Property -Force | select Name | where { $_.Name -ne "Parent" -and $_.Name -ne "Properties" }
```
ctrlbold: I would actually just do  $server.Configuration 

When I make Get-DbaDatabase, I'll send $server.Databases immediately to the pipeline -- it looks like that's what Get-SqlDatabase does too, actually. So ultimately, this command can have as few as 5 lines. I'll submit a PR for Get-DbaDatabase so you can see. 

@PowerDBAKlaas please review my next command. I'd like to make sure we're on the same page.

@sirsql thanks so much and welcome to the team! 👍 💯 🥇

functions/Get-DbaConfiguration.ps1
ctrlbold: This new PR does it -- #508

functions/Get-DbaConfiguration.ps1
```diff
@@ -0,0 +1,108 @@
+﻿FUNCTION Get-DbaConfiguration
+{
+<#
+.SYNOPSIS
+Returns all server level system configuration (sys.configuration/sp_configure) information
```
ctrlbold: I really like your docs btw!

functions/Get-DbaConfiguration.ps1
```diff
@@ -0,0 +1,108 @@
...
+		[System.Management.Automation.PSCredential]$SqlCredential,
+		[switch]$Detailed
+	)
+	BEGIN
+	{
+		$output = @()
+	}
+	PROCESS
+	{
+		FOREACH ($Server in $SqlServer)
```
ctrlbold: I also noticed-- this should be named differently. like foreach ($instance in $sqlserver) 


then a few lines below, $server = connect-sqlserver -sqlserver $instance

functions/Get-DbaConfiguration.ps1
sirsql: Sort removed in latest commit

functions/Get-DbaConfiguration.ps1
sirsql: Removed switch

functions/Get-DbaConfiguration.ps1
sirsql: Changes to push directly to the pipeline

functions/Get-DbaConfiguration.ps1
sirsql: Updated function to Get-DbaSpConfigure

functions/Get-DbaConfiguration.ps1
sirsql: Fixed

functions/Get-DbaConfiguration.ps1
sirsql: Fixed

sirsql:
Several changes based on feedback.

- Changed name to Get-DbaSpConfigure to better reflect what it pulls
- Outputs directly to the pipeline, no more loading up a variable
- Removed the sorts
- Removed the detailed switch, now returns all columns
- Fixed the write-verbose so that it is now a write-warning
- Fixed case where $server was being used incorrectly

Have left in the loop as it's the only way to capture the parent of the configuration, and this is important should a user wish to set them.Important reference point.

PowerDBAKlaas:
Ha, now I understand what you're saying.
It's the configname.
Of course that's a valid reason.

@ctrlbold we leave the ComputerName - SqlInstance -InstanceName for march, right?

ctrlbold:
Naw, I just accepted it ;) March is right around the corner and I didn't have time to switch everything back. I'll update the website banner when i get a chance.

PowerDBAKlaas:
what do you mean? It should be changed now?

ctrlbold:
The code is good as is. And we'll accept PRs with both SqlServer and SqlInstance then standardize in march.

PowerDBAKlaas:
Hi Nic

I have an additional question: wouldn't it be nice to have an extra property 'IsDefault'?
$false for every configuration.property where [RunningValue] or [ConfiguredValue] is not the DefaultValue.
That may be a handy property to filter on for a review,
and I see that's what sp_Blitz and SSMS standard reports do too.

sirsql:
That's a great idea, I'll add that.

wsmelton:
@sirsql if you want I had written code that contained all the default values of SQL Server versions (pulled from BOL) a few years ago. We could build this out into an internal function to store that info, then utilize in this command to more easily find the "default".

[blog article](http://blog.wsmelton.info/2013-12-04-ps-check-instanceconfig-part-1/)
[Actual code](https://github.com/wsmelton/scripts/blob/master/posh/Check-InstanceConfig.ps1)

Mind you it was years ago so it probably can be cleaned up and optimized.

sirsql:
Thank @wsmelton I was just checking and finding out that the server default values are not stored, so a function like that will be really useful. I'll have a dig through your code.

sirsql:
Added the default configurations for each version in a new internal function. Added a call to that in the Get-DbaSpConfigure function, now returns the defaults, and whether or not the running value matches the default.

ctrlbold:
Fantastic, useful command with wonderful output! Great teamwork too. Ran it against all my servers from 2000 to vnext with success! Wonderful work, @sirsql

PR #519, opened 2017-01-10
Initial commit for Get-DbaServerPermission

Staggerlee011:
v1


ctrlbold:
hey @Staggerlee011, thanks for the PR! This looks like a subset of what Get-DbaPermission, however. What would differentiate it?

Staggerlee011:
hey @ctrlbold 

Get-DbaPermissions doesn't return any of the roles that users are in only object lvl permissions, where this returns both,  There was a conversation in slack about trying to create something like sp_dbpermissions where a few people thought it was a good idea..

PR #520, opened 2017-01-10
Get-DbaDatabasePermission,Get-DbaServerPermission

Staggerlee011:
v1


ctrlbold:
This also looks like a Subset of Get-DbaPermission. Was there anything that sets it apart? If so, we should consider adding to Get-DbaPermission.

Staggerlee011:
Same again i guess, Both just mean to have an easy to understand permission layout that contains both roles and specific permissions for people to see, happy to drop them if you think they are needed. just thought i would share :) in case they where liked :)

ctrlbold:
@Staggerlee011  @PowerDBAKlaas did yall get this all figured out?

PR #521, opened 2017-01-10
Find-DbaLoginInGroup

Staggerlee011:
v1


wsmelton:
@Staggerlee011 this does not look like one of the templates was used to create the function? It is missing the help block and all.

PR #522, opened 2017-01-10
Get-DbaLinkedServer

Staggerlee011:
v1

### DIFF
--- a/functions/Get-DbaLinkedServer.ps1
+++ b/functions/Get-DbaLinkedServer.ps1
@@ -10,11 +10,8 @@ Retrieves information about each linked server on the instance
 .PARAMETER SqlServer
 The SQL Server that you're connecting to.
 
-.PARAMETER Credential
+.PARAMETER SqlCredential
 Credential object used to connect to the SQL Server as a different user
-
-.PARAMETER Databases
-Return information for only specific databases
 
 .NOTES
 Author: Stephen Bennett ( https://sqlnotesfromtheunderground.wordpress.com/ )
@@ -32,7 +29,7 @@ https://dbatools.io/Get-DbaLinkedServer
 .EXAMPLE
 Get-DbaLinkedServer -SqlServer DEV01
 
-Returns all Linked Servers for the SQL instance DEV01
+Returns all Linked Servers for the SQL Server instance DEV01
 
 #>
 	[CmdletBinding()]
@@ -54,25 +51,15 @@ Returns all Linked Servers for the SQL instance DEV01
 	        catch
 	        {
 	            Write-Warning "Failed to connect to: $srv"
-                break
+                continue
 	        }
-            if($server -contains "\")
-            {
-                $computername = ($server.Name).Split("\")[0]
-                $inst =  ($server.Name).Split("\")[1]
-            }
-            else
-            {
-                $computername = $server.Name
-                $inst =  ""
-            }
 
             foreach ($ls in $server.LinkedServers)
             {               
 
                     $output = [PSCustomObject]@{
-		                ComputerName = $computername
-                        SqlInstance = $inst
+                        ComputerName = $server.NetName
+                        SqlInstance = $server.InstanceName
 		                LinkedServerName = $ls.Name
                         RemoteServer = $ls.DataSource
                         ProductName = $ls.ProductName 
--- a/functions/Get-DbaLinkedServer.ps1
+++ b/functions/Get-DbaLinkedServer.ps1
@@ -1,0 +1,90 @@
+﻿function Get-DbaLinkedServer
+{
+<#
+.SYNOPSIS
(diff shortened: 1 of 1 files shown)


functions/Get-DbaLinkedServer.ps1
```diff
@@ -0,0 +1,90 @@
...
+                        RemoteServer = $ls.DataSource
+                        ProductName = $ls.ProductName 
+                        Impersonate = $ls.LinkedServerLogins.Impersonate
+                        RemoteUser = $ls.LinkedServerLogins.remoteuser
+                        Rpc = $ls.Rpc
+                        RpcOut = $ls.RpcOut
+                        LinkedServer = $ls
+                        }
+     
+                    Select-DefaultField -InputObject $output -Property ComputerName, SqlInstance, LinkedServerName, RemoteServer, ProductName, Impersonate, RemoteUser, Rpc, RpcOut
```
SirCaptainMitch: Not sure if I am missing something on this line, but this errors out. 

`Select-DefaultField : The term 'Select-DefaultField' is not recognized as the name of a cmdlet, function, script file, or 
operable program.`

wsmelton: @SirCaptainMitch that is an internal function...if you are testing it just make sure you have that command in your dbatools\internals folder. It was just added about 6 days ago.

functions/Get-DbaLinkedServer.ps1
```diff
@@ -0,0 +1,90 @@
...
+	        }
+            if($server -contains "\")
+            {
+                $computername = ($server.Name).Split("\")[0]
+                $inst =  ($server.Name).Split("\")[1]
+            }
+            else
+            {
+                $computername = $server.Name
+                $inst =  ""
```
wsmelton: Best to have this as `$inst = $null`

functions/Get-DbaLinkedServer.ps1
```diff
@@ -0,0 +1,90 @@
...
+        {
+            try
+	        {
+	            Write-Verbose "Connecting to $Instance"
+                $server = Connect-SqlServer -SqlServer $Instance -SqlCredential $sqlcredential
+	        }
+	        catch
+	        {
+	            Write-Warning "Failed to connect to: $srv"
+                break
```
wsmelton: This should be `continue`. Using `break` will cause it to go out of the foreach loop if one of many servers is passed in and you hit one that throws an exception.

functions/Get-DbaLinkedServer.ps1
wsmelton: As well you can actually  just remove the else all together. 

Also you don't need to use the extra parenthesizes to split it, and just another way to do it:

```powershell
if ($server.Contains("\") {
   $computername = $server.Name.Split("\")[0]
   $inst = $server.Name.Split("\")[1]
}
```

The above will ensure you still get the computer name, and if you deal with a default instance, the `$inst` will be `null` and not an empty string.

functions/Get-DbaLinkedServer.ps1
```diff
@@ -0,0 +1,90 @@
...
+You should have received a copy of the GNU General Public License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+	
+.LINK
+https://dbatools.io/Get-DbaLinkedServer
+
+.EXAMPLE
+Get-DbaLinkedServer -SqlServer DEV01
+
+Returns all Linked Servers for the SQL instance DEV01
```
wsmelton: This should read

> Returns all Linked Servers for the SQL Server instance DEV01

functions/Get-DbaLinkedServer.ps1
```diff
@@ -0,0 +1,90 @@
...
+.DESCRIPTION
+Retrieves information about each linked server on the instance
+
+.PARAMETER SqlServer
+The SQL Server that you're connecting to.
+
+.PARAMETER Credential
+Credential object used to connect to the SQL Server as a different user
+
+.PARAMETER Databases
```
wsmelton: This parameter should be removed since it is not being used.

functions/Get-DbaLinkedServer.ps1
```diff
@@ -0,0 +1,90 @@
...
+.SYNOPSIS
+Gets all linked servers and summary of information from the sql servers listed
+
+.DESCRIPTION
+Retrieves information about each linked server on the instance
+
+.PARAMETER SqlServer
+The SQL Server that you're connecting to.
+
+.PARAMETER Credential
```
wsmelton: Update the parameter name to `$SqlCredential` to match the parameters for the function.

ctrlbold:
These get commands should return full on SMO objects ala Get-DbaDatabase (which I haven't finished ;) I will approve and fix

PR #527, opened 2017-01-11
Handle NULL values passed into Out-DbaDataTable

scinobe:
When NULL values were being passed to Out-DbaDataTable through the pipe
and an error was generated.  Testing for NULL did not match.  I test for
.length instead and it works.  When the length of the property value is
0, it does not add the value to the data table row item.  The row will
then contain a NULL value for that field.

### DIFF
--- a/functions/Out-DbaDataTable.ps1
+++ b/functions/Out-DbaDataTable.ps1
@@ -109,7 +109,10 @@ Similar to above but $dbalist gets piped in
 				}
 				else
 				{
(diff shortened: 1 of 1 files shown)


ctrlbold:
Tested with

```
Get-DbaLastGoodCheckDb -SqlServer (Get-SqlRegisteredServerName -SqlServer sql2014) | Out-DbaDataTable
```

Lookin good!

PR #530, opened 2017-01-12
Get-DbaDatabaseFreespace, added Try block to server connect & fixed bug with NULL Sizes

scinobe:
The server connect code was not in a try block.  Script would crash out
if it couldn't find the server.  Also, sometimes Full Text indexes would
return NULL values for size information.  The previous code was reusing values that were not
correct.  I had to check for NULL then set to zero.  The introduction of
a Round function on a previous update also threw errors with NULL
values.


ctrlbold:
Works very well on my servers. Thanks, @scinobe!

PR #532, opened 2017-01-14
Fix restore sql backup from directory norecovery

### LINKED ISSUE #516: Restore-SqlBackupFromDirectory NoRecovery
Is this a *feature* OR *bug*:
Bug
- Powershell Version:  4.0
- OS Version:  Windows 2012 R2

**Steps to Reproduce**
 Restore-SqlBackupFromDirectory -SqlServer TestServer -Path '\\UNCPath\Backups\TestServer$InstanceName\' -Databases TestDatabase-NoRecovery 

When trying to restore a database with and leaving it in NoRecovery by specifying the -NoRecovery parameter, the database is recovered.  When leaving off that option, it throws an error, but still recovers the database.  Also noticed that databases in full recovery mode with Differential backups and Log backups, do not get those restores.  It looks like from the output that just the full restore is being done.

**Action Results**

**Expected Results**

**Additional comments**

ConstantineK:
Fixes #516  

 - Passes NoRecovery to all restores, allowing the final test to determine whether to bring the database online

How to test this code: 
- Standard SQL Server setup
- Ola hallengren backup scripts https://ola.hallengren.com/
- A folder filled with backups generated from ola's script

- [X]  Windows 7

### DIFF
--- a/functions/Restore-SqlBackupFromDirectory.ps1
+++ b/functions/Restore-SqlBackupFromDirectory.ps1
@@ -75,7 +75,7 @@ All user databases contained within \\fileserver\share\sqlbackups\SQLSERVERA wil
 
 #>	
 	#Requires -Version 3.0
-	[CmdletBinding()]
+	[CmdletBinding(SupportsShouldProcess=$true)]
 	Param (
 		[parameter(Mandatory = $true)]
 		[Alias("ServerInstance","SqlInstance")]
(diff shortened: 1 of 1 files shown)


ctrlbold:
hey ck, your fix worked and great logic. Ultimately, however, the root cause of this script breaking -- me modifying /internal/Restore-Database without taking this command into account (since it's undocumented and will disappear to become part of @Stuart-Moore's series of restore commands) -- introduced other bugs.

The simplest fix is to add back the old internal restore-database command in the BEGIN block. It's not a long term solution but that's no problem because this command will not remain long term.

I submitted a separate PR quickfix that addresses all of the issues.

ConstantineK:
Sounds good, sorry this didn't do it all :D

PR #533, opened 2017-01-14
Restore-SqlDatabaseFromDirectory - fix

ctrlbold:


ctrlbold:
Tested multiple times and really just put back code that disappeared.

PR #537, opened 2017-01-16
Initial commit of Get-DbaPageFileSetting

PowerDBAKlaas:
Fixes # 

 - 
 - 
 - 

How to test this code: 
- [ ] 
- [ ] 

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [x]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally

### DIFF
--- a/functions/Get-DbaPageFileSetting.ps1
+++ b/functions/Get-DbaPageFileSetting.ps1
@@ -46,91 +46,104 @@ Returns a custom object displaying ComputerName, AutoPageFile, FileName, Status,
 		[PSCredential] [System.Management.Automation.CredentialAttribute()]$Credential
 	)
 BEGIN {}
-PROCESS {
+PROCESS
+    {
         foreach ($Computer in $ComputerName)
         {
-            Write-Verbose "Connecting to $Computer"
 			$reply = Resolve-DbaNetworkName -ComputerName $Computer -erroraction silentlycontinue
-            if ( $reply.ComputerName )
+            if ( $reply.ComputerName ) # we can reach $computer
             {
                 $computer = $reply.ComputerName
-            }
-            Write-Verbose "Connecting to $Computer via CIM (WSMan)"
-            $CompSys = Get-CimInstance -ComputerName $Computer -Query "SELECT * FROM win32_computersystem" -ErrorVariable $MyErr
-            if ( $CompSys )
-            {
-                Write-Verbose "Successfully retrieved PageFile information on $Computer via WSMan"
-                if ($CompSys.automaticmanagedpagefile -eq $False)
+                Write-Verbose "Connecting to $Computer via CIM (WSMan)"
+                $CompSys = Get-CimInstance -ComputerName $Computer -Query "SELECT * FROM win32_computersystem" -ErrorVariable $MyErr
+                if ( $CompSys ) # we have computersystem class via WSMan
                 {
-                    $PF = Get-CimInstance -ComputerName $Computer -Query "SELECT * FROM win32_pagefile" # deprecated !
-                    $PFU = Get-CimInstance -ComputerName $Computer -Query "SELECT * FROM win32_pagefileUsage"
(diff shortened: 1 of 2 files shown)


functions/Get-DbaPageFileSetting.ps1
```diff
@@ -0,0 +1,136 @@
...
+                        if ($CompSys.automaticmanagedpagefile -eq $False)
+                        {
+                            $PF = Get-CimInstance -CimSession $CIMsession -Query "SELECT * FROM win32_pagefile" # deprecated !
+                            $PFU = Get-CimInstance -CimSession $CIMsession -Query "SELECT * FROM win32_pagefileUsage"
+                            $PFS = Get-CimInstance -CimSession $CIMsession -Query "SELECT * FROM win32_pagefileSetting"
+                        }
+                    }
+                    else
+                    {
+                    Write-Warning "$computer Operating System too old. No Pagefile information available."
```
ClaudioESSilva: Missing some indentation here and the line after.

ClaudioESSilva: but when I change the settings so this become $false I receive this message but I don't receive it when it goes through the IF block. It is supposed to be like this?
If query throught DCOM the behavior shouldn't be the same? Or thought DCOM we can't access the properties?

IF block:
image

else block (without change settings):
image

PowerDBAKlaas: I don't understand what setting you changed.

The process is:
- try WSMan
- if that fails, try to get win32_computersystem via DCom
- if we get such class and it has an .automaticmanagedpagefile property, we have what we want
- if the object is there, but it has no .automatic... property, it's an OS that has no page file (pre 2003R2).

functions/Get-DbaPageFileSetting.ps1
```diff
@@ -0,0 +1,136 @@
...
+                    $PFU = Get-CimInstance -ComputerName $Computer -Query "SELECT * FROM win32_pagefileUsage"
+                    $PFS = Get-CimInstance -ComputerName $Computer -Query "SELECT * FROM win32_pagefileSetting"
+                }
+            }
+			else
+            {
+				Write-Verbose "No WSMan connection to $Computer"
+                Write-Verbose "Getting computer information from server $Computer via CIM (DCOM)"
+			    $sessionoption = New-CimSessionOption -Protocol DCOM
+			    $CIMsession = New-CimSession -ComputerName $Computer -SessionOption $sessionoption -Credential $Credential
```
PowerDBAKlaas: added 
Get-CimSession | Remove-CimSession

ClaudioESSilva: I think this could be a little bit overwhelming. Imagine you have some CimSession in use for other module/command on the same session. That combo will remove all of them.
I would prefer to do that on the progress block after each one and remove only the one inside $CIMSession. What do you think?

functions/Get-DbaPageFileSetting.ps1
ClaudioESSilva: I don't have changed anything in the pagefile settings.
I want to test the ELSE block and I expected that it will work like the IF (1st screenshot).
It is not suppose? I tested on sql2008 machine.

functions/Get-DbaPageFileSetting.ps1
PowerDBAKlaas: change to Remove-Cimsession $Cimsession inside foreach loop.

We need to decide on a common strategy for this :-)

ctrlbold:
I believe additional clarification is needed. @ClaudioESSilva - please work with @PowerDBAKlaas on Slack and let me know if we can get this one in by today.

PowerDBAKlaas:
we're finally there @ctrlbold 
thank you very much for the assistance, @ClaudioESSilva

ctrlbold:
Beautiful!! 😍

PR #539, opened 2017-01-17
Test-DbaDatabaseFreespace - Sql200 is not supported, Skip over without breaking loop

### LINKED ISSUE #529: Get-DbaDatabaseFreespace doesn't handle sql 2000
Is this a *feature* OR *bug*:
bug

**Expected Results**
it's okay not to handle sql 2000, but you gotta say it dont spray it

SirCaptainMitch:
Fixes #529 
Tested on Windows 10, powershell 5, against sql2000, 2005, 2012, 2016 

 - adds an if for version 8 and skips over sql2000 without breaking the loop 

How to test this code: 
- [ ] Test with Windows Auth, against sql2000. Best if there are multiple servers 
- [ ] Test with SQL Login Auth, against sql2000. Best if there are multiple servers 

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally

### DIFF
--- a/functions/Get-DbaDatabaseFreespace.ps1
+++ b/functions/Get-DbaDatabaseFreespace.ps1
@@ -104,10 +104,19 @@ Returns database files and free space information for the db1 and db2 on localho
 		foreach ($s in $SqlServer)
 		{
 			#For each SQL Server in collection, connect and get SMO object
+
+			
+
 			Write-Verbose "Connecting to $s"
 			$server = Connect-SqlServer $s -SqlCredential $SqlCredential
 			#If IncludeSystemDBs is true, include systemdbs
 			#only look at online databases (Status equal normal)
+
+			if ($server.VersionMajor -eq '8'){
+				Write-Warning "SQL Server 2000 is not supported. Skipping: $($server.Name)"
+				continue
+			}
+
 			try
 			{
 				if ($databases.length -gt 0)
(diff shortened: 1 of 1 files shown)


functions/Get-DbaDatabaseFreespace.ps1
```diff
@@ -108,6 +108,12 @@ Returns database files and free space information for the db1 and db2 on localho
 			$server = Connect-SqlServer $s -SqlCredential $SqlCredential
 			#If IncludeSystemDBs is true, include systemdbs
 			#only look at online databases (Status equal normal)
+
+			if ($server.VersionMajor -eq '8'){
```
ctrlbold: While this works, it should be 8 instead of '8' because that is a string of 8. Will fix after merge.

functions/Get-DbaDatabaseFreespace.ps1
```diff
@@ -108,6 +108,12 @@ Returns database files and free space information for the db1 and db2 on localho
...
+				Write-Warning "SQL Server 2000 is not supported. Skipping: $($server.Name)"
+				continue
```
ctrlbold: We can now just do $server since that puts brackets which is what we like, will also update after. 

Thanks for the fix!

PR #540, opened 2017-01-18
Get-DbaHelpIndex - new function/script

sirsql:
Fixes # 

 - New function Get-DbaHelpIndex which returns index information for a given database and or table
 - Should work 2005 and up (do not have revs below 2012 for testing) - throws error if SQL 2000
 - A few switches for additional functionality - lots of info on that through get-help

How to test this code: 
- [ ] 
- [ ] 

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally

### DIFF
--- a/functions/Get-DbaHelpIndex.ps1
+++ b/functions/Get-DbaHelpIndex.ps1
@@ -1,0 +1,596 @@
+﻿function Get-DbaHelpIndex
+{
+<#
+.SYNOPSIS
+Returns size, row and configuration information for indexes in databases.
+
+.DESCRIPTION
+This function will return detailed information on indexes (and optionally statistics) for all indexes in a database, or a given index should one be passed along.
+As this uses SQL Server DMVs to access the data it will only work in 2005 and up (sorry folks still running SQL Server 2000).
+The data includes:
+	- ObjectName: the table containing the index
+	- IndexType: clustered/non-clustered/columnstore and whether the index is unique/primary key
+	- KeyColumns: the key columns of the index
+	- IncludeColumns: any include columns in the index
+	- FilterDefinition: any filter that may have been used in the index 
+	- DataCompression: row/page/none depending upon whether or not compression has been used
+	- IndexReads: the number of reads of the index since last restart or index rebuild
+	- IndexUpdates: the number of writes to the index since last restart or index rebuild
+	- SizeKB: the size the index in KB
+	- IndexRows: the number of the rows in the index (note filtered indexes will have fewer rows than exist in the table)
+	- IndexLookups: the number of lookups that have been performed (only applicable for the heap or clustered index)
+	- MostRecentlyUsed: when the index was most recently queried (default to 1900 for when never read)
+	- StatsSampleRows: the number of rows queried when the statistics were built/rebuilt
+	- StatsRowMods: the number of changes to the statistics since the last rebuild
+	- HistogramSteps: the number of steps in the statistics histogram
+	- StatsLastUpdated: when the statistics were last rebuilt
+
+.PARAMETER SqlServer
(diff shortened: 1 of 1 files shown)


functions/Get-DbaHelpIndex.ps1
```diff
@@ -0,0 +1,596 @@
...
+Get-DbaHelpIndex -SqlServer localhost -Database MyDB -ObjectName dbo.Table1 -FormatResults
+	Returns the index information for the table dbo.Table1 in the MyDB database, and returns the numerical data with separators to make it more readable (ie 1234 becomes 1,234)
+	
+.EXAMPLE
+Get-DbaHelpIndex -SqlServer localhost -Database MyDB -IncludeStats -FormatResults
+	Returns the index information for all index in the MyDB database, as well as statistics, and formats the numerical data to be more redable	
+	
+#>
+	
+	[CmdletBinding(SupportsShouldProcess = $true)]
```
PowerDBAKlaas: I don't think it does support shouldprocess.
-Whatif or -Confirm have no effect, I just get output straight away.

wsmelton: For `Get` commands we can remove this, as it is not needed.

functions/Get-DbaHelpIndex.ps1
```diff
@@ -0,0 +1,596 @@
...
+
+DECLARE @IndexUsageStats TABLE
+    (
+      object_id INT ,
+      index_id INT ,
+      user_scans BIGINT ,
+      user_seeks BIGINT ,
+      user_updates BIGINT ,
+      user_lookups BIGINT ,
+      last_user_lookup DATETIME2(0) ,
```
sirsql: I don't have 2005 to test. Can change it to datetime.

functions/Get-DbaHelpIndex.ps1
sirsql: Will remove

functions/Get-DbaHelpIndex.ps1
```diff
@@ -0,0 +1,596 @@
...
+	- HistogramSteps: the number of steps in the statistics histogram
+	- StatsLastUpdated: when the statistics were last rebuilt
+
+.PARAMETER SqlServer
+SQLServer name or SMO object representing the SQL Server to connect to.
+
+.PARAMETER SqlCredential
+PSCredential object to connect as. If not specified, current Windows login will be used.
+
+.PARAMETER Databases
```
sirsql: It can accept multiple databases, and uses Get-ParamSqlDatabases which provides the databases parameter. Using database would require another helper function, or remove the potential for the drop down list.

functions/Get-DbaHelpIndex.ps1
ClaudioESSilva: The datetime2 is not the only issue. Also some columns on the DMV only appear later. For example the filter (not mentioned here, should be nice if you can update with it :-) ), and compression.
We have other command (https://github.com/sqlcollaborative/dbatools/blob/master/functions/Find-SqlDuplicateIndex.ps1) that use those and I had to use 2 different sql queries in order to support both.

functions/Get-DbaHelpIndex.ps1
PowerDBAKlaas: oh yes of course, this is in the dynamicparams
sorry

sirsql:
I got a hold of a copy of 2005. Going to close this PR and get to work on ensuring compat with it. Will put out a new one when I've been able to test that.

PR #586, opened 2017-01-19
Initial commit of Get-DbaAgentOperator

PowerDBAKlaas:
 - New function to list the SQL Agent Operators
 - 
 - 

How to test this code: 
- [ ] 
- [ ] 

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [x]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [x] Works remotely
- [x] Works locally

### DIFF
--- a/functions/Get-DbaAgentOperator.ps1
+++ b/functions/Get-DbaAgentOperator.ps1
@@ -1,0 +1,87 @@
+﻿FUNCTION Get-DbaAgentOperator
+{
+<#
+.SYNOPSIS
+Returns all SQL Agent operators on a SQL Server Agent.
+
+.DESCRIPTION
+This function returns SQL Agent operators.
+
+.PARAMETER SqlServer
+SQLServer name or SMO object representing the SQL Server to connect to.
+This can be a collection and receive pipeline input.
+
+.PARAMETER SqlCredential
+PSCredential object to connect as. If not specified, currend Windows login will be used.
+
+.NOTES 
+Author: Klaas Vandenberghe ( @PowerDBAKlaas )
+Date: 2017-01-16
+
+dbatools PowerShell module (https://dbatools.io, clemaire@gmail.com)
+Copyright (C) 2016 Chrissy LeMaire
+This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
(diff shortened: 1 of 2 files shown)


functions/Get-DbaAgentOperator.ps1
```diff
@@ -0,0 +1,87 @@
...
+			}
+			catch
+			{
+				Write-Warning "Failed to connect to $Instance"
+				continue
+			}
+			
+			$operators = $Instance.jobserver.operators
+			
+			if (!$operators)
```
ClaudioESSilva: An instance without operators is falling on else block. Should fall on IF

functions/Get-DbaAgentOperator.ps1
```diff
@@ -0,0 +1,87 @@
...
+			{
+				$Instance = Connect-SqlServer -SqlServer $Instance -SqlCredential $sqlcredential
+			}
+			catch
+			{
+				Write-Warning "Failed to connect to $Instance"
+				continue
+			}
+			
+			$operators = $Instance.jobserver.operators
```
ClaudioESSilva: We should validate if SQL Server Agent is running?
I don't had the service up and this run without any kind of warning. I would appreciate and message like "The service is down. Please start SQL Server Agent service"

ClaudioESSilva: If instance is an Express edition we should say that SQL Server Agent is not available on this edition.

PowerDBAKlaas: For now I added a check if we get a SQL agent, but not if it's sqlexpress or a stopped service or ...
I'm not sure we should include so much tests in here.
If test-connection fails, we also won't check AD or DNS cache to say "the computer exists. Please turn power on"
Maybe we can write a separate function to bundle those checks?

functions/Get-DbaAgentOperator.ps1
PowerDBAKlaas: switched to $operator.count check

functions/Get-DbaAgentOperator.ps1
```diff
@@ -0,0 +1,98 @@
...
+			}
+			catch
+			{
+				Write-Warning "Failed to connect to $Instance"
+				continue
+			}
+            Write-Verbose "Connecting to SQL Agent on $Instance"
+			try
+			{
+				$Jobserver = $Instance.jobserver
```
ClaudioESSilva: This never fail. So the flow never gonna hit the catch block. This case we have to check if $jobserver is empty or null.
But, I tested to stop the service and even when SQL Agent is stopped smo can reach operators. So set back how it was seems the best.
Sorry @PowerDBAKlaas

ctrlbold:
Parrrrrrrty!

PR #593, opened 2017-01-22
Initial commit for Get-DbaClusterSqlInstance

Stuart-Moore:
Gets installed instances back via WMI over CIM

Also created function page on the website.

Fixes # 

 - 
 - 
 - 

How to test this code: 
- Get-DbaClusterSqlInstance -ClusterName ClusterOfChoice.contoso.com
Should return a list of installed SQL Server instances

'ClusterOfChoice.contoso.com' | Get-DbaClusterSqlInstance
also works

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [x] Works locally


functions/Get-DbaClusterSqlInstance.ps1
```diff
@@ -0,0 +1,84 @@
...
+						$CimOption = New-CimSessionOption -Protocol DCOM
+						$CimSession = New-CimSession -Credential:$ClusterCredential -ComputerName $Cluster -SessionOption $CimOption
+						$Results = $CimSession | Get-CimInstance -class "MSCluster_Resource" -namespace "root\mscluster"| where {$_.type -eq "SQL Server"}  					
+					}
+					catch
+					{
+						Write-Exception $_
+					}
+				}
+				$results | %{ $_.PrivateProperties.VirtualServerName+"\"+$_.PrivateProperties.InstanceName}
```
PowerDBAKlaas: Couldn't it be useful to return the status too?
Would this also include SSAS?

functions/Get-DbaClusterSqlInstance.ps1
```diff
@@ -0,0 +1,84 @@
...
+{
+<#
+.SYNOPSIS
+Returns the registered SQL Server Instances on a windows Server cluster
+	
+.DESCRIPTION
+By default, this command returns for each Cluster passed in:
+Pulls Cluster details via WMI/CIM so doesn't require Failover Cluster PowerShell module to be installed.
+Multiple clusters can be piped in.
+Returns ALL instance registered on cluster. Does not check that they are viable, correctly installed or currenly running.
```
PowerDBAKlaas: typo: "currenly"

functions/Get-DbaClusterSqlInstance.ps1
```diff
@@ -0,0 +1,84 @@
...
+.SYNOPSIS
+Returns the registered SQL Server Instances on a windows Server cluster
+	
+.DESCRIPTION
+By default, this command returns for each Cluster passed in:
+Pulls Cluster details via WMI/CIM so doesn't require Failover Cluster PowerShell module to be installed.
+Multiple clusters can be piped in.
+Returns ALL instance registered on cluster. Does not check that they are viable, correctly installed or currenly running. 
+	
+.PARAMETER Cluster
```
PowerDBAKlaas: This confuses me:
if you submit a clustername, you can end up on multiple computers, right?
Is "the Sql Server that..." the correct description, then?

functions/Get-DbaClusterSqlInstance.ps1
```diff
@@ -0,0 +1,84 @@
+function Get-DbaClusterSqlInstance
```
Stuart-Moore: Nope.

functions/Get-DbaClusterSqlInstance.ps1
Stuart-Moore: Oops

functions/Get-DbaClusterSqlInstance.ps1
Stuart-Moore: No, you'll end up on the node hosting the cluster service (it's a seperate role within the WSFC and it's hosted on a node the same as any other role).

Each node could return this information, but best practice is to ask the one in charge

functions/Get-DbaClusterSqlInstance.ps1
```diff
@@ -0,0 +1,84 @@
...
+.DESCRIPTION
+By default, this command returns for each Cluster passed in:
+Pulls Cluster details via WMI/CIM so doesn't require Failover Cluster PowerShell module to be installed.
+Multiple clusters can be piped in.
+Returns ALL instance registered on cluster. Does not check that they are viable, correctly installed or currenly running. 
+	
+.PARAMETER Cluster
+The SQL Server that you're connecting to.
+
+.PARAMETER ClusterCredential
```
Stuart-Moore: Certainly can. Was just thinking about seperating it out

functions/Get-DbaClusterSqlInstance.ps1
Stuart-Moore: I'm leaving the status to other command ( function that does one thing well, etc). 

So status would be checked by pushing them through to one of the existing functions. this was why I wrote the basics, wanted to dump the output through to Get-DbaUptime to see if we'd had a mass failover/flap during a core router outage.

functions/Get-DbaClusterSqlInstance.ps1
PowerDBAKlaas: Oh, sorry, I overwrote that comment, but it got lost apparently
I wanted this to say:
We could have more functions later that collect the nodes (I wrongly assumed first that's what this function did), or work with SQL Server FCI.
So I think it would be better to make clear this is about WSFC cluster and resources, not nodes.
Get-DbaWsfcSqlResource?

functions/Get-DbaClusterSqlInstance.ps1
Stuart-Moore: SSAS should just be another type of Cluster resource.
I don't have a clustered instance to hand for that, 

And do the other tools accept a SSAS instance happily? or would that need filtering before piping into something else?

functions/Get-DbaClusterSqlInstance.ps1
PowerDBAKlaas: Yes, that's what I meant. Not multiple, but sometimes computer1, and sometimes computer2. So still, should we describe that as "the sql server", while the parameter's name is "cluster"

Stuart-Moore: You can do, I was trying to be accurate.

Tbh, if the cluster is setup properly then just using a node name will work (does with my testing, and my opinion is that mine are setup correctly ;) )

functions/Get-DbaClusterSqlInstance.ps1
PowerDBAKlaas: To make it all more consistent, we chose to use 'Credential' for anything that isn't 'SqlCredential'.
Of course, there may be exceptions.
Like adding 'source' and 'destination' if we need two for the copy- functions, for example.
I guess in this case, there can't be much confusion?

functions/Get-DbaClusterSqlInstance.ps1
PowerDBAKlaas: It depends, but in this case, I seem to see status is just a property of the class you are querying, so it's free.
https://msdn.microsoft.com/en-us/library/aa371464(v=vs.85).aspx

And including ssas could be (I don't have any either) just a redefinition of the where clause, so, almost free.
Indeed, a filter before piping would be an easy way to pipe only SqlServer or only SSAS to the next function.

I think I prefer that over having separate functions. But then again, I don't work with clusters, so you may have more valid arguments for the other way.

functions/Get-DbaClusterSqlInstance.ps1
Stuart-Moore: Works for me

Yes, having a look at Get-DbaClusterActiveNode was on my list of things to do. And that looks prime for splitting out to make scaling up the number of nodes easier to do.

functions/Get-DbaClusterSqlInstance.ps1
Stuart-Moore: I'll ask on slack if anyone's got a clustered SSAS install and wouldn't mind sending me the WMI return

Added Status and StatusCode to the output. only issue with that, is that's it's only the SQL Server status as seen by the cluster. Doesn't mean that everything is accessible, as that relies on other things, which hopefully are correct with the right dependencies. Looping back through another command is still going to be the definitive proof.

PowerDBAKlaas: Yes, combining with other functions will clearly be the cherry on the cake.
We're hoping you'll write the whole series. :-)
Game of Clusters, season 1,2,3 & 4, by Stuart Moore

Does this [status] need a more descriptive name then to distinct it from the others?

functions/Get-DbaWsfcSqlResource.ps1
```diff
@@ -0,0 +1,99 @@
...
+
+.LINK
+https://dbatools.io/Get-DbaWsfcSqlResource
+
+.EXAMPLE
+Get-DbaWsfcSqlResource -ClusterName MyProdCluster.contoso.com
+
+Returns a list of all SQL Server instances that have been setup on the cluster MyProdCluster.contoso.com
+.EXAMPLE
+$Credential = Get-Credential
```
PowerDBAKlaas: Since the function now returns an object, this example probably needs an -expandproperty ?

Stuart-Moore: See the PR I'm just about to push. That example will now work straight through.

Or will if both PRs go through.

ctrlbold:
NooooOOoOOoooo! I had a whole big thing I typed out and lost it. Anyway, I really love the idea of this command! I also appreciate that it does not rely on the Cluster module. 

I'd like to see it offer a lot more - imagine how $smoserver offers properties and collections of things. We could have the same things, that basically show this:

image

I have some code that can help with the storage (though it uses WMI and the cluster module, so it'd need to be converted)

https://blog.netnerds.net/2013/07/powershell-v3-get-storage-information-for-clustered-sql-servers/

As an aside, I'd like for all cluster command have cluster in the name. Wscf is not an acronym I ever use (even when searching Google) and I primarily use these classic Clusters.

Stuart-Moore:
It's wsfc, you've transposed the last 2 letters. 2nd result in google for
me is the  MSDN article on windows server failover clustering.

But I'll let you and klaas argue that one as it's already been renamed once
;)

In my normal pattern I see this as the first in a string of commands. So
the output from this could be pushed into Get-DbaWSClusterInstanceDetails
to then return the FC manager role details.

Either push in the entire cluster results from this one, or where-object it
down to just the role you want
I may award about trying to replicate too much of the functionality of the
existing tools, as those are the proper supported way of managing clusters.
I'd see these functions as making some DBA tasks easier. For instance there
isn't an easy way to do what this command does and get a nice simple list
of sql instances back

PowerDBAKlaas:
@Stuart-Moore don't encourage us to argue, please. 😄 

I thought "Wsfc" would be nice. I believe @wsmelton used it too a few days ago. But I've hardly worked with clusters, so don't let my vote weigh in. If experienced users go "what is wsfc?" than maybe a more explanatory name is preferrable.

I do prefer to have one function to collect all useful properties from an object we visit, since a simple | select , with or without -expandproperty can turn the objects into a list. It'll be more versatile that way. 

From there on, your idea about piping the output to the next function is of course perfect. There may be enough possibilities specific for clusters, and also anything that collects security, performance, and more, or even sets properties.

Stuart-Moore:
Any thoughts on the name?
And now I've been introduced to Select-DefaultField I'm happy to include more, but hide them in the output for now. Any particular ones people are thinking of? I'd go with IDs and GUID type ones as those will be wanted by other functions down the line

ctrlbold:
(I changed it to Select-DefaultView since that's more what it is) (don't have more input just yet but I'd love to see a ton 👍 )

ctrlbold:
I'm going to close this PR without merging and will **not** delete its branch. I want this command, but it will need to be redesigned and beefed up.

PR #604, opened 2017-01-28
Get/Set-DbaProcess - now awesome

ctrlbold:


ctrlbold:
Here's proofs that they work
image

I've been wanting to do this for a while and some small changes to @niphlod's new commands require proper output 👍

PR #629, opened 2017-02-02
Find-DbaStoredProcedure - Speed up find sproc

ctrlbold:
We need some description :)


ctrlbold:
Thanks bud! Love your improvements <3

PR #631, opened 2017-02-03
Measure-DbaBackupThroughput

ctrlbold:
I don't doubt my code is messed up somewhere -- please anyone take a look. I am bad at math.

- [ ] Measure-DbaBackupThroughput -SqlInstance sql2012  | Format-Table
- [ ] Measure-DbaBackupThroughput -SqlInstance sql2005 -Databases SharePoint_Config | Format-Table
- [ ] Measure-DbaBackupThroughput -SqlInstance sql2016 -Since (Get-Date).AddDays(-30) | Format-Table
- [ ] Measure-DbaBackupThroughput -SqlInstance sql2005 -Databases SharePoint_Config -Since (Get-Date).AddDays(-30)

### DIFF
--- a/functions/Measure-DbaBackupThroughput.ps1
+++ b/functions/Measure-DbaBackupThroughput.ps1
@@ -142,6 +142,7 @@ Will fill this in
 				$measurestart = $db.Group.Start | Measure-Object -Minimum
 				$measureend = $db.Group.End | Measure-Object -Maximum
 				$measuresize = $db.Group.TotalSizeMB | Measure-Object -Average
+				$avgduration = $db.Group | ForEach-Object { New-TimeSpan -Start $_.Start -End $_.End } | Measure-Object -Average TotalSeconds
 				
 				[pscustomobject]@{
 					Server = $db.Group.Server | Select-Object -First 1
@@ -153,6 +154,7 @@ Will fill this in
 					MinBackupDate = $measurestart.Minimum
 					MaxBackupDate = $measureend.Maximum
 					BackupCount = $db.Count
+					AvgDuration = New-TimeSpan -Start (Get-Date) -End (Get-Date).AddSeconds($avgduration.Average)
 				}
 			}
 		}
--- a/functions/Measure-DbaBackupThroughput.ps1
+++ b/functions/Measure-DbaBackupThroughput.ps1
@@ -134,7 +134,7 @@ Will fill this in
 				$agghistories += $history | Select-Object Server, Database, MBps, TotalSizeMB, Start, End
 			}
 			
-			$groups = $agghistories | Group-Object Database
+			$groups = $agghistories | Sort-Object Database | Group-Object Database
 			
 			foreach ($db in $groups)
 			{
--- a/functions/Measure-DbaBackupThroughput.ps1
+++ b/functions/Measure-DbaBackupThroughput.ps1
@@ -1,0 +1,160 @@
+﻿function Measure-DbaBackupThroughput
+{
+<#
+.SYNOPSIS
+Determines how quickly SQL Server is backing up databases to media.
+
+.DESCRIPTION
+Returns backup history details for some or all databases on a SQL Server. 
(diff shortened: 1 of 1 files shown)


ctrlbold:
image

yehh

ctrlbold:
image

added avgduration. I'm not so sure I did this part right.

ctrlbold:
Doing total rewrite, don't judge except for AvgDuration. That one isn't working 👎

ctrlbold:
Okay almost done. Can someone help me with the math? There's no way my backup wrote 2404 MB a second. It's shared storage. Perhaps this has something to do with compression?

```
Database                          AvgThroughputMB                AvgSizeMB AvgDuration                       MinThroughputMB          MaxThroughputMB MinBackupDate            MaxBackupDate
WSS_Content                 2404.65                 197278.4 01:15:48.5000000                           142.49                  4345.43 1/8/2017 2:35:45 AM      1/29/2017 3:52:14 AM```

**Comment by niphlod** [comment]
lemme try

**Comment by niphlod** [comment]
my timings and averages are aligned correctly...and I don't see a flaw in the math.

**Comment by ctrlbold** [comment]
Thanks for looking @niphlod - do you have compression enabled? I mean, maybe this is why compression is so awesome is that it's effectively writing at that speed but not really. So this tests your backup throughput but not your storage throughput :brainfry:

**Comment by niphlod** [comment]
math is right doesn't mean command is right. I'm checkin the underlying data.

**Comment by niphlod** [comment]
yep. the underlying data reports the uncompressed size. Get-DbaBackupHistory should need to be modified to report the size of the compressed backup if you want disk throughput.

**Comment by ClaudioESSilva** [comment]
Get-DbaBackupHistory is unaware of compression.

https://github.com/sqlcollaborative/dbatools/blob/master/functions/Get-DbaBackupHistory.ps1#L230

**Comment by niphlod** [comment]
yes but why ? meaning: it's for a specific reason or what ? all that it takes is that for < 2008 the size for TotalSizeMB is get from Backup_Size and from there onwards from compressed_backup_size ...

**Comment by niphlod** [comment]
also, be aware that there's an hiccup in get-dbadatabasehistory .... try running it in a case-sensitive instance :P . Should I log a bug ?

**Comment by ClaudioESSilva** [comment]
@niphlod Yes for sure!

Regarding the compression is different columns:

An example from a script that I use a lot.
```
CASE 
				WHEN bup.compressed_backup_size < bup.backup_size THEN 'Compressed'
				ELSE 'Not Compressed'
			  END AS isBackupCompressed
```

niphlod:
@ClaudioESSilva : 

> yes but why ? meaning: it's for a specific reason or what ? all that it takes is that for < 2008 the size for TotalSizeMB is get from Backup_Size (column) and from there onwards from compressed_backup_size (column)...

ClaudioESSilva:
Sorry misread it 😳 

**I think** he just leaved the sql 2008+ support for later and never finished.

niphlod:
yeah there's also a nasty RANK() that could be easily avoided if we want this to support even 2000...

ctrlbold:
I'd have zero idea how to implement that within a timely manner. Any help? It was necessary for something.

niphlod:
@ctrlbold : I'll try to fix get-dbabackupheader in an hour or two if nothing urgent comes up.

ctrlbold:
@niphlod hell yeah, thanks man!

ctrlbold:
Was this fixed, @niphlod ?

niphlod:
the bug totally. the fact that we could work out it even for mssql 2000 no. we can log an enhancement for it if there is traction for it but I'd refrain from doing it for each and every command. we should have a list of supported targets for each command and see if enough people want it.

ctrlbold:
naw, i dont wanna put in too much effort for 2000. It's always best try within a reasonable amount of time, if not, nbd. thank you for fix! I couldn't remember and my github profile reminded me.

PR #633, opened 2017-02-03
Find dbaloginingroup fix

ctrlbold:


ctrlbold:
hey @niphlod and @Staggerlee011 -- I removed the pipe holdup and cleaned up a couple things but I don't think the -Login param is working as expected. Please fix in Find-DbaLoginInGroup-Fix branch.

image

ctrlbold:
Also, login should be the whole login, no? like BASE\ctrlb instead of ctrlb

ctrlbold:
dang, this command is making other tests fail until it's in dev. I'll merge then file bug report.

niphlod:
the only culrpit I see is that the loginname is not reported with the domain. everything else works as expected

niphlod:
sent a fix over .

niphlod:
this should be fixed already. @ctrlbold ping me if not

PR #668, opened 2017-02-07
Get-DbaMsDtc - improve error handling

PowerDBAKlaas:
add localhost as default computername
remove check for computername count
split registry checks
remove erroraction stop
remove semicolons

Fixes # 

 - 
 - 
 - 

How to test this code: 
- [ ] 
- [ ] 

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally


ctrlbold:
hey @PowerDBAKlaas - im having a number of issues with my local machine - even after I Ran As

image

functions/Get-DbaMsdtc.ps1
```diff
@@ -90,77 +94,73 @@ Get DTC status for the computer srv0042 and show in a grid view
...
-					networkDTCAccessTransactions = $reg.networkDTCAccessTransactions
-					XATransactions = $reg.XATransactions
-				}
-			}
-		}
+    ForEach ($Computer in $Computername)
+    {
+      $dtcservice = $null
+      Write-Verbose "$functionName - Getting DTC on $computer via WinRM"
+      $dtcservice = Get-Ciminstance -ComputerName $Computer -Query $query
```
FriedrichWeinmann: This can easily fail with blood red errors.
You may want to consider handling those, maybe try/catch and write a nice warning.

functions/Get-DbaMsdtc.ps1
```diff
@@ -90,77 +94,73 @@ Get DTC status for the computer srv0042 and show in a grid view
...
+      if ( $null -eq $dtcservice )
+      {
+          Write-Verbose "$functionName - Failed To get DTC via WinRM. Getting DTC on $computer via DCom"
+          $SessionParams = @{ }
+          $SessionParams.ComputerName = $Computer
+          $SessionParams.SessionOption = (New-CimSessionOption -Protocol Dcom)
+          $Session = New-CimSession @SessionParams
```
FriedrichWeinmann: This can easily fail with blood red errors.
You may want to consider handling those and terminate the execution for this computer - it will not produce output anymore anyway.
You may want to consider the new "Stop-Function" function as way to handle this.

functions/Get-DbaMsdtc.ps1
```diff
@@ -71,17 +71,21 @@ Get-DbaMsdtc -ComputerName srv0042 | Out-Gridview
...
+		[Parameter(ValueFromPipeline = $true)]
 		[Alias("cn", "host", "Server")]
-		[string[]]$Computername
+		[string[]]$Computername = $env:COMPUTERNAME
 	)
 	
-	BEGIN
+BEGIN
 	{
+    $functionName = "Get-DbaMsDtc"
```
FriedrichWeinmann: FYI: You can get your own function's name with this line instead:
(Get-PSCallstack)[0].Command
One thing less you have to maintain later on. Will survive all renames.

functions/Get-DbaMsdtc.ps1
```diff
@@ -90,77 +94,73 @@ Get DTC status for the computer srv0042 and show in a grid view
...
+      $reg = $CIDs = $null
+      $CIDHash = @{ }
+      Write-Verbose "$functionName - Getting MSDTC Security Registry Values on $Computer"
+      $reg = Invoke-Command -ComputerName $Computer -ScriptBlock $DTCSecurity
+      if ( $null -eq $reg )
+      {
+        Write-Warning "$functionName - Can't connect to MSDTC Security registry on $Computer"
+      }
+      Write-Verbose "$functionName - Getting MSDTC CID Registry Values on $Computer"
+      $CIDs = Invoke-Command -ComputerName $Computer -ScriptBlock $DTCCIDs
```
FriedrichWeinmann: This command shouldn't be executed if the previous Invoke-Command failed: Adds delay and no resaon for it to work when it didn't the first time.
Also any errors should be caught and handled.

functions/Get-DbaMsdtc.ps1
```diff
@@ -90,77 +94,73 @@ Get DTC status for the computer srv0042 and show in a grid view
...
+          $dtcservice = Get-Ciminstance -CimSession $Session -Query $query
+      }
+      if ( $null -eq $dtcservice )
+      {
+          Write-Warning "$functionName - Can't connect to CIM on $Computer"
+      }
+      $reg = $CIDs = $null
+      $CIDHash = @{ }
+      Write-Verbose "$functionName - Getting MSDTC Security Registry Values on $Computer"
+      $reg = Invoke-Command -ComputerName $Computer -ScriptBlock $DTCSecurity
```
FriedrichWeinmann: This can error out.
A little try/catch should help make this a less bloody user experience

functions/Get-DbaMsdtc.ps1
```diff
@@ -71,17 +71,21 @@ Get-DbaMsdtc -ComputerName srv0042 | Out-Gridview
...
 		[Alias("cn", "host", "Server")]
-		[string[]]$Computername
+		[string[]]$Computername = $env:COMPUTERNAME
 	)
 	
-	BEGIN
+BEGIN
 	{
+    $functionName = "Get-DbaMsDtc"
+    $ComputerName = $ComputerName | ForEach-Object {$_.split("\")[0]} | Select-Object -Unique
```
FriedrichWeinmann: Would you mind explaining to me what is the purpose behind this line? I can't make heads or tails of it :(

functions/Get-DbaMsdtc.ps1
ctrlbold: Thanks for this FriedrichWeinmann! @Stuart-Moore, @PowerDBAKlaas and I were discussing it on the Style board. Will add it there.

ctrlbold:
@PowerDBAKlaas - please ping me when the necessary changes have been made.

functions/Get-DbaMsdtc.ps1
wsmelton: This line ensures that if the same computer name is passed into the command multiple times it will only process it once.

PowerDBAKlaas: yes @FriedrichWeinmann  we had some discussion on this and decided to go for this as a general approach, putting user friendliness first. So both pulling computernames out in case a user feeds us instance names, and making computername unique before looping through.

Thanks, @wsmelton

functions/Get-DbaMsdtc.ps1
PowerDBAKlaas: well, I'm not sure about this one.
I don't have enough different configurations to test on.
So I'm not sure if on older OS, or some versions, one registry key can exist and the other not.
If anyone can verify that or point to documentation that explains the history or existence of those keys, I'd be grateful.

functions/Get-DbaMsdtc.ps1
PowerDBAKlaas: yes, error handling needs to be addressed in general.
I used to put -ErrorAction Silentlycontinue everywhere, but then I realised it would be better not to, and left it out without putting in a better way.

When using try catch around a variable assignment, we have lots of cases the catch is useless, since the assignment succeeds, even when assigning $null. So we replaced that with Silently continue and then an if else, checking if there is something meaningful in the variable.

functions/Get-DbaMsdtc.ps1
FriedrichWeinmann: Well, in that case you may want to change it up a little:
Simply put, it isn't going to work when used with a pipeline (Since it runs begin only once on the first object piped in).

Instead:
In Begin, create a list of computernames already processed
During Process, within the foreach loop, add the computer to that list if it isn't already tehre or skip it.
This way it works well in both modes

functions/Get-DbaMsdtc.ps1
FriedrichWeinmann: Well ... since it is supposed to go on either way, how about tying it in with the Stop-Function methodology?
Use try/catch, write a warning when it is _not_ set to silent or write an error when it is.

Thoughts anybody:
 - Manually implement this each time?
 - Integrate it into Stop-Function, even though the code continues on the next line (currently it either terminates the function or calls continue)
 - Build a dedicated internal function for this?
The first option will need no overhead, but has lots of ()small pieces of) duplication.
The second option already has the infrastructure for that, but is not inuitively named
The third option would be the cleanest, but take more implementation (I've got another idea brewing, where it could be rolled into the general verbosity/debug output management, in which case this would probably vastly outshine the second option)

functions/Get-DbaMsdtc.ps1
FriedrichWeinmann: Actually, I was thinking about catching whenever the previous command was _unable to connect at all_. In that case it is safe to assume the second remoting attempt will fail too.
In that scenario you'd need to make sure the first script will not be throwing errors, in order to avoiding muddling the water (a $null return should work).
Then you call the first Invoke-Command with try/catch and on catch disable the second attempt.

ctrlbold:
worked across the board for all of my servers!

PR #671, opened 2017-02-07
New Function Get-DbaDatabase

gbargsley:


functions/Get-DbaDatabase.ps1
```diff
@@ -0,0 +1,162 @@
...
+.PARAMETER System
+Returns only system databases
+
+.PARAMETER User
+Returns only user databases
+	
+.PARAMETER State
+Returns only online databases
+
+.PARAMTER Owner
```
FriedrichWeinmann: Missing e breaks parameter help. Should this parameter really be named owner? I'd kind of expect to be allowed to ask for a specific owner, rather than "everything not owned by sa". Maybe something like "HasCustomOwner".

functions/Get-DbaDatabase.ps1
```diff
@@ -0,0 +1,162 @@
...
+		[parameter(Position = 0, Mandatory = $true, ValueFromPipeline = $True)]
+		[Alias("ServerInstance", "SqlServer")]
+		[object[]]$SqlInstance,
+		[System.Management.Automation.PSCredential]$SqlCredential,
+		[switch]$System,
+		[switch]$User,
+        [switch]$Owner,
+        [switch]$Encrypted,
+		[ValidateSet('Online', 'Offline', 'Emergency')]
+		[string]$State,
```
FriedrichWeinmann: Emergency option has no effect. Should this be allowed as an option then?

functions/Get-DbaDatabase.ps1
ctrlbold: Thanks for jumping right in @FriedrichWeinmann! 💯 

I'm thinking if we do use Owner, it should probably be like "give me all the databases with this owner".

functions/Get-DbaDatabase.ps1
```diff
@@ -0,0 +1,174 @@
...
+			{
+				Write-Warning "Failed to connect to: $instance"
+				continue
+			}
+			
+			$defaults = 'Name', 'Status', 'ContainmentType', 'RecoveryModel', 'CompatibilityLevel', 'Collation', 'Owner', 'EncryptionEnabled'
+			
+			if ($System)
+			{
+				$inputobject = $server.Databases | Where-Object { $_.IsSystemObject -eq $true }
```
wsmelton: Can make this a bit cleaner if you write these as `$server.Databases | Where-Object IsSystemObject`.

Since [the property](https://msdn.microsoft.com/en-us/library/microsoft.sqlserver.management.smo.database.issystemobject.aspx) is a bool value PowerShell does not require you to specify `$true` or `$false`.

functions/Get-DbaDatabase.ps1
```diff
@@ -0,0 +1,174 @@
...
+				$inputobject = $server.Databases | Where-Object { $_.Name -in $databases }
+			}
+			
+			if ($state -eq "Online")
+			{
+                $inputobject = $server.Databases | Where-Object { $_.status -eq 'Normal' }
+			}
+			
+			if ($state -eq "Offline")
+			{
```
wsmelton: If you do an "IF" more than once for the same property it will be less lines of code to just use the switch statement:

```powershell
switch ($state) {
   "Online" {$inputobject =....}
   "Offline" {$inputobject = ....}
}

File: functions/Get-DbaDatabase.ps1 (hunk 3 above)
Comment (ctrlbold) [inline functions/Get-DbaDatabase.ps1]: Nooo kidding! that will be prettier! Checking to ensure it works in v3, but it seems like it would. We need to add this to the style guide

Comment (ctrlbold) [inline functions/Get-DbaDatabase.ps1]: Excellent, it does!

Comment (FriedrichWeinmann) [inline functions/Get-DbaDatabase.ps1]: Actually, that should be the default wherever possible. Whenever you use braces behind where-object, you take a significant performance impact. Doing it once may not be that bad, but it quickly adds up when processing multiple items with that.

File: functions/Get-DbaDatabase.ps1 [hunk 5]
```diff
@@ -0,0 +1,174 @@
...
+			$defaults = 'Name', 'Status', 'ContainmentType', 'RecoveryModel', 'CompatibilityLevel', 'Collation', 'Owner', 'EncryptionEnabled'
+			
+			if ($System)
+			{
+				$inputobject = $server.Databases | Where-Object { $_.IsSystemObject -eq $true }
+			}
+			
+			if ($User)
+			{
+				$inputobject = $server.Databases | Where-Object { $_.IsSystemObject -eq $false }
```
FriedrichWeinmann: For better performance, remove the braces:
`$inputobject = $server.Databases | Where-Object IsSystemObject -eq $false`

ctrlbold: I used this syntax on one of the SPN commands and it crapped the bed so I went back to {}. I do find it's inconsistent. You guys have never run into that issue?

FriedrichWeinmann: Never had one so far.
Admittedly, most of the times I use my own function that does the same (we have still plenty of customers with 6.1 machines and PS2).
If you distrust it too much, I can grab my own filter function and do a pull request for it.
The extra Runspace needed for the scriptblock just adds so much overhead, it really adds up. In this case it's probably trivial, but as a general issue, it needs to be addressed.

functions/Get-DbaDatabase.ps1
gbargsley: Removed, thanks.

functions/Get-DbaDatabase.ps1
gbargsley: So you guys are saying I should make it the following:
`$inputobject = $server.Databases | Where-Object { $_.IsSystemObject }`

functions/Get-DbaDatabase.ps1
gbargsley: That is working, thanks for the suggestion.

functions/Get-DbaDatabase.ps1
gbargsley: I changed to DatabaseOwner to be more descriptive.

functions/Get-DbaDatabase.ps1
wsmelton: No, you can make it just:

```powershell
$inputobject = $server.Databases | Where-Object IsSystemObject
```

Think of it in the same way that we do:

```powershell
if ($SomeTrueValue) {
#do something because it is true
}
```

gbargsley: Got it and made the update.

functions/Get-DbaDatabase.ps1
```diff
@@ -0,0 +1,159 @@
...
+        [ValidateSet('Full', 'Simple', 'BulkLogged')]
+        [string]$RecoveryModel
+	)
+	
+	DynamicParam { if ($SqlInstance) { return Get-ParamSqlDatabases -SqlServer $SqlInstance[0] -SqlCredential $SqlCredential } }
+	
+	BEGIN
+	{
+		$databases = $psboundparameters.Databases
+	}
```
PowerDBAKlaas: Following @Stuart-Moore 's example, you can add $FunctionName = 'Get-DbaDatabase' in BEGIN
and add Write-Verbose on every step start and completion further on.
Like:
Write-Verbose "$FunctionName - Connecting to $instance"

functions/Get-DbaDatabase.ps1
```diff
@@ -0,0 +1,159 @@
...
+Returns all databases that have TDE enabled
+
+.PARAMETER RecoveryModel
+Returns databases for Full, Simple or Bulk Logged recovery models
+
+
+
+	
+.NOTES
+dbatools PowerShell module (https://dbatools.io, clemaire@gmail.com)
```
PowerDBAKlaas: You may add your name as the author here. 👍

ctrlbold:
@gbargsley please ping me when the necessary changes have been made. We are so close!

functions/Get-DbaDatabase.ps1
PowerDBAKlaas: Update:
@FriedrichWeinmann  taught us an even better way:
`$FunctionName = (Get-PSCallstack)[0].Command`
so we'll all use that.

functions/Get-DbaDatabase.ps1
```diff
@@ -0,0 +1,159 @@
+﻿FUNCTION Get-DbaDatabase
+{
+<#
+.SYNOPSIS
+Gets a SQL database object for each database that is present in the target instance of SQL Server.
```
gbargsley: Clarified the verbiage.

functions/Get-DbaDatabase.ps1
```diff
@@ -0,0 +1,159 @@
...
+ SQL Server. If the name of the database is provided, the command will return only this specific database object.
+	
+.PARAMETER SqlInstance
+SQL Server name or SMO object representing the SQL Server to connect to. This can be a collection and recieve pipeline input.
+
+.PARAMETER SqlCredential
+PSCredential object to connect as. If not specified, current Windows login will be used.
+
+.PARAMETER System
+Returns only system databases
```
gbargsley: Changed parameters to Include.

functions/Get-DbaDatabase.ps1
```diff
@@ -0,0 +1,159 @@
...
+
+.PARAMETER User
+Returns only user databases
+	
+.PARAMETER State
+Returns only online databases
```
gbargsley: Corrected phrasing and changed how state parameter is enabled.

functions/Get-DbaDatabase.ps1
gbargsley: Done

functions/Get-DbaDatabase.ps1
gbargsley: I tried to add this and got errors when trying to import.  Not sure I understand how the @FriedrichWeinmann  option is supposed to work and where the code goes.

functions/Get-DbaDatabase.ps1
```diff
@@ -0,0 +1,159 @@
...
+			{
+				Write-Warning "Failed to connect to: $instance"
+				continue
+			}
+			
+			$defaults = 'Name', 'Status', 'ContainmentType', 'RecoveryModel', 'CompatibilityLevel', 'Collation', 'Owner', 'EncryptionEnabled'
+			
+			if ($System)
+			{
+				$inputobject = $server.Databases | Where-Object { $_.IsSystemObject }
```
gbargsley: I am not sure I understand.  From what I can tell from the code is it does not try to filter down a full list of DB's it hits the if statement and loads just the DB's from that parameter.

functions/Get-DbaDatabase.ps1
```diff
@@ -0,0 +1,159 @@
...
+                "ReadOnly" {$inputobject = $server.Databases | Where-Object { $_.ReadOnly }}
+                "ReadWrite" {$inputobject = $server.Databases | Where-Object { $_.ReadOnly -eq $false }}
+            }
+            
+            if ($Encrypted)
+			{
+				$inputobject = $server.Databases | Where-Object { $_.EncryptionEnabled }
+			}
+			
+            switch ($RecoverModel)
```
gbargsley: Fixed

ctrlbold:
Great work, all!

PR #672, opened 2017-02-07
Fix get dbadatabasefreespace

scinobe:
Re-submitting code with the addition of the File Type.  ROWS vs LOG


ConstantineK:
Hopefully I didnt gum it up too bad merging these commits.

ctrlbold:
Works perfect across my whole farm! thanks to you both.

PR #708, opened 2017-02-11
Updating TabExpansionPlusPlus Showcase

FriedrichWeinmann:
Updated the TEPP demo implementation by ...
 - Preventing it being applied to all commands with parameters of the specified name, irrespective of source
 - Only runs when the necessary commands are available, so neither TEPP nor PowerShell v5 become mandatory
 - Prefers using the TEPP implementation, because those can be discovered using Get-ArgumentCompleter


ctrlbold:
closing to reopen

PR #709, opened 2017-02-11
Check for SMO and DMO XPs Enabled on an instance

### LINKED ISSUE #448: Need to add SMO / DMO Disabled check to Connect-SqlServer
Is this a *feature* OR *bug*:
Kind of a bug? 

**Additional comments**
If sp_configure 'SMO and DMO XPs', 0;  then SSMS bails and doesn't even try to connect. People may think that since we're command line, we can bypass a lot of the GUI stuff and thus they still won't need SMO. We should check and bail too.

image

shaneis:
Many thanks to @ConstantineK for the help with this one. Git and I had a few...words...to say about each other.

Fixes #448 

 - Added configuration check for 'SMO and DMO XPs enabled' and errors out if found.
 - 
 - 

How to test this code: 
- [ ] Attempt to connect to an instance with 'SMO and DMO XPs enabled' set to 1.
- [ ] Attempt to connect to an instance with 'SMO and DMO XPs enabled' set to 0.

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally

### DIFF
--- a/internal/Connect-SqlServer.ps1
+++ b/internal/Connect-SqlServer.ps1
@@ -1,4 +1,4 @@
-Function Connect-SqlServer
+﻿Function Connect-SqlServer
 {
 <# 
 .SYNOPSIS 
@@ -70,6 +70,14 @@ Internal function that creates SMO server object. Input can be text or SMO.Serve
 	
 	$server = New-Object Microsoft.SqlServer.Management.Smo.Server $SqlServer
 	$server.ConnectionContext.ApplicationName = "dbatools PowerShell module - dbatools.io"
+
+    if ($server.Configuration.SmoAndDmoXPsEnabled.RunValue -eq 0)
+    {
+        Write-Error "Accessing this server via SQL Management Objects (SMO) or Distributed Management Objects (DMO) is currently not permitted.
+                     Enable the option 'SMO and DMO XPs' on your instance using sp_configure to continue.
+                     Note that this will require 'Show Advanced Options' to be enabled using sp_configure as well."
+        break
+    }
 	
 	try
 	{


wsmelton:
If we state something has to be changed (sp_configure setting) we should offer the command in PowerShell to do it ;)

> Enable the option 'SMO and DMO XPs' on your instance using sp_configure to continue.

Compared to:

> Enable the option 'SMO and DMO XPs' on your instance with `Set-DbaSpConfigure -SqlServer MyInstance -Name 'SMO and DMO XPs' -Value 1`

shaneis:
@wsmelton Good idea actually. I like it a lot better.
If nobody has any objections with this, I'll aim to have a commit Wednesday. 
Busy tomorrow and all that jazz

ctrlbold:
@wsmelton, @shaneis  - Ultimately, I realized that it was usnig SMO to check for SMO -- same goes for `Set-DbaSpConfigure -SqlServer MyInstance -Name 'SMO and DMO XPs' -Value 1` - that command uses SMO as well. And I had to comment it out with a note. Please revisit using pure T-SQL instead.

wsmelton:
I'm not sure I follow, please revisit what? 

The setting 'SMO and DMO XPs' in no way prevents you from hitting an instance utilizing SMO itself. This setting is strictly around the use of **extended procedures** that utilize SMO, which is why SSMS will bark if it is found disabled (or set to zero).

image

So are we saying just because SSMS gets it's jollies using extended stored procedures with SMO that we should not allow dbatools to function at all?

ctrlbold:
This isn't about SSMS, but rather not having to check within every command to see if we can use SMO XPs when we need it.

To my knowledge, we use a several SMO XP's -- any command that uses Test-SqlPath uses it, Find-DbaOrphanedFile and so on. While it's possible to make an index of commands that use xps, either explicitly or implicitly, I don't know if anyone can commit lifelong time or interest to maintain the index and do the research as to what implicitly uses the XPs.

ctrlbold:
okay, just did some testing. thank you for bringing this up @wsmelton 

I was able to perform an entire migration with the XPs being disabled on both sides, so we can skip the check in its entirety for connect-sqlserver. i was also able to execute find-dbaorphanedfile with success, so i actually have no idea which XPs are SMO specific XPs and Microsoft does not provide an index on the first page of google search.

PR #715, opened 2017-02-12
Find-DbaUserObject - added some checks and fixes

ctrlbold:


functions/Find-DbaUserObject.ps1
```diff
@@ -37,292 +37,294 @@ You should have received a copy of the GNU General Public License
...
+Shows all user owned (non-sa, non-dbo) objects and verbose output 
 #>
 	[CmdletBinding()]
 	Param (
 		[parameter(Position = 0, Mandatory = $true, ValueFromPipeline = $True)]
-		[Alias("ServerInstance", "SqlInstance", "SqlServers")]
-		[string[]]$SqlServer,
+		[Alias("ServerInstance", "SqlServer", "SqlInstances")]
+		[object[]]$SqlInstance,
 		[System.Management.Automation.PSCredential]$SqlCredential,
```
FriedrichWeinmann: You may at least want to add a 'Credential' alias for this parameter. While this module is primarily aimed at SQL Admins, it _is_ the PowerShell standard.

functions/Find-DbaUserObject.ps1
```diff
@@ -37,292 +37,294 @@ You should have received a copy of the GNU General Public License
...
-                        $out
-                }
-            }  
-
-        }
-    }
+	process
+	{
+		if ($Pattern -like "*\*")
+		{
```
FriedrichWeinmann: non-pipeline parameters should be processed in the begin loop.
Why you'd escape the escape character (and do it only once) is a bit of a mystery.
If you do not want to to be actual regex, you could also use the `[regex]::Escape("...")` method, to escape all of it.

functions/Find-DbaUserObject.ps1
```diff
@@ -37,292 +37,294 @@ You should have received a copy of the GNU General Public License
...
+			$slashcount = ($Pattern -split "\\" | Measure-Object | Select-Object -ExpandProperty Count) - 1
+			if ($slashcount -eq 1)
+			{
+				Write-Verbose "Too few slashes, adding extra as required by regex"
+				$Pattern = $Pattern.Replace('\', '\\')
+				Write-Verbose "Pattern now $pattern"
+			}
+		}
+		
+		Write-Verbose "Starting"
```
FriedrichWeinmann: Inconsistent behavior and little extra value:
If you want to announce once that the function starts, move it to the begin Block.
If you want to mention it for each Instance it should process, move it inside the foreach loop and include the instance-name.

functions/Find-DbaUserObject.ps1
ctrlbold: Thanks for the review!!

You'll find that some commands do have Credential. We have decided in dbatools.io/tstyle that we'll be reserving Credential for Windows/Linux and SqlCredential for SQL Server itself.

FriedrichWeinmann: I found that's its unresolved and in discussion

FriedrichWeinmann: Check your Trello ^^

functions/Find-DbaUserObject.ps1
ctrlbold: Re: adding a begin - it's true. i was bein lazy 👎 

This was intended to fix people who pass ad\username instead of ad\\username. Do you have the regex that'd only match what i'm looking to fix? I dont want to escape everything but i do want to catch the most common error.

functions/Find-DbaUserObject.ps1
ctrlbold: agreed, removed

functions/Find-DbaUserObject.ps1
ctrlbold: LOL, you check it! (I'll move that card.)

Issue #685, opened 2017-02-08
Get-DbaSpn - Users expect it to act like setspn -l domain\account


ctrlbold:
### Is this a *feature* OR *bug*:

Both? Needs docs

### Details
Basically, someone thought some entries were missing 

```
> setspn -l
MSSQLSvc/CORPSQL1.system.CORPDOMAIN:50141
MSSQLSvc/CORPSQL1:50141
MSSQLSvc/CORPSQL1\PORTOCENTRE:50141
MSSQLSvc/CORPSQL1.system.CORPDOMAIN\PORTOCENTRE:50141

> Get-DbaSpn -ComputerName CORPSQL1 | select SPN
MSSQLSvc/CORPSQL1.system.CORPDOMAIN
MSSQLSvc/CORPSQL1.system.CORPDOMAIN:1433

> setspn -l
mssqlsvc/CORPSQL2
mssqlsvc/CORPSQL2.system.CORPDOMAIN:1433
mssqlsvc/CORPSQL2:1433
mssqlsvc/CORPSQL2.system.CORPDOMAIN\portocentre
mssqlsvc/CORPSQL2\portocentre
mssqlsvc/CORPSQL2:49850
mssqlsvc/CORPSQL2.system.CORPDOMAIN:49580

> Get-DbaSpn -ComputerName CORPSQL2 | select SPN
MSSQLSvc/CORPSQL2.system.CORPDOMAIN:1433
MSSQLSvc/CORPSQL2.system.CORPDOMAIN:49580
```

Ultimately, he had to run

```
Get-DbaSpn -AccountName domain\sqlserver | Where-Object SPN -like *CORPSQL1* | Select SPN
```

To get it to work as he expected. Which is fine, but we need to clarify both in Get-DbaSpn's Help and on the webpage.

forensicsguy20012004:
Yeah I think I found something along these lines as well

result of Get-DbaSpn

> ComputerName  : Myserver.domain.net
InstanceName  : MSSQLSERVER
SqlProduct : SQL Server 20xx Enterprise Edition (64-bit)
InstanceServiceAccount : NT SERVICE\MSSQLSERVER
RequiredSPN: MSSQLSvc/Myserver.domain.net:port
IsSet : False
Cluster : False
TcpEnabled  : True
Port  : MyPort
DynamicPort : False
Warning  : None
**Error : SPN missing**

However with setspn -l Myserver:
> MSSQLSvc/Myserver.domain.net:Port
[...]

PR #718, opened 2017-02-13
Get-Spn to better align with setspn -l

### LINKED ISSUE #713: BUG: Get-DbaSpn
bug

Latest version release of dbatools

result of Get-DbaSpn

> ComputerName : Myserver.domain.net
InstanceName : MSSQLSERVER
SqlProduct : SQL Server 20xx Enterprise Edition (64-bit)
InstanceServiceAccount : NT SERVICE\MSSQLSERVER
RequiredSPN: MSSQLSvc/Myserver.domain.net:port
IsSet : False
Cluster : False
TcpEnabled : True
Port : MyPort
DynamicPort : False
Warning : None
Error : SPN missing

However with setspn -l Myserver:

> MSSQLSvc/Myserver.domain.net:Port
MSSQLSvc/Myserver.domain.net
Since my SQL is running at [NT SERVICE\MSSQLSERVER], Get-DbaSpn does not show a valid SPN but the old school setspn -l does show a valid SPN.

Looks like something that just needs to be investigated a little bit further. Either way...AWESOME WORK.

I made comment of this same issue on: https://github.com/sqlcollaborative/dbatools/issues/685. I think you could probably close this issue if you address what I found in https://github.com/sqlcollaborative/dbatools/issues/685

thanks

ctrlbold:
Any suggestions on the new parameter name? 
fixes #713 and #685

### DIFF
--- a/functions/Get-DbaSpn.ps1
+++ b/functions/Get-DbaSpn.ps1
@@ -17,7 +17,7 @@ The accounts you want to retrieve set SPNs for.
 .PARAMETER Credential
 User credential to connect to the remote servers or active directory. This is a required parameter.
 	
-.PARAMETER ShowAccountNameAll
+.PARAMETER ByAccount
 Shows all SPNs registered by the specified AccountName otherwise, only results will be shown for the specified ComputerName (which is localhost by default)
 
 .NOTES 
@@ -53,23 +53,14 @@ Returns a custom object with SearchTerm (ServerName) and the SPNs that were foun
         [string[]]$ComputerName = $env:COMPUTERNAME,
         [Parameter(Mandatory = $false)]
 		[string[]]$AccountName,
-		[switch]$ShowAccountNameAll,
+		[switch]$ByAccount,
 		[Parameter(Mandatory = $false)]
         [PSCredential]$Credential
-    )
-	process
-	{	
-		if ($ComputerName)
-		{
-			if ($ComputerName[0].EndsWith('$'))
-			{
-				$AccountName = $ComputerName
-				$ComputerName = $null
-			}
-		}
-		
-		if ($ShowAccountNameAll)
-		{
+	)
+	begin
+	{
+		Function Process-Account ($AccountName, $ByAccount) {
+			
 			ForEach ($account in $AccountName)
 			{
 				$ogaccount = $account
@@ -130,7 +121,6 @@ Returns a custom object with SearchTerm (ServerName) and the SPNs that were foun
 							$serviceclass = ($spn -Split "\/")[0]
 						}
 					}
-					
 					[pscustomobject] @{
 						Input = $ogaccount
 						AccountName = $ogaccount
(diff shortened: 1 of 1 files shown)


ctrlbold:
What I did was set ComputerName to $env:COMPUTERNAME by default so if only an $accountname was specified, it'd be $env:COMPUTERNAME + $accountname, like setspn. 

This effectively removed the functionality to search for all SPNs set by $accountname so I added a new switch (crappy name, feel free to suggest alts) named $ShowAccountNameAll that will show all SPNs registered to that $accountname on the domain.

functions/Get-DbaSpn.ps1
```diff
@@ -71,76 +68,81 @@ Returns a custom object with SearchTerm (ServerName) and the SPNs that were foun
 			}
 		}
 		
-		ForEach ($account in $AccountName)
+		if ($ShowAccountNameAll)
```
FriedrichWeinmann: Please use regions to split different sections of your code.
When using PSS:
Mark all of one logical section.
Press "CTRL" + "R"
Give it an acceptable name.
Makes it easier to read and maintain

functions/Get-DbaSpn.ps1
```diff
@@ -71,76 +68,81 @@ Returns a custom object with SearchTerm (ServerName) and the SPNs that were foun
...
-					ServiceClass = "MSSQLSvc" # $serviceclass
-					Port = $port
-					SPN = $spn
-                }
 			}
+			continue
 		}
 		
-		
 		foreach ($server in $ComputerName)
```
FriedrichWeinmann: It would be customary to use $Computer, rather than $Server here.

FriedrichWeinmann:
Set it to approved, since I'm going to bed now - you'll know yourself well enough when it's as done as it gets :)

functions/Get-DbaSpn.ps1
```diff
@@ -48,20 +50,15 @@ Returns a custom object with SearchTerm (ServerName) and the SPNs that were foun
     [cmdletbinding()]
 	param (
         [Parameter(Mandatory = $false,ValueFromPipeline = $true)]
-        [string[]]$ComputerName,
+        [string[]]$ComputerName = $env:COMPUTERNAME,
         [Parameter(Mandatory = $false)]
 		[string[]]$AccountName,
+		[switch]$ShowAccountNameAll,
```
ctrlbold: I like ByAccount, thanks! 

The parametersets would work to keep ByAccount and AccountName together - but there are definitely times when users would want to check a remote host with another AccountName. Brain isn't working, pass for now.

functions/Get-DbaSpn.ps1
ctrlbold: 1.0 ;)

functions/Get-DbaSpn.ps1
ctrlbold: done!

functions/Get-DbaSpn.ps1
```diff
@@ -48,20 +50,15 @@ Returns a custom object with SearchTerm (ServerName) and the SPNs that were foun
...
 	process
-	{
-		if (!$ComputerName -and !$AccountName)
-		{
-			$ComputerName = $env:computername
-			$AccountName = "$env:USERDOMAIN\$env:USERNAME"
-		}
-		
+	{	
 		if ($ComputerName)
```
ctrlbold: true, i had to move it a bunch of stuff to ensure this still worked, please recheck my logic.

@drewfurgiuele - please check this PR and approve. I changed the functionality.

functions/Get-DbaSpn.ps1
wsmelton: Regions are not that useful at seperating logic in your code unless the editor that you use supports it. Which not every editor used by contributors is going to support it, so it is not really any different than just adding comments.

functions/Get-DbaSpn.ps1
```diff
@@ -71,76 +68,81 @@ Returns a custom object with SearchTerm (ServerName) and the SPNs that were foun
...
+				
+				Write-Verbose "Looking for account $account..."
+				
+				try
+				{
+					$Result = $adsearch.FindOne()
+				}
+				catch
+				{
+					Write-Warning "AD lookup failure. This may be because the hostname ($computer) was not resolvable within the domain ($domain) or the SQL Server service account ($serviceaccount) couldn't be found in domain."
```
ctrlbold: Can you give the syntax plz? LOL :/. Then I'll remember it.

ctrlbold:
Will add Stop-Process in 1.1

PR #725, opened 2017-02-13
Initial commit for Find-DbaCommand, connects to #692

### LINKED ISSUE #692: Find-DbaCommand
This new command will be a speedier version to search dbatools commands (e.g. with tags, help text, etc) because get-help performance is sloooowww

niphlod:
Fixes #692

Most of this command revolves around what we want to parse and how to parse it. Once that is figured out, it'll be pretty easy to implement. Right now "tag" and "author" are searched into .NOTES, with the simplest of the rule possible: a line starting with "Tags:" and one starting with "Author:".
I'd stick with the basics: we'll need a bit of rewrite here and there for modules but once the good behaviour is promoted with a default template it'll be easy for everybody

TODO: 
- integrating with TE++ for autocompletion of tags
- agreeing on where to put the index file
- adding useful filters (?)

How to test this code: 
- [ ] when the index is not found, it'll be in the "Documents" folder
- [ ] when the index is not found, it is built automatically
- [ ] you can do a full-text search with -Pattern
- [ ] you can search by tag and by author

- [ ]  Powershell 3
- [ ]  Windows 7

Tests for tester:
- [ ] Works locally

### DIFF
--- a/functions/Find-DbaCommand.ps1
+++ b/functions/Find-DbaCommand.ps1
@@ -10,10 +10,15 @@ Finds dbatools commands searching through the inline help text, building a conso
 .PARAMETER Tag
 Finds all commands tagged with this tag
 
+.PARAMETER Author
+Finds all commands tagged with this author
+
+.PARAMETER Rebuild
+Rebuilds the index
+
 .NOTES
 Tags: find
 Author: niphlod
-
 
 dbatools PowerShell module (https://dbatools.io, clemaire@gmail.com)
 Copyright (C) 2016 Chrissy LeMaire
--- a/functions/Find-DbaCommand.ps1
+++ b/functions/Find-DbaCommand.ps1
@@ -16,6 +16,15 @@ Finds all commands tagged with this author
 .PARAMETER Rebuild
 Rebuilds the index
 
+.PARAMETER Pattern
+Searches help for all commands in dbatools for the specified pattern and displays all results
+
+.PARAMETER Confirm
+Confirms overwrite of index
+
+.PARAMETER WhatIf
+Displays what would happen if the command is run
+
 .NOTES
 Tags: find
 Author: niphlod
--- a/functions/Find-DbaCommand.ps1
+++ b/functions/Find-DbaCommand.ps1
@@ -107,7 +107,7 @@ Finds all commands searching the entire help for "snapshot", rebuilding the inde
                 $x = Get-DbaHelp $c.Key
                 $helpcoll.Add($x)
             }
-            $dest = Get-DbaIdxPath
+            $dest = Get-ConfigValue -Name 'Path.TagCache' -Fallback "$(Resolve-Path $PSScriptRoot\..)\dbatools-index.json"
             if ($Pscmdlet.ShouldProcess($dest, "Recreating index"))
             {
                 $helpcoll | ConvertTo-Json | Out-File $dest
--- a/functions/Find-DbaCommand.ps1
+++ b/functions/Find-DbaCommand.ps1
@@ -1,0 +1,146 @@
+Function Find-DbaCommand
+{
+<#
+.SYNOPSIS
+Finds dbatools commands searching through the inline help text
+
+.DESCRIPTION
+Finds dbatools commands searching through the inline help text, building a consolidated json index and querying it because Get-Help is too slow
+
+.PARAMETER Tag
+Finds all commands tagged with this tag
+
+.NOTES
+Tags: find
+Author: niphlod
+
+
+dbatools PowerShell module (https://dbatools.io, clemaire@gmail.com)
+Copyright (C) 2016 Chrissy LeMaire
+
+This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+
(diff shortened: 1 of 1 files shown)


ctrlbold:
love it! works so well. Going to add some tags in time for our next non-snowball release 👍 

Can you please put the json file in `$PSScriptRoot\internal\` instead?

niphlod:
No problem but are we sure it's always writable?

Il 13 feb 2017 23:41, "Chrissy LeMaire" <notifications@github.com> ha
scritto:

functions/Find-DbaCommand.ps1
```diff
@@ -0,0 +1,146 @@
...
+        function Get-DbaIndex() {
+            $dbamodule = Get-Module dbatools
+            $allcommands = $dbamodule.ExportedCommands
+            $helpcoll = New-Object System.Collections.Generic.List[System.Object]
+            foreach($c in $allcommands.GetEnumerator()) {
+                $x = Get-DbaHelp $c.Key
+                $helpcoll.Add($x)
+            }
+            $dest = Get-DbaIdxPath
+            if ($Pscmdlet.ShouldProcess($dest, "Recreating index"))
```
FriedrichWeinmann: You may want to move this to the top of the helper function.
Lots of processing time wasted if the user negates here.

functions/Find-DbaCommand.ps1
```diff
@@ -0,0 +1,146 @@
...
+            ## fetch the synopsis
+            $thebase.Synopsis = $thishelp.Synopsis
+
+            ## store notes
+            $as = $thishelp.alertSet | Out-String -Width 120
+
+            ## fetch the tags
+            $tags = $tagsrex.Match($as).Groups[1].Value
+            if($tags) {
+                $thebase.Tags = $tags.Trim().Split(',')
```
FriedrichWeinmann: I'd split first and trim then.
That way, tags can safely have a whitespace behind the separating comma.

functions/Find-DbaCommand.ps1
niphlod: sure, TBF I added it at the last moment just because out-file without a shouldprocess gets tagged by scriptanalyzer.

ctrlbold:
@niphlod - since I will be doing the rebuilds, yes. I'd prefer putting the convenience and neatness of our access over a theoretical -rebuild by a user that they wont need because I'll rebuild as part of each release.

FriedrichWeinmann:
@ctrlbold , @niphlod : I'm almost done with my configuration system (final internal tests in fact). You could place the actual path as a configuration value that _defaults_ to internal, but can be changed by the user.
That would suit both sides, since then the user can call it even in the few situations where doesn't have the write access (basically, if it's installed for system on a computer he doesn't have local admin for). Wouldn't change much code either.

ctrlbold:
@FriedrichWeinmann  works for me!

functions/Find-DbaCommand.ps1
```diff
@@ -0,0 +1,145 @@
...
+
+For rigorous typers: finds all commands searching the entire help for "snapshot"
+
+.EXAMPLE
+Find-DbaCommand -Tag copy
+
+Finds all commands tagged with "copy"
+
+.EXAMPLE
+Find-DbaCommand -Author "chrissy"
```
ctrlbold: lol!

ctrlbold:
Will be adding some tags on Thursday to make this useful even before v1 💃

FriedrichWeinmann:
There should already be some on my config functions :)

functions/Find-DbaCommand.ps1
```diff
@@ -0,0 +1,154 @@
...
+
+Finds every command whose author contains our beloved "chrissy"
+
+.EXAMPLE
+Find-DbaCommand -Pattern "snapshot" -Rebuild
+
+Finds all commands searching the entire help for "snapshot", rebuilding the index (good for developers)
+#>
```
PowerDBAKlaas: Maybe you can add an example using multiple params, like:
Find-DbaCommand -Author JFK -Tag speeding

functions/Find-DbaCommand.ps1
```diff
@@ -0,0 +1,154 @@
...
+            Get-DbaIndex
+            Write-Verbose "Rebuild done in $($swrebuild.Elapsedmilliseconds)ms"
+
+        }
+        $consolidated = Get-Content $idxfile | ConvertFrom-Json
+        $result = $consolidated
+        if($Pattern.length -gt 0) {
+            $result = $result | Where-Object { $_.psobject.properties.value -like "*$Pattern*" }
+        }
+        if($Tag.length -gt 0) {
```
PowerDBAKlaas: Could you do a split on $Tag too and return the commands that have multiple tags?

Find-DbaCommand -Tag copy,user

PowerDBAKlaas:
Could we have Pester tests that check if Author and Tags are present in every file, and written following the guidelines?
@SQLDBAWithABeard @ctrlbold

ctrlbold:
I really like that idea @PowerDBAKlaas

ctrlbold:
currently working on retro tagging

SQLDBAWithABeard:
Sure can

ClaudioESSilva:
Maybe worth have a list of "approved tags" so we can reuse and not duplicate them? Does this make sense?

niphlod:
note to everyone. merging without needed mods makes life harder for the original contributor :P .

Going to issue another PR with the proposed fixes

PR #733, opened 2017-02-14
Configuration System

FriedrichWeinmann:
Implemented a configuration system for system values the dbatools use.
Currently much information is hardcoded in individual functions, that actually should be configurable.
This PR addresses that:
 - All configuration options stored in a central directory
 - All configured values can be discovered, using Get-DbaConfig
 - Users can add configuration values to their profile and thus override module defaults


FriedrichWeinmann:
Also added a framework to include a framework to add C# code to, in case more of that comes along.
Redesigned the way optional components become optional as well while I was at it - now you don't need to update the .psm1 each time, which was ... not so good.

ctrlbold:
Thanks so much, @FriedrichWeinmann! can you take a look at the test failures? I'm getting this on almost every test.

Error occurred in test script 'C:\projects\dbatools\Tests\Test-DbaDatabaseOwner.Tests.ps1' 719ms
   CommandNotFoundException: The term 'C:\projects\dbatools\internal\configurations' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.

FriedrichWeinmann:
on it

FriedrichWeinmann:
Alright, got to the root of the problem with the tests:
They invoked every single file within internal, without making sure those are files first.
Not very successful on folders ...

Btw, these tests desperately need some looking after :(
I'd love to, but only so many hours in a given day. If I have the spare time after finishing my current workload (including that mirror command, which I haven't forgotten in all this), I'll look at things myself, but ... well, pre-freeze priorities ;)

ctrlbold:
@FriedrichWeinmann we have ppl who are doing it -- i think the only reason the tests are failing is because we're messing with the directory structure, no?

FriedrichWeinmann:
That was the reason _this_ time.
We have a lot of redundancy in there that could use some eliminating. For example, while they are called by appveyor.pester.ps1, they all test content of the same module in the same version.
Still we process and import the module in _each_ test, rather than consolidating it in the primary file before calling invoke-pester.
The issue now was called due to a totally, 110% superfluous line in the source (I killed it completely). When we already have the entire module imported, it serves no purpose to explicitly import some of its files.
This is not a level 1 emergency, no - I'd have jumped right into the action if it were - but it still deserves cleaning up, so we don't multiply code unnecessarily. These are the templates new tests will be based upon, so if we leave this unchecked, we might just run into an issue that may have been easy to fix now but will be a lot of work then.
Sorry about calling it `desperately` - that was an exaggeration, probably based on my own frustration hunting down this particular issue - that was an overreaction.

ctrlbold:
This looks super cool! Is it possible to change the namespace? I know it's small but dbatools is all lower case - I rarely mention it / ppl to change cuz no bigs but for our internal stuff, I'd like it properly cased.

From:
SqlCollective.DbaTools

To:
sqlcollaborative.dbatools

Will you be able to write a short and sweet blog post on the background/intention for our developers? No need to be longer than Drew or Stuart's.

Thank you so much! 🚎

ctrlbold:
I'm especially pumped about that XML file!

ctrlbold:
Oh, and stop modifying the psd1 or I'll come after you with 🛠 - I can add entries at any time and even before you write the commands, it helps everyone with the branches and I'm happy to do it.

FriedrichWeinmann:
* ducksbehindcorner *
Yes, My Lord!

FriedrichWeinmann:
I promise to be good from now on, Master! ;)
Seriously, this habit is _soooo_ hard to lose :(
I'll try to remember

FriedrichWeinmann:
While it hurts my soul and shakes my faith in humanity - neigh, existence itself! - I guess I could change the casing on the namespaces. It'd be about the equivalent to asking a priest to join the next witches' sabbath so he can help officiate over the child sacrifice, but ... ;)
Again, one of those habit things.
No technical reason to lowercase it though, so gimme a sec (or better a few minutes) and the casing shall be done, and done right.

ctrlbold:
You like Linux, right? Think of it like Linux.. lower case everythang 🐧

FriedrichWeinmann:
The best thing about penguins are their livers. Preferably Well Done ;)

ctrlbold:
I had puffin in Iceland. Broke my heart but dammmmn was it amazing!

PR #734, opened 2017-02-14
Read-DbaBackupHeader - fixed for sql2005 which does not have CompressedBackupSize

ctrlbold:


ctrlbold:
@FriedrichWeinmann - super appreciate it, thank you!

PR #745, opened 2017-02-15
Ignore tests that appveyor fails but locals pass

ctrlbold:


ctrlbold:
Thankfully, that worked

PR #750, opened 2017-02-15
Restore-DbaBackup - Quickfix UNC vs Local

ctrlbold:


functions/Restore-DbaDatabase.ps1
```diff
@@ -178,8 +178,31 @@ folder for those file types as defined on the target instance.
...
+		else
+		{
+			if ($Path.StartsWith("\\") -eq $false)
+			{
+				# Many internal functions parse using Get-ChildItem. 
+				# We need to use Test-SqlPath and other commands instead
+				# Prevent people from trying 
+				
+				Write-Warning "Currently, you can only use UNC paths when running this command remotely. We expect to support local paths shortly."
+				continue
```
FriedrichWeinmann: Danger!
Continue will affect the calling function if any! Use return instead.
Or better yet: Use Stop-Function.

ctrlbold:
hey @Stuart-Moore  - take a look at this PR when you return. I didn't wanna mess too much so I just prevented ppl from using remote servers with local paths. ie

`Restore-DbaDatabase -SqlServer sql2016 -WithReplace -Path C:\temp\SQL2016_Cube_Query_History_FULL_20170206_115448.bak`

We have Test-SqlPath that tests from the SQL Server's perspective but will need to create new internal functions to do the equiv of gci *.bak etc.

functions/Restore-DbaDatabase.ps1
ctrlbold: Not in BEGIN. Try it. It stops everything. If you use return in BEGIN, it moves on to PROCESS

ctrlbold: Sorry man, I'm about to keel over from stuff I gotta do and I haven't looked into the syntax of Stop-Function.

functions/Restore-DbaDatabase.ps1
```diff
@@ -221,8 +221,30 @@ folder for those file types as defined on the target instance.
...
     PROCESS
     {
-        
-        foreach ($f in $path)
+		$base = $SqlServer.Split("\")[0]
+		
+		if ($base -eq "." -or $base -eq "localhost" -or $base -eq $env:computername -or $base -eq "127.0.0.1")
+		{
+			# they can use whatever path they like
+		}
```
FriedrichWeinmann: Move the base processing back into begin.
store whether it's local or not in
$IsLocal
Will not catch IPv6 localhost

functions/Restore-DbaDatabase.ps1
```diff
@@ -221,8 +221,30 @@ folder for those file types as defined on the target instance.
...
+		else
+		{
+			if ($Path.StartsWith("\\") -eq $false)
+			{
+				# Many internal functions parse using Get-ChildItem.
```
FriedrichWeinmann: This will fail if the user passes a list of paths without using the pipeline (cause then it's an array).
Move this into the `foreach ($f in $path)` loop below. Add $IsLocal to the check, since you can't do if/else across begin/process.

functions/Restore-DbaDatabase.ps1
```diff
@@ -221,8 +221,30 @@ folder for those file types as defined on the target instance.
...
+		{
+			if ($Path.StartsWith("\\") -eq $false)
+			{
+				# Many internal functions parse using Get-ChildItem. 
+				# We need to use Test-SqlPath and other commands instead
+				# Prevent people from trying 
+				
+				Write-Warning "Currently, you can only use UNC paths when running this command remotely. We expect to support non-UNC paths for remote servers shortly."
+				return
+
```
FriedrichWeinmann: When you move this into the foreach loop, you _can_ use continue. Here's the equivalent in Stop-Function:
` Stop-Function -Message "Currently, you can only use UNC paths when running this command remotely. We expect to support non-UNC paths for remote servers shortly." -Silent $Silent -Category InvalidArgument -Target $f -Continue`
This will write a warning and continue with the next path (if any) if not in silent mode. In silent mode it will throw an interrupting error and crash the function.
You'd also need to add a `$Silent` switch parameter to the function.

ctrlbold: mm out of braindwidth to add anything new. will change return to continue but it seems like it'd give the same thing over and over.

FriedrichWeinmann: Naw. When you call "continue" within a loop (such as the foreach-loop here), it will stop the current iteration and move to the next item. If it was the last item, it will gracefully leave the loop and continue with End {}

FriedrichWeinmann: So so long as you move the thing to within the loop, continue is safe.

functions/Restore-DbaDatabase.ps1
ctrlbold: thanks, will add ipv6 localhost in future $islocal scriptwide var

functions/Restore-DbaDatabase.ps1
ctrlbold: i get how understand works within the continue, but i assuming that the person will pass the same type of path over and over again so it's just sorta a waste to keep trying but maybe that's a bad assumption. changed to continue.

plz plz can i has approval, im so done with this PR

ctrlbold:
YEAHHH PARTYYYY!

PR #751, opened 2017-02-15
Added new function Get-DbaQueryStoreConfig

Evdlaar:
Function retrieves the Query Store configuration for every database
inside a SQL Server 2016 Instance and returns it through a table.

### DIFF
--- a/functions/Get-DbaQueryStoreConfig.ps1
+++ b/functions/Get-DbaQueryStoreConfig.ps1
@@ -1,0 +1,176 @@
+Function Get-DbaQueryStoreConfig
+{
+<#
+.SYNOPSIS
+Get the Query Store configuration for all Query Store enabled databases.
+	
+.DESCRIPTION
+Retrieves and returns the Query Store configuration for every database that has the Query Store feature enabled.
+
+Default output includes
+	
+.PARAMETER SqlServer
+The SQL Server that you're connecting to.
+
+.PARAMETER Credential
+Credential object used to connect to the SQL Server as a different user.
+
+.PARAMETER Databases
+Return information for only specific databases
+
+.PARAMETER Exclude
+Return information for all but these specific databases
+
+.NOTES
+Author: Enrico van de Laar ( @evdlaar )
+
+dbatools PowerShell module (https://dbatools.io)
+Copyright (C) 2016 Chrissy LeMaire
+This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+You should have received a copy of the GNU General Public License along with this program. If not, see http://www.gnu.org/licenses/.
+
+
+.LINK
+https://dbatools.io/Get-QueryStoreConfig
+
+.EXAMPLE
+Get-DbaQueryStoreConfig -SqlServer ServerA\sql
+
+Returns Query Store configuration settings for every Query Store enabled database on the ServerA\sql instance.
+
+	
+#>
+	[CmdletBinding()]
+	Param (
+		[parameter(Mandatory = $true, ValueFromPipeline = $true)]
+		[Alias("ServerInstance", "SqlInstance")]
+		[string[]]$SqlServer,
+		[PsCredential]$Credential
+	)
+
+	DynamicParam {
+		if ($SqlServer) {
+			return Get-ParamSqlDatabases -SqlServer $SqlServer[0] -SqlCredential $Credential
+		}
+	}
+
+	BEGIN
+	{
+		# Convert from RuntimeDefinedParameter object to regular array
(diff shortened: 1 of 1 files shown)


functions/Get-DbaQueryStoreConfig.ps1
```diff
@@ -0,0 +1,176 @@
...
+				else
+				{
+					Write-Warning "Can't connect to $servername. Moving on."
+					Continue
+				}
+			}
+
+            $sqlVersion = $server.Version
+
+            $sqlVersion = $sqlVersion.ToString().Split(".")[0]
```
wsmelton: You can exclude this line altogether by pulling the `VersionMajor` property instead of having to parse the output of `Version`:

image

functions/Get-DbaQueryStoreConfig.ps1
```diff
@@ -0,0 +1,176 @@
...
+
+            if($sqlVersion -ilt "13")
+                {
+
+                Write-Warning "The SQL Server Instance has a lower SQL Server version than SQL Server 2016. Skipping server."
+                break
```
wsmelton: I would add the `$servername` to this warning so it clear which one is being skipped.

functions/Get-DbaQueryStoreConfig.ps1
```diff
@@ -0,0 +1,176 @@
...
+                }
+
+            
+            # We have to exclude all the system databases since they cannot have the Query Store feature enabled
+			$dbs = $server.Databases | Where-Object { $_.name -ne 'TempDb' -and $_.name -ne 'master' -and $_.name -ne 'msdb' -and $_.name -ne 'model' }
+
```
wsmelton: Seem to recall we have an internal function for getting databases for a given instance...or maybe we are adding that for v1.0 of the module.

functions/Get-DbaQueryStoreConfig.ps1
```diff
@@ -0,0 +1,176 @@
...
+			if ($databases.count -gt 0)
+			{
+				$dbs = $dbs | Where-Object { $databases -contains $_.Name }
+			}
+
+			if ($exclude.count -gt 0)
+			{
+				$dbs = $dbs | Where-Object { $exclude -notcontains $_.Name }
+			}
+
```
wsmelton: Just preference maybe, but if you look at [this line from Copy-SqlJob](https://github.com/sqlcollaborative/dbatools/blob/master/functions/Copy-SqlJob.ps1#L118) it has a bit cleaner method of filtering out based on the include and exclude filters. You could translate that to the database list and would save a few lines.

functions/Get-DbaQueryStoreConfig.ps1
```diff
@@ -0,0 +1,176 @@
...
+                $QSCDT_Row.CaptureMode = $db.QueryStoreOptions.QueryCaptureMode
+                $QSCDT_Row.CleanupMode = $db.QueryStoreOptions.SizeBasedCleanupMode
+                $QSCDT_Row.StaleQueryThreshold = $db.QueryStoreOptions.StaleQueryThresholdInDays
+
+                $QSCDT.Rows.Add($QSCDT_Row)
+                       
+			}
+        }
+
+    $QSCDT | Format-Table -Property Instance, Database, ActualState, FlushInterval, CollectionInterval, MaxSize, CurrentSize, CaptureMode, CleanupMode, StaleQueryThreshold -Wrap -AutoSize
```
wsmelton: `Format-Table` should be removed, this causes issues if the output needs to be piped to another command at some point. As well, we should output objects and format commands do not do that.

You can take a look at `Get-DbaProcess` to see use of an [internal function for outputting objects](https://github.com/sqlcollaborative/dbatools/blob/master/functions/Get-DbaProcess.ps1#L184).

functions/Get-DbaQueryStoreConfig.ps1
ctrlbold: yep! It'll be 1.0

functions/Get-DbaQueryStoreConfig.ps1
ctrlbold: In addition, please use continue instead of break, looking good so far!

functions/Get-DbaQueryStoreConfig.ps1
ctrlbold: This is an interesting one, while we could add the fields we need to the datatable that is $db, I'd go with a pscustomobject

You're going to love custom objects -- very easy. No need to define column names or types. 

In this case, I'd do

[pscustomobject]@{
Instance = $servername
ActualState = $db.QueryStoreOptions.ActualState
Database = $db.name
FlushInterval = $db.QueryStoreOptions.DataFlushIntervalInSeconds
etcetc
}

etc and so on. You can order it how you need in the object itself. In addition, you do not need to assign it to any variable - just output right to the pipeline. You can see that in action here:

https://github.com/sqlcollaborative/dbatools/blob/master/functions/Get-DbaMaxMemory.ps1#L73

As for Format-Table, Shawn is correct -- the end user should be the one to use it. It turns data into something pretty but not usable. You can use Select-DefaultView to hide objects (like SMO Database objects) but they are optional.

ctrlbold:
wtf no idea why it repeated -- sorry!

PR #753, opened 2017-02-15
Adding Stop-Function to Restore-DbaDatabase, Updated module components import

FriedrichWeinmann:
A simple change, just because I could leave 'well' be ;)


FriedrichWeinmann:
Damn, this git is troublesome.
Added totally unrelated changes:

Fixed library import
Changed subordinate call in configurations, because with the new import $PSScriptRoot is no longer available
Updated Get-GenericArgumentCompleter, in order to support the redesigned optional files component.

ctrlbold:
Find-DbaCommand still works <3 So I'm good.

PR #760, opened 2017-02-16
Write-Message

FriedrichWeinmann:
Introduces a central node for writing message, verbose, debugging or warning messages.


FriedrichWeinmann:
Note on how this changes a function:
 - The function will need a `-Silent` switch-parameter.
 - The message needs a level (integer-valued). Verbose is by default 5, Debug is 8 and Host is 2

Before:
`Write-Verbose "Message"`

After
`Write-Message -Message "Message" -Level 5 -Silent $Silent

ctrlbold:
wait, end users would need to use that and we pass it on? Why silent?

FriedrichWeinmann:
End users don't use it.
Silent is the global switch to shut up. It will _not_ affect verbose, debug and logging.
It will suppress visible warnings and messages to host

FriedrichWeinmann:
It is also used in Stop-Function

ctrlbold:
cool k 🍶

FriedrichWeinmann:
The level thing is so the user can choose (using the configuration system) how verbose the module is. So he can easily set all functions to write all messages to Write-Information (Or Write-Host if he's on PS3-4).
Or just some of them.
Also control what messages go to Write-Verbose or Write-Debug

PR #762, opened 2017-02-16
Dba backup the return

Stuart-Moore:
Fixes # 

 - Striping is back either by file count in a single folder or across multiple directories
 - Appears to be remote/local safe (all path checks done via SQL Server)
 - 

How to test this code: 
- [ ] 
- [ ] 

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally

### DIFF
--- a/functions/Backup-DbaDatabase.ps1
+++ b/functions/Backup-DbaDatabase.ps1
@@ -48,6 +48,9 @@ Accepted values are Full, Log, Differential, Diff, Database
 THis is the number of striped copies of the backups you wish to create
 This value is overwritten if you specify multiple Backup Directories
 
+.PARAMETER CreateFolder
+If switch enabled then each databases will be backed up into a seperate folder on each of the specified backuppaths
+
 .PARAMETER DatabaseCollection
 Internal parameter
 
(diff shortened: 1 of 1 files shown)


functions/Backup-DbaDatabase.ps1
```diff
@@ -108,9 +108,14 @@ Backs up AdventureWorks2014 to sql2016's C:\temp folder
 		if ($SqlInstance.length -ne 0)
 		{
 			if ($PSCmdlet.ParameterSetName -ne 'Pipe')
-			{
+			{	
+				Write-Verbose "$Functionname - Databases paramter used"
```
bluekeys: Typo: Should be parameter not paramter

functions/Backup-DbaDatabase.ps1
```diff
@@ -71,27 +75,47 @@ Backs up AdventureWorks2014 to sql2016's C:\temp folder
...
 		[string]$BackupFileName,
+		[parameter(ParameterSetName = "NoPipe")][parameter(ParameterSetName = "Pipe")]
 		[switch]$NoCopyOnly,
+		[parameter(ParameterSetName = "NoPipe")][parameter(ParameterSetName = "Pipe")]
 		[ValidateSet('Full', 'Log', 'Differential', 'Diff', 'Database')]
 		[string]$Type = "Database",
-		[parameter(ParameterSetName = "Pipe", Mandatory = $true, ValueFromPipeline = $true)]
-		[object[]]$DatabaseCollection
+		[parameter(ParameterSetName = "Pipe",Mandatory = $true, ValueFromPipeline = $true)]
+		[object[]]$Arrdatabases,
```
ctrlbold: Hey @Stuart-Moore - naming is hard and in PS v4, we could actually hide this from the end user, but unfortunately in v3 we can't. Once we get TabExtension++ setup, this will actually be $Database but for now, please keep it DatabaseCollection as DatabaseCollection  is more PowerShelly than Arrdatabases.

In addition, we're getting Pester failures for missing help. Please add help for CreateFolder and SqlInstance.

functions/Backup-DbaDatabase.ps1
```diff
@@ -71,27 +75,47 @@ Backs up AdventureWorks2014 to sql2016's C:\temp folder
 #>
 	[CmdletBinding(DefaultParameterSetName = "Default")]
 	param (
-		[parameter(ParameterSetName = "NoPipe", Mandatory = $true)]
+		[parameter(ParameterSetName = "NoPipe", Mandatory = $true)][parameter(ParameterSetName = "Pipe")]
```
ctrlbold: If someone is piping in full on SMO objects, then they'll have their connected server inside the object, so $server = $database.parent thus they won't need -SqlInstance. That's why $Sqlinstance is not in the Pipe ParameterSetName.

functions/Backup-DbaDatabase.ps1
```diff
@@ -71,27 +75,47 @@ Backs up AdventureWorks2014 to sql2016's C:\temp folder
...
+		$PipeDbs = @()
+		
+
+
 		if ($SqlInstance.length -ne 0)
 		{
-			$databases = $psboundparameters.Databases
+			if ($PSCmdlet.ParameterSetName -ne 'Pipe')
+			{	
+				Write-Verbose "$Functionname - Databases paramter used"
```
ctrlbold: typo here as well and hey! I didnt know you could get $pscmdlet.parametersetname

functions/Backup-DbaDatabase.ps1
Stuart-Moore: Are we always going to get SMO objects?

I've not pushed it yet, but have a version that could be called with:
'fgs','historyrest' | Out-File C:\dbatool\dbs.txt 
Get-Content C:\dbatool\dbs.txt | Backup-DbaDatabase -SqlInstance localhost\sqlexpress2016 -type full -verbose

TBH, wouldn't be an issue to pick the filtering based on the type of input object (like I'm doing with Restore-DbaDatabase)

functions/Backup-DbaDatabase.ps1
Stuart-Moore: I'll add that to the list of words I can't spell

functions/Backup-DbaDatabase.ps1
Stuart-Moore: Another blindspot

Yes, it's pretty handy

functions/Backup-DbaDatabase.ps1
Stuart-Moore: I'll fix the help

Yes, name could be better. I think I went with something different as the others get clobbered by the  dynamic parameters bit. but now that's escaped it might not be an issue

functions/Backup-DbaDatabase.ps1
ctrlbold: I'd pick full SMO objects over text if we had to do just one but if we'd support both, you can do a check on the object type and if it's string, ensure that -SqlInstance is specified.

Stuart-Moore:
This is really a rollback. Can't square generic object passing and the way that binding seems to be done for SMO objects. Not going to have time this weekend to spend on it, so this is at least a workable version with all the features in.

PR #766, opened 2017-02-17
New function Set-DbaQueryStoreConfig

Evdlaar:
Allows the configuration of the Query Store on one or multiple
databases.

### DIFF
--- a/functions/Set-DbaQueryStoreConfig.ps1
+++ b/functions/Set-DbaQueryStoreConfig.ps1
@@ -1,0 +1,177 @@
+Function Set-DbaQueryStoreConfig
+{
+<#
+.SYNOPSIS
+Configure Query Store settings for a specific or multiple databases.
+	
+.DESCRIPTION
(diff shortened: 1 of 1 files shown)


ctrlbold:
Looks great! Thank you :bowtie: 

Going to merge then do a little bit of cleanup from an old template that we have to update.

ctrlbold:
hey @Evdlaar, 
I merged it then fixed the encoding. Can you please resync dev to grab a fresh copy?

The command is awesome -- the only issue I see is that it requires every property (this may be an issue of SMO, if so, no prob). Would you be able to require at least one (I'd try `if (-not ($param1, $param2, $param3, etc))` and then when you're doing the sets, only set the ones that have been passed?

Evdlaar:
Yeah thought about that last night as well.
Will fix that though it needs some additional thought. In cases where QS is set to off I think all options need to be supplied. Will check!

PR #773, opened 2017-02-18
New Command - Invoke dbalogshippingrecovery

sanderstad:


functions/Invoke-DbaLogShippingRecovery.ps1
```diff
@@ -0,0 +1,307 @@
...
+#>
+    [CmdletBinding(DefaultParameterSetName = "Default", SupportsShouldProcess = $true)]
+    param
+	(
+        [Parameter(Position=1, Mandatory=$true)]
+		[Alias("ServerInstance", "SqlInstance")]
+        [string]$SqlServer,
+        [Parameter(Position=2, Mandatory=$false, ValueFromPipeline=$true)]
+        [string[]]$Database = $null,
+        [Parameter(Position=3, Mandatory=$false)][object]$SqlCredential,
```
FriedrichWeinmann: This parameter should be type `[System.Management.Automation.PSCredential]` instead of `[object]`
It is passed to connect-sqlserver which expect the more restrictive type.

functions/Invoke-DbaLogShippingRecovery.ps1
```diff
@@ -0,0 +1,307 @@
...
+        [Parameter(Position=4, Mandatory=$false)][switch]$NoRecovery
```
FriedrichWeinmann: Please add a switch parameter named 'Silent':
[switch]$Silent
(That'll be needed for other recommendations down the road)

functions/Invoke-DbaLogShippingRecovery.ps1
```diff
@@ -0,0 +1,307 @@
...
+	)
+
+    BEGIN
+    {
+        Write-Output "Attempting to connect to Sql Server.."
```
FriedrichWeinmann: Please replace all instances of Write-Output that you use to message the state/progress of your function with:
`Write-Message -Message "<Message Text here>" -Level 5 -Silent $Silent`
Choose an appropriate level:
2: Equivalent to Write-Host/Write-Information (By default, the users reads this)
5: Equivalent to Write-Verbose (By default, the user only reads this if specifying -Verbose as parameter)
8: Equivalent to Write-Debug (By default,, the user only sees this if specifying -Debug as parameter)
The smaller the value, the larger the probability it will appear on the screen of the user. It also ties in into the logging system (well ... as soon as I finish that later today).

functions/Invoke-DbaLogShippingRecovery.ps1
```diff
@@ -0,0 +1,307 @@
...
+
+        # Retrieve the log shipping information from the secondary instance
+        try{
+            Write-Verbose "Retrieving log shipping information from the secondary instance"
+            $logshipping_details = Invoke-Sqlcmd2 -ServerInstance $SqlServer -Database 'msdb' -Query $query
+        }
+        catch
+        {
+            Write-Error $_.Exception.Message
+            Break
```
FriedrichWeinmann: break is dangerous and should not be used to stop a function. Instead of these two lines, use this:
`Stop-Function -Message "<Message Text here>" -Silent $Silent -InnerErrorRecord $_ -Target $db`
`return`

functions/Invoke-DbaLogShippingRecovery.ps1
```diff
@@ -0,0 +1,307 @@
...
+                    # Check if the backup source directory can be reached
+                    if(Test-Path $ls.backup_source_directory)
+                    {
+                        # Get the last file from the backup source directory
+                        $latestfile = Get-ChildItem -Path $ls.backup_source_directory -filter ("*" + $ls.primary_database + "*") | where {($_.Extension -eq '.trn') } | Sort-Object LastWriteTime -Descending | Select-Object -First 1
+                    
+                        # Check if the latest file is also the latest copied file
+                        if($latestfile.Name -ne [string]$ls.last_copied_file.Split('\')[-1])
+                        {
+                            Write-Verbose "Synchronizing the latest transaction log backup file(s)"
```
FriedrichWeinmann: Please replace all instances of Write-Verbose with:
`Write-Message -Message "<Message Text here>" -Level 5 -Silent $Silent`

functions/Invoke-DbaLogShippingRecovery.ps1
```diff
@@ -0,0 +1,307 @@
...
+                if($server.Databases[$ls.secondary_database].Status -in ('Normal, Standby', 'Standby', 'Restoring'))
+                {
+
+                    Write-Output ("Started Recovery for '" + $ls.secondary_database + "'")
+
+                    # Check if the backup source directory can be reached
+                    if(Test-Path $ls.backup_source_directory)
+                    {
+                        # Get the last file from the backup source directory
+                        $latestfile = Get-ChildItem -Path $ls.backup_source_directory -filter ("*" + $ls.primary_database + "*") | where {($_.Extension -eq '.trn') } | Sort-Object LastWriteTime -Descending | Select-Object -First 1
```
FriedrichWeinmann: Avoid using Aliases, such as "where", instead of the full cmdlet "Where-Object".
If you use PowerShell Studio as an editor, you can expand aliases with the "Tab" key, dunno about the other editors.

functions/Invoke-DbaLogShippingRecovery.ps1
```diff
@@ -0,0 +1,307 @@
...
+                                Write-Verbose ("Waiting for files to get copied to '" + $ls.backup_destination_directory + "'")
+
+                            }
+
+                            # Check if the file has been copied
+                            $query = "SELECT last_copied_file FROM msdb.dbo.log_shipping_secondary WHERE primary_database = '" + $ls.primary_database + "'"
+                            $latestcopy = Invoke-Sqlcmd2 -ServerInstance $SqlServer -Database 'msdb' -Query $query
+
+                            # Waiting for the copy action to complete
+                            if($PSCmdlet.ShouldProcess($ls.backup_destination_directory, "Waiting for the copy action to complete.. "))
```
FriedrichWeinmann: Does waiting for an action to complete require a ShouldProcess check?
If the user needs to choose something here, maybe it should be a switch-parameter instead

functions/Invoke-DbaLogShippingRecovery.ps1
ctrlbold: oh man, @FriedrichWeinmann - i thought that we werent going to burden the end user with the silent switch. Can't we handle the silence through some other type of internal variable?

FriedrichWeinmann: Actually, it _needs_ to be on the front end - no way around that - and you don't need to worry about the effect on end users. Just make sure it's the last variable in the variable list (so it's last in the parameter tabbing cycle)) and all users that don't care about that are going to assume it's one of those myriad of automatic parameters (such as -InformationVariable).
Been using it for years and none of my coworkers using my toolkit ever even commented on it.

The thing about it is that it's an opt-in behavior so the default is off. Internal variables on that comes with a world of hurt through complexity and implementation. The thing is a function that the common user calls may have to use another public function which must be able throw errors so that the calling function realizes that. We can't globally toggle it off or on since some functions running under the hood may need silent mode when the user still shouldn't see a thing.

functions/Invoke-DbaLogShippingRecovery.ps1
ctrlbold: We are switching to the new Stop-Function @sanderstad, and yours is one of the first :)

@FriedrichWeinmann - shouldn't Stop-Function handle the return? If not, then it's not stopping?

FriedrichWeinmann: It can handle almost all things, but it cannot gracefully and the calling function - that's the limitation of the PowerShell engine's scope system, which means a Stop-Function that is used in a place where it needs to end the calling function cannot do this without throwing an exception (and we wanted to go easy on the red messages).
I tried a good long time getting it to work (something similar to `return return`). Nothing was possible except for using the truly horrid engine statements (such as break) which have seriously bad side effects.
So when we place Stop-Function in a location where we want to actually kill the calling function - as opposed to skip this item and process the next - Stop-Function handles the automatic integrations into anything we want to happen when a function breaks. And if we want to gracefully stop the function the return on the next line does it.

functions/Invoke-DbaLogShippingRecovery.ps1
sanderstad: I did that in the first place but I looked at some other functions and they used the object one so I changed it again. Shouldn't have done that

functions/Invoke-DbaLogShippingRecovery.ps1
sanderstad: I didn't know about that function. No problem I'll change those

functions/Invoke-DbaLogShippingRecovery.ps1
sanderstad: Changing it

functions/Invoke-DbaLogShippingRecovery.ps1
sanderstad: Like the write-output, changing it so the Write-Message

functions/Invoke-DbaLogShippingRecovery.ps1
sanderstad: The reason I had to put the where within the shouldprocess check is that when you execute it with -whatif the function will halt. The reason for that is, is because the I use the log shipping jobs to handle the last actions of the log shipping. 
With the whatif parameter those jobs are not executed and the msdb database is never updated with the new values which results in an infinite loop.
It was the prettiest solution I had for now

FriedrichWeinmann: Hm, just a thought - not at all sure whether it's the right solution - but ... wouldn't it make sense to place the entire, or almost entire process block into a single big shouldprocess conditional? Maybe add more information on individual steps of the process to the help?

(Slight background info: I'm quite deep into PowerShell itself, but I'm a lot shallower as a dba. From a PS perspective, the above proposed solution would be correct, but I can't really know what a fulltime dba - the main audience for this module - would expect out of function behavior)

sanderstad: I can probably put more actions into one shouldprocess check. I'm going to look into that.

Normally log shipping is set up using sql server agent jobs. The jobs have either backup, copy or restore action. The copy and restore jobs are on the secondary instance. Those jobs use the sqllogship.exe which takes care of a lot of checks and internals.
Te reason I use the jobs to execute the copy and restore is because I'd have to make lots of extra code. One of the steps I take is disabling the copy and restore jobs after they've been executed which removes the possibility of errors.
If I'd have to do a disaster recovery with log shipping I'd do the same kind of actions but manually. In my production environment I have over 25 databases being log shipped which would probably take quite some time to restore.

functions/Invoke-DbaLogShippingRecovery.ps1
FriedrichWeinmann: Don't you worry, it's still sparkling new and nothing you must have known beforehand.
What Write-Message and Stop-Function do, is centralize our message and control flow.
Meaning, we have a central location that can ...
 - Determine how the user is informed about things
 - Handle logging / debugging information
This has the benefit, that we can place the choice easily into the hands of the user (using the no less new configuration system - think of that as a kind of environment-variables for dbatools in more shiny) on just how the system handles all that, without having to actually implement the consequence of that into each function individually.

For example, a user can configure, what level of message he wants written directly to screen, which he wants in Verbose and which in Warning. In case you wonder, the defaults are:
Write-Host/Write-Information: 1-3 (Automatically switches to Write-Information on v5)
Write-Verbose: 4-6
Write-Debug: 1-9 (So if you want to only write Debug, use something between 7-9)
The level system gives us the ability to control verbosity in an bit more granular fashion. That said, the level system does not affect _warnings_, which will always be shown unless the function is explicitly silenced.

functions/Invoke-DbaLogShippingRecovery.ps1
wsmelton: No worries once implement standards guide it will be the same on all functions...

functions/Invoke-DbaLogShippingRecovery.ps1
wsmelton: If this is going to be used and replace the normal, then a wiki page needs to be created with all of the explanation and examples. I see this having to be repeated so might as well get a page for it now.

functions/Invoke-DbaLogShippingRecovery.ps1
wsmelton: @sanderstad if you run your code against PSScriptAnalyzer it will help pick up these type of things. It is a module from PSGallery that you just run `Invoke-ScriptAnalyzer myfile.ps1`

functions/Invoke-DbaLogShippingRecovery.ps1
FriedrichWeinmann: Yupp. And once I had a final conversation on some implementation specifics with Chrissy later tonight, that is just what I'm going to write (And if that discussion causes changes I'll be the one who writes the changes for that on all functions that already use it).
Gonna make more than single page for it, too :)

functions/Invoke-DbaLogShippingRecovery.ps1
```diff
@@ -0,0 +1,305 @@
...
+
+        # Retrieve the log shipping information from the secondary instance
+        try{
+            Write-Message -Message "Retrieving log shipping information from the secondary instance" -Level 5 -Silent $Silent
+            $logshipping_details = Invoke-Sqlcmd2 -ServerInstance $SqlServer -Database 'msdb' -Query $query 
+        }
+        catch
+        {
+            Stop-Function -Message "Error retrieving the log shipping details" -Silent $Silent -InnerErrorRecord $_ -Target $SqlServer
+            Write-Error $_.Exception.Message
```
FriedrichWeinmann: You don't need to use Write-Error anymore: Stop-Function does that for you.
Our default is to not show red exceptions in dbatools. There will be an opt-in into red exceptions (that's where _$Silent_ comes in). Feel free to add the exception message to the message text, if you want to pass that to the user.

sanderstad:
Removed the Write-Error

functions/Invoke-DbaLogShippingRecovery.ps1
```diff
@@ -0,0 +1,390 @@
...
+        catch
+        {
+            Stop-Function -Message ("Error retrieving the log shipping details: " + $_.Exception.Message) -Silent $Silent -InnerErrorRecord $_ -Target $SqlServer
+            
+            return
+        }
+        
+        # Checking the status of the SQL Server Agent service
+        Write-Message -Message ("Checking the status of the SQL Server Agent" + $ls.secondary_database + "'") -Level 2 -Silent $Silent
+        $agentservice = Get-Service | Where-Object -like -value '*sql*agent*' -Property 'name'
```
ctrlbold: Hey Sander,
This line will only account for 

- services on localhost
- any service named sqlagent

It does not handle: remote servers or multiple instances.

ctrlbold:
hey @sanderstad it added a bunch of files. I think it's because I committed straight to master and you were using master as a base instead of development. I'm gonna try to uncommit then add your changes back.

ctrlbold:
too many files - moved to https://github.com/sqlcollaborative/dbatools/pull/907/files

PR #776, opened 2017-02-19
Disabled the need to set all Query Store configuration items when executing the function

Evdlaar:
In this version not all Query Store configuration parameters need to be
supplied. The function will now only modify the configuration for
parameters supplied.


ctrlbold:
hey @Evdlaar  - this PR is causing a conflict so I'm going to take the resulting code, paste it in and submit it to development. I'm assuming it was created from an old branch -- please sync sqlcollaborative\development (sorry my screenshot is a bit off, I'm connected directly to the repo, pretend it says sqlcollaborative\development like the first screenshot)

image


image


image

PR #777, opened 2017-02-19
New function Copy-DbaQueryStoreConfig

Evdlaar:
Copies the configuration of a Query Store configured database to other
databases on the same or different servers.


ctrlbold:
Gonna change a couple things that should have been in our template then we will be set for including all 3 in the next release. Awesome job!! 🌮!

PR #778, opened 2017-02-20
Querystore commands updates

ctrlbold:
@Evdlaar please check out the code and approve if it functions as you intended. I added support for WhatIf and updated some variable names.

### DIFF
--- a/functions/Copy-DbaQueryStoreConfig.ps1
+++ b/functions/Copy-DbaQueryStoreConfig.ps1
@@ -16,8 +16,8 @@ The database from which you want to copy the Query Store configuration.
 .PARAMETER SourceSqlCredential
 Credential object used to connect to the source SQL Server as a different user.
 
-.PARAMETER SourceSqlCredential
-Credential object used to connect to the source SQL Server as a different user.
+.PARAMETER DestinationSqlCredential
+Credential object used to connect to the destination SQL Server as a different user.
 
 .PARAMETER Destination
 The target server where the databases reside on which you want to enfore the copied Query Store configuration from the SourceDatabase.
@@ -25,6 +25,9 @@ The target server where the databases reside on which you want to enfore the cop
 .PARAMETER DestinationDatabase
 The databases that will recieve a copy of the Query Store configuration of the SourceDatabase.
 
+.PARAMETER AllDatabases
+Run command against all user databases	
+	
 .PARAMETER Exclude
 Copy Query Store configuration for all but these specific databases.
 
--- a/functions/Copy-DbaQueryStoreConfig.ps1
+++ b/functions/Copy-DbaQueryStoreConfig.ps1
@@ -28,6 +28,16 @@ The databases that will recieve a copy of the Query Store configuration of the S
 .PARAMETER Exclude
 Copy Query Store configuration for all but these specific databases.
 
+.PARAMETER WhatIf
+Shows what would happen if the command were to run
+	
+.PARAMETER Confirm
+Prompts for confirmation of every step. For example:
+
+Are you sure you want to perform this action?
+Performing the operation "Changing Desired State" on target "pubs on SQL2016\VNEXT".
+[Y] Yes  [A] Yes to All  [N] No  [L] No to All  [S] Suspend  [?] Help (default is "Y"):
+
 .NOTES
 Author: Enrico van de Laar ( @evdlaar )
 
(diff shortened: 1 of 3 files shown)


functions/Copy-DbaQueryStoreConfig.ps1
```diff
@@ -38,120 +54,116 @@ You should have received a copy of the GNU General Public License along with thi
...
-			}
-
-        # Grab the Query Store configuration from the SourceDatabase through the Get-DbaQueryStoreConfig function
-        $SourceQSConfig = Get-DbaQueryStoreConfig -SqlServer $SourceSqlServer -Databases $SourceDatabase
-
+	
+		Write-Verbose "Connecting to source: $Source"
+		try
+		{
+			$sourceserver = Connect-SqlServer -SqlServer $Source -SqlCredential $SourceSqlCredential
```
PowerDBAKlaas: $Source can contain multiple objects, yet here's a connection made to $source in BEGIN{}.
Probably $Source should accept only one SqlInstance?

functions/Copy-DbaQueryStoreConfig.ps1
```diff
@@ -38,120 +54,116 @@ You should have received a copy of the GNU General Public License along with thi
...
+			
+		}
+		catch
+		{
+			Write-Warning "Can't connect to $Source."
+			continue
```
PowerDBAKlaas: What will "continue" do when we're not in a loop?

functions/Copy-DbaQueryStoreConfig.ps1
```diff
@@ -38,120 +54,116 @@ You should have received a copy of the GNU General Public License along with thi
...
-			$targetserver = Connect-SqlServer -SqlServer $TargetSqlServer -SqlCredential $Credential
-
+			Write-Warning "You must specify databases to execute against using either -DestinationDatabase, -Exclude or -AllDatabases"
+			continue
+		}
+		
+		Write-Verbose "Connecting to destination: $Destination"
+		try
+		{
+			$destserver = Connect-SqlServer -SqlServer $Destination -SqlCredential $Credential
```
PowerDBAKlaas: $Destination can accept multiple objects.
Probably it needs to be altered in the param definition?
If not, there needs to be a foreach to connect to multiple destinations.

functions/Copy-DbaQueryStoreConfig.ps1
```diff
@@ -38,120 +54,116 @@ You should have received a copy of the GNU General Public License along with thi
...
-            
-            # Set the Query Store configuration through the Set-DbaQueryStoreConfig function
-            Set-DbaQueryStoreConfig -SqlServer $TargetSqlServer -Databases $($db.name) -State $SourceQSConfig.ActualState -FlushInterval $SourceQSConfig.FlushInterval -CollectionInterval $SourceQSConfig.CollectionInterval -MaxSize $SourceQSConfig.MaxSize -CaptureMode $SourceQSConfig.CaptureMode -CleanupMode $SourceQSConfig.CleanupMode -StaleQueryThreshold $SourceQSConfig.StaleQueryThreshold
-                      
-		}  
-
-
-    }
-	
+
```
PowerDBAKlaas: I see "SupportsShouldProcess" is added, but I can't find the implementation of -Whatif logic.

functions/Get-DbaQueryStoreConfig.ps1
```diff
@@ -7,11 +7,11 @@ Get the Query Store configuration for Query Store enabled databases.
...
-.PARAMETER SqlServer
+.PARAMETER SqlInstance
 The SQL Server that you're connecting to.
 
-.PARAMETER Credential
-Credential object used to connect to the SQL Server as a different user.
+.PARAMETER SqlCredential
+SqlCredential object used to connect to the SQL Server as a different user.
 
 .PARAMETER Databases
```
PowerDBAKlaas: parameter should be singular: Database

functions/Copy-DbaQueryStoreConfig.ps1
Evdlaar: Correct. The way it is implemented means that the configuration of only one databases gets pushed to (multiple) target database(s).

functions/Copy-DbaQueryStoreConfig.ps1
Evdlaar: Fixed in next update

functions/Copy-DbaQueryStoreConfig.ps1
Evdlaar: Fixed in next update

functions/Copy-DbaQueryStoreConfig.ps1
Evdlaar: I believe @ctrlbold changed this so it matches a new template?
-WhatIf logic is not yet implemented.

ctrlbold: No, no it is. The Set at the bottom handles all of it.

PowerDBAKlaas: @ctrlbold I thought it might,
couldn't the version checks be left to Set- too then?

ctrlbold: They could, but i think it's better to catch them earlier.

Evdlaar: I did remove them in next version though since they are handled by both the Get and the Set command seperatly

PR #788, opened 2017-02-21
Fix last backup order

ClaudioESSilva:
- If the most recent backup is done to an existing and older backup
media set id, when runing the function the output shows not the most
recent regarding date time but the most recent regarding the higher
backupset.media_set_id.

I will add more info into PR

### DIFF
--- a/functions/Get-DbaBackupHistory.ps1
+++ b/functions/Get-DbaBackupHistory.ps1
@@ -188,7 +188,7 @@ Lots of detailed information for all databases on sqlserver2014a and sql2016.
 								  a.MediaSetId,
 								  a.Software
 								FROM (SELECT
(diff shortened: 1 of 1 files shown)


ClaudioESSilva:
@Stuart-Moore correct me if I'm wrong 😃 

Example:
image

Using query on SSMS
image

Stuart-Moore:
Yes, always trust the backup date, not the file date, mediaset number or anything else.

Even the date lies sometimes, the only true ordering is the LSN.

ctrlbold:
Good to go, y'all? This is over my 🍔

ctrlbold:
Oh, @Stuart-Moore said yes, so yes :D merged! 👍

PR #790, opened 2017-02-21
Get dbasecurityaudit new

gbargsley:


functions/Get-DbaSecurityAudit.ps1
```diff
@@ -0,0 +1,82 @@
...
+				continue
+			}
+
+            if ($server.versionMajor -lt 10)
+            {
+                Write-Warning "Server Audits are only supported in SQL Server 2008 and above. Quitting."
+                return
+            }
+			
+            $serveraudits = $server.Audits
```
ctrlbold: We want to keep these audits as their full on SMO objects the way we did with Get-DbaDatabase (this is only true for some things - which ones? I'd say the ones that aren't calculated or rich SMO objects.)

So what you can do is this 

```
$server.Audits | Select-DefaultView Name, Enabled, FilePath, FileName
```

We still need to get the server name in there, however. I'll be back with the code in 5.

ctrlbold:
Perfect!

PR #792, opened 2017-02-21
Tested a number of parameters and fixed some bugs

cvdiermen:
Line 120 & 203 converted $LastUsed to boolean value and used ToString()
for accepting 0 as a valid value. Otherwise $LastUsed value 0 would
result in boolean value False.

Line 205 Replace variable $Since with variable $Daysback, because $Since
is used for another purpose further down.

Line 207 Used positive value  $LastUsed in message instead of negative
value $Since

Line 233 Fixed incorrect message


ctrlbold:
Thanks so much, @cvdiermen! 🌮

PR #811, opened 2017-02-23
Backup-Dbadatabase - added skip for snapshots

ctrlbold:


ctrlbold:
Thank you, @SQLDBAWithABeard ! :D

PR #814, opened 2017-02-23
Initial commit of Get-DbaLocaleSetting

PowerDBAKlaas:
Gets the locale settings on one or more computers.

How to test this code: 
- [ ] 
- [ ] 

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally


PowerDBAKlaas:
Probably needs a Selecct-Defaultview to make something user friendly.
Any suggestions?
@ctrlbold @wsmelton @niphlod @ClaudioESSilva @Stuart-Moore @ConstantineK

ctrlbold:
@PowerDBAKlaas  playing now

ctrlbold:
Very useful stuff! 

But yes, it needs some column name love. You can either create an object or use a lot of "as"es in the Select-DefaultView (that adds aliases and hides the original, unsure which you want).

Would you want it to have no "s" or "i" unless there are overlapping things then you can do BlahString or BlahNumeric ?

PowerDBAKlaas:
Well there are methods to set the values via CIM too, so it may be useful to have the real object at some point. For that reason I would choose Select-Defaultview with indeed a lot of " as ".

But I'd like to have an idea on which columns the team considers the most useful.
BTW I just discovered I've got one computer that has an aberrant firstweekoftheyear, so I'd put that definitely in 😄

ctrlbold:
I like them all, honestly 👍

PowerDBAKlaas:
There may be inconsistent output:
I take all values in the subkey, so there's a risk older OS or so may have a different collection.

ctrlbold:
i have some win2003's, win2000 and xps if you wanna try.

ctrlbold:
Where are we at with this, @PowerDBAKlaas ?

PowerDBAKlaas:
@ctrlbold  It works.
Maybe we can include it and prettify later?

ctrlbold:
k!

ctrlbold:
later on, what you can do is an array of pretty column names, and then an array of replacement column names. and if it matches, then it turns pretty, if it doesnt, then it goes thru with the ugz i and s

PowerDBAKlaas:
OK,
only ...
with the what?

Issue #816, opened 2017-02-23
Get-DbaSqlBuildReference, new command


niphlod:
This is going to be the base which other commands, that will follow closely, will build.
The idea is knowing from the release number when it was released, at which SP level and CU level it is, etc, kinda what dba are used to look for on sqlserverupdates.com or sqlserverbuilds.blogspot.com or buildnumbers.wordpress.com. 
Plus, it'll hold infos about MS Lifecycle Management.
It'll try to stay updated fetching from dbatool's own maintained datasource (i.e. a json file) OR warn about how much "stale" the reported data can be.

`Test-DbaSqlBuild -SqlServer foo -MinimumVersion bar` and/or `Test-DbaSqlBuild -SqlServer foo -MaxReleasesBehind 2`
are just some of the examples of useful functions that will benefit from Get-DbaSqlBuildReference

niphlod:
this is linked to https://github.com/sqlcollaborative/dbatools/pull/840

PR #840, opened 2017-02-27
Get-DbaSqlBuildReference - from @niphlod 

ctrlbold:
aweeeeeeeeesome!


niphlod:
linked to https://github.com/sqlcollaborative/dbatools/issues/816

niphlod:
- [ ] need to put a global "updatedon" prop on the json
- [ ] need to return a global "updatedon" prop on the returned resultset
- [ ] need to warn about "updatedon" being too far in the past
- [ ] need to download from a central location (sqlcollaborative.github.io gh-pages ?) the freshest buildlist
- [ ] need to adhere on names before concocting higher-level cmdlets such as `Test-DbaSqlBuild`
- [ ] need to adhere on naming conventions
- [ ] need to figure out if the actual structure is easy-on-the-eyes and easy-to-parse (i.e. maybe all props should always be a list rather than a string if they contain one element only OR maybe they should always be a string (joined/splitted by `,`))

gbargsley:
I just ran it across my environment and looks great.  I notice it does not list the actual product version.  That might be helpful.  Also, what is MatchType and Warning represent?  And on my 2012 instances I just applied SP3 CU7 last Friday and that is the latest that I am aware of.  The SP3 shows latest, but CULevel shows CU7, but does not say latest.

gbargsley:
It also looks at shows that if only a SP is applied and no SP/CU has been applied then it still shows latest CU from previous SP.

For instance our production 2014 instances are SP2 12.0.5000 but the function reports 2014, SP2 LATEST and CU11.

niphlod:
@gbargsley : we can output the actual version but I thought it's kinda the deal... you have everywhere a version string and with this cmdlet you'd like another representation for it. 
Note that the original SMO is on the "instance" prop that is hidden, so another way to fetch it could be `select { $_.Instance.Version }`

MatchType refers to the fact that the exact version string has been found vs the nearest. 
That's another point on my todolist i.e. 
- we have a 11.0.1234 and 11.0.1240 on the index 
- the returned version from the server is 11.0.1237
In that case I'd like to return infos for 11.0.1234 (the nearest on the lower end) but Matchtype would be "Approx" rather than "Exact".

Warning ATM gets filled when we have no clue about it (not even a "major" match)

niphlod:
> It also looks at shows that if only a SP is applied and no SP/CU has been applied then it still shows latest CU from previous SP.
You're right . Added to the list of fixes.

niphlod:
everyone: should we have a "latest" for CUs too ?

niphlod:
6 points on the todolist have been enhanced/fixed. There's also a -Build now that takes a simple string and a `Version` attribute that holds the input value that is used as the base value to make comparisons against.

ctrlbold:
Hey @niphlod - would you like me to wait until you're done with the rest of the todos before merging? The PR came in before March 1, so nbd, you're in. Why not host file in /bin (and even https github dbatools bin)? too large or too ridiculous to put on everyone's machine when we just need it for whatever?

niphlod:
the logic part needs testing. the final release needs to take care of the possibility of new releases coming in . If we are releasing - for the first period - at a faster rate than SQL versions, we can live shipping the json with the module. 
After we can agree on a "scheme" for this and maybe other commands that are dependant on "to be refreshed once in a while" resources.

functions/Get-DbaSqlBuildReference.ps1
```diff
@@ -0,0 +1,210 @@
...
+The SQL Server that you're connecting to.
+
+.PARAMETER Credential
+Credential object used to connect to the SQL Server as a different user.
+
+.PARAMETER Build
+Instead of connecting to a real instance, pass a string identifying the build to get the info back.
+
+.PARAMETER Silent
+Replaces user friendly yellow warnings with bloody red exceptions of doom!
```
PowerDBAKlaas: Too much poetry. 😄  the warning and error colors are not always yellow and red.

functions/Get-DbaSqlBuildReference.ps1
```diff
@@ -0,0 +1,210 @@
...
+$cred = Get-Credential sqladmin
+Get-DbaSqlBuildReference -SqlInstance sqlserver2014a -SqlCredential $cred
+
+Does the same as above but logs in as SQL user "sqladmin"
+
+.EXAMPLE
+Get-DbaSqlBuildReference -Build "12.00.4502"
+
+Returns information about a build identified by  "12.00.4502" (which is SQL 2014 with SP1 and CU11)
+
```
PowerDBAKlaas: Please include examples:
- piping sqlinstance(s) to the function
- submitting multiple sqlinstances

functions/Get-DbaSqlBuildReference.ps1
```diff
@@ -0,0 +1,210 @@
...
+					ComputerName  = $server.NetName
+					NameLevel = $Detected.Name
+					SPLevel = $Detected.SP
+					CULevel = $Detected.CU
+					KBLevel = $Detected.KB
+					SupportedUntil = $Detected.SupportedUntil
+					MatchType = $MatchType
+					Warning = $Detected.Warning
+					Version = $LookFor
+					Instance = $server
```
PowerDBAKlaas: Do we need the complete server object as [Instance] when there's also [SqlInstance] and [InstanceName]?

functions/Get-DbaSqlBuildReference.ps1
```diff
@@ -0,0 +1,210 @@
...
+			}
+			Select-DefaultView -InputObject $object -Property SqlInstance, Version, NameLevel, SPLevel, CULevel, KBLevel, SupportedUntil, MatchType, Warning
+		}
+		foreach($buildstr in $Build) {
```
PowerDBAKlaas: So here are different types of object mingled.
That causes difficulties when redirecting output.

functions/Get-DbaSqlBuildReference.ps1
niphlod: straight copy-pasted, sorry. will try to come up with something better but, really, this would need to be documented and then copied over any "silent" parameter spec in cmdlets using the new logging infrastructure

functions/Get-DbaSqlBuildReference.ps1
niphlod: if in the need of piping down instance, that's the only way, and in fact that is hidden "as usual"

functions/Get-DbaSqlBuildReference.ps1
niphlod: Know that but asked specifically in the channel what to do (e.g. different cmdlets) but when accepting a simple string .... returning a made-up instance seemed unproper.

PowerDBAKlaas: oh yes, that is improper.

It's clearly two different things, the server properties vs the references. Two functions is the only proper way I see.

niphlod: Happy either way ... let's have an "internal" with the "core logic" and two main "functions", one for instances and the other for strings ? @ctrlbold  : comments ?

PowerDBAKlaas: The output of both can still be objects, no?

niphlod: yeah. 
Technically we could also have just a cmdlet that works on strings and show people how easy it is to pipe connect-dbasqlserver to it.

ctrlbold: what klaas said! ;D

ctrlbold: Love the idea to pipe to it!

ctrlbold:
Hey everyone involved, can we get this merged by Monday?

niphlod:
I'll take a stab at it tomorrow and see how it'll become. At this point is just "modelling after expected usage" rather than "the code doesn't work the way it's supposed to be".

niphlod:
it now works just against strings, with a provided example to make it work against all registered servers. This can be a good pre-1.0, where we'll eventually add the "refresh index from a configured location".
@PowerDBAKlaas , @ctrlbold : is the current "format" acceptable ?

functions/Get-DbaSqlBuildReference.ps1
```diff
@@ -0,0 +1,166 @@
...
+		[string[]]$Build,
+		[switch]$Silent
+	)
+
+	BEGIN {
+		function Find-DbaSqlBuildReferenceIndex {
+			[CmdletBinding()]
+			Param()
+			$orig_idxfile = "$moduledirectory\bin\dbatools-buildref-index.json"
+			$DbatoolsData = Get-DbaConfigValue -Name 'Path.DbatoolsData'
```
ctrlbold: Yessssss!

ctrlbold:
Thanks, all! 💪

PR #841, opened 2017-02-27
 New-DbaSsisCatalog

Staggerlee011:
Creates the SSIS Catalog

### DIFF
--- a/functions/New-DbaSsisCatalog.ps1
+++ b/functions/New-DbaSsisCatalog.ps1
@@ -1,0 +1,117 @@
+﻿Function New-DbaSsisCatalog {
+<#
+.SYNOPSIS 
+Enables the SSIS Catalog on a SQL Server 2012 +
+
+.DESCRIPTION
+After installing the SQL Server Engine and SSIS you still have to enable the SSIS Catalog. This function will enable the catalog and gives the option of supplying the password. 
+
+.PARAMETER SqlServer
+SQL Server you wish to run the function on.
+
+.PARAMETER SqlCredential
+Credenitals used to connect to the SQL Server
+
+.PARAMETER SsisCredential
+Optional Password that will be used for the security key in SSISDB
+
+.NOTES 
+Author: Chrissy LeMaire (@cl), netnerds.net
+Requires: sysadmin access on SQL Servers
+
+dbatools PowerShell module (https://dbatools.io, clemaire@gmail.com)
+Copyright (C) 2016 Chrissy LeMaire
+
+This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+
+This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+.LINK
+https://dbatools.io/New-DbaSsisCatalog
+
+.EXAMPLE   
+New-DbaSsisCatalog -SqlServer DEV01 -SsisCredential MySuperPassword
+
+Creates the SSIS Catalog on server DEV01
+
+#>
+	[CmdletBinding(SupportsShouldProcess = $true)]
+	Param (
+		[parameter(Mandatory = $true, ValueFromPipeline = $true)]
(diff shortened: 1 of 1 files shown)


functions/New-DbaSsisCatalog.ps1
```diff
@@ -0,0 +1,117 @@
...
+SQL Server you wish to run the function on.
+
+.PARAMETER SqlCredential
+Credenitals used to connect to the SQL Server
+
+.PARAMETER SsisCredential
+Optional Password that will be used for the security key in SSISDB
+
+.NOTES 
+Author: Chrissy LeMaire (@cl), netnerds.net
```
wsmelton: Need to add your name here :)

functions/New-DbaSsisCatalog.ps1
```diff
@@ -0,0 +1,117 @@
...
+Creates the SSIS Catalog on server DEV01
+
+#>
+	[CmdletBinding(SupportsShouldProcess = $true)]
+	Param (
+		[parameter(Mandatory = $true, ValueFromPipeline = $true)]
+		[Alias("ServerInstance", "SqlInstance")]
+		[string[]]$SqlServer,
+        [System.Management.Automation.PSCredential]$SqlCredential,
+        [string]$SsisPassword
```
wsmelton: Oh, never plain text 🦈 Make this a secure string, or can use the credential object to make it easier for folks.

functions/New-DbaSsisCatalog.ps1
```diff
@@ -0,0 +1,117 @@
...
+
+            #if SQL 2012 or higher only validate databases with ContainmentType = NONE
+            if (!($server.versionMajor -gt 10))
+		    {
+                write-Warning "This version of SQL Server cannot have SSIS catalog"
+                break
+            }
+            ## need to check if ssis installed
+            if ($true)
+            {
```
wsmelton: Maybe I'm not reading it write but isn't something missing here?

functions/New-DbaSsisCatalog.ps1
```diff
@@ -0,0 +1,117 @@
...
+                Write-Warning "You do not have SSIS installed on this instance"
+            }
+
+            ## check if SSIS and Engine running on box
+            $services = Get-DbaSqlService -ComputerName $server.NetName 
+            $ssis_scv = $Services | Where {$_.ServiceType -eq "SSIS" -and $_.State -eq "Running"}
```
wsmelton: Need to fix case of `$Servers`, should be `$services`.

functions/New-DbaSsisCatalog.ps1
```diff
@@ -0,0 +1,117 @@
...
+                {
+                    if (!($SsisCredential))
+                    {
+                        $SSISDB = New-Object Microsoft.SqlServer.Management.IntegrationServices.Catalog ($SSISServer, "SSISDB")           
+                        $SSISDB.Create()  
+                    }
+                    else
+                    {
+                        $SSISDB = New-Object Microsoft.SqlServer.Management.IntegrationServices.Catalog ($SSISServer, "SSISDB", $SsisPassword)           
+                        $SSISDB.Create()
```
wsmelton: Just a thought, but it is recommended practice to backup the master key after you create the SSIS catalog. Might add in a warning? I don't think we can validate if the master key has ever been backed up or not (not sure there is a history of it).

functions/New-DbaSsisCatalog.ps1
```diff
@@ -0,0 +1,117 @@
...
+            }
+
+            ## check if SSIS and Engine running on box
+            $services = Get-DbaSqlService -ComputerName $server.NetName 
+            $ssis_scv = $Services | Where {$_.ServiceType -eq "SSIS" -and $_.State -eq "Running"} 
+            
+            if (-not $ssis_scv)
+            {
+                write-warning "SSIS is not running on the server you supplied"
+                break
```
wsmelton: This should be continue so it does not cause the function to stop...which think we have some new internal functions now instead of using break and such.

functions/New-DbaSsisCatalog.ps1
ctrlbold: @Staggerlee011 :O ! we had a week long convo about this, what happened?

functions/New-DbaSsisCatalog.ps1
Staggerlee011: agree, i have the basics of a backup cert function, just need to check a few things. ill add a warning for here (when the backup cert is done maybe put that in here.)

ctrlbold:
fixed and merging 👍

PR #842, opened 2017-02-27
 Get-DbaDatabaseEncryption

Staggerlee011:
Get-DbaDatabaseEncryption - lists al encryption used in databases on
server


ctrlbold:
hey @PowerDBAKlaas  - do you think this should be named Get-DbaEncryptionInfo? And if so, would it need more stuff?

PowerDBAKlaas:
I think the name is correct; since it is specific about the database level it should contain "...Database...".
Maybe functions concerning AlwaysEncrypted, disk encryption or what MS might add in future versions, will be added later.

I would, however, try to return the original object with some add-member. Question is if we can return those 4 objects in one function then. Can you examine that, @Staggerlee011 ?

PR #848, opened 2017-02-27
New Test-DbaIdentityUsage

scinobe:
This module can execute against a server, set of servers, database or
set of databases and return the IDENTITY seed columns, their current
value, the max allowable value, and the % usage of that range.

### DIFF
--- a/functions/Test-DbaIdentityUsage.ps1
+++ b/functions/Test-DbaIdentityUsage.ps1
@@ -1,0 +1,182 @@
+﻿Function Test-DbaIdentityUsage
+{
+<# 
+.SYNOPSIS 
+Displays information relating to IDENTITY seed usage.  Works on SQL Server 2008-2016.
+
+.DESCRIPTION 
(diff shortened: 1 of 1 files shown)


functions/Test-DbaIdentityUsage.ps1
```diff
@@ -0,0 +1,182 @@
...
+
+        $databases = $psboundparameters.Databases
+		$exclude = $psboundparameters.Exclude
+        
+        $threshold = 0
+        $threshold = $psboundparameters.Threshold
+
+        $collection = New-Object System.Collections.ArrayList
+
+        $sql = "Select DB_NAME(), Object_Name(id.object_id) As [table_name]
```
ClaudioESSilva: This query does not take into account columns that starts with negative values.
I will give you a query that I use (on slack)

Output example:
image

ctrlbold:
Nice work! Love this command - total spirit of PS + SQL! 🥇

PR #852, opened 2017-02-27
Connect-SqlServer - Bugfix Credentials not used for parameter connections

FriedrichWeinmann:
Fixed an issue where Connect-SqlServer wouldn't use credentials when used in parameter mode


ctrlbold:
Thank you!

PR #863, opened 2017-02-28
Create PSScriptAnalzyer rules file

wsmelton:
Rules file that will be used with contributors wanting to use VS Code mostly, but can be used for those just wanting to run `Invoke-ScriptAnalzyer` against their PS1.

 - Just adding a rules file to the bin directory

How to test this code: 

```powershell
Invoke-ScriptAnalzyer -Path C:\GitHub\dbatools\functions\Backup-DbaDatabase.ps1 -Settings C:\GitHub\dbatools\bin\PSScritpAnalzyerRules.psd1
```

### DIFF
--- a/bin/PSScriptAnalyzerRules.psd1
+++ b/bin/PSScriptAnalyzerRules.psd1
@@ -1,0 +1,24 @@
+@{
+    IncludeRules = @(
+                    'PSAvoidUsingCmdletAliases',
+                    'PSAvoidUsingWriteHost',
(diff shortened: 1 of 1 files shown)


wsmelton:
I can't see what rules Appveyor is set to use, it is meant to copy that so the rules file as-is might contain more than what we want to check. We can adjust as needed.

ctrlbold:
😍  AppVeyor juust hits up the functions/ dir

PR #865, opened 2017-02-28
Improve test sqltempdbconfiguration

ClaudioESSilva:
- Add support for multiple instances
- if server is sql 2000 will be skipped. Before it was giving error executing query.

### DIFF
--- a/functions/Test-SqlTempDBConfiguration.ps1
+++ b/functions/Test-SqlTempDBConfiguration.ps1
@@ -16,7 +16,7 @@ Other rules can be added at a future date. Only results that don't match best pr
 use the -Detailed switch.
 
 .PARAMETER SqlServer
-The SQL Server instance.You must have sysadmin access and server version must be SQL Server version 2000 or higher.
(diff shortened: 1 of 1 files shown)


functions/Test-SqlTempDBConfiguration.ps1
```diff
@@ -60,178 +60,210 @@ Checks tempdb on the localhost machine. All rest results are shown.
 	Param (
 		[parameter(Mandatory = $true)]
 		[Alias("ServerInstance", "SqlInstance")]
-		[object]$SqlServer,
+		[string[]]$SqlServer,
```
ctrlbold: plz do [object[]] so we can send multiple string or SMO object

ClaudioESSilva: Thanks @ctrlbold.
Done. Also changed to PSCredential instead of object

ctrlbold:
works! thanks claudio! 💃

PR #866, opened 2017-02-28
Improve test dbajobowner

ClaudioESSilva:
Login's id property is empty on sql server 2000 so the validation need a little fallback to set the default to 'sa' as the help says


ctrlbold:
works! 🍔

PR #867, opened 2017-02-28
Fix test dbaservername

ClaudioESSilva:
Output was not the same when -Detailed switch was used


ctrlbold:
hey @ClaudioESSilva - I tested it and it worked, but then I updated it a bit. Please run

`Get-SqlRegisteredServerName -SqlServer sql2014| Test-DbaServerName  | Where RenameRequired -eq $true`

You should see two servers that need to be renamed. Then

`Get-SqlRegisteredServerName -SqlServer sql2014| Repair-DbaServerName`

But do not rename! Just say no when prompted.

ClaudioESSilva:
Hey @ctrlbold I confirm.
I saw what you said.

but accidently I hit the... I'm kidding 😄

ctrlbold:
lol good ;)

Issue #552, opened 2017-01-18
Find-DbaLongRunningQuery


SirCaptainMitch:
Is this a *feature* OR *bug*:
Enhancement 

https://trello.com/c/kK7JNqMG

**Additional comments**
@wsmelton 
"Purpose would not be to find blocking tree or anything...that you can get by calling sp_whoisactive with the right parameters. This would just be pulling session duration since start and outputting if any are over particular threshold, or just return all. More or less what you can get from whoisactive, maybe just in format to do quick checks or something."

potatoqualitee:
Created as Get-DbaQueryExecutionTime

I think I'd actually like to add an alias for this Find-DbaLongRunningQuery. Maybe.... prolly not 🤔

PR #870, opened 2017-02-28
New function - Find slowest stored procedure and sql statement executions per database.

scinobe:
Return the slowest executions of stored procedures as a whole and
individual SQL statements.  SQL Statements can be ad hoc or within the
stored procedures.  Default values:  100 results, with 100 executions or
more with a minimum execution time of 500ms.  These defaults are to
prevent massive amounts of data coming back in databases with thousands
or more query plans.


ctrlbold:
Fixes #552

ctrlbold:
Bon!

PR #873, opened 2017-03-01
Improved internals

FriedrichWeinmann:
Updated Stop-Function and implemented Test-FunctionInterrupt, in order facilitate graceful code interruption.
Updated Stop-Function to require less mandatory input
Updated Write-Message to accept meaningful string names as Level.
New functionality: Write-Message can now be used to write a specific message only once per session, using the new -Once parameter.

### DIFF
--- a/bin/library.ps1
+++ b/bin/library.ps1
@@ -1,6 +1,23 @@
+﻿
+#region Test whether the module had already been imported
+$ImportLibrary = $true
+try
+{
+    $null = New-Object sqlcollective.dbatools.Configuration.Config
+    
+    # No need to load the library again, if the module was once already imported.
+    $ImportLibrary = $false
+}
+catch
+{
+    
+}
+#endregion Test whether the module had already been imported
 
-#region Source Code
-$source = @'
+if ($ImportLibrary)
+{
+    #region Source Code
+    $source = @'
 using System;
 
 namespace sqlcollective.dbatools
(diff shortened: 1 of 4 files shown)


bin/library.ps1
```diff
@@ -506,17 +603,73 @@ namespace sqlcollective.dbatools
...
-	continue
+    #endregion Source Code
+    
+    try
+    {
+        Add-Type $source -ErrorAction Stop
+    }
+    catch
+    {
+        throw
```
ctrlbold: This was causing issues for people when it throws if they import twice. Please add the continue back or resolve it somehow.

FriedrichWeinmann: It is resolved
Check the part above the the source code

ctrlbold: yay but also we never throw, can you put one of your pretty warnings? ;)

FriedrichWeinmann: I could, but that would be ... problematic.
You see, without the library, all that neat error handling isn't gonna work (in fact, it will cover the sky in red!). The only way I can stop Import-Module from completing its work is by throwing an exception.

So that's the one case where we don't get around a throw (But: It's beyond unlikely to occur. The library is extremely basic and I made sure to keep it that way). I can _also_ add a nice message if you want, so that the user has some better idea what happened.

ctrlbold: yes, say write-warning and continue? :D I never had the problem and cant repro so I can't say for sure if this wouldnt throw, yenno?

PR #876, opened 2017-03-01
Updated Restore-DbaDatabase. Fixes #468

### LINKED ISSUE #468: Restore-Database missing a call for KillAllProcesses
Is this a *feature* OR *bug*:
BUG / Suggestion

- Powershell Version: 3.0
- OS Version: 2012

**Steps to Reproduce**
Using SMO in a project with a customer. Restoring a backup, to replace the database, while relocating the files, failed for me. 
I found that I had to call the KillAllProcesses method on the sqlserver object, and parsing the name of the database that I was about to replace.

**Action Results**

**Expected Results**

**Additional comments**

sanderstad:
Implemented killing active sessions, database to single user and back to multiple users after restore


Stuart-Moore:
issue 486 doesn't refer to THIS function!  That's for the other Restore-Database, this is a completely seperate function

The killing of processes is done in Restore-DBFromFilteredArray which is the function that actually does the work, line 76 onwards

Your process also kills process to do the OutputScriptOnly action, which isn't required

The database is restored to whatever state it was in when it was backed up so there's no reason to force it to something else unless the user asks for that.

And  the $FunctionName should be in all the write- statement, that's the agreed way of doing things. (see dbatools-style for that discussion).

Cheers
Stuart

Zhunya:
Why you girls and boys hate turning database offline... this is 3-4-5 times you change way how you handle connection to database to be restored.

sanderstad:
Closed the pull request due to incorrect changes

niphlod:
there's also set-dbadatabasestate if you don't wanna code over and over "state changes"

PR #879, opened 2017-03-01
Get-DbaBackupHistory - added back fix for emailed issue

ctrlbold:
User emailed saying that all of the backup history files were combined when -Last was used. This was the fix. @Stuart-Moore, please check.


ctrlbold:
sorry what i meant @Stuart-Moore was please make sure this still works with yuor stuff. it was all combined before.

PR #883, opened 2017-03-01
Update to Write-Message and Stop-Function

FriedrichWeinmann:
Added Silent as default, not mandatory
Also added new level "Warning", deprecating the -Warning switch


ctrlbold:
beautiful changes 💅

PR #887, opened 2017-03-01
MasterInitial Commit of Find-DbaDatabaseGrowthEvent 

SQLvariant:
Returns any database AutoGrow events in the Default Trace for every
database on the SQL instance.


ctrlbold:
Nice work, buddy!

PR #900, opened 2017-03-03
Get-DbaPrivilege - Fix wrong resolve of SID

### LINKED ISSUE #898: Get-DbaPrivilege, failing to resolve users
### *bug*

the command uses secedit to export SIDs (which is good) but then fails:
- if the user is not a "common sid" (e.g. Administrators)
- if the user is not a domain user

That's because Convert-SIDToUserName needs to be run remotely as well, because SIDs for local users on the remote server may not match local ones.

PowerDBAKlaas:
Fixes #898 

- execute Convert-SIDToUserName remote
- force output to be an array
- remove try catch (the call never results in a terminating error
- add silentlycontinue to handle empty collection

How to test this code: 
- [ ] 
- [ ] 

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally

### DIFF
--- a/functions/Get-DbaPrivilege.ps1
+++ b/functions/Get-DbaPrivilege.ps1
@@ -51,17 +51,19 @@ function Get-DbaPrivilege
     [PSCredential] [System.Management.Automation.CredentialAttribute()]$Credential
   )
 
-BEGIN
+  BEGIN
   {
-    function Convert-SIDToUserName ([string] $SID ) {
-      $objSID = New-Object System.Security.Principal.SecurityIdentifier ("$SID") 
-      $objUser = $objSID.Translate( [System.Security.Principal.NTAccount]) 
-      $objUser.Value
+    $ResolveSID = @"
+    function Convert-SIDToUserName ([string] `$SID ) {
+      `$objSID = New-Object System.Security.Principal.SecurityIdentifier (`"`$SID`") 
+      `$objUser = `$objSID.Translate( [System.Security.Principal.NTAccount]) 
+      `$objUser.Value
     }
+"@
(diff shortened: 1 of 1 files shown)


functions/Get-DbaPrivilege.ps1
```diff
@@ -51,17 +51,19 @@ function Get-DbaPrivilege
...
   )
 
-BEGIN
+  BEGIN
   {
-    function Convert-SIDToUserName ([string] $SID ) {
-      $objSID = New-Object System.Security.Principal.SecurityIdentifier ("$SID") 
-      $objUser = $objSID.Translate( [System.Security.Principal.NTAccount]) 
-      $objUser.Value
+    $ResolveSID = @"
```
ctrlbold: Hey Klaas, Is there a reason this has to be in text? Could it be either be a [scriptblock]::Create(, or { } or just passing it whole to Invoke-Command even if it's duplicated code?

PowerDBAKlaas: I'm not sure, @ctrlbold . This is based on an example I found.
Surely repeating the whole block three times is not better?
When I have the time I can search for more concise methods, but this was the first that seems to work fine.
@niphlod @gbargsley can you confirm this resolves correctly?

ctrlbold: true, ill test :)

niphlod:
it works!

PR #908, opened 2017-03-04
Internals: Updated enumeration and Test-DbaDeprecation

FriedrichWeinmann:
Changes:
 - Added a new Alias for level 2 in the MessageLevel enumeration: Output
 - Added function: "Test-DbaDeprecation", which shall be used to manage renamed functions/parameters as well as deprecated parameters.

Test-DbaDeprecation
 - Will write a warning once per session when using a function/parameter/parameteralias we are going to remove
 - Can be tested against, making sure the content is actually removed in the release we announced to drop it.

### DIFF
--- a/bin/library.ps1
+++ b/bin/library.ps1
@@ -625,6 +625,11 @@ namespace sqlcollective.dbatools
             Host = 2,
 
             /// <summary>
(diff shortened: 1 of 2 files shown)


ctrlbold:
yesssss

PR #909, opened 2017-03-04
Get-DbaDatabaseFile

Stuart-Moore:
Fixes # 

 - New function
 - Returns the files and filegroups for specifed dbs (or entire instance)
 - Or takes a fragment of a physical name and returns which db owns it.

How to test this code: 
- [ ] Get-DbaDatabaseFile -SqlInstance sql2016 
- [ ] Get-DbaDatabaseFile -SqlInstance sql2016 -database master, model
- [ ] Get-DbaDatabaseFile -SqlInstance sql2016 -DataFilePath mastlog

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally


ctrlbold:
hey stuart,
Thank you so much! I have a PITA request :/ Unfortunately, the command only works on 2012 and up. We need to support all the way back to 2000, like smo.

```
$s = Connect-DbaSqlServer -SqlServer sql2000
$s.Databases[0].LogFiles
$s.Databases[0].FileGroups
```
Also, the SQL connection is in BEGIN, which means that it will not support multiple servers being passed.

And I'm unsure what's up with WebAnalyticsServiceApplicat.. in this photo.

image

Stuart-Moore:
Now works back to 2005 and I'm working on 2000

I'd not really intended to take in multiple sql instances, but not a problem to change that

Will mean the output will need to be changed though. Is the output from this intended for human or other command consumption?

If it's the later I'll just drop out as nested objects so people can slice and dice with where/select/group etc

Humans make it more difficult. Would probably end up being nested hashtables to allow things like:
$results["instance"]["Database"]["Primary"].values

If anyone knows how to write pretty collections like SMO uses I'm happy to give them a go.

The web analytics db is just something Sharepoint does at times, adding random filegroups. Code is reporting correctly

ctrlbold:
ahh I have an idea, @Stuart-Moore - I think getting it like SMO could be accomplished by something like a [pscustomobject] cast.

```
the last part would be like
foreach ($row in $finalobject) 
{
[pscustomobject]@{
	ComputerName = $server.NetName
	InstanceName = $server.ServiceName
	SqlInstance = $server.DomainInstanceName
	Database = $row.Name
	Data = $row.Value.Primary # or prolly not?, unsure how to unwrap it
	Log = $row.Value.Log
}
```

or something. my last two row names may suck.

Stuart-Moore:
I've gone with this atm:
            foreach ($db in $DatabaseCollection)
            {
                Write-Verbose "$FunctionName - Querying database $($db.name)"
                $results = Invoke-SqlCmd2 -ServerInstance $server.name -Query $sql -Database $($db.name)
 
                $Grouped = $results | Group-Object -Property fgname
                $FileGroups = @()
                Foreach ($Name in $Grouped)
                {
                    $GroupName = $Name.Name
                    if ($GroupName -eq '')
                    {
                        $GroupName = "LOGS"
                    }

                    $FileGroups += [PSCustomObject]@{
                                    Name = $GroupName
                                    ID = $Name[0].group[0].data_space_id
                                    IsDefault = $Name.group[0].FGIsDefault
                                    IsReadonly =  $Name.group[0].FGIsReadOnly
                                    Size = ($Name.group.Size | Measure-Object -sum).sum
                                    Files = $Name.group | Select-Object AvailableSpace,BytesReadFromDisk,BytesWrittenToDisk,FileName,Growth,GrowthType,ID,IsOffline,IsPrimaryFile,IsReadOnly,IsReadOnlyMedia,IsSparse,MaxSize,NumberOfDiskReads,NumberOfDiskWrites,Size,UsedSpace,VolumeFreeSpace,Name,State
                            }
                }
               [PSCustomObject]@{
                        SqlInstance = $server.name
                        Database = $db.name
                        FileGroups = $FileGroups
                    }
            }

Which then allows this:
$t =Get-DbaDatabaseFile -SqlInstance localhost\sqlexpress2016 -Databases fgtest
$t.filegroups
$t.FileGroups.Files

Which is  OK I think, and then also allows:
$t =Get-DbaDatabaseFile -SqlInstance localhost\sqlexpress2016
($t | where-Object {$_.Database -eq 'RestoreTime'}).FileGroups

to work even with multiple instances (comparing test/prod for instance)

Just going to pull yours into non gitted copy so I can have a play without losing my sql2000 stuff (working now)

PowerDBAKlaas:
Sorry @Stuart-Moore I think you're mixing too much here.
Either you want to return databases, filegroups or files. When you join those three in a hierarchical object model, you're rebuilding SMO.
$t.filegroups ans $t.filegroups[0].files is already possible when $t = Get-DbaDatabase.

I think we miss a Get-DbaFileGroups.
Collecting file information is done most part in Get-DbaDatabaseFreeSpace, which needs to be renamed (and maybe also rewritten).

ctrlbold:
@PowerDBAKlaas  - didn't we want to rebuild SMO? The way he outputs it is faster because it doesn't wait, it goes one by one instead of the way that SMO enumerates and waits.

@Stuart-Moore - I'm going to rework it to look more like SMO once you're totally done. Lemme know.

PowerDBAKlaas:
@ctrlbold  what I mean to say is that Get-DbaDatabaseFile should return one object per file, preferrably the file object.
The current PSCustomobject is actually a modified database object with dependent filegroups and files added as a collection, with a subset of properties,not easy reusable for other purposes.
If Get-DabDatabaseFileGroup and Get-DbaDatabaseFile are created and do what their name tells us, that'll be valuable for any purpose, including @Stuart-Moore 's requirements.

ctrlbold:
I see what you're saying. It does the collecting for you. I kind of liked that because we could easily outfit our previous commands that use SMO File enumeration, but I think you're right.

Stuart-Moore:
So, split this function into 2 as follows:

Get-DatabaseFileGroup - Just return the filegroup info for the specified instances and databases
Get-DatabaseFile - Just return the filegroup info for the specified instances and databases, with no mention of filegroups, or just include the fg names?

PowerDBAKlaas:
FileGroup is in File as [Parent]. We could add-Member or alias in Select-Defaultview.
I think it will be handy to have a filegroup property in the file object.

Another idea: if we also list log files, could we add missing properties from datafiles to logfiles as $null or a dummy value, and vice versa. If both object types have the same properties we can return them in one go, while it will still be different types?

potatoqualitee:
Latest commit looks like this! @PowerDBAKlaas , Check it!

image

potatoqualitee:
It will be easy to use this for a similar Get-DbaDatabaseFileGroup.

potatoqualitee:
Also, sorry @Stuart-Moore - I removed the fragment search - it returned a differently formatted object and seemed out of scope. It's slightly less efficient but a `| Where-Object` seems more appropriate.

potatoqualitee:
k merging - ive hella tested it. anyone who wants to make changes, please submit a patch PR!

PR #916, opened 2017-03-05
Managing Remote Management

FriedrichWeinmann:


FriedrichWeinmann:
Features I still want in it:
 - Option to globally disable any combination of connection types
 - Split up the configuration/Cache part into new commands: Get-DbaWmConnection, Set-DbaWmConnection, Remove-DbaWmConnection, New-DbaWmConnection
 - Option to automatically register the first working credential

Bugfix:
 - Current Windows Credentials cannot be registered as bad credentials at the moment. They should register as that, when they do not work.

FriedrichWeinmann:
Question:
What's your opinion on bad credentials:
When a user uses bad credentials (that are not previously known to be bad), what should the function do, beyond registering those credentials as bad:
 - Terminate with a warning / error?
 - Continue with the default or known credentials?
 - Terminate with an error if silent, continue with warning and default or known credentials if non-silent?

niphlod:
That's going to be hard. 
From the security perspective, continuing to use credentials on "resource y" that are know to not work on "resource x" could lead to e.g. locked account.
On the other end you get the "smooth usage" generally preferred in dbatools, which basically warns and continues anyway. 
From a UX standpoint, the latter wins. 
From a security and scripting standpoint, the first does.
I'm happy dbatools is "user-friendly(er)" than other modules but I find myself "covering-coding" when scripting out automations (kinda set-that AND then get-that to see if it has been correctly set).

FriedrichWeinmann:
Oh, no question, if you spam wrong credentials at every server, you're likely to got locked out.
That's a bit of a brains situation and already the case as we live it right now.
The thing is, my function will alleviate that by _refusing_ to use credentials it _knows_ are bad. However, it can only know this on a per-server basis (no way to know what the user has actually access to).
Currently, it will completely interrupt mass processing, if you set it to Silent and warn if the user runs it without (So if you use it in silent mode when used from another function, you'll get the error).
A user can intervene mass miss-authentication by CTRL-C.

functions/Test-DbaWMConnection.ps1
```diff
@@ -0,0 +1,428 @@
...
+        The results will be reported and registered, future calls from Get-DbaWmObject will thus recognize the results and optimize the query.
+    
+    .EXAMPLE
+        PS C:\> Test-DbaWMConnection -ComputerName sql2014 -Credential $null -DisableConnectionType PowerShellRemoting, Wmi -OverrideCredential -SetCredential -Type CimDCOM, CimRM
+    
+        This test will run a connectivity test against the computer sql2014
+        - It will use windows authentication of the current user
+        - It will only test Cim over DCOM and Cim over WinRM
+        - It will disable the use of PowerShellRemoting and WMi as a connectivity type
+        - It will configure all connections to this server to use the current windows credentials, instead of what ever may be specified by the user or another function.
```
ClaudioESSilva: How is that done? You say "or another function". If I create a function and use WMI, for example, how will this command overwrite that one? 🤔
Is this only for modules inside dbatools?

ClaudioESSilva:
@FriedrichWeinmann In the case of _refusing to use credentials it knows are bad_, lets say my account was not well configured yet. In this case I will need to call this command alone to -ClearCache?

functions/Test-DbaWMConnection.ps1
FriedrichWeinmann: These functions will be public, that is, users can use it directly, either manually or in scripts. So can our own functions of course. Basically, it's an internal function that is so useful, that we decided to make it available to our users directly, for their own needs.
Of course, settings of the connection management only affect calls made through our own connection functions. When a user directly uses Get-WmiObject, then that will use wmi and only with the connection information specified by the user, as usual. We do not override the regular commands. The very next line points out that this affects Get-DbaWmObject only.

Do I need to be more explicit than that?

FriedrichWeinmann:
@ClaudioESSilva yes, if your credentials become valid during a session, but were already blacklisted by the cache, you need to clear that cache, in order to be able to use it. That is what currently is done via the '-ClearCache' parameter and will probably be moved to another function before final release. This odd packaging of functionalities into Test-DbaWmConnection was merely done due to draconian limitations imposed by our BDFL. She might have reconsidered by now.

PowerDBAKlaas:
@FriedrichWeinmann I'd like to raise following cautions:

The order of preference would be:
1. Get-Ciminstance ( sessionless, because that's usually faster)
2. New-Cimsession over WSMan
3. New-Cimsession over DCom

There should be no cases where 3. fails and Get-WmiObject succeeds since PS V3 is our basic requirement.

I also wonder if using 4. Invoke-Command when 2. has failed may work in some special configurations, but if so we should use Get-Ciminstance and not Get-WmiObject.

1. can be skipped immediately if:
- credentials are supplied (- the user doesn't want to use his current credentials, or - we try to connect to an IP), or
 - we need Invoke-CimMethod, or
 -we know we need multiple calls to the same computer

Also consider that a failing test for a particular class does not mean this connection type is not working.

Sometimes we need to create a session and reuse it multiple times in the calling function. The session should therefore not be closed after the first call; maintaining a session is the whole purpose of it since creating and removing is costly, and useless if we only execute Get-Ciminstance without credential.

FriedrichWeinmann:
Good points, altogether.
There is in fact a scenario where PowerShell remoting works and Cim over WinRM fails. Basically, it's part of the listener configuration. Doesn't happen in standard, but happened in some of the earlier examples you can still find on msdn.
Also: Does CimRM work on a target with PSv2?
Not using the sessions on Get-CimInstance when using default credentials should indeed be done. Gonna implement that.
I've considered reusing the sessions. Probably won't do it in the first release, but will add it afterwards (yet still for 1.0).

Wasn't too sure about completely dropping WMI out of the system, however when I add global options on enabling/disabling individual paths, I was intending to disable path 3 and 4 by default, making them opt-in measures.

Was using Get-WmiObject in the scriptblock of option 4 out of mental laziness. The target may be an old computer with PS2. Probably should make it an if/else thing, but the performance impact is so negligible compared to the remoting overhead (the call itself after all is local, and Get-CimInstance's main speed advantage is over the network) ...

I've considered the tests and will update them for other eventualities as I can discern them from regular failings. However, as it is, the connection will not refuse to run against targets whose connection failed: It will choose the failing connections last is all. A global ping test to avoid going through all hoops on a wrong computername is on the todo list (and it will be possible to globally or locally disable this test).
Basically, currently, a request will try each step once on a request, even if it knows the step failed before.
Was considering implementing a timeout on individual protocols ("if it fails, don't try it again for 5 minutes"), will probably add it.

PowerDBAKlaas:
Targets PS V2 should make no difference for the cmdlets. It does make a difference in namespaces and classes available. So we'll have to use win32_networkadapter for a while since cim_networkadapter may result in a lot more errors. But the CIM cmdlets should work anywhere WMI does.
Not the other way around though: Get-WmiObject will certainly fail on Linux and on Nano Server.

When using Invoke-Command it's indeed another situation, but if we take that into consideration, it would mean we have change requirements everywhere because whatever we try against a PS V2 target will fail if we use PS V3 functionality.

ctrlbold:
Hello all, where do we stand with this PR?

ctrlbold:
Hey @FriedrichWeinmann -- is this ready? Also, DisableConnectionType needs to be documented in Test-DbaWmConnection - it's causing a failure of the pester test.

ctrlbold:
hey @FriedrichWeinmann - I know you've been working hard on some SUPER COOL shit that blows my mind, so I hate to ask, but is this nearing completion? I'm eager to see it completed cuz I wanna play. Thank you 🙇

FriedrichWeinmann:
Working on it full-time tonight, might make it, might take another

ctrlbold:
hey @FriedrichWeinmann - can you address the xml/dbatools.Format.ps1xml issue? Also, ping me when this is ready! Thanks

FriedrichWeinmann:
Alright, I'll kill this PR and rebuild it from the ground up - too many changes incoming and can't be bothered to fix it on the run ;)

PR #926, opened 2017-03-07
Remove obsolete condition on Restore-SqlBackupFromDirectory

ie-zero:
Restore-SqlBackupFromDirectory function have twice identical condition
on a IF statement. So one is removed.


ctrlbold:
Thank you, @ie-zero !

PR #935, opened 2017-03-10
Get-DbaTable

Staggerlee011:
v1

### DIFF
--- a/functions/Get-DbaTable.ps1
+++ b/functions/Get-DbaTable.ps1
@@ -38,15 +38,11 @@ Return all tables in the Test1 database
 Get-DbaTable -SqlInstance DEV01 -Database MyDB -Table MyTable
 Return only information on the table MyTable from the database MyDB
 .EXAMPLE
-Get-DbaTable -SqlInstance DEV01 -Database MyDB -Table MyTable
-using pipeline
-.EXAMPLE
-Get-DbaTable -SqlInstance DEV01 -Database MyDB -Table MyTable
-with multiple instances
-.EXAMPLE
 Get-DbaTable -SqlInstance DEV01 -Table MyTable
-REturns information on table called MyTable if it exists in any database on the server, under any schema
-
+Returns information on table called MyTable if it exists in any database on the server, under any schema
+.EXAMPLE
+@('localhost','localhost\namedinstance') | Get-DbaTable -Database DBA -Table Commandlog
+Returns information on the CommandLog table in the DBA database on both instances localhost and the named instance localhost\namedinstance
 
 #>
 	[CmdletBinding()]
@@ -123,7 +119,7 @@ REturns information on table called MyTable if it exists in any database on the 
 			{
 				if ($databases.length -gt 0)
 				{
-					$dbs = $server.Databases | Where-Object { $databases -contains $_.Name }
+					$dbs = $server.Databases | Where-Object { $databases -contains $_.Name -and $_.status -eq 'Normal' }
 				}
 				elseif ($IncludeSystemDBs)
 				{
--- a/functions/Get-DbaTable.ps1
+++ b/functions/Get-DbaTable.ps1
@@ -1,9 +1,8 @@
 ﻿function Get-DbaTable
 {
 <#
-<#
 .SYNOPSIS
-Returns a summary of information on the table
+Returns a summary of information on the tables
 .DESCRIPTION
 Shows table information around table row and data sizes and if it has any table type information. 
 .PARAMETER SqlInstance
@@ -12,7 +11,7 @@ collection and recieve pipeline input
 .PARAMETER SqlCredential
 PSCredential object to connect as. If not specified, currend Windows login will be used.
 .PARAMETER Database
-Define the database you wish to search
+Define the databases you wish to search
 .PARAMETER IncludeSystemDBs
 Switch parameter that when used will display system database information
 .PARAMETER Table
@@ -38,6 +37,14 @@ Return all tables in the Test1 database
 .EXAMPLE
 Get-DbaTable -SqlInstance DEV01 -Database MyDB -Table MyTable
 Return only information on the table MyTable from the database MyDB
+.EXAMPLE
+Get-DbaTable -SqlInstance DEV01 -Database MyDB -Table MyTable
+using pipeline
+.EXAMPLE
+Get-DbaTable -SqlInstance DEV01 -Database MyDB -Table MyTable
+with multiple instances
+.EXAMPLE
+
 
 #>
 	[CmdletBinding()]
@@ -46,8 +53,8 @@ Return only information on the table MyTable from the database MyDB
 		[object[]]$SqlInstance,
 		[System.Management.Automation.PSCredential]$SqlCredential,
 		[switch]$IncludeSystemDBs,
-		[switch]$Silent,
-		[string[]]$Table
+		[string[]]$Table,
+		[switch]$Silent
 	)
 	
 	DynamicParam { if ($SqlInstance) { return Get-ParamSqlDatabases -SqlServer $SqlInstance[0] -SqlCredential $SourceSqlCredential } }
@@ -57,6 +64,8 @@ Return only information on the table MyTable from the database MyDB
 		$databases = $psboundparameters.Databases
 		$exclude = $psboundparameters.Exclude
 		
+        $defaultprops = @("Parent", "Schema", "Name", "IndexSpaceUsed", "DataSpaceUsed", "RowCount", "HasClusteredIndex", "IsFileTable", "IsMemoryOptimized", "IsPartitioned", "FullTextIndex", "ChangeTrackingEnabled")
+
 		$fqtns = @()
 		
 		if ($Table)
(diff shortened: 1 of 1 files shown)


functions/Get-DbaTable.ps1
```diff
@@ -0,0 +1,176 @@
...
+Return only information on the table MyTable from the database MyDB
+
+#>
+	[CmdletBinding()]
+	param ([parameter(ValueFromPipeline, Mandatory = $true)]
+		[Alias("ServerInstance", "SqlServer")]
+		[object[]]$SqlInstance,
+		[System.Management.Automation.PSCredential]$SqlCredential,
+		[switch]$IncludeSystemDBs,
+		[switch]$Silent,
```
FriedrichWeinmann: Please make silent the last parameter in the function. Most users will never use it and the order the parameters are defined in is also the one the user tabs through the possible parameters

functions/Get-DbaTable.ps1
```diff
@@ -0,0 +1,176 @@
...
+			}
+		}
+	}
+	
+	PROCESS
+	{
+		foreach ($instance in $sqlinstance)
+		{
+			#For each SQL Server in collection, connect and get SMO object
+			Write-Message -Level Verbose -Message "Connecting to $instance"
```
FriedrichWeinmann: Duplicate Message, will also be written in the try block

functions/Get-DbaTable.ps1
```diff
@@ -0,0 +1,176 @@
...
+			
+			try
+			{
+				Write-Message -Level Verbose -Message "Connecting to $instance"
+				$server = Connect-SqlServer -SqlServer $instance -SqlCredential $sqlcredential
+			}
+			catch
+			{
+				Stop-Function -Message "Failed to connect to: $instance" -Continue -Target $instance
```
FriedrichWeinmann: Please add the parameter `-InnerErrorRecord $_`
This will attach the original exception as a nested exception, allowing seasoned users to drill down into an issue.

functions/Get-DbaTable.ps1
```diff
@@ -0,0 +1,176 @@
...
+				}
+				
+				if ($exclude.length -gt 0)
+				{
+					$dbs = $dbs | Where-Object { $exclude -notcontains $_.Name }
+				}
+			}
+			catch
+			{
+				Stop-Function -Message "Unable to gather dbs for $instance" -Target $instance -Continue
```
FriedrichWeinmann: Please add the parameter `-InnerErrorRecord $_`
This will attach the original exception as a nested exception, allowing seasoned users to drill down into an issue.

functions/Get-DbaTable.ps1
```diff
@@ -0,0 +1,176 @@
...
+					{
+						if ($fqtn.schema)
+						{
+							try
+							{
+								$db.Tables | Where-Object { $_.name -eq $table -and $_.Schema -eq $schema } | Select-DefaultView -Property Parent, Schema, Name, IndexSpaceUsed, DataSpaceUsed, RowCount, HasClusteredIndex, IsFileTable, IsMemoryOptimized, IsPartitioned, FullTextIndex, ChangeTrackingEnabled
+							}
+							catch
+							{
+								Write-Message -Level Warning "Could not find table name: $($fqtn.table) schema: $($fqtn.schema)"
```
FriedrichWeinmann: Please avoid positional binding, specify `-Message "Could ..."` instead.
When catching a non-critical error (one where you do not interrupt execution with Stop-Function for), you can still log the error to help troubleshooting. To do so, add `-ErrorRecord $_` as a parameter to Write-Message (nly when inside a catch block)

functions/Get-DbaTable.ps1
```diff
@@ -0,0 +1,176 @@
...
+						}
+						else
+						{
+							try
+							{
+								$db.Tables | Where-Object { $_.name -eq $table } | Select-DefaultView -Property Parent, Schema, Name, IndexSpaceUsed, DataSpaceUsed, RowCount, HasClusteredIndex, IsFileTable, IsMemoryOptimized, IsPartitioned, FullTextIndex, ChangeTrackingEnabled
+							}
+							catch
+							{
+								Write-Message -Level Warning "Could not find table name: $($fqtn.table) schema: $($fqtn.schema)"
```
FriedrichWeinmann: Please avoid positional binding, specify `-Message "Could ..."` instead.
When catching a non-critical error (one where you do not interrupt execution with Stop-Function for), you can still log the error to help troubleshooting. To do so, add `-ErrorRecord $_` as a parameter to Write-Message (nly when inside a catch block)

functions/Get-DbaTable.ps1
```diff
@@ -0,0 +1,176 @@
+﻿function Get-DbaTable
+{
+<#
+<#
```
PowerDBAKlaas: looks like one <# too many?

functions/Get-DbaTable.ps1
```diff
@@ -0,0 +1,176 @@
...
+.SYNOPSIS
+Returns a summary of information on the table
```
PowerDBAKlaas: "on one or more tables" ?

functions/Get-DbaTable.ps1
```diff
@@ -0,0 +1,176 @@
...
+.DESCRIPTION
+Shows table information around table row and data sizes and if it has any table type information. 
+.PARAMETER SqlInstance
+SQLServer name or SMO object representing the SQL Server to connect to. This can be a
+collection and recieve pipeline input
+.PARAMETER SqlCredential
+PSCredential object to connect as. If not specified, currend Windows login will be used.
+.PARAMETER Database
+Define the database you wish to search
```
PowerDBAKlaas: or databases

functions/Get-DbaTable.ps1
```diff
@@ -0,0 +1,176 @@
...
+	
+.LINK
+https://dbatools.io/Get-DbaTable
+.EXAMPLE
+Get-DbaTable -SqlInstance DEV01 -Database Test1
+Return all tables in the Test1 database
+.EXAMPLE
+Get-DbaTable -SqlInstance DEV01 -Database MyDB -Table MyTable
+Return only information on the table MyTable from the database MyDB
+
```
PowerDBAKlaas: please include examples

- using pipeline
- with multiple instances
- submitting table but no database

functions/Get-DbaTable.ps1
```diff
@@ -0,0 +1,176 @@
...
+				
+				if ($dotcount -eq 1)
+				{
+					$schema = $t.Split(".")[0]
+					$table = $t.Split(".")[1]
+				}
+				
+				if ($dotcount -eq 2)
+				{
+					$database = $t.Split(".")[0]
```
PowerDBAKlaas: a shortcut is possible ( but not necessary ):
$database, $schema, $table = $t.split(".")

functions/Get-DbaTable.ps1
```diff
@@ -0,0 +1,176 @@
...
+					$schema = $t.Split(".")[1]
+					$table = $t.Split(".")[2]
+				}
+				
+				$fqtn = [PSCustomObject] @{
+					Database = $database
```
PowerDBAKlaas: By just reading the code, I think $database and / or $schema may retain the value of the previous iteration, when first table is "db1.dbo.table1" and the next "table2". Did you test that?

functions/Get-DbaTable.ps1
```diff
@@ -0,0 +1,176 @@
...
+			catch
+			{
+				Stop-Function -Message "Unable to gather dbs for $instance" -Target $instance -Continue
+			}
+			
+			foreach ($db in $dbs)
+			{
+				Write-Message -Level Verbose -Message "Processing $db"
+				
+				$d = $server.Databases[$db]
```
PowerDBAKlaas: $dbs can be $null, indexing will result in an error.
So you should catch that first,
maybe `if ($dbs.count -gt 0)` ?

functions/Get-DbaTable.ps1
```diff
@@ -0,0 +1,176 @@
...
+				if ($fqtns)
+				{
+					foreach ($fqtn in $fqtns)
+					{
+						if ($fqtn.schema)
+						{
+							try
+							{
+								$db.Tables | Where-Object { $_.name -eq $table -and $_.Schema -eq $schema } | Select-DefaultView -Property Parent, Schema, Name, IndexSpaceUsed, DataSpaceUsed, RowCount, HasClusteredIndex, IsFileTable, IsMemoryOptimized, IsPartitioned, FullTextIndex, ChangeTrackingEnabled
```
PowerDBAKlaas: Since you need the property collection 3 times, it's better to put that in a variable $defaultprops or so, in BEGIN{}

functions/Get-DbaTable.ps1
```diff
@@ -0,0 +1,187 @@
...
+				Stop-Function -Message "Failed to connect to: $instance" -Continue -Target $instance -InnerErrorRecord $_
+			}
+			
+			#If IncludeSystemDBs is true, include systemdbs
+			#only look at online databases (Status equal normal)
+			try
+			{
+				if ($databases.length -gt 0)
+				{
+					$dbs = $server.Databases | Where-Object { $databases -contains $_.Name }
```
PowerDBAKlaas: sorry @Staggerlee011 , I forgot this in the first review:
shouldn't the check for status -eq 'Normal' also be added when $databases are submitted?

ctrlbold: Thanks much for the review, @PowerDBAKlaas!

ctrlbold:
Thank you both! ❤️

PR #937, opened 2017-03-10
CIM for DiskAlignment

ctrlbold:
Can someone who knows something about CIM fix this? It's almost fixed.

### DIFF
--- a/functions/Test-DbaDiskAlignment.ps1
+++ b/functions/Test-DbaDiskAlignment.ps1
@@ -94,10 +94,10 @@ Function Test-DbaDiskAlignment
         [switch]
         $Detailed,
         
-        [System.Data.SqlClient.SqlCredential]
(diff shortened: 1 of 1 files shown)


PowerDBAKlaas:
this function too has almost all code wrapped in a nested function in BEGIN{}
I just struggled with that in Get-DbaDiskspace
I think it creates problems with passing the necessary parameters,
like $cimsession, which is created in PROCESS{}, but is it handed over to the nested function?

FriedrichWeinmann:
Yeah, I'm currently working on it. Will remove all scope boundary violations and change some other things around.

FriedrichWeinmann:
And that should do it, please test the overhauled Test-DbaDiskAlignment

ctrlbold:
Thanks, @FriedrichWeinmann !

PR #940, opened 2017-03-10
Hush Resolve network name

ctrlbold:


functions/Resolve-DbaNetworkName.ps1
```diff
@@ -173,12 +173,12 @@ Function Resolve-DbaNetworkName
 					Write-Verbose "$functionName - Getting computer information from $Computer via CIM (WSMan)"
 					if ($Credential)
 					{
-						$CIMsession = New-CimSession -ComputerName $Computer -ErrorAction SilentlyContinue -Credential $Credential
-						$conn = Get-CimInstance -Query "Select * FROM Win32_computersystem" -CimSession $CIMsession -ErrorAction SilentlyContinue
+						$CIMsession = New-CimSession -ComputerName $Computer -ErrorAction Stop -Credential $Credential
```
PowerDBAKlaas: ?? is this because of the error in Test-DbaDiskAlignment?
I'd think this generates more error, not less

The error is in line 204, the only one without silentlycontinue

FriedrichWeinmann: Nope, these are done within a try block, so the catch will trigger.
I've added the same for the rest of the Cim-Calls, hushing everything. As a stop-gap for tomorrow it should suffice.

PR #946, opened 2017-03-13
Bugfix for Error Record Category

FriedrichWeinmann:
The input validation was not sufficient, causing the function to resolve to a bad error category, when ...
 - An inner error record was provided
 - That error record would not contain a proper CategoryInfo
A new validation condition is applied, ensuring that only when a valid categoryinfo property is on the record it will be used to override the default


ctrlbold:
Thanks fred!

PR #949, opened 2017-03-14
Master

SQLRockstar:
one fix


ConstantineK:
Hey @SQLRockstar!
It looks like we might have a little PR mix-up, it looks this PR is merging `abandoned-for-now-code` into master, when instead you probably want to merge `new-command` into `development`.

I am going to hit you up on Slack and see if we can work this out :)

ConstantineK:
Going to quickly close this and move the commits to the next PR, it looks like whether a PR is fork or no fork is not something that can be changed after the fact.

PR #950, opened 2017-03-14
Test-DbaLinkedServerConnection

ConstantineK:
Adding for a friend :)


ConstantineK:
And even more clearly, you cant change the source branch, only the destination for a PR :)

PR #952, opened 2017-03-14
Fix Get-SqlRegisteredServerName Group Parameter Docs

propellor:
Examples were referencing parameter 'Groups' which is incorrect - should be 
Group.  Web page for the command also references groups, which needs to be corrected as
well.  Group parameter is dynamic parameter which causes it to not show
in get-help or intellisense, which leaves the user in the dark.  I added
a note of this to the description.  Not sure if the dynamic parameter
not showing in get-help and intellisense can be fixed.


ctrlbold:
Thanks so much, @propellor !

Issue #936, opened 2017-03-10
Test-DbaSpn and Get-DbaSpn, minor bug in domain restrictions


niphlod:
### bug

When checking the existence of the account all of "SPN" commandlets strip the domain part from the account, assuming the account is registered under the same domain of the server running MSSQL.
It *can* happen (and if something strange happens, niphlod has it) that you have a domain forest like so
- corp.com
- subcorp.corp.com

when the server is registered under subcorp.corp.com, but the user is on corp.com, the root DN for the search is not taken into account.

Get-DbaSPN shows the correct account (i.e. corp\mssql_user) ... Test-DbaSPN too .... but then when you pipe it down to Set-DbaSPN it gets stripped. (yeah, I know there is a "domain" parameter, but the wording in this case is misleading)

A little more serious, what happens with Test-DbaSpn is that the SPN is reported to not being there ... but it's because the object is not found on subcorp.corp.com rather than because the SPN is missing at all.

```.PARAMETER Domain
If your server resides on a different domain than what your current session is authenticated against, you can specify a domain here. This
parameter is optional.
```
it should instead be
```.PARAMETER Domain
If your server OR THE SERVICE ACCOUNT resides on a different domain than what your current session is authenticated against, you can specify a domain here. This
parameter is optional.
```

That being said (i.e. that technically documenting the shortcoming in this particular corner-case is fine) the cmdlet *could* use [System.DirectoryServices.ActiveDirectory.Forest]::GetCurrentForest() to search properly the accounts in all of dbatool's spn commandlets (i.e. "retaining the info" about the domain where the account is reported to be and searching in the correct DN)

niphlod:
figured out patches to fix them. As soon as https://github.com/sqlcollaborative/dbatools/pull/1027 is merged I'll post a PR

PR #953, opened 2017-03-14
new internal Get-ADObject

niphlod:
This is a straight plug from Test-DbaValidLogin (which seems the most
correct implementation when searching) but could help also other cmdlets
in the need to work with AD. It's a mere first draft that points to
maximizing performances:
- no need to spend time in collecting all forests' netbios names
- shortcuirciting the dns to dn translation
- possibly caching it (ATM just in script scope)

BTW: in my env Test-DbaValidLogin passes ~45 seconds in the BEGIN {} just enumerating
all the domains (think a forest of 6 main + 30ish crosstrusted) 

A quick plug of this new logic into Test-DbaValidLogin shortcircuits a lot of machinery and the whole command finishes up in less than 10 seconds.

To be fair it was built to address more easily https://github.com/sqlcollaborative/dbatools/issues/936 but every speedup in any command is welcome ^_^

### DIFF
--- a/internal/Get-DbaADObject.ps1
+++ b/internal/Get-DbaADObject.ps1
@@ -1,0 +1,131 @@
+Function Get-DbaADObject
+{
+<#
+.SYNOPSIS
+Get-DbaADObject tries to facilitate searching AD with dbatools, which ATM can't require AD cmdlets.
+
+.DESCRIPTION
+As working with multiple domains, forests, ldap filters, partitions, etc is quite hard to grasp, let's try to do "the right thing" here and
+facilitate everybody's work with it.
+
+.PARAMETER ADObject
+ATM it takes something in the format "DOMAIN\LdapFilter" but let's see how to make it better
+
+.NOTES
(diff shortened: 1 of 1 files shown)


niphlod:
hum. this could be further simplified (and speeded up a bit without the convoluted logic) if we allow System.DirectoryServices.AccountManagement , which is included in .net 3.5 if I'm not mistaken. 
Anyone against it ? @ctrlbold , @ConstantineK , @wsmelton ?

wsmelton:
I'm not against it if it is included in majority of OS'

niphlod:
seems so from [this](https://msdn.microsoft.com/en-us/library/system.directoryservices.accountmanagement(v=vs.110).aspx) link so I'll go down that route. Will first rewrite this to be more useful and then reengineer commands using different kinds of AD searches to use this component instead of their own baked solutions.

ctrlbold:
Can you integrate this with the required commands, @niphlod ? Merging now.

niphlod:
this was in no shape to be merged .....  ^_^

BTW: what are "the required commands" ?

PR #962, opened 2017-03-16
Get-DbaSqlBuildReference, tests

niphlod:
 - tests for Get-DbaSqlBuildReference

For now it's just checking the data in the json. As this is effectively
my first shot at Pester, please do tell if anything doesn't add up
properly


niphlod:
I was unsure about it but then I asked myself that without this we may end in the same situation of other sources... staleness. 
With this, AT LEAST every 45 days we should check for new releases coming / different EOL dates as per MLP. 
Remediation would just need to update the LastUpdated at the top of the json, after confirming the data is correct.

PR #972, opened 2017-03-17
Remove-DbaBackup, fixes #970

### LINKED ISSUE #970: Remove-DbaBackup, Slow Folder Removal and Failed Deletion of Files
Bug

### System Details

- Operating system name and version:

```
Evaluate $PSVersionTable in PowerShell and paste the output here
```

- Output of dbatools version:

```
Evaluate (Get-Module dbatools -ListAvailable).Version and paste output here
```

### Steps to Reproduce

When the option to remove emp
foldersdeleted
ty folders is used it takes a long time to delete a small amount of folders.  See attached.

Also, when you delete a large amount of tlogs backups for example if the file is deleted prior to the process getting to that file it throws a red error.

### Attached Logs or any exception errors

- Use of Start-Transcript can help collection of console output and exceptions
- Certain commands will generate an exception log that you can find in following directory: `dir $env:USERPROFILE\Documents\dbatools-exceptions.txt`

niphlod:
Fixes #970 

 - avoid long waits to build up file lists, everything is pipelined, so concurrent alteration of the underlying filestructure are less likely to impact the command itself
 - new logging subsystem
 - returns the removed files and directory fullnames for empty folder removals
- avoids top-down recursion when removing empty folders (which ultimately was the reason behind the relative slowness described)

tl;dr: on 302k files nested into 5k dirs on a local SSD the process passed from 20min to 6min

### DIFF
--- a/functions/Remove-DbaBackup.ps1
+++ b/functions/Remove-DbaBackup.ps1
@@ -13,11 +13,11 @@ backups have been archived to your archive location before removal.
 Also included is the ability to remove empty folders as part of this cleanup activity.
 
 .PARAMETER Path
-Name of the base level folder to search for backup files. 
+Name of the base level folder to search for backup files.
 Deletion of backup files will be recursive from this location.
 
 .PARAMETER BackupFileExtension
-Extension of the backup files you wish to remove (typically bak or log)
+Extension of the backup files you wish to remove (typically bak, trn or log)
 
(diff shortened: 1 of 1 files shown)


niphlod:
@FriedrichWeinmann : I'd need a crash course on how to handle possible throws with the new system. stop-function is the solution there ?

PR #986, opened 2017-03-23
Get-DbaADObject, streamlines the process for everybody.

niphlod:
Builds on top of ideas had on https://github.com/sqlcollaborative/dbatools/pull/953 . Integrations in cmdlets using AD is coming in

### DIFF
--- a/internal/Get-DbaADObject.ps1
+++ b/internal/Get-DbaADObject.ps1
@@ -43,6 +43,11 @@ Get-DbaADObject -ADObject "contoso\sqlserver2014$" -Type Group
 
 Seaches in the contoso domain for a sqlserver2014 computer (remember the ending $ for computer objects)
 
+.EXAMPLE
+Get-DbaADObject -ADObject "contoso\ctrlb" -Type User -Silent
+
+Seaches in the contoso domain for a ctrlb user, suppressing all error messages
+
 #>
 	[CmdletBinding()]
 	Param (
@@ -76,8 +81,7 @@ Seaches in the contoso domain for a sqlserver2014 computer (remember the ending 
 		foreach($ADObj in $ADObject) {
 			$Splitted = $ADObj.Split("\")
 			if ($Splitted.Length -ne 2) {
-				Stop-Function -Message "You need to pass ADObject in DOMAIN\object format" -Silent $Silent
-				continue
+				Stop-Function -Message "You need to pass ADObject in DOMAIN\object format" -Continue -Silent $Silent
 			}
 			$Domain, $obj = $Splitted
 			try {
--- a/internal/Get-DbaADObject.ps1
+++ b/internal/Get-DbaADObject.ps1
@@ -60,7 +60,7 @@ Seaches in the contoso domain for a ctrlb user, suppressing all error messages
 		try {
 			Add-Type -AssemblyName System.DirectoryServices.AccountManagement
 		} catch {
-			write-warning 'Failed to load the required module'
+			Stop-Function -Message "Failed to load the required module $($_.Exception.Message)" -Silent $Silent -InnerErrorRecord $_
 		}
 		switch ($Type) {
 			"User" {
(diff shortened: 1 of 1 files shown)


internal/Get-DbaADObject.ps1
```diff
@@ -24,108 +29,65 @@ This program is distributed in the hope that it will be useful, but WITHOUT ANY
...
-				$LDAP = ($script:domains | Where-Object NetBios -eq $domain).LDAP
-				$ad = New-Object System.DirectoryServices.DirectoryEntry $LDAP
-				$searcher = New-Object System.DirectoryServices.DirectorySearcher
-				$searcher.SearchRoot = $ad
-				$searcher.Filter = $filter
+		foreach($ADObj in $ADObject) {
+			$Splitted = $ADObj.Split("\")
+			if ($Splitted.Length -ne 2) {
+				Stop-Function -Message "You need to pass ADObject in DOMAIN\object format" -Silent $Silent
+				continue
```
FriedrichWeinmann: Instead of calling the continue statement in the next line, use the `continue` parameter. Otherwise this code will never run

internal/Get-DbaADObject.ps1
```diff
@@ -24,108 +29,65 @@ This program is distributed in the hope that it will be useful, but WITHOUT ANY
...
-			{
-				if ($x -eq ($array.Length - 1)) { $separator = "" }
-				else { $separator = "," }
-				[string]$dn += "DC=" + $array[$x] + $separator
-			}
-			return $dn
+		try {
+			Add-Type -AssemblyName System.DirectoryServices.AccountManagement
+		} catch {
+			write-warning 'Failed to load the required module'
```
FriedrichWeinmann: Should be using `Write-Message -Level Warning` instead

internal/Get-DbaADObject.ps1
```diff
@@ -24,108 +29,65 @@ This program is distributed in the hope that it will be useful, but WITHOUT ANY
 You should have received a copy of the GNU General Public License along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 .EXAMPLE
-Get-DbaADObject -ADObject "contoso\(&(objectCategory=User)(sAMAccountName=ctrlb))"
+Get-DbaADObject -ADObject "contoso\ctrlb" -Type User
```
FriedrichWeinmann: Maybe an example with the silent parameter might be helpful, if it's intended for internal usage

internal/Get-DbaADObject.ps1
niphlod: on hindsight, warning is not the correct way to handle this ..... this exception would need to stop the function with a critical error. Without that type, we can't do nothing at all

FriedrichWeinmann: Yeah, you got a point there, that needs a Stop-Function instead

wsmelton: I've come across this on other commands I've looked at, certain places original author used Write-Warning and would need to be replaced with `Stop-Function` and not just `Write-Message -Warning`. We will have to look at those more closely to make sure each is used properly.

FriedrichWeinmann: Aye. Note please however that `Write-Message -Warning` is deprecated in favor of `Write-Message -Level Warning`.

niphlod: but in a BEGIN block how can Stop-Function be used ? forcing the silent parameter to $true ?

FriedrichWeinmann: If you use Stop-Function in begin, use this at the beginning of both Process and end:
That'll kill those blocks as soon as they begin

niphlod:
@FriedrichWeinmann everything should be ok now

ctrlbold:
Waiting for confirmation from @FriedrichWeinmann

niphlod:
done with everything I could, except for 

> For the Messaging sign off you'd need to add some verbose messages that show the current progress.
> For the CodeStyle sign off you'd need another who understands the code style standards. I'm not current on the requirements for that.

which ATM is out of my league :D

PR #988, opened 2017-03-26
Get-DbaBackupHistory

FriedrichWeinmann:
Redesigned function to
- Implement current messaging system
- Implement current flow control system
- Output strong types
- Use the new DateTime & TimeSpan objects

Library update
- New DateTime objects to support custom display styles
- New TimeSpan object to support trimmed Timespan style without loosing
accuracy


functions/Get-DbaBackupHistory.ps1
```diff
@@ -1,449 +1,441 @@
...
+                            $obj.FileList = $server.ConnectionContext.ExecuteWithResults($Filesql).Tables.Rows
+                            $obj.Position = $group.Group[0].Position
+                            $obj.FirstLsn = $group.Group[0].First_LSN
+                            $obj.DatabaseBackupLsn = $group.Group[0].database_backup_lsn
+                            $obj.CheckpointLsn = $group.Group[0].checkpoint_lsn
+                            $obj.LastLsn = $group.Group[0].Last_Lsn
+                            $obj.SoftwareVersionMajor = $group.Group[0].Software_Major_Version
+                            $GroupResults += $obj
+                        }
+                        $results = $GroupResults | Sort-Object -Property End -Descending
```
ctrlbold: Could this just be `$GroupResults | Sort-Object -Property End -Descending` and we skip the foreach result in results?

functions/Get-DbaBackupHistory.ps1
```diff
@@ -1,449 +1,441 @@
...
+	logical_name as LogicalName,
+	physical_name as PhysicalName
+from
+    msdb.dbo.backupfile
+where
+    backup_set_id='$($Group.group[0].BackupSetID)'
+"@
+                            Write-Message -Level Debug -Message "FileSQL: $FileSql"
+                            
+                            $obj = New-Object sqlcollective.dbatools.Database.BackupHistory
```
ctrlbold: mmm i dont want to get people in the habit of doing generic $obj. can you name this $history or something? merci!

FriedrichWeinmann: Done, truly, a bad name in general.
Used to it because I myself always use it in a very restricted situation (Exactly as here: filling in properties on a result object before passing it to an array). Personal habit, so I didn't consider it on a project-wide scale.

functions/Get-DbaBackupHistory.ps1
FriedrichWeinmann: done. Didn't touch the original loop when I should have read more carefully.

ConstantineK:
Going through the code now, but I had a small piece of meta-feedback that I think may be relevant for pull requests as we move more towards a "dbatools framework" (instead of a bag of functions.) 

I would prefer to draw a brighter line between framework changes and function changes (and have them in separate PRs). This has a few benefits around testing and time to review, and keeps changes thematically separate.

If the work was done all at once (like in this case) and we want to submit both right away (instead of submit framework change, get accepted, now use that change in your code), we can use git cherry-pick to separate the commits and note the dependency in the PR submission (and submitting the framework one first 😆 ). 

This would require the user do an additional merge (if there is any code changes to the framework items) or even refactor the child function, so really its user's choice of how they want to "chain" pull requests.

internal/configurations/formatting.handler.ps1
```diff
@@ -0,0 +1,121 @@
...
+    Param (
+        [string]
+        $Value
+    )
+    
+    $Result = New-Object PSOBject -Property @{
+        Success = $True
+        Message = ""
+    }
+
```
FriedrichWeinmann: Well ... given that this is a hashtable declaration ... yes, there is a reason:
It's a technical requirement.
I could work around this by escaping the linebreak symbol (that's what the final backtick in a line does), but that's not neat and I really don't like people confusing Scriptblock/Hashtable declarations with braces denoting loops or scope-boundaries (such as when declaring a function). There's a vast technical difference right there.

Either way, the property-parameter expects a hashtable, so I give it one and start specifing it right behind it, which is where PowerShell expects values bound to a parameter.

bin/library.ps1
```diff
@@ -1328,118 +1498,2224 @@ namespace sqlcollective.dbatools
...
+            }
+
+            /// <summary>
+            /// 
+            /// </summary>
+            /// <param name="obj"></param>
+            /// <returns></returns>
+            public bool Equals(TimeSpan obj)
+            {
+                re
```
FriedrichWeinmann: Applied your update - stylistically a lot superior to my version, so I'm all in favor-

FriedrichWeinmann: Yeah, it's more concise and really the better solution. Please forgive me the comma error - I'm a German and thus handicapped where commas are concerned ;)
Done

PR #994, opened 2017-03-27
Test-DbaValidLogin, fixes #987

### LINKED ISSUE #987: Test-DbaValidLogin, incorrect parsing of groups
### bug

When the sqlserver passed is not in the format  sqlserver\instance, an incorrect filter is applied on groups.

I'm going to raise a PR with a fix soon, incorporating this fix and https://github.com/sqlcollaborative/dbatools/pull/953 and the new logging system.

niphlod:
plus a big refactor

Fixes #987

 - uses Get-DbaADObject instead of its own routine
 - changes ExcludeDomains to IgnoreDomains (for some reason it clashed with -exclude dynamicparam)
 - implements the new logging system
- it's generally faster
- normalizes everything to the netbios notation _every dba is accustomed to_  (e.g. for people having disjoint domains, that are the ones that have an ldap path representation that doesn't match the nebios one)
- treats include,exclude and ignoredomains IMHO in a more consistent way
- filters better the logins that can't be validated

How to test this code: 
- [ ] 
- [ ] 

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally

### DIFF
--- a/functions/Test-DbaValidLogin.ps1
+++ b/functions/Test-DbaValidLogin.ps1
@@ -5,7 +5,7 @@ Function Test-DbaValidLogin
 Test-DbaValidLogin finds any logins on SQL instance that are AD logins with either disabled AD user accounts or ones that no longer exist
 
 .DESCRIPTION
-The purpose of this function is to find SQL Server logins that are used by active directory users that are either disabled or removed from the domain. It allows you to 
+The purpose of this function is to find SQL Server logins that are used by active directory users that are either disabled or removed from the domain. It allows you to
 keep your logins accurate and up to date by removing accounts that are no longer needed.
 
 .PARAMETER SQLServer
@@ -23,11 +23,14 @@ Excludes any login you pass into it from the results.
 .PARAMETER FilterBy
 By default the function returns both Logins and Groups. you can use the FilterBy parameter to only return Groups (GroupsOnly) or Logins (LoginsOnly)
 
-.PARAMETER ExcludeDomains
-By default we traverse all domains in the forest and all trusted domains. You can exclude domains by adding them to the ExcludeDomains
+.PARAMETER IgnoreDomains
+By default we traverse all domains in the forest and all trusted domains. You can exclude domains by adding them to the IgnoreDomains
 
 .PARAMETER Detailed
 Returns a more detailed result, showing if the login on SQL Server is enabled or disabled and what type of account it is in AD
+
+.PARAMETER Silent
+Use this switch to disable any kind of verbose messages
 
 .NOTES
 Author: Stephen Bennett: https://sqlnotesfromtheunderground.wordpress.com/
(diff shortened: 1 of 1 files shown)


niphlod:
I'm unsure (on top of the usual things) about the -Detailed switch on this, what should we do there ?

PowerDBAKlaas:
@niphlod I guess we best go the Select-Defaultview way wherever the difference is only in the output selection.
If it makes a difference in query logic or any other amount of work the function will perform, -Detailed would be a valid switch, as we can speed up the function by skipping steps when not supplying it.
I think we need to make that decision when other alterations are made and the definitive extent of this function becomes clear.
There are indeed more improvements to make here, but I too ignored them up to now since the focus is on bug fixing.

niphlod:
what kind of improvements ? 

in this case the detailed just does a "isenabled" property enumeration, so I guess select-defaultview is the way to go. but the disappearance of the -detailed switch would effectively break compatibility so.... how to manage the in-between ? :D

PowerDBAKlaas:
There was a card on Trello somewhere about this one, but I can't find it.
I recall asking for PasswordIsExpired.
[ADLoginDetails] should be replaced with the individual properties. They can be collected by binary 'and' against the useraccountcontrol, which results in the 22 properties included in this binary string, in only 1/4 of the code.
It would be a waste to go through AD, and then report only some properties, and needing another function to go back and find the other properties. In other words, we could Test-DbaValidLogin make us believe our account is perfect, and still having troubles with it because the password is expired.

niphlod:
useraccountcontrol is already checked here. it goes to AD just once .

PowerDBAKlaas:
well yes, but only a few properties are collected and then 2, 3 or 4 properties are pasted together into one string. So we could have all 22 with the same effort, but we don't.

niphlod:
I *think* the 17 we have cover all possible cases for users and groups out of the 22 totally documented, some are only relative to domain controllers. 
So the complete scenario is covered, the real "problem" is the string representation of it ?
i.e. you'd like to have separate "enabled", "smartcardrequired", "passwordrequired","passwordexpires" properties ?

niphlod:
btw, I can totally leverage this script https://blogs.msdn.microsoft.com/muaddib/2014/03/30/powershell-script-to-query-useraccountcontrol-flags/

PowerDBAKlaas:
Yes, that's it. Exactly, except for the Write-Host part, of course 😄 
But then the next question: would it be better to take this part out and make a separate function, that will be called in here, but maybe also in a Get- and other functions?
Indeed, that composed string bothers me most at the moment.

niphlod:
Imho the current is human-good.
We can use additional raw properties for computer-good, the cost is
entirely negligible



Il 28 mar 2017 1:26 PM, "PowerDBAKlaas" <notifications@github.com> ha
scritto:

niphlod:
ok @PowerDBAKlaas , tell me what you want. We can have the script returning these constants as an array, so people can filter with them ?

```
$MappingRaw = @{
	1        = 'SCRIPT' 
	2        = 'ACCOUNTDISABLE'
	8        = 'HOMEDIR_REQUIRED'
	16       = 'LOCKOUT'
	32       = 'PASSWD_NOTREQD'
	64       = 'PASSWD_CANT_CHANGE' 
	128      = 'ENCRYPTED_TEXT_PASSWORD_ALLOWED'
	256      = 'TEMP_DUPLICATE_ACCOUNT'
	512      = 'NORMAL_ACCOUNT'
	2048     = 'INTERDOMAIN_TRUST_ACCOUNT'
	4096     = 'WORKSTATION_TRUST_ACCOUNT'
	8192     = 'SERVER_TRUST_ACCOUNT'
	65536    = 'DONT_EXPIRE_PASSWD'
	131072   = 'MNS_LOGON_ACCOUNT'
	262144   = 'SMARTCARD_REQUIRED'
	524288   = 'TRUSTED_FOR_DELEGATION'
	1048576  = 'NOT_DELEGATED'
	2097152  = 'USE_DES_KEY_ONLY'
	4194304  = 'DONT_REQUIRE_PREAUTH'
	8388608  = 'PASSWORD_EXPIRED'
	16777216 = 'TRUSTED_TO_AUTHENTICATE_FOR_DELEGATION'
	67108864 = 'PARTIAL_SECRETS_ACCOUNT'
}
$MappingHuman = @{
	1        = 'The logon script is executed'
	2        = 'The user account is disabled'
	8        = 'The home directory is required'
	16       = 'The account is currently locked out'
	32       = 'No password is required'
	64       = 'The user cannot change the password'
	128      = 'The user can send an encrypted password'
	256      = 'This is an account for users whose primary account is in another domain. (AKA local user account)'
	512      = 'This is a default account type that represents a typical user'
	2048     = 'This is a permit to trust account for a system domain that trusts other domains'
	4096     = 'This is a computer account for a Windows or Windows Server that is a member of this domain'
	8192     = 'This is a computer account for a system backup domain controller that is a member of this domain'
	65536    = 'The password for this account will never expire'
	131072   = 'This is an Majority Node Set (MNS) logon account'
	262144   = 'The user must log on using a smart card'
	524288   = 'The service account, under which a service runs, is trusted for Kerberos delegation. Any such service can impersonate a client requesting the service'
	1048576  = 'The security context of the user will not be delegated to a service even if the service account is set as trusted for Kerberos delegation'
	2097152  = 'Restrict this principal to use only Data Encryption Standard (DES) encryption types for keys'
	4194304  = 'This account does not require Kerberos preauthentication for logon'
	8388608  = 'The user password has expired'
	16777216 = 'The account is enabled for delegation'
	67108864 = 'Computer account for a read-only domain controller (RODC)'
}
```

PowerDBAKlaas:
I would like the relevant settings as individual properties set to $true or $false:
PasswordRequired:  True
PasswordExpired:     False
TrustedForDelegation: True
PasswordNeverExpires: True
...

Get-ADUSer returns them more strictly translated:
PasswordNotRequired:  False
DoesNotRequirePreAuth: False

It may be more consistent to stick to that. On the other hand I dislike those double negations.
What's your opinion on this, @niphlod ?


The ones you found exclusively for DC's we can leave out.
Filtering and sorting are indeed useful extra functionalities we lose when combining multiple settings in one sentence.

niphlod:
chicken and egg problem.

These are the bool switches available for set-aduser
```
-AccountNotDelegated
-AllowReversiblePasswordEncryption
-CannotChangePassword
-ChangePasswordAtLogon
-Enabled
-PasswordNeverExpires
-PasswordNotRequired
-SmartcardLogonRequired
-TrustedForDelegation
```

would that satisfy your requirements ? IMHO it's more usable to check for a `status -contains 'PASSWORD_EXPIRED'` rather than having 9 different props with true and false

PowerDBAKlaas:
I think that list will do. It'll be more recognizable for those used to work with AD.

I don't agree with the contains; a user would need to know what to look for in advance or explore what possible values can be in there and how they are spelled out.
One property should hold one atomic piece of information. You wouldn't let Get-DbaDatabaseState return "Online, read_only and multi_user" as one property,right? Or "0.825 MB Used and 1725.25MB Free" as DatabaseSpace,...

niphlod:
correct. ok, will implement the aforementioned properties as hidden bools.

PS: @ctrlbold as this is probably "breaking backward compatibility" both for the arguments and the output, remember that docs needs proper care before putting whatever will be the result in master.

wsmelton:
Overall it sounds like yall are trying to turn this into something the AD module already does. Remember that on average a DBA is not going to have permissions to query a production AD (I don't in most environments I support).

You might be able to get the basic information (what's under the general tab) of a user but not sure about all those other properties. In testing we just need to test the permission issue and how that effects the function. We just want to ensure the base requirement of the command is still meet. I don't have my lab domain to see what (by default) a basic domain user can query.

I'm also in agreement on the details property, it should be individual.

wsmelton:
Ok, so did find [defaults in AD 2008+](http://techgenix.com/active-directory-information-exposed-users/) are pretty open but I have hit domains that they locked it down like they should.

PowerDBAKlaas:
So we need some extra checks and make sure we don't mix up "access denied" and "user does not exist".
@niphlod  will Get-DbaADObject catch that?

niphlod:
yep, it was one of the original bugs reported on this cmdlet and has already been patched (basically, the 'unknown' logic, line 150 and following for example). Guess that those bool would then be $null.

PowerDBAKlaas:
Great!

But I think booleans don't accept $null

niphlod:
"bools" as in these props
```
-AccountNotDelegated
-AllowReversiblePasswordEncryption
-CannotChangePassword
-ChangePasswordAtLogon
-Enabled
-PasswordNeverExpires
-PasswordNotRequired
-SmartcardLogonRequired
-TrustedForDelegation
```
the value for those would be:
-  $null if we can't access the useraccountcontrol property of the ad object 
OR 
- either $true or $false depending on the said value

niphlod:
done, please check

functions/Test-DbaValidLogin.ps1
```diff
@@ -281,42 +250,41 @@ Tests all logins excluding any that are from the subdomain Domain
...
-						Found = $exists
-						Enabled = $enabled
-						DisabledInSQLServer = $login.IsDisabled
-						ADLoginDetails = 'AD group'
-					}
+				$rtn = [PSCustomObject]@{
+					Server = $server.DomainInstanceName
+					Domain = $domain
+					Login = $groupname
+					Type = "Group"
```
PowerDBAKlaas: We'll better rename some properties to make clear we're talking about AD.
Maybe
[FoundInAD] or [ExistsInAD]?
[ADType]

functions/Test-DbaValidLogin.ps1
```diff
@@ -281,42 +250,41 @@ Tests all logins excluding any that are from the subdomain Domain
...
-						Domain = $domain
-						Login = $groupname
-						Type = "Group"
-						Found = $exists
-						Enabled = $enabled
-					}
+				if ($Detailed) {
+					Select-DefaultView -InputObject $rtn -ExcludeProperty UserAccountControl
+				} else {
+					Select-DefaultView -InputObject $rtn -ExcludeProperty UserAccountControl,AccountNotDelegated,AllowReversiblePasswordEncryption,CannotChangePassword,PasswordNeverExpires,SmartcardLogonRequired,TrustedForDelegation
```
PowerDBAKlaas: I would also like to rename the function itself, to make clear we check AD.
A login with sql authentication can be a 'valid login' too, so I think '..AD...' should be somewhere in the function name.
Do you have any suggestions?

niphlod:
I think once it's named "Test-DbaValidADLogin" the "exist" part should be clear.

ctrlbold:
Thanks a metric ton to you all! 👍

PR #1022, opened 2017-04-09
Fix issue 1008 - Add SizeMB property to default of Get-DbaDatabase

### LINKED ISSUE #1008: Add to default output of Get-DbaDatabase

feature asap

### System Details
please add size in MB to default output of Get-DbaDatabase

wsmelton:
Fixes #1008 

 - Added "SizeMB" to the default view, the Size property defaults to MB units
 - Added messaging functions (Stop-Function) where appropriate to replace Write-Warning
 - Cleaned up where-object to remove `$_.property` references, makes it easier to read

How to test this code: 
- [ ] 
- [ ] 

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally

### DIFF
--- a/functions/Get-DbaDatabase.ps1
+++ b/functions/Get-DbaDatabase.ps1
@@ -135,36 +135,36 @@ Returns databases on multiple instances piped into the function
             }
 
             if ($NoUserDb) {
-                $inputobject = $server.Databases | Where-Object { $_.IsSystemObject }
+                $inputobject = $server.Databases | Where-Object IsSystemObject
             }
 
             if ($NoSystemDb) {
-                $inputobject = $server.Databases | Where-Object { $_.IsSystemObject -eq $false }
+                $inputobject = $server.Databases | Where-Object IsSystemObject -eq $false
             }
 
             if ($databases) {
-                $inputobject = $server.Databases | Where-Object { $_.Name -in $databases }
+                $inputobject = $server.Databases | Where-Object Name -in $databases
             }
 
             if ($status) {
-                $inputobject = $server.Databases | Where-Object { $_.Status -eq $status }
+                $inputobject = $server.Databases | Where-Object Status -eq $status
             }
 
             if ($Owner) {
-                $inputobject = $server.Databases | Where-Object { $_.Owner -in $Owner }
+                $inputobject = $server.Databases | Where-Object Owner -in $Owner
             }
 
             switch ($Access) {
-                "ReadOnly" { $inputobject = $server.Databases | Where-Object { $_.ReadOnly } }
-                "ReadWrite" { $inputobject = $server.Databases | Where-Object { $_.ReadOnly -eq $false } }
+                "ReadOnly" { $inputobject = $server.Databases | Where-Object ReadOnly }
+                "ReadWrite" { $inputobject = $server.Databases | Where-Object ReadOnly -eq $false }
             }
 
             if ($Encrypted) {
-                $inputobject = $server.Databases | Where-Object { $_.EncryptionEnabled }
+                $inputobject = $server.Databases | Where-Object EncryptionEnabled
             }
 
             if ($RecoveryModel) {
-                $inputobject = $server.Databases | Where-Object { $_.RecoveryModel -eq $RecoveryModel }
+                $inputobject = $server.Databases | Where-Object RecoveryModel -eq $RecoveryModel
             }
 
             # I forgot the pretty way to do this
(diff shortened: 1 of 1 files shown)


functions/Get-DbaDatabase.ps1
```diff
@@ -74,181 +76,152 @@ Returns only the system databases on the local default SQL Server instance
...
+    )
+
+    DynamicParam { if ($SqlInstance) { return Get-ParamSqlDatabases -SqlServer $SqlInstance[0] -SqlCredential $SqlCredential } }
+
+    BEGIN {
+        $databases = $psboundparameters.Databases
+        $exclude = $psboundparameters.Exclude
+
+        if ($NoUserDb -and $NoSystemDb) {
+            Stop-Function -Message "You cannot specify both NoUserDb and NoSystemDb" -Continue -Silent $Silent
```
FriedrichWeinmann: Calling the -Continue parameter without a loop will break the calling function.
Instead use like this:
```
Stop-Function -Message "You cannot specify both NoUserDb and NoSystemDb"
return
```
Also read first comment, which is related to this.

functions/Get-DbaDatabase.ps1
```diff
@@ -74,181 +76,152 @@ Returns only the system databases on the local default SQL Server instance
...
+        }
+    }
+
+    PROCESS {
```
FriedrichWeinmann: Add a line here with the following:
This will detect whether Stop-Function was called in the previous block and gracefully terminate the process block as well.

functions/Get-DbaDatabase.ps1
```diff
@@ -74,181 +76,152 @@ Returns only the system databases on the local default SQL Server instance
...
+        foreach ($instance in $SqlInstance) {
+            try {
+                $server = Connect-SqlServer -SqlServer $instance -SqlCredential $sqlcredential
+            }
+            catch {
+                Stop-Function -Message "Failed to connect to: $instance" -Continue -Silent $Silent
```
FriedrichWeinmann: To improve troubleshooting, add:
`-InnerErrorRecord $_ -Target $instance`
This will include the original exception and the instance for which it failed to the error log

functions/Get-DbaDatabase.ps1
wsmelton: Debugging this code this can actually be removed because this if statement can never be hit since parameter sets are being used.

Just run `Get-DbaDatabase -SqlInstance MyServer -NoUserDb -NoSystemDb` returns the error:

> Parameter set cannot be resolved using the specified named parameters.

So if we want a more robust message to be displayed then the parameter sets would have to be removed.

wsmelton:
What section of the wiki are you referring to that notes how much verbosity or comments are needed in a given function? I'm indifferent on this one since I did not write it, just adding the property Chrissy wanted.

In reference to the verbose and comments from my opinion...most of the `Get-xxx` functions we have in this module are basic-level in reference to complexity. They are simply returning an object from SMO or specific properties, so none of them really have much in the way of verbose messages or a single comment. So if we are changing the rules on this then a specific level/amount of verbose and comments need to be explained and applied to all. [If this is all in the wiki and I just don't see it then this is a mute point.]

functions/Get-DbaDatabase.ps1
FriedrichWeinmann: Aye. That's an either-or thing. Guess that needs to be escalated to @ctrlbold as it's a major design philosophy question (both paths are technically valid)

FriedrichWeinmann:
Here's the link to the wiki entry. It comes with a full example, if there are any ambiguities, please point them out. As usual, from my perspective everything should be clear, but then the author might take some things for granted. Second and third opinmions most welcome:
https://github.com/sqlcollaborative/dbatools/wiki/Implementing-the-messaging-system

functions/Get-DbaDatabase.ps1
wsmelton: Added in last commit.

functions/Get-DbaDatabase.ps1
wsmelton: Added in last commit.

potatoqualitee:
This was awesome! Works well, thank you so much @wsmelton! Thank you also for the review, @FriedrichWeinmann

PR #1023, opened 2017-04-10
Update issues template with examples and more info

wsmelton:
 - Add some examples to the issue template to try and get folks attention, so they would add more details.

If this is to much/overboard we can scrap it.


ctrlbold:
Looks good, thank you!

PR #1036, opened 2017-04-11
Get-DbaLogin command 

### LINKED ISSUE #945: Get-DbaLogin - needed for 1.0

feature

### System Details

We need Get-DbaLogin like the other basic commands. Please add the last login time to SMO object and the default output

```
SELECT MAX(login_time) AS [Last Login Time], login_name [Login]
FROM sys.dm_exec_sessions where login_name = '$($login.name)'
```

per https://mohammaddarab.com/find-last-login-date-sql-server-login/

mshamann:
Fixes #945 

Tested on Windows 10, Windows Server 2016, SQL 2008R2, SQL 2016 SP1 and powershell 5.1

 - Adds Get-DbaLogin cmdlet to get the logins. 
 - Returns smo object to the piplines

How to test this code: 
- [ ] Get-DbaLogin -SqlInstance $sql2016-SqlCredential $sqlcred -Logins dbatoolsuser,thecaptain | format-table
- [ ] Get-DbaLogin -SqlInstance sql2016 -SqlCredential $sqlcred  | format-table

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally


ctrlbold:
Hell yeah! Thanks @SirCaptainMitch!

mshamann:
heh, sorry it took a while. I kept having to divert to other projects.

functions/Get-DbaLogin.ps1
```diff
@@ -0,0 +1,135 @@
...
+{
+<#
+.SYNOPSIS 
+Function to get an SMO login object of the logins for a given SQL Instance. Takes a server object from the pipe 
+
+.DESCRIPTION
+The Get-DbaLogin function returns an SMO Login object for the users passed, if there are no users passed it will return all logins.  
+
+.PARAMETER SqlInstance
+The SQL Server instance.You must have sysadmin access and server version must be SQL Server version 2000 or higher.
```
PowerDBAKlaas: here it is worth mentioning the user can submit multiple instances.

functions/Get-DbaLogin.ps1
```diff
@@ -0,0 +1,135 @@
...
+
+.PARAMETER SqlCredential
+Allows you to login to servers using SQL Logins as opposed to Windows Auth/Integrated/Trusted. 
+
+.PARAMETER Logins
```
PowerDBAKlaas: this parameter should be singular, even though it can handle multiple logins.

functions/Get-DbaLogin.ps1
```diff
@@ -0,0 +1,135 @@
...
+The SQL Server instance.You must have sysadmin access and server version must be SQL Server version 2000 or higher.
+
+.PARAMETER SqlCredential
+Allows you to login to servers using SQL Logins as opposed to Windows Auth/Integrated/Trusted. 
+
+.PARAMETER Logins 
+Pass a single login, or a list of them. Comma delimited. 
+
+.PARAMETER NoSystemLogins
+A switch to exlude system logins from returning. (Logins with '##' in the front or the 'sa' login)
```
PowerDBAKlaas: we have to rethink the definition and name of this switch: ## are not system logins.

functions/Get-DbaLogin.ps1
```diff
@@ -0,0 +1,135 @@
...
+
+.PARAMETER Detail
```
PowerDBAKlaas: the parameter name is actually -Detailed

functions/Get-DbaLogin.ps1
```diff
@@ -0,0 +1,135 @@
...
+.EXAMPLE   
+Get-DbaLogin -SqlInstance SQl2016 -SqlCredential $sqlcred 
+Gets all the logins for a given SQL Server using a passed credential object and returns the SMO login objects 
+
+.EXAMPLE 
+Get-DbaLogin -SqlInstance SQl2016 -SqlCredential $sqlcred -Logins dbatoolsuser,TheCaptain 
+Get specific user objects from the server
+
+.EXAMPLE 
+@('sql2016', 'sql2014') |  Get-DbaLogin -SqlCredential $sqlcred
```
PowerDBAKlaas: 'sql2016','sql2014' |
 is enough, no @() required

functions/Get-DbaLogin.ps1
wsmelton: This would also be a dynamic parameters unless we are going to stop doing that now.... @ctrlbold

PowerDBAKlaas: @wsmelton , it was actually dynamic param, but @mshamann took that out already on our request.

functions/Get-DbaLogin.ps1
wsmelton: Neither are "NT SERVICE\xxx" but are we going to exclude those as well?

PowerDBAKlaas: true,

if we use -NoSystem... it should really mean the systemobjects. So the switch can stay, but the statement should be `Where -not IsSystemObject` instead of the ## and 'sa'

we probably better leave all other excludes to the user with a 
`Get-DbaLogin | where ...`

functions/Get-DbaLogin.ps1
```diff
@@ -0,0 +1,135 @@
...
+	
+#>
+	[CmdletBinding()]
+	Param (
+		[parameter(Position = 0, Mandatory = $true, ValueFromPipeline = $true)]
+		[Alias("ServerInstance", "SqlServer")]
+		[object[]]$SqlInstance,
+		[System.Management.Automation.PSCredential]$SqlCredential,
+		[Object[]]$Logins,
+		[Switch]$NoSystemLogins,
```
PowerDBAKlaas: sorry, @mshamann , @ClaudioESSilva just caught an important issue:
the code I suggest above is PS V4, and we only require V3, so it actually has to be 
`$serverlogins = $serverlogins | where {...`

same goes for the other suggestion below.

mshamann: I think, in this case we can just remove the -nosystem and let the end user do that. I think it removes a lot of needless complexity in determining what counts as a "system" object. 

otherwise, do we add another switch for NoService accounts? 

I see it being much better facilitated by saying "You can just filter it with a Where-Object { .... " 

What do you think @ctrlbold @PowerDBAKlaas @wsmelton ?

wsmelton: I would agree on dropping the switches for filtering of system objects, although we could look at adding switch parameters similar to the `Get-SqlLogin` command:

- Disabled
- Locked
- HasAccess

Then add a LoginType parameter, which for the `Get-SqlLogin` this seems to have the following set of values:
AsymmetricKey, Certificate, ExternalGroup, ExternalUser, SqlLogin, WindowsGroup, WindowsUser

mshamann: Agreed, flush the NoSystem, and filter off SMO properties instead. I can also add the login type.

SirCaptainMitch:
Refactored based on Review.

functions/Get-DbaLogin.ps1
mshamann: @PowerDBAKlaas i'm not sure why this part isn't clearing, but I did fix this.

functions/Get-DbaLogin.ps1
PowerDBAKlaas: Hmmmm, I don't like this type of switches very much. There are so many options. Now we have a -Disabled switch, but what if I only want to see the enabled ones? Or the 'Not locked', 'has no access', 'system', 'no system', ,...

I still have to know how to use `| where `, and it's actually not more work than to look up which switches are available.
Furthermore, there's a more complex logic needed, @SirCaptainMitch . What if multiple switches are supplied? Now only the first one is applied. That's not correct.
And you always have to be careful what switches exclude each other, and catch that too.

I'm in favour of dropping them all. @wsmelton ?

potatoqualitee:
Awesome work!! Thank you, @mshamann! This command will be included in the next release 🤗

PR #1041, opened 2017-04-13
Export-DbaJob - Can be used as template for other exports

ctrlbold:


ctrlbold:
@PowerDBAKlaas made a good point, I'm just gonna create a universally pipable object.

PR #1046, opened 2017-04-13
Export-DbaServerAudit

ctrlbold:


wsmelton:
Does this need to be named DbaSqlAudit? Because it only exports the audits at the server level. What about the audits at the database level? 

As stated in the [docs for audits](https://docs.microsoft.com/en-us/sql/relational-databases/security/auditing/sql-server-audit-database-engine) 

```SQL Server audit lets you create server audits, which can contain server audit specifications for server level events, and database audit specifications for database level events.```

ctrlbold:
Yep! We had this convo for Garry's command and I forgot. Will change thank you!

ctrlbold:
@PowerDBAKlaas made a good point, I'm just gonna create a universally pipable object.

PR #1052, opened 2017-04-15
Test-DbaSpn, accomodate get-dbaadobject

### LINKED ISSUE #936: Test-DbaSpn and Get-DbaSpn, minor bug in domain restrictions

niphlod:
Fixes #936 

Additional PRs in the next few days to make every spn cmdlet use get-dbaadobject

While rewriting it, I used the new messaging system and spotted a few inconsintencies between the documented use of parameters and their mandatory usage.

Also, can the original authors/testers vouch for the $domain parameter ? I found it to be unused and, generally, less than well documented in the way it's used down the code. Maybe it should disappear.

### DIFF
--- a/functions/Test-DbaSpn.ps1
+++ b/functions/Test-DbaSpn.ps1
@@ -1,3 +1,4 @@
+#ValidationTags#FlowControl,Pipeline#
 Function Test-DbaSpn
 {
 <#
@@ -23,11 +24,15 @@ or not.
 The server name you want to discover any SQL Server instances on. This parameter is required.
 
 .PARAMETER Credential
-The credential you want to use to connect to the remote server and active directory. This parameter is required.
+The credential you want to use to connect to the remote server and active directory.
 
 .PARAMETER Domain
 If your server resides on a different domain than what your current session is authenticated against, you can specify a domain here. This
 parameter is optional.
+
+.PARAMETER Silent
+Use this switch to disable any kind of verbose messages
+
 
 .NOTES
 Tags: SQLWMI, SPN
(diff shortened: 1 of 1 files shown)


niphlod:
here too, too fast. if $domain isn't needed, it must be cut out.

potatoqualitee:
You can do a combo PR if you like

niphlod:
mwhahaha, ok, I need to submit a bunch for the other cmdlets, too.

PR #1053, opened 2017-04-15
Get-DbaSpn, accomodate get-dbaadobject

### LINKED ISSUE #936: Test-DbaSpn and Get-DbaSpn, minor bug in domain restrictions

niphlod:
Fixes #936

Additional PRs in the next few days to make every spn cmdlet use get-dbaadobject

While rewriting it, I used the new messaging system and spotted a few inconsintencies between the documented use of parameters and their mandatory usage.

Also, can anyone vouch for the usage of computername with the ending $ ? 
What's the whole deal behind Process-Account and the ByAccount switch ??
I can't figure out both what ByAccount changes in the flow AND why is it there in the first place, but I may be tired

### DIFF
--- a/functions/Get-DbaSpn.ps1
+++ b/functions/Get-DbaSpn.ps1
@@ -21,6 +21,9 @@ User credential to connect to the remote servers or active directory.
 .PARAMETER ByAccount
 Shows all SPNs registered by the specified AccountName otherwise, only results will be shown for the specified ComputerName (which is localhost by default)
 
+.PARAMETER Silent
+Use this switch to disable any kind of verbose messages
+
 .NOTES
 Tags: SPN
 Author: Drew Furgiuele (@pittfurg), http://www.port1433.com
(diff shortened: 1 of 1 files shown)


potatoqualitee:
hey @drewfurgiuele - you got any time for this?

drewfurgiuele:
Review added. Good stuff.

potatoqualitee:
Thank yall sooooooooo much!

niphlod:
huh, not so fast .... there are some pieces missing then (the ByAccount bit)

potatoqualitee:
i was actually gonna merge, test then patch. well, ask for a patch ;) Can you grab that real quick? I'd like to do a release in an hour or three

niphlod:
in two I'll have reached a developing station. till then, no joy.

potatoqualitee:
np! just sittin, eatin sushi and listening to sander's presso ;)

Issue #1013, opened 2017-04-04
Export-SqlUser does script permissions under system databases


ClaudioESSilva:
### Is this a *bug*:
If we want to export user permissions on a specific system database like msdb, this command does not goes on those databases.

- SQL Server version for source/target
All

### Steps to Reproduce
Try to execute the command through msdb database

### Action Results
Will not script out the system databases permissions for the user(s).

### Expected Results
Permissions under system databases should be scripted

ben-thul:
I already have a proposed change to this file (for another enhancement). I propose adding an optional parameter that allows the inclusion of system databases. Currently, they are _explicitly_ excluded.

ClaudioESSilva:
@ben-thul thanks for your update.

Which another enhancement are you talking about? 😄 

The will be add the same experience on the other commands, allow all databases and people can use -Exclude to add all that they don't want.

ben-thul:
If there are already plans in the works for something more systemic, I'll wave off of the change in re: system databases. It's only in my local repo for now and was a trivial change.

In regards to the other change, the current script assumes that securables in the database are schemas, users, or objects. That assumption, however, doesn't produce the correct DDL for all ObjectClasses (as defined [here](https://msdn.microsoft.com/en-us/library/microsoft.sqlserver.management.smo.objectclass.aspx)). The change I have extends the function to deal with most of the object types provided in the MSDN link.

ClaudioESSilva:
@ben-thul don't "wave off" 😄 we will be glad that you submit a PR with those changes 😉 
Submit a PR and we can do a code review 👍 

Regarding the other changes, which ones are you including? 
A quick reminder, this is only for User (not Login) which means database level.

ben-thul:
Sure thing. I just did the coding last night, but I've gotta test it yet. I'll probably get to it over the weekend. Also, acknowledged about this being for database-scoped permissions. As for the list, I added support for all of the following:

- ApplicationRole
- AsymmetricKey
- AvailabilityGroup
- Certificate
- DatabaseRole
- FullTextCatalog
- FullTextStopList
- MessageType
- ObjectOrColumn (added support for column-level permissions here)
- RemoteServiceBinding
- SearchPropertyList
- Service
- ServiceContract
- ServiceRoute
- SqlAssembly
- SymmetricKey
- UserDefinedType
[...]

PR #1054, opened 2017-04-18
Added support for more object types, system dbs.

ben-thul:
Added support for many more object types to export-sqluser.

Also addressed Issue #1013

### DIFF
--- a/functions/Export-SqlUser.ps1
+++ b/functions/Export-SqlUser.ps1
@@ -96,8 +96,7 @@ https://dbatools.io/Export-SqlUser
 		[System.Management.Automation.PSCredential]$SqlCredential,
 		[Alias("NoOverwrite")]
 		[switch]$NoClobber,
-		[switch]$Append,
-		[switch]$IncludeSystemDatabases = $false
+		[switch]$Append
 	)
 	
 	DynamicParam 
@@ -154,7 +153,7 @@ https://dbatools.io/Export-SqlUser
 
         if ($databases.Count -eq 0)
         {
-            $databases = $sourceserver.Databases | Where-Object {$exclude -notcontains $_.Name -and $_.IsSystemObject -eq $IncludeSystemDatabases -and $_.IsAccessible -eq $true}
+            $databases = $sourceserver.Databases | Where-Object {$exclude -notcontains $_.Name -and $_.IsAccessible -eq $true}
         }
         else
         {
@@ -165,7 +164,7 @@ https://dbatools.io/Export-SqlUser
 		    }
             else
             {
-                $databases = $sourceserver.Databases | Where-Object {$exclude -notcontains $_.Name -and $_.IsSystemObject -eq $IncludeSystemDatabases -and $_.IsAccessible -eq $true -and ($databases -contains $_.Name)}
+                $databases = $sourceserver.Databases | Where-Object {$exclude -notcontains $_.Name -and $_.IsAccessible -eq $true -and ($databases -contains $_.Name)}
             }
         }
 
@@ -182,8 +181,9 @@ https://dbatools.io/Export-SqlUser
                 }
                 else
                 {
-                    $scriptVersion = $versions[$DestinationVersion]
+                    $scriptVersion = $versions[$DestinationVersion]   
                 }
+                $versionNameDesc = $versionName[$scriptVersion.ToString()]
 
                 #Options
                 [Microsoft.SqlServer.Management.Smo.ScriptingOptions] $ScriptingOptions = New-Object "Microsoft.SqlServer.Management.Smo.ScriptingOptions";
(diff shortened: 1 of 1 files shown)


functions/Export-SqlUser.ps1
```diff
@@ -96,7 +96,8 @@ https://dbatools.io/Export-SqlUser
 		[System.Management.Automation.PSCredential]$SqlCredential,
 		[Alias("NoOverwrite")]
 		[switch]$NoClobber,
-		[switch]$Append
+		[switch]$Append,
+		[switch]$IncludeSystemDatabases = $false
```
ClaudioESSilva: I vote for $true.
Thinking on users that have msdb roles (for example)

functions/Export-SqlUser.ps1
```diff
@@ -153,7 +154,7 @@ https://dbatools.io/Export-SqlUser
 
         if ($databases.Count -eq 0)
         {
-            $databases = $sourceserver.Databases | Where-Object {$exclude -notcontains $_.Name -and $_.IsSystemObject -eq $false -and $_.IsAccessible -eq $true}
+            $databases = $sourceserver.Databases | Where-Object {$exclude -notcontains $_.Name -and $_.IsSystemObject -eq $IncludeSystemDatabases -and $_.IsAccessible -eq $true}
```
ClaudioESSilva: Include means adding to existing one. Which means that if I say $IncludeSystemDatabases = $true I want to validate system databases along with user database.
The way the validation is done will show only system databases **or** user databases, never both.
imo, both should be the default.
What do you think?

functions/Export-SqlUser.ps1
```diff
@@ -250,50 +251,156 @@ https://dbatools.io/Export-SqlUser
...
 
-                                $outsql += "$($GrantDatabasePermission) $($DatabasePermission.PermissionType) TO [$($DatabasePermission.Grantee)] $WithGrant"
+                                $outsql += "$($GrantDatabasePermission) $($DatabasePermission.PermissionType) TO [$($DatabasePermission.Grantee)]$($WithGrant) AS [$($DatabasePermission.Grantor)];"
                             }
 
 
 	                        #Database Object Permissions
-                            foreach ($ObjectPermission in $db.EnumObjectPermissions() | Where-Object {@("sa","dbo","information_schema","sys") -notcontains $_.Grantee -and $_.Grantee -notlike "##*" -and $dbuser.Name -contains $_.Grantee})
+                            # NB: This is a bit of a mess for a couple of reasons
+                            # 1. $db.EnumObjectPermissions() doesn't enumerate all object types
```
ClaudioESSilva: Nice to know 😄 
Thanks for the comment.

functions/Export-SqlUser.ps1
potatoqualitee: Thank you so much, ben! 

In addition, a switch will auto default to $false so = $false is not needed :) But `[switch]$IncludeSystemDatabases = $false` would require users to do `-IncludeSystemDatabases` so this should change to -NoSystemDatabases if we go that route (i vote we do as well)

potatoqualitee: Oh, we'll also need the Help parameter to be added - currently, the Pester tests that check to ensure all params have documentation are failing.

functions/Export-SqlUser.ps1
potatoqualitee: Yes, thank you! The other commands that use this will need to be updated as well.

functions/Export-SqlUser.ps1
```diff
@@ -250,50 +251,156 @@ https://dbatools.io/Export-SqlUser
...
+                            # 2. Some (x)Collection types can have EnumObjectPermissions() called
+                            #    on them directly (e.g. AssemblyCollection); others can't (e.g.
+                            #    ApplicationRoleCollection). Those that can't we iterate the
+                            #    collection explicitly and add each object's permission.
+
+                            $obj_perms = @()
```
potatoqualitee: With regards to Performance, += rebuilds the object each time an item is added. It would be [a lot faster](https://foxdeploy.com/2016/03/23/coding-for-speed/) to use 

```
$perms = New-Object System.Collections.ArrayList
$null = $perms.Add($whatever)
```

functions/Export-SqlUser.ps1
ben-thul: @ClaudioESSilva, @potatoqualitee: Acknowledged and understood. I'm inclined to change the default to `$true`.

functions/Export-SqlUser.ps1
ben-thul: You're right of course. I'm thinking that reverting to the old behavior and then appending system databases (pending accessibility checks and all that) to the list should do the trick.

functions/Export-SqlUser.ps1
ben-thul: Oh boy... that's a can of worms. If the number of places is "more than one" (as you imply that it is), I vote that this be abstracted out into its own function within the project. DRY principle and all that. Thoughts?

functions/Export-SqlUser.ps1
ben-thul: An easy enough change. Thanks for the feedback!

potatoqualitee:
Excellent work @ben-thul and @ClaudioESSilva ! Merging.

PR #1055, opened 2017-04-18
Get-DBASqlBuildReference, added latest build infos

niphlod:
Just an update to include 2014SP2CU5 and 2014SP1CU12


potatoqualitee:
Thank you both!

PR #1061, opened 2017-04-19
Get-DbaBackupHistory - fixed missing attributes, made Last a full chain

potatoqualitee:
fixed missing attributes with the Lasts, made -Last the full chain. Could be prettier and pure sql but w/e


potatoqualitee:
uhh, nm this just broke, brb

PR #1076, opened 2017-04-20
added process to allow install of dbatools from git repo

ChocolateMonkey:
without destroying the relationship

Borrowed process idea from posh git install option
https://github.com/dahlbyk/posh-git

Fixes # 

 - install.ps1 checks for location of run, if contains .git then process will perform Import-Module based on local version of code, process will also modify $profile to reload that same code.

How to test this code: 
- [ ] clone repo
- [ ] cd to dbatools
- [ ] .\install.ps1

- [ ]  Powershell 3
- [x]  Windows 7

- [x]  Powershell 5

Tests for tester:
- [ ] Works on my machine
- [ ] existing function not affected


utils.ps1
```diff
@@ -0,0 +1,298 @@
+# Need this variable as long as we support PS v2
+$ModuleBasePath = Split-Path $MyInvocation.MyCommand.Path -Parent
+
```
wsmelton: All of the functions in this file need to be separated out into multiple PS1 files, we do not use a single file to hold mulitple functions anymore. The way these are being used would all be placed in the `dbatools\internal` directory.

dbatools.psm1
```diff
@@ -125,3 +125,5 @@ Set-Alias -Name Test-SqlDiskAllocation -Value Test-DbaDiskAllocation
 Set-Alias -Name Get-DiskSpace -Value Get-DbaDiskSpace
 Set-Alias -Name Get-SqlMaxMemory -Value Get-DbaMaxMemory
 Set-Alias -Name Set-SqlMaxMemory -Value Set-DbaMaxMemory
+
+. $PSScriptRoot\utils.ps1
```
wsmelton: This line should be removed. Once your functions are added to the appropriate folder there is other code that will handle importing them so you can call them.

dbatools.psd1
```diff
@@ -67,6 +67,7 @@
 	
 	# Functions to export from this module
 	FunctionsToExport = @(
+		'Add-DbaToolsToProfile',
```
wsmelton: While @potatoqualitee is the only individual that can alter the psd1 file I don't think this needs to be a public function. Internal is sufficient.

utils.ps1
```diff
@@ -0,0 +1,298 @@
...
+        }
+        if (!$importedInProfile -and !$underTest) {
+            $importedInProfile = Test-DbaToolsImportedInScript $PROFILE.AllUsersCurrentHost
+        }
+        if (!$importedInProfile -and !$underTest) {
+            $importedInProfile = Test-DbaToolsImportedInScript $PROFILE.AllUsersAllHosts
+        }
+
+        if ($importedInProfile) {
+            Write-Warning "Skipping add of DbaTools import to file '$profilePath'."
```
wsmelton: All of these (warnings/verbose/errors/host output) need to be adjusted to use the internal function `Write-Message`. You can see [the wiki](https://github.com/sqlcollaborative/dbatools/wiki/Implementing-the-messaging-system) for more info on how our messaging system works.

potatoqualitee:
thanks so much @ChocolateMonkey! This neglected lil command has needed an update for so long. Shawn had some good points in his review - if we can get them addressed quickly, your updates can be included in the next release 💯

utils.ps1
potatoqualitee: Correct, in addition, we do not support v2 - dbatools is v3 and up, thankfully ;)

dbatools.psm1
potatoqualitee: Indeed - all internal functions should be internal and all external functions (user) functions should be in external

dbatools.psd1
potatoqualitee: 💪 I am ;) And this is indeed an internal function. In the future when you do add an external/user function, I'll add it to the psd1. Reason being, it always causes conflicts so having one designated person resolves that issue.

utils.ps1
```diff
@@ -0,0 +1,298 @@
...
+            }
+            $global:Error.RemoveRange(0, $numNewErrors)
+        }
+    }
+}
+
+<#
+.SYNOPSIS
+    Configures your PowerShell profile (startup) script to import the DbaTools
+    module when PowerShell starts.
```
potatoqualitee: Unfortunately, this deviates from standard PowerShell practice -- with PowerShell v3+ so long as dbatools is in the user's profile, it will dynamically load as needed. If anything, you could prompt the user if they'd like to modify their psmodulepath (HKCU, probably!) to include their git repo

utils.ps1
```diff
@@ -0,0 +1,298 @@
...
+    if ($PSCmdlet.ShouldProcess($profilePath, "Add 'Import-Module DbaTools' to profile")) {
+        Add-Content -LiteralPath $profilePath -Value $profileContent -Encoding UTF8
+    }
+    if ($StartSshAgent -and $PSCmdlet.ShouldProcess($profilePath, "Add 'Start-SshAgent -Quiet' to profile")) {
+        Add-Content -LiteralPath $profilePath -Value 'Start-SshAgent -Quiet' -Encoding UTF8
+    }
+}
+
+<#
+.SYNOPSIS
```
potatoqualitee: Many of these functions will not be needed - hit me up on Slack (I'm cl) and we can work together to make install.ps1 awesome 🤗

ChocolateMonkey: Probably not 😁 I just completely lifted and shifted the entire file from posh-git and found and replaced posh-git with dbatools 🙂

utils.ps1
ChocolateMonkey: This is completely lifted from posh-git, really I just wanted the add to profile bit, but I'll see what I can do.

utils.ps1
potatoqualitee: thanks for joining on Slack! Let me know when you've got time and we'll fix up install.ps1 for you :) Till then, I'll just throw in a catch to ensure it doesnt destroy your repo.

PR #1087, opened 2017-04-24
Write-DbaDataTable and Out-DbaDataTable enhancements

JohnRoos:
Updated Write-DbaDataTable and Out-DbaDataTable with features from Write-ObjectToSQL.

Out-DbaDataTable
- Added more data types
- Added special data types for situations where the type does not fit into a table very well without type conversion, like the TimeSpan object
- Added logic for resolving script properties before type is checked
- Removed check for property value before evaluating type, since it should not be needed anymore

Write-DbaDataTable
- Added hash table for type conversion between PowerShell and SQL Server data types
- Updated help message to suggest user to use Out-DbaDataTable if wrong type of input object is used
- Function now converts data types much better than before when creating target table. No more varchar where it shouldn't be (except for timespan which is special).
- Replaced varchar with nvarchar when creating target table
- Strings are now converted into nvarchar(MAX) since there is no way of knowing how long the strings will be from the pipeline (or further down the DataTable)
- Added example showing the type conversion in action (hint: the process object has a lot of properties)
- Left some commented out verbose logging in there for now. Should be removed later

How to test this code: 
- Use the examples as test. The latest example in Write-DbaDataTable provides a good test for type conversion.
- The following object is also a good test to use (and to make sure that the converted types in SQL are correct):
$obj = New-Object -TypeName psobject -Property @{
    guid = [system.guid]::NewGuid()
    timespan = New-TimeSpan -Start 2017-01-01 -End 2017-01-02
    datetime = Get-Date
    char = [System.Char]'Z'
    true = $true
    false = $false
    null = [bool]$null
    string = "some text"
    UInt64 = [System.UInt64]123456
}

Tested on Windows 10 and SQL Server 2014.

### DIFF
--- a/functions/Out-DbaDataTable.ps1
+++ b/functions/Out-DbaDataTable.ps1
@@ -60,32 +60,69 @@ Similar to above but $dbalist gets piped in
 			param ($type)
 			
 			$types = @(
-				'System.Boolean',
-				'System.Byte[]',
-				'System.Byte',
-				'System.Char',
-				'System.Datetime',
-				'System.Decimal',
-				'System.Double',
-				'System.Guid',
-				'System.Int16',
-				'System.Int32',
-				'System.Int64',
-				'System.Single',
-				'System.UInt16',
-				'System.UInt32',
-				'System.UInt64')
-			
-			if ($types -contains $type)
+                'Int32',    
+                'UInt32',   
+                'Int16',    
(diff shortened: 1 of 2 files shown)


potatoqualitee:
💃

potatoqualitee:
Thank you so much @JohnRoos and welcome to the team!!

PR #1090, opened 2017-04-25
New command New-DbaAgentJob

sanderstad:


functions/New-DbaAgentJob.ps1
```diff
@@ -0,0 +1,476 @@
...
+Copyright (C) 2016 Chrissy LeMaire
+This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
+This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+You should have received a copy of the GNU General Public License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+.LINK
+https://dbatools.io/New-DbaAgentJob
+
+.EXAMPLE   
+New-DbaAgentJob -SqlServer 'sql1' -JobName 'Job1' -Description 'Just another job'
```
potatoqualitee: Job1 would not need quotes - please either remove quotes or name it 'Example Job' or 'Job 1' so that the quotes are justified 👍

functions/New-DbaAgentJob.ps1
```diff
@@ -0,0 +1,476 @@
...
+.EXAMPLE   
+New-DbaAgentJob -SqlServer 'sql1' -JobName 'Job1' -Disabled
+Creates the job but sets it to disabled
+
+.EXAMPLE
+New-DbaAgentJob -SqlServer 'sql1' -JobName 'Job1' -EventLogLevel 'OnSuccess'
+Creates the job and sets the notification to write to the Windows Application event log on success
+
+.EXAMPLE
+New-DbaAgentJob -SqlServer 'SSTAD-PC' -JobName 'Job1' -EmailLevel 'OnFailure' -EmailOperatorName 'dba'
```
potatoqualitee: We've moved to SqlInstance - this should be updated. In addition, no quotes are needed anywhere in this example.

functions/New-DbaAgentJob.ps1
```diff
@@ -0,0 +1,476 @@
...
+.EXAMPLE   
+"sql1", "sql2", "sql3" | New-DbaAgentJob -JobName 'Job1'
+Creates a job with the name "Job1" on multiple servers
+#>
+
+    [CmdletBinding(DefaultParameterSetName = "Default", SupportsShouldProcess = $true)]
+
+    param(
+        [parameter(Mandatory = $true, ValueFromPipeline=$true)]
+		[object[]]$SqlServer,
```
potatoqualitee: Please copy/paste from another command that has the appropriate parameter aliases too and also change this to $SqlInstance as noted above.

functions/New-DbaAgentJob.ps1
```diff
@@ -0,0 +1,476 @@
...
+        [object]$PageLevel = '',
+        [Parameter(Mandatory = $false)]
+        [ValidateScript({
+            if(($EmailLevel -in 1, "OnSuccess", 2, "OnFailure", 3, "Always") -and ($_.Length -ge 1))
+            {
+                $true
+            }
+            else
+            {
+                Throw "Please set the e-mail operator when the e-mail level parameter is set."
```
potatoqualitee: We've gone away from throws since they are scary. @FriedrichWeinmann , can you please confirm that Stop-Function would be appropriate here?

functions/New-DbaAgentJob.ps1
```diff
@@ -0,0 +1,476 @@
...
+        [ValidateSet(0, "Never", 1, "OnSuccess", 2, "OnFailure", 3, "Always")]
+        [object]$DeleteLevel = '',
+        [bool]$Force,
+        [switch]$Silent
+    )
+
+    BEGIN
+    {
+        # Check the input for the job id or job name
+        if($JobName.Length -lt 1)
```
potatoqualitee: This should be `!$jobname` or `-not $jobname` instead.

functions/New-DbaAgentJob.ps1
```diff
@@ -0,0 +1,476 @@
...
+        {
+            Stop-Function -Message "Please enter a job name." -Silent $Silent -InnerErrorRecord $_ -Target $JobName
```
potatoqualitee: The -Silent is inherited and should not be explicitly specified.

`Stop-Function -Message "Please enter a job name." -InnerErrorRecord $_ -Target $JobName`

Same goes for all commands below

functions/New-DbaAgentJob.ps1
```diff
@@ -0,0 +1,476 @@
...
+            return
+        }
+
+        # Check of the event log level is of type string and set the integer value
+        if(($EmailLevel.GetType().Name -eq 'String') -and ($EventLogLevel.Length -ge 1))
```
potatoqualitee: This could be  `-and ($EventLogLevel))` I believe

functions/New-DbaAgentJob.ps1
```diff
@@ -0,0 +1,476 @@
...
+                "Always" { $DeleteLevel = 3 } 
+            }
+        }
+
+        [object[]]$JobsCreated = $null
+    }
+
+    PROCESS
+    {
+        # Try connecting to the instance
```
potatoqualitee: This should be foreach ($instance in $sqlinstance) so that it handles multiple servers at once

functions/New-DbaAgentJob.ps1
```diff
@@ -0,0 +1,476 @@
...
+        Write-Message -Message "Attempting to connect to Sql Server.." -Level 2 -Silent $Silent
+        try 
+        {
+            $Server = Connect-SqlServer -SqlServer $SqlServer -SqlCredential $SqlCredential
+        }
+        catch 
+        {
+            Stop-Function -Message "Could not connect to Sql Server instance" -Silent $Silent -Target $SqlServer
```
potatoqualitee: @FriedrichWeinmann - this would not give enough info to the user. How do we do it? include $_ in -Message?

functions/New-DbaAgentJob.ps1
```diff
@@ -0,0 +1,476 @@
...
+        catch 
+        {
+            Stop-Function -Message "Could not connect to Sql Server instance" -Silent $Silent -Target $SqlServer 
+            return
+        }
+
+        # Check if the job already exists
+        if(-not $Force -and (($Server.JobServer.Jobs).Name -contains $JobName))
+        {
+            Stop-Function -Message "Job '$($JobName)' already exists on '$($SqlServer)'" -Silent $Silent -Target $SqlServer
```
potatoqualitee: Please remove extra quotes as defined in the style guide. This can be "$jobname already exists on $instance"

functions/New-DbaAgentJob.ps1
```diff
@@ -0,0 +1,476 @@
...
+        }
+
+        # Create the job object
+        try 
+        {
+            $Job = New-Object Microsoft.SqlServer.Management.Smo.Agent.Job($Server.JobServer, $JobName)
+        }
+        catch 
+        {
+            Stop-Function -Message ("Something went wrong creating the job. `n$($_.Exception.Message)") -Silent $Silent -Target $JobName
```
potatoqualitee: No () is needed around the message.

functions/New-DbaAgentJob.ps1
```diff
@@ -0,0 +1,476 @@
...
+                }
+                else 
+                {
+                    Stop-Function -Message ("The e-mail operator name '$($EmailOperatorName)' does not exist on instance '$($SqlServer)'. Exiting..") -Silent $Silent -Target $JobName
+                    return
+                }
+            }
+            else 
+            {
+                Write-Message -Message "Invalid combination of e-mail operator name '$($EmailOperatorName)' and email level $($EmailLevel). Not setting the notification." -Warning -Silent $Silent
```
potatoqualitee: No extra quotes plz :)

functions/New-DbaAgentJob.ps1
```diff
@@ -0,0 +1,476 @@
...
+            }
+            catch
+            {
+                Write-Message -Message ("Something went wrong creating the job. `n$($_.Exception.Message)") -Level 2 -Silent $Silent 
+                
+            }
+        }
+        
+        # Add the job to the list
+        $JobsCreated += $Job
```
potatoqualitee: No list needed - please output immediately - right here can just be $job

functions/New-DbaAgentJob.ps1
```diff
@@ -0,0 +1,476 @@
...
+    }
+
+    END
+    {
+        Write-Message -Message "Finished creating job(s)." -Level 2 -Silent $Silent
```
potatoqualitee: Please consider named levels i the future - like Output, Verbose, VeryVerbose, etc.

functions/New-DbaAgentJob.ps1
```diff
@@ -0,0 +1,476 @@
...
+                "Always" { $EmailLevel = 3 } 
+            }
+        }
+
+        # Check of the net send level is of type string and set the integer value
+        if(($NetsendLevel.GetType().Name -eq 'String') -and ($NetsendLevel.Length -ge 1))
+        {
+            switch ($NetsendLevel)
+            {
+                {($_ -eq "Never") -or ($_ -eq '')} { $NetsendLevel = 0 }
```
PowerDBAKlaas: I think this would be better as the final entry:
default { 0 }

also:
$netsendlevel = switch($netsendlevel) { 'OnSuccess' { 1 } 'Onfailure' {2} 'Always' { 3 } default {0}}
will replace those 8 lines

same for the other switches

functions/New-DbaAgentJob.ps1
```diff
@@ -0,0 +1,476 @@
...
+                "OnFailure" { $DeleteLevel = 2 }
+                "Always" { $DeleteLevel = 3 } 
+            }
+        }
+
+        [object[]]$JobsCreated = $null
+    }
+
+    PROCESS
+    {
```
PowerDBAKlaas: the parameter definition accepts multiple sql servers, yet there is no foreach loop
this connects to $sqlserver which can be a collection
either add foreach or limit sqlserver to be only one instance (probably easier and best option for now)

did you test the examples given?

functions/New-DbaAgentJob.ps1
```diff
@@ -0,0 +1,476 @@
...
+            return
+        }
+
+        # Check if the job already exists
+        if(-not $Force -and (($Server.JobServer.Jobs).Name -contains $JobName))
+        {
+            Stop-Function -Message "Job '$($JobName)' already exists on '$($SqlServer)'" -Silent $Silent -Target $SqlServer 
+            return
+        }
+        elseif($Force -and (($Server.JobServer.Jobs).Name -contains $JobName))
```
PowerDBAKlaas: shouldn't there be some more `if ($PSCmdlet.ShouldProcess(`
around the job removal?

functions/New-DbaAgentJob.ps1
FriedrichWeinmann: `Stop-Function -Message "Could not connect to Sql Server instance" -Silent $Silent -Target $SqlServer`
becomes
`Stop-Function -Message "[$SqlServer] Could not connect to Sql Server instance: $($_.Exception.Message)" -Silent $Silent -Target $SqlServer -InnerErrorRecord $_ -Continue`
(The `-Continue ` is a requirement when the loop you asked for comes)

functions/New-DbaAgentJob.ps1
FriedrichWeinmann: Nope, it's a validate script in the parameter binding, no way to do this without error _in this location_.
Validation could be moved into the begin block however, where we could indeed use Stop-Function to handle this.

Btw, we could implement the various Level-parameters as an enumerated type (as with the `-Level` parameter on Write-Message), saving lots of ValidateSets and removing the numbers from the auto-complete list (while still being valid input).

functions/New-DbaAgentJob.ps1
```diff
@@ -0,0 +1,476 @@
...
+                    return
+                }
+            }
+            else 
+            {
+                Write-Message -Message "Invalid combination of page operator name '$($PageOperatorName)' and page level $($PageLevel). Not setting the notification." -Warning -Silent $Silent
+            }
+        }
+
+        if($DeleteLevel -ge 0)
```
PowerDBAKlaas: if we remove the '' as default for the parameter, and there's a validateset,
and we add default {0} where the strings are switched to int in the BEGIN{}, 
these if () are redundant

functions/New-DbaAgentJob.ps1
```diff
@@ -0,0 +1,476 @@
...
+                Write-Message -Message "Job created with UID '$($Job.JobID)'" -Level 5 -Silent $Silent
+            }
+            catch
+            {
+                Write-Message -Message ("Something went wrong creating the job. `n$($_.Exception.Message)") -Level 2 -Silent $Silent 
+                
+            }
+        }
+        
+        # Add the job to the list
```
PowerDBAKlaas: there is no list, it's just one job

functions/New-DbaAgentJob.ps1
FriedrichWeinmann: General note on testing and the silent parameter:
I have found the inheritance issue not working perfectly _in tests_ when importing the .psm1 and the test function not being a part of the module (like pasted directly or imported separately. In this case the `$Silent` parameter may not be passed correctly. This can be worked around by either of these two methods:
 - Testing as part of module (Just paste new files in the correct folder, then rmo/ipmo)
 - Defining a `$Silent = $false` variable on the commandline

potatoqualitee: I always prefer people testing as part of the module.

potatoqualitee: Thanks for the note btw!

functions/New-DbaAgentJob.ps1
PowerDBAKlaas: ooh, now I see: the comment says "check event log level type..." and the code says `if $emaillevel.gettype...`

functions/New-DbaAgentJob.ps1
```diff
@@ -0,0 +1,476 @@
...
+        [int]$StartStepId,
+        [Parameter(Mandatory = $false)]
+        [string]$CategoryName,
+        [Parameter(Mandatory = $false)]
+        [int]$CategoryId,
+        [Parameter(Mandatory = $false)]
+        [string]$OwnerLoginName,
+        [Parameter(Mandatory = $false)]
+        [ValidateSet(0, "Never", 1, "OnSuccess", 2, "OnFailure", 3, "Always")]
+        [object]$EventLogLevel = '',
```
sanderstad: The reason I did that is because I'm checking if the value submitted is text and than do a switch to get the right integer value. I could probably set it to 0 which is the default for sql server

PowerDBAKlaas: I think there should be no default at all, 
if the switch statement has a default, that will be assigned when the parameter is not set too

functions/New-DbaAgentJob.ps1
sanderstad: Typically a type there. I fixed that locally already. Will be committing the new version soon

functions/New-DbaAgentJob.ps1
sanderstad: Yeah that is kind of redundant isn't it? It's removed

functions/New-DbaAgentJob.ps1
sanderstad: That is a good idea. I'll implement that

functions/New-DbaAgentJob.ps1
sanderstad: I did test the examples. There is now a foreach loop to loop through the instances. I assumed we'd only want people to add multiple servers using the pipe lines.

FriedrichWeinmann: Most may do so, but not all and it's an artificial limit that needs not be and is easily removed. General practice is that all functions that accept pipeline input must also accept an array as input.
All other habits confuse users (think about Export-Csv for a stellar example of failing this, though they can't really help it. Still confused most beginners when they first hit it).

functions/New-DbaAgentJob.ps1
sanderstad: I only use the switch when the value is a text type. If it's a number I want to keep it as it is.
The default in the switch would only be called if the value had invalid parameter  value which cannot occur due to the validateset.

functions/New-DbaAgentJob.ps1
PowerDBAKlaas: I agree with @FriedrichWeinmann 

I now see the only example submitting multiple instances is indeed through the pipeline.
That leads to this next request 😄 : @sanderstad , can you add an example supplying multiple instances at the -SqlInstance parameter, please?
and also mention in the comment based help that SqlInstance accepts more than one?

functions/New-DbaAgentJob.ps1
PowerDBAKlaas: and if you execute the switch when `if $eventloglevel -notin (1,2,3)` ?

functions/New-DbaAgentJob.ps1
wsmelton: Yeah, I can say I have tried not including the silent within a command and it does fail with an error specific to that parameter not being provided.

potatoqualitee:
hey @sanderstad - can you please fix your git conflicts? please dont forget that I am the only one who is assigned the editing of psm1 and psd1 files.

potatoqualitee:
It appears that the easiest way will be to undo your commits that deleted both of these files.

potatoqualitee:
or just close this PR and resubmit with fresh branch

potatoqualitee:
I see what happened -- instead of deleting, try just right clicking and dismissing changes like this

image

merci

sanderstad:
Going to submit it through another PR to fix the extra files issue

PR #1093, opened 2017-04-25
New command New-DbaAgentJobStep

sanderstad:


functions/New-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,341 @@
...
+                $Server = Connect-SqlServer -SqlServer $instance -SqlCredential $SqlCredential
+            }
+            catch {
+                Stop-Function -Message "Could not connect to Sql Server instance"  -Target $instance 
+                return
+            }
+
+            # Check if the job exists
+            if (($Server.JobServer.Jobs).Name -notcontains $JobName) {
+                Write-Message -Message "Job $JobName doesn't exists on $instance" -Warning
```
FriedrichWeinmann: In Set-DbaAgentJobStep, you use Stop-Function when this happens, here you write a warning. I'd go with the Stop-Function implementation, as it's better to use in silent mode and allows for cleaner code as we don't need the else-braces.

functions/New-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,341 @@
...
+                }
+
+                if ($RetryInterval) {
+                    Write-Message -Message "Setting job step retry interval to $($RetryInterval)" -Level Verbose 
+                    $JobStep.RetryInterval = $RetryInterval
+                }
+
+                if ($OutputFileName) {
+                    Write-Message -Message "Setting job step output file name to $OutputFileName" -Level Verbose 
+                    $JobStep.OutputFileName = $OutputFileName
```
wsmelton: So it seems that SMO does not have the ability to allow you to set the output file to append (as you can in the UI). As this may be requested I am not sure exactly where/how the best method would be to do it. 

As-is this command will only set the output file for the job step and overwrite every time the job is executed.

wsmelton: Ok, found it. We need to add the ability to set the [JobStepFlags property](https://msdn.microsoft.com/en-us/library/microsoft.sqlserver.management.smo.agent.jobstepflags.aspx).

I think I would just do a specific parameter as `-JobStepFlags` because you can have multiple options (e.g. log to a file and a table). Use validate set so only these can be selected and then the help we may just need to add a more descriptive explanation of them?

| Member name                        | Description                                                                 |
|------------------------------------|-----------------------------------------------------------------------------|
| AppendAllCmdExecOutputToJobHistory | Job history, including command output, is appended to the job history file. |
| AppendToJobHistory                 | Job history is appended to the job history file.                            |
| AppendToLogFile                    | Job history is appended to the SQL Server log file.                         |
| AppendToTableLog                   | Job history is appended to a log table.                                     |
| LogToTableWithOverwrite            | Job history is written to a log table, overwriting previous contents.       |
| None                               | Job history is not appended to a file.                                      |
| ProvideStopProcessEvent            | Job processing is stopped.                                                  |

sanderstad: The reason I did not implement that one is because the documentation of the SMO said this:
> The JobStepFlags property is reserved for future use. 

It works though and it was not that difficult to implement.

sanderstad:
Close and reopen new PR

PR #1094, opened 2017-04-25
New command Set-DbaAgentJobStep

sanderstad:


functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,332 @@
...
+{
+<#
+.SYNOPSIS 
+Set-DbaAgentJobStep updates a job step.
+
+.DESCRIPTION
+Set-DbaAgentJobStep updates a job step in the SQL Server Agent with parameters supplied.
+
+.PARAMETER SqlServer
+SQL Server instance. You must have sysadmin access and server version must be SQL Server version 2000 or greater.
```
FriedrichWeinmann: official naming convention was settled on `SqlInstance` I believe

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,332 @@
...
+        [Parameter(Mandatory = $false)]
+        [int]$CmdExecSuccessCode,
+        [Parameter(Mandatory = $false)]
+        [ValidateSet('QuitWithSuccess', 'QuitWithFailure', 'GoToNextStep', 'GoToStep')]
+        [string]$OnSuccessAction,
+        [Parameter(Mandatory = $false)]
+        [ValidateScript({
+            if(($OnSuccessAction -ne 'GoToStep') -and ($_ -ge 1))
+            {
+                Throw "Parameter OnSuccessStepId can only be used with OnSuccessAction 'GoToStep'."
```
FriedrichWeinmann: Either ignore in that case or handle in begin block with the more smooth `Stop-Function` handling.
If we choose to keep the interrupt, the comment based help should reflect this action.

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,332 @@
...
+        })]
+        [int]$OnSuccessStepId,
+        [Parameter(Mandatory = $false)]
+        [ValidateSet('QuitWithSuccess', 'QuitWithFailure', 'GoToNextStep', 'GoToStep')]
+        [string]$OnFailAction,
+        [Parameter(Mandatory = $false)]
+        [ValidateScript({
+            if(($OnFailAction -ne 'GoToStep') -and ($_ -ge 1))
+            {
+                Throw "Parameter OnFailStepId can only be used with OnFailAction 'GoToStep'."
```
FriedrichWeinmann: Either ignore in that case or handle in begin block with the more smooth `Stop-Function` handling.
If we choose to keep the interrupt, the comment based help should reflect this action

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,332 @@
...
+        [Parameter(Mandatory = $false)]
+        [string]$OutputFileName,
+        [Parameter(Mandatory = $false)]
+        [string]$ProxyName,
+        [switch]$Silent
+    )
+
+    PROCESS
+    {
+        # Try connecting to the instance
```
FriedrichWeinmann: Since you accept multiple instances as parameter, be sure to allow for non-pipeline input by adding a foreach loop:
```
foreach ($instance in $SqlInstance)
{
    # ...
}
```

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,332 @@
...
+        Write-Message -Message "Attempting to connect to Sql Server.." -Level 2 -Silent $Silent
```
FriedrichWeinmann: Consider telling the user which server to connect to. Also, connection actions are usually level 5 (or `Verbose`)
This guide details the messaging system and how to implement it:
https://github.com/sqlcollaborative/dbatools/wiki/Implementing-the-messaging-system
With plenty of examples. Example on how this might look instead:
`Write-Message -Level Verbose -Message "Attempting to connect to $instance ..." -Target $instance`

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,332 @@
...
+        try 
+        {
+            $Server = Connect-SqlServer -SqlServer $SqlServer -SqlCredential $SqlCredential
+        }
+        catch 
+        {
+            Stop-Function -Message "Could not connect to Sql Server instance" -Silent $Silent -Target $SqlServer 
+            return
```
FriedrichWeinmann: As above, it might be nice to include the server you are trying to connect to. Also more details on why it failed:
`Stop-Function -Message "Could not connect to $instance : $($_.Exception.Message)" -Target $instance -InnerErrorRecord $_`
This way, the error text gets appended _and_ the error log will have the original exception, making it that much easier to troubleshoot. If you implement the loop, also use the `-Continue` parameter, to keep processing the next server if one failed.

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,332 @@
...
+        }
+
+        # Check if the job exists
+        if(($Server.JobServer.Jobs).Name -notcontains $JobName)
+        {
+            Stop-Function -Message "Job '$($JobName)' doesn't exists on '$($SqlServer)'" -Silent $Silent -Target $SqlServer
```
FriedrichWeinmann: As Chrissy already mentioned in another PR of yours, extra quotes are not desired within message texts, though the information is appreciated and should stay

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,332 @@
...
+                return
+            }
+            else 
+            {
+                # Get the job step
+                $JobStep = $Server.JobServer.Jobs[$JobName].JobSteps[$StepName]
+
+                #region job step options
+                # Setting the options for the job step
+                if($NewName.Length -ge 1)
```
FriedrichWeinmann: could be `if ($NewName)` instead. Same is true for most of the other conditionals you use here

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,332 @@
...
+                    $JobStep.Subsystem = $Subsystem
+                }
+
+                if($Command.Length -ge 1)
+                {
+                    Write-Message -Message "Setting job step command to '$($Command)'" -Level 5 -Silent $Silent
+                    $JobStep.Command = $Command
+                }
+
+                if($CmdExecSuccessCode -ge 0)
```
FriedrichWeinmann: Will also be applied *when the parameter is not bound*. Beware of that comparison, valuetypes are initialized with their default value when parameters are not bound (which in case of integer is '0'). Instead use this:
`if ($PSBoundParameters.ContainsKey("CmdExecSuccessCode"))`
In order to check whether a parameter was bound

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,332 @@
...
+                    # Check if the database is present on the server
+                    if(($Server.Databases).Name -contains $DatabaseName)
+                    {
+                        Write-Message -Message "Setting job step database name to '$($DatabaseName)'" -Level 5 -Silent $Silent
+                        $JobStep.DatabaseName = $DatabaseName
+                    }
+                    else 
+                    {
+                        Write-Message -Message ("The database is not present on instance '$($SqlServer)'.") -Warning -Silent $Silent 
+                        return
```
FriedrichWeinmann: Unneeded braces around the string.
The `-Warning` parameter has been deprecated by Write-Message (it will write a warning and tell you so if you run it). Use `-Level Warning` instead.
Also if you are interrupting an execution for this, consider using `Stop-Function` instead. Is it worth interrupting the flow for this?

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,332 @@
...
+                    try
+                    {
+                        Write-Message -Message ("Changing the job step") -Level 2 -Silent $Silent
+                        
+                        # Change the job step
+                        $JobStep.Alter()
+                    }
+                    catch
+                    {
+                        Write-Message -Message ("Something went wrong changing the job step. `n$($_.Exception.Message)") -Level 2 -Silent $Silent
```
FriedrichWeinmann: Instead of writing a warning, use stop-function here. Even if it's the last line and nothing comes after it, Stop-Function exists to handle unattended mode, for example as part of scripts. If this fails, the calling script must now, so it needs the exception thrown by Stop-Function in silent mode. Even if not, it handles debugging information and should be used for troubleshooting and support information.

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,302 @@
...
+        }
+
+        # Check the parameter on success step id
+        if (($OnFailAction -ne 'GoToStep') -and ($OnFailStepId -ge 1)) {
+            Stop-Function -Message "Parameter OnFailStepId can only be used with OnFailAction 'GoToStep'."  -Target $instance 
+            return
+        }
+    }
+
+    process {
```
FriedrichWeinmann: Needs to check for function interruption due to validation in begin:
as first act in process

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,302 @@
...
+    begin{
+        # Check the parameter on success step id
+        if (($OnSuccessAction -ne 'GoToStep') -and ($OnSuccessStepId -ge 1)) {
+            Stop-Function -Message "Parameter OnSuccessStepId can only be used with OnSuccessAction 'GoToStep'."  -Target $instance 
+            return
+        }
+
+        # Check the parameter on success step id
+        if (($OnFailAction -ne 'GoToStep') -and ($OnFailStepId -ge 1)) {
+            Stop-Function -Message "Parameter OnFailStepId can only be used with OnFailAction 'GoToStep'."  -Target $instance
```
FriedrichWeinmann: Not processing any item yet, so `-Target $Instance` is not appropriate yet (and the variable $instance doesn't exist yet).

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,302 @@
...
+        [string]$OutputFileName,
+        [Parameter(Mandatory = $false)]
+        [string]$ProxyName,
+        [switch]$Silent
+    )
+
+    begin{
+        # Check the parameter on success step id
+        if (($OnSuccessAction -ne 'GoToStep') -and ($OnSuccessStepId -ge 1)) {
+            Stop-Function -Message "Parameter OnSuccessStepId can only be used with OnSuccessAction 'GoToStep'."  -Target $instance
```
FriedrichWeinmann: Not processing any item yet, so `-Target $Instance` is not appropriate yet (and the variable $instance doesn't exist yet).

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,302 @@
...
+        if (($OnFailAction -ne 'GoToStep') -and ($OnFailStepId -ge 1)) {
+            Stop-Function -Message "Parameter OnFailStepId can only be used with OnFailAction 'GoToStep'."  -Target $instance 
+            return
+        }
+    }
+
+    process {
+        foreach ($instance in $sqlinstance) {
+            # Try connecting to the instance
+            Write-Message -Message "Attempting to connect to Sql Server.." -Level Output
```
FriedrichWeinmann: Add the server you are connecting to to the message, otherwise the information doesn't tell the user all that much

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,302 @@
...
+    process {
+        foreach ($instance in $sqlinstance) {
+            # Try connecting to the instance
+            Write-Message -Message "Attempting to connect to Sql Server.." -Level Output 
+            try {
+                $Server = Connect-SqlServer -SqlServer $instance -SqlCredential $SqlCredential
+            }
+            catch {
+                Stop-Function -Message "Could not connect to Sql Server instance" -Target $instance 
+                return
```
FriedrichWeinmann: Instead of the interrupt version, here you want to use the continue implementation of stop function, as only processing of this item has failed. Also include the exception in it and include the server name in the message:
` Stop-Function -Message "Could not connect to Sql Server instance $instance: $($_.Exception.Message)" -Target $instance -InnerErrorRecord $_ -Continue`

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,302 @@
...
+            }
+
+            # Check if the job exists
+            if (($Server.JobServer.Jobs).Name -notcontains $JobName) {
+                Stop-Function -Message "Job $JobName doesn't exists on $instance" -Target $instance 
+                return
```
FriedrichWeinmann: Instead of the interrupt version, here you want to use the continue implementation of stop function, as only processing of this item has failed:
` Stop-Function -Message "Job $JobName doesn't exists on $instance" -Target $instance -Continue`

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,302 @@
...
+            }
+            else {
```
FriedrichWeinmann: No need for this `else`. If the if-condition is met the continue will prevent this code from running. No need to add extra braces and indentations to the code.

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,302 @@
...
+                # Check if the job step exists
+                if (($Server.JobServer.Jobs[$JobName].JobSteps).Name -notcontains $StepName) {
+                    Stop-Function -Message "Step $StepName doesn't exists on $instance" -Target $instance 
+                    return
```
FriedrichWeinmann: Instead of the interrupt version, here you want to use the continue implementation of stop function, as only processing of this item has failed:
` Stop-Function -Message "Step $StepName doesn't exists on $instance" -Target $instance -Continue`

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,302 @@
...
+                }
+                else {
```
FriedrichWeinmann: No need for this `else`. If the if-condition is met the continue will prevent this code from running. No need to add extra braces and indentations to the code.

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,302 @@
...
+
+                    if ($DatabaseName) {
+                        # Check if the database is present on the server
+                        if (($Server.Databases).Name -contains $DatabaseName) {
+                            Write-Message -Message "Setting job step database name to $DatabaseName" -Level Verbose 
+                            $JobStep.DatabaseName = $DatabaseName
+                        }
+                        else {
+                            Stop-Function -Message ("The database is not present on instance $instance.") -Target $SqlServer 
+                            return
```
FriedrichWeinmann: Instead of the interrupt version, here you want to use the continue implementation of stop function, as only processing of this item has failed. Also no need to wrap the string in braces here:
` Stop-Function -Message "The database is not present on instance $instance." -Target $instance -Continue`

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,302 @@
...
+
+                    if (($DatabaseUserName) -and ($DatabaseName)) {
+                        # Check if the username is present in the database
+                        if (($Server.Databases[$DatabaseName].Users).Name -contains $DatabaseUserName) {
+                            Write-Message -Message "Setting job step database username to $DatabaseUserName" -Level Verbose 
+                            $JobStep.DatabaseUserName = $DatabaseUserName
+                        }
+                        else {
+                            Stop-Function -Message ("The database user is not present in the database $DatabaseName on instance $instance.") -Target $SqlServer 
+                            return
```
FriedrichWeinmann: Instead of the interrupt version, here you want to use the continue implementation of stop function, as only processing of this item has failed. Also no need to wrap the string in braces here:
` Stop-Function -Message "The database user is not present in the database $DatabaseName on instance $instance." -Target $instance -Continue`

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,302 @@
...
+
+                    if ($ProxyName) {
+                        # Check if the proxy exists
+                        if (($Server.JobServer.ProxyAccounts).Name -contains $ProxyName) {
+                            Write-Message -Message "Setting job step proxy name to $ProxyName" -Level Verbose 
+                            $JobStep.ProxyName = $ProxyName
+                        }
+                        else {
+                            Stop-Function -Message ("The proxy name $ProxyName doesn't exist on instance $instance.") -Target $SqlServer 
+                            return
```
FriedrichWeinmann: Instead of the interrupt version, here you want to use the continue implementation of stop function, as only processing of this item has failed. Also no need to wrap the string in braces here:
`Stop-Function -Message "The proxy name $ProxyName doesn't exist on instance $instance." -Target $instance -Continue`

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,302 @@
...
+                        }
+                    }
+                    #region job step options
+
+                    # Execute 
+                    if ($PSCmdlet.ShouldProcess($SqlServer, ("Changing the job step $StepName"))) {
+                        try {
+                            Write-Message -Message ("Changing the job step") -Level Output
```
FriedrichWeinmann: Please don't wrap string in braces

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,302 @@
...
+                        }
+                        else {
+                            Stop-Function -Message ("The proxy name $ProxyName doesn't exist on instance $instance.") -Target $SqlServer 
+                            return
+                        }
+                    }
+                    #region job step options
+
+                    # Execute 
+                    if ($PSCmdlet.ShouldProcess($SqlServer, ("Changing the job step $StepName"))) {
```
FriedrichWeinmann: Please don't wrap string in braces

functions/Set-DbaAgentJobStep.ps1
```diff
@@ -0,0 +1,302 @@
...
+                        try {
+                            Write-Message -Message ("Changing the job step") -Level Output 
+                        
+                            # Change the job step
+                            $JobStep.Alter()
+                        }
+                        catch {
+                            Write-Message -Message ("Something went wrong changing the job step. `n$_.Exception.Message)") -Level Output
```
FriedrichWeinmann: Should use `Stop-Function` with the ` -Continue`  parameter instead. Avoid artificial linebreaks and remember that the properties of variables are not evaluated in strings. So rather than this:
`"$_.Exception.Message"`
use this:
`"$($_.Exception.Message)"`

potatoqualitee:
🏅 🏅 🏅 🏅 🏅 !

PR #1110, opened 2017-04-28
export-sqllogin 1.0 update

potatoqualitee:


potatoqualitee:
whys that, @niphlod ? it does not output to file, only screen.

niphlod:
sorry, was on the run and on a mobile phone. forgot that it is "informational" only.

potatoqualitee:
yay thank you merging 👍

PR #1111, opened 2017-04-28
Update for 1.0 - Copy DatabaseMail output to be an object of status

wsmelton:
Update for 1.0 - fixing output to be pscustomobject

 - Added pscustomobject of each item being copied and output that object
 - Changed all write-output to be verbose output utilizing messaging system.

### On Source Server
- [ ] Change the system settings for DBMail to other values (e.g Set "Account Retry Deaily" to 75 and "Maximum File Size" to 100)
- [ ] Create a DBMail profile
- [ ] Create a DBMail Account (minimum just name, email and server/port)

Run command:
```powershell
$results = Copy-SqlDatabaseMail -Source Server -Destination Server -Verbose
```

### On Destination Server
- [ ] Confirm System Configurations were updated
- [ ] Confirm profile was created
- [ ] Confirm account was created

### General verification of output
- [ ] Verify each item has expected status (should be an object for each item copied in the command)
- [ ] Run the command again without use of -Force and verify the "Skipped" status shows as expected for those objects receiving warning.

### Expected output scenarios

1) Initial run where everything is copied and created:
image

2) Secondary run where accounts and profiles already exist, so you get warnings that the -Force parameter should be used:
image

## Has been tested on minimum requirements:
- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

## Has been tested on maximum requirements:
- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database


functions/Copy-SqlDatabaseMail.ps1
```diff
@@ -115,34 +115,60 @@ Shows what would happen if the command were executed.
...
             }
+            $copyMailConfigStatus
         }
 
         function Copy-SqlDatabaseAccount {
             $sourceAccounts = $sourceServer.Mail.Accounts
             $destAccounts = $destServer.Mail.Accounts
 
-            Write-Message -Message "Migrating accounts" -Level Output -Silent $Silent
+            Write-Message -Message "Migrating accounts" -Level Verbose -Silent $Silent
```
potatoqualitee: The only thing I'd change is this -Silent $Silent is not needed since it's inherited (thats why you add it at the top)

potatoqualitee:
Looks soooo damn good! can you show others the pic you sent me or no? if no, ill take one.

functions/Copy-SqlDatabaseMail.ps1
wsmelton: Removed

potatoqualitee:
Looks great! The only thing I changed - within error messages, i like $sourceserver but i changed it to .name in the output. i also removed the end, it was an unnecessary practice i picked up because i could.

PR #1114, opened 2017-04-29
Added ToString for dbainstanceparameter class

FriedrichWeinmann:
Updates DbaInstanceParameter to include a ToString() method that will show nicely what the target is named


potatoqualitee:
Thanks bud!

PR #1117, opened 2017-04-29
Added helper function to test for bound parameters

FriedrichWeinmann:
Adds helper function to test whether a parameter was bound, independent of the actual value bound


potatoqualitee:
Ouuu SO GOOD! Need to announce to the channel

PR #1118, opened 2017-04-29
Internal command - Was-Bound - Renamed Reverse to Not

FriedrichWeinmann:


potatoqualitee:
I'm going to leave this here a little bit (maybe) so ppl get to see there's a new internal command called `Was-Bound`

PR #1120, opened 2017-04-29
Set-SqlTempdbConfiguration - cleaned up, made output more useful

potatoqualitee:
image

tested local and remote

### DIFF
--- a/functions/Set-SqlTempDBConfiguration.ps1
+++ b/functions/Set-SqlTempDBConfiguration.ps1
@@ -110,7 +110,7 @@ Returns PSObject representing tempdb configuration.
 	param (
 		[parameter(Mandatory = $true)]
 		[Alias("ServerInstance", "SqlServer")]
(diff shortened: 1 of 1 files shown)


functions/Set-SqlTempDBConfiguration.ps1
```diff
@@ -54,7 +54,7 @@ File path to create tempdb data files in. If not specified, current tempdb locat
 .PARAMETER LogPath
 File path to create tempdb log file in. If not specified, current tempdb location will be used.
 
-.PARAMETER Script
+.PARAMETER OutScript
```
wsmelton: I'd prefer to see this called [OutputScriptOnly](https://github.com/sqlcollaborative/dbatools/blob/development/functions/Restore-DbaDatabase.ps1#L76), which I think makes it more clear as to what the switch does.

potatoqualitee: I like that, k

PR #1125, opened 2017-04-30
[New] Functions around pulling system/hardware information of a server

### LINKED ISSUE #837: [Commands] Server/Computer Information
﻿### Is this a *feature* OR *bug*:
[Feature]

### Action Results

In most cases folks could have CMDB systems that contain information about their servers, or some spreadsheet that contains the information they have saved on some network share. This is just basically for that purpose to get a fresh list, or inventory servers after they were migrated possibly.

### Expected Results

Just rough ideas...parameter names are just for example

```powershell
Get-DbaServerInfo -ServerName MyServer
```

Output basic info from win32_operatingsystem or other various bits that we would want to document for a given server. Could also be info we want for troubleshooting (build number, time zone mismatch to domain, etc.)

image

Date values would be cleaned up to display in proper format, and could also get the description of the time zone from `win32_timezone`. 

image

Parameters could be used to get additional information, or output multiple data sets:
```powershell
Get-DbaServerInfo -ServerName MyServer -IncludeDiskInfo 
```

This could output basic information and (maybe) include a data set with the disk information (e.g. disk type, partition count (or detail out volume letter and size) and also include allocation unit size). This parameter might just be for only outputting disk information, so would just say `-OnlyDiskInfo` or something like that. 

It may be best to break this up into a series of commands. We could create a wrapper command to output it all in multiple data sets, that could then just be piped to `ConvertTo-Html` to easily make a basic report for the server. I know this would bleed into what dbareports is for, but I would want this information without having to deploy SSRS or PowerBI.

We have commands under the Test verb that check various bits of the server level information, but none that really just output for folks. Main use case I have would be try to inventory a client environment to figure out what type of OS, maybe hardware, but mostly disk storage the server is configured. This is pertinent to doing migrations to new hardware or in my case new data centers.

wsmelton:
Fixes #837 but adding this as a place holder so duplicate work is not started by another.

### Changes proposed in this pull request:
 Initial idea is to have at most 3 commands:
- Get-DbaSystemHardware
- Get-DbaSystemOS
- Get-DbaSystemDrives

Then possibly have a wrapper function like Get-DbaSystemInfo (? not sure on output format if this would work or not). At most though, this is going to be dependent upon the remote management functions under #916 being completed by @FriedrichWeinmann (no pressure).


potatoqualitee:
hey @wsmelton - #916 has been merged and you can proceed as needed 👍

wsmelton:
Well let's get this 🎉 going 💪

potatoqualitee:
hey :shawn: plz resubmit when files are ready

Issue #1128, opened 2017-04-30
dbadatetime, null if jan 1, 0001


potatoqualitee:
## Is this:
_Just place "x" to indicate selection (e.g. "[x]")_

- [x] *feature* 
- [ ] *bug*:

## Details
SQL Agent is outrageous and gives the standard time of 1/1/0001 12:00:00 AM when something has never occurred. Can dbadatetime return null if it matches the above? Currently, it returns 0001-01-01 00:00:00.000

FriedrichWeinmann:
Solved in:
https://github.com/sqlcollaborative/dbatools/pull/1132

Issue #1121, opened 2017-04-29
dbasize should support -1 (Unlimited)


FriedrichWeinmann:
## Is this:
_Just place "x" to indicate selection (e.g. "[x]")_

- [x] *feature* 
- [ ] *bug*:

## System Details

in Database country, -1 size means unlimited. Dbasize should be able to do that:
 - Accept -1 as value
 - Present "Unlimited" as string interpretation of value

FriedrichWeinmann:
Solved in:
https://github.com/sqlcollaborative/dbatools/pull/1132

PR #1132, opened 2017-04-30
Fixes #1119, #1121 and #1128

### LINKED ISSUE #1119: DbaInstanceParameter must accept .
## Is this:

- [ ] *feature* 
- [x] *bug*:

##  Details

-SqlInstance . is valid but this chokes. It should handle . and not translate it, if possible. If a translation is required, then localhost will do. I think that . is named pipes whereas the rest are tcp

```
Invoke-DbaDiagnosticQuery : Cannot process argument transformation on parameter 'SqlInstance'. Cannot convert value "." to type "Sqlcollective.Dbatools.Parameter.DbaInstanceParameter[]". Error: "Cannot convert value "." to type
"Sqlcollective.Dbatools.Parameter.DbaInstanceParameter". Error: "Failed to parse instance name: .""
```

FriedrichWeinmann:
- DbaInstanceParameter knows "."
 - DbaSize knows -1 as unlimited
 - Dbadate, DbaDateTime and DbaTime have a new static method: Generate. It creates an actuall $null if the SQL Equivalent of null time was given (a datetime object with 0 ticks)

### DIFF
--- a/bin/library.ps1
+++ b/bin/library.ps1
@@ -2045,6 +2045,12 @@ namespace Sqlcollective.Dbatools
             {
                 InputObject = Name;
 
+                if (Name == ".")
+                {
+                    _ComputerName = Name;
+                    return;
+                }
+
                 string tempString = Name;
 
                 // Case: Default instance | Instance by port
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Straight up magic wizardry! :wizard:

PR #1133, opened 2017-05-01
Computermanagement v2

FriedrichWeinmann:
Alright, now after a long wait, here's the second iteration of the computer management framework.

This implements automatic connection caching, lots of customization, up to four paths to reach your target, credential management and lots of pre-validation and good performance (at least in my tests so far).

Caveat:
 - It currently can only retrieve classes or invoke queries.
 - It _cannot_ list namespace contents
 - it _cannot_ invoke methods
Not because I don't want to do this, but because I'd like the basic framework to have some mileage before I start into adding still more onto it. Complexity-creep is what's delayed this to the current point.

I fully expect to get back to these points before long - I know it to be not complex, but as I wrote, I'd prefer some more testing and live use before I add more complexity to it.

### DIFF
--- a/bin/library.ps1
+++ b/bin/library.ps1
@@ -450,6 +450,9 @@ namespace Sqlcollective.Dbatools
             /// <param name="Credential">The bad credential that must be punished</param>
             public void AddBadCredential(PSCredential Credential)
             {
+                if (DisableBadCredentialCache)
+                    return;
+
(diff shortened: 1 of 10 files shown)


potatoqualitee:
Super pumped! Will have a look tonight so we can get it in the dev branch and folks can start adjusting their commands and testing.

PR #1160, opened 2017-05-08
Debugging v2

FriedrichWeinmann:
- Introducing the developer mode (`Set-DbaConfig "developer.mode.enable" $true`)
 - Redesigned error logging (mostly behind-the-scenes stuff, smoother and mroe resilient)
 - Messaging system can handle multiple error records at a time
 - Exception text of passed-on exceptions will be passed along
 - Improved target object logging
 - New Command: `Get-DbatoolsLog` : Grants direct access to the in-memory log
 - New Command: `New-DbatoolsSupportPackage` : Creates a support file with lots of data helpful in troubleshooting stuff
 - New optional internal Command: `Expand-Archive` : Copy of the PSv5 version, available in PS v3+4
 - New optional internal Command: `Compress-Archive` : Copy of the PSv5 version, available in PS v3+4


ConstantineK:
Just starting to look over this commit, is the move of Invoke-SqlCmd2 from optional to internal also part of this? 

Just asking because I see a move in this commit but no reference to it in the PR/commit history notes for this part.

FriedrichWeinmann:
oh, that's internal housekeeping, keeping stuff in the folder where they should go.
Didn't comment upon it because it was a non-consequential change.

Historically, we's kept Invoke-Sqlcmd2 optional, because we didn't want to overwrite the original one if present. At that time it was a straight copy from the official repo and it allowed people to update that command independently from dbatools - if so desired - and have dbatools used the later version.
Then we did a change to the command so now it's mandatory, since we can't rely on users having the latest version otherwise (no version info on the script-file itself)

ConstantineK:
Sounds good!

potatoqualitee:
💃 💃 💃 💃

PR #1161, opened 2017-05-09
Remove DbaInstanceParameter

potatoqualitee:


potatoqualitee:
Removed all instances of DbaInstanceParameter  within connect and then the functions/*

PR #1187, opened 2017-05-11
New-DbaLogShippingPrimaryDatabase

sanderstad:
Function to save information for log shipping about the primary database


sanderstad:
This function executes the procedure "master.dbo.sp_add_log_shipping_primary_database " which does a couple of things:

- Add a backup job which will be disabled
- If not already created, create an alert job which will be enabled
- Set values in several tables. The following tables will be written:
SELECT * FROM msdb.dbo.log_shipping_monitor_primary
SELECT * FROM msdb.dbo.log_shipping_primary_databases
SELECT * FROM msdb.dbo.log_shipping_monitor_alert
SELECT * FROM msdb.dbo.log_shipping_primary_databases

Downside to testing this function is that the values cannot be easily removed. At least I don't know if a manual removal will be successful.
The way I removed it essentially is by executing the entire Invoke-LogShipping for that database and remove it through the GUI. This is something I'm going to look into.

If everything works out fine the results would look similar to this
result_new-dbalogshippingprimarydatabase

PR #1189, opened 2017-05-11
remove-dbamasterkey

potatoqualitee:


FriedrichWeinmann:
Gimme a sec and I'll do a live edit for reference

FriedrichWeinmann:
Done

potatoqualitee:
mmmm no, it'd be output at least, but i really think it should be a warning since its not doing anything that they intended to.

potatoqualitee:
ive kept your bottom changes but reverted it back to stop-function for what i'd personally like to see. otherwise, it feels as though it's silently failing.

FriedrichWeinmann:
You think? I mean we're talking here about potentially mass-spamming it just to be sure without any sure targeting ...
I'd rather work with output objects in that case with different states. Maybe make them optional for skipped entries (switch parameter to optin or optout)

potatoqualitee:
im less concerned about mass spamming than silently failing :/

potatoqualitee:
they can always use -Silent too?

FriedrichWeinmann:
Ne-va-yeh
Silent suppresses real errors in return for exceptions.
Maybe add a `-Lazy` parameter to make it use Verbose-Scope messages instead, if you're insistent that the default should be terrifying.

FriedrichWeinmann:
Better Idea:
Instead add a `-Mode` parameter with the options of Lazy, Strict or report (Strict is default)
Strict = Warning with Stop-Function
Lazy = Message in verbose
Report = Output objects so user can action on it

potatoqualitee:
No way, that's a whole infra change :(

potatoqualitee:
Silent or no

Issue #1143, opened 2017-05-05
Get-dbadatabasebackuphistory output issues


Stuart-Moore:
Output property backupsetupid should be backupsetid (when this is fixed please also look for the commented out fix in restore-dbadatabase) 

Filelist property isn't specified properly. Needs to return a hash table or an object as that's whats expected 

Both of these are prevent pipeline the output from this function into the restore chain. Have mitigated other issues in that side already.

FriedrichWeinmann:
```PowerShell
Get-DbaBackupHistory -SqlServer ProdServer1 -Databases Hr,Finance | Restore-DbaDatabase -SqlServer StageSrv2 -DestinationDataDirectory D:\Staging
```
Should restore the database. Currently throws errors on the file relocate part of the restore. Have worked around most of the other things with the output changes.

Stuart-Moore:
This issue also linked - https://github.com/sqlcollaborative/dbatools/issues/1196

FriedrichWeinmann:
@Stuart-Moore is this still an issue or superseded by the many changes since? (DM me in Slack if it's still ongoing, otherwise feel free to close it)

Stuart-Moore:
Nah, fixed a while ago. then fixed again last week. Test in CI build to catch it next time someone tinkers.

PR #1199, opened 2017-05-13
Fixing BackupHistory

### LINKED ISSUE #1196: Grouping issue with Get-DbaBackupHistory
Pulling @asharovarov  comment from PR 1186 as it wasn't fixed in that merge 
He asked:
if you are looking at backup history could you please check behavior for multi-file backups. I think it previously it was returning 1 record with multiple files. Now it returns a record for each file. Is it expected behavior?

That behavior leads to multiple MOVE clauses for the same DB file in Restore-DbaDatabase. It doesn't cause any issues. But doesn't look clean.

Stuart-Moore:
Fixes #1196 and #1143

I think I've modified the type definition right, and stuff looks 'righter' at my end. Fred need to confirm though!

And while I'm going this, whats the management view on the different return names? Easy enough to standardise, just which is the preferred one:
Raw:
Name
----
BackupSetID
checkpoint_lsn
Database
database_backup_lsn
DeviceType
Duration
End
first_lsn
FullName
last_lsn
MediaSetId
Path
position
Server
Software
software_major_version
Start
TotalSize
Type
Username

object:
Name
----
BackupSetId
CheckpointLsn
ComputerName
Database
DatabaseBackupLsn
DeviceType
Duration
End
FileList
FirstLsn
FullName
InstanceName
LastLsn
Path
Position
Software
SoftwareVersionMajor
SqlInstance
Start
TotalSize
Type
UserName

### DIFF
--- a/bin/library.ps1
+++ b/bin/library.ps1
@@ -1235,7 +1235,7 @@ namespace Sqlcollective.Dbatools
             /// <summary>
             /// Where is the backup stored
             /// </summary>
-            public string Path;
+            public string[] Path;
 
             /// <summary>
             /// What is the total size of the backup
@@ -1250,7 +1250,7 @@ namespace Sqlcollective.Dbatools
             /// <summary>
             /// The ID for the Backup job
             /// </summary>
-            public string BackupSetupId;
+            public string BackupSetId;
 
             /// <summary>
             /// What kind of backup-device was the backup stored to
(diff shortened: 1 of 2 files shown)


FriedrichWeinmann:
Looking good, but - as discussed - superseded by https://github.com/sqlcollaborative/dbatools/pull/1200

PR #1200, opened 2017-05-13
Get-DbaBackupHisotry - Fixing type trouble

FriedrichWeinmann:
Undid the dbadatetime issues, class update to better fit the actual needs.


FriedrichWeinmann:
Thanks for the confirm

PR #1214, opened 2017-05-15
Rename missed files

wsmelton:
@potatoqualitee Sent you vai DM in Slack the updated manifest files with what I found missing.

Changes proposed in this pull request, renamed a few more files
 - Just renamed the file CsvSqlImport.ps1 to match function name -> Import-DbaCsvToSql (also added Dba prefix to function name).
 - Rename of internal function vai manifest update Chrissy had done: Invoke-Dbacmd2 -> Invoke-DbaCmd (no reason to put the "2" since it is our version of the command now).
 - Renamed Get-DbaServerAuditSpecification -> Get-DbaSqlAuditSpecification
 - Renamed Get-DbaAudit -> Get-DbaSqlAudit

The last two are to simply match up to what we named the copy functions of the same objects.

How to test this code: 
- [ ] No code changes just renames.

### DIFF
--- a/functions/Get-DbaServerAudit.ps1
+++ b/functions/Get-DbaServerAudit.ps1
@@ -1,11 +1,11 @@
-FUNCTION Get-DbaSqlAudit
+FUNCTION Get-DbaServerAudit
 {
 <#
 .SYNOPSIS
 Gets SQL Security Audit information for each instance(s) of SQL Server.
 
 .DESCRIPTION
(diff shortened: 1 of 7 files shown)


potatoqualitee:
Thanks so much @wsmelton ! 

In addition to renaming Invoke-DbaSqlcmd, can you also update the pester names? They are causing failures.

PR #1215, opened 2017-05-15
New command New-DbaLogShippingPrimarySecondary

sanderstad:


sanderstad:
To test this function you need to have run the prerequisite function "New-DbaLogShippingPrimaryDatabase" to make sure the primary database is registered.

Use the following query to see if the values are written
`SELECT * FROM msdb.dbo.log_shipping_primary_secondaries`

The result should look something like the printscreen below:
result_new-dbalogshippingprimarysecondary

PR #1217, opened 2017-05-15
Get-dbadatabase status improvements

Stuart-Moore:
Fixes # 

 - Status parameter now works with comma combined statuses (express is bad for this ie; "normal, AutoClosed")
 - Allows multiple statuses to be filtered for. Wanted this for the continue restore functionality to get dbs in states that will allow more log restores (recovering, standby)
 - No other changes

How to test this code: 
- [ ] Get-DbaDatabase -SqlInstance localhost\sqlexpress2016 -Status standby,recovery
- [ ] 

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [x] Works remotely
- [x] Works locally

### DIFF
--- a/functions/Get-DbaDatabase.ps1
+++ b/functions/Get-DbaDatabase.ps1
@@ -99,6 +99,7 @@ Returns databases on multiple instances piped into the function
         [parameter(ParameterSetName = "Encrypted")]
         [switch]$Encrypted,
         [parameter(ParameterSetName = "Status")]
+        [ValidateSet('EmergencyMode', 'Normal', 'Offline', 'Recovering', 'Restoring', 'Standby', 'Suspect')]
         [string[]]$Status,
         [parameter(ParameterSetName = "Access")]
         [ValidateSet('ReadOnly', 'ReadWrite')]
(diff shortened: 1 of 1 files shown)


functions/Get-DbaDatabase.ps1
```diff
@@ -99,8 +99,7 @@ Returns databases on multiple instances piped into the function
         [parameter(ParameterSetName = "Encrypted")]
         [switch]$Encrypted,
         [parameter(ParameterSetName = "Status")]
-        [ValidateSet('EmergencyMode', 'Normal', 'Offline', 'Recovering', 'Restoring', 'Standby', 'Suspect')]
-        [string]$Status,
```
potatoqualitee: Is it possible to keep the validateset? It helps with tabbing. We could still keep the string[]

Stuart-Moore: It doesn't like that :( :
image

advice appears to be to user DynamicParameters, which are on the way out?

or enums? Are you happy they'd work far enough back?

potatoqualitee: DynamicParams are not on the way out, we're just doing them differently using TEPP, but I believe the answer would be 

```
[ValidateSet('EmergencyMode', 'Normal', 'Offline', 'Recovering', 'Restoring', 'Standby', 'Suspect')]
[string[]]$Status
```

Stuart-Moore: Apologies, something strange was going on in my session. That error was what I was getting with ValidateSet

Bounced box and it works now. Will push up

potatoqualitee:
sweet, thank you!

PR #1223, opened 2017-05-15
Added ExecutionMode

FriedrichWeinmann:
Added a new enumeration and its type accelerator-alias `[DbaMode]`
To be used when allowing the user to choose how to handle non-terminating issues.

Three modes have been implemented:
 - Strict: Terminate on non-critical error
 - Lazy: Continue on non-critical error
 - Report: Continue, but add a report object to the results. Should not be the default mode.

Ideally, `Report` should emit similar objects. That is, the output should be able to handle all items just the same with all objects having the same properties.
This may not always be possible though.

### DIFF
--- a/bin/library.ps1
+++ b/bin/library.ps1
@@ -2308,6 +2308,31 @@ namespace Sqlcollective.Dbatools
         }
     }
 
+    namespace General
+    {
+        /// <summary>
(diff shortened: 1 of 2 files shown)


potatoqualitee:
Thanks so much@ Can you update one command to show usage?

PR #1236, opened 2017-05-16
Cases - changed case for param, process, begin, end

potatoqualitee:


wsmelton:
Looks like you changed the case of "END" in some of the T-SQL statements, which that should be upper case to match the case of the statements.

wsmelton:
A few like `Get-DbaBackupHistory.ps1` shows you changed the column names that were `End` to `end`. So it  will miss-match the case of the property names, if the results of that query are returned.

potatoqualitee:
I'll redo these, thank you@ gonna close this

PR #1241, opened 2017-05-17
Move registering Tepp to function file

FriedrichWeinmann:
Redesign so that each function can register its own Tepp in its own file, rather than having to update everything in one central location.

 - Added function to do the registering
 - Implemented in Get-DbaBackupHistory as an example

Sorry, but in order to make that happen I had to reshuffle the loading order in the psm1


potatoqualitee:
Thank you!

Issue #1246, opened 2017-05-17
Get-DbaDatabaseFile, wrong properties returned 


niphlod:
## Is this:
- [ ] *feature* 
- [x] *bug*:

When querying for UsedSpace for database snapshots size_on_disk_bytes from dm_io_virtual_file_stats should be used  to discern the actual usage of the sparse file.

Also, for Size, Used and Available I think calcs are plain wrong for all databases (not only snapshots)

niphlod:
hint: the correct calcs to update for every database would be 

`$result.Size / 128.0 * 1024 * 1024`

more on the snapshot later... I'd need someone to test them on a 2005 instance

backintheday41:
Just adding verification that some calculations are incorrect:

    PS> Get-DbaDatabaseFile -SqlInstance Server1 -Databases Db1 | Select TypeDescription,Size
    
    TypeDescription Size    
    --------------- ----    
    ROWS            38.68 TB
    LOG             1.38 TB 
    
    PS> Get-DbaDatabase -SqlInstance Server1 -Databases Db1 | Select SizeMB
    
        SizeMB
        ------
    328143.9375

Get-DbaDatabase is returning the correct size.

dbaduck:
Also on the Size, it gets returned as a string, so when you want to sort on the Size you get a weird sort order.  I think it should come out as a number and the label should be Size_GB or something it should be something that is in the documentation that indicates that it will be in GB, but it should be a number so it can be sorted.  Thanks.

dbaduck:
I think that the Size property should act like the Get-DbaDiskSpace function has the data for the sizes.  Whenever I do a Format-Table the sizes are all right aligned and the header have inGB instead of having GB or MB in the string.  But even if you wanted to keep the GB or MB then it should be right justified so that it looks better and lines up. Thanks.

wsmelton:
@dbaduck The right alignment is default nature of PowerShell.

With regards to having the unit measurement in the output is part of the custom types in the module. If you check the code you will notice we set the [type of the output property](https://github.com/sqlcollaborative/dbatools/blob/development/functions/Get-DbaDatabaseFile.ps1#L213). That is as by design.

wsmelton:
However, @potatoqualitee @FriedrichWeinmann should note that this output prevents you from doing any comparison like:

```powershell
[...]

Issue #618, opened 2017-02-01
Restore-DbaFromDatabaseSnapshot - suggestions for more detailed output


potatoqualitee:
### Is this a *feature* OR *bug*:
ehancement

- can add date of snapshot to output?
- add original exception to notes? can add innerexception with more info?

![image](https://cloud.githubusercontent.com/assets/8278033/22490334/fbaa6d5e-e81c-11e6-8261-788ae6f29e10.png)

![image](https://cloud.githubusercontent.com/assets/8278033/22490368/36507098-e81d-11e6-97c2-534def0cb31e.png)

Can we also perhaps have a prompt to kick users out? this is why I couldn't restore my snap

PR #1253, opened 2017-05-17
Snapshot Management, first round

niphlod:
- implemented new messaging and flow control
- changed "server" to SqlInstance in output props
- adopted dbadatetime
- adopted dtainstanceparameter
- whatever I remembered from styleguide (surely I missed out something)

Things to check:
- [ ] TEPP
- [ ] Pipeline conventions from get-dbadatabasesnapshot (had to revert to a simple pscustomobject)
- [ ] Size property isn't really useful, SMO returns the size of the original database even for the snapshot. Get-DbaDatabaseFile could help there but I found out it needs further refinements ( #1246 ).

PS: I failed to repro #618 since it seems that remove-sqldatabase drops
the connections correctly. Also with the new messaging system I *think*
we don't wanna put exceptions in the 'Notes' Column anymore

### DIFF
--- a/functions/Get-DbaDatabaseSnapshot.ps1
+++ b/functions/Get-DbaDatabaseSnapshot.ps1
@@ -1,3 +1,4 @@
+#ValidationTags#FlowControl#
 Function Get-DbaDatabaseSnapshot
 {
 <#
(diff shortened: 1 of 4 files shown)


potatoqualitee:
thanks @niphlod - i see a few issues that I'll fix after the merge 👍

PR #1257, opened 2017-05-18
Test Redesign to pre-v2

FriedrichWeinmann:
Changes how the entire test suite works.
Primarily centralized all imports to avoid Import-Module/Export-Module conflicts.

### DIFF
--- a/tests/Export-DbaUser.Tests.ps1
+++ b/tests/Export-DbaUser.Tests.ps1
@@ -11,75 +11,36 @@
 
 #Thank you Warren http://ramblingcookiemonster.github.io/Testing-DSC-with-Pester-and-AppVeyor/
 
-if(-not $PSScriptRoot)
-{
+if (-not $PSScriptRoot) {
     $PSScriptRoot = Split-Path $MyInvocation.MyCommand.Path -Parent
 }
-$Verbose = @{}
-if($env:APPVEYOR_REPO_BRANCH -and $env:APPVEYOR_REPO_BRANCH -notlike "master")
-{
-    $Verbose.add("Verbose",$True)
+$Verbose = @{ }
+if ($env:APPVEYOR_REPO_BRANCH -and $env:APPVEYOR_REPO_BRANCH -notlike "master") {
+    $Verbose.add("Verbose", $True)
 }
-
-
 
 $sut = (Split-Path -Leaf $MyInvocation.MyCommand.Path).Replace('.Tests.', '.')
-Import-Module $PSScriptRoot\..\functions\$sut -Force
-Import-Module PSScriptAnalyzer
-## Added PSAvoidUsingPlainTextForPassword as credential is an object and therefore fails. 
-## We can ignore any rules here under special circumstances agreed by admins :-)
-## We expect some context using comments about the reason for ignoring a rule
-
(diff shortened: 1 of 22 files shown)


FriedrichWeinmann:
Note, this brings a few issues, which I'll address as the tests fail:
 - Help for internal functions is tested. That should be done anyway, but some of them don't have any.
 - Disabled all Script Analyzer Rule exclusions from the tests. Should not be disabled globally but on a per-file basis

Will see what fails and address as I go

potatoqualitee:
@SQLDBAWithABeard and @niphlod - says you recently edited these files. Can you please confirm that you understand what's going on going forward? 

Thank you!

potatoqualitee:
Thanks @niphlod - I'll be updating the tests so they don't fail. And start SqlServer and pass out that info so that ppl can make real tests 👍

niphlod:
@cl to make sure sql server(s) are ready you can borrow this snippet
https://github.com/niphlod/pydal/blob/master/appveyor.yml#L37

Il 18 mag 2017 12:51, "Chrissy LeMaire" <notifications@github.com> ha
scritto:

potatoqualitee:
Cool thank you! I'll add that

SQLDBAWithABeard:
Yes, is time the tests moved from the initial get the script analyser run to a more concrete method

potatoqualitee:
@SQLDBAWithABeard !! Look i wrote my first tests and theyr'e awesome! they perform actual restores from the appveyor-lab repo!!

image

image

SQLDBAWithABeard:
So cool, next you need to learn mocking and remove the warning from the output but check it is called and I would use -ea silentlycontinue to avoid the "sea of red"

potatoqualitee:
can you take a look at this [test](https://github.com/sqlcollaborative/dbatools/blob/testresolutions/tests/Restore-DbaDatabase.Tests.ps1) and tell me if it stayed within scope? i wanted it to be thorough but ended up sorta testing other commands, too.

SQLDBAWithABeard:
I see an error in your test - look at these two tests below using output from restore-dbdatabase

First one passes - second one doesnt

PS Presentations:\TUGAIT 2017 - Pester> Describe "one test" {
    It "Should pass" {
        ($results.Restorecomplete -contains $False) | should be $False
    }
        It "Should pass but wont" {
        $results.Restorecomplete | should be $true
    }
}


Describing one test
  [+] Should pass 97ms
  [-] Should pass but wont 37ms
    Expected: {True}
    But was:  {True True True True True True True True True True True True True True True True True True True True True True True Tr
ue True True True True True True True True True True True True True True True True True True True True True True True True True True
 True True True True True True True}
    at <ScriptBlock>, : line 6
    6:         $results.Restorecomplete | should be $true

potatoqualitee:
What's strange is that it depends on the version of PowerShell 👎 @gbargsley had the same problem you do, but it works perfectly on my version. How can i say "Shoudl not contain $false" i think i tried that and it scuked

SQLDBAWithABeard:
But the scope i fine for the integration test

SQLDBAWithABeard:
You can do it like I showed above 

It "Should pass" {
($results.Restorecomplete -contains $False) | should be $False
}

Pester contains only works for file contents so you have to use PowerShell contains and test for true or false

Good eh :-)

potatoqualitee:
i like that! thank you!!!!

PR #1262, opened 2017-05-18
Formatting help for 2 Copy functions

wsmelton:
Just formatted help in two commands that I missed. I will be doing this while updating the output now.


potatoqualitee:
Thank youuuu!

PR #1272, opened 2017-05-19
Restore quieter

Stuart-Moore:
Fixes some extraneous data that @potatoqualitee  noticed was being returned during a restore (issue raised on slack so no github ticket)


potatoqualitee:
thanks so much!

PR #1282, opened 2017-05-20
Read-DbaBackupHeader Update to 1.0.0 standards

FriedrichWeinmann:


potatoqualitee:
This is really pretty! Since we are gonna have the SqlInstance on all commands, can we just automatically process in psm1? Any functions\Copy-* would be source and destination then the rest would be sql instance. Just gotta standardize that in next release.

potatoqualitee:
Ill make a test for this today 👍

PR #1288, opened 2017-05-20
Dbainstanceparameter

potatoqualitee:
changed object and string to Dbainstanceparameter for source, destination and sqlinstance


potatoqualitee:
merged! we're looking so good. thanks for the kick in the pants, shawn!

PR #1289, opened 2017-05-20
singular params in examples

potatoqualitee:


potatoqualitee:
Singular!!!!!!! @wsmelton is working on putting the params back - they arent working right now since i ripped out dynamic params function

PR #1295, opened 2017-05-22
Initial commit of Get-MaintenanceSolutionLog

PowerDBAKlaas:
Read the performance of Ola's MaintenanceSolution from the log files.
For now, only IndexOptimize is included.

How to test this code: 
- [ ] 
- [ ] 

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally


potatoqualitee:
hey @PowerDBAKlaas ! Welcome back 😸 

This looks like an extremely useful command! I think, however, that it should be named  Get-MaintenanceSolutionLog and then have a switch like -Type where it gets specific logs. Otherwise, we have a super long name and we have a few of them. You can start out with only support for IndexOptimize.

Also, as we spoke about, This looks like a good command to use -Database. It should autopopulate - you simply have to add the parameter like a normal parameter.

PowerDBAKlaas:
It was my Original idea to have the -Type parameter.
But after the IndexOptimize, when I looked for the next interesting type, I found the others aren't very useful.
I think we have great alternatives for the Backup information, by @Stuart-Moore a.o.,
and also one for CHECKDB.
sp_purge_jobhistory only says: "140 entries deleted",
'Output File Cleanup' lists all deleted log files,
and sp_delete_backuphistory is an empty file on my systems.
So then I changed my opinion and the function name.
Do you see added value in any of the other log file types, @potatoqualitee @ClaudioESSilva @ConstantineK @niphlod @FriedrichWeinmann @SQLDBAWithABeard @wsmelton ?

wsmelton:
All of those write to the CommandLog table if it is set to true in the jobs (`@LogToTable='Y'`). I would also include that as a type. I rarely ever go and look at the SQL Agent output files because the table is easier to check for issues. I'd have to see an issue again to know if those Agent log files show any more details.

PowerDBAKlaas:
@wsmelton this is indeed a 'plan B' function. In cases where LogtoTable='N' (or was in the recent past), or where the logtable is cleaned up more frequently than the log files, or where there's more information needed than provided by logtable.
The latter seems to be true only for indexoptimize.
The files I'm reading now aren't the actual agent log files, but the log files from Ola's Stored Procedures.
The real SQL logs were a 3rd option, but as you say, not very likely to add anything.

Of course, a query on the log table is easier for most purposes. I just started with this one because I needed it myself, and didn't think of combining it.
Would it be best to include logtable queries in this same function? With a -sourcetype 'CommandLogTable' | 'LogFiles' ? If those files are only useful for indexoptimize, I think I'd rather keep them in separate functions, because the logic and choices will differ a lot depending on the sourcetype.
In that case that long name is sadly to stay, and another function Get-MaintenanceSolutionCommandLogTable could be the other one? @potatoqualitee ?

PowerDBAKlaas:
@potatoqualitee I looked at DB selection, but now I feel this one is not a good place to implement it; the file has to be read line by line from start to end anyway, so selecting fewer DBs doesn't result in less work. It's also harder to filter out subsequent lines that belong to a non-selected DB in the function, than it is to pipe to Where afterwards, on the output objects.

PowerDBAKlaas:
@potatoqualitee I had a merge conflict and by resolving I'm afraid I undid some of your changes.
sorry :-(
Can you check again, please?

potatoqualitee:
k but im overruling this name. if only one thing is useful, then it doesnt need to be qualified. Also, the length just kills me ;)

PowerDBAKlaas:
I know, if Ola would rename Maintenancesolution to sol_O, that would be so much easier 😆

potatoqualitee:
💃

PR #1312, opened 2017-05-25
Tests update

FriedrichWeinmann:
Massively overhaul & finetuning of tests.
 - No major structure changes
 - Updated tests to correctly call renamed functions
 - Did other troubleshooting of small things
 - Gave all empty files a default test, to make it easier to remember that something should be added


FriedrichWeinmann:
Does _not_ fix `Get-XpDirTreeRestoreFile.Tests.ps1`
And it worked flawlessly on the local machine, so ... why?!

PR #1314, opened 2017-05-25
Added missing -SqlCredential parameter

ClaudioESSilva:
Added 4 missing -SqlCredential parameters on Get-DbaBackupHistory inside calls
Without them is impossible to run the command, for example using -Last switch, with Sql Authentication


potatoqualitee:
$server is a connected SMO object and should not require sqlcredential. If that is not working as expected, we need to take a closer look at Connect-Sqlinstance. Lemme know.

ClaudioESSilva:
Everything works as expected. I messed with versions different functions. 👎  My bad. 😅 
Will delete the branch.

PR #1315, opened 2017-05-25
Help fixes

FriedrichWeinmann:
Resolves some bad help fixes:
 - Two instances of renamed parameter where the help was missed
 - Type validation when an array of an enumerated type was required


FriedrichWeinmann:
grml
Tests somehow don't work out as intended.
Giving it a rest for tonight, gonna continue on it this weekend.

PR #1322, opened 2017-05-27
Fix bug in RegEx replace for source and destination

wsmelton:
 - Change ` [Regex]::Escape("'$destination'")` to ` "'$destination'"` in all commands.

How to test this code: 
- [ ] None really unless you have an instance with special characters in the name (e.g. SomeServer\Instance.Name


potatoqualitee:
you are kicking ass! thank you, :shawn:

PR #1331, opened 2017-05-28
Implement TEPP parameter - Database (4 of x)

wsmelton:
Adjusting code for TEPP parameter, Database.

Tried adding the Snapshot parameter to TEPP but it isn't working from my machine. This is not just a straight SMO property, so maybe it is not supported?


FriedrichWeinmann:
Also appears to negatively impact `Restore-DbaDatabase`: Got one more failed test than usual

PR #1352, opened 2017-05-29
adding tepp for database

wsmelton:
Adjusting for Database parameters and some formatting.


potatoqualitee:
thank, shawn! one thing to note is that all tags should be singular. merging anyway cuz i'd like to consolidate all tags at a later date.

PR #1356, opened 2017-05-30
Fix Split method Errror on Resolve-DbaNetworkName

ClaudioESSilva:
Change from Object to DbaInstanceParameter
Added validation of object type and get the ComputerName correctly

thanks to @FriedrichWeinmann


potatoqualitee:
Thanks, @ClaudioESSilva ! Since the ComputerName is cast as a dbainstanceparameter, is it possible that that gettype check will ever fail? I think it was useful when it was [object] but now that it's not generic, that check is no longer needed.

PR #1357, opened 2017-05-30
Some improvements

ClaudioESSilva:
- Change from object to PSCredential
- Added $Silent parameter and changed all Write-Warning | Write-Verbose to Write-Message using new messaging system (Write-Message)
- Added support for mutiple instance.
- Added PSCustomObject as output

New output:
image


potatoqualitee:
fantasitc!

PR #1367, opened 2017-05-31
Smo Server Extension v2

FriedrichWeinmann:
New features:
 - Defaults to master
 - New optional parameter: All tables. Returns all tables instead of only the first one.

Breaking change:
 - Reordered the input order

Examples of usage:
Run query against master:
`$server.Query($query)`
Run query against other db:
`$server.Query($query, "db1")`
Run query against other db and return all tables:
`$server.Query($query, "db1", $true)`

Note:
There is no sane way to support default database and explicit `$AllTables`. Parameters in method calls are positional and the order is respected. I could work around this, but it'd be a hack causing trouble down the road.

### DIFF
--- a/internal/type-extensions.ps1
+++ b/internal/type-extensions.ps1
@@ -1,9 +1,13 @@
-# Implement query accelerator
+# Implement query accelerator for the server object
 Update-TypeData -TypeName Microsoft.SqlServer.Management.Smo.Server -MemberName Query -MemberType ScriptMethod -Value {
     Param (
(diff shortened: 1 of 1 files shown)


FriedrichWeinmann:
Oh, if you think it should default to all tables and have an option to return less, I can implement it that way, too ...

potatoqualitee:
💃 💃 💃 💃 💃

potatoqualitee:
this is super pretty, thank you!

PR #1368, opened 2017-05-31
Test-DbaSpn, local account fixes

### LINKED ISSUE #1366: Test-DbaSpn's output sometimes missing a variable
## Is this:

- [ ] *feature* 
- [x] *bug*:

## Details

```
WARNING: [Test-DbaSpn][02:05:07] AD lookup failure. This may be because the domain cannot be resolved for the SQL Server service account (NT Service\MSSQL$STANDARDRTM).
WARNING: [Test-DbaSpn][02:05:07] AD lookup failure. This may be because the domain cannot be resolved for the SQL Server service account ().
WARNING: [Test-DbaSpn][02:05:07] AD lookup failure. This may be because the domain cannot be resolved for the SQL Server service account (NT Service\MSSQL$STANDARDRTM).
WARNING: [Test-DbaSpn][02:05:07] AD lookup failure. This may be because the domain cannot be resolved for the SQL Server service account ().
```

Merci!

niphlod:
Fixes #1366 (or, it *should*)

Changes: 
- when the serviceaccount is detected as LocalSystem or is of type "NT SERVICE\xxx" replaces the object where to register the SPN with the computer object.

Clusters should be fine, 'cause - theoretically - they should not run on a local account. 

Remains to be seen how the cmdlet handles instances with a MSAs, but I don't have any available at hand to verify.

### DIFF
--- a/functions/Get-DbaSpn.ps1
+++ b/functions/Get-DbaSpn.ps1
@@ -66,9 +66,13 @@ Returns a custom object with SearchTerm (ServerName) and the SPNs that were foun
 			ForEach ($account in $AccountName)
 			{
 				Write-Message -Message "Looking for account $account..." -Level Verbose
+				$searchfor = 'User'
+				if($account.EndsWith('$')) {
+					$searchfor = 'Computer'
+				}
 				try
 				{
-					$Result = Get-DbaADObject -ADObject $account -Type User -Credential $Credential -Silent
+					$Result = Get-DbaADObject -ADObject $account -Type $searchfor -Credential $Credential -Silent
 				}
 				catch
 				{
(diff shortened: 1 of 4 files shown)


potatoqualitee:
merci! will test and patch.

potatoqualitee:
unfortunatelyl, didn't work 

image

i checked a few times to ensure I was using your changes. I think we should actually not even process or report on the "NT *" accounts. You?

niphlod:
why not ? in that case the SPN needs to be set on the computer object instead of the user, no biggie. The problem there seems that even with the mods the "change" in behaviour doesn't get triggered. 
The relevant line is https://github.com/sqlcollaborative/dbatools/pull/1368/files#diff-2d4ae1885405fbad08a4b571ad0cc6f0R282

niphlod:
whoopsie, found the reason

PR #1379, opened 2017-05-31
Fix problem with CHECKDB after restore not working

ClaudioESSilva:
- Added refresh when not finding database after the restore from the job.
Before: 
image

After:
image

- Added $Silent parameter to support messaging system (Write-Message)
- Change all Write-Warning | Write-Verbose | Write-Output to Write-Message system
- Added PSCustomObject as output
- Change NoDBCCCheck and added Alias to old NoCheck

### DIFF
--- a/functions/Remove-DbaDatabaseSafely.ps1
+++ b/functions/Remove-DbaDatabaseSafely.ps1
@@ -39,7 +39,7 @@ If specified this is the server that the Agent Jobs will be created on. By defau
 .PARAMETER Database
 The database name to remove or an array of database names eg $Database = 'DB1','DB2','DB3'
 
-.PARAMETER NoCheck
+.PARAMETER NoDBCCCheck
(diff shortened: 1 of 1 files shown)


functions/Remove-DbaDatabaseSafely.ps1
```diff
@@ -246,7 +250,7 @@ If there is a DBCC Error it will continue to perform rest of the actions and wil
 				}
 				
 				catch {
-					Write-Exception $_
+					Stop-Function -Message $_ -Continue
```
potatoqualitee: Stop-Function -Message "Error occured: $_" -Target $agentservice -InnerExceptionRecord $_

potatoqualitee: Oh, then  next line, return

potatoqualitee:
Hey @FriedrichWeinmann - can you take a look and update if I missed anything?

potatoqualitee:
I merged, but if you can do a PR for it if you have time. if not, nbd, its working for us currently.

PR #1386, opened 2017-06-02
Move the C# code in library.ps1 into a separate file

zippy1981:
 - If the C# code is read from a CS file we have the power of intellisense, and other static analysis.

How to test this code: 
- [ ] Make sure I didn't break anything
- [ ] 

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally


Stuart-Moore:
PRs should be based on development not master. 

Do you want to resubmit?

zippy1981:
@Stuart-Moore I edited it instead.

Stuart-Moore:
Thats also works

Asked Fred to review this as it's his baby

wsmelton:
If it is moved from the library file you will have to adjust everything that referenced it. Based on the appveyor this broke a lot of stuff.

zippy1981:
Interesting. I'll look at the appvoyer. I think this is very worthwhile to
have so I'll put some effort into doing it right before giving up.

wsmelton:
What is the benefit? Just for my inlightenment...

zippy1981:
The first benefit is it's a C# file, we can get intellisense  when we work
on it.

The second benefit is we can load it into Linqpad or a C# repl and debug it
easier.

The third benefit is we can a
Create a csproj with that file and some unit tests and test it with a C#
unit test framework. Then we can use the full power of visual studio,
resharper and all the static analysis tools out there.

FriedrichWeinmann:
Heya Zippy,

took a look and yeah, that's not gonna work. Totally not.
You can't use `$PSScriptRoot` to target the .cs file, since we don't invoke them as path, but by reading the text and invoking it from psm1. So it doesn't find the .cs file and that's just that.

For intellisense ... I don't know what tools you are using. My editor (PowerShell Studio) receives the the compiled .dll file and uses that (I compile it from my Visual Studio project).

What I am planning to do with the library is to:
 - Make it a more spread out project with multiple components
 - Compile it as part of the build process and ship the dll files as part of the finished module (which cuts down the initial module import time by a fair portion of what it currently takes)
 - By splitting it up, I make it easier to contribute to parts of it, significantly lowering the contribution barrier for the library.

However, this is not something to do quick&dirty, so I was looking for a sufficiently large timeslot to make it happen right.

I don't see the benefit in the meantime to split things up like you just did. If you need the source for your own intellisense, make a copy for your own use - it's not like it changes too frequently.

So, while I won't dictate it, I'm opposed to this PR, as it adds complexity without significant benefits you cannot realize locally for yourself without changing the project entire.

Sorry, but that's the way I feel about it.
Cheers,
Fred

wsmelton:
Code picks up intelligence for me just fine on the types. It is a bit delayed but has gotten better as the PS extension is been updated. 

If you are referring to intellisense for the C# code, that is a minor thing and I have that in Code as well with the right extension. 

Overall with what would have to be changed to the project as a whole,  I would have to agree with Fred that we don't touch this right now.

FriedrichWeinmann:
Hi Zippy,

we had a discussion on it and decided against the change proposed by your PR.
The basic consensus was, that adding complexity without functional improvement for the _user_ cannot be justified by a slight advantage for some developers, when this advantage can also be gained using other tools/methods.

Thank you for the idea anyway, as this lead to still some more discussions, sparking more ideas.

Cheers,
Fred

PR #1387, opened 2017-06-04
Fixed tepp for Snapshot parameter

wsmelton:
Odd enough in order to pass in a property where you need to filter, had to set as a variable and then pass in that variable to TEPP for it to pick up. Not sure why that is.

 - Adjusted files to accommodate snapshot parameter for tepp 
 - Fixed filter I missed seeing with Get-DbaDatabaseSnapshot as well.


potatoqualitee:
Good catch!

wsmelton:
Oddly enough the spelling error was not the fix. I had to set to a variable and pass it in that way. Only way it picks them up.

PR #1398, opened 2017-06-05
Get-DbaSsisEnvironment

BartekR:
Fixes # none

New function. Previous PR closed, this one uses functions mentioned during the reviev, uses foreach loops, objects, exclusions, Silence swith and probably some other neat things. Previously I put two functions in the PR, this time it's first of them, second on the way.

 - Get-DbaSsisEnvironment - gets SSIS environment variables, even if encrypted

How to test this code: 
- see EXAMPLES; also I have Pester integration tests, but not sure how to prepare the setup for the tests.

- [ ]  Powershell 3 - no (4)
- [ ]  Windows 7 - no (Windows Server 2012R2)
- [ ]  SQL Server 2000 - no (minimum required SQL Server 2012 - SSIS Catalog was introduced then)

- [ ]  SQL Server vNext - no
- [ ]  Windows 10 - no
- [ ]  Azure Database - no

Tests for tester:
- [ ] Works remotely
- [ ] Works locally


potatoqualitee:
thanks so much @BartekR ! I have assigned @PowerDBAKlaas to review this for you. Thank you, Klaas!

functions/Get-DbaSsisEnvironment.ps1
```diff
@@ -0,0 +1,229 @@
...
+                    #$ssisVariables = $connection.Databases['SSISDB'].ExecuteWithResults($sql).Tables[0]
+                    $ssisVariables = Invoke-DbaSqlCmd -ServerInstance $instance -Database SSISDB -Query $sql -As DataTable
+                    
+                    foreach($variable in $ssisVariables) {
+                        if($variable.sensitive -eq $true) {
+                            $value = $variable.decrypted
+                        } else {
+                            $value = $variable.value
+                        }
+                        [PSCustomObject]@{
```
PowerDBAKlaas: I don't know if you already have further plans, but if setting variables is possible, this function would better return the real variable objects.
You could do that with Add-member for each variable instead of PSCustomObject.

functions/Get-DbaSsisEnvironment.ps1
```diff
@@ -0,0 +1,229 @@
+<#
+.SYNOPSIS
```
PowerDBAKlaas: Actually, since we return one object per variable,
this function should be named Get-DbaSsisEnvironmentVariable.

functions/Get-DbaSsisEnvironment.ps1
```diff
@@ -0,0 +1,229 @@
...
+This command gets specified SSIS Environment and all its variables
+
+.DESCRIPTION
+This command gets all variables from specified environment from SSIS Catalog. All sensitive valus are decrypted.
+
+.PARAMETER SqlInstance
```
PowerDBAKlaas: ... where the SSISDB Catalog was created.

functions/Get-DbaSsisEnvironment.ps1
```diff
@@ -0,0 +1,229 @@
...
+            catch
+            {
+                Stop-Function -Message "Failed to connect to: $instance" -ErrorRecord $_ -Target $instance -Continue -Silent $Silent
+                return
+            }
+            
+
+            if ($connection.versionMajor -lt 11)
+            {
+                Stop-Function -Message "SSISDB catalog is only available on Sql Server 2012 and above, exiting." -Silent $Silent
```
PowerDBAKlaas: it should not exit, but continue to the next sql instance

functions/Get-DbaSsisEnvironment.ps1
```diff
@@ -0,0 +1,229 @@
...
+                Write-Message -Message "Connecting to $instance Integration Services" -Level 5 -Silent $Silent
+                $SSIS = New-Object "$ISNamespace.IntegrationServices" $connection
+            }
+            catch
+            {
+                Stop-Function -Message "Could not connect to Integration Services on $instance" -Silent $Silent
+                return
+            }
+
+            Write-Message -Message "Fetching SSIS Catalog and its folders" -Level 5 -Silent $Silent
```
PowerDBAKlaas: Let's call it SSISDB Catalog consistently.

functions/Get-DbaSsisEnvironment.ps1
```diff
@@ -0,0 +1,229 @@
...
+            $catalog = $SSIS.Catalogs | Where-Object { $_.Name -eq "SSISDB" }
+
+            # get all folders names if none provided
+            if($null -eq $Folder) {$Folder = $catalog.Folders.Name}
+
+            # filter unwanted folders
+            if ($FolderExclude) {
```
PowerDBAKlaas: I think the logic on 'Folder', 'FolderExclude', 'Environment' and 'EnvironmentExclude' is much more complex.
To catch all possibilities, it probably needs parameter sets, or we should choose 2 instead of 4 parameters.

If a user enters the same value in -Folder and -FolderExclude -> we will arrive at ```foreach ($f in $folder)```, but $folder will be empty,
If a user enters an environmentname in -Environment and multiple folders in -Folder, no environments will be returned from the folder that does not contain those, while the user may expect to see all environments,...

In short, I think we need to catch an empty collection,
and also explain in the examples and description that those parameters act as additive filters and only environments that pass all filters will be returned.

functions/Get-DbaSsisEnvironment.ps1
```diff
@@ -0,0 +1,229 @@
...
+
+    }
+
+    process {
+
+        foreach($instance in $SqlInstance)
+        {
+            try
+            {
+                Write-Message -Message "Connecting to $instance" -Level 5 -Silent $Silent
```
wsmelton: You can remove the `-Silent $Silent` from all the Write-Message and Stop-Function calls. This will get picked up by default if `-Silent` is passed in by the calling function.

functions/Get-DbaSsisEnvironment.ps1
BartekR: My mistake. Should be 'SSIS Catalog', will change it and use 'SSIS Catalog' everywhere as it's the name one can find in docs

functions/Get-DbaSsisEnvironment.ps1
BartekR: Removed. Previously I had some problems with Pester tests when I omitted `-Silent` switch but tested now and looks OK.

wsmelton: Did you write a pester test for this function? If so would be awesome to include in your PR.

potatoqualitee:
hey @PowerDBAKlaas  - can you take a look at @BartekR's changes and approve? then we can merge 💃

BartekR:
I'm in contact with Klaas. Wee agreed that when I finish the tests I will let him know and he will check it. I will finish them in max two days.

potatoqualitee:
oh, awesome on both counts :D looking forward to it! thank you both

functions/Get-DbaSsisEnvironmentVariable.ps1
```diff
@@ -0,0 +1,281 @@
+<#
+.SYNOPSIS
+This command gets specified SSIS Environment and all its variables
+
+.DESCRIPTION
+This command gets all variables from specified environment from SSIS Catalog. All sensitive valus are decrypted.
+The function communicates directly with SSISDB database, "SQL Server Intergation Services" service isn't queried there.
+Each parameter (besides SqlInstance and SqlCredential) acts as the filter to only include or exclude particular element
+
+.PARAMETER SqlInstance
```
PowerDBAKlaas: from the parameter definition I see multiple sql instances are accepted after all.
Can you make that clear in the parameter help than, please?

functions/Get-DbaSsisEnvironmentVariable.ps1
```diff
@@ -0,0 +1,281 @@
...
+                                {
+                                    $value = $variable.decrypted
+                                }
+                                else
+                                {
+                                    $value = $variable.value
+                                }
+
+                                [PSCustomObject]@{
+                                    Server          = $instance
```
PowerDBAKlaas: Can you add 
- ComputerName
- InstanceName
- SqlInstance
instead of Server, please?
Those are our standards. You'll find them in most Get- functions.

functions/Get-DbaSsisEnvironmentVariable.ps1
```diff
@@ -8,7 +8,7 @@ The function communicates directly with SSISDB database, "SQL Server Intergation
 Each parameter (besides SqlInstance and SqlCredential) acts as the filter to only include or exclude particular element
 
 .PARAMETER SqlInstance
-The SQL Server instance.
+The SQL Server instance or set of instances separated by commas
```
PowerDBAKlaas: commas is only when typed in directly.
a user can send multiple instances in a variable or a subexpression
so I wouldn't put it like this

potatoqualitee:
Hell yes, thanks for the approval, @PowerDBAKlaas ! And thanks for the command, @BartekR ! Welcome to the team 💃

PR #1399, opened 2017-06-05
Added new Get-DbaAgentSchedule command

DevOpsFu:
Fixes # 

 - Addition of new `Get-DbaAgentSchedule` command

How to test this code: 
- [ ] See the command documentation for usage

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [x]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally

### DIFF
--- a/functions/Get-DbaAgentSchedule.ps1
+++ b/functions/Get-DbaAgentSchedule.ps1
@@ -47,12 +47,12 @@ FUNCTION Get-DbaAgentSchedule
 	
 	process {
 		foreach ($instance in $SqlInstance) {
-			Write-Verbose "Attempting to connect to $instance"
+			Write-Message -Level Verbose -Message "Attempting to connect to $instance"
 			try {
 				$server = Connect-SqlServer -SqlServer $instance -SqlCredential $SqlCredential
 			}
 			catch {
-				Write-Warning "Can't connect to $instance or access denied. Skipping."
+				Write-Message -Level Warning -Message "Can't connect to $instance or access denied. Skipping."
 				continue
 			}
 
--- a/functions/Get-DbaAgentSchedule.ps1
+++ b/functions/Get-DbaAgentSchedule.ps1
@@ -1,0 +1,78 @@
+FUNCTION Get-DbaAgentSchedule
+{
+<#
+	.SYNOPSIS
+	Returns all SQL Agent Shared Schedules on a SQL Server Agent.
+
+	.DESCRIPTION
+	This function returns SQL Agent Shared Schedules.
+
+	.PARAMETER SqlInstance
+	SqlInstance name or SMO object representing the SQL Server to connect to.
+	This can be a collection and receive pipeline input.
+
+	.PARAMETER SqlCredential
+	PSCredential object to connect as. If not specified, currend Windows login will be used.
+
+	.PARAMETER Silent
+	Use this switch to disable any kind of verbose messages
+
+	.NOTES
+	Author: Chris McKeown (@devopsfu), http://www.devopsfu.com
+	Tags: Agent, Schedule
+	Website: https://dbatools.io
+	Copyright: (C) Chrissy LeMaire, clemaire@gmail.com
+	License: GNU GPL v3 https://opensource.org/licenses/GPL-3.0
+
+	.LINK
+	https://dbatools.io/Get-DbaAgentSchedule
+
+	.EXAMPLE
+	Get-DbaAgentSchedule -SqlInstance localhost
+	Returns all SQL Agent Shared Schedules on the local default SQL Server instance
+
+	.EXAMPLE
+	Get-DbaAgentSchedule -SqlInstance localhost, sql2016
+	Returns all SQL Agent Shared Schedules for the local and sql2016 SQL Server instances
+
+#>
+	[CmdletBinding()]
+	Param (
+		[parameter(Position = 0, Mandatory = $true, ValueFromPipeline = $True)]
+		[Alias("ServerInstance", "Instance", "SqlServer")]
+		[object[]]$SqlInstance,
+		[System.Management.Automation.PSCredential]$SqlCredential,
+		[switch]$Silent
+	)
+	
+	process {
+		foreach ($instance in $SqlInstance) {
+			Write-Verbose "Attempting to connect to $instance"
+			try {
+				$server = Connect-SqlServer -SqlServer $instance -SqlCredential $SqlCredential
+			}
+			catch {
+				Write-Warning "Can't connect to $instance or access denied. Skipping."
+				continue
+			}
+
+			Write-Message -Level Verbose -Message "Getting Edition from $server"
+			Write-Message -Level Verbose -Message "$server is a $($server.Edition)"
+			
+			if ($server.Edition -like 'Express*') {
+				Stop-Function -Message "There is no SQL Agent on $server, it's a $($server.Edition)" -Continue
+			}
+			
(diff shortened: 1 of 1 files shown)


potatoqualitee:
@PowerDBAKlaas  - please confirm command name, I get confused ;)

functions/Get-DbaAgentSchedule.ps1
```diff
@@ -0,0 +1,78 @@
...
+	)
+	
+	process {
+		foreach ($instance in $SqlInstance) {
+			Write-Verbose "Attempting to connect to $instance"
+			try {
+				$server = Connect-SqlServer -SqlServer $instance -SqlCredential $SqlCredential
+			}
+			catch {
+				Write-Warning "Can't connect to $instance or access denied. Skipping."
```
wsmelton: This should be `Write-Message -Level Warning -Message ""`

functions/Get-DbaAgentSchedule.ps1
```diff
@@ -0,0 +1,78 @@
+FUNCTION Get-DbaAgentSchedule
```
wsmelton: This should be `Get-DbaAgentSharedSchedule` as that is what you are pulling.

DevOpsFu: @wsmelton I wonder if we should consider renaming `New-DbaAgentSchedule` then - if you create a schedule using that command, it comes back in the collection retrieved by this one. Having said that, the `New-DbaAgentSchedule` also seems to require and attach itself to a job even though it says it doesn't do the latter in the synopsis. I need the ability to create a shared schedule without referencing a job, so I might add some more stuff to this PR purely based on shared schedules

PowerDBAKlaas: @potatoqualitee @DevOpsFu @wsmelton 
There are 2 different type of schedules in SQLServer:
jobserver.sharedschedule
jobserver.job.jobschedule
Given there are also other schedules outside, it's important to clearly name them.
As @wsmelton  says, `Get-DbaAgentSharedSchedule` for this one (if the New- also works on this type, indeed that needs to be `New-DbaAgentSharedSchedule`, @DevOpsFu )
Another function should be named `Get-DbaAgentJobSchedule`, isn't Sander working on that?

potatoqualitee: thanks @PowerDBAKlaas - please remember to rename the file as well, @DevOpsFu. @sanderstad  is not working on Gets so this is perfect 👍

DevOpsFu: Will do - I'll add the equivalent `get-` and `set-` forms of the `DbaSharedSchedule` commands also. Should I close this PR for now or leave it and just update it?

DevOpsFu: Actually the schedules created by `New-DbaAgentSchedule` (which requires a Job param to associate the schedule with) are returned by the call to `JobServer.SharedSchedules` so it seems that they're being created as shared schedules anyway, even if only one job references them. Digging further into it, I'm not sure if it's possible to create a schedule that _can't_ be shared.

potatoqualitee: Yep, just tested it and the job schedule is pointing to the shared collection

````
PS C:\WINDOWS\system32> $server.jobserver.jobs.jobschedules |Get-Member


  TypeName: Microsoft.SqlServer.Management.Smo.Agent.JobSchedule

[1:58] 
PS C:\WINDOWS\system32> $server.jobserver.SharedSchedules | Get-Member


  TypeName: Microsoft.SqlServer.Management.Smo.Agent.JobSchedule
````

Comment (PowerDBAKlaas) [inline functions/Get-DbaAgentSchedule.ps1]: @DevOpsFu so, is it certain that the collection of all jobserver.job.schedules is exactly the same as jobserver.sharedschedules?

@potatoqualitee in that case, it should stay Get-DbaAgentSchedule and I think this PR is fine.
@wsmelton  do you agree?

I also guess we have the complete set of functions on schedules then, not?
We have New- Get- Set- Remove- and no renames are needed; DbaAgentSchedule covers all.
@sanderstad can you confirm that? Do any of your functions need adaptations?

Comment (sanderstad) [inline functions/Get-DbaAgentSchedule.ps1]: For the schedule smo there is no distinction between a schedule and a shared schedule. I can easily create a schedule and add it to another job based on the ID. 

The code below will use two existing jobs called "Job1" an "Job2" to add a schedule to. 
```
Import-Module SqlServer

# Create the server object
$server = New-Object Microsoft.SqlServer.Management.Smo.Server '.'

# Set some names
$jobName1 = 'Job1'
$jobName2 = 'Job2'
$scheduleName = 'TestSharedSchedule'

# Get the jobs
$smoJob1 = $Server.JobServer.Jobs[$jobName1]
$smoJob2 = $Server.JobServer.Jobs[$jobName2]

# Get the count for the schedules
$smoJob1.JobSchedules.Count
$smoJob2.JobSchedules.Count

# Create the job schedule
$smoSchedule = New-Object Microsoft.SqlServer.Management.Smo.Agent.JobSchedule($smoJob1, $scheduleName)

# Set some values for the schedule
$smoSchedule.ActiveStartDate = Get-Date
$smoSchedule.ActiveEndDate = '9999-12-31'
$smoSchedule.FrequencyTypes = "Daily"
$smoSchedule.FrequencyInterval = 1
$smoSchedule.ActiveStartTimeOfDay = 230000
$smoSchedule.ActiveEndTimeOfDay = 235959

# Create the schedule for the job
$smoSchedule.Create()

$smoJob2.AddSharedSchedule($smoSchedule.ID)

Write-Host "`nNormal schedules`n" -ForegroundColor Cyan
$($server.JobServer.Jobs.JobSchedules.Name)

Write-Host "`nShared Schedules`n" -ForegroundColor Cyan
$($server.JobServer.SharedSchedules.Name)
```

At first the jobs have no schedule:
image

After the script:
image

What we might want to do is add a feature to the set-dbaagentjob function to add a shared schedule based on the ID or the name of the schedule. 

I don't think the other functions need to be changed.

DevOpsFu: Ultimately, what I'm driving at is being able to create a new schedule without having to specify which job it's associated with. I've found that this can be done by specifying the `JobServer` object as the parent object when creating a new schedule, rather than having to specify a Job object.

In `[New-DbaAgentSchedule](https://github.com/sqlcollaborative/dbatools/blob/master/functions/New-DbaAgentSchedule.ps1#L7)` it specifies that creating a schedule does not attach it to the job - this is actually not the case; creating a new schedule using this function does associate the schedule with the referenced job.

I see two options here - either make the `Job` parameter optional (and if it isn't specified then create the schedule with the `JobServer` as the parent), or split the functions into `New-DbaAgentSchedule` and `New-DbaAgentJobSchedule` , for example.

What are your thoughts? I'm conscious that this conversation is getting a bit long - feel fee to move it into a Slack channel if you prefer?

functions/Get-DbaAgentSchedule.ps1
DevOpsFu: @wsmelton Corrected.

functions/Get-DbaAgentSchedule.ps1
DevOpsFu: @sanderstad and I spoke via Slack earlier. He'll be modifying the other `*-DbaAgentSchedule` commands to allow creation of a schedule without an associated job.

I've also addressed @wsmelton 's comment. Given that, can this be merged?

potatoqualitee: Yep! Thanks, so much and welcome to the team, @DevOpsFu! Thank you for your awesome code 💯

PR #1400, opened 2017-06-06
TEPP for Job parameter

wsmelton:
Add TEPP and format for Job parameter. A few I added Login parameter so we could have TEPP with it as well.


potatoqualitee:
currrent job, good call!

PR #1407, opened 2017-06-07
DbaAgent functions - work without job values

sanderstad:
New-DbaAgentJob was cleaned in the parameter section. Two parameters are
added (schedule and scheduleid) to make it possible to attach a shared schedule when the new job gets created. The return of the object was also improved.

New-DbaAgentSchedule got a cleanup in the parameter section. It not longer needs a job to create a schedule. The job parameters can still be used to assign the schedule to a job.

Remove-DbaAgentJob cleanup in the parameter section and the job parameter was set to be an object type instead of a string.

Remove-DbaAgentSchedule had the help improved and a cleanup in the
parameter section. It not longer relies on the job and has had a full rewrite to support that.

Set-DbaAgentJob had a cleanup in the parameter section.  Two parameters
are added (schedule and scheduleid) to make it possible to attach a shared schedule to a job.

### DIFF
--- a/functions/New-DbaAgentJob.ps1
+++ b/functions/New-DbaAgentJob.ps1
@@ -133,7 +133,7 @@ Creates a job with the name "Job One" on multiple servers using the pipe line
     param (
         [parameter(Mandatory = $true, ValueFromPipeline = $true)]
         [Alias("ServerInstance", "SqlServer")]
-        [object[]]$SqlInstance,
+        [DbaInstanceParameter[]]$SqlInstance,
 
         [Parameter(Mandatory = $false)]
         [System.Management.Automation.PSCredential]$SqlCredential,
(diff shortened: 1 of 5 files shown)


functions/New-DbaAgentJob.ps1
```diff
@@ -379,16 +410,23 @@ Creates a job with the name "Job One" on multiple servers using the pipe line
                     # Make sure the target is set for the job
                     Write-Message -Message "Applying the target (local) to job $Job" -Level Verbose
                     $smoJob.ApplyToTargetServer("(local)")
```
sanderstad: This is done to set a target for the job. Only in the case of multi server administration with the agent will this need to be changed.
It doesn't have anything to do with the connection to the sql instance for the creation of that job

DevOpsFu:
**Tested and confirmed working:**
- New-DbaAgentSchedule with no job param supplied, to create a shared schedule not attached to any jobs
- New-DbaAgentSchedule with a job param to attach a new schedule to an existing job
- New-DbaAgentJob with no associated schedule
- New-DbaAgentJob with an associated schedule, referenced by Name

**Tests failed:**
- New-DbaAgentJob referencing an existing schedule by ID (the same schedule as referenced by name above). When adding by name, I see that it correctly resolves the schedule ID:

> VERBOSE: [Set-DbaAgentJob][18:12:19] Adding schedule id 16 to job

but if I pass `ScheduleId 16` into `New-DbaAgentJob` then I get:

> WARNING: [Set-DbaAgentJob][18:16:13] Schedule 16 cannot be found on instance...

wsmelton:
You may need to refresh your connection or the job server object. We have seen that with some other commands.

DevOpsFu:
@wsmelton Yep, I've had this issue using the objects directly. @sanderstad could we insert a call to refresh() before any job retrieval operations?

potatoqualitee:
hey @sanderstad - can you add @DevOpsFu's request and we'll merge and test for this week's version?

DevOpsFu:
@sanderstad This still doesn't work for me after creating a new schedule.

PR #1415, opened 2017-06-07
Get-DbaCmObject - Bugfix WMI with Query

FriedrichWeinmann:
WMI used to only work with object requests, not query


potatoqualitee:
💃 💃 💃 💃

PR #1416, opened 2017-06-08
Resolve-DbaNetworkName Adjusted ComputerName to be 1.0 compliant

wsmelton:
- Adjusted ComputerName to be a string instead of the DbaInstanceParameter type. If you pass in a computer that does not exist it will throw an error that it cannot convert the value and then can't parse the instance name. Then output the expected warning messages.

- Adjusted the alias to match the example, SqlServer to SqlInstance.

- Removed all the CIM variations and put in `Get-DbaCmObject`...think I did it right but should be checked closely.

- Added messaging system

- Removed `$FunctionName` since messaging system handles that in output.

### DIFF
--- a/functions/Resolve-DbaNetworkName.ps1
+++ b/functions/Resolve-DbaNetworkName.ps1
@@ -1,94 +1,84 @@
-Function Resolve-DbaNetworkName
-{
-  <#
-      .SYNOPSIS
-      Returns information about the network connection of the target computer including NetBIOS name, IP Address, domain name and fully qualified domain name (FQDN).
+function Resolve-DbaNetworkName {
+	<#
+	  .SYNOPSIS
+		Returns information about the network connection of the target computer including NetBIOS name, IP Address, domain name and fully qualified domain name (FQDN).
 
-      .DESCRIPTION
-      Retrieves the IPAddress, ComputerName from one computer.
-      The object can be used to take action against its name or IPAddress.
+	  .DESCRIPTION
+		Retrieves the IPAddress, ComputerName from one computer.
+		The object can be used to take action against its name or IPAddress.
 
-      First ICMP is used to test the connection, and get the connected IPAddress.
+		First ICMP is used to test the connection, and get the connected IPAddress.
 
-      If your local Powershell version is not higher than 2, WMI is tried to get the computername.
-      If not, CIM is used, first via WinRM, and if not successful, via DCOM.
+		Multiple protocols (e.g. WMI, CIM, etc) are attempted before giving up.
 
-      .PARAMETER ComputerName
-      The Server that you're connecting to.
-      This can be the name of a computer, a SMO object, an IP address or a SQL Instance.
+	  .PARAMETER ComputerName
+		The Server that you're connecting to.
+		This can be the name of a computer, a SMO object, an IP address or a SQL Instance.
 
-      .PARAMETER Credential
-      Credential object used to connect to the SQL Server as a different user
+	  .PARAMETER Credential
+		Credential object used to connect to the SQL Server as a different user
 
-      .PARAMETER Turbo
-      Resolves without accessing the serer itself. Faster but may be less accurate.
+	  .PARAMETER Turbo
+		Resolves without accessing the serer itself. Faster but may be less accurate.
 
-      .PARAMETER Silent
-      Use this switch to disable any kind of verbose messages.
+	  .PARAMETER Silent
+		Use this switch to disable any kind of verbose messages.
 
-      .NOTES
-      Author: Klaas Vandenberghe ( @PowerDBAKlaas )
+	  .NOTES
+	  	Tags: Network, Resolve
+		Original Author: Klaas Vandenberghe ( @PowerDBAKlaas )
 
-      dbatools PowerShell module (https://dbatools.io)
-      Copyright (C) 2016 Chrissy LeMaire
-      This program is free software: you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, either version 3 of the License, or (at your option) any later version.
-      This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details.
-      You should have received a copy of the GNU General Public License along with this program. If not, see http://www.gnu.org/licenses/.
+		Website: https://dbatools.io
+		Copyright: (C) Chrissy LeMaire, clemaire@gmail.com
+		License: GNU GPL v3 https://opensource.org/licenses/GPL-3.0
 
-      .LINK
-      https://dbatools.io/Resolve-DbaNetworkName
+	  .LINK
+		https://dbatools.io/Resolve-DbaNetworkName
 
-      .EXAMPLE
-      Resolve-DbaNetworkName -ComputerName ServerA
+	  .EXAMPLE
+		Resolve-DbaNetworkName -ComputerName ServerA
 
-      Returns a custom object displaying InputName, ComputerName, IPAddress, DNSHostName, Domain, FQDN for ServerA
-	
-      .EXAMPLE
-      Resolve-DbaNetworkName -SqlInstance sql2016\sqlexpress
+		Returns a custom object displaying InputName, ComputerName, IPAddress, DNSHostName, Domain, FQDN for ServerA
 
-      Returns a custom object displaying InputName, ComputerName, IPAddress, DNSHostName, Domain, FQDN for the SQL instance sql2016\sqlexpress
-	
-      .EXAMPLE
-      Resolve-DbaNetworkName -SqlInstance sql2016\sqlexpress, sql2014
+	  .EXAMPLE
+		Resolve-DbaNetworkName -SqlInstance sql2016\sqlexpress
 
-      Returns a custom object displaying InputName, ComputerName, IPAddress, DNSHostName, Domain, FQDN for the SQL instance sql2016\sqlexpress and sql2014
+		Returns a custom object displaying InputName, ComputerName, IPAddress, DNSHostName, Domain, FQDN for the SQL instance sql2016\sqlexpress
 
-      Get-SqlRegisteredServerName -SqlInstance sql2014 | Resolve-DbaNetworkName
-	
-      Returns a custom object displaying InputName, ComputerName, IPAddress, DNSHostName, Domain, FQDN for all SQL Servers returned by Get-SqlRegisteredServerName
+	  .EXAMPLE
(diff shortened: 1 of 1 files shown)


functions/Resolve-DbaNetworkName.ps1
```diff
@@ -1,90 +1,84 @@
...
+		Returns a custom object displaying InputName, ComputerName, IPAddress, DNSHostName, Domain, FQDN for all SQL Servers returned by Get-SqlRegisteredServerName
   #>
 	[CmdletBinding()]
 	param (
 		[parameter(ValueFromPipeline)]
-		[Alias('cn', 'host', 'ServerInstance', 'Server', 'SqlServer')]
-		[DbaInstanceParameter[]]$ComputerName = $env:COMPUTERNAME,
-    	[PSCredential] [System.Management.Automation.CredentialAttribute()]$Credential,
+		[Alias('cn', 'host', 'ServerInstance', 'Server', 'SqlInstance')]
+		[string[]]$ComputerName = $env:COMPUTERNAME,
```
FriedrichWeinmann: Strongly disagree with this change.
Removes a lot of flexibility of usage and constitutes a breaking change!

functions/Resolve-DbaNetworkName.ps1
```diff
@@ -1,90 +1,84 @@
...
+	process {
+		foreach ($Computer in $ComputerName) {
 			$conn = $ipaddress = $CIMsession = $null
-			
-			if ($Computer.GetType() -eq [Microsoft.SqlServer.Management.Smo.Server])
-			{
+
+			if ($Computer.GetType() -eq [Microsoft.SqlServer.Management.Smo.Server]) {
 				$Computer = $Computer.NetName
 			}
```
FriedrichWeinmann: This case will never happen, can be removed

functions/Resolve-DbaNetworkName.ps1
```diff
@@ -1,90 +1,84 @@
...
-			{
+
+			if ($Computer -eq 'localhost' -or $Computer -eq '.') {
 				$Computer = $env:COMPUTERNAME
 			}
-			
+
 			if ($Computer.GetType() -eq [Sqlcollective.Dbatools.Parameter.DbaInstanceParameter]) {
 				$Computer = $Computer.ComputerName
 			}
```
FriedrichWeinmann: Either it will never have this type (right now) or it will always be of that type.
This is residue from a previous hotfix to keep it working and can be removed

functions/Resolve-DbaNetworkName.ps1
```diff
@@ -93,175 +87,127 @@ Function Resolve-DbaNetworkName
 				$Computer = ($Computer -Split ('\:'))[0]
 				$Computer = ($Computer.Split('\,'))[0]
```
FriedrichWeinmann: This will not be necessary with the DbaInstanceParameter class as expected type

functions/Resolve-DbaNetworkName.ps1
```diff
@@ -93,175 +87,127 @@ Function Resolve-DbaNetworkName
...
 						$resolved = [System.Net.Dns]::GetHostEntry($Computer)
 						$ipaddress = $resolved.AddressList[0].IPAddressToString
 						$fqdn = $resolved.HostName
 					}
-					catch
-					{
-						Write-Warning "$functionName - DNS name not found"
+					catch {
+						Write-Message -Level Warning -Message "DNS name not found"
 						continue
```
FriedrichWeinmann: Wouldn't this be a case to use `Stop-Function`?

functions/Resolve-DbaNetworkName.ps1
```diff
@@ -93,175 +87,127 @@ Function Resolve-DbaNetworkName
...
-					{
-						$CIMsession = New-CimSession -ComputerName $Computer -ErrorAction Stop -Credential $Credential
-						$conn = Get-CimInstance -Query "Select * FROM Win32_computersystem" -CimSession $CIMsession -ErrorAction Stop
+
+			if ($host.Version.Major -gt 2) {
+				Write-Message -Level Verbose -Message "Your PowerShell Version is $($host.Version.Major)"
+				try {
+					Write-Message -Level Verbose -Message "Getting computer information from $Computer"
+					if (Was-Bound "Credential") {
+						$conn = Get-DbaCmObject -ClassName win32_ComputerSystem -Computer $Computer -Credential $Credential
```
FriedrichWeinmann: Add the `-Silent` switch if you want to catch something

functions/Resolve-DbaNetworkName.ps1
```diff
@@ -93,175 +87,127 @@ Function Resolve-DbaNetworkName
...
 					}
-					else
-					{
-						$conn = Get-CimInstance -Query "Select * FROM Win32_computersystem" -ComputerName $Computer -ErrorAction Stop
+					else {
+						$conn = Get-DbaCmObject -ClassName win32_ComputerSystem -Computer $Computer
```
FriedrichWeinmann: Add the `-Silent` switch if you want to catch something

functions/Resolve-DbaNetworkName.ps1
```diff
@@ -93,175 +87,127 @@ Function Resolve-DbaNetworkName
...
-							Domain = $fqdn.Replace("$hostname.", "")
+							Domain      = $fqdn.Replace("$hostname.", "")
 						}
 					}
-					catch
-					{
-						Write-Warning "$functionName - No .NET.Dns information from $Computer"
+					catch {
+						Write-Message -Level Warning -Message "No .NET.Dns information from $Computer"
 						continue
```
FriedrichWeinmann: Better use `Stop-Function` here. Also pass along exceptions

functions/Resolve-DbaNetworkName.ps1
```diff
@@ -93,175 +87,127 @@ Function Resolve-DbaNetworkName
...
+
+			try {
+				Write-Message -Level Verbose -Message "Resolving $($conn.DNSHostname) using .NET.Dns GetHostEntry"
 				$hostentry = ([System.Net.Dns]::GetHostEntry($conn.DNSHostname)).HostName
 			}
-			catch
-			{
-				Write-Warning "$functionName - .NET.Dns GetHostEntry failed for $($conn.DNSHostname)"
+			catch {
+				Write-Message -Level Warning -Message ".NET.Dns GetHostEntry failed for $($conn.DNSHostname)"
```
FriedrichWeinmann: Should we continue, if this fails?
If not: `Stop-Function`
Also please pass along exceptions

functions/Resolve-DbaNetworkName.ps1
wsmelton: So why does it break when you pass in a computer name that does not exist on the network?

wsmelton: `Resolve-DbaNetworkName -ComputerName DoesNotExist` throws and exception that it cannot convert the value.

functions/Resolve-DbaNetworkName.ps1
wsmelton: I don't follow? I understood we don't use the Silent switch in functions and then it picks that up if the user passes it in from calling the Resolve command itself.

FriedrichWeinmann: The main purpose of `-Silent` is to toggle on exceptions.
We don't want to scare users, so by default, we don't _show the user_ the exception.
The only functions that inherit the `$Silent` switch of the original function are:
`Write-Message`
`Stop-Function`
Which are the functions that manage verbosity and exception handling.

If you are calling another dbatools function, and want to determine whether it worked or not by using try/catch, you need an exception. Otherwise it won't throw one (since that's our default).

WIthout `-Silent` on Get-DbaCmObject, this function would always assume the call worked, since it never threw an exception.

FriedrichWeinmann: `Silent` isn't a great name for this mechanism, I simply have never found a less bad one (totally open for suggestions though)

potatoqualitee: Agreed about the Silent. What bout -Debug? or -AllowThrow?

FriedrichWeinmann: Debug is taken obviously.
AllowThrow doesn't explain why the console suddenly turned silent (Remember, silent does both: Remove verbosity _and_ throw exceptions).

functions/Resolve-DbaNetworkName.ps1
wsmelton: Reverted back to original type

functions/Resolve-DbaNetworkName.ps1
wsmelton: Removed

functions/Resolve-DbaNetworkName.ps1
wsmelton: Removed

functions/Resolve-DbaNetworkName.ps1
wsmelton: Removed

functions/Resolve-DbaNetworkName.ps1
wsmelton: What about "-Quiet"...or I like `-ShutTheFrontDoor`.

functions/Resolve-DbaNetworkName.ps1
wsmelton: Updated

functions/Resolve-DbaNetworkName.ps1
wsmelton: Done

functions/Resolve-DbaNetworkName.ps1
FriedrichWeinmann: `-Quiet` is established as "Don't return object, just $true or $false"

functions/Resolve-DbaNetworkName.ps1
wsmelton: updated and included error record

functions/Resolve-DbaNetworkName.ps1
wsmelton: Added

functions/Resolve-DbaNetworkName.ps1
wsmelton: `-Shush` or maybe `-Hush` which the term itself I think means different things in various languages but [English](http://www.dictionary.com/browse/shush).

potatoqualitee: butt hat's not what it's doing - it's allowing throws which is totally diff :/ otherwise, the user just gets warnings but ps cant tell the command didtn execute well

potatoqualitee:
got approval to merge from :brain: !

PR #1417, opened 2017-06-08
Get-DbaService

FriedrichWeinmann:
Adds a new command to dbatools:

 - Get-DbaService

Allows scanning for services using the connection management.

Example usage:
```PowerShell
Get-DbaService -Name LanmanServer
Get-ADComputer -Filter * | Get-DbaService -Name Browser
Get-DbaService -ComputerName "server1","server2","server3" -Name Lanman%
```


potatoqualitee:
hey fred - i'm happy to overlap with some of the SqlServer module functions, but I'm concerned about overlapping with Windows itself. Would this command be better as an internal function? I was actually wondering that about a few of the other Cm commands - we love them, but would end-users want them exposed? I'll ask in the #dbatools-style channel too for feedback

FriedrichWeinmann:
Heya Chrissy,

I totally don't mind putting it in internal instead. Wasn't too sure about this one and just went with intuition (`'Hey, this might be cool ...'`).
The other *-Cm* commands however are firmly in the public sector:
`Get-DbaCmObject` (and the upcoming `Invoke-DbaCmCommand`) are plain too useful, so I'd keep them exposed. The rest is all about configuring the way dbatools connects to computers, so I'd keep them public.

For the meta commands - such as this one, as well as the pending `Stop-DbaService`, `Start-DbaService` and `Restart-DbaService` - I can totally live with making them internal.

FriedrichWeinmann:
Another point though:
This does only partially overlap with the windows cmdlets: Get-/Start-/Stop-/Restart-Service use the service control manager's own remoting options. This means, that service management using those depends on the firewall and service allowing remote management over this old interface.

This set of commands however achieves this over Cim/Wmi/PowerShell Remoting. So while it _does_ provide the same functionality, it takes a radically different path to do so. It can be very legitimately argued that this command is out-of-scope for dbatools to _publish_ . On the other hand, how often do dbas have no control over the firewall settings they are stuck with?

wsmelton:
I could see these of use for say discovery purposes. If say I get some support ticket for a client (or user) on a server, one that I don't recognize, then I'd want to ensure SQL Server is running on it. We could expand these as internal functions or public ones and say build one that is wrapper to use various methods to say SQL Server is on this machine or it isn't.

Just thinking out loud.

FriedrichWeinmann:
Well ... we have Get-DbaSqlServer for that I think.

wsmelton:
There is one for `Get-DbaSqlService` but that is just [basic services](https://github.com/sqlcollaborative/dbatools/blob/master/functions/Get-DbaSqlService.ps1).

FriedrichWeinmann:
Well ... I suppose this is as ready as it's going to be, merging.

PR #1428, opened 2017-06-11
cert functionality

potatoqualitee:


functions/Add-DbaComputerCertificate.ps1
```diff
@@ -78,36 +75,39 @@ Adds the local C:\temp\cer.cer to the local computer's LocalMachine\My (Personal
...
 		}
 		
-		if (!(Test-Path -Path $path)) {
-			Write-Message -Level Warning -Message "$Path does not exist"
-			return
+		if ($Path) {
+			if (!(Test-Path -Path $Path)) {
+				Write-Message -Level Warning -Message "$Path does not exist"
+				return
+			}
```
FriedrichWeinmann: Input Validations of non-piped values should be part of `begin`
Use Stop-Function to interrupt due to bad parameterization

functions/Add-DbaComputerCertificate.ps1
```diff
@@ -78,36 +75,39 @@ Adds the local C:\temp\cer.cer to the local computer's LocalMachine\My (Personal
...
+				$Certificate = New-Object System.Security.Cryptography.X509Certificates.X509Certificate2
+				$Certificate.Import($bytes, $password, [System.Security.Cryptography.X509Certificates.X509KeyStorageFlags]::DefaultKeySet)
 			}
-			
-			if ($Certificate) {
-				$tempdir = ([System.IO.Path]::GetTempPath()).TrimEnd("\")
-				$tempfile = "$tempdir\cert.cer"
-				$certdata = $certificate.Export([System.Security.Cryptography.X509Certificates.X509ContentType]::PFX, $password)
+			catch {
+				Stop-Function -Message "Can't import certificate." -ErrorRecord $_ -Continue
```
FriedrichWeinmann: `-Continue` parameter should only be used within a loop.
Given general advice probably a non-issue though in this instance

functions/Add-DbaComputerCertificate.ps1
```diff
@@ -78,36 +75,39 @@ Adds the local C:\temp\cer.cer to the local computer's LocalMachine\My (Personal
...
+				Write-Message -Level Warning -Message "$Path does not exist"
+				return
+			}
 		}
 		
-		foreach ($computer in $computername) {
-			if (![dbavalidate]::IsLocalhost($computer) -and !$Password) {
-				Write-Message -Level Output -Message "You have specified a remote computer. A password is required for private key encryption/decryption for import."
-				$Password = Read-Host -AsSecureString -Prompt "Password"
+		if (![dbavalidate]::IsLocalhost($computer) -and !$Password) {
```
FriedrichWeinmann: Will fail painfully.
The loop through computers only starts below.
Also note:
`$computer.IsLocalhost` can also be used on the parameter-class. I included that because it was most likely to be needed there. It calls the same validation method to determine whether it is localhost, so the results should always be identical.

potatoqualitee: Ouu, i like $computer.IsLocalhost - ohh, i need to check that. I moved it around.

functions/Add-DbaComputerCertificate.ps1
potatoqualitee: Admittedly, I shortcutted it... fffffffineeee

functions/Add-DbaComputerCertificate.ps1
potatoqualitee: whoops, i moved stuff around. thanks for the catch!

functions/Add-DbaComputerCertificate.ps1
potatoqualitee: this is a loop tho cuz foreach computer in computername?

FriedrichWeinmann: True, but according to my browser, you are only starting that loop below this line.

PR #1434, opened 2017-06-13
Moved TEPP cache building to runspace

FriedrichWeinmann:
Overhauled the TEPP system to run _by default_ in the background and not as part of the direct connection process.
This prevents delays in execution, especially on bad connections, due to requesting information not pertinent to the command the user was meaning to run.
However, it is possible to opt into the previous model of TEPP updates (for example in case number of connections is an issue)

It also adds some configuration options to TEPP, in order to provide better flexibility:
`'TabExpansion.UpdateInterval'` - How frequent the TEPP cache is updated (default: 3 minutes)
`'TabExpansion.UpdateTimeout'` - How long after the last connection the cache is being updated (default: 30 minutes)
`'TabExpansion.Disable'` - Disables Tab Expansion (not fully implemented yet)
`'TabExpansion.Disable.Asynchronous'` - Disables background updates of the TEPP cache
`'TabExpansion.Disable.Synchronous'` - Disables foreground updates of the TEPP cache (enabled by default)

### DIFF
--- a/bin/library.ps1
+++ b/bin/library.ps1
@@ -2212,9 +2212,12 @@ namespace Sqlcollective.Dbatools
             /// </summary>
             public static void Kill()
             {
-                LogWriter.Runspace.Close();
-                LogWriter.Dispose();
-                LogWriter = null;
+                if (LogWriter != null)
+                {
+                    LogWriter.Runspace.Close();
+                    LogWriter.Dispose();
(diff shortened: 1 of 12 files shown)


potatoqualitee:
This is straight up wizardry! I'll merge so we can all test.

PR #1437, opened 2017-06-13
Asynchronous TEPP: Fixed execution on import

FriedrichWeinmann:
Forgot to submit the part where the background update script was loaded on import.
Without that, no TEPP updates ...


potatoqualitee:
merging this too!

PR #1446, opened 2017-06-14
Rename command - Get-DbaDatabaseFreespace

wsmelton:
Rename Get-DbaDatabaseFreespace to `Get-DbaDatabaseSpace`.

Also updated the index for `Find-DbaCommand`.


potatoqualitee:
hey @wsmelton - thank you! just fyi, each publish gets a good ol refresh. we've updated so many tho, that's a good idea for me to do in this branch

PR #1451, opened 2017-06-16
Get-DbaSqlBuildReference Upgrades

FriedrichWeinmann:
Fixed file handling in writable location (used to never update it after first copy. Issue when updating the module)
Redesigned object/property handling (Use Version type rather than suffer string manipulation)
Fixed Scope Violations (evil)
Overhauled messaging and flow control (Just adapting it to the current state, since a lot has changed)
Implemented DbaInstanceParameter (Another addition that came later)
Implemented integrated connections to instances. (Trivial with the instance parameter)

### DIFF
--- a/functions/Get-DbaSqlBuildReference.ps1
+++ b/functions/Get-DbaSqlBuildReference.ps1
@@ -122,7 +122,7 @@ Function Get-DbaSqlBuildReference {
 			
 			$LastUpdated = Get-Date -Date $result.LastUpdated
 			if ($LastUpdated -lt (Get-Date).AddDays(-45)) {
-				Write-Message -Level Warni -Silent $Silent -Message "Index is stale, last update on: $(Get-Date -Date $LastUpdated -Format s)"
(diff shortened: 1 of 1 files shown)


functions/Get-DbaSqlBuildReference.ps1
```diff
@@ -1,197 +1,249 @@
...
 			}
+			
+			# Else if the module version of the file no longer exists, but the writable version exists, return the writable version
+			else {
+				$result = Get-Content $writable_idxfile -Raw | ConvertFrom-Json
+			}
+			
+			$LastUpdated = Get-Date -Date $result.LastUpdated
+			if ($LastUpdated -lt (Get-Date).AddDays(-45)) {
+				Write-Message -Level Warni -Silent $Silent -Message "Index is stale, last update on: $(Get-Date -Date $LastUpdated -Format s)"
```
wsmelton: typo "Warni"

FriedrichWeinmann: good catch, fixed

niphlod:
failing tests should not be a problem. When this is merged, I have one PR in line to let the index be fetched without having to update the release.

potatoqualitee:
thanks all! 🙇

PR #1452, opened 2017-06-16
Write-DbaDataTable now works with SQL Azure databases (PAAS)

Stuart-Moore:
Fixes # 

 - Add SQL PAAS support

There's a slightly hacky workaround at L264 to cope with smo not populating the tables object properly on first request. It works reliably, just not the prettiest

L307 - I could add the databasename to both connection strings. Shouldn't break anything, but didn't know if there was a design decision to not do that from the start?

How to test this code: 
- [ ] See example in function
- [ ] 

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally


Stuart-Moore:
Now also includes a fix for when users pre quote their objects. Currently passing in 

$table = '[MyDb].[dbo].[MyTable]'
Write-DbaDataTable -sqlserver $server -SqlCredential $mycreds -InputObject $csv -Table $table

will stop, as we double protect the objects (so it's looking for [[MyDb]]')

And the object existence checks fail as the SMO objects won't match [MyDb] as they return MyDb

potatoqualitee:
Hell yes, thank you @ConstantineK !!! @Stuart-Moore - I'mm test this when i get a moment but surely before friday

PR #1455, opened 2017-06-19
Output added for Copy-DbaDatabaseAssembly

wsmelton:
 - formatted help
 - Add message system
 - update camelCase
 - Add output object


potatoqualitee:
😍

PR #1457, opened 2017-06-19
Get-DbaSqlBuildReference, allow online updates

niphlod:
 - added a new Update switch to fetch the index from online assets

### DIFF
--- a/functions/Get-DbaSqlBuildReference.ps1
+++ b/functions/Get-DbaSqlBuildReference.ps1
@@ -120,7 +120,8 @@ Function Get-DbaSqlBuildReference {
 				if ($module_time -gt $data_time) {
 					Copy-Item -Path $orig_idxfile -Destination $writable_idxfile -Force -ErrorAction Stop
 					$result = $module_content
-				} else {
+				}
+				else {
 					$result = $data_content
 					$offline_time = $data_time
 				}
@@ -131,7 +132,7 @@ Function Get-DbaSqlBuildReference {
 						$webdata_content = $WebContent.Content | ConvertFrom-Json
 						$webdata_time = Get-Date $webdata_content.LastUpdated
 						if ($webdata_time -gt $offline_time) {
-							Write-Message -Level Output -Message "Index updated correctly, last update on: $(Get-Date -Date $webdata_time -f s), was $(Get-Date -Date $offline_time -f s)"
+							Write-Message -Level Output -Message "Index updated correctly, last update on: $(Get-Date -Date $webdata_time -Format s), was $(Get-Date -Date $offline_time -Format s)"
 							$WebContent.Content | Out-File $writable_idxfile -Encoding utf8 -ErrorAction Stop
 							$result = Get-Content $writable_idxfile -Raw | ConvertFrom-Json
 						}
--- a/functions/Get-DbaSqlBuildReference.ps1
+++ b/functions/Get-DbaSqlBuildReference.ps1
@@ -25,6 +25,11 @@ Function Get-DbaSqlBuildReference {
 		Returns information about a build identified by  "12.00.4502" (which is SQL 2014 with SP1 and CU11)
 	
 	.EXAMPLE
+		Get-DbaSqlBuildReference -Build "12.00.4502" -Update
+		
+		Same as previous, but checks online for a more up to date index
+	
+	.EXAMPLE
 		Get-DbaSqlBuildReference -Build "12.0.4502","10.50.4260"
 		
 		Returns information builds identified by these versions strings
@@ -64,6 +69,9 @@ Function Get-DbaSqlBuildReference {
 		$SqlCredential,
 		
 		[switch]
+		$Update,
+		
+		[switch]
 		$Silent
 	)
 	
@@ -74,6 +82,9 @@ Function Get-DbaSqlBuildReference {
 			Param (
 				[string]
 				$Moduledirectory,
+				
+				[bool]
+				$Update,
 				
 				[bool]
 				$Silent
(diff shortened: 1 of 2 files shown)


functions/Get-DbaSqlBuildReference.ps1
```diff
@@ -105,13 +116,26 @@ Function Get-DbaSqlBuildReference {
...
 				$data_time = Get-Date $data_content.LastUpdated
 				
+				$offline_time = $module_time
 				if ($module_time -gt $data_time) {
 					Copy-Item -Path $orig_idxfile -Destination $writable_idxfile -Force -ErrorAction Stop
 					$result = $module_content
-				}
-				
-				else {
+				} else {
```
FriedrichWeinmann: Was there another deterioration in our style guideline?
If not, please refrain from placing else on the same line as the closing brace :shudder:

functions/Get-DbaSqlBuildReference.ps1
```diff
@@ -105,13 +116,26 @@ Function Get-DbaSqlBuildReference {
...
+					$offline_time = $data_time
+				}
+				# If Update is passed, try to fetch from online resource and store into the writeable
+				if ($Update) {
+					$WebContent = Get-DbaSqlBuildReferenceIndexOnline -Silent $Silent
+					if ($null -ne $WebContent) {
+						$webdata_content = $WebContent.Content | ConvertFrom-Json
+						$webdata_time = Get-Date $webdata_content.LastUpdated
+						if ($webdata_time -gt $offline_time) {
+							Write-Message -Level Output -Message "Index updated correctly, last update on: $(Get-Date -Date $webdata_time -f s), was $(Get-Date -Date $offline_time -f s)"
```
FriedrichWeinmann: Don't use aliases - even for parameters - in a script

functions/Get-DbaSqlBuildReference.ps1
```diff
@@ -122,12 +146,35 @@ Function Get-DbaSqlBuildReference {
...
 			$result.Data | Select-Object @{ Name = "VersionObject"; Expression = { [version]$_.Version } }, *
 		}
 		
+		function Get-DbaSqlBuildReferenceIndexOnline {
+			[CmdletBinding()]
+			Param (
+				[bool]
+				$Silent
+			)
+			$url = 'https://sqlcollaborative.github.io/assets/dbatools-buildref-index.json'
```
FriedrichWeinmann: Should be a configuration entry. Information should not be hardcoded. For example, large organizations could provide an internal path and thus allow machines without online access to update.
I'd recommend creating a new configuration file "assets", then place the config there and call it with something like this:
`Get-DbaConfigValue -Name 'assets.sqlbuildreference'`

functions/Get-DbaSqlBuildReference.ps1
niphlod: huh ... aliases ??

FriedrichWeinmann: `-f` instead of `-Format`

functions/Get-DbaSqlBuildReference.ps1
niphlod: good thinking. But I don't want to disobey 'cause "disobedience shall be answered by the wrath of Fred." :-P I'm missing how to code the "handler" part for it.

functions/Get-DbaSqlBuildReference.ps1
niphlod: will adhere but I lost where this is formalized ^_^ pretty recently edited files never had this explicit guideline

functions/Get-DbaSqlBuildReference.ps1
niphlod: dang! Fixed.

functions/Get-DbaSqlBuildReference.ps1
FriedrichWeinmann: Handler is optional. It's mostly input validation. You could try to convert the input into an uri, in order to determine whether it's legit, but you can also just skip validation if you want to.

niphlod:
moved $url to configuration, too. Thanks for the review

internal/configurations/assets.ps1
```diff
@@ -0,0 +1,6 @@
+<#
+This is designed for all online assets you need configurations for.
+#>
+
+# The default path where dbatools stores persistent data
+Set-DbaConfig -Name 'assets.sqlbuildreference' -Value 'https://sqlcollaborative.github.io/assets/dbatools-buildref-index.json' -Default -DisableHandler -Description "The url where dbatools fetches the up to date builreference index"
```
FriedrichWeinmann: Typo: `Builreference`
You may also want to include the product the build reference is for in the description (just to be 100% blunt and direct)

niphlod: ok

functions/Get-DbaSqlBuildReference.ps1
```diff
@@ -25,6 +25,11 @@ Function Get-DbaSqlBuildReference {
 		Returns information about a build identified by  "12.00.4502" (which is SQL 2014 with SP1 and CU11)
 	
 	.EXAMPLE
+		Get-DbaSqlBuildReference -Build "12.00.4502" -Update
```
PowerDBAKlaas: Can you add 'update' in the .PARAMETER section, please?
Maybe some extensive explanation what happens when using it; does it overwrite a file?

niphlod: done

PR #1465, opened 2017-06-21
Added parameter for enabled

jpomfret:
Fixes # 

 -  Add parameter for Enabled

How to test this code: 
- [ ]  Get-DbaAgentJob -SqlInstance localhost -Enabled True
Returns all enabled jobs from localhost
- [ ]  Get-DbaAgentJob -SqlInstance localhost -Enabled False
Returns all disabled jobs from localhost

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [X]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally

### DIFF
--- a/functions/Get-DbaAgentJob.ps1
+++ b/functions/Get-DbaAgentJob.ps1
@@ -4,11 +4,12 @@ FUNCTION Get-DbaAgentJob {
 			Gets SQL Agent Job information for each instance(s) of SQL Server.
 
 		.DESCRIPTION
-			The Get-DbaAgentJob returns connected SMO object for SQL Agent Job information for each instance(s) of SQL Server.
+			The Get-DbaAgentJob returns connected SMO object for SQL Agent Job information for each instance(s) of SQL Server. The default behavior
+			is to return enabled jobs only.
 
 		.PARAMETER SqlInstance
-			SQL Server name or SMO object representing the SQL Server to connect to. This can be a collection and recieve pipeline input to allow the function
-			to be executed against multiple SQL Server instances.
+			SQL Server name or SMO object representing the SQL Server to connect to. This can be a collection and recieve pipeline input 
+			to allow the function to be executed against multiple SQL Server instances.
 
 		.PARAMETER SqlCredential
 			SqlCredential object to connect as. If not specified, current Windows login will be used.
@@ -20,7 +21,7 @@ FUNCTION Get-DbaAgentJob {
 			The job(s) to exclude - this list is auto populated from the server.
 
 		.PARAMETER IncludeDisabled
-			Default is to only return enabled jobs, switch will also return disabled jobs.
+			Switch will include the disabled jobs in output.
 
 		.PARAMETER Silent
 			Use this switch to disable any kind of verbose messages
@@ -39,17 +40,27 @@ FUNCTION Get-DbaAgentJob {
 		.EXAMPLE
 			Get-DbaAgentJob -SqlInstance localhost
 
-			Returns all SQL Agent Job on the local default SQL Server instance
+			Returns all enabled SQL Agent Jobs on the local default SQL Server instance
 
 		.EXAMPLE
 			Get-DbaAgentJob -SqlInstance localhost, sql2016
 
-			Returns all SQl Agent Job for the local and sql2016 SQL Server instances
+			Returns all enabled SQl Agent Jobs for the local and sql2016 SQL Server instances
 
+		.EXAMPLE
+			Get-DbaAgentJob -SqlInstance localhost -Job BackupData, BackupDiff
+
+			Returns all enabled SQL Agent Jobs named BackupData and BackupDiff from the local SQL Server instance.
+
+		.EXAMPLE
+			Get-DbaAgentJob -SqlInstance localhost -ExcludeJob BackupDiff
+
+			Returns all enabled SQl Agent Jobs for the local SQL Server instances, except the BackupDiff Job.
+			
 		.EXAMPLE
 			Get-DbaAgentJob -SqlInstance localhost -IncludeDisabled
 
-			Includes disabled SQl Agent Jobs in the returned set for the local SQL Server instances
+			Returns all SQl Agent Jobs for the local SQL Server instances, including disabled jobs.
 	#>
 	[CmdletBinding()]
 	param (
@@ -90,7 +101,7 @@ FUNCTION Get-DbaAgentJob {
 			foreach ($agentJob in $jobs) {
 				Add-Member -InputObject $agentJob -MemberType NoteProperty -Name ComputerName -value $agentJob.Parent.Parent.NetName
 				Add-Member -InputObject $agentJob -MemberType NoteProperty -Name InstanceName -value $agentJob.Parent.Parent.ServiceName
-				Add-Member -InputObject $agentJob -MemberType NoteProperty -Name SqlInstance -value $agentJob.Parent.Parent.DomainInstanceName
+				Add-Member -InputObject $agentJob -MemberType NoteProperty -Name SqlInstance -value $agentJob.Parent.Parent.DomainInstanceName			
 				
 				Select-DefaultView -InputObject $agentJob -Property ComputerName, InstanceName, SqlInstance, Name, Category, OwnerLoginName, 'IsEnabled as Enabled', LastRunDate, DateCreated, HasSchedule, OperatorToEmail
 			}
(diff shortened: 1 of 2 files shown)


functions/Get-DbaAgentJob.ps1
```diff
@@ -52,6 +60,8 @@ FUNCTION Get-DbaAgentJob {
 		$SqlCredential,
 		[object[]]$Job,
 		[object[]]$ExcludeJob,
+		[ValidateSet('True', 'False')]
+		[string]$Enabled,
```
wsmelton: This should be a bool value, or maybe just a switch. We should not have them type in the word but rather use the variables of `$true` or `$false`. 

However, another point of view is a parameter named `IncludeDisabled` and then by default, the command only returns enabled jobs. If the user wants to include disabled jobs, then they provide the switch. 

Thoughts from @potatoqualitee  ?

functions/Get-DbaAgentJob.ps1
```diff
@@ -67,18 +77,26 @@ FUNCTION Get-DbaAgentJob {
 			}
 
 			$jobs = $server.JobServer.Jobs
+			if ($Enabled) {
+				if($Enabled -eq 'True') { 
+					$jobs = $Jobs | Where-Object IsEnabled -eq $true
+				} else {
+					$jobs = $Jobs | Where-Object IsEnabled -eq $false
+				}
+			}
```
wsmelton: This block of logic needs to be after the `$ExcludeJob` filter, originally after line 75. 

I guess technically it does not matter but just a matter of flow, we filter the `$jobs` down first based on the `-Job` and `-ExcludeJob` objects passed in, then filter on the enabled or disabled.

functions/Get-DbaAgentJob.ps1
wsmelton: I know have the time I am pulling jobs from a server I am only concerned with the enabled ones initially.

jpomfret: I like that idea better, didn't think of that. I'll get an update together, thanks for the review!

functions/Get-DbaAgentJob.ps1
```diff
@@ -19,8 +19,8 @@ FUNCTION Get-DbaAgentJob {
 		.PARAMETER ExcludeJob
 			The job(s) to exclude - this list is auto populated from the server.
 
-		.PARAMETER Enabled
-			True will return only enabled jobs, false will return disabled jobs.
+		.PARAMETER IncludeDisabled
+			Default is to only return enabled jobs, switch will also return disabled jobs.
```
wsmelton: I would say note the default output is only enabled jobs in the `.DESCRIPTION`. You can then keep the definition of this parameter to the point, something like: `Switch will include the disabled jobs in output.`

functions/Get-DbaAgentJob.ps1
```diff
@@ -47,9 +47,9 @@ FUNCTION Get-DbaAgentJob {
 			Returns all SQl Agent Job for the local and sql2016 SQL Server instances
 
 		.EXAMPLE
-			Get-DbaAgentJob -SqlInstance localhost -Enabled True
+			Get-DbaAgentJob -SqlInstance localhost -IncludeDisabled
```
wsmelton: This needs to be separate `.EXAMPLE`. Look at some of the other Get commands will show you the common examples we include. It is best to try and do an example for each parameter. We obviously forgot to show examples for the Job and ExcludeJob parameter so if you want to add those as well would be good.

functions/Get-DbaAgentJob.ps1
```diff
@@ -77,20 +76,16 @@ FUNCTION Get-DbaAgentJob {
...
-			}
 			
 			if ($Job) {
 				$jobs = $jobs | Where-Object Name -In $Job
 			}
 			if ($ExcludeJob) {
 				$jobs = $jobs | Where-Object Name -NotIn $ExcludeJob
 			}
+			if (!$IncludeDisabled -and !$job) {
+					$jobs = $Jobs | Where-Object IsEnabled -eq $true
```
wsmelton: 💯

functions/Get-DbaAgentJob.ps1
wsmelton: Disregard this one, reading it wrong on the diff view.

However, if you can include the examples for the `-Job` and `-ExcludeJob` would make it complete examples.

jpomfret:
Looking at issue #1472 and seeing if I can add some schedule/last run info

functions/Get-DbaAgentJob.ps1
jpomfret: Consider them added :)

PowerDBAKlaas:
@jpomfret shedule info does not belong in the AgentJob objects.

functions/Get-DbaAgentJob.ps1
```diff
@@ -4,11 +4,12 @@ FUNCTION Get-DbaAgentJob {
 			Gets SQL Agent Job information for each instance(s) of SQL Server.
 
 		.DESCRIPTION
-			The Get-DbaAgentJob returns connected SMO object for SQL Agent Job information for each instance(s) of SQL Server.
+			The Get-DbaAgentJob returns connected SMO object for SQL Agent Job information for each instance(s) of SQL Server. The default behavior
+			is to return enabled jobs only.
```
PowerDBAKlaas: A function should do what is says, in this case, get the Agent Jobs, and not implement an arbitrary filter by default just because it is convenient for some people.

functions/Get-DbaAgentJob.ps1
```diff
@@ -67,18 +87,22 @@ FUNCTION Get-DbaAgentJob {
...
 			$jobs = $server.JobServer.Jobs
+			
 			if ($Job) {
 				$jobs = $jobs | Where-Object Name -In $Job
 			}
 			if ($ExcludeJob) {
 				$jobs = $jobs | Where-Object Name -NotIn $ExcludeJob
 			}
-
+			if (!$IncludeDisabled -and !$job) {
```
PowerDBAKlaas: The -Enabled switch as implemented before was a better solution.

functions/Get-DbaAgentJob.ps1
wsmelton: SO you are saying we should allow a parameter to accept the string version of "true" and "false"?

functions/Get-DbaAgentJob.ps1
wsmelton: Looking at other commands then to follow the same pattern this should actually be changed to "-NoDisabledJobs" and then by default we return all of them.

- `Get-DbaDatabase` has NoUserDb and NoSystemDb
- `Get-DbaProcess` has NoSystemSpids
- `Get-DbaTable` has IncludeSystemDbs

wsmelton:
@niphlod I would party agree with that, however, in this manner, we are filtering on a property which is a pattern most of the built-in functions for PowerShell also provide...at least for the most common. And with SQL Agent jobs looking at disabled/enabled is a common filter almost everyone uses.

functions/Get-DbaAgentJob.ps1
niphlod: as far as we figure out a logic and every cmdlet tries to follow it, it's good (i.e. -nooranges or -apples are both logic)

jpomfret:
So is the best solution here not to add an -enabled variable or a -includedisable switch but to just have users pipe into where-object?

wsmelton:
The general discussion is what to call the variable, which would then dictate how the logic works.

General synopsis I see is whether to call it `-NoDisabledJobs`, which is what other commands are doing in dbatools. However, others also just do `-Disabled` but in the logic of the command, this would be to only return the disabled jobs, not filter them from the full list.

I favor more using `NoDisabledJobs` as that makes readable sense what it is for, and would mean not passing that in all jobs are returned. We have some commands that just use `-Disabled` but it is only one or two, and the logic behind that parameter is only returning the disabled objects.

The boss will likely need to decide which way she wants to go @potatoqualitee  🔨

PowerDBAKlaas:
@niphlod indeed we probably need to reconsider some other functions and make them more consistent
@wsmelton `-NoDisabledJobs` may certainly be the most clear, or did we have the `-Exclude...` for this?

potatoqualitee:
Hoo, boy! Where we at with this, yall?

wsmelton:
@jpomfret I believe we are in favor of using -NoDisabledJob as the switch name.

jpomfret:
Alright. I'll get it submitted probably on Thursday (on holiday til wed night).

potatoqualitee:
thanks yall <3

PR #1466, opened 2017-06-21
Export-DbaLogin - Missing -Message parameter

ClaudioESSilva:
One of the Write-Message does not have the -Message named parameter.
Which leads to:

```
Write-Message : A positional parameter cannot be found that accepts argument 'NT SERVICE\MSSQLSERVER is skipped because it is a local machine name.'.
At line:139 char:6
+ ...             Write-Message -Level Warning "$userName is skipped becaus ...


potatoqualitee:
Thanks C-dawg

PR #1468, opened 2017-06-22
Library - Changed entire code maintenance and build process

FriedrichWeinmann:
This is a complete overhaul of the entire library structure.
It is now maintained in a Visual Studio project structure with individual files per class/enumeration/etc.

The main purpose is to make it easier to contribute to the library. It'll also slightly improve module import speed (except for the first time).

Important change!
 - In development branch, the library is compiled to the module directory. If the directory is in a UAC protected folder, it may need elevation to load the module for the first time.

Also renamed the typo in the namespace : `SqlCollective.Dbatools` has become `Sqlcollaborative.Dbatools`

Finally added two control variables that are respected by the library:
 - `$dbatools_strictsecuritymode` : Setting this to $true will cause dbatools to always load the library directly from the module directory. This is more secure, but less convenient when it comes to updating the module, as all consoles using it must be closed.
 - `$dbatools_alwaysbuildlibrary` : Setting this to $true will cause the module to always build the library from source, rather than reuse the binaries. Mostly for developers working on the library.
Both of these variables can be added to your profile if needed, in order to adjust the way the module handles the library


bin/projects/dbatools/dbatools/Configuration/Config.cs
```diff
@@ -0,0 +1,67 @@
+﻿using System;
```
zippy1981: Can we make one namespace declaration per file and move all the usings up top:

```
using System;
using System.Collections;

namespace Sqlcollaborative.Dbatools.Configuration
{
```

bin/projects/dbatools/dbatools/Utility/DbaDate.cs
```diff
@@ -0,0 +1,237 @@
...
+            /// <summary>
+            /// Constructs a generic timestamp object wrapper from an input timestamp object.
+            /// </summary>
+            /// <param name="Timestamp">The timestamp to wrap</param>
+            public DbaDate(DateTime Timestamp)
+            {
+                _timestamp = Timestamp;
+            }
+
+            /// <summary>
```
zippy1981: I think you should kill the empty comment blocks. Add them later when you have actually comments. Is there a reason you are creating empty ones?

potatoqualitee:
Oh, good. You're looking too, thanks @zippy1981! Please decide and merge at your earliest convenience, @FriedrichWeinmann - that way, I can include this framework change in my tests this weekend.

bin/projects/dbatools/dbatools/Configuration/Config.cs
FriedrichWeinmann: Absolutely. That's a relic from the auto-conversion Visual Studio was using in conjunction with the limitations of the single-file library.
Was gonna clean that up once the proof of concept has been worked out. If you have automated tools for that, go ahead.

bin/projects/dbatools/dbatools/Utility/DbaDate.cs
FriedrichWeinmann: Yes, there is a reason.
Those classes were autogenerated from the original DateTime struct. I added the comments in order to shut Visual Studio up about them (I was willing to forego proper comments on a DateTime wrapper), so I'd see where I'd actual failed to provide proper help (because I consider those comments to be absolutely mandatory for original dbatools content).
Again a legacy of the old single-file project the old library mandated. I suppose they could be dropped now, but I don't see the reason since they'll flood the build log without added benefit that I can see. I'm fine either way though, so long as it doesn't mess with development.

PR #1470, opened 2017-06-22
test fixes i hope

potatoqualitee:


niphlod:
haha, that's why I recommended a proper before_start segment on appveyor.yml. 
Tests shouldn't wait. the ENV should, before starting the test process.
The underlying reasoning is that - one day, hopefully soon - everyone can run dbatools tests in their own lab.

potatoqualitee:
You mean this that's already there? ;)

````
before_test:
   # "Get Pester manually"
  - ps: Install-Module -Name Pester -Repository PSGallery -Force | Out-Null
  # "Setting up the local SQL Server environments"
  - ps: .\Tests\appveyor.sqlserver.ps1
````

Do you have a proposal to fix it? I find that it's just lying. Says it's really started but something's up.

**Comment by niphlod** [comment]
that relies on the service being up but it's not really always the case where sql server is really ready in alternating/constrained VM performances.
I found out that a "more reliable" way to see if it's up or not is more on the lines of 
```
while($LASTEXITCODE -ne 0)
        {
          & sqlcmd -S "(local)\SQL2016" -b -U "sa" -P "Password12!" -Q "select 1" -d "master"
          sleep 10; $val++; Write-Host Waiting ... $val; if($val -ge 10) {break}
        }
```

potatoqualitee:
Sweet! I just added this

````
do {
	$null = (& sqlcmd -S localhost\sql2016 -b -Q "select 1" -d master)
}
while ($lastexitcode -ne 0 -and $s++ -lt 10)
````

niphlod:
gosh I remember someone smacking my forehead with do - while 🗡

niphlod:
btw: without a sleep it's going to be a really tight loop.

potatoqualitee:
it worked out cuz the timeout was like 15 seconds, but i added more sleeps for stability 👍

PR #1475, opened 2017-06-23
Out-DbaDatatable - Added support for dbasize & dba timespan types

FriedrichWeinmann:
- DbaTimespan is now understood as a timespan
 - DbaTimespanPretty is now understood as a timespan
 - DbaSize is now properly understood and can be converted as Int64 (default), Int32 or String
 - New Parameter: SizeType - controls how DbaSize is interpreted
 - Added 1.0 Progress Tracking tags


potatoqualitee:
Thanks, @FriedrichWeinmann !

PR #1476, opened 2017-06-23
Get-DbaDatabaseFile, adjustment for sparse files

### LINKED ISSUE #1246: Get-DbaDatabaseFile, wrong properties returned

niphlod:
Fixes #1246 

 - take file paths from master_files instead of database_files
 - take the file size from dm_io_virtual_file_stats
 - take file properties from master_files instead of database_files

This solves wrong properties returned for sparse files, most notable on database snapshots

How to test this code: 
- [ ] run the previous version of this cmdlet and compare with the newer output
- [ ] confirm that for snapshots the returned file is the sparse one with the correct properties returned
- [ ] confirm that for anything else than sparse files the output from the previous version and this is the same

### DIFF
--- a/functions/Get-DbaDatabaseFile.ps1
+++ b/functions/Get-DbaDatabaseFile.ps1
@@ -78,18 +78,19 @@ function Get-DbaDatabaseFile {
 				df.Type,
 				df.type_desc as TypeDescription,
 				df.name as LogicalName,
-				df.physical_name as PhysicalName,
+				mf.physical_name as PhysicalName,
 				df.state_desc as State,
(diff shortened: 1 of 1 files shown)


potatoqualitee:
thanks, niph!

PR #1482, opened 2017-06-24
[New Command] Get-DbaSqlManagementObject

dbaduck:
Fixes # 

 - 
 - 
 - 

How to test this code: 
- [X] Run with nothing on the parameters and you should get all the versions on the computer 
- [X] Run with a -VersionNumber of a version between 9 and 14 and you will either get back a result with that version or you will get nothing 
- [X] Run with a -VersionNumber of 8 and you will not be able to because there is a ValidateSet for SMO Versions of 9-14. 

- [X]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000 - This is not related to SQL Server versions only SMO

- [X]  SQL Server vNext
- [X]  Windows 10
- [ ]  Azure Database - Not relevant to SQL Server only SMO

Tests for tester:
- [ ] Works remotely
- [ ] Works locally


functions/Get-DbaSqlManagementObject.ps1
```diff
@@ -0,0 +1,88 @@
...
+	Get-DbaSqlManagementObject -VersionNumber 13
+	Returns just the version specified. If the version does not exist then it will return nothing.
+	
+
+#>
+	[CmdletBinding()]
+	param (
+		[Parameter(Mandatory = $false, Position=0)]
+		[ValidateSet(9,10,11,12,13,14)]
+		[int]$VersionNumber = 0,
```
wsmelton: I don't know that I would do a validation set for this because it will mean we have to maintain this command each time a new version of SQL Server comes out.

I would simply validate it is 9 or greater in the begin block and then move on. 

@potatoqualitee or @PowerDBAKlaas thoughts?

functions/Get-DbaSqlManagementObject.ps1
```diff
@@ -0,0 +1,88 @@
+﻿function Get-DbaSqlManagementObject
+{
+<#
+	.SYNOPSIS
+	Gets SQL Mangaement Object versions installed on the machine.
+
+	.DESCRIPTION
+	 The Get-DbaSqlManagementObject returns an object with the Version and the Add-Type Load Template for each version on the server.
```
wsmelton: Just cosmetics, but would add a line break (make it 2 line sentence) in this to make it more readable in PowerShell console and such.

functions/Get-DbaSqlManagementObject.ps1
```diff
@@ -0,0 +1,88 @@
...
+	This is the specific version number you are looking for. The function will look 
+	for that version only.
+	
+	.PARAMETER Silent
+	Use this switch to disable any kind of verbose messages
+	
+	.NOTES
+	Original Author: Ben Miller (@DBAduck - http://dbaduck.com)
+
+	Tags: SMO
```
wsmelton: Just to match format of most other commands can you change this to be just:

```powershell
Tags: SMO
Original Author....
```

No need for the line spacing and such.

functions/Get-DbaSqlManagementObject.ps1
```diff
@@ -0,0 +1,88 @@
...
+	.EXAMPLE
+	Get-DbaSqlManagementObject
+	Returns all versions of SMO on the computer
+	
+	.EXAMPLE
+	Get-DbaSqlManagementObject -VersionNumber 13
+	Returns just the version specified. If the version does not exist then it will return nothing.
+	
+
+#>
```
wsmelton: It varies between commands but we (at least me :) ) are trying to format what we can in the same way.

So for the help can you format it (tabs, line breaks, etc.) the same as [this command](https://github.com/sqlcollaborative/dbatools/blob/development/functions/Copy-DbaAgentAlert.ps1).

functions/Get-DbaSqlManagementObject.ps1
```diff
@@ -0,0 +1,88 @@
...
+		[Parameter(Mandatory = $false, Position=0)]
+		[ValidateSet(9,10,11,12,13,14)]
+		[int]$VersionNumber = 0,
+		[switch]$Silent
+	)
+	
+	process {
+        Write-Message -Level Verbose -Message "Looking for SMO in the Global Assembly Cache"
+		
+		$smolist = Get-ChildItem -Path "$($env:SystemRoot)\assembly\GAC_MSIL\Microsoft.SqlServer.Smo" | Sort-Object Name -Desc | select -ExpandProperty Name
```
wsmelton: Can't recall if we have any standard way of doing this but shorts it up a bit just doing something like below. You also do not need to use the subexpression for your environment variable.

```powershell
$smolist = (Get-ChildItem -Path "$env:SystemRoot\assembly\GAC_MSIL\Microsoft.SqlServer.Smo" | Sort-Object Name -Desc).Name
```

wsmelton: I also don't know if we want to stick with this being the only method to search for SMO assemblies. 

You may want to put a try catch around this and also include an alternate method as a failsafe of sorts. You can also search the registry key for it:

```powershell
$keyPath = HKLM\SOFTWARE\Microsoft\Microsoft SQL Server\SharedManagementObjects\CurrentVersion
(Get-ItemProperty -Path $keyPath -Name Version).Version
```

functions/Get-DbaSqlManagementObject.ps1
```diff
@@ -0,0 +1,88 @@
...
+		
+		$VersionList = [System.Collections.ArrayList]@()
+		
+		foreach ($version in $smolist)
+		{
+			$array = $version.Split("__")
```
wsmelton: If the alternate method above is used you would have to account for the registry value being returned in a different format "13.0.1601.5"

functions/Get-DbaSqlManagementObject.ps1
```diff
@@ -0,0 +1,88 @@
...
+					}
+					break
+				}
+			}
+			
+		} #foreach
+		
+		foreach ($version in $VersionList)
+		{
+			$version.LoadTemplate = "Add-Type -AssemblyName `"Microsoft.SqlServer.Smo, Version=$($version.Version), Culture=neutral, PublicKeyToken=89845dcd8080cc91`""
```
wsmelton: Not that this command will likely have performance issues as is, but seems like we could include this template property in your first foreach loop to prevent having to go over it all again.

As well, we tend to output as we build the object and not use the return, but again for this particular function, it may not be required.

functions/Get-DbaSqlManagementObject.ps1
dbaduck: OK, what should be the result in the begin when it is -lt 9?  Throw an exception?  Just want to do it the way it should be done to be consistent with dbatools.

functions/Get-DbaSqlManagementObject.ps1
dbaduck: I guess I could add the registry thing, but if it is in the registry it will be in the GAC.  So not sure this is valuable.

functions/Get-DbaSqlManagementObject.ps1
dbaduck: Good point, I can change that.

functions/Get-DbaSqlManagementObject.ps1
dbaduck: If it is in the registry it will be in the GAC, so not sure that putting this in should be there. I am open to whether or not you want it in.

functions/Get-DbaSqlManagementObject.ps1
wsmelton: [Found example](https://github.com/sqlcollaborative/dbatools/blob/0057052198ad6983479256aa34299b231650aa50/functions/Disable-SqlLogonTrigger.ps1#L85), just have it throw a message about support or SMO only exist for 2005 or higher should work.

functions/Get-DbaSqlManagementObject.ps1
wsmelton: Yeah, can't say I have tried it in every environment I've been in to know if it would matter. We can just leave it as is for right now if an issue comes up around this method we can just address it then.

functions/Get-DbaSqlManagementObject.ps1
dbaduck: Got it.  I will put that in.

functions/Get-DbaSqlManagementObject.ps1
dbaduck: OK. Great.

functions/Get-DbaSqlManagementObject.ps1
dbaduck: I have fixed the things up like the command in the link. It looks much more readable and standard. Thanks.

functions/Get-DbaSqlManagementObject.ps1
dbaduck: Fixed this too.

functions/Get-DbaSqlManagementObject.ps1
dbaduck: Fixed this.

functions/Get-DbaSqlManagementObject.ps1
PowerDBAKlaas: Agreed, as long as we don't have versions with letters or symbols, `-gt 9` is solid

PR #1486, opened 2017-06-26
Add CreateDate to default view of Get-DbaLogin function

galador:
Fixes # 1479

 - Adds the "CreateDate" property to the default view when running Get-DbaLogin 

How to test this code: 
- Run Get-DbaLogin
- Verify that CreateDate shows up in default output of the function.

### DIFF
--- a/functions/Get-DbaLogin.ps1
+++ b/functions/Get-DbaLogin.ps1
@@ -133,7 +133,7 @@ function Get-DbaLogin {
 				Add-Member -InputObject $serverLogin -MemberType NoteProperty -Name InstanceName -Value $server.ServiceName
 				Add-Member -InputObject $serverLogin -MemberType NoteProperty -Name SqlInstance -Value $server.DomainInstanceName
 
-				Select-DefaultView -InputObject $serverLogin -Property ComputerName, InstanceName, SqlInstance, Name, LoginType, LastLogin, HasAccess, IsLocked, IsDisabled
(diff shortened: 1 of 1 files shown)


potatoqualitee:
helll yessss

potatoqualitee:
That's nice as hell! works well all the way back to 2005. one issue, can you please submit a PR to address sql 2000? either skip it (but leave the column as null) or execute some sql that work for sql2k

currently getting 

```
Exception calling "ExecuteScalar" with "1" argument(s): "An exception occurred while executing a Transact-SQL statement or batch."
At line:131 char:5
+ ...             Add-Member -InputObject $serverLogin -MemberType NoteProp ...
    + FullyQualifiedErrorId : ExecutionFailureException

ComputerName : SQL2000
````

thank you 🙇

galador:
@potatoqualitee 

Looks like it's probably been like that for a little bit now. I actually don't have access to a SQL 2000 box...  I looked online, and it seems that there's not a reliable way to get the last login for SQL 2000. 

[This](https://social.msdn.microsoft.com/Forums/sqlserver/en-US/9b298293-2b48-437f-b7fd-3b9f8654aa9d/last-login-datetime-in-sql-server-2000?forum=sqlsecurity) says to be reliable, you need to have auditing enabled.

[This](https://stackoverflow.com/questions/1278609/sql-server-find-last-time-user-was-connected) says that there should be a field in `sys.syslogins`, but it doesn't seem to get updated on a new login.

I'll put in a PR to only pull that field in 2005+

PR #1493, opened 2017-06-27
Get-DbaLogin -- Only show `LastLogin` on versions > 2000

galador:
Fixes issue brought up in #1486 where SQL 2000 would throw an error.

 - LastLogin not reliably tracked in SQL 2000, so only show on 2005+

How to test this code: 
- Run `Get-DbaLogin sql2000` (SQL 2000 instance), and don't see an exception :)


potatoqualitee:
hey @galador - almost! I added a commit to handle if it's sql 2000 - we still want to ensure the column exists even if it's $null

potatoqualitee:
great work!

potatoqualitee:
And prompt, too :D thank you

PR #1495, opened 2017-06-27
Fixing improper merge artifacts

ConstantineK:
Had some issues with the VS Code merge, and even with testing it failed (must have had the command in memory already.)

This code removes the symbols that are traditionally used to delineate between the two branches being merged.


ConstantineK:
Please hold on merging, have another error in the file.

ConstantineK:
Added another commit to fix a bug? in Get-DbaBackupHistory as it was throwing errors when I had null LSN's in my database.

PR #1499, opened 2017-06-28
fixed library build command

FriedrichWeinmann:
Fixes # 

Fixes the issue with building the library when importing the module from anywhere but the module root.
I really should have seen that one coming, looking at the error in retrospect ...


potatoqualitee:
busy 🐝 !

PR #1501, opened 2017-06-28
Asynchronous TEPP - fixed conflict with main connection

FriedrichWeinmann:
Was accidentally using the same connection as the connection returned to the calling function of `Connect-SqlInstance`, rather than a copy of it.

# Fixes
Fixes errors such as this:
```
Exception calling "ExecuteWithResults" with "1" argument(s): "There is already an open DataReader associated with this Command which must be closed first."
At line:447 char:6
+ ...             $historyObject.FileList = $server.ConnectionContext.Execu ...
    + FullyQualifiedErrorId : InvalidOperationException
```

### DIFF
--- a/internal/Connect-SqlInstance.ps1
+++ b/internal/Connect-SqlInstance.ps1
@@ -129,7 +129,7 @@
         }
 		
 		# Register the connected instance, so that the TEPP updater knows it's been connected to and starts building the cache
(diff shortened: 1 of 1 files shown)


potatoqualitee:
ohh I see that copy there! nice nice. you should blog about this lil workaround! Why did you pick that over other workarounds?

FriedrichWeinmann:
Mostly because it's a more complete copy, but also because it was less effort. This way I only had to change 2 lines.

PR #1502, opened 2017-06-28
TEPP for configuration system / TEPP Bugfix

FriedrichWeinmann:
# Features
 - Added TEPP for the Configuration system commands
 - Added `-FullName` parameter to `Get-DbaConfig`. The legacy remnants made the usage less intuitive using `-Module` and `-Name` parameters

# Fixes
Fixed TEPP autocompletion being not shown when doing regular module import.
 - Internal commands could not be accessed.

Worked around that by making the commands publicly available but invisible / inaccessible to the user
 - Functions `Where-DbaObject` and `New-DbaTeppCompletionResult` were scoped to global, rather than local
 - The same functions were then set to visibility `"private"`

This makes them available to all code but not to users.

### DIFF
--- a/functions/Get-DbaConfig.ps1
+++ b/functions/Get-DbaConfig.ps1
@@ -7,6 +7,10 @@
 		.DESCRIPTION
 			Retrieves configuration elements by name.
 			Can be used to search the existing configuration list.
+	
+		.PARAMETER FullName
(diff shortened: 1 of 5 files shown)


potatoqualitee:
hell to the yes, thanks fred!

PR #1507, opened 2017-06-28
removed crap we dont want

zippy1981:
Fixes # Gitignore for the library

 - We are down to just two .gitignore files

How to test this code: 
- [ ] Make sure nothing broke
- [ ] Make sure after importing the module `git status` is clean

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally


.gitignore
```diff
@@ -11,6 +11,4 @@ dbatools.psprojs
 # ignore library
 bin/dbatools.dll
 bin/dbatools.xml
-bin/dbatools.pdb
-bin/projects/dbatools/dbatools/obj/*
-bin/projects/dbatools/*.user
```
wsmelton: If someone starts using the VS Project included in the repository the files generated on a build would get picked up by removing line 15 and 16. Is that intentional?

FriedrichWeinmann: He's excluding those on a lower level (see below)

PR #1517, opened 2017-06-29
Output added for Copy-DbaServerAuditSpecification

wsmelton:
 - formatted help
 - Add message system
 - Add logic to handle reverse copy, copying from higher version to lower version of SQL Server
 - update camelCase
 - Add output object

image

On reverse copy on versions:
image


potatoqualitee:
Man, can't thank you enough for all of this :shawn: 8 beers on me at Summit!

PR #1521, opened 2017-06-29
New Find-Commands & Find-DbaSP improv

ClaudioESSilva:
Find-DbaStoredProcedure - Add schema name and change to Write-Message
Find-DbaTrigger - Returns all triggers that contain a specific
case-insensitive string or regex pattern. Take into account the
different levels. Server, Database, Object
Find-DbaView - Returns all views that contain a specific
case-insensitive string or regex pattern.


potatoqualitee:
yay

potatoqualitee:
heck yahhhh nice work!!

PR #1522, opened 2017-06-29
SpConfig TEPP

potatoqualitee:


potatoqualitee:
hey @wsmelton - i havent tested these, do you have time? I needed them and noticed the configs were missing so I added them back

wsmelton:
Sure pulling it down now.

functions/Get-DbaSpConfigure.ps1
```diff
@@ -55,55 +55,58 @@ function Get-DbaSpConfigure {
 		[parameter(Position = 0, Mandatory = $true, ValueFromPipeline = $True)]
 		[Alias("ServerInstance", "SqlServer", "SqlServers")]
 		[DbaInstanceParameter[]]$SqlInstance,
-		[System.Management.Automation.PSCredential]$SqlCredential
+		[System.Management.Automation.PSCredential]$SqlCredential,
+		[object[]]$Config
```
wsmelton: I think this needs to match the property you have to return: `ConfigName` and not just `Config`. The object output for the value is set as `ConfigName`.

wsmelton: Go ahead and add `Silent` switch. Line 68 looks to be the only change required for `Write-Message`.

wsmelton:
Added few notes. The reference for naming the TEPP parameter ConfigName applies to the Set command as well. I forgot to add the review on that command as well.

wsmelton:
Adjusted the parameter name while you were sleeping 👿 

If you disagree you can always put it back. I did add `Config` as an alias on both commands, updated help, and examples.

potatoqualitee:
LOL, thank youuuuu. I've been sleeping since about 7pm yesterday

PR #1535, opened 2017-07-02
Was-Bound - Multiplayer upgrades

FriedrichWeinmann:
Enables `Was-Bound` to understand multiple parameters as input.
Implements both `-And` and `-Or` logic (was any one bound or were all bound) as well as the negation.
Defaults to `-Or`


potatoqualitee:
thanks! will test now

PR #1555, opened 2017-07-03
Fixed connection hogging by the logging system

FriedrichWeinmann:
Passing smo objects to the logging system as target would have it reuse the same connection the current commands were using when logging stuff.

This would cause commands - especially lengthy commands such as `Start-DbaMigration` that reuse connections to fail, due to the connection being busy.

This update converts all SMO objects that are passed as targets into non-SMO objects.
This is of obvious detriment to the logged data's level of detail, but I don't see a better compromise.

### DIFF
--- a/internal/Stop-Function.ps1
+++ b/internal/Stop-Function.ps1
@@ -108,8 +108,21 @@
         [string]
         $ContinueLabel
     )
-    
-    $records = @()
+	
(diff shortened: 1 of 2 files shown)


potatoqualitee:
fixing-a-sea-of-blood 😂

PR #1566, opened 2017-07-04
Get-DbaBackupHistory, introduced devicetype

niphlod:
Some love plus the introduction of filtering by
backupmediafamily.device_type. Some users (like me) have different VSS
agents running on some SQL servers and they create a full entry to the
msdb database.

Fixes # 

 - introduced devicetype

How to test this code: 
- [ ] compare with previous, then try to filter with devicetype to see if they match


potatoqualitee:
thanks for the examples and tests! works.

PR #1572, opened 2017-07-04
Get-DbaErrorLog - Redesigned for better performance

FriedrichWeinmann:
Redesigned the function from the keel up, in order to provide a significant boost in performance:
 - All instances are targeted in parallel
 - All processing is done on the target server, rather than the client
 - Used `Get-WinEvent` rather than `Get-EventLog`
 - Processing logs on the target system means, we no longer need to copy them to local
 - Processing logs line by line without keeping the entire log in memory eliminates crashes on gigantic logs
 - Added throttling that is actually respected

### DIFF
--- a/functions/Get-DbaErrorLog.ps1
+++ b/functions/Get-DbaErrorLog.ps1
@@ -1,6 +1,4 @@
-﻿
-function Get-DbaErrorLog
-{
+﻿function Get-DbaErrorLog {
 <#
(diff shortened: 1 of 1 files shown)


potatoqualitee:
LOL on fire tag

potatoqualitee:
git isnt seeing this branch so I'mma merge and add help, thanks all!

PR #1573, opened 2017-07-05
Get-DbaBackupHistory, filter by lsn

niphlod:
Speeds up -Last retrieval without processing the entire history

 - Introduced LastLsn parameter 

How to test this code: 
- [ ] Try fetching the last chain with -Last 
- [ ] Observe that the same resultset before this fix is the same but the time needed is hugely less


potatoqualitee:
lol hugely less!

niphlod:
yeah. would have gone for "super arci mega ultra extra less" but I felt compelled to restrain myself.

potatoqualitee:
thanks, bud! Tested and all works well - the results also look consistent with the previous veresion.

@Stuart-Moore - just fyi this command changed slightly.

Issue #1571, opened 2017-07-04
New-DbaAgentSchedule - Issue using example 6


mvillegascuellar:
## Is this:
_Just place "x" to indicate selection (e.g. "[x]")_

- [ ] *feature* 
- [x] *bug*:

When trying to do a test with example 6 from the help I get a message indicating to enter a start date, which I thought it would work from the very begining as the code comes from the help. So I entered the -Force parameter equals to 1 and it fails:

```
PS C:\Users\Administrator\Documents> New-DbaAgentSchedule -SqlInstance localhost -Schedule daily -FrequencyType Daily -FrequencyInterval 1
WARNING: [New-DbaAgentSchedule][22:00:16] Please enter a start date or use -Force 1 to use defaults.
[New-DbaAgentSchedule][22:00:16] Finished adding the job schedule(s).
PS C:\Users\Administrator\Documents> New-DbaAgentSchedule -SqlInstance localhost -Schedule daily -FrequencyType Daily -FrequencyInterval 1 -Force 1
New-DbaAgentSchedule : Cannot process argument transformation on parameter 'SqlCredential'. userName
At line:1 char:110
+ ... st -Schedule daily -FrequencyType Daily -FrequencyInterval 1 -Force 1
+                                                                         ~
    + CategoryInfo          : InvalidData: (:) [New-DbaAgentSchedule], ParameterBindingArgumentTransformationException
    + FullyQualifiedErrorId : ParameterArgumentTransformationError,New-DbaAgentSchedule
```

potatoqualitee:
dude, you're awesome! thanks for the reports. I've assigned the creator, @sanderstad, to take a look 👍

PR #1575, opened 2017-07-05
Fix example newdbaagentschedule #1571

sanderstad:
The error message still referred to the old way to assign the -Force
parameter. 
The example also didn't work, added -Force parameter.


potatoqualitee:
thank you!

PR #1587, opened 2017-07-05
Get-DbaDefaultPath

potatoqualitee:
Required because the SMO properties are sometimes blank (long story)


potatoqualitee:
🌮 party!

PR #1589, opened 2017-07-05
Fixed SQL query for Get-DbaTempdbUsage function

### LINKED ISSUE #1104: Get-DbaTempdbUsage - Improve output to be more accurate
## Is this:

- [ ] *feature* 
- [x] *bug*:

## System Details

```
Doesn't matter I guess, it's to do with the way I wrote the T-SQL.
```

- Output of dbatools version:

```
The output just says "awesome" but that's not helpful here.
```

```
First noticed in: Microsoft SQL Server 2012 (SP3)
```

#### Example:
1. Long running query with an `ORDER BY` that would require tempdb usage (or any other tempdb inducing query)...
2. Run Command:
```powershell
Get-DbaTempdbUsage -SqlServer '<server name>'
```
3. Received following error
```
No error received but will not show currently executing query.
```

#### Example:
The results are wrong! Wrong dammit!! Wrong I say!!!

#### Example:
```powershell
Get-DbaTempdbUsage -SqlServer '<server name>'
```
3. Receive following output (if applicable)
```
Results will show the tempdb usage of the currently executing query as well.
```

## Details
T-SQL currently uses `sys.dm_db_session_space_usage` which shows us the tempdb usage for the session but, looking for a way to return the offending query that works pre SQL Server 2012, when reading the documents I read this line and stop dead!

> The counters are updated only when a task ends; they do not reflect running tasks.
[Reference](https://docs.microsoft.com/en-us/sql/relational-databases/system-dynamic-management-views/sys-dm-db-session-space-usage-transact-sql#remarks)

Luckily there's also this guy - `sys.dm_db_task_space_usage` - which the docs tell us...

> Page counters are initialized to zero (0) at the start of a request. These values are aggregated at the session level when the request is completed
[Reference](https://docs.microsoft.com/en-us/sql/relational-databases/system-dynamic-management-views/sys-dm-db-task-space-usage-transact-sql#remarks)

So all we need to do is aggregate the counters from this DMV and add it to the values in `sys.dm_db_session_space_usage` to get our **real current total usage**.

Unfortunately this means that previous results of this command did not include any tempdb usage from currently executing queries.

Soz y'all 😢

shaneis:
Should take into account currently running sessions and parallel queries

Fixes #1104  

 - fixed SQL query to account for tempdb usage in currently running queries
 - Aggregates parallel queries to stop "false duplicates" showing
 - Added more information and columns to the result set

How to test this code: 
- [ ] Run on an active or inactive SQL Instance, should return sessions using tempdb, system or user.

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally

### DIFF
--- a/functions/Get-DbaTempdbUsage.ps1
+++ b/functions/Get-DbaTempdbUsage.ps1
@@ -38,54 +38,89 @@ Function Get-DbaTempdbUsage {
 	param (
 		[parameter(Mandatory = $true, ValueFromPipeline = $true)]
 		[Alias("ServerInstance", "SqlServer")]
-		[DbaInstanceParameter[]]$SqlInstance,
-		[PSCredential][System.Management.Automation.CredentialAttribute()]$SqlCredential,
+		[object[]]$SqlInstance,
+		[System.Management.Automation.PSCredential]$SqlCredential,
 		[switch]$Silent
 	)
 	
 	process {
 		foreach ($instance in $SqlInstance) {
 			try {
-				$server = Connect-SqlInstance -SqlInstance $instance -SqlCredential $sqlcredential
+				$server = Connect-SqlServer -SqlServer $instance -SqlCredential $sqlcredential
 			}
 			catch {
-				Stop-Function -Message "Failure" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
+				Stop-Function -Message "Failed to connect to: $instance" -Continue -Target $Instance
 			}
 			
 			if ($server.VersionMajor -le 9) {
 				Stop-Function -Message "This function is only supported in SQL Server 2008 or higher." -Continue
 			}
 			
-			$sql = "SELECT SERVERPROPERTY('MachineName') AS ComputerName, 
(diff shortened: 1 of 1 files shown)


ConstantineK:
Looks great and definitely returns information about currently running stuff, love the query. 

The only thing I was a tiny bit confused about was that this didnt return information about static objects that are taking up tempdb's internal space (eg if I make a gigantic and terrible temp table, its not returned in this), but after considering it for a bit this is (generally) more useful.

I dont know if a rename is useful, at first blush I dont have good new names besides Get-DbaTempDbQueryUsage or some other such monstrosity.

wsmelton:
Maybe a parameter like StaticUsage. 

Overall I have no interest in knowing that type of usage in tempdb. General troubleshooting I do is checking what is currently using tempdb. 


_Sent from my Samsung SM-G900V using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

potatoqualitee:
Fixed!

PR #1592, opened 2017-07-06
Get-DbaTable - Fix table only

ClaudioESSilva:
If we just use table name (no schema) the output is empty

### DIFF
--- a/functions/Get-DbaTable.ps1
+++ b/functions/Get-DbaTable.ps1
@@ -76,26 +76,43 @@ Returns information on the CommandLog table in the DBA database on both instance
 		
 		if ($Table) {
 			foreach ($t in $Table) {
-				$dotcount = ([regex]::Matches($t, "\.")).count
-				
+                $splitName = [regex]::Matches($t, "(\[.+?\])|([^\.]+)").Value
+				$dotcount = $splitName.Count
+				Write-Verbose "$dotcount"
 				$database = $NULL
 				$Schema = $NULL
 
-				if ($dotcount -eq 0) {
-				    $tbl = $t
-				}
+                switch ($dotcount) {
+                    1 {
+                        $tbl = $t
+                    }
+                    2 {
+                        $schema = $splitName[0]
+                        $tbl = $splitName[1]
+                    }
+                    3 {
+                        $database = $splitName[0]
+		                $schema = $splitName[1]
+		                $tbl = $splitName[2]
+                    }
+                    default {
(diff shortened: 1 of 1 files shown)


PowerDBAKlaas:
image

😮

PowerDBAKlaas:
`$tbl,$schema,$database = $t.Split(".")[-1..-3]`

can replace this whole block:

> $dotcount = ([regex]::Matches($t, "\.")).count
				$database = $NULL
				$Schema = $NULL
				if ($dotcount -eq 0) {
				    $tbl = $t
				}
				if ($dotcount -eq 1) {
					$schema = $t.Split(".")[0]
					$tbl = $t.Split(".")[1]
				}
				if ($dotcount -eq 2) {
					$database = $t.Split(".")[0]
					$schema = $t.Split(".")[1]
					$tbl = $t.Split(".")[2]
				}

But it will still be wrong when a database has dots in its name.

ClaudioESSilva:
Thanks for pointing that @PowerDBAKlaas !
I have committed a change. Please review it.

image

Thanks to @FriedrichWeinmann for the string with a bunch of characters that we call regex 😅

functions/Get-DbaTable.ps1
```diff
@@ -76,22 +76,43 @@ Returns information on the CommandLog table in the DBA database on both instance
 		
 		if ($Table) {
 			foreach ($t in $Table) {
-				$dotcount = ([regex]::Matches($t, "\.")).count
-				
+                $splitName = [regex]::Matches($t, "(\[.+?\])|([^\.]+)").Value
```
PowerDBAKlaas: OK,
now I need two lines to replace those 35:

> $test = $case -match '\[?(?<one>[^\]]*)?\]?.?\[?(?<two>[^\]]*)?\]?.?\[?(?<three>[^\]]*)\]?'
$database,$schema,$Table = switch (!$Matches.three + !$Matches.two + !$Matches.one){ 2 { $null, $null, $matches.one ;break } 1 { $null, $matches.one, $matches.two ;break } default { $matches.one, $matches.two, $matches.three }}

😄 
well, that handles the regex, the dotcount, the variable assignments and the [] replacements,
a catch for invalid parameter may still be needed

ClaudioESSilva: As we talked, I have added some verbosity to the help for Table parameter and added new example.

potatoqualitee:
🐒 🕺  thanks, yall! Fantastic teamwork 😍

PR #1607, opened 2017-07-07
Adding TEPP parameters

wsmelton:
This pretty much covers the basic set of dynamic parameters from a rough search, and ones for the Copy commands.

A few I'm not sure how to handle in a clean method that are missing from this PR...placing here for reference:
- Assembly (for each database)
- CMSGroup
- XEvent
- Policy (for PBM)
- DataCollection

Commands for the above:
```
Copy-SqlDatabaseAssembly.ps1
Copy-SqlCentralManagementServer.ps1
Get-SqlRegisteredServerName.ps1
Watch-SqlDbLogin.ps1
Copy-SqlDataCollector.ps1
Copy-SqlExtendedEvent.ps1
Get-DbaXEventsSession.ps1
Copy-SqlPolicyManagement.ps1
```

Some of the ones above you need a specific assembly to gather the info. I don't know if there may be any shortcuts or magic. If it is done the same way as you would any other function I can work on adding those.


potatoqualitee:
kawww!!

potatoqualitee:
thanks baw

PR #1613, opened 2017-07-07
Fixed Remove-DbaAgentSchedule fixes #1601

### LINKED ISSUE #1601: Remove-DbaAgentSchedule
## Is this:

- [ ] *feature* 
- [x] *bug*:

#### Example:
Running Remove-DbaAgentSchedule against a server with 1100 duplicate schedules I get the error below.

## Attached screen shots/console output

image

sanderstad:
Fixed issue with using wrong parameter with Stop-Function calls
Changed the proces of the schedule removal. It will now allow removing
multiple schedules instead of just one.

### DIFF
--- a/functions/Remove-DbaAgentSchedule.ps1
+++ b/functions/Remove-DbaAgentSchedule.ps1
@@ -111,51 +111,45 @@ Remove the schedule on multiple servers using pipe line
 						# Loop through each of the schedules and drop them
 						Write-Message -Message "Removing schedule $s on $instance" -Level Output
 
-						# Get the job ids 
-						$jobGuids = $Server.JobServer.SharedSchedules[$s].EnumJobReferences()
-
 						#Check if multiple jobs use the schedule
 						if ($jobCount -ge 1) {
-							if ($jobCount -gt 1 -and $Force) {
+							# Get the job object
+							$smoSchedules = $server.JobServer.SharedSchedules | Where-Object {($_.Name -eq $s)}
+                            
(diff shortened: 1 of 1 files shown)


gbargsley:
Reviewing the recent changes.  

Do you mean to remove the schedule even if it is related to a job?  This example there were two of the same schedule associated to the same job and it removed both.

Should that be with -force?  I know we did something similar for my situation last week, just was not sure if that was the default behavior you wanted.

image

potatoqualitee:
yo @sanderstad - can you take a look when you get a minute?

sanderstad:
I changed a lot in the code and I had to struggle a little to make sure it worked the same way it did before.

What the function does is check how many jobs are connected to the schedule. If this is more than one, and -Force is not being used, the function will error out saying you need the -Force parameter.
If there are is more than one job connected and -Force is being used the function will iterate through each of the jobs and remove the schedule.

What happened in your situation was that it was the other way around. You had one job with multiple schedules with the same name. You can have duplicate names and SQL Server accepts that because the uniqueness is in the schedule ID.
So what I did for that situation is combining the if statement to check whether there are more jobs with -Force or if the job count was equal to 1. It will iterate through the job, in your case iterate the same job lots of times and remove the schedule.

To make sure that schedules without jobs would also be removed I also select the schedules with the same name that have no jobs attached and iterate through the result to remove them. Because they're not attached to a job there will be no errors like in the other scenarios with attached jobs.

In theory not a lot of things changed in the behavior. Technically it works totally different and the procedures to remove the schedules are cleaner and easier to follow.

potatoqualitee:
k so good? 😆

Issue #1604, opened 2017-07-06
Install-DbaFirstResponderKit Replace Message


gbargsley:
## Is this:
_Just place "x" to indicate selection (e.g. "[x]")_

- [ ] *feature* 
- [x] *bug*:

## Steps to Reproduce
When running the command in the screen shot I get Confirm Folder Replace even when it was empty and I know the zip file and folder are not in the AppData\Local\Temp.

It appears to be like it is doing it twice.

It would also be nice if the output comment was the stored procedure was UPDATED or INSTALLED so you know.
## Attached screen shots/console output

![image](https://user-images.githubusercontent.com/169602/27925501-8cb14852-624a-11e7-937c-14d9355d96f2.png)

niphlod:
are you on the latest ? I can't reproduce from development

potatoqualitee:
another closed 💃

PR #1614, opened 2017-07-07
Install-DbaFirstResponderKit, added output

niphlod:
Related to #1604

 - added output to check if solution was installed/updated


potatoqualitee:
oh man i lik ethis! Can you please add it to Install-DbaWhoisActive 🙏 ?

PR #1618, opened 2017-07-07
Copy-DbaDatabase, fix for #1615

### LINKED ISSUE #1615: Copy-DbaDatabase, owner not set on destination
## Is this:

- [ ] *feature* 
- [x] *bug*:

## System Details
When the restore process completes, Update-SqlDbOwner gets called .... but it gets passed the SMO connection which is unaware of the newly created database. Everything gets trapped by Copy-DbaDatabase, which doesn't return that the process of updating the owner has been correctly done (because on the other end Update-SqlDbOwner happily skips over without an error).

This is a huge "security nightmare" if one has to trust Copy-DbaDatabase ... and it happens only for backuprestore.

niphlod:
Fixes #1615  
Let's go with "if there is a cmdlet and is not an internal one, use it instead" ?

 - instead of fixing update-sqldbowner, let's go with the "standard" Set-DbaDatabaseOwner

How to test this code: 
- [ ] copy-dbadatabase -backuprestore with a source db owner, verify that the destination owner is not the user copy-dbadatabase is running from, but correctly the one set on the source

### DIFF
--- a/functions/Copy-DbaDatabase.ps1
+++ b/functions/Copy-DbaDatabase.ps1
@@ -996,14 +996,15 @@ function Copy-DbaDatabase {
 
 					$dbFinish = Get-Date
 					if ($NoRecovery -eq $false) {
+						# needed because the newly restored database doesn't show up
+						$destServer.Databases.Refresh()
 						$dbOwner = $sourceServer.Databases[$dbName].Owner
-						if ($dbOwner -eq $null) {
+						if ($null -eq $dbOwner -or $destServer.Logins.Name -notcontains $dbOwner) {
 							$dbOwner = Get-SaLoginName -SqlInstance $destServer
 						}
-						Write-Message -Level Verbose -Message "Updating database owner"
-						try {
-							$null = Set-DbaDatabaseOwner -SqlInstance $destServer -Database $dbName -TargetLogin $dbOwner -Silent
-						} catch {
+						Write-Message -Level Verbose -Message "Updating database owner to $dbOwner"
+						$OwnerResult = Set-DbaDatabaseOwner -SqlInstance $destServer -Database $dbName -TargetLogin $dbOwner -Silent
+						if ($OwnerResult.Length -eq 0) {
 							Write-Message -Level Warning -Message "Failed to update database owner"
 						}
 					}
@@ -1019,7 +1020,7 @@ function Copy-DbaDatabase {
 
 					$dbOwner = $sourceServer.Databases[$dbName].Owner
 
-					if ($dbOwner -eq $null) {
+					if ($null -eq $dbOwner -or $destServer.Logins.Name -notcontains $dbOwner) {
 						$dbOwner = Get-SaLoginName -SqlInstance $destServer
 					}
 
--- a/functions/Copy-DbaDatabase.ps1
+++ b/functions/Copy-DbaDatabase.ps1
@@ -1004,6 +1004,7 @@ function Copy-DbaDatabase {
 						try {
 							$null = Set-DbaDatabaseOwner -SqlInstance $destServer -Database $dbName -TargetLogin $dbOwner -Silent
 						} catch {
+							Write-Message -Level Warning -Message "Failed to update database owner"
 						}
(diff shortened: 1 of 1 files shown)


functions/Copy-DbaDatabase.ps1
```diff
@@ -995,10 +995,17 @@ function Copy-DbaDatabase {
...
+						$dbOwner = $sourceServer.Databases[$dbName].Owner
+						if ($dbOwner -eq $null) {
+							$dbOwner = Get-SaLoginName -SqlInstance $destServer
+						}
 						Write-Message -Level Verbose -Message "Updating database owner"
-						$result = Update-Sqldbowner -source $sourceServer -destination $destServer -dbname $dbName
+						try {
+							$null = Set-DbaDatabaseOwner -SqlInstance $destServer -Database $dbName -TargetLogin $dbOwner -Silent
+						} catch {
+							Write-Message -Level Warning -Message "Failed to update database owner"
```
wsmelton: These needs to be updated to output the error, something that will be done for the command as a whole but in this situation it shows why it fails to update the owner.

The `Set-DbaDatabaseOwner` has not been updated to include the Silent parameter.

functions/Copy-DbaDatabase.ps1
```diff
@@ -995,10 +995,17 @@ function Copy-DbaDatabase {
...
-
 					if ($NoRecovery -eq $false) {
+						$dbOwner = $sourceServer.Databases[$dbName].Owner
+						if ($dbOwner -eq $null) {
+							$dbOwner = Get-SaLoginName -SqlInstance $destServer
+						}
 						Write-Message -Level Verbose -Message "Updating database owner"
-						$result = Update-Sqldbowner -source $sourceServer -destination $destServer -dbname $dbName
+						try {
+							$null = Set-DbaDatabaseOwner -SqlInstance $destServer -Database $dbName -TargetLogin $dbOwner -Silent
```
wsmelton: The output I get for this shows the update did not occur.
```
VERBOSE: [Copy-DbaDatabase][05:40:04] Updating database owner
VERBOSE: Connecting to SQL2016-01
VERBOSE: Gathering databases to update
VERBOSE: Updating 0 database(s).
VERBOSE: Performing the operation "Updating BrokerEnabled on avail" on target "SQL2016-01".
VERBOSE: [Copy-DbaDatabase][05:40:04] Successfully updated BrokerEnabled to True for avail on SQL2016-01
```

functions/Copy-DbaDatabase.ps1
```diff
@@ -995,10 +995,17 @@ function Copy-DbaDatabase {
 					}
 
 					$dbFinish = Get-Date
-
 					if ($NoRecovery -eq $false) {
+						$dbOwner = $sourceServer.Databases[$dbName].Owner
```
wsmelton: This should be set to `$destServer.Databases[$dbName].Owner` shouldn't it?

This assumes the login exist on the destination. Pulling it from the source then `$dbOwner` will never be null so `Get-SaLoginName` will never be called.

niphlod:
will check this in a bit: to be fair I'm working with a modified version of set-dbadatabaseowner which I'm rewriting so some _could_ be related to this

functions/Copy-DbaDatabase.ps1
niphlod: I copied the logic from below but I think it's an error. I'm going to update the code here and there to:
- get the owner from the source
- if it's null (some weird error may happen) OR the owner from source doesn't exist in destination, THEN fetch the sa from the destination

niphlod:
ok @wsmelton : latest commit fixes things but it's dependant on https://github.com/sqlcollaborative/dbatools/pull/1632

potatoqualitee:
there you go, niph!

potatoqualitee:
@wsmelton - can i get an approval?

potatoqualitee:
that;s merged and i'll be writing tests for copy-dbadatabase. if i have any issues, ill file an issue.

Issue #1586, opened 2017-07-05
rename variables starting with No to Exclude 


potatoqualitee:
`-NoSystemDbs` -> `-ExcludeSystemDbs`

potatoqualitee:
will be hell D:

niphlod:
again, tackling this via code reviews. Don't think this should remain opened, but happy to leave it if @potatoqualitee or @wsmelton ping me here :P

wsmelton:
It was primarily for the commands that had not been updated...fairly sure we got them all but not 100%.

niphlod:
part of ongoing effort to standardize things. Here too, given the sheer number of functions, either we make a whitelist and check in CI or we'll die trying.

PR #1631, opened 2017-07-10
Get-DbaTcpPort applied standard

wsmelton:
 - Formatted help
 - Added Silent parameter
 - Add message commands
 - Update catch blocks
 - Removed version check for 2000 to use `-MinimumVersion` parameter for `Connect-SqlInstance`
 - Added Query() method and adjusted property reference for the column returned
 - Changed filter from `DisplayName -eq "TCP/IP"` to be `Name -eq Tcp`, just reads better.
 - Changed `NoIpv6` to `ExcludeIpv6` as referenced in #1586 
 - Adjusted camelCase
 - Removed array declaration that was not being used, that I could see and test still ran.

### DIFF
--- a/functions/Get-DbaTcpPort.ps1
+++ b/functions/Get-DbaTcpPort.ps1
@@ -58,7 +58,7 @@ function Get-DbaTcpPort {
 			Returns an object with server name, IPAddress (just ipv4), port and static ($true/$false) for every server listed in the Central Management Server on sql2014
 	#>
 	[CmdletBinding()]
-	Param (
+	param (
 		[parameter(Mandatory = $true, ValueFromPipeline = $true)]
 		[Alias("ServerInstance", "SqlServer")]
 		[DbaInstanceParameter[]]$SqlInstance,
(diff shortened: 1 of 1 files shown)


gbargsley:
@wsmelton Doing testing and I cannot get Port information from a named instance on my local or information from remote servers.  All servers in my test are 2012 and up.  2016 being the two remote instances.

image

wsmelton:
That is something @FriedrichWeinmann or @potatoqualitee will have to address.

`Connect-DbaSqlServer` offers a `-NetworkProtocol` but the internal command does not have that option.

potatoqualitee:
This command should not be using the -NetworkProtocol - prepending TCP: is how it's worked since it's inception.

wsmelton:
@potatoqualitee that works fine if you are not testing against a named instance...
image

Named parameters are not parsed properly and it fails
image

gbargsley:
@wsmelton  What are the next steps on this PR?  Can I test it differently that is more in line with the functions design?

wsmelton:
@gbargsley it is going to fail against a named instance and I can't fix that... @zippy1981 or @FriedrichWeinmann will have to intervene to address parsing an instance name when it has one of the prefixed protocols included.

As it stands unless @potatoqualitee has concern we can go ahead and approve it and let it go because it is out of our control. Once the parameter type is fixed it will start working, nothing we can do in code to workaround it.

FriedrichWeinmann:
Expect a fix to the input interpretation within 24 hours (and I thought I _had_ fixed that :grml:)
Opened an issue for that:
https://github.com/sqlcollaborative/dbatools/issues/1687

potatoqualitee:
merging then, thanks @FriedrichWeinmann !

PR #1633, opened 2017-07-10
Adjust New-DbaAgentSchedule Parameter validation sets

### LINKED ISSUE #1570: New-DbaAgentSchedule - FrequencyType parameter showing numbers
## Is this:

- [ ] *feature* 
- [ x] *bug*:

When I try to create a new Agent Schedule, the parameter "FrequecyType" autocompletes numbers (1,128,16,32,4,64,8) and then the real Descriptive frequencies. I think numbers should be avoided for clarity of the command.

wsmelton:
Fixes #1570 

 - Formatted command
 - Cleaned up help and corrected a few misspellings
 - Cleaned up param block, added validation sets that appeared to have been missed
 - Removed validation sets with numbers, only strings.
 - Cleaned up some of the warning messages to be a bit more clear in less words
 - Cleaned up use of Stop-Function, removed adding the exception message to the Message parameter as that is handled by Stop-Function when you pass the error record.
 - Also changed `-InnerErrorRecord` to just `-ErrorRecord` as that parameter was updated.

I only changed logic in the BEGIN block, but because I formatted the whole document it will appear I made many more changes than I really did.

### DIFF
--- a/functions/New-DbaAgentSchedule.ps1
+++ b/functions/New-DbaAgentSchedule.ps1
@@ -1,610 +1,587 @@
 function New-DbaAgentSchedule {
-    <#
-.SYNOPSIS 
-New-DbaAgentSchedule creates a new schedule in the msdb database.
-
-.DESCRIPTION
-New-DbaAgentSchedule will help create a new schedule for a job. 
-If the job parameter is not supplied the schedule will not be attached to a job.
-
-.PARAMETER SqlInstance
-SQL Server instance. You must have sysadmin access and server version must be SQL Server version 2000 or greater.
-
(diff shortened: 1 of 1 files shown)


potatoqualitee:
hey @sanderstad - can you please take a look and approve?

wsmelton:
On use of `-not $var` or `!$var` I'm not sure what our decision was on the style guide. @potatoqualitee  ?

potatoqualitee:
Style guide left it open, but later discussions proved -not to be more reliable.

potatoqualitee:
@sanderstad please approve when youve got time

PR #1640, opened 2017-07-10
Copy-DbaLogin, fixes #303

### LINKED ISSUE #303: Copy-SqlLogin - Allow choice of default database
Allow choice of default database when "source instance" database isn't available

Adding a parameter to allow default database to be either master or tempdb could be good. Frankly I'd rather give a login access to tempdb than to master, in general.

Request from Steve Armistead

niphlod:
Fixes #303 

 - if the default database doesn't exist on destserver, use master if the login is a sysadmin, else tempdb

How to test this code: 
- Copy a login with a default database that doesn't exist on the destination and:
  - [ ] see that it gets "master" as default if it's a sysadmin
  - [ ] see that it gets "tempdb" as default  if it's not a sysadmin
- Copy a login with a default database that exist on the destination and:
  - [ ] see that it gets the proper database as default

### DIFF
--- a/functions/Copy-DbaLogin.ps1
+++ b/functions/Copy-DbaLogin.ps1
@@ -296,8 +296,17 @@ function Copy-DbaLogin {
 					$destLogin.Language = $sourceLogin.Language
 
 					if ($destServer.databases[$defaultDb] -eq $null) {
-						Write-Message -Level Warning -Message "$defaultDb does not exist on destination. Setting defaultdb to master."
-						$defaultDb = "master"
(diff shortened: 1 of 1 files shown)


gbargsley:
Both of the first two tests worked, but output informational message saying TempDB and master did not exist on destination.  That is not accurate.  

image

niphlod:
dang. wait a bit

niphlod:
@gbargsley : should be ok now

potatoqualitee:
😍

PR #1643, opened 2017-07-11
Connect commands - removed SetDefaultInitFields

potatoqualitee:
Forcing a few new autloads (that dont envoke enumerations) was good in theory but has caused some issues. In addition to evaluating the Server version, I would have also needed to evaluate the Client SMO version instead. 

I also encountered an issue with trying to export something yesterday and it was like "IsSystemObject doesn' exist!" I overlooked something there.

The original intent of this was to prevent enumerations for things like IsSystemObject but it just didn't work out. Bye!

### DIFF
--- a/functions/Connect-DbaSqlServer.ps1
+++ b/functions/Connect-DbaSqlServer.ps1
@@ -1,10 +1,10 @@
 Function Connect-DbaSqlServer {
     <#
 .SYNOPSIS
-Creates an efficient SMO SQL Server object.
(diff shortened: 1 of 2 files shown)


potatoqualitee:
It's important to note, however, that this should potentially be added back in the future. I added this to get Copy-DbaDatabase to copy 1000 commands in 8 minutes. Without it, it took infinity. 

I just don't have the focus right now to retest with Server & Client.

PR #1647, opened 2017-07-11
initial commit of Set-DbaPrivilege

PowerDBAKlaas:
 - finds SQL Service account and adds it to LPIM, IFI and/or BatchLogon privilege(s)

It would be nice to have `-Whatif` in this function, but I haven't found a way to implement that in `Invoke-Command.`
Maybe this function can be extended to allow other accounts to be granted privileges, or else a rename may be appropriate

How to test this code: 
- [ ] 
- [ ] 

- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Works remotely
- [ ] Works locally

### DIFF
--- a/functions/Set-DbaPrivilege.ps1
+++ b/functions/Set-DbaPrivilege.ps1
@@ -57,6 +57,7 @@ function Convert-UserNameToSID ([string] `$Acc ) {
 `$objUser = New-Object System.Security.Principal.NTAccount(`"`$Acc`")
 `$strSID = `$objUser.Translate([System.Security.Principal.SecurityIdentifier])
 `$strSID.Value
+}
 "@
 		$ComputerName = $ComputerName.ComputerName | Select-Object -Unique
 	}
@@ -73,7 +74,7 @@ function Convert-UserNameToSID ([string] `$Acc ) {
                 if ( $SQLServiceAccounts.count -ge 1 ) {
                     Write-Message -Level Verbose -Message "Setting Privileges on $Computer"
 				    Invoke-Command2 -Raw -ComputerName $computer -Credential $Credential -ArgumentList $ResolveAccountToSID, $SQLServiceAccounts, $BatchLogon, $IFI, $LPIM -ScriptBlock {
-					    Param ($ResolveAccountToSID, $SQLServiceAccounts)
+					    Param ($ResolveAccountToSID, $SQLServiceAccounts, $BatchLogon, $IFI, $LPIM)
 					    . ([ScriptBlock]::Create($ResolveAccountToSID))
 					    $temp = ([System.IO.Path]::GetTempPath()).TrimEnd("");
                         $tempfile = "$temp\secpolByDbatools.cfg"
@@ -82,7 +83,7 @@ function Convert-UserNameToSID ([string] `$Acc ) {
 					        ForEach ( $acc in $SQLServiceAccounts ) {
                                 $SID = Convert-UserNameToSID -Acc $acc;
                                 if ( $BLline -notmatch $SID ) {
-                                    (Get-Content $tempfile) -replace "SeBatchLogonRight = ","SeBatchLogonRight = *$SID" |
+                                    (Get-Content $tempfile) -replace "SeBatchLogonRight = ","SeBatchLogonRight = *$SID," |
                                     Set-Content $tempfile
                                     Write-Message -Level Verbose -Message "Added $acc to Batch Logon Privileges on $Computer"
                                 }
@@ -96,7 +97,7 @@ function Convert-UserNameToSID ([string] `$Acc ) {
 					        ForEach ( $acc in $SQLServiceAccounts ) {
                                 $SID = Convert-UserNameToSID -Acc $acc;
                                 if ( $IFIline -notmatch $SID ) {
-                                    (Get-Content $tempfile) -replace "SeManageVolumePrivilege = ","SeManageVolumePrivilege = *$SID" |
+                                    (Get-Content $tempfile) -replace "SeManageVolumePrivilege = ","SeManageVolumePrivilege = *$SID," |
                                     Set-Content $tempfile
                                     Write-Message -Level Verbose -Message "Added $acc to Instant File Initialization Privileges on $Computer"
                                 }
@@ -110,7 +111,7 @@ function Convert-UserNameToSID ([string] `$Acc ) {
 					        ForEach ( $acc in $SQLServiceAccounts ) {
                                 $SID = Convert-UserNameToSID -Acc $acc;
                                 if ( $LPIMline -notmatch $SID ) {
-                                    (Get-Content $tempfile) -replace "SeLockMemoryPrivilege = ","SeLockMemoryPrivilege = *$SID" |
+                                    (Get-Content $tempfile) -replace "SeLockMemoryPrivilege = ","SeLockMemoryPrivilege = *$SID," |
                                     Set-Content $tempfile
                                     Write-Message -Level Verbose -Message "Added $acc to Lock Pages in Memory Privileges on $Computer"
                                 }
@@ -119,8 +120,8 @@ function Convert-UserNameToSID ([string] `$Acc ) {
                                 }
                             }
                         }
-                        secedit /import /cfg $tempfile /overwrite /quiet
-				    } -ErrorAction SilentlyContinue
+                        $null = secedit /configure /cfg $tempfile /db secedit.sdb /areas USER_RIGHTS
+                    } -ErrorAction SilentlyContinue
 				    Write-Message -Level Verbose -Message "Removing secpol file on $computer"
 				    Invoke-Command2 -Raw -ComputerName $computer -Credential $Credential -ScriptBlock { $temp = ([System.IO.Path]::GetTempPath()).TrimEnd(""); Remove-Item $temp\secpolByDbatools.cfg -Force > $NULL }
 			    }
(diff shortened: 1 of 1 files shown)


potatoqualitee:
@PowerDBAKlaas  - please add the required param help!

Describe : Test help for Set-DbaPrivilege
Context  : Test parameter help for Set-DbaPrivilege
Name     : It gets help for parameter: BatchLogon : in Set-DbaPrivilege
Result   : Failed
Describe : Test help for Set-DbaPrivilege
Context  : Test parameter help for Set-DbaPrivilege
Name     : It gets help for parameter: IFI : in Set-DbaPrivilege
Result   : Failed
Describe : Test help for Set-DbaPrivilege
Context  : Test parameter help for Set-DbaPrivilege
Name     : It gets help for parameter: LPIM : in Set-DbaPrivilege
Result   : Failed


merci! :belgium:

PowerDBAKlaas:
added parameter help 🚒

gbargsley:
Did some testing and this failed for my local computer.

image

potatoqualitee:
Does this need a RunAsAdmin check, @PowerDBAKlaas ? I built an internal command called  `Test-RunAsAdmin`

niphlod:
not so sure it's even usable locally.

PowerDBAKlaas:
I think this says that remoting isn't enabled. @gbargsley ?

gbargsley:
I am running as windows admin on my laptop.  How do I know if remoting is running.  Other commands have been working on my laptop.
@PowerDBAKlaas

PowerDBAKlaas:
`Test-WSMan`
or
`Enter-PSSession localhost`

gbargsley:
Not sure what this shows.

image

potatoqualitee:
sessions like that won't work on localhost - our CM stuff or Invoke-Command2 handle if its localhost and just execute the commands directly - handles both empty credentials and localhost gracefully.

gbargsley:
Ok, so I cannot run this test against my local.  I will need to perform against a remote server?

potatoqualitee:
No, @PowerDBAKlaas needs to update the command - havent looked at the code. But i know Get-DbaPrivelege uses Invoke-Command2 so it shouldn't be too hard.

PowerDBAKlaas:
sessions work perfectly like that when remoting is enabled, @potatoqualitee , this is `Invoke-command2` after Test-PSRemoting, exactly like `Get-DbaPrivilege`
@gbargsley  you need to `enable-psremoting` on your localhost.

remoting will use WSMan, even on localhost.

PowerDBAKlaas:
@potatoqualitee `Test-RunasAdmin` quits if the user doesn't run elevated on localhost, right?
I saw something that 'reloads' the console to run as administrator, I think. I'm not sure it was someone from the team or not. Have you seen that?

potatoqualitee:
hey @PowerDBAKlaas - yeah I have a realoader in `Install-DbaWatchUpdate`

:O I had no idea you needed to enable remoting for Invoke-Command! 💩 How should I rewrite Invoke-Command2?

gbargsley:
Weird I run as administrator then login with my admin account it works.  I run as different user and login with my admin account it does not work.  I hate our security people.  Let me try the test again.

gbargsley:
I show the command running, but no output is generated.  And I do not show settings in the local security policy that was not there already.

image

image

PowerDBAKlaas:
@potatoqualitee I don't know if `Invoke-Command2` needs a rewrite. If we use remoting, remoting needs to be enabled. With CIM we fall back to DCOM, wich works, but it's still a pity we silently work around remoting. To catch the same problem with `Invoke-command` would require an elaborate logic to repeat the ScriptBlock without `Invoke-Command`, just for localhost. That's not economic.

But maybe `Test-RunasAdmin` could lead to 'Do you want to elevate this session?[y/n]' or something. That's probably not for now, but something to think about. I will add `Test-RunasAdmin` tomorrow. It will clearly state the user is not running in elevated mode, and that's an improvement.

@gbargsley I'm not sure. Maybe there's a flaw in the function logic.
I do have to catch the situation where no privilege switch is given, since it will just export secpol and then remove it again. That's useless.
Other than that I would expect it always returns a message when using `-Verbose`.

I'll look at it in +/- 12 hours from now :-)

PowerDBAKlaas:
added check to:
- stop if no privilege is chosen to be altered
- stop if the user tries to execute on localhost and is not running with elevated permissions

@gbargsley can you repeat your tests, please?

gbargsley:
Looks like it is getting stuck running.  This screen has been running for 2 minutes.  The output came after about 3 seconds and now it is just sitting there.


image

PowerDBAKlaas:
😮 
better ctrl+c it
must be hanging on the secedit 😞

gbargsley:
Is there any logging I can turn on to help troubleshoot?

gbargsley:
Here is -verbose.

image

gbargsley:
And a question for you.  When I run computer name actual computer name things fail.  When I use localhost I get further.  Is that supposed to be like that?

PowerDBAKlaas:
there should be a log in `Get-ChildItem $env:SYSTEMROOT\security\logs`
'scesrv.log'
but on my system it isn't created
yet the function executes 🤔 

@gbargsley can you show an example of computername failing? What statement do you execute and what's the error?

gbargsley:
image

PowerDBAKlaas:
@gbargsley seems as if remoting isn't enabled on L7DR15466

gbargsley:
But it is on localhost?  Isn't that the same thing?

PowerDBAKlaas:
yes, and remoting needs to be enabled on localhost.
`Invoke-Command` uses the WSMan protocol, on localhost or to the other side of the world makes no difference 😄

gbargsley:
But if it works on localhost without me doing anything.  What would I do different for computername.

PowerDBAKlaas:
those screenshots are all from the same computer?

do you get the same output for 
`Test-WSMan localhost`
and
`Test-WSMan L7DR15466`
?

gbargsley:
image

potatoqualitee:
hey @PowerDBAKlaas - I'm about to add -Force too All Add-Member - if you have some addmemebers, please add -Force as well. I'm getting errors when piping without it.

wsmelton:
The only way to test a group policy is with the GP analyzer. I'm not that is available in CLI form, have never tried or searched. You generally have that available within the GPMC that is only available one a server OS. 

I don't see this being a command that will be stable in any environment. There are to many controls at the domain level that can either prevent this or revert 15 minutes later. I always ask this to be enabled via a GP and not manually done on a given server.

potatoqualitee:
Where we at with this one, @PowerDBAKlaas ? Keep waiting?

gbargsley:
I cannot test this based on my computer security restrictions.  Last convo I had with @PowerDBAKlaas he thought that the graceful errors I am getting is good.  Not sure what needs testing.

potatoqualitee:
heyyyyyy go get some rest, @gbargsley ! 😷

PowerDBAKlaas:
@gbargsley I made a minor change. If you would be so kind to give it one more go with this new version, just repeating the same test, on 'localhost'. If it hangs again, can you see what's in $env:windir\security\logs\Scesrv.log ? If it fails and you get an error and your prompt back, that would be good news.

@potatoqualitee @wsmelton current status: the function works on my systems. It fails on some, as Garry demonstrated above. That's probably always because of user rights, and similar to other commands. The one that bothers me is when it just hangs. Maybe the `/quiet` can overcome that.
I find it a very useful function, but I really want to solve this problem.

It would also be nice if someone has an idea on a 'replacement' for ShouldProcess for Invoke-Command.

gbargsley:
@PowerDBAKlaas 
Ok, with the latest version I get failed to connect to computer name and it seems to try and work against localhost.  Is this valid for testing or is more work needed on this PR.

image

PowerDBAKlaas:
That's looking good, @gbargsley  Thanks!
So it wasn't a permissions issue after all, or did your security change?

@potatoqualitee I think this behaviour is acceptable, meaning it's probably the best we can make of it. As @wsmelton says, it'll always depend on some settings beyond our control.
Can you think of more tests?

potatoqualitee:
I can't think of a test but can I request one design change? Have it be -Type with a validated set instead of `-IFI` etc.

PowerDBAKlaas:
Sure. I considered that and decided to go the other way, but I can't remember why. 🤔 Probably because of the structure it once had. Seems totally valid now.
Allthough, `-Type` doesn't seem to be the correct name; they're the same type.
 `-Privilege`?

potatoqualitee:
It's the Type of Privilege - Type is used a ton within Microsoft's modules and so we adopted it hardcore in ours as well.

PowerDBAKlaas:
OK, that's done.
I'm using `Test-RunasAdmin`, is that still right or should I replace that?
And what do I use for credential (Windows)
`
		[PSCredential][System.Management.Automation.CredentialAttribute()]
		$Credential,`

potatoqualitee:
It is now `[PSCredential]$Credential,` until we can figure out a thorough solution.

It appears the new Test-RunAsAdmin is now `$null = Test-ElevationRequirement -ComputerName $instance -Continue`

potatoqualitee:
I will merge and fix 💃 MERCI!

PR #1651, opened 2017-07-11
Export-DbaScript - Improvement to allow export data

ClaudioESSilva:
- The parameter was $ScriptingOptionObject but on the code we were using $ScriptingOptionsObject (optionS plural)

image

- Even so, as it is we couldn't script data. To do this we need to use a object of type Scripter. This adds that object and assign the options ($ScriptingOptionObject) when they are passed.

To test:

$options = New-DbaScriptingOption
$options.ScriptData = $true

Get-DbaTable -SqlInstance "server" -Database "db1" -Table "dbo.tbl1" | Export-DbaScript -ScriptingOptionsObject $options -Path "D:\Temp\ScriptObject.sql"

### DIFF
--- a/functions/Export-DbaScript.ps1
+++ b/functions/Export-DbaScript.ps1
@@ -87,7 +87,7 @@
 	param (
 		[parameter(Mandatory = $true, ValueFromPipeline = $true)]
 		[object[]]$InputObject,
-		[Microsoft.SqlServer.Management.Smo.ScriptingOptions]$ScriptingOptionObject,
+		[Microsoft.SqlServer.Management.Smo.ScriptingOptions]$ScriptingOptionsObject,
 		[string]$Path,
 		[ValidateSet('ASCII', 'BigEndianUnicode', 'Byte', 'String', 'Unicode', 'UTF7', 'UTF8', 'Unknown')]
 		[string]$Encoding = 'UTF8',
@@ -140,6 +140,11 @@
 			
 			$server = $parent
 			$servername = $server.name.replace('\', '$')
+
+            if ($ScriptingOptionsObject) {
+                $scripter = New-Object Microsoft.SqlServer.Management.Smo.Scripter "$($server.name)" 
+                $scripter.Options = $ScriptingOptionsObject
+            }
 			
 			if (!$passthru) {
 				if ($path) {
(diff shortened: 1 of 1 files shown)


gbargsley:
The help syntax and parameter still show -ScriptingOptionSObject.
Testing with your code I get your output in ScriptObject.sql, but if I script out object in SSMS I get constraints and defaults which are not output in your process.  Is that an issue? I tried adding WithDependencies and that did not work.  And I am not getting any data output if that is the desired effect.

image


image

ClaudioESSilva:
The example is wrong. Need the 'S'. Fixed.

Which means that you are not using the most recent version. Otherwise you should get an error saying that parameter not exists. Like this:
image

Using your example it works
image

image

gbargsley:
Ok, my mistake.  I have tested again and I do get the data output.  I have three questions.
1.) I still do not see constraints and defaults in the output file for the table
2.) If I have an output file with the same name it is appending by default.  Should it overwrite by default and have an append parameter?
3.) What if I do a table with 1 million rows?  Will the output file handle that, should there be a parameter to limit output or have some kind of progress indicator?

image

ClaudioESSilva:
Hi @gbargsley ! Thanks for your testing on this!
1.) you need to use $options.DriAllConstraints = $true
2.) Good call. We use the -Append & -NoClobber in other Export- commands. Added that! Also add the help and a new example.
3.) This one we need to discuss it. In my opinion we can output a warning if the property -.ExportData is set to $true saying that can be slow depending on the number of existing rows. I don't think we can use a progress bar here.
@PowerDBAKlaas , @wsmelton, @ConstantineK, @niphlod, @potatoqualitee, @SQLDBAWithABeard thoughts?

gbargsley:
1. 1. Append in Example two is spelled wrong
2. 2. Get-Credential in Example 3 is spelled wrong
3. Example 5 needs true and false in the text to be consistent, variable or text

Testing shows -append works.  -NoClobber throws error that file exists.  Should it be a graceful error and not red.

I will wait until a decision is made on the large export.

image

ClaudioESSilva:
Fixed typos.
Added a validation on existing file for friendly warning

PowerDBAKlaas:
hmmm, @ClaudioESSilva I think the great win for this function would be to offer parameters and switches to let it compose the scriptingoptions. Without that, we can easily
`(Get-DbaAgentJob -SqlInstance Mysqlserver).script()`
The `-ScriptData` would then become a real dynamic param, and checks for the number of rows and size can be added. Then we have a useful warning, saying "Those 14 tables selected contain 2.4TB of data together! This could take a while. Do you still want to script the data?[y/n]"

If a user wants to learn how to write a scriptingoptions object, he'll likely know how to execute the scripter too instead of using this function as the lockstone of his script.

PR #1666, opened 2017-07-12
added helper info and stopped throw

potatoqualitee:


functions/Get-DbaDatabase.ps1
```diff
@@ -231,21 +231,26 @@ function Get-DbaDatabase {
...
-				Add-Member -Force -InputObject $db -MemberType NoteProperty BackupStatus -value $Notes
-
-				Add-Member -Force -InputObject $db -MemberType NoteProperty -Name ComputerName -value $server.NetName
-				Add-Member -Force -InputObject $db -MemberType NoteProperty -Name InstanceName -value $server.ServiceName
-				Add-Member -Force -InputObject $db -MemberType NoteProperty -Name SqlInstance -value $server.DomainInstanceName
-				Select-DefaultView -InputObject $db -Property $defaults
-
+			}
+			catch {
+				Write-Message -Level Warning -Message "Collection has been modified. Please use parens (Get-DbaDatabase ....) | when working with commands that modify the collection such as Remove-DbaDatabase" -Continue
```
FriedrichWeinmann: Use stop-function here, write-Message doesn't have a -Continue parameter.
Pass the error record along (`-ErrorRecord $_`)
Specify a `-Target $instance` to help debugging
Filter by exception type, there may be more than one and you don't want to cloak other errors by judging to soon. Also don't decide the error by message, since that'll change based on localization.

potatoqualitee:
Thanks @fred1

PR #1669, opened 2017-07-12
Export-DbaLogin, speedup enumeration

niphlod:
 - enums the database login mapping per instance, in order to save time enumerating login mappings which takes forever

How to test this code: 
- [ ] export before and after this PR, check that everything matches 
- [ ] export before and after this PR, passing a database (or a database list), check that everything matches (for the run including this PR, you can skip coffee break)


potatoqualitee:
Hell yes! thank you both!

Issue #1593, opened 2017-07-06
Get-DbaTable


ClaudioESSilva:
## Is this:
_Just place "x" to indicate selection (e.g. "[x]")_

- [ ] *feature* 
- [x] *bug*:

Try to find just by table name (no schema) and the output is blank. With schema worked.

#### Example:
1. Run Command:
```powershell
Get-DbaTable -SqlInstance DEV01 -Table MyTable
```
2. Received empty output. Even if the "MyTable" exists on one or multiple databases.

### Solution
Solved on PR: https://github.com/sqlcollaborative/dbatools/pull/1592

potatoqualitee:
Thanks for reporting and solving everyone :D :D :D :D

PR #1684, opened 2017-07-14
Change Get-DbaTable's input parameters to be additive.

galador:
I started looking at this function because of https://github.com/sqlcollaborative/dbatools/issues/1593, but it looks like the problem described there actually got fixed, but the issue is still open.

However, when I ran it, I noticed some weird results (the last command below). If you specify both `-Database` and the database name as a three-part name in `-Table`, you still get results for that table from "database1", even though the `-Database` filter should exclude anything not from "database2"

```powershell
# -- I have a table with the name `table1` in multiple databases
# Shows two tables 
Get-DbaTable -SqlInstance bhamby -Table "table1"

# Also shows two tables 
Get-DbaTable -SqlInstance bhamby -Table "dbo.table1"

# Shows one table from the correct database
Get-DbaTable -SqlInstance bhamby -Table "database1.dbo.table1"

# Shows one table from the database listed in `-Table` (expected to get no results)
Get-DbaTable -SqlInstance bhamby -Table "database1.dbo.table1" -Database database2
```

 - Changes the input parameters to be additive. Before, if you specified a database in both the table name (as a three-part name), and another database name in the `-Database` parameter, it would still show both copies of the table in the different databases.
 - If tables are not found, change try/catch to `if(count)`, because assignment of nothing will not trigger a throw. I changed the warning that there are no tables from `Warning` to `Verbose`... my thought is that, if a table isn't found, then just don't show it in the output. ﻿`¯\_(ツ)_/¯`
 - While processing, if a three-part name is specified, and it is not the database currently being processed, skip that table specification.
 - Formatted code with VS Code based on settings at https://dbatools.io/vscode/ (by the way, there needs to be a comma on the end of line 15 in the workspace settings listing on that page)

There was some discussion in Slack about how this should work (starting around [here](https://sqlcommunity.slack.com/archives/C1M2WEASG/p1499983857354510) (it's actually the message above that one, but the linking is weird because my question was a snippet))

How to test this code: 
- Run with various combinations of `-SqlInstance`, `-Database`, and `-Table` (with and without schema and database)

### DIFF
--- a/functions/Get-DbaTable.ps1
+++ b/functions/Get-DbaTable.ps1
@@ -1,10 +1,10 @@
 ﻿function Get-DbaTable {
-	<#
+<#
 .SYNOPSIS
 Returns a summary of information on the tables
 
 .DESCRIPTION
-Shows table information around table row and data sizes and if it has any table type information. 
+Shows table information around table row and data sizes and if it has any table type information.
 
 .PARAMETER SqlInstance
 SQLServer name or SMO object representing the SQL Server to connect to. This can be a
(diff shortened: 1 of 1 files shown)


potatoqualitee:
LOL "¯\_(ツ)_/¯"

Fantastic pull request, thank you! Our QA @gbargsley will take a look shortly 💃

functions/Get-DbaTable.ps1
```diff
@@ -77,126 +77,146 @@ Returns information on the CommandLog table in the DBA database on both instance
...
-								Write-Message -Level Warning -Message "Could not find table name: $($fqtn.tbl) schema: $($fqtn.schema)" -ErrorRecord $_
+							else {
+								$skipped = $false
 							}
 						}
-						else {
-							try {
-								$tables = $db.Tables | Where-Object { $_.name -eq $tbl }
+
+						$tables = $db.Tables | Where-Object { $_.Name -eq $fqtn.Table }
```
PowerDBAKlaas: This is a valid solution, @galador , but it's probably faster to avoid a few `if` and pipe to `Where` only once.

Something like this, maybe:
`$tables = $db.tables | Where-Object { $_.name -eq $fqtn.tbl -and $fqtn.schema -in ($schema,$null) -and $fqtn.database -in ($_.Parent.Name,$null)}`

potatoqualitee:
Thanks so much all! I will merge this and then test the change that @PowerDBAKlaas suggested with the commands in Garry's screenshot.

Awesome work, @galador!

potatoqualitee:
Hey klaas - i tried it and got 0 results so i discarded it but added anew lil column.

PowerDBAKlaas:
sorry, I didn't test that, I just wrote it from the top of my head.
that's why I said "somehting like this" 😄 

I think I used `$fqtn.tbl` what doesn't exist, it should be `$fqtn.table`

galador:
@potatoqualitee @PowerDBAKlaas 

I think the right line would be (there was also a missing `$_.` on schema):

```powershell
$tables = $db.tables | Where-Object {$_.name -eq $fqtn.Table -and $fqtn.Schema -in ($_.Schema, $null) -and $fqtn.Database -in ($_.Parent.Name, $null)}```

I tested it on my machine, and it worked how I expected. I can make another pull request, if needed.

PowerDBAKlaas:
haaa, that's the correct one 😆 
well done @galador 
I think it would indeed be best to submit a new PR,
@potatoqualitee ?
Then @gbargsley  can have another round of testing. May be interesting to see a performance comparison too.

potatoqualitee:
Yep, I'm down!

PR #1696, opened 2017-07-16
Get-DbaAvailabilityGroup quick fox of Replicas property

wsmelton:
Adjust spelling error and Replicas property to be the actual property from SMO.


potatoqualitee:
🌮

PR #1705, opened 2017-07-17
Detach/Attach Functions

potatoqualitee:
Still need some examples then we're set


potatoqualitee:
@sanderstad - these commands should make your day

PR #1729, opened 2017-07-18
DbaInstanceParameter Added fallback for netname when not available

FriedrichWeinmann:
Fixes # 

 - In SMO Versions where NetName is not available, it will fall back to using `ComputerNamePhysicalNetBIOS` instead

### DIFF
--- a/bin/projects/dbatools/dbatools/Parameter/DbaInstanceParameter.cs
+++ b/bin/projects/dbatools/dbatools/Parameter/DbaInstanceParameter.cs
@@ -400,7 +400,8 @@ namespace Sqlcollaborative.Dbatools.Parameter
                 case "microsoft.sqlserver.management.smo.server":
                     try
                     {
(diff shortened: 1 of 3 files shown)


potatoqualitee:
so good so good

bin/projects/dbatools/dbatools/Parameter/DbaInstanceParameter.cs
```diff
@@ -400,7 +400,8 @@ public DbaInstanceParameter(object Input)
                 case "microsoft.sqlserver.management.smo.server":
                     try
                     {
-                        _ComputerName = (string)tempInput.Properties["NetName"].Value;
```
zippy1981: Can we do a keyExists() check instead of a try catch? throwing an exception is expensive (it locks every core, granted for nanoseconds). exceptions should be for stuff we really don't expect to happen.

PR #1757, opened 2017-07-19
updates test to address issue with non refreshed databases property

potatoqualitee:


potatoqualitee:
This failed once on one test

Issue #1712, opened 2017-07-17
Awesome Availability Groups Commands


potatoqualitee:
### Problem to solve
Building set of commands around building and managing an Availability Group.

### Additional information
Let's make a list of ideas and corresponding command names

We can focus on both building blocks _and_ whole solutions.

The following complete solutions have been proposed:

- [ ] Finding the date of the most recent backup across all replicas of an AG database
- [ ] Restore a backups (with logs, etc) to all nodes; Add a restored DB to the AG
- [ ] Adding an existing DB to an AG, using existing backups

wsmelton:
Prerequisite to this will be the ability to do partial restores with the `Restore-DbaDatabase`. I would need the ability to start a restore without having to anchor with full backup first.

@Stuart-Moore

wsmelton:
@potatoqualitee @DevOpsFu (and @PowerDBAKlaas to make sure naming convention makes sense) 

Just putting this post as one we can edit to build out the list of commands we actually need

Commands:
- [x] Get-DbaAgReplica
- [x] Get-DbaAgDatabase
- [ ] Get-DbaAgListener
- [ ] Get-DbaAgReplicaState (based on SMO object to get latency info)
- [ ] Get-DbaAgHealth (based on XE session)
- [ ] Test-DbaAgHealth (seen use cases where based on how the AG was built the XE session does not get turned on)
- [ ] Set-DbaAgHadr (enable hadr for a given instance, initial step to building an AG is it has to be enabled for the instance; could have switch for -ExcludeRestart to say don't restart the instance but default it will enable it and then restart it to apply)
- [ ] Set-DbaAgPrimary (proposed in issue #1432 which has "server" in the name but thinking we don't need to include that?)
- [ ] New-DbaEndpoint (generic but required to build an AG; will also need to ensure the output of this command flows well for the AG build process)
- [ ] New-DbaAgGroup
- [ ] New-DbaAgListener (need to determine best process for this, scenarios like Azure IaaS you don't follow the same process as you would on-premise or in AWS)

PowerDBAKlaas:
Looks like a fine collection, @wsmelton 
Only the `New-DbaAgDatabase`, does it create a DB, or is it more an `Add-` or `Join-`?

potatoqualitee:
i do like join or add!

wsmelton:
I would be both, you have to add the database to the replica and then join it to the AG. You would join on each replica. So that name could go either way @potatoqualitee

[...]

PR #1770, opened 2017-07-19
Get-DbaAgDatabase - Initial version

wsmelton:
 - New command

image

Also partly in preparation for working on #1712


wsmelton:
What is the opinion on having the `Replica` property? That is not a property in the default SMO object but it seemed like it should be there, though it is the same as ComputerName. Just didn't feel right now having that property.

gbargsley:
I think it is good info.  Would it be possible to add primary/secondary to each AG in the output?

potatoqualitee:
Will include - if you want to add features, please do so and create a new Pr 💃

PR #1804, opened 2017-07-21
Major update to issue and pr template

wsmelton:
We worked template after finding out how well markdown comments can come in handy.

As well, no one was really using the previous version to the extent we had planned or thought it would help.


potatoqualitee:
these are great!

PR #1836, opened 2017-07-22
Update Permission commands to standards - and cleanup

wsmelton:
## Type of Change

 - [X] Breaking change (effects a command or functionality)
 - [ ] Pester test is included

### Purpose
 
- Address applying formatting standards for `Sync-DbaLoginPermission` and `Update-SqlPermissions`
- Removing duplicate code from `Sync-DbaLoginPermission` (`Update-SqlPermissions` was moved to internal function)
- Found bug when you specify `-Login` for `Sync-DbaLoginPermission` a warning message still outputs around the current user being sync'd.

### Approach

I started with just the purpose of cleaning up `Update-SqlPermissions` (e.g. expanded help to include the parameter list as minimum, applied formatting, etc). In the desire to test that command I checked the references and found `Sync-DbaLoginPermission` is the only command that uses it. However, once I got into that command it was not actually using the internal command, it was calling the copy of that function within the begin block.

So, went ahead and cleaned up the LoginPermission command and then removed the copy of the update function within that file. In testing the LoginPermission command picked up the warning message was outputting when it shouldn't. If you specify a single login to sync it barks about sync'n the current login, which was not specified.

### Screenshots

False warning that was being displayed

image

Corrected logic where warning does not show when a specific login is passed in, also added proper grammar to verbose messages so they all have the same pattern.

image


potatoqualitee:
you're beyond awesome! thank you :shawn:!

PR #1837, opened 2017-07-23
Get-DbaDatabase - add LogReuseWaitStatus to default view

wsmelton:
## Type of Change

 - [ ] Pester test is included

### Purpose
 
Add the LogResuseWaitStatus to the default view.

### Approach

The log reuse is one of the most common values to check for databases in SQL Server. This is used to check why the log file is filling up or growing out of control in most cases.

Being that we provide the last log backup it makes sense to also include this value. I chose to add the property alongside the recovery model as it is most directly related to that property.

### Commands to test

`Get-DbaDatabase -SqlInstance -Database master`
### Screenshots

image


potatoqualitee:
Very cool!

PR #1862, opened 2017-07-23
Set-DbaStartupParameter - fixes #1380

### LINKED ISSUE #1380: Set-DbaStartupParameter missing SqlCredential
## Is this:

- [ ] *feature* 
- [x] *bug*:

## Bug description
Set-DbaStartupParameter does not as it is support specifying explicit SqlCredentials. Thus both Smo connections and PowerShell remoting are handled using the same credentials.
This should be split up, so database connections can be established using their own, dedicated credentials.

potatoqualitee:


potatoqualitee:
please review/test @gbargsley

Stuart-Moore:
Looks good to me :+1:

will merge

PR #1878, opened 2017-07-24
Fix logshipping functions2

sanderstad:
Lots of bugs removed
Cleaned up code
Removed usage of Invoke-DbaSqlCmd


potatoqualitee:
thanks bud!

PR #1882, opened 2017-07-25
Get-DbaServerprotocol use DbaCmObject

PowerDBAKlaas:
## Type of Change

 - [ ] Pester test is included

### Purpose
 
- Reuse functionality of Get-DbaCmObject
- use messaging standard
- improve speed
### Approach

- replace `Get-CimInstance` with `Get-DbaCmObject`
- replace `Write-Verbose` with `Write-Message`

### DIFF
--- a/functions/Get-DbaServerProtocol.ps1
+++ b/functions/Get-DbaServerProtocol.ps1
@@ -71,63 +71,29 @@ PROCESS
           if ( $Server.ComputerName )
           {
               $Computer = $server.ComputerName
-              Write-Verbose "$FunctionName - Getting SQL Server namespace on $Computer via CIM (WSMan)"
-              $namespace = Get-CimInstance -ComputerName $Computer -NameSpace root\Microsoft\SQLServer -ClassName "__NAMESPACE" -Filter "Name Like 'ComputerManagement%'" -ErrorAction SilentlyContinue |
-                          Where-Object {(Get-CimInstance -ComputerName $Computer -Namespace $("root\Microsoft\SQLServer\" + $_.Name) -ClassName ServerNetworkProtocol -ErrorAction SilentlyContinue).count -gt 0} |
+              Write-Verbose "$FunctionName - Getting SQL Server namespace on $Computer"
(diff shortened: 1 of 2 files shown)


potatoqualitee:
@PowerDBAKlaas - I wrote a Test for this command that passes 👍 But I don't like the output from the new Cm commands and would like to get that addressed prior to merging.

@FriedrichWeinmann - I submitted a PR to remove the exclamation point but will probably need to up the library's version #1884. Either way, the command works, but it complains, seemingly unnecessarily

image

potatoqualitee:
Will wait to merge with the other one

wsmelton:
If you don't like the output of `Get-DbaCmObject` then it should be addressed in that command. But what output are you talking about?

potatoqualitee:
Could not find valid protocol warning at the top that comes from the cm command

wsmelton:
If you issue that same command with Verbose and post that output?

wsmelton:
The stop function in that line show set to continue so yeah the phrase in that message is inaccurate.

FriedrichWeinmann:
Well now ... that's mightily odd. Looks like I seriously need some quality time with `Get-DbaCmObject`, both for the performance issue, as well as this. Obviously, throwing an exception _and_ returning the requested information is a bit contradictory.
Looking into it.

functions/Get-DbaServerProtocol.ps1
```diff
@@ -55,12 +58,12 @@ Param (
   [parameter(ValueFromPipeline)]
   [Alias("cn","host","Server")]
   [string[]]$ComputerName = $env:COMPUTERNAME,
```
wsmelton: Need to set this to `DbaInstanceParameter`

functions/Get-DbaServerProtocol.ps1
```diff
@@ -71,67 +74,33 @@ PROCESS
           if ( $Server.ComputerName )
           {
               $Computer = $server.ComputerName
```
wsmelton: Same question as other PR do we need to add a `Resolve-DbaNetworkName`?

potatoqualitee: looks like someone added it 💃

potatoqualitee:
merging to prevent future conflicts. a release or two with slower commands is less headache 🙉

PR #1886, opened 2017-07-25
Update hardcoded url to parameter

slimrichard:
The error was giving an incorrect path and is hardcoded to a version. This should be the same parameter used to attempt the download.

Really small change.

## Type of Change


potatoqualitee:
Great patch, thanks so much for the fix @slimrichard ! 🌮

PR #1887, opened 2017-07-25
Get-DbaSqlservice use DbaCmObject

### LINKED ISSUE #1810: Get-DbaSqlService - rework and cleanup
### Problem to solve
- Current code of the command seems to take some time to run in certain environments. [ @gbargsley noted on about 250 servers it can take 25-31 minutes to complete.]
- Needs to be adjusted to utilize the CMI/WMI commands that are now in the module.

PowerDBAKlaas:
## Type of Change
 - [ ] Bug fix (non-breaking change, fixes #1810)
 - [ ] Pester test is included

### Purpose
- reuse Cim functionality
- use messaging standard
- output sqlservice object instead of PSCustomObject
- improve speed 

### Approach
- replace `Get-cimInstance` by `Get-DbaCmObject`
- replace `Write-verbose` by `Write-Message`
- return sqlservice object

### Learning
+ instead of speeding up, these changes are actually slowing it down:
on my domain, it went from 14-15 sec to 21-31 sec for 20 computers


potatoqualitee:
Unfortunately,  @FriedrichWeinmann, we can confirm the slowness in the lab. 

`'sql2005','sql2008','sql2008\sql2k8','sql2012','sql2014','sql2016a','sql2016b','sql2016c' | Get-DbaSqlService` 

Development branch: 5.28 seconds
SqlService branch: 10.71

@PowerDBAKlaas - I will keep this PR open until the speed issue is resolved.

FriedrichWeinmann:
Will look into it!
The proportions should rather be the opposite actually ... hm ...
May take a bit - got a lot on my plate from my company, may need a few days before I can take a look, sorry :(

functions/Get-DbaSqlService.ps1
```diff
@@ -58,7 +61,9 @@ Param (
   [Alias("cn","host","Server")]
   [string[]]$ComputerName = $env:COMPUTERNAME,
```
wsmelton: This needs to be `DbaInstanceParameter`

functions/Get-DbaSqlService.ps1
```diff
@@ -75,84 +80,39 @@ PROCESS
             if ( $Server.ComputerName )
             {
                 $Computer = $server.ComputerName
-                Write-Verbose "$FunctionName - Getting SQL Server namespace on $Computer via CIM (WSMan)"
-                $namespace = Get-CimInstance -ComputerName $Computer -NameSpace root\Microsoft\SQLServer -ClassName "__NAMESPACE" -Filter "Name Like 'ComputerManagement%'" -ErrorAction SilentlyContinue |
-                            Where-Object {(Get-CimInstance -ComputerName $Computer -Namespace $("root\Microsoft\SQLServer\" + $_.Name) -Query "SELECT * FROM SqlService" -ErrorAction SilentlyContinue).count -gt 0} |
+                Write-Message -Level Verbose -Message "Getting SQL Server namespace on $Computer"
+                $namespace = Get-DbaCmObject -ComputerName $Computer -NameSpace root\Microsoft\SQLServer -Query "Select * FROM __NAMESPACE WHERE Name Like 'ComputerManagement%'" -ErrorAction SilentlyContinue |
```
wsmelton: Don't know if we should add a `Resolve-DbaNetworkName` as we do that in some of the other commands that call DbaCmObject?

As well, line 71 `$FunctionName` can be removed (appears to still be there from GitHub view).

wsmelton:
@potatoqualitee are we going to wait for the dependent commands to be fixed before merging this update?

wsmelton:
@PowerDBAKlaas I'm just curious after noting this command in current version of development branch takes about 8 seconds to return the SSIS services on my machine:

image

Is there any difference if you have this use WMI.ManagedComputer instead of CIM?

```powershell
if (-not (Get-Module dbatools)) {
	ipmo .\dbatools.psm1
}
$scriptBlock = {
	$wmi.Services | ForEach-Object {
		[PSCustomObject]@{
			ServiceName  = $_.Name
			DisplayName  = $_.DisplayName
			StartName    = $_.ServiceAccount
			ServiceType  = $_.Type
			State        = $_.ServiceState
			StartMode    = $_.StartMode
		}
	}
}

Invoke-ManagedComputerCommand -ComputerName manatarms -ScriptBlock $scriptBlock | Where-Object ServiceType -eq 'SqlServerIntegrationService'
```

The above code returned information under a second:

image

PowerDBAKlaas:
If I recall correctly, I had issues when the highest namespace doesn't contain any services:
after a partial install, or an install and remove, or something.
I think the managedcomputer command doesn't catch that.

wsmelton:
How did you test that? If Wmi.ManagedComputer does not see the service then Configuration Manager does not function and you can't modify any of the services in a supported manner. If that is the case there is no guarantee the services are functioning anyway at that point.

I would also think you are going the same route, just using CIM to get what ManagedComputer already has...unless I read your code wrong.

PowerDBAKlaas:
`$namespace = Get-CimInstance -ComputerName $Computer -NameSpace root\Microsoft\SQLServer -ClassName "__NAMESPACE" -Filter "Name Like 'ComputerManagement%'" -ErrorAction SilentlyContinue |
                            Where-Object {(Get-CimInstance -ComputerName $Computer -Namespace $("root\Microsoft\SQLServer\" + $_.Name) -Query "SELECT * FROM SqlService" -ErrorAction SilentlyContinue).count -gt 0} |
                            Sort-Object Name -Descending | Select-Object -First 1`

This selection solved it in my case. I had 'ComputerManagement10', 'ComputerManagement11' and 'ComputerManagement12' on one computer, but 
+ ...10 didn't return the higher versions of the services.
+ ...12 didn't have any 'SqlService' objects.
So I came up with "I need the highest namespace that does have sqlservice class objects".

I think this had something to do with it too: https://www.powerdba.eu/powershell-for-the-dba/managedcomputererror

wsmelton:
I can't say I've ever seen or had that issue before.

nvarscar:
Got the following yesterday when was testing Get-DbaSqlService (and Stop-*) with more than 10 machines at a time:
image
There were no credentials specified in the function call.
This error only exists in the current session and do not reappear in the new PS session.

potatoqualitee:
I  uhm, I don't know when the Cm commands will be available and there are now conflicts with this command. My apologies, Klaas, but I've got to close this one.

nvarscar:
To be fair, new Get-DbaSqlService uses huge bits of code from this branch, so basically this code was actually merged, but from a different source 😃

PR #1893, opened 2017-07-25
Extra feature for log shipping

sanderstad:
Added option to prompt the user when the secondary database is not
initialized.
The user will not get a choice when -GenerateFullBackup and/or -Force
are not being used.


potatoqualitee:
bon bon bon! 🍖

PR #1897, opened 2017-07-25
Initial Commit of Set-DbaCompression and update to Test-DbaCompression

### LINKED ISSUE #1714: New Command Set-DbaCompression
## Is this:

- [x] *feature* 
- [ ] *bug*:

## System Details

- Operating system name and version:

```
Evaluate $PSVersionTable in PowerShell and paste the output here
```

- Output of dbatools version:

```
Evaluate (Get-Module dbatools -ListAvailable).Version and paste output here
```

#### Example:
1. Created database/object/files/etc
2. Run Command:
```powershell
Get-DbaFunction -SqlInstance MyServer | Get-DbaSomeOtherFunction -FilterSwitch -StelleEyedMissileMan
```
3. Received following error
```powershell
Get-DbaFunction: parameter StelleEyedMissileMan cannot be bound due to electrical system failure
```

#### Example:
The launch of Apollo 12 failed due to electrical system failure

#### Example:
1. Command will set your Table/Index compression to the best recommended PAGE or ROW compression setting
2. Run Command:
```powershell
Get-DbaFunction -SqlInstance MyServer | Set-DbaCompression
```
3. Receive following output (if applicable)
```powershell
Your compression settings have been updated to the best recommended settings.
```

## Attached screen shots/console output

- Use of Start-Transcript can help collection of console output and exceptions
- Provide screen shots of the output from your console if available.

### Import - Attach dbatools logs
- Latest release of dbatools now includes a messaging and logging system. You can find this information via the `Get-DbaConfig` function. Use of the Set-DbaConfig controls the configuration of the system.
- Ensure this returns **true**: `(Get-DbaConfig -Name errorlogenabled).value`
- Locate the log path for the error logs using this command: `(Get-DbaConfig -Name dbatoollogpath).Value`
- **Attach the latest xml and log files found in the path shown.**

js0505:
## Type of Change

 - [ ] Bug fix (non-breaking change,)
 - [x] New feature (non-breaking change, adds functionality  fixes #1714)
 - [ ] Pester test is included

### Purpose
 
Added new function Set-DbaCompression
Updated function Test-DbaCompression


gbargsley:
I am attempting to run tests against Test-DbaCompression and get no results.  With the -verbose turned on it shows the DB's, but no compressions suggestions.

image

js0505:
gbargsley - can you just take the SQL out and run it in SSMS and let me know what it does.  I have attached my output.
test-dbacompression

js0505:
I also just updated Set-DbaCompression as I did not remove the '_' from the output table and results were not populating correct.  I should be able to merge into this PR.

potatoqualitee:
Thanks Jason! So far, I'm getting this on two databases on one server

image

These two databases have their compat set to SQL 2005 or above (tested 2005 compat and it threw no errors)

> if ($db.Compatibility -eq 'Version80' ) { Stop-Function -Continue .... }

potatoqualitee:
````
ROWestimatePctoforig
PAGEestimatePctoforig
CompressionTypeRecommendation
sizecurKB
sizereqKB
percentcompression
```

Please update this to RowEstimatePercentOriginal, PercentCompression, SizeRequirementKB, SizeCurrentKB (or whatever it is, I'm unsure of the column names, which means it should be updated/clarified)

Thanks much! Looking super forward to adding these!

js0505:
Will provide update here in a bit!  Thanks for the notes and additional testing!  This helps out A LOT!

js0505:
Also with the last commit, I cleaned up the naming as well.

gbargsley:
I am getting ? on CompressionTypeRecommendation.  Is that expected?

js0505:
For test yes, because that needs to be reviewed by the end user.  For Set I had to put NONE for now.  Unless we want to come up with better math within the SQL script.

gbargsley:
So page and row are even in savings so you put a ?.

js0505:
Yea, do we want to do a default of page but make it a parameter passed so the users can have an option to switch to row?

potatoqualitee:
hey jason
i copied the new code and continue to see a sea of red

image

I believe this is caused by the -ErrorRecord attached to the Write-Message (error record is only stop-function)

js0505:
Checking

potatoqualitee:
the thing that's weird, though, is that those databases uses to process the sql just fine - not sure what changed 

Also, it should detect whatever is causing those failures before they pass thru and just say "database skipped cuz whatever"

js0505:
Okay, have you closed your entire session and relaunched fresh? I was able to reproduce it on mine.  I left a code session open but was testing everything in PS.  I go to run a long test in code and I got the failures as well.  I closed all my sessions in code and PS and opened up code and I was no longer having the issue.

gbargsley:
I am not sure if I am understanding the output.  Is this saying Row Estimate is 51 percent but you are recommending PAGE compression?

image

js0505:
gbargsley - Size of the _sample_ with the requested compression setting. This includes any fragmentation.

FriedrichWeinmann:
Found out where the odd errors are coming from:
It's an issue in how Write-Message handles error messages. Fix incoming

functions/Set-DbaCompression.ps1
```diff
@@ -0,0 +1,508 @@
...
+	.EXAMPLE
+		Set-DbaCompression -SqlInstance ServerA -Database DBName -MaxRunTime 60 -PercentCompression 25 | Out-GridView
+		Set the compression run time to 60 minutes and will start the compression of of tables/indexes
+        that have a difference of 25% or higher between current and recommended and the results into and nicely formated GridView.
+	
+	.EXAMPLE
+		Set-DbaCompression -SqlInstance ServerA -MaxRunTime 60 -PercentCompression 25
+		Set the compression run time to 60 minutes and will start the compression of of tables/indexes; across all databases;
+        that have a difference of 25% or higher between current and recommended.
+
```
wsmelton: Need to add an example using the `-IncludeSystemDbs`, `-ExcludeDatabase` and `-SqlCredential` (just for completeness).

functions/Set-DbaCompression.ps1
```diff
@@ -0,0 +1,508 @@
...
+#>
+	[CmdletBinding(DefaultParameterSetName = "Default")]
+	param (
+        [parameter(Mandatory = $true, ValueFromPipeline = $true)]
+		[Alias("ServerInstance", "SqlServer")]
+		[DbaInstanceParameter[]]
+		$SqlInstance,
+		
+		[System.Management.Automation.PSCredential]
```
wsmelton: We are setting this to just `[PSCredential]` to prevent issues on Windows 7 ( think it was).

functions/Set-DbaCompression.ps1
```diff
@@ -0,0 +1,508 @@
...
+		$SqlCredential,
+		
+		[Alias("Databases")]
```
wsmelton: No reason to use this alias on new commands.

functions/Set-DbaCompression.ps1
```diff
@@ -0,0 +1,508 @@
...
+		$Database,
+		
+		[object[]]
+		$ExcludeDatabase,
+		
+		[switch]
+		$IncludeSystemDBs,
+		
+		[switch]
+		$Silent,
```
wsmelton: This needs to be the last parameter in the list.

functions/Set-DbaCompression.ps1
```diff
@@ -0,0 +1,508 @@
...
+				Write-Message -Level VeryVerbose -Message "Connecting to $instance" -Target $instance
+				$server = Connect-SqlInstance -SqlInstance $instance -SqlCredential $SourceSqlCredential -MinimumVersion 10
+			}
+			catch {
+				Stop-Function -Message "Failed to process Instance $Instance" -ErrorRecord $_ -Target $instance -Continue
+			}
+			
+            $Server.ConnectionContext.StatementTimeout = 0
+                
+                if ($Server.EngineEdition -eq 'Standard' -and $Server.VersionMajor -lt '13')
```
wsmelton: Would you mind adding a comment to just explain above this line why you validate in the connection for version 10 but then check for less than 13 here? Since this is due to MS changes in Standard Edition I'm assuming, worth just noting that to avoid confusion and make sure it is not removed.

functions/Set-DbaCompression.ps1
```diff
@@ -0,0 +1,508 @@
...
+                    Stop-Function -Message "Only SQL Server Enterprise Edition supports compression on $Server" -Target $Server -Continue
+                    }
+			#If IncludeSystemDBs is true, include systemdbs
+			#look at all databases, online/offline/accessible/inaccessible and tell user if a db can't be queried.
+			try {
+				if ($Database) {
+					$dbs = $server.Databases | Where-Object Name -In $Database
+				}
+				elseif ($IncludeSystemDBs) {
+					$dbs = $server.Databases | Where-Object Status -eq 'Normal'
```
wsmelton: This will end up including databases that are on a secondary replica (AG or mirroring) but not readable. It is best to check for `IsAccessible` as that guarantees you can query the database, the status does not.

wsmelton: `-eq` should also be `-EQ` because it is a parameter now and not just a comparison operator.

functions/Set-DbaCompression.ps1
```diff
@@ -0,0 +1,508 @@
...
+                
+                if ($Server.EngineEdition -eq 'Standard' -and $Server.VersionMajor -lt '13')
+                    {
+                    Stop-Function -Message "Only SQL Server Enterprise Edition supports compression on $Server" -Target $Server -Continue
+                    }
+			#If IncludeSystemDBs is true, include systemdbs
+			#look at all databases, online/offline/accessible/inaccessible and tell user if a db can't be queried.
+			try {
+				if ($Database) {
+					$dbs = $server.Databases | Where-Object Name -In $Database
```
wsmelton: Most of the commands, not sure we have updated the templates to show this, collect all the databases in `$dbs` and then filter down so we don't continuously go and try to pull the list of databases on an instance. System that I support has 1300 databases, just returning that list does take a few seconds.

So it would best to change this section to something like:
```powershell
$dbs = $server.Databases
if ($Database) {
 $dbs = $dbs | where-object...
}
if ($ExcludeDatabase) {
 $dbs = $dbs | where-object...
}
...
```

functions/Set-DbaCompression.ps1
```diff
@@ -0,0 +1,508 @@
...
+			}
+			catch {
+				Stop-Function -Message "Unable to gather list of databases for $instance" -Target $instance -ErrorRecord $_ -Continue
+			}
+			
+
+			foreach ($db in $dbs) {
+				try {
+					Write-Message -Level Verbose -Message "Querying $instance - $db"
+					If ($db.status -ne 'Normal' -or $db.IsAccessible -eq $false)
```
wsmelton: Little bit of a thang...but can you lowercase the `if`. ⛏

functions/Set-DbaCompression.ps1
```diff
@@ -0,0 +1,500 @@
...
+        Remember Uptime is critical, the longer uptime, the more accurate the analysis is.
+        You would probably be best if you utilized Get-DbaUptime first, before running this command.
+		
+		Set-DbaCompression script derived from GitHub and the tigertoolbox 
+        (https://github.com/Microsoft/tigertoolbox/tree/master/Evaluate-Compression-Gains)
+	
+	.PARAMETER SqlInstance
+		SqlInstance name or SMO object representing the SQL Server to connect to. This can be a collection and recieve pipeline input
+	
+	.PARAMETER SqlCredential
```
js0505: Fixed SqlInstance and SqlCredential

functions/Set-DbaCompression.ps1
js0505: Done

functions/Set-DbaCompression.ps1
```diff
@@ -0,0 +1,500 @@
...
+		PSCredential object to connect under. If not specified, current Windows login will be used.
+	
+	.PARAMETER Database
+		The database(s) to process - this list is autopopulated from the server. If unspecified, all databases will be processed.
+	
+	.PARAMETER ExcludeDatabase
+		The database(s) to exclude - this list is autopopulated from the server
+	
+	.PARAMETER IncludeSystemDBs
+		Switch parameter that when used will display system database information
```
js0505: Done

functions/Set-DbaCompression.ps1
js0505: I was able to have a successful test with [String], is that right?

functions/Set-DbaCompression.ps1
js0505: updated both command to not utilize Alias

functions/Set-DbaCompression.ps1
js0505: Added comment!  Woop! Will push tonight or tomorrow morning.

gbargsley:
I have run it on a couple systems.  I am not sure I understand the formula for what to compress.  It takes a really long time on one of my databases.

functions/Set-DbaCompression.ps1
js0505: Updated

functions/Set-DbaCompression.ps1
js0505: removed systemdb reference as that is invalid in this command

functions/Set-DbaCompression.ps1
js0505: done

js0505:
gbargsley: 
Column Percent_Update shows the percentage of update operations on a specific table, index, or partition, relative to total operations on that object. The lower the percentage of Updates (that is, the table, index, or partition is infrequently updated), the better candidate it is for page compression.

Column Percent_Scan shows the percentage of scan operations on a table, index, or partition, relative to total operations on that object. The higher the value of Scan (that is, the table, index, or partition is mostly scanned), the better candidate it is for page compression.

Column Compression_Type_Recommendation can have four possible outputs indicating where there is most gain, if any: ‘PAGE’, ‘ROW’, ‘NO_GAIN’ or ‘?’. When the output is ‘?’ this approach could not give a recommendation, so as a rule of thumb I would lean to ROW if the object suffers mainly UPDATES, or PAGE if mainly INSERTS, but this is where knowing your workload is essential. When the output is ‘NO_GAIN’ well, that means that according to sp_estimate_data_compression_savings no space gains will be attained when compressing, as in the above output example, where compressing would grow the affected object.

Note: Note that this script will execute on the context of the current database. Also be aware that this may take awhile to execute on large objects, because if the IS locks taken by the        sp_estimate_data_compression_savings cannot be honored, the SP will be blocked.

gbargsley:
Sounds good @js0505.  I will do some more tests today.  I just get antsy when commands run for a long time and it shows nothing on the screen.

potatoqualitee:
Test-DbaCompression looks fantastic on my side, @js0505 ! So long as @gbargsley can validate the results, the behavior is great on my side.

gbargsley:
I just want to make sure the output is as expected by the logic.  For instance ServerConfig table has high scan and low update.  Would that not be a candidate for PAGE instead of ?


image

js0505:
@gbargsley correct.

potatoqualitee:
hey @js0505 - so sorry but cant you please rename the files and the functions? To DbaDbCompression 

thank you 🙇

js0505:
Sure NP

gbargsley:
Would it be beneficial to add logic or another function to reverse what you just did if you see issues on performance?

potatoqualitee:
Unfortunately, this PR has two commands - @gbargsley - can you also approve Set?

gbargsley:
My test was on both.  How do I approve by file?

potatoqualitee:
Oh good 👍  I didn't realize. Will add now and update names.

PR #1899, opened 2017-07-26
[new command] Get-DbaOperatingSystem

wsmelton:
## Type of Change

 - [x] Pester test is included

### Purpose
 
Addressed part of the discussion from #837 
A need to pull common Operating System information, similar when doing an inventory or server review.
### Approach

A command that will pull the most common (at least initial version) properties of the Operating System class `Win32_OperatingSystem`. I did break the rule a bit and include the active power plan in this command, which is a different class...if that is not allowed we can remove it for a command specific to that class.
### Commands to test

```powershell
# 1
Get-DbaOperatingSystem -ComputerName sql2016, sql2016c,sql2016a

# 2
Get-DbaOperatingSystem -ComputerName sql2016 | select *

# 3 - Accepts piping
'sql2016' | Get-DbaOperatingSystem | select *
```
### Screenshots

(1)

image

(2)

image

### Learning

I tried to make the command return the full object you get from `Get-DbaCmObject` but that overwrites the use of `Select-DefaultView` within the command. The default view of `Win32_OperatingSystem` was the only thing to output when I tried. So I had to revert to building a custom object.

In the end, we can always expand this to additional properties if folks feel a need.

### DIFF
--- a/functions/Get-DbaOperatingSystem.ps1
+++ b/functions/Get-DbaOperatingSystem.ps1
@@ -1,0 +1,71 @@
+function Get-DbaOperatingSystem {
+	<#
+		.SYNOPSIS
+			Gets operating system information from the server.
+
+		.DESCRIPTION
+			Gets operating system information from the server and returns as an object.
+
+		.PARAMETER ComputerName
+			Target computer(s). If no computer name is specified, the local computer is targeted
+
+		.PARAMETER Credential
+			Alternate credential object to use for accessing the target computer(s).
+
+		.PARAMETER Silent
+			Use this switch to disable any kind of verbose messages
+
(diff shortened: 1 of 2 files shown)


PowerDBAKlaas:
strange, in the remake of `Get-DbaServerProtocol` `Get-DbaClientProtocol` and `Get-DbaSqlService` I did the same, and the combination of `Get-DbaCmObject` and `Select-Defaultview` seems to work fine.
Or are there different versions?

potatoqualitee:
beautiful screenshot, beautiful test! Will merge and patch as necessary - appveyor saying that it failed the test but it failed cuz it couldn't build something else.

PR #1901, opened 2017-07-26
Get-DbaOperatingSystem updates

potatoqualitee:
@wsmelton - LOVE this command! It's immediately useful to me. I made a couple small changes including translating Language to human readable and showing more columns by default.

### DIFF
--- a/functions/Get-DbaOperatingSystem.ps1
+++ b/functions/Get-DbaOperatingSystem.ps1
@@ -47,7 +47,7 @@ function Get-DbaOperatingSystem {
 	process {
 		foreach ($computer in $ComputerName) {
 			Write-Message -Level Verbose -Message "Attempting to connect to $computer"
-			$server = Resolve-DbaNetworkName -ComputerName $computer -Credential $Credential
+			$server = Resolve-DbaNetworkName -ComputerName $computer.ComputerName -Credential $Credential
 
 			$computerResolved = $server.ComputerName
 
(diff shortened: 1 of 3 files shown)


functions/Get-DbaOperatingSystem.ps1
```diff
@@ -68,37 +68,36 @@ function Get-DbaOperatingSystem {
...
-				Manufacturer = $os.Manufacturer
-				Organization = $os.Organization
-				Architecture = $os.OSArchitecture
-				Version = $os.Version
-				Build = $os.BuildNumber
-				InstallDate = [DbaDateTime]$os.InstallDate
-				LastBootTime = [DbaDateTime]$os.LastBootUpTime
-				LocalDateTime = [DbaDateTime]$os.LocalDateTime
-				TimeZone = $tz.Caption
+			$language = Get-Language $os.OSLanguage
```
wsmelton: I don't see the purpose in this one but your the boss.

Most folks know what the language id should be for their environment.

As well we could have used `Get-Culture` which is builtin cmdlet for PowerShell 1.0 and up to grab the same information from the machine. Although don't think it includes remote support so would have had to use `Invoke-Command`, but either way you get the same info and the correct OS level (and doesn't require another command to keep up with internally :) [Which `us_english` is SQL Server's spelled out version, the OS uses `en-US` format]

image

functions/Get-DbaOperatingSystem.ps1
```diff
@@ -68,37 +68,36 @@ function Get-DbaOperatingSystem {
...
-		} #end foreach instance
+				Language	   = $language.Name
+				LanguageAlias  = $language.Alias
+				LanguageId	   = $language.ID
+				CodeSet	       = $os.CodeSet
+				CountryCode    = $os.CountryCode
+				Locale		   = $os.Locale
+			} | Select-DefaultView -ExcludeProperty Codeset, CountryCode, Locale, LanguageID, LanguageAlias
+			
+			#$defaults = 'ComputerName','InstanceName','SqlInstance','Manufacturer','Architecture','Build','Version','InstallDate','LastBootTime','LocalDateTime'
```
wsmelton: Don't leave code commented out....leaves a bad taste :)

potatoqualitee: Wanted your approval first ;)

PR #1907, opened 2017-07-26
add startup info to Get-DbaSqlModule

potatoqualitee:


potatoqualitee:
tested and works on other systems

PR #1908, opened 2017-07-26
Install-DbaMaintenanceSolution

cviorel:
Download and Install SQL Server Maintenance Solution created by Ola Hallengren

## Type of Change

 - [ ] Pester test is included

### DIFF
--- a/functions/Install-SQLMaintenanceSolution_v2.ps1
+++ b/functions/Install-SQLMaintenanceSolution_v2.ps1
@@ -44,12 +44,21 @@ function Install-DbaMaintenanceSolution {
         http://dbatools.io/Install-DbaMaintenanceSolution
 	
 	.EXAMPLE
-        This will create the Ola Hallengren's Solution objects. Existing objects are not affected in any way.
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Thanks so much @cviorel! I shall accept and patch. It'll pretty much just need a rename and I'll add the required help from another command 👍 Hope to see more from you!

niphlod:
reminder: as discussed in the channel, before merging, it'd be good to have it match other install-* cmdlets and having a param to just install the base procedures instead of the whole solution, as there will be many that tuned ola's default jobs to match each specific need.

cviorel:
@niphlod, that's a good point, I will add this logic to the script.
Thanks!

potatoqualitee:
@cviorel - please add the following validated parameter set which allows the user to pass more than one solution

```
[ValidateSet('Backup', 'IntegrityCheck','IndexOptimize')]
[string[]]$Solution,
```

then
```
if (!$Solution) { 
   $Solution = 'Backup', 'IntegrityCheck','IndexOptimize'
}
```
then `foreach ($part in $solution) { if $part -eq 'Backup' ..... }`

Merci! 🍺

potatoqualitee:
Gonna merge and create tests! Thanks so much!

PR #1910, opened 2017-07-26
Write-Message - Fixed bug parsing error messages

FriedrichWeinmann:
Fixes a bug where Write-Message would generate an exception due to inability to use the `-NotLike` operator.
Basically it'd choke on all exception texts that included a `[`-bracket without closing it. A `-` inbetween the brackets would also prevent this detection (e.g.: `[a-b]`).


potatoqualitee:
thanks so much for fixing this 💯  and thank you for the approval, @wsmelton !

PR #1933, opened 2017-07-28
Get-DbaAvailabilityGroup Change IsPrimary comparison order

### LINKED ISSUE #1932: Get-DbaAvailabilityGroup IsPrimary Always Returns False

### Version Information
 - Operating System (Name|Version): Windows Server 2016 Standard
 - PowerShell Version: 5.1.14393.1198
 - SQL Server (Edition|Version): 2016 Enterprise

## Steps to Reproduce
`Get-DbaAvailabilityGroup -sqlinstance SERVERNAME -AvailabilityGroup AGNAME -IsPrimary | select IsPrimary`

 - [x] Attach output from PowerShell console (if possible/allowed)
IsPrimary = False

### Problem to solve
IsPrimary always returns false.

awickham10:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #1932)
 - [ ] Pester test is included

### Purpose
 
IsPrimary comparison was not working (always returning false). Switched order to force PowerShell to use implicit comparison function in DbaInstanceParameter class.

### Commands to test

`Get-DbaAvailabilityGroup -sqlinstance SERVERNAME -AvailabilityGroup AGNAME -IsPrimary | select IsPrimary`

### DIFF
--- a/functions/Get-DbaAvailabilityGroup.ps1
+++ b/functions/Get-DbaAvailabilityGroup.ps1
@@ -90,7 +90,7 @@ function Get-DbaAvailabilityGroup {
 				if ($IsPrimary) {
 					$defaults = 'ComputerName','InstanceName','SqlInstance','Name as AvailabilityGroup','IsPrimary'
 					$value = $false
-					if ($serverName -eq $ag.PrimaryReplicaServerName) {
+					if ($ag.PrimaryReplicaServerName -eq $serverName) {
 						$value = $true
 					}
 					Add-Member -Force -InputObject $ag -MemberType NoteProperty -Name IsPrimary -Value $value


potatoqualitee:
Awesome! Thanks @awickham10! The test failed but it's something we're working on - not your code.

@wsmelton can you give this a look? It's not something I can write a pester test for.

functions/Get-DbaAvailabilityGroup.ps1
```diff
@@ -90,7 +90,7 @@ function Get-DbaAvailabilityGroup {
 				if ($IsPrimary) {
 					$defaults = 'ComputerName','InstanceName','SqlInstance','Name as AvailabilityGroup','IsPrimary'
 					$value = $false
-					if ($serverName -eq $ag.PrimaryReplicaServerName) {
+					if ($ag.PrimaryReplicaServerName -eq $serverName) {
```
wsmelton: This makes no difference which side of the operator you put the values to be compared.

Do you have screenshots of it failing before this change and working after?

wsmelton: Just for reference this command when it was rewritten was tested against SQL Server 2016 and returned information correctly.

awickham10: The side of the operator does make a difference when dealing with implicit casts. The data type of the left side will be used to make the comparison. If you have $serverName first it will try to compare a DbaInstanceParameter object to a string, which will not match. If you put $ag.PrimaryReplicaServerName first it will cast DbaInstanceParameter to a string, which will invoke the class' implicit operator string function and return just the server name instead of the entire DbaInstanceParameter object.

Take a look at this screen shot for an example of how the order matters when dealing with objects of different types. Unfortunately I am on vacation now and don't readily have access to my server where I was having this issue. 

image

wsmelton: image

wsmelton: Yeah, @awickham10 edited my initial comment but we should actually be using `$server.Name` as the comparison and not the string passed into the command. Good catch on this one.

potatoqualitee: Thanks @awickham10 and @wsmelton! Are we ready to merge?

PR #1947, opened 2017-07-29
Fixing Snapshot tests

potatoqualitee:
Silent just allows Exceptions to be thrown but I don't see any try
catches. This param will soon be renamed to AllowException /cc @niphlod


potatoqualitee:
awesome, i was able to repro errors locally - will now fix

image

niphlod:
hum, ok. -Silent just remained in because I was trying to see how pester handles exceptions on the callbacks.

niphlod:
they now pass (for whatever reason) and the incorrect failure due to setup is signalled.

potatoqualitee:
hell yes!!!!! thank you!

PR #1950, opened 2017-07-30
Quick fix on compare replica name to server name

wsmelton:
## Type of Change

 - [ ] Breaking change (affects multiple commands or functionality)
 - [ ] Pester test is included

### Purpose
 
Correct the comparison to identify the `IsPrimary`. Originally it compared the replica name to the Instance name passed into the command. There is a potential the name passed in could be a DNS alias which would not match the replica name.
### Approach

Ensures that we are matching the actual server name to the replica name.


potatoqualitee:
thank you both!

PR #1951, opened 2017-07-30
Copy-DbaAgentJob - Set logic to Force owner to sa if login is missing on destination

### LINKED ISSUE #204: Copy-SqlAgentJob dealing with jobs where owner does not exist on destination
When copying jobs maybe add a switch (or include it in force) that if the owner of a job does not exist the job will simply be assigned to the "sa" account.

Job ownership is one of those things in best practice that jobs shouldn't be owned by users as those accounts are most commonly dropped if the person leaves and such. Situations where SSRS are involved is a special case that those jobs should be owned by the SSRS service account. [Although moving of SSRS jobs should probably be skipped. SSRS migration is a process all byitself.]

wsmelton:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #204 )
 - [ ] Pester test is included

### Purpose
 
Address issue where owner login of source job does not exist on destination instance. Previous we skipped those jobs, now it will set the job owner to "sa" of the destination server, if `-Force` is provided.

### Commands to test

1. Create job on your source instance with owner of that job set to a login that does not exist on your destination
2. Copy the job using `Copy-DbaAgentJob`

### Screenshots
image

image

image

### DIFF
--- a/functions/Copy-DbaAgentJob.ps1
+++ b/functions/Copy-DbaAgentJob.ps1
@@ -167,12 +167,14 @@
 			$missingLogin = $serverJob.OwnerLoginName | Where-Object { $destServer.Logins.Name -notcontains $_ }
 
 			if ($missingLogin.Count -gt 0) {
-				$missingLogin = ($missingLogin | Sort-Object | Get-Unique) -join ", "
-				$copyJobStatus.Status = "Skipped"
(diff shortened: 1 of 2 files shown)


potatoqualitee:
great enhancement!

PR #1962, opened 2017-07-31
Add AWS Support for Get-DbaComputerSystem

wsmelton:
## Type of Change

 - [ ] Pester test is included

### Purpose
 
In environments that happen to be in AWS Infrastructure As A Service (IaaS) there is additional information, or meta data, that you can pull on the EC2 instance itself. This is available via http://169.254.169.254/latest/meta-data

image

So in cases where you are pulling information and it is an EC2 instance you can pass the `-IncludeAws`. I had thought of doing a `IsAws` but didn't seem right when we can't do that for all of the providers. Azure does not offer this meta-data right now from within the VM itself, only AWS.

### Approach

Adding this information will allow the DBA to know what type of EC2 instance they have. Knowing this can make a difference in troubleshooting certain performance issues with SQL Server.

### Commands to test

Knowing that our testers may not have access to AWS instances I'm including screenshots of the information below.

### Screenshots

EC2 instance as seen from AWS Console

image

_The server sql2014-01 is a remote machine. I am calling this from a utility server also hosted in AWS that has network access to the environment._
`Get-DbaComputerSystem -ComputerName sql2014-01 -IncludeAws`

image


potatoqualitee:
Very cool, thanks Shawn!

Would it be possible for you to start using the official repo's branches? It would make it easier to test. If you aren't comfortable, nbd.

PR #1984, opened 2017-08-02
Fix Get-DbaRegisteredServerName Group issue

### LINKED ISSUE #1975: Get-SQLRegisteredServerName and Groups again
Get-SQLRegisteredServerName not accepting groups again. well, it does but it doesnt work

wsmelton:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #1975 )
 - [ ] Pester test is included

### Purpose
 
Fixing a bug when passing in a group to the command. The dynamic part of the parameter is not working for me anymore but it should accept the value now.
### Approach

Found a few issues where we were checking for an empty string/whitespace in an inaccurate method. Adjusted that to properly validate if the string was null, empty, or whitespaced.

As well added proper param block to the internal command so it is more clear parameters being passed to it, and corrected spelling error on parameter name (`$Bbase` to `$Base`).

### Commands to test

`Get-DbaRegisteredServerName -SqlInstance Server -Group SomeGroup`

`Get-DbaRegisteredServerName -SqlInstance Server`
### Screenshots

Passing `-Group`

image

No `-Group`

image

### Learning

Issue was being hit at line 99 with validating two variables that contained the group being passed in. Line 93 was also wrong so updated it to validate the `$Base`.

image

A bit of research and chatting with Klaas. Came across [proper way to check for empty string](https://blogs.technet.microsoft.com/heyscriptingguy/2015/10/07/dude-a-string-is-a-string-in-powershell/) and updated line 93 to be `$Base -eq $null -or [string]::IsNullOrWhiteSpace($Base)`

### DIFF
--- a/functions/Get-DbaRegisteredServerName.ps1
+++ b/functions/Get-DbaRegisteredServerName.ps1
@@ -130,6 +130,10 @@ function Get-DbaRegisteredServerName {
 			if ($Group -ne $null) {
 				foreach ($currentGroup in $Group) {
 					$cms = Find-CmsGroup -CmsGrp $cmsStore.DatabaseEngineServerGroup.ServerGroups -Stopat $currentGroup
+					if ($cms -eq $null) {
+						Write-Message -Level Output -Message "No groups found matching that name"
+						continue
+					}
 					$servers += ($cms.GetDescendantRegisteredServers()).ServerName
 				}
 			}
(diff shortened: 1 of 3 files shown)


gbargsley:
Testing now.  Your two tests work, but I get errors trying to get to a group within a group.

image

wsmelton:
Ok... @gbargsley 

- I've added output to better inform that it didn't find anything trying to use just the sub-group name. Right now it will only support the full path being provided to that sub-group. I can't get the logic right at the moment and would have to work through it with someone that has more smarts 🔢 

- I've also updated help to include an example of looking just for servers under a sub-folder.

- Also added test to validate output of command. It will create the following structure:

image

potatoqualitee:
Thanks yall 👍 !! Love seeing all these PRs. Will resolve all tomorrow.

PR #1987, opened 2017-08-02
Tests/fix backupdbadatabase

niphlod:
dir is part of lab setup so better to leave it

this is until we have a proper "lab initialization script"


potatoqualitee:
sounds good, passes checks 🏁

PR #1991, opened 2017-08-02
get-dbadatabase, tests fixes

niphlod:
master can be in full recovery model. a better test is for tempdb


Zhunya:
- [x] Tested (New/Restore/Remove) on database with one data file (SQL Server 2016)
- [x] Tested (New/Restore/Remove) on database with 5 data files (SQL Server 2012)
- [x] TEPP work good with this version (with old, it was not working for -Snapshot)

PR #1994, opened 2017-08-02
fix for snapshot functions, plus tests

niphlod:
## Type of Change

 - [x] Pester test is included

### Purpose
making new-dbadatabasesnapshot and remove-dbadatabasesnapshot more resilient.

### Approach
new-dbadatabasesnapshot failed on ultra-speedy instances (when snap creation was sub-second). Now it includes milliseconds, too. 
Also, filename and paths diverged a bit when -Path was passed or not. That logic has been streamlined.
remove-dbadatabasesnapshot relied on the $db object to be present after dropping, but it was not always the case when the snapshot removal took a few seconds. The SnapshotOf property gets now filled before the actual drop.

----
pester tests are included/updated accordingly


potatoqualitee:
Thanks to you both!! 🏆 🏆🏆

PR #1996, opened 2017-08-03
Fix the casting and alter error

sanderstad:


sanderstad:
Results after fix:
image

gbargsley:
Here is a before and after run.  What should I see different?

image

sanderstad:
The force parameter should now remove existing job steps. 
There were some other errors that I fixed in the first commit of the PR

potatoqualitee:
@gbargsley - tearin it up! Thanks so much, man. And thank you, @sanderstad for the quick fix! 🏆 🏆

PR #2013, opened 2017-08-04
Get-DbaFile Test - makes test universal

potatoqualitee:


jpomfret:
Looks good to me, tested against a couple of instances

image

potatoqualitee:
omg appveyor is totally lying, im gonna have to do some more tests this is weird

niphlod:
@potatoqualitee : you create "dbatoolsci_getfile$random" ... search for that file mask only (rather than dbatoolsci), probably other tests leave something behind.

potatoqualitee:
i did - in this case, it was returning some files instead of true/false. i guess it's evaluated differently for som reason? i see why but wished it'da failed locally

PR #2015, opened 2017-08-04
Remove-DbaFromDatabaseSnapshot, fixes

niphlod:
## Type of Change
 - [x] Pester test is included

### Purpose
fix original size log issue, be sure queries are issued in the correct context

### Approach
- added Snapshot property (to recognize what snapshot has been picked
up)
- corrected some queries
- fixed bug in restoring original log sizes
- removed an empty return value record
- nicer error message
- added tests


potatoqualitee:
tests work! looks great. 

Is that how I do this? 😂

PR #2022, opened 2017-08-05
Get-DbaDatabaseState and Set-DbaDatabaseState, complete revamp

niphlod:
## Type of Change
 - [x] Pester test is included

### Purpose
Adhere to codestyle, flowcontrol and messaging standards. Added tests

### DIFF
--- a/functions/Get-DbaDatabaseState.ps1
+++ b/functions/Get-DbaDatabaseState.ps1
@@ -1,5 +1,6 @@
-Function Get-DbaDatabaseState {
-    <#
+#ValidationTags#Messaging,FlowControl,Pipeline#
+function Get-DbaDatabaseState {
+	<#
 .SYNOPSIS
 Gets various options for databases, hereby called "states"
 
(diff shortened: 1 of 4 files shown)


potatoqualitee:
one test continues to fail :(

niphlod:
investigating... in the meantime, could you please inspect why enumerating $db.useraccess takes that long on appveyor ?

potatoqualitee:
I'm not even sure where to start. Upon researching, I came across a post that mentioned @dbaduck. Ben, would you have any idea why $db.useraccess is reasonably fast on our own machines, but insanely slow (about 2 minutes) on appveyor?

wsmelton:
Which insurance are you hiring in Appveyor? Does it make a difference on either one (faster, slower, same)?

potatoqualitee:
(I'm merging but we can still keep talking about it)

dbaduck:
I am not familiar with Appveyor. This property is a pretty easy property but may have issues when the database is offline, etc. but there should be no reason for it to be really slow when it is not always show. I try to detect things before by using IsAccessible before interrogating those properties.

niphlod:
@wsmelton both are slow. My first guess would be on the loaded SMO making a difference.
@dbaduck: it's always marginally slower with offline dbs but for whatever reason it takes a hell of a time there rather than in my (and it seems in other) test rigs.

Issue #1190, opened 2017-05-11
Start-SqlMigration login errors with Credentials


elguapo79:
Trying to migrate across domains, so I'm using a SQL user, with -SourceSqlCredential, and I can tell it validates because if I intentionally put the wrong PW in it tells me it failed, but when it tries to actually do anything, all it says is Login failed for user ''. The user is not associated with a trusted SQL Server Connection.  I double checked by running Connect-DbaSqlServer and that works fine also.

Attaching debug output as requested by Fred
[output.zip](https://github.com/sqlcollaborative/dbatools/files/993850/output.zip)

FriedrichWeinmann:
Hi,
sorry for getting back at you only after 3 days.
I checked the data you sent and can totally see the same things you described in the logs.
Much as I hate to admit it, it didn't make much sense to me from looking at the source - should have worked without a hitch.

When you connect using `Connect-DbaSqlServer`, do you use the default settings (that is: IP Address and Credentials) and let the function do the rest with default settings?

If that's it, I'll have to ask someone deeper on the DBA end of business look at the fine-print in the differences between the public function and what we use internally (not quite the same thing).

That said, you can use the connection object `Connect-DbaSqlServer` returns as input to `Start-SqlMigration`. In that case simply don't specify credentials, our commands will reuse that connection object to access the source server.

elguapo79:
Yes, that’s correct – using Connect-DbaSqlServer I just used the IP and credentials and it worked fine.

I tried as you mentioned below, still getting the same error.  Check out the screenshot.  It connects, but gives the error on the Migration, but then I can still use the object I created and query off it, so the connection is working, but for some reason the migration script doesn’t like it

[cid:image002.png@01D2CD60.E1F01D90]

Jason Allen

Zhunya:
Have same cross-domain authentication problems when trying to Copy-DbaDatabase/Restore-DbaDatabase. 

From my machine:
![windows powershell 2017-06-15 17 35 00](https://user-images.githubusercontent.com/14327077/27186812-1ecf5684-51f2-11e7-91ca-765952dade7c.png)

On remote host:
![snipping tool 2017-06-15 17 42 17](https://user-images.githubusercontent.com/14327077/27186842-345d1810-51f2-11e7-95ba-1cbfbafa45dc.png)

wsmelton:
[...]

PR #2023, opened 2017-08-06
Address issue where we connect using Windows Auth when Credential is provided

### LINKED ISSUE #2012: Test-DbaIdentityUsage attempts using current Windows account before specified -SqlCredential
## Bug Report
```
$cred = Get-Credential
Test-DbaIdentityUsage -SqlInstance MyInstance -SqlCredential $cred
```

I noticed that, when running PowerShell as a Windows user without login privileges to the destination SQL instance that I receive a login failure in the SQL errorlog even when specifying the `-SqlCredential` parameter. What is happening is that the command is trying the active Windows user running PowerShell first before trying what is specified via the `-SqlCredential` parameter. This would likely go unnoticed if the Windows account running PowerShell has permissions to run the cmdlet on the specified `-SqlInstance`. I noticed this due to the Windows account I was using _not_ having permissions to the SQL instance and automatic email notifications set up to alert whenever a failed login gets logged to the SQL errorlog. This has been confirmed with Profiler traces.

I have tested `Connect-SqlInstance` as suggested by @potatoqualitee and @wsmelton, but it does not suffer from the same problem.

```
#This works - does not first attempt to login with Windows account
$cred = Get-Credential
Connect-SqlInstance -SqlInstance MyInstance -SqlCredential $cred
```

Destination servers running Windows Server 2008 R2 and SQL Server 2008 R2 show this in the errorlog when ran from the context of a Windows user without SQL login permissions:
image

Destination servers running Windows Server 2012 R2 and SQL Server 2014 show this in the errorlog when ran from the context of a Windows user without SQL login permissions:
image

```
#This also works - does not first attempt to login with Windows account
$cred = Get-Credential
$srv = Connect-DbaSqlServer -Credential $cred -SqlInstance MyInstance
Test-DbaIdentityUsage -SqlInstance $srv
```

Connecting to the instance using `Connect-DbaSqlServer` first and then passing that into `Test-DbaIdentityUsage` works without the errorlog login failure when ran from the context of a user without SQL instance login permissions.

Here is a SQL Profiler trace screenshot showing the process when executed from the context of a user that does not have SQL login permissions:
image

see above

### Version Information

Source Operating System: Windows Server 2012 R2

dbatools version:
```
PS C:\Windows\system32> (Get-Module dbatools).version

Major  Minor  Build  Revision
-----  -----  -----  --------
0      9      22     -1      
```
Local SQL Server version: 2014
Remote SQL Server version: 2014

## Steps to Reproduce
```
$cred = Get-Credential
Test-DbaIdentityUsage -SqlInstance MyInstance -SqlCredential $cred
```

Also, running just `Test-Dba-IdentityUsage -SqlInstance MyInstance` gives the login failure directly to the screen when executed from the context of a user without SQL login permissions:

```
PS C:\Windows\system32> Test-DbaIdentityUsage -SqlInstance MyInstance
WARNING: [Test-DbaIdentityUsage][12:44:20] Failure | Can't connect to MyInstance: System.Data.SqlClient.SqlException (0x80131904): Login failed for user 'DOMAIN\username'.
```

### Problem to solve
Don't have the`Test-DbaIdentityUsage` cmdlet use the current PowerShell session's Windows credential if the `-SqlCredential` parameter is set.

wsmelton:
## Type of Change

 - [x] Bug fix ( non-breaking change, fixes #2012 )
 - [ ] Breaking change (affects multiple commands or functionality)
 - [ ] Pester test is included

### Purpose
 
It was found after debugging and stepping through `Connect-SqlServer` that when a credential object is passed in, or the `-MinimumVersion` parameter was used in a command that an attempt is made to connect to the given SQL Server instance. This allows for an unknown login attempt by the account running the PowerShell session/host.

The main issue where this is found is when the account running the PowerShell host does not have access to the given SQL Server instance. The instance will log multiple login failed attempts just by executing a single command. If a credential (e.g. SQL Login) is provided it still attempts to login as the account running the PowerShell host. [See learning section below.]
### Approach

By provided a known bogus server name then no connection will be made until the `$server.ConnectionContext.Connect()` line of either command is executed.
### Commands to test

Appveyor test should run enough as the Pester test utilize `Connect-DbaSqlServer` and then the commands run in the integration test call `Connect-SqlInstance`.
### Screenshots

Will provide a blog post that details how this was identified in debugging.
### Learning
This was one of those "after I saw it I knew/recall why this happens".

The general reasoning is just executing this command in a PowerShell host will cause an attempted connection to the `localhost`:
```powershell
New-Object Microsoft.SqlServer.Management.Smo.Server
```

If the `localhost` does not have SQL Server running, then you will never notice anything. If the `localhost` has SQL Server running, but the account executing it does not have access then a login failed message will be logged, **but you will not see any error posted to the PowerShell host**.

So as we do this in our code (or close to it):
```powershell
$server = New-Object Microsoft.SqlServer.Management.Smo.Server $SqlInstance
```
This will try to immediately connect to the `$SqlInstance` as soon as that line is run. Then, as well, any subsequent line that accesses the `$server` object will cause additional login attempts to be performed.

The only workaround to this is to pass in a bogus server to the initial `New-Object` so it connects to nothing. That is the sole purpose in this `([System.Guid]::NewGuid())`. PowerShell 5 offers `New-Guid` but in order to support back to PowerShell 3 we have to use the `System.Guid` 's `NewGuid()` method.

### DIFF
--- a/functions/Connect-DbaSqlServer.ps1
+++ b/functions/Connect-DbaSqlServer.ps1
@@ -210,7 +210,7 @@ function Connect-DbaSqlServer {
             return $SqlInstance
         }
 		
-        $server = New-Object Microsoft.SqlServer.Management.Smo.Server ([guid]::NewGuid())
+        $server = New-Object Microsoft.SqlServer.Management.Smo.Server ([System.Guid]::NewGuid())
 		
         if ($AppendConnectionString) {
             $connstring = $server.ConnectionContext.ConnectionString
--- a/functions/Connect-DbaSqlServer.ps1
+++ b/functions/Connect-DbaSqlServer.ps1
@@ -1,4 +1,4 @@
-Function Connect-DbaSqlServer {
+function Connect-DbaSqlServer {
     <#
     .SYNOPSIS
         Creates an efficient SMO SQL Server object.
@@ -210,7 +210,7 @@ Function Connect-DbaSqlServer {
             return $SqlInstance
         }
 		
-        $server = New-Object Microsoft.SqlServer.Management.Smo.Server $SqlInstance
+        $server = New-Object Microsoft.SqlServer.Management.Smo.Server ([guid]::NewGuid())
 		
         if ($AppendConnectionString) {
             $connstring = $server.ConnectionContext.ConnectionString
@@ -268,7 +268,7 @@ Function Connect-DbaSqlServer {
                         $server.ConnectionContext.set_SecurePassword($Credential.Password)
                     }
                 }
-				
+				$server.ConnectionContext.ServerInstance = $SqlInstance
                 $server.ConnectionContext.Connect()
             }
             catch {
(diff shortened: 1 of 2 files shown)


wsmelton:
There is a potential this will also fix #1190

wsmelton:
All main test completed successfully. The failures are with the snapshot test.

PR #2024, opened 2017-08-06
Initial commit of New-DbaDatabase

matt-darwin:
## Type of Change

 - [ ] Pester test is included

### Purpose
 
Standardise database creation, and introduce best practise approach to seperate user data files from system files in the PRIMARY filegroup.

### Approach

Creates a new database with fixed sized file and log growth, allows specification of the number of files to create in the user filegroup, this is set to default filegroup.  Allows easy setting of collation, database owner and recovery model.

### Commands to test

New-DbaDatabase

### Screenshots

new-dbadatabase

### Learning

Adapted from my blog post function New-SQLDatabase:- https://naturalselectiondba.wordpress.com/2015/06/10/standardising-new-database-creation-using-powershell-and-smo/

### Todo

Test for recovery model being correctly passed in (i.e. simple, full or bulk logged).  Check to see if default primary file sizes are smaller than existing model file sizes which will prevent the db from being created; need to take the existing model size if so.  Input parameters need looking at, and descriptions creating in the header.

### DIFF
--- a/functions/New-DbaDatabase.ps1
+++ b/functions/New-DbaDatabase.ps1
@@ -1,0 +1,313 @@
+function New-DbaDatabase 
+{
+  <#
+.SYNOPSIS 
+New-DbaADatabase creates a new database
+
+.DESCRIPTION
+New-DbaDatabase creates a new database with a single user filegroup, and the PRIMARY filegroup reserved for system objects.
+It allows creation with multiple files, and sets all growth settings to be fixed size rather than percentage growth.
+
+.PARAMETER SqlInstance
+The SQL server instances to be connected to; can be passed in via the pipeline.
+
+.PARAMETER Force
+The force parameter will ignore some errors in the parameters and assume defaults.
+
+.PARAMETER WhatIf 
+Shows what would happen if the command were to run. No actions are actually performed. 
+
+.PARAMETER Confirm 
+Prompts you for confirmation before executing any changing operations within the command. 
+
+.PARAMETER Silent 
+Use this switch to disable any kind of verbose messages
+
+.NOTES 
+Original Author: Matthew Darwin (@evoDBA, naturalselectiondba.wordpress.com)
+Tags: Database
+	
+Website: https://dbatools.io
+Copyright: (C) Chrissy LeMaire, clemaire@gmail.com
+License: GNU GPL v3 https://opensource.org/licenses/GPL-3.0
+
+.LINK
+https://dbatools.io/New-DbaAgentJob
+
+.EXAMPLE   
+New-DbaAgentJob -SqlInstance sql1 -DatabaseName 'TestDatabase' -UserDataFileSize 128 -UserDataFileMaxSize 1024 -UserDataFileGrowth 128 `
+-LogSize 128 -LogGrowth 128
+Creates a database named TestDatabase on instance sql1 with a user filegroup with a single file of 128MB
+
+#>
+
+    [cmdletbinding(SupportsShouldProcess = $true, ConfirmImpact = "Low")]
+    Param 
+    ( 
+    # set variables for the database
+            [parameter(Mandatory = $true, ValueFromPipeline = $true)]
+		    [Alias("ServerInstance", "SqlServer")]
+		    [DbaInstanceParameter[]]$SqlInstance
+          , [PSCredential]$SqlCredential  
+          , [parameter(Mandatory = $true)]
+            [string]$DatabaseName                      
+          , [double]$NumberOfFilesInUserFilegroup = 1
+
+    # optional db variables
+          , [boolean]$UseDefaultFileLocations = $true
+          , [string]$NonDefaultFileLocation
+          , [string]$NonDefaultLogLocation
+          , [string]$Collation
+          , [string]$RecoveryModel
+          , [string]$DatabaseOwner = "sa"
+
+    #set the user data size, maxsize and growth
+          , [double]$UserDataFileSize 
+          , [double]$USerDataFileMaxSize
+          , [double]$UserDataFileGrowth #use 0 for no growth
+
+    #set the log size and growth
+          , [double]$LogSize
+          , [double]$LogGrowth
+
+    #set the primary file size in MB (will be converted to kb later)
+          , [double]$PrimaryFileSize = 10
+          , [double]$PrimaryFileGrowth = 10
+          , [double]$PrimaryFileMaxSize = 100
+          		
+    #Switches
+          , [switch]$Force
+		  , [switch]$Silent
+    )
+
+    begin
+    {
+        #Check file directories passed in if not using defaults
+        if ($UseDefaultFileLocations -eq $false -and ($NonDefaultFileLocation -eq $Null -or $NonDefaultLogLocation -eq $Null))
+            {
+                Stop-Function -Message "Non Default file locations selected, but are not supplied" -Category InvalidData -ErrorRecord $_ -Target $instance -Continue
+            }
+    }
+
+
+    process 
+    {
+        
+        if (Test-FunctionInterrupt) { return }
+
+        #Verbose message to show the server
+        Write-Message -message "Connecting to server $SqlInstance" -Level Verbose
+
+        #instantiate the sql server object
+	    try {
+		    $SQLServer = Connect-SqlInstance -SqlInstance $SqlInstance -SqlCredential $SqlCredential
+	    }
+	    catch {
+		    Stop-Function -Message "Failure" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
+	    }
+
+        #check to see if the database already exists.
+        if ($SQLServer.Databases[$DatabaseName].Name -ne $Null)
+        {
+            Stop-Function -Message "Database $DatabaseName already exists on $TargetServer" -Target $instance -Continue
+        }
+
+        # if we are using the default file locations, get them from the server
+        if ($UseDefaultFileLocations -eq $true)
+        {
+            #get the default file locations; if the master is in the default location we use that path as the default file will not be set
+            $LocalDataDrive = if ($SQLServer.DefaultFile -eq [DBNULL]::value)
+                                        {$SQLServer.MasterDBPath} 
+                                    else
+                                        {$SQLServer.DefaultFile}
+            $LocalLogDrive = $SQLServer.DefaultLog
+        }
+        elseif ($UseDefaultFileLocations -eq $false)
+        {
+            $LocalDataDrive = $NonDefaultFileLocation
+            $LocanewlLogDrive = $NonDefaultLogLocation
+        }
+        
+        #output message in verbose mode
+        Write-Message -message "Set local Data drive to $LocalDataDrive and local log drive to $LocalLogDrive" -level verbose
+
+        #create the new db object
+
+        try
+        {
+            write-message -message "Creating smo object for new database $DatabaseName" -level verbose
+            $NewDB = New-Object Microsoft.SqlServer.Management.Smo.Database($SQLServer, $DatabaseName)
+        }
+        catch
+        {
+            Stop-Function -Message "Error creating database object for $DatabaseName on server $Sqlserver" -ErrorRecord $_ -Target $instane -Continue
+        }
+
+        #add the primary filegroup and a primary file
+        try
+        {
+            write-message -message "Creating PRIMARY filegroup" -level Verbose
+            $PrimaryFG = new-object Microsoft.SqlServer.Management.Smo.Filegroup($NewDB, "PRIMARY")
+            $NewDB.Filegroups.Add($PrimaryFG)
+        }
+        catch
+        {
+            Stop-Function -Message "Error creating Primary filegroup object" -ErrorRecord $_ -Target $instance -Continue
+        }
+
+        #add the primary file
+        try
+        {
+            $PrimaryFileName = $DatabaseName + "_PRIMARY"
+            Write-Message -message "Creating file name $PrimaryFileName in filegroup PRIMARY" -level verbose
+
+            #create the filegroup object
+            $PrimaryFile = new-object Microsoft.SqlServer.Management.Smo.DataFile($PrimaryFG, $PrimaryFileName)
+            $PrimaryFile.FileName = $LocalDataDrive + "\" + $PrimaryFileName + ".mdf"
+            $PrimaryFile.Size = ($PrimaryFileSize * 1024)
+            $PrimaryFile.GrowthType = "KB"
+            $PrimaryFile.Growth = ($PrimaryFileGrowth * 1024)
+            $PrimaryFile.MaxSize = ($PrimaryFileMaxSize * 1024)
+            $PrimaryFile.IsPrimaryFile = "true"
+            #add the file to the filegroup
+            $PrimaryFG.Files.Add($PrimaryFile)
+        }
+        catch
+        {
+            Stop-Function -Message "Error adding file to Primary filegroup" -ErrorRecord $_ -Target $instance -Continue
+        }
+
+        #add the user data file group
+        try
+        {
+            $UserFilegroupName = $DatabaseName + "_MainData"
+            write-Message -message "Creating user filegroup $UserFileGroupName" -level Verbose
+            
+            $UserFG = new-object Microsoft.SqlServer.Management.Smo.Filegroup($NewDB, $UserFilegroupName)
+            $NewDB.Filegroups.Add($UserFG)
+        }
+        catch
+        {
+            Stop-Function -Message "Error creating user filegroup" -ErrorRecord $_ -Target $instance -Continue
+        }
+
+        #add the required number of files to the filegroup in a loop
+        #set the filecounter
+        $FileCounter = 1
+
+        #open a loop while the filecounter is less than the required number of files
+        While ($FileCounter -le $NumberOfFilesInUserFilegroup)
+        {
+            #Set the file name
+            try
+            {
+                $UserFileName = $UserFileGroupName + "_" + [string]$FileCounter
+                Write-Message -message "Creating file name $UserFileName in filegroup $UserFileGroupName" -Level Verbose
+                #create the smo object for the file
+                $UserFile = new-object Microsoft.SQLServer.Management.Smo.Datafile($UserFG, $UserFileName)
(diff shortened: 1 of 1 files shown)


matt-darwin:
Fixed missing loop for multiple instance support.

functions/New-DbaDatabase.ps1
```diff
@@ -0,0 +1,317 @@
...
+          , [switch]$Force
+		  , [switch]$Silent
+    )
+
+    begin
+    {
+        #Check file directories passed in if not using defaults
+        if ($UseDefaultFileLocations -eq $false -and ($NonDefaultFileLocation -eq $Null -or $NonDefaultLogLocation -eq $Null))
+            {
+                Stop-Function -Message "Non Default file locations selected, but are not supplied" -Category InvalidData -ErrorRecord $_ -Target $instance -Continue
```
FriedrichWeinmann: Slight error in using `Stop-Function`:
```PowerShell
# Replace this
Stop-Function -Message "Non Default file locations selected, but are not supplied" -Category InvalidData -ErrorRecord $_ -Target $instance -Continue

# With this
Stop-Function -Message "Non Default file locations selected, but are not supplied" -Category InvalidData -ErrorRecord $_
return
```
 - `Continue` should only be used in process within a loop, as it can otherwise kill the function or script calling it
 - `-Target` is to help with debugging, by containing the item currently being processed. You aren't processing an item yet (It's the begin block), so it shouldn't be there yet

functions/New-DbaDatabase.ps1
```diff
@@ -0,0 +1,317 @@
...
+
+            #Verbose message to show the server
+            Write-Message -message "Connecting to server $Instance" -Level Verbose
+
+            #instantiate the sql server object
+	        try {
+		        $SQLServer = Connect-SqlInstance -SqlInstance $Instance -SqlCredential $SqlCredential
+	        }
+	        catch {
+		        Stop-Function -Message "Failure" -Category ConnectionError -ErrorRecord $_ -Target $Instance -Continue
```
FriedrichWeinmann: Please include a better message. `Failure` doesn't tell the user a thing.
Keep in mind, by default, the user will not be seeing verbose messages, but _will_ see warnings by Stop-Function. So ... include a bit mroe about what just broke :)

functions/New-DbaDatabase.ps1
```diff
@@ -0,0 +1,317 @@
...
+                Stop-Function -Message "Error adding log file to database." -ErrorRecord $_ -Target $instance -Continue
+            }
+
+            #set database settings; collation, owner, recovery model
+
+            #set the collation
+            if ($Collation.Length -eq 0)
+            {
+                $Message = "USing default server collation"
+                Write-Verbose $Message
```
FriedrichWeinmann: `Write-Verbose` should not be used, instead use `Write-Message -Level Verbose`

functions/New-DbaDatabase.ps1
```diff
@@ -0,0 +1,317 @@
...
+            }
+            else
+            {
+                $Message = "Setting collation to $Collation"
+                Write-Verbose $Message
```
FriedrichWeinmann: `Write-Verbose` should not be used, instead use `Write-Message -Level Verbose`

functions/New-DbaDatabase.ps1
```diff
@@ -0,0 +1,317 @@
...
+                $NewDB.Collation = $Collation
+            }
+
+            #set the recovery model
+            if ($RecoveryModel.Length -eq 0)
+            {
+                $Message = "Using default recovery model from the Model database"
+                Write-Verbose $Message
```
FriedrichWeinmann: `Write-Verbose` should not be used, instead use `Write-Message -Level Verbose`

functions/New-DbaDatabase.ps1
```diff
@@ -0,0 +1,317 @@
...
+                }
+                catch
+                {
+                    Stop-Function -Message "Error setting default filegorup to $UserFileGroupName" -ErrorRecord $_ -Target $instance -Continue
+                }
+
+            }
+
+            #Write completed message
+            Write-Message -message "Completed creating database $DatabaseName on server $sqlInstance" -level Verbose
```
FriedrichWeinmann: Use `$instance` variable instead for the message content. `$sqlinstance` may be more than one instance

functions/New-DbaDatabase.ps1
```diff
@@ -0,0 +1,317 @@
...
+
+.PARAMETER WhatIf 
+Shows what would happen if the command were to run. No actions are actually performed. 
+
+.PARAMETER Confirm 
+Prompts you for confirmation before executing any changing operations within the command. 
+
+.PARAMETER Silent 
+Use this switch to disable any kind of verbose messages
+
```
FriedrichWeinmann: - Lots of parameters missing from this help
 - A minimal example would be nice ("What is the minimum information you _must_ pass in order for something to happen?")
 - An example showing pipeline usage would be great

functions/New-DbaDatabase.ps1
matt-darwin: Sorry that one was just copied direct from New-DbaAgentJob.  I was assuming the category ConnectionError and the message Failure were a standard.

functions/New-DbaDatabase.ps1
```diff
@@ -0,0 +1,469 @@
...
+            elseif ($UseDefaultFileLocations -eq $false)
+            {
+                $LocalDataDrive = $NonDefaultFileLocation
+                $LocanewlLogDrive = $NonDefaultLogLocation
+            }
+
+            #create the file locations if they do not already exist
+            try
+            {
+                if ((test-path -path $LocalDataDrive) -eq $false)
```
wsmelton: You will need to adjust this to support remote calls to the system. Or use some other method of creating the directories on the given instance. 

As is this will create the instance on the local machine I execute the command from.

functions/New-DbaDatabase.ps1
```diff
@@ -0,0 +1,469 @@
...
+    #set the primary file size in MB (will be converted to kb later)
+          , [double]$PrimaryFileSize = 10
+          , [double]$PrimaryFileGrowth = 10
+          , [double]$PrimaryFileMaxSize = 100
+          		
+    #Switches
+          , [switch]$Force
+		  , [switch]$Silent
+    )
+
```
wsmelton: Would be cool to provide a properties parameter that accepts a hash table of the properties that can be set, if folks want to override model database settings.

functions/New-DbaDatabase.ps1
```diff
@@ -0,0 +1,469 @@
...
+        {
+            Stop-Function -Message "PrimaryFilesize of $PrimaryFileSize is greater than the PrimaryFileMaxSize setting of $PrimaryFileMaxSize" -Category InvalidData `
+            -ErrorRecord $_
+        }
+
+        #Check that the user data file max size is greater than the user file growth
+        if ($PrimaryFileGrowth -gt $PrimaryFileGrowth)
+        {
+            Stop-Function -Message "PrimaryFileGrowth of $PrimaryFileGrowth is greater than the PrimaryFileMaxSize setting of $PrimaryFileMaxSize" -Category InvalidData `
+            -ErrorRecord $_
```
wsmelton: There won't be any error record to pass in these checks. So if you want the function to stop need to add a return for each block. Unless your test showed otherwise?

functions/New-DbaDatabase.ps1
wsmelton: They are the standard we have in all the commands currently. So it is fine to keep it that way for now.

functions/New-DbaDatabase.ps1
```diff
@@ -0,0 +1,469 @@
...
+          , [string]$NonDefaultFileLocation
+          , [string]$NonDefaultLogLocation
+          , [string]$Collation
+          , [string]$RecoveryModel
+          , [string]$DatabaseOwner = "sa"
+
+    #set the user data size, maxsize and growth
+          , [double]$UserDataFileSize 
+          , [double]$USerDataFileMaxSize
+          , [double]$UserDataFileGrowth #use 0 for no growth
```
matt-darwin: The reason I named these User is to differentiate between the Primary file sizes. Don't mind either way. For logs, as there's only one log doesn't really make a difference.

functions/New-DbaDatabase.ps1
matt-darwin: Good point, this was a recent addition that I've only tested locally. Will have a look at the best options for this.

functions/New-DbaDatabase.ps1
```diff
@@ -0,0 +1,469 @@
...
+True/False, if true will use the default file locations for data and log files on the SQL Server instance, if false, these are specified
+in the $NonDefaultFileLocation and $NonDefaultLogLocation parameters.
+
+.PARAMETER NonDefaultFileLocation
+The location that data files will be placed if UseDefaultFileLocations is set to false.
+
+.PARAMETER NonDefaultLogLocation
+The location the log file will be placed if UseDefaultFileLocations is set to False.
+ 
+.PARAMETER Collation
```
matt-darwin: This is available in SQL Server as a drop down, so must be exposed meta data somewhere, just will be a BIG list...

functions/New-DbaDatabase.ps1
```diff
@@ -0,0 +1,469 @@
...
+            [string]$DatabaseName                      
+          , [double]$NumberOfFilesInUserFilegroup = 1
+
+    # optional db variables
+          , [boolean]$UseDefaultFileLocations = $true
+          , [string]$NonDefaultFileLocation
+          , [string]$NonDefaultLogLocation
+          , [string]$Collation
+          , [string]$RecoveryModel
+          , [string]$DatabaseOwner = "sa"
```
matt-darwin: How would that work with multiple instances?

functions/New-DbaDatabase.ps1
matt-darwin: No, just copy and paste laziness happening! Will fix up. Anything else in the error handling worth tweaking as that's the majority of changes I've made for DBATools?

functions/New-DbaDatabase.ps1
wsmelton: It looks at the first one. If you have multiple instances you only allow one owner...so we have to assume that single login will exist on all of them. Should be validated as well, if the login provided is on the specific instance.

If it is not found I'd say default to "sa". However in that regard you need to use the internal command `Get-SaLoginName` as it handles returning the real name in case "sa" was renamed (like it should be).

functions/New-DbaDatabase.ps1
wsmelton: Why are we forcing folks to use multiple data files? This should be their option to have a primary file and secondary files. Whether it was the intention of a "best practice" or not, we cannot force a specific configuration on every user.

wsmelton: Consider in cases where I have to create a database per physical period or month for a finance application. I'm not worried about performance or restore times, just have a need to create a database that will likely never be over 1GB in size.

In larger cases where I'm creating an initial data warehouse database, then I'd prefer the ability to have a primary file/group and secondary.

wsmelton:
Outside of just creating the basic database with files, are we going to include the support of setting the properties of the database? Or just let the secondary commands perform those task (e.g. enabling QueryStore for database in 2016+)

potatoqualitee:
wow, nice work, thank you @matt-darwin! 🌯 🌮 🍔 

@wsmelton, thank you so much spending so much time on this one. I'll leave it in your hands - just let me know when it's approved. Also, @PowerDBAKlaas can you take a look at the functionality if you have time this week and ensure it aligns with your vision as well.

functions/New-DbaDatabase.ps1
matt-darwin: Was simply from best practise; personally I think the only downside of _not_ having user data in a seperate filegroup is that it's a pain to create the initial statement, regardless of the db size/performance considerations/recovery strategy. Please bear in mind that this function was ported from a function that was used to enforce a specific business set of standards, so adapting what was already written for this so may take a little time to sort some of this stuff out!

Happy to make changes, though, to make this optional, will most likely have to go down an object as input to specify the configuration for multiple file groups though (e.g. Filegroup, File, Location), which isn't necessarily a bad thing.  I have an item on my list to break out the file/filegroup section to a seperate child function in any case, so we could then use those in this function, and set up an object that specifies how many filegroups, how many files in those filegroups, and which drives they should sit on?  At the same time, I'm aware that a bit of the purpose of this is to try and return a complicated DB creation script into a simple one liner, so the object set up for that may need to be quite simple.

Has this comment turned into waffle? Yes... I think it has....

functions/New-DbaDatabase.ps1
matt-darwin: I'm wondering if this should actually be a seperate function? In T-SQL they would all be seperate Alter... Set commands after the creation.

functions/New-DbaDatabase.ps1
```diff
@@ -0,0 +1,469 @@
...
+$scred = Get-Credential, then pass $scred object to the -SqlCredential parameter. 
+To connect as a different Windows user, run PowerShell as that user.
+
+.PARAMETER DatabaseName
+The name of the new database to be created.
+
+.PARAMETER NumberOfFilesInUserFilegroup
+The number of files to create in the user filegroup for the database.  Default is 1.
+
+.PARAMETER UseDefaultFileLocations
```
PowerDBAKlaas: Isn't this parameter redundant? 
If no DataFileLocation or LogFileLocation is given, default has to be used.

PowerDBAKlaas:
@potatoqualitee , for a DBA looking for automation and standardization, having a set of defined settings may be the most interesting part, as @wsmelton  suggests. However, since there are a lot of properties, differing per sql version, dependent on one another, with hierarchical and chronological sequences to be respected, I feel this deserves a second function `Configure-DbaDatabase` or `Set-DbaDBConfiguration`.
In this `New-DbaDatabase` we can probably stick to 'unevitable' settings, like the files and the db Owner.
Thanks, @matt-darwin !

functions/New-DbaDatabase.ps1
matt-darwin: I think it's a statement of intent, really.

matt-darwin:
@PowerDBAKlaas I think I agree; I was reviewing the actual options for the CREATE DATABASE tsql statement (https://docs.microsoft.com/en-us/sql/t-sql/statements/create-database-sql-server-transact-sql) and there's not as much as I immediately think, some of those things are ALTER DATABASE .... SET statements that follow up.

There are a couple that probably belong in this function, and then perhaps we could make the call to whatever that function looks like optional at the end of this?

I'm also very much thinking that the filegroup section needs to be moved out into a child function, which would then allow new creation of filegroups in the future as well.  Not had much time to do any coding recently, but when I do I think this will be the initial focus alongside the further tweaks suggested.

potatoqualitee:
Thanks so much @PowerDBAKlaas, on point as always ❤️  

@matt-darwin New-DbaDbFileGroup? does sound great

matt-darwin:
@potatoqualitee Yeah pretty much, and probably New-DbaDataFile and New-DbaLogFile (to cover that moment when your log is full and your drive is full) as well.

The New-DbaDatabase function could then effectively be a wrapper that puts all the relative components together.

Got my work cut out.....

PowerDBAKlaas:
hmmm,
maybe instead of having a function for every setting ( there's also growth settings, set default filegroup,...),
an XML or JSON could be a convenient source to describe the desired DB

matt-darwin:
I was thinking that the file size/growth related settings would work as they do in the current function (there's really no reason to have multiple files in the same filegroup with different sizes), but set the filegroup function up so that you could either pass in a single parameter for location, or a hash table with each file/location.

So ultimately we'd end up with something like the following:-

- **New-DbaFile** - effectively runs the code inside the while loop
- **New-DbaFilegroup** - runs the code that creates the filegroup, and then has a while loop that runs New-DbaFile the required number of times; works either off a counter for number of files with stocks settings, or a hash table for filename/filelocation.  Perhaps also checks to see if TF1117 is set, and switches it on if the filegroup added has more than 1 file.  Has a switch for IsDefaultFilegroup and sets that, but this needs to only run if the database has been created because you can't set it before then).
- **New-DbaLogFile** - creates a log file.  If we need an additional log, we wouldn't be creating at db creation so I don't think this needs any kind of input other than parameters.
- **New-DbaDatabase** - runs all the above to create a new database, and has the options that can be set in the T-SQL CREATE DATABASE command, along with the ability to change the owner, set the recovery model, and change the database compatibility mode.  I think these are the three that are _most likely_ to be set often at creation time?

The additional ALTER DATABASE ... SET options could then be in a seperate function, that takes an object as input for the required commands, this could then be passed in to New-DbaDatabase and if it exists we run the function and apply those settings.

functions/New-DbaDatabase.ps1
galador: You can get the available collations by connecting to a server, and using the `EnumCollations()` method of the server object. You're right, though, it is a BIG list... the server I happened to have open returns 3887 entries...

```powershell
> $server = Connect-DbaSqlServer sql2014
> $server.EnumCollations().Rows.Count
3887
```

wsmelton: That sounds like a good command to me then :+1: `Get-DbaCollations`

potatoqualitee:
*reviews comments, runs 🏃 * 😂 lemme know how the progress is going, @matt-darwin

matt-darwin:
Apologies, been a bit tied up recently; my baby boy was born on the 18th! As soon as I get time I'll get back on this!

potatoqualitee:
heyyy, congrats! 👶 no problem, we'll be here when you return!

wsmelton:
Don't know if we want to scrap this PR and have it redone?  Since it has been here for so long.

potatoqualitee:
I wonder that as well, @wsmelton - I can also merge, not publish the code and then we'll still have it to work with. Perhaps, with your proposed changes, this could be v1.

potatoqualitee:
hey @matt-darwin - we are gonna go ahead and close this PR :( But we do welcome and need this command so when you are ready with the changes, we'll be happy to review and merge 🍹 🌻

potatoqualitee:
Thanks everyone for their time, and thank you @matt-darwin ! 🙇  I'll be merging this to a new branch and working from there.

Issue #2014, opened 2017-08-04
Backup-DbaDatabase not always doing diffs


potatoqualitee:
Got a report from a user via email and was able to repro with

`Backup-DbaDatabase -SqlInstance localhost -Type Diff`

![image](https://user-images.githubusercontent.com/8278033/28984892-6c21ce90-7960-11e7-8a94-ac15181402c3.png)

Specifying the database makes diff work, though as does piping (but only a single object) 🤔  (Though Type is misleading and will always say Full)

piping 1 object
![image](https://user-images.githubusercontent.com/8278033/28985171-9035c312-7961-11e7-95bc-870f5d8f47b6.png)

piping multiple objects

![image](https://user-images.githubusercontent.com/8278033/28985215-c009e276-7961-11e7-9d96-405faea9509a.png)

Stuart-Moore:
That's strange. Will have a look

PR #2027, opened 2017-08-06
Fixing only doing diffs on first piped in db

Stuart-Moore:
Fixing #2014 

Was doing a diff on the 1st db, then resetting the type.

Now an obvious change in var name to make sure it doesn't happen

## Type of Change

 - [ ] Pester test is included

### Purpose
Fixing Issue #2014

### Commands to test
Get-DbaDatabase -SqlInstance Server\Instance | Backup-DbaDatabase -Type Diff

### Learning
Make variable names really obvious. And reassign if it's going to be used again in a loop.

Don't break things before going on holiday as the broadband in the Cairngorms is a bit slow :)


potatoqualitee:
sweet, thanks @Stuart-Moore !!

PR #2028, opened 2017-08-06
Added test for Get-DbaDatabaseEncryption

jpomfret:
## Type of Change

 - [X] Pester test is included

### Purpose
 
Added test for Get-DbaDatabaseEncryption, this is my first real attempt at adding a test so let me know if I there are more tests I should add or if this one should be written differently.

### Screenshots

image


wsmelton:
Overall it looks fine to me. I would suggest though look at `Get-DbaDatabaseEncryption` and see if you can come up with additional `It` blocks to cover more of the code.

One could be using the `SqlCredential` to pass into the command...we need to be doing this for all of them. You would need to utilize T-SQL and maybe `Get-DbaLogin` to create the login if it doesn't exist. Then drop that login afterwards.

potatoqualitee:
@wsmelton - we have to keep contention in mind :( That would actually be a perfect test for Connect-SqlInstance. I'll make one this week or someone else can jump in if they are interested. I'm mostly slammed.

potatoqualitee:
Thank you, @jpomfret !! Love that another developer is on board with Pester 💃  💃  💃  💃

PR #2033, opened 2017-08-07
Get-DbaDatabaseState Test - renamed to temporarily disable

potatoqualitee:
Hey @niphlod - I'm seeing nearly 100% failure with this test, likely due to contention. Your tests are awesome, but can you perhaps limit some `if($env:APPVEYOR)` to set them not to run?


niphlod:
chicken and egg problem ... integration tests assure that the function won't break. 
I'm not testing anything twice, so there are none to "nicely disable".
Disabling some tests mean that we won't catch errors as soon as they happen.

potatoqualitee:
Can we perhaps put in a sleep? I didn't get a chance to see which one was causing the issue, just that it kept dying in a generic area.

potatoqualitee:
It looks like it may even be in the setup

https://ci.appveyor.com/project/sqlcollaborative/dbatools/build/0.8.7136

niphlod:
If the context is generic it surely is in the setup, but as you can see from the code there's a safeguard against it. It _must_ fail on setting some db properties rather than the sheer creation. I'll investigate more this evening

potatoqualitee:
Thanks, bud!

potatoqualitee:
this damn thing... took the biggest 💩 when I merged
https://ci.appveyor.com/project/potatoqualitee/dbatools/build/0.9.964

Maybe you can restart the SQL services and sleep a moment before `Restore-DbaFromDatabaseSnapshot` since that's where it all begins to fall apart? `If ($appveyor) { Restart ...`

potatoqualitee:
Trying now, wish us luck! https://ci.appveyor.com/project/potatoqualitee/dbatools/build/0.9.967

PR #2041, opened 2017-08-08
Dynamic whois active install

### LINKED ISSUE #1974: Install-DbaWhoIsActive make dynamic on version
## Feature Request
Update command to be dynamic when downloading the latest version of the procedure. Right now it has hard coded value.

### Problem to solve

As time goes and updates are made to sp_whoisactive would require us to also maintain that versioning in the function. Updating this to be more dynamic offers less management on code for our side.
### Additional information
Example of getting the latest version. As it stands going to the sit there are 2 versions listed, the first one is showing as the latest version so it would be assumed that going forward this logic should work.
```powershell
$baseUrl = "http://whoisactive.com/downloads"
$latest = ((Invoke-WebRequest -uri http://whoisactive.com/downloads).Links | where href -match "who_is_active" | select href -First 1).href
$baseUrl + "/" + $latest
```

alevyinroc:
## Type of Change

 - [ ] Pester test is included

### Purpose
 

Fixes #1974. Makes the sp_WhoisActive install more dynamic by selecting the first link on the download page, instead of hardcoding the filename/version number in the function. Also allows the user to specify any local filename for a previously-downloaded copy, either as a zip file or unpacked SQL script.

This is a breaking change because the `-Update` switch has been removed. See the Approach section for its replacement.

### Approach

If the `-LocalFile` parameter isn't supplied, the function goes out to http://whoisactive.com/downloads/, locates the first link, downloads, unpacks and installs.

If `-LocalFile` *is* supplied, that file is read as the installation of `sp_WhoisActive`. If it's a `.zip` file, it'll be unzipped first, otherwise it's assumed to be a plain text SQL script file.

I've also modified the unzipping step to use `Expand-Archive` if it's available, otherwise fall back to using the shell.

### Commands to test

Help has been updated with the appropriate examples

### DIFF
--- a/functions/Install-DbaWhoIsActive.ps1
+++ b/functions/Install-DbaWhoIsActive.ps1
@@ -72,7 +72,7 @@ function Install-DbaWhoIsActive {
 		[DbaInstanceParameter[]]$SqlInstance,
 		[PsCredential]$SqlCredential,
 		[parameter(Mandatory=$false)]
-		[ValidateScript({Test-Path -Path $_ -PathType file})]
+		[ValidateScript({Test-Path -Path $_ -PathType Leaf})]
 		[string]$LocalFile,
 		[object]$Database,
 		[switch]$Silent
@@ -82,7 +82,7 @@ function Install-DbaWhoIsActive {
 		$temp = ([System.IO.Path]::GetTempPath()).TrimEnd("\")
 		$zipfile = "$temp\spwhoisactive.zip"
 
-		if ($LocalFile -eq $null) {
+		if ($LocalFile -eq $null -or $Localfile.Length -eq 0) {
 			$baseUrl = "http://whoisactive.com/downloads"
 			$latest = ((Invoke-WebRequest -uri http://whoisactive.com/downloads).Links | where-object {$PSItem.href -match "who_is_active"} | Select-Object href -First 1).href	
 			if ($PSCmdlet.ShouldProcess($env:computername, "Downloading sp_WhoisActive")) {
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Hey @alevyinroc - thanks for the update. My only concern is removing the SQL file, because we'll be hitting Adam's site to download the zip with every execution.

My approach was also faulty because it kept the "temp" file for too long. Can you implement a compromise? Don't delete the extracted SQL file but check the site to see if it has a newer version and compare?

alevyinroc:
@potatoqualitee That's workable. We know the SQL filename that's inside the ZIP file matches the name of the ZIP file, so I can compare the names after stripping the extensions and if they don't match, do the full download.

I just wish we had a more permanent place to put the downloads.

Another thought! What if I were to also run `exec $database.dbo.sp_whoisactive @help=1` and parse the output of that to compare the version installed to the version that's posted online? If the versions match, don't download (and exit gracefully). If the versions don't match, or the stored proc doesn't exist at all, go ahead and download it or install from the previously-saved download?

niphlod:
@alevyinroc : I think we have. If @potatoqualitee (and @FriedrichWeinmann ) agree with it, it can be placed where dbatools stores user data already (see Path.DbatoolsData), or the whole logic in Get-DbaSqlBuildReference [here](https://github.com/sqlcollaborative/dbatools/blob/development/functions/Get-DbaSqlBuildReference.ps1#L83)

potatoqualitee:
that sounds awesome, @alevyinroc ! Great solution. @niphlod has a good point, too. One thing that we must make sure we do is force the end-user to download it at least once. We have only been granted permission to download, not to include in the repo itself.

niphlod:
yeah @potatoqualitee , the logic I was referring to is the one that saves the file in the userdata dir , which is guaranteed to be writable and "dbatools only property".

potatoqualitee:
Perfect <3 Please do implement, Andy.

potatoqualitee:
andy, i'll merge this then you can do a PR with the new changes when you are ready 👍

PR #2043, opened 2017-08-08
More Test changes -  mostly restarts

potatoqualitee:


potatoqualitee:
Removing read-only tests only for appveyor seems to soothe its soul

PR #2049, opened 2017-08-08
Connect-SqlInstance - fix it so that "$server" returns name and not guid

potatoqualitee:


FriedrichWeinmann:
Looks good, but don't forget Connect-DbaSqlServer :)

PR #2050, opened 2017-08-08
Update/New Commands - Get-DbaSqlService, Stop/Start/Restart-DbaSqlService

nvarscar:
## Type of Change

 - [x] Pester test is included

### Purpose
Submitting updated code to match the dbatools/sqlservice branch.
Also, adding some more error handling to the Update-DbaSqlServiceStatus procedure

### Approach
Moved corresponding lines from the dbatools/sqlservice branch that use Get-DbaCmObject instead of Get-CimInstance

### Commands to test
Get-DbaSqlService, Stop/Start/Restart-DbaSqlService

### Screenshots
n/a

### Learning
Get-DbaCmObject seems to be a much better option.


nvarscar:
@potatoqualitee how does one re-run the test if it failed due to contention?

potatoqualitee:
I push a lame commit that removes or add empty space 👎  Sometimes, I just go on faith after it works locally.

potatoqualitee:
@PowerDBAKlaas - when you get a moment, please let me know if you're happy with these changes. I'd like to get it merged. If not, let me know too!

PowerDBAKlaas:
@potatoqualitee  I thought there were conflicts with PR #1887 
Are those solved?
And is there any progress on the speed issue? If this new version with `Get-DbaCmObject` remains slower than the old version, should we do this?

functions/Get-DbaSqlService.ps1
```diff
@@ -65,57 +68,38 @@ Function Get-DbaSqlService {
 	Param (
 		[parameter(ValueFromPipeline = $true)]
 		[Alias("cn", "host", "Server")]
-		[string[]]$ComputerName = $env:COMPUTERNAME,
+		[DbaInstanceParameter[]]$ComputerName = $env:COMPUTERNAME,
 		[Alias("Instance")]
 		[string[]]$InstanceName,
 		[PSCredential]$Credential,
-		[ValidateSet("Agent", "Browser", "Engine", "FullText", "SSAS", "SSIS", "SSRS", "AnalysisServer", "ReportServer", "Search", "SqlAgent", "SqlBrowser", "SqlServer", "SqlServerIntegrationService")]
+		[ValidateSet("Agent", "Browser", "Engine", "FullText", "SSAS", "SSIS", "SSRS", "AnalysisServer", "ReportServer", "FullTextFilter Daemon Launcher", "FullText Search", "SqlAgent", "SqlBrowser", "SqlServer", "SqlServerIntegrationService")]
```
wsmelton: I have to ask...why are we repeating service names here?

internal/Change-DbaSqlServiceStartupMode.ps1
```diff
@@ -1,26 +1,43 @@
 Function Change-DbaSqlServiceStartupMode {
```
wsmelton: This is not an approved verb for a command and should be changed to something else....likely `Invoke-...`

wsmelton:
Error received trying to import these commands/changes:
image

image


Looking at Stop-DbaSqlService is anyone else seeing this when they pull down this code?

image

functions/Get-DbaSqlService.ps1
nvarscar: The idea was to have all the service names standardized by using corresponding SMO object as an example. Before that, the function was using custom service names. This approach allows for using both standards (also, making sure there's no code breaks).

nvarscar:
@wsmelton I do see it if I pull it... not sure what's going on here, will try to re-upload

internal/Change-DbaSqlServiceStartupMode.ps1
wsmelton: Actually I guess `Edit` would be the right verb. As well since this is internal we can make it shorter and just say something like `Edit-ServiceStartMode`. It is known/understood that we work with services around SQL Server.

PowerDBAKlaas: Why not `Set-`?

potatoqualitee: Yes, set please

functions/Get-DbaSqlService.ps1
wsmelton: There is already code break...the command is in a previous PR to be rewritten already. We need to pick one because you have to tab 15 times to get to `SqlServerIntegrationService`. I would go for the short names and see no reason to have it standardized to a corresponding SMO object.

I believe the original purpose of the shorter names was to map those to the publicly-known name the services are referenced by.

internal/Change-DbaSqlServiceStartupMode.ps1
```diff
@@ -1,26 +1,43 @@
...
+		Sets all SQL services on sql1 to Manual startup.
+		
+		.EXAMPLE
+		$services = Get-DbaSqlService -ComputerName sql1 
+		Change-DBASqlServiceStartupMode -ServiceCollection $services -Mode 'Automatic'
+		
+		Sets all SQL services on sql1 to Automatic startup.
 
 #>
 	[CmdletBinding(SupportsShouldProcess = $true)]
```
wsmelton: The ValidateScript should be removed from this command as it is in an internal command there is no reason to validate it anyway.

nvarscar: @wsmelton this internal command represents a method in a public function and outputs a meaningful error when user is trying to run (Get-DbaSqlService).ChangeStartMode('NonExistingMode'). Please LMK if there's a better way of doing it - I probably could've moved it to the method itself instead of running it within validation.

nvarscar:
@potatoqualitee @wsmelton @PowerDBAKlaas thanks a lot for your input! Everything has been updated according to your suggestions (please let me know if I still should remove ValidateScript in internal/Set-ServiceStartMode.ps1 )

wsmelton:
If you feel the validation is needed the  do that in your code. Adding that to the parameter overwrites the ability of our message system to not output anything. 

So it still needs to be removed, but again if you want to keep validation (although I see no purpose in it for an internal command) just put that in a begin block or something. 

_Sent from my Htc HTC6545LVW using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

nvarscar:
@wsmelton fixed. Thanks!

PR #2072, opened 2017-08-10
Change Get-DbaRegisteredServerName to take a -FullObject parameter.

galador:
* `-FullObject` returns the actual SMO object
 * Changed to use `Resolve-DbaNetworkName` to get the NetBIOS name / IP address for the Registered Servers
 * Various documentation updates.

## Type of Change

 - [ ] Pester test is included

### Purpose
 
Allows the function to return the full RegisteredServer object from SMO, but keeps output to ServerName by default for backwards compatibility.

Also, while testing, I found several unrelated issues with Example documentation, so I updated that, too! :)

Everything I tested worked in my environment, but I am open to feedback / suggestions!

### Approach

Previously, the function would get the RegisteredServer objects from SMO, but immediately extract the ServerName field. Now, I'm keeping the full SMO object, and filtering at the end.

I also changed to use `Resolve-DbaNetworkName` to handle NetBIOS / IP Address lookups, instead of the roll-your-own version in the function.

### Commands to test

There are several example commands in the documentation.  I also searched for `Get-DbaRegisteredServerName`, and found other functions that use it within their documentation, to make sure the changes didn't break those. Examples from the docs:

`Get-DbaRegisteredServerName -SqlInstance sql2016 | Get-DbaBackupHistory`
`Get-DbaRegisteredServerName -SqlInstance sql2014 | Get-DbaServerInstallDate` 
`Get-DbaRegisteredServerName -SqlInstance sql2014 | Get-DbaTcpPort -ExcludeIpv6 -Detailed -Verbose` 
`Get-DbaRegisteredServerName -SqlInstance sql2014 | Get-DbaUptime`
`Get-DbaRegisteredServerName -SqlInstance sql2014 | Resolve-DbaNetworkName`

### DIFF
--- a/functions/Get-DbaRegisteredServerName.ps1
+++ b/functions/Get-DbaRegisteredServerName.ps1
@@ -202,20 +202,20 @@ function Get-DbaRegisteredServerName {
 		}
 
 		if ($IpAddress) {
-			$ret = ($servers | Select-Object IPAddress)
+			$ret = @($servers | Select-Object IPAddress)
 
 			if (!$NoCmsServer) {
-				$ret += ($cmsServers | Select-Object IPAddress)
+				$ret += @($cmsServers | Select-Object IPAddress)
 			}
 
 			$ret | Select-Object -Unique -ExpandProperty IPAddress
 		}
 
 		elseif ($NetBiosName) {
-			$ret = ($servers | Select-Object ComputerName)
+			$ret = @($servers | Select-Object ComputerName)
 
 			if (!$NoCmsServer) {
-				$ret += ($cmsServers | Select-Object ComputerName)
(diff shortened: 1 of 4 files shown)


potatoqualitee:
Thanks for this! Those CMS commands are a beesh. This is a nice middle ground until this command gets replaced. I was thinking in the future, we can add a .ToString() method that resolves to the name like this 💃

LOL just saw you added a commit! Once it passes the tests, I'll merge 🏆

galador:
I think there will still be one failing test, this one, I accidentally missed some logic for adding the CMS itself back to the output. Just about to commit again. :)

potatoqualitee:
lookin good!

PR #2075, opened 2017-08-10
Logship fixes - remote execution issues Invoke-DbaLogShipping

### LINKED ISSUE #2074: Invoke-DbaLogShipping
## Bug Report
When running 
receive error on test-path when using remote credentials, also double hop issue when running via winrm when creating directories with new-item.

Test-Path : Cannot retrieve the dynamic parameters for the cmdlet. The FileSystem provider supports credentials only on the New-PSDrive cmdlet. Perform the operation again without specifying credentials.
At line:1297 char:15
 if (-not (Test-Path -Path $DatabaseCopyDestinationFolder -Credentials

- running development branch, previous release "960" Invoke-DbaLogShipping does not work
- believe the issue is related to test-path not supporting credentials properly.
- Can duplicate in new/fresh PowerShell.exe session

### Version Information
 - WIndows Server 2016
 - PowerShell Version: 5

 - SQL Server 2016 standard

## Steps to Reproduce
Run remotely via winrm or on primary node.
Invoke-DbaLogShipping -SourceSqlInstance WIN2016\PRIMARY -DestinationSqlInstance WIN2016-DR -Database trans -BackupNetworkPath \\WIN2016\logship -Force -CompressBackup -CopyDestinationFolder \\server-dr\logship -SourceCredential $Source_Credentials -SourceSqlCredential $Source_Credentials -DestinationCredential $Dest_Credentials -Verbose -DestinationSqlCredential $Dest_Credentials

### Problem to solve
Update test-path to Test-DbaSqlPath, also good for to keep with common functions

Also when running remotly via winrm issue with double hops occur when using New-item.
Recommend updating to use New-DbaSqlDirectory

lukecl:
## Type of Change
 - [ ] Bug fix non-breaking change, fixes #2074

### Purpose
Fix remote credential issues, also standardizes functions 

### Approach
update new-item to New-DbaSqlDirectory and test-path to Test-DbaSqlPath

### Commands to test
details in issue #2074

### Learning
I'm testing dbatools remotely via a control server which is showing where credentials are not being passed. double hops.

### DIFF
--- a/functions/Invoke-DbaLogShipping.ps1
+++ b/functions/Invoke-DbaLogShipping.ps1
@@ -133,7 +133,7 @@ The parameter works as a prefix where the name of the database will be added to 
 The default is "LSBackup_[databasename]"
 
 .PARAMETER CopyRetention
-The copy retention period in minutes. Default is 4320 / 72 hours
+The copy retention period in minutes. Default is 4320 / 72 hours.
 
 .PARAMETER CopySchedule
 Name of the backup schedule created for the copy job.
--- a/functions/Invoke-DbaLogShipping.ps1
+++ b/functions/Invoke-DbaLogShipping.ps1
@@ -664,7 +664,7 @@ The script will show a message that the copy destination has not been supplied a
 		}
 
 		# Check the backup network path
-		if (-not ((Test-Path $BackupNetworkPath -PathType Container -IsValid -Credential $SourceCredential) -and (Get-Item $BackupNetworkPath).PSProvider.Name -eq 'FileSystem')) {
+		if ((Test-DbaSqlPath -Path $BackupNetworkPath -SqlInstance $SourceSqlInstance -SqlCredential $SourceCredential) -ne $true){
 			Stop-Function -Message "Backup network path $BackupNetworkPath is not valid or can't be reached." -Target $SourceSqlInstance 
 			return
 		}
@@ -679,7 +679,8 @@ The script will show a message that the copy destination has not been supplied a
 			$CopyDestinationFolder = "$($DestinationServer.Settings.BackupDirectory)\Logshipping"
 
 			# Check to see if the path already exists
-			if (Test-Path $CopyDestinationFolder -Credential $DestinationCredential) {
+            
+			if (Test-DbaSqlPath -Path $CopyDestinationFolder -SqlInstance $DestinationSqlInstance -SqlCredential $DestinationCredential) {
 				Write-Message -Message "Copy destination $CopyDestinationFolder already exists" -Level Verbose
 			}
 			else {
@@ -744,7 +745,7 @@ The script will show a message that the copy destination has not been supplied a
 				} # else not force
 			} # if test path copy destination
 		} # if not copy destination
-		elseif (-not ((Test-Path $CopyDestinationFolder -PathType Container -IsValid -Credential $DestinationCredential) -and (Get-Item $CopyDestinationFolder).PSProvider.Name -eq 'FileSystem')) {
+		elseif ((Test-DbaSqlPath -Path $CopyDestinationFolder -SqlInstance $DestinationSqlInstance -SqlCredential $DestinationCredential) -ne $true) {
 			Stop-Function -Message "Copy destination folder $CopyDestinationFolder is not valid or can't be reached." -Target $DestinationSqlInstance 
 			return
 		}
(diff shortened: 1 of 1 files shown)


functions/Invoke-DbaLogShipping.ps1
```diff
@@ -664,7 +664,7 @@ The script will show a message that the copy destination has not been supplied a
 		}
 
 		# Check the backup network path
-		if (-not ((Test-Path $BackupNetworkPath -PathType Container -IsValid -Credential $SourceCredential) -and (Get-Item $BackupNetworkPath).PSProvider.Name -eq 'FileSystem')) {
+		if ((Test-DbaSqlPath -Path $BackupNetworkPath -SqlInstance $SourceSqlInstance -SqlCredential $SourceCredential) -ne $true){
```
wsmelton: This changes the intention of this check the way I am looking at it.

`Test-DbaSqlPath` looks at the path as it pertains to SQL Server service and that account being able to access it. The original statement was checking if the path existed using an external credential (if passed in, `$SourceCredential`). The SourceCredential is utilized to access that path under Windows Authentication.  Since `Test-DbaSqlPath` directly connects to the SQL Server instance that Source credential does not (or may not) have access to the given SQL Server instance, but regardless you are validating the path based on the rights of the SQL Server service account.

The fix for this is to utilize `Invoke-Command2` to actually ensure the `Test-Path` is run on the remote system, as well that command handles the credential. You can see this done on line 703 of this command:
```powershell
Invoke-Command2 -ComputerName $DestinationServerName -Credential $DestinationCredential -ScriptBlock 
```

functions/Invoke-DbaLogShipping.ps1
```diff
@@ -1092,13 +1093,12 @@ The script will show a message that the copy destination has not been supplied a
...
 
-
 				# Checking if the database network path exists
-				if (-not (Test-Path -Path $DatabaseBackupNetworkPath -Credential $SourceCredential)) {
+				if ((Test-DbaSqlPath -Path $DatabaseBackupNetworkPath -SqlInstance $SourceSqlInstance -SqlCredential $SourceCredential) -ne $true) {
 					# To to create the backup directory for the database 
 					try {
 						Write-Message -Message "Backup network path not found. Trying to create it.." -Level Verbose
-						New-Item $DatabaseBackupNetworkPath -ItemType Directory -Credential $SourceCredential | Out-Null
+                        New-DbaSqlDirectory -Path $DatabaseBackupNetworkPath -SqlInstance $SourceSqlInstance -SqlCredential $SourceCredential
```
wsmelton: While this may work in most cases `New-DbaSqlDirectory` utilizes `xp_create_subdir`. It again utilizes permissions based on SQL Server which may not have rights to the directory, while the `$SourceCredential` will. Although the reverse can also be the same as the `$SourceCredential` may  not have rights to connect to the given SQL Server instance. 

I would say the safer process would be to leave the use of `New-Item` and just wrap it in the `Invoke-Command2` call.

A variable could be added in the begin block to determine if a remote connection has to be made but in generally we just go ahead and assume it will be.

potatoqualitee:
thanks so much, @lukecl! I will await @sanderstad's review.

To further @wsmelton's comment about the approach for the new directory, I think your approach could solve some Windows admin vs SQL Admin permissions issues. Would you be able to modify your PR so that it first attempts to create the new directory using New-DbaSqlDirectory then fails back to New-Item if that bombs? 🏖 🍹 💯 

Thanks again!

lukecl:
Thanks for the comments.
Would it be better to test if credentials are being used, if they are then use the Test-DbaSqlPath and New-DbaSqlDirectory functions, if not using credentials then fallback to the original statements minus the credential parameter?
From what i have read New-item and Test-path does not support credentials even though its documented, MS then released New-PSDrive in PS V3 to deal with credentials and remote drive mappings.

Using invoke-command2 as a wrapper could be become problematic, uses winrm to submit the command to the remote server, if using NTLM will need to add the hosts into trusted servers, also maintain firewall ports between hosts, and submit windows hostname vs SQL instance name, seems like alot of effort and complexity from no real benefit.

I like using the approach that Copy-SqlDatabase uses, Test-DbaSqlPath and New-DbaSqlDirectory which standardises the mechanics (principals are the same), also good those functions test if the SQL instance can actually access file system locations.

potatoqualitee:
Hey @sanderstad - can you give this PR a look? Thank you 🙇

functions/Invoke-DbaLogShipping.ps1
sanderstad: There will be situations where SQL Server account will not have access to certain directories. In most cases the service account that needs the access is the agent service account.

As I'm writing this I'm thinking we probably need to supply the user with information to check the directories after they're created. The dba/user should grant the agent service account the necessary privileges to the just created directories, otherwise the log shipping will not work..
In normal circumstances you use directories that are already being used by sql server or you've already set the privileges. 

I agree with the comment to make sure that in case the Test-DbaSqlPath does not work, we still try to create the directory in the original way using Invoke-Command2.

sanderstad:
There will be situations where SQL Server account will not have access to certain directories. In most cases the service account that needs the access is the agent service account.

As I'm writing this I'm thinking we probably need to supply the user with information to check the directories after they're created. The dba/user should grant the agent service account the necessary privileges to the just created directories, otherwise the log shipping will not work..
In normal circumstances you use directories that are already being used by sql server or you've already set the privileges.

I agree with the comment to make sure that in case the Test-DbaSqlPath does not work, we still try to create the directory in the original way using Invoke-Command2.

potatoqualitee:
hey @lukecl will you have time to update this PR within the next few days?

lukecl:
Hi All, i have tested this and working well with remote credentials.

potatoqualitee:
@ConstantineK - do you have a moment to fix the conflicts on this PR?

ConstantineK:
Normally I would say yes @potatoqualitee , but this code has 9 conflicts currently and many of them are non-trivial to suss out, aka you would need direct knowledge of the code's purpose to solve them; I am going to reach out to @lukecl on Slack and see how comfortable he is with the merge UI.

lukecl:
Lots of Fixes - dd217d25a4528aff709e3abd7b9a41faddbe8509
Fixed in above commit

PR #2077, opened 2017-08-10
[New Command] Invoke-CycleErrorLog

wsmelton:
## Type of Change

 - [ ] Pester test is included (_Will work on this while discussion is made on what this command should be called._ )

### Purpose
 
We all need a way to cycle the error log for SQL Server in PowerShell. I mean it is just technically two words to execute as T-SQL but what fun is that...
### Approach

There is no direct SMO method that can be used to perform this task so it is basically just executing the same T-SQL that we always do. I do output an object so you get some result as to whether it was cycled or not, for reporting purposes if needed.
### Commands to test

See the help in the command.
### Screenshots

A thousand words is not really needed for this command...


potatoqualitee:
awessssome! Thank you @wsmelton ! And thank you for testing, @gbargsley 

@PowerDBAKlaas - how do you like the name? We were kind of at a loss.

PowerDBAKlaas:
`Invoke-DbaCycleErrorLog` would be more appropriate, I feel.
the SP's work on `_errorlog` and `agent_errorlog` respectively

MS has confused us by using different names for those logs depending on the interface a DBA uses 😞 
I'm in general in favour of naming it 'log' or 'eventlog' or something, since there are not only errors in it, but in this case, keeping the name of the SP's a DBA may know, may be the most recognizable choice, and make a clearer distinction from truncate log or VLF or anything else.

potatoqualitee:
I like that, thank you, @PowerDBAKlaas! @wsmelton can you rename and I'll merge? When I rename, it gives me all the credit/blame 👎 

Thanks for this function, btw! WIll come in handy.

wsmelton:
I not sure I understand this comment:

> depending on the interface

The `sp_cycle_errorlog` and sp_cycle_agent_errorlog` are separate commands and separate components of SQL Server. In general most folks don't even both with cycling the agent log so figured it was not worth a dedicated command, and just included it in this one.

The function was already named `Invoke-DbaCycleErrorLog` but noticed I didn't have the file matching that name so it has been updated.

wsmelton:
I'll add a quick/basic test for this one and get it added to this PR.

PowerDBAKlaas:
@wsmelton  I mean that in SSMS, if you navigate to 'server > Management > SQL Server Logs' you arrive at the same collection as via 'server > SQL Server Agent > Error Logs'
So, one time they call it 'error logs' and the other time they don't.

niphlod:
@PowerDBAKlaas : nope
'server > Management > SQL Server Logs'  > ERRORLOG
'server > SQL Server Agent > Error Logs' > SQLAGENT.OUT

PowerDBAKlaas:
yes, but I'm only talking about the interface. Both context menus open the same viewer, it's just one of both logs that is preselected, and you can check and uncheck all of them. Yet, the folder once has 'error' in it's name and once it hasn't.

PR #2080, opened 2017-08-11
Get-DbaDiskSpace needs to check for additional name filters to positively identify service names. 

pollusb:
I really wonder why Export-DbaSpConfigure is not existing in development and it does in master.

## Type of Change

 - [ ] Pester test is included

### Purpose
Bug 1 - changed break for return. It was breaking my loop calling this function.
Bug 2 - changed -CheckForSql option in Get-DbaDiskSpace to work with SQL 2000, it also works with SQL 2008


ConstantineK:
Looks like the antecedent is interesting, I am going to reject your changes to Export-DbaSpConfigure and leave your other changes for further review, and in that way fix the merge conflict and remove it as a candidate file.

pollusb:
You rejected the change to Export-DbaSpConfigure because the file does not exists?

ConstantineK:
Actually, the file does exist @pollusb, the interface is being a bit misleading. 
What actually happened is your version and our version conflicted, but your version apparently was changed in such a way that it appeared it had an entirely different antecedent, which made it appear...interesting.

I have added an additional commit to your branch to manage the merge conflict so that your actual changes can be reviewed on their own merit apart from some weird git stuff which was not your goal to send in as a pull request.

potatoqualitee:
Thanks bud! Looks good on my machine and passed the new Test I added 💯

pollusb:
Thanks,
J'ai bien hâte de contribuer de nouveau.

PR #2091, opened 2017-08-13
Cache WhoisActive download to avoid repeated downloads

alevyinroc:
## Type of Change
 - [ ] Pester test is included

### Purpose
Caches the `sp_WhoisActive` zipfile in the local dbatools data directory so that it isn't re-downloaded repeatedly from the internet.

Per conversation in [PR 2041](https://github.com/sqlcollaborative/dbatools/pull/2041)

### Approach
1. Goes to the website & gets the name of the latest version
1. Looks for a file by the same name in `Path.DbatoolsData`
1. If found, uses that local file for the installation
1. If not found, downloads the file from the internet and saves it to `Path.DbatoolsData` for future use

### Commands to test
See help

### DIFF
--- a/functions/Install-DbaWhoIsActive.ps1
+++ b/functions/Install-DbaWhoIsActive.ps1
@@ -79,6 +79,7 @@ function Install-DbaWhoIsActive {
 	)
 	
 	begin {
+        $DbatoolsData = Get-DbaConfigValue -Name "Path.DbatoolsData"
 		$temp = ([System.IO.Path]::GetTempPath()).TrimEnd("\")
 		$zipfile = "$temp\spwhoisactive.zip"
 
@@ -99,7 +100,7 @@ function Install-DbaWhoIsActive {
 						$url = $baseUrl + "/" + $latest
 						try {
 							Invoke-WebRequest $url -OutFile $zipfile -ErrorAction Stop
-							Copy-Item -Path $latest -Destination $LocalCachedCopy
+							Copy-Item -Path $zipfile -Destination $LocalCachedCopy
 						}
 						catch {
 							#try with default proxy and usersettings
@@ -223,7 +224,9 @@ function Install-DbaWhoIsActive {
 		}
 	}
 	end {
-        Get-Item $sqlfile | Remove-Item
+        if ($PSCmdlet.ShouldProcess($env:computername, "Post-install cleanup")) {
+            Get-Item $sqlfile | Remove-Item
+        }
 		Test-DbaDeprecation -DeprecatedOn "1.0.0" -Silent:$false -Alias Install-SqlWhoIsActive
 	}
 }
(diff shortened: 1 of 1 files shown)


functions/Install-DbaWhoIsActive.ps1
```diff
@@ -79,27 +79,38 @@ function Install-DbaWhoIsActive {
...
 			$baseUrl = "http://whoisactive.com/downloads"
 			$latest = ((Invoke-WebRequest -uri http://whoisactive.com/downloads).Links | where-object {$PSItem.href -match "who_is_active"} | Select-Object href -First 1).href	
-			if ($PSCmdlet.ShouldProcess($env:computername, "Downloading sp_WhoisActive")) {
-				try {
-					Write-Message -Level Verbose -Message "Downloading sp_WhoisActive zip file, unzipping and installing."
-					$url = $baseUrl + "/" + $latest
+			$LocalCachedCopy = Join-Path -Path $DbatoolsData -ChildPath $latest;
+
+			if (Test-Path -Path $LocalCachedCopy -PathType Leaf) {
+				Write-Message -Level Verbose -Message "Locally-cached copy exists, not downloading."
```
wsmelton: Would reword this to say maybe 

> Locally-cached copy found, skipping download

Most of the commands we use "skipping" instead of "not".

functions/Install-DbaWhoIsActive.ps1
```diff
@@ -79,27 +79,38 @@ function Install-DbaWhoIsActive {
...
+				if ($PSCmdlet.ShouldProcess($env:computername, "Copying sp_WhoisActive from local cache for installation")) {
+					Copy-Item -Path $LocalCachedCopy -Destination $zipfile;
+				}
+			} else {
```
wsmelton: The `else` should be on a new line.

functions/Install-DbaWhoIsActive.ps1
```diff
@@ -79,27 +79,38 @@ function Install-DbaWhoIsActive {
...
+						try {
+							Invoke-WebRequest $url -OutFile $zipfile -ErrorAction Stop
+							Copy-Item -Path $zipfile -Destination $LocalCachedCopy
+						}
+						catch {
+							#try with default proxy and usersettings
+							(New-Object System.Net.WebClient).Proxy.Credentials = [System.Net.CredentialCache]::DefaultNetworkCredentials
+							Invoke-WebRequest $url -OutFile $zipfile -ErrorAction Stop
+						}
+					} catch {
```
wsmelton: Catch should be on a new line to match format of the file.

wsmelton:
General comment, in using a cached copy how do we know it is the latest version?

I mean Adam does not release new versions all that often but if we are going to cache it we should have some mechanism to know it is the latest version...if possible. As well, we need to offer a switch, maybe just `-Force` that ops in to having it downloaded every time and not dependent upon checking for a local copy.

alevyinroc:
When I download the zipfile and save it locally, I preserve the filename Adam uses on his website. On subsequent runs, I grab the web page and check the name of the zipfile linked to by the first link on the page. If that name matches the cached download's name, I assume there's no newer version, and skip the download.

A `-Force` is a good idea in the event the local copy is corrupted somehow.

potatoqualitee:
awesome thanks! will merge and retest.

PR #2092, opened 2017-08-13
Add new function `Get-DbaAvailableCollation`

galador:
## Type of Change

 - [ ] Pester test is included

### Purpose
Add new function `Get-DbaCollations`, which returns the available collations for each SQL Server. It was [mentioned in the PR for `New-DbaDatabase`](https://github.com/sqlcollaborative/dbatools/pull/2024#discussion_r132829596) that this could be a good command to add.

### Approach
Create a new cmdlet. :)

### Commands to test
Make sure the following works on all the various supported versions of SQL Server and Windows. The command with the `-Detailed` switch uses some .NET APIs to pull the Code Page and Locale names. 

Since PowerShell requires .NET, and these functions have been in .NET since 2.0, I *think* it should be safe for all supported platforms.

`Get-DbaCollations -SqlInstance sql2016 -Detailed`


wsmelton:
Need to remove the psd1 file from this PR.

functions/Get-DbaCollations.ps1
```diff
@@ -0,0 +1,104 @@
...
+					Description      = $_.Description     
+					CollationVersion = $_.CollationVersion
+				}
+
+				if ($Detailed) {
+					$collation.CodePageName = (Get-CodePageDescription $collation.CodePage)
+					$collation.LocaleName = (Get-LocaleDescription $collation.LocaleID)
+				}
+
+				$availableCollations += $collation
```
wsmelton: An example of this is [Get-DbaDatabase](https://github.com/sqlcollaborative/dbatools/blob/development/functions/Get-DbaDatabase.ps1#L313)

functions/Get-DbaCollations.ps1
```diff
@@ -0,0 +1,104 @@
...
+		[Alias("ServerInstance", "SqlServer")]
+		[DbaInstanceParameter[]]$SqlInstance,
+		[PSCredential]$SqlCredential,
+		[switch]$Detailed,
+		[switch]$Silent
+	)
+	
+	process {
+		function Get-CodePageDescription ($CodePage) {
+			$encoding = [System.Text.Encoding]::GetEncoding($CodePage)
```
wsmelton: If I recall we've got another command that does something similar to this but we found that this would only get the code page for the local machine. So in cases where this is for a remote instance, this may not be accurate.

functions/Get-DbaCollations.ps1
```diff
@@ -0,0 +1,104 @@
...
+		}
+
+		function Get-LocaleDescription ($LocaleId) {
+			if ($LocaleId -eq 66577) {
+				#No longer supported by Windows, but still shows up in SQL Server
+				#http://www.databaseteam.org/1-ms-sql-server/982faddda7a789a1.htm
+				return "Japanese_Unicode"
+			}
+
+			$culture = [System.Globalization.CultureInfo]::GetCultureInfo($LocaleId)
```
wsmelton: Same thing as CodePage, this would only get the local id for the local machine and not a remote instance.

functions/Get-DbaCollations.ps1
wsmelton: Found it, we had that issue getting the language ID for the OS, so drew up an internal command [Get-Language](https://github.com/sqlcollaborative/dbatools/blob/development/internal/Get-Language.ps1). If you wanted you could do the same as maybe `Get-Codepage` and `Get-Locale` ?

functions/Get-DbaCollations.ps1
galador: It can, but the data type is a .NET DataTable, which appears to include links to the other rows on each row. It was much slower to keep the same object, and add properties to it, than create a new object with just the fields we care about.

The difference was about 2.5-3 seconds for the DataTable and <1 second for the custom object.

functions/Get-DbaCollations.ps1
```diff
@@ -0,0 +1,104 @@
...
+				}
+
+				$availableCollations += $collation
+			}
+
+			if ($Detailed) {
+				Select-DefaultView -InputObject $availableCollations -Property Instance, Name, CodePage, CodePageName, LocaleID, LocaleName, Description
+			}
+			else {
+				Select-DefaultView -InputObject $availableCollations -Property Instance, Name, CodePage, LocaleID, Description
```
galador: I was originally going to do it that way, but it was about 3x slower to do it that way (from >1 second to about 3 seconds). Not a huge deal for one call, but that could add up.

functions/Get-DbaCollations.ps1
galador: Unless I'm missing something, doesn't `Get-Language` do the exact same thing I'm doing? I could (and probably will, no use in duplication) change my function to use `Get-Language`, but wouldn't that still just return the cultures that are on the local machine?

potatoqualitee:
Oh, and don't forget to rename the file, too 👍 When I rename files, it gives me all the code credit 👎

functions/Get-DbaCollations.ps1
wsmelton: Is the additional information in that data table useless or no potential to be used in other ways by users? If I recall the command that reads the error log for SQL Server returned a data table as well....though it was pretty fast to process on servers I tested.

If the data in the object is of no use to anyone I'm fine with using the custom object. I would put a block comment noting as much so future reference can be made as to why we chose to "break" the norm.

potatoqualitee:
🍳 cellent :D Going to merge and patch - just gonna make it conform with our ComputerName, InstanceName, SqlInstance format. Thanks so much, @galador !

PR #2105, opened 2017-08-15
Fix Copy-DbaSsisCatalog with Named Instances

slimrichard:
The get-service check is passed the raw connection string so fails if a named instance is used. Added a sanitization to grab just the servername part.

## Type of Change

### Purpose
The function would fail if the destination was a named instance.

### Approach
Added a quick check for a \ and if one is found grab just the server part for the service check.

### DIFF
--- a/functions/Copy-DbaSsisCatalog.ps1
+++ b/functions/Copy-DbaSsisCatalog.ps1
@@ -109,7 +109,10 @@ Deploy entire SSIS catalog to an instance without a destination catalog.  Passin
             param (
                 [Object]$Computer
             )
-            $result = Get-Service -ComputerName $Computer -Name msdts*
+            if ($Computer -like "*\*") {
+                $Computer = $Computer -split '\\' | Select-Object -First 1
+            }
+	    $result = Get-Service -ComputerName $Computer -Name msdts*
             if ($result.Count -gt 0) {
                 $running = $false
                 foreach ($service in $result) {
--- a/functions/Copy-DbaSsisCatalog.ps1
+++ b/functions/Copy-DbaSsisCatalog.ps1
@@ -109,14 +109,11 @@ Deploy entire SSIS catalog to an instance without a destination catalog.  Passin
             param (
                 [Object]$Computer
             )
-            if ($Computer -like "*\*") {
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Thanks so much, Richard! Someone will take a look soon.

wsmelton:
@slimrichard if you are interested, this command could use some rework because the function you corrected can actually be completely replaced by using  `Get-DbaSqlService`.

As well, you can get the computer name by utilizing a property returned by `DbaInstanceParameter` type we have in dbatools. So line 265 can be updated to call the `Get-DbaSqlService`. A fail safe of sorts that we do in other commands is to resolve the network name of that server, in case aliases are being used. So we use `Resolve-DbaNetorkName` to make sure we get the `FullComputerName` and pass that into `Get-DbaSqlService` since it uses CIM/WMI.

wsmelton:
I know we had some other commands for SSIS catalog but not sure they are in a state to implement in this command either. To replace other internal functions like `New-CatalogFolder` for `New-DbaSsisCatalog` [code reference](https://github.com/sqlcollaborative/dbatools/blob/development/functions/New-DbaSsisCatalog.ps1).

slimrichard:
Hi @wsmelton I have replaced the get-service call with the requested get-dbasqlservice. This function handles named instances and returns the computername (I tested against aliases and clusters and it functioned well). I have left the original logic in the Get-RemoteIntegrationService function as it has logic I think is useful. As for the other changes I will need to set up a lab to make more substantial changes as I am doing this all through the github GUI so need a better environment for testing.

wsmelton:
Looking over that whole command there is a good bit that can/needs to be cleaned up and re-formatted based on how we have done other commands.

A lot of the logic could be cleaned up in that command. It shouldn't be creating connections manually and various things of that nature.

wsmelton:
A big one is a credential needs to be added for checking remotely for the service. That would be a new parameter. 

I'll see if I can't make a rough list in an issue and we can track what is fixed there.

wsmelton:
Going to go ahead and merge this and we can work on the other updates needed for the command in a later PR.

PR #2109, opened 2017-08-15
Add new function Get-DbaRegisteredServer, fixes #2083

### LINKED ISSUE #2083: Get-DbaRegisteredServerName should return object by default
### Problem to solve
This function is built backwards and has now logic opposite to other functions and best practices.

### Additional information

Since registered servers appear to be objects, this should be the default outcome of this function. It should then be renamed to `Get-DbaregisteredServer`. The user who wants just the names, has to `$myservers = Get-DbaRegisteredServer -... | Select -Expandproperty Name`

galador:
* Add `Get-DbaRegisteredServer`, which returns the full SMO object
 * Removed `Get-DbaRegisteredServerName`, and updated documentation
 * Change tests to reference `Get-DbaRegisteredServer`

## Type of Change

 - [x] Pester test is included *(changed old test to use new base function)*

### Purpose
Add `Get-DbaRegisteredServer`, which returns the full SMO object, instead of having a switch on `Get-DbaRegisteredServerName`.

### Concerns
1) I didn't want to get rid of `Get-DbaRegsiteredServerName` quite yet, because it is referenced in several places in the documentation for other functions. I tried adding a `SqlInstance` property to the SMO object, to see if the other functions would use that, but that still doesn't pass it through. Is there a way to fix that?
1) `Get-DbaRegisteredServerName` will include the name of the CMS instance you pass into the function, if you don't include a group (or if you don't include the `-NoCmsServer` switch). I'm not sure how to make this work with a full object, or if it would be needed. Thoughts?
1) @potatoqualitee, you changed the old version to only pull the NetBiosName / IP if those switches are passed in, for speed. I was going to add the IP/NetBIOS name on the full object before I saw that commit. I've just left it in the "Name" version for now. However, I didn't see much difference in speed (running `Measure-Command`) between a version with IP/NetBIOS and one without. I was testing on a CMS with about 100 instances (some of which are offline) listed in it. Any idea what the slowdown was?


wsmelton:
I would say `Get-DbaRegsiteredServerName` needs to be dropped. The examples can easily be updated across the commands that do reference it in examples.

Referring to `-NoCmsServer`, need to rename that to `-ExcludeCmsServer` but it should be just like filtering out a database. Just take your input object and filter out the CMS server from it. Something like `$cmsObject = $cmsObject | where Name -NE $CMSServer` or whatever that property is called. It is how the exclusions are done with databases and other SMO objects.

potatoqualitee:
Agreed, @wsmelton - it'll be dropped along with Restore-DbaBackupFromDirectory

potatoqualitee:
And thanks so much for all the work on these, @galador !

galador:
I'll work on updating the documentation tomorrow, then.

galador:
I pushed some updates for the documentation. I decided to add a `-NameOnly` parameter, to make it easier to pass the results to other functions. I figured this was a big enough use case to add a switch instead of having to always remember to add `| Select-Object -ExpandProperty ServerName -Unique` when piping to the other functions.

There was also a `Get-DbaRegisteredServerStore` function that @potatoqualitee added to get the CMS server object, but `Get-DbaRegisteredServer[Name]` was never changed to use it. I made that update, too.

cc/ @wsmelton @PowerDBAKlaas

PowerDBAKlaas:
@galador hmm, I don't think this switch is any more justified than with `Get-DbaInstance`, `Get-DbaDatabase`, `Get-ADUser`, `Get-ADComputer`, `Get-ChildItem` and all other commands.

'Having to always remember': ` Get-... | Select ...` is the PowerShell way, and any PS user should know this by heart after one day practice. Remembering which commands have a switch and which don't will be harder.

functions/Get-DbaRegisteredServer.ps1
```diff
@@ -0,0 +1,164 @@
...
+				Add-Member -Force -InputObject $server -MemberType NoteProperty -Name ComputerName -Value $null
+				Add-Member -Force -InputObject $server -MemberType NoteProperty -Name IPAddress -Value $null
+			}
+		}
+
+		if ($NameOnly) {
+			return $servers | Select-Object -ExpandProperty ServerName -Unique
+		}
+
+		return $servers
```
PowerDBAKlaas: Results should be returned as early as possible.
All steps can be executed in the `process` instead of in the `end` block.

galador:
Perhaps a better question would be: why can't the SMO RegsiteredServer object be converted to a `DbaInstanceParameter`? 

I think the type could just be added to the C# code that handles that custom parameter (somewhere around [here](https://github.com/sqlcollaborative/dbatools/blob/master/bin/projects/dbatools/dbatools/Parameter/DbaInstanceParameter.cs#L398) ), but I'm not sure if there's any other stuff to make that magic work.

FriedrichWeinmann:
You've put the finger right on the spot where it would need to be implemented :)
There is no additional magic required, on type mismatch during parameter binding, PowerShell will check for a valid constructor. All you'd need to make sure is that all the contracted fields are filled out as required (some are mandatory, some are optional - the attributes should tell you). I'll put adding RegisteredServer as input type on the list for my weekend tasks.

galador:
@FriedrichWeinmann I added code to allow `DbaInstanceParameter` to convert an SMO `RegisteredServer` in commit 3c517a8f6c 

Could you take a look?

wsmelton:
This PR should just be the changes to `Get-DbaRegisteredServerName`. All the other changes need to be a separate PR.

PR for the help updates
PR for the library updates.

galador:
Wouldn't making this one commit into three introduce the potential for breakage? All of these changes are related, and I see this all as one unit of work. If I made three separate PRs (that aren't pulled into `development` at the same time), there could be a chance that someone could pull code or docs that simply don't work.

Changing the function (and nothing else) would break the docs, and users couldn't pass the results from this function to others that have a `SqlInstance` parameter.

Changing the library (and nothing else) would work, but would be somewhat pointless, since the use case wouldn't exist yet.

Changing the docs (and nothing else) would cause user confusion, since none of the example would work.

wsmelton:
Yes, all them separate would depend on each other, but putting it all in one PR also makes it more difficult to revert changes that could break more than just these commands.

You are touching the library that regardless if you are only making a change to the library for one set of command(s) the library is a dependent on every command. You can do your PR in stages if you wish to prevent each one from failing test...I do this often and have in the past with the AG commands. I submit one AG that I need and I don't work on the next until it gets merged.

The main reason we do one command to one PR is for testing, we do not just depend on Appveyor for everything as we cannot test every command against the Appveyor environment.

galador:
I created a PR ( https://github.com/sqlcollaborative/dbatools/pull/2169 ) to split the SMO object conversion as step 1. I guess I'll close this one.

PR #2113, opened 2017-08-16
Get-DbaOperatingSystem - Adjusting try/catch and add PowerShellVersion property

wsmelton:
## Type of Change

 - [ ] Pester test is included

### Purpose
 
- Add `PowerShellVersion` property to output.
- Adjusted collection code to be separate try/catch blocks so we can more easily see/identify what area failed.

### Approach

When you get OS level information it can be convenient to know what version of PowerShell you are working with on the remote instance. Especially if working with SQL Agent jobs and such. 

I had thought of passing in the full PSVersion object but seemed mute as most info anyone wants to know is the major and minor version of PowerShell.
### Commands to test

`(Get-DbaOperatingSystem -ComputerName Server1).PowerShellVersion`

### DIFF
--- a/functions/Get-DbaOperatingSystem.ps1
+++ b/functions/Get-DbaOperatingSystem.ps1
@@ -39,7 +39,7 @@ function Get-DbaOperatingSystem {
 	[CmdletBinding()]
 	param (
 		[Parameter(ValueFromPipeline = $true)]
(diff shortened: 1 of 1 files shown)


potatoqualitee:
has a test, passes a test 👍

PR #2114, opened 2017-08-16
Write Exception to log when function Test-PSRemoting Fails

### LINKED ISSUE #2048: Get-DbaPrivilege needs more detailed information/error logging when WinRM is not accessible
## Bug Report

## Steps to Reproduce

- Stop WinRM service on local machine
- Execute Get-DbaPrivilege for local machine

The error says "Failed to connect to <computername>". It is missleading on what the issue actually is, I think the error should be logged to a log file like the dbatools log, and in the warning indicate the user that for further details go to the log file.

mvillegascuellar:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #2048) 
 
### Purpose
Write on the dbatools error log the real error message so the user is not missleaded by the generic error message that the command throws when the WinRM service is stopped.

### Approach
Add instruction to call Write-Exception function.

### Commands to test
Details on issue #2048.


wsmelton:
@mvillegascuellar If you would not mind we need to have this updated to utilize the messaging system we have available in dbatools. You can see some information on how that is done [here](https://github.com/sqlcollaborative/dbatools/wiki/Implementing-the-messaging-system).

The catch block in this one is a bit unique but would go something like this:

```powershell
Stop-Function -Messsage "Remote testing failed for computer $ComputerName" -Target $ComputerName -ErrorRecord $_ -Continue
return $false
```

mvillegascuellar:
Hi @wsmelton, I did the change, instead of writing the exception to the dbatools log file, I called the stop-function as you suggested and the output is now displayed directly to the user console and no message is written to the dbatools log file. Now the output looks like this:

image

Before, when I called the Write-Exception function the output looked like this:

image

I think this last one looks cleaner and the real error is written into the dbatools error log. I would also like to change the Write-Verbose calls from function Test-PSRemoting to Write-Message, what do you think?

wsmelton:
@mvillegascuellar all updates of Write-Verbose/Debug/Exception/Output should be changed over to Write-Message.

One main thing in this PR though is something is a bit odd on why the other commits are showing up. Unless @ConstantineK knows how to clean this up, it will likely be easiest to just submit a new PR. If you go that route please make sure the branch you create is based on the development branch and not master.

The commits showing up in this PR makes me think you have been synchronizing your local branch with master and not development.

mvillegascuellar:
Done @wsmelton, I created a new clean PR #2143 
This PR can be dissmissed.

PR #2115, opened 2017-08-16
New function - New-DbaLogin

### LINKED ISSUE #2034: Pester test that tests Connect-SqlInstance for Credential Support
This would be fun

Login with the credential via Connect-SqlInstance (since internals are exposed) then do a Get-DbaProcess using the same credential to ensure that credential is logged in and doing stuff.

nvarscar:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #2034)
 - [x] Pester test is included

### Purpose
Adding new function New-DbaLogin that supports creation of all kinds of logins. Doesn't cover permissions inside databases or server-level permissions.

### Approach
1 new function with all the blows and whistles
3 new internal functions to support hash-related operations
1 new pester test to test it all and also fix #2034 - please LMK if you want to see it as a separate pester test file

### Commands to test
Examples in the help
Pester tests
Couldn't test on 2000 and lower - password hash algorithm is slightly different there, but not sure if it works at all, since 2000 had different syntax and SMO
Credentials are not working on 2005, not sure if errors are handled correctly (do we still care about 2005?)

### Screenshots
image

### Learning
About generating hashed passwords:
http://sqlity.net/en/2460/sql-password-hash/
https://learn-powershell.net/2013/03/25/use-powershell-to-calculate-the-hash-of-a-file/


wsmelton:
@nvarscar you mention this fixes #2034 but I do not see a pester test for `Connect-SqlInstance` in this PR.

nvarscar:
Hey @wsmelton , as I wrote in the description, there is 1 new pester test to test this new command, which also does the testing of Connect-SqlInstance. I also asked to let me know if you guys want to see it as a separate pester test file, since it is an internal command and I wasn't sure if I should create a separate pester for it. There is a dedicated section in the existing pester that tests Connect-SqlInstance behavior and I could easily move it to a separate file.
```
+	Context "Connect with a new login" {
 +		It "Should login with newly created Sql Login (also tests credential login) and get instance name" {
 +			$cred = New-Object System.Management.Automation.PSCredential ("tester", $securePassword)
 +			$s = Connect-SqlInstance -SqlInstance $script:instance1 -SqlCredential $cred
 +			$s.Name | Should Be $script:instance1
 +		}
 +		It "Should return existing process running under the new login and kill it" {
 +			$cred = New-Object System.Management.Automation.PSCredential ("tester", $securePassword)
 +			$results = Get-DbaProcess -SqlInstance $script:instance1 -Login $cred.UserName
 +			$results | Should Not BeNullOrEmpty
 +			$results = $results | Stop-DbaProcess
 +			$results.Status | Should be 'Killed'
 +		}
 +	}
```

wsmelton:
Yeah, it needs to be a test by itself. The only thing test for `New-DbaLogin` would need to test is if a new login was created and various properties are set as called by the function.

The test for Connect-SqlInstance also should not depend on `New-DbaLogin` to create a login, you would have to create it the old fashion way with T-SQL or utilize SMO if you want. I know we do that for some but the Connect command is so important that test should be able to hold its own whether public commands are borked up or not. This will allow us to know Connect command is not actually broken.

potatoqualitee:
This is 💣 af! Fantastic work, Kilrill

PR #2124, opened 2017-08-18
Get-DbaDetailedPermission

potatoqualitee:
I need this for my own audit and asked my buddy to compile the DISA STIG perms into a command that won't leave junk behind. 

We weren't sure what to call it and decided on Get-DbaDetailedPermission. This command offers different info from Get-DbaPermission

### DIFF
--- a/bin/stig.sql
+++ b/bin/stig.sql
@@ -1,0 +1,809 @@
+USE tempdb;
+                GO
+
+                IF EXISTS (SELECT 1 FROM sys.schemas WHERE name = 'STIG')
+                BEGIN
+	                DROP SCHEMA STIG
+                END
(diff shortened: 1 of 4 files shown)


wsmelton:
Naming ideas because I don't care for the "detailed" word... 

- `Get-DbaDbSecurityAudit` (I mean just name it by subject matter and what it would be used for, the Db some of can be filtered) 
- `Invoke-DbaSecurityAudit` (since you require and external file, the SQL script)

potatoqualitee:
Klaas and I decided on Get-DbaUserLevelPermission

wsmelton:
If user level shouldn't it allow you to filter on a user?

PR #2128, opened 2017-08-18
Out-DbaDatatable updates -  kinda fixes #2095

### LINKED ISSUE #2095: Out-DbaDataTable Datatype Change later on in Object
I'm having an issue with `Test-DbaLatstBackup | Out-DbaDatatable` because I have a column that starts out as DateTime then later on in the object is an array of DateTime (which I generally handle with `-join "','"`)

Commenting out line 261 which sets an explicit datatype fixed the issue but that's a bad solution. Would it be possible to evaluate all rows before determining a datatype?

Also, can you please add `-or ... System.String[]` to line 280?

Thanks much,
Chrissy

potatoqualitee:


potatoqualitee:
passed tests

PR #2130, opened 2017-08-18
Get-DbaAgentLog - added try to catch instances not running agent

potatoqualitee:


wsmelton:
Just wondering but are you referring to the database instance itself? SQL Agent does not have to be running in order to read the log for it.

image

PR #2138, opened 2017-08-22
Tests: build matrix

niphlod:
## Type of Change
 - Pester tests reorganization (in appveyor)

### Purpose
Better tests stability (as agreed upon #dbatools-pester ) 

### Approach
Soooo, I reorganized things to be in 3 scenarios: *only* 2008R2, *only* 2016, *default* (as it was up until now). Setups for each scenario have been *clearly* separated, and a new file holding a mapping on which tests to run in each scenario too (which has also the "pro" of ending the "rename to .onhold" to avoid pester runs on appveyor, and a "future" path towards "which test can I run on my own lab ( i.e. _that hasn't two local instances active_ )"). Adding a scenario in the future in this way shouldn't be a PITA. Pardon the darkgreen coloring, I need _at least_ for a few weeks a clear separation of "what I did" vs "everything else" when bugs will arise. I also fixed a few rogue tests that were slipping away in the process. Let's hope all this work leads to more stability.


potatoqualitee:
This is phenomenal! 😍  Thanks niph!

PR #2143, opened 2017-08-23
Stop Function when function Test-PSRemoting Fails

### LINKED ISSUE #2048: Get-DbaPrivilege needs more detailed information/error logging when WinRM is not accessible

mvillegascuellar:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #2048)
 
### Purpose
Write a more accurate error message when the internal function Test-PSRemoting fails, so the user is not missleaded by the generic error message that the command throws when the WinRM service is stopped.

Also add standard messaging system to function Test-PSRemoting instead of Write-Verbose messages.

This is a clean PR that replaces PR #2114 

### Approach
Add standard messaging system

### Commands to test
Details on issue #2048

### DIFF
--- a/internal/Test-PSRemoting.ps1
+++ b/internal/Test-PSRemoting.ps1
@@ -15,25 +15,24 @@ https://www.petri.com/test-network-connectivity-powershell-test-connection-cmdle
   )
  
   Begin {
-    Write-Verbose -Message "Starting $($MyInvocation.Mycommand)"  
+    Write-Message -Level Verbose -Message "Starting $($MyInvocation.Mycommand)"
   } #begin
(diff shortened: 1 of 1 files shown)


wsmelton:
Failure on Appveyor does not pertain to this change so going to go ahead and merge.

Issue #2126, opened 2017-08-18
Restore-DBFromFilteredArray calls Stop-DbaProcess with incorrect parameter. Additionally verbose output gives the impression that restore succeeded, but testing of the object returned from Restore-DbaDatabase for the RestoreComplete value returns False.


pbarryuk:
### General Troubleshooting steps

- [x] Verified running latest release of dbatools?
- [x] Verified errors are not related to permissions?
- [x] Can duplicate in new/fresh PowerShell.exe session?

### Version Information
 - Operating System (Name|Version): Windows 10 Enterprise
 - PowerShell Version: 

Major  Minor  Build  Revision
-----  -----  -----  --------
5      1      14393  953     

 - SQL Server (Edition|Version): SQL Server 2016 Enterprise Edition (Build 13.0.4446)

## Steps to Reproduce

Trying to restore a database using restore-dbadatabase with the following parameters:

- -FileMapping (passing an  array to move the files on restore)

- -WithReplace

The restore detected sessions using the 

 - [ ] Attach any screenshots (if possible/allowed)
 - [ ] Attach logs generated by dbatools (`Get-DbatoolsLog`)
 - [x] Attach output from PowerShell console (if possible/allowed)

VERBOSE: [Get-FilteredRestoreFile][13:42:08] Got a Full/Diff backups, now find all Tlogs needed
VERBOSE: [Get-FilteredRestoreFile][13:42:08] Filtered 0 log backups down to 0
VERBOSE: [Get-FilteredRestoreFile][13:42:08] 0 Transaction Log backups found
VERBOSE: [Restore-DbaDatabase][13:42:08]  dbs to restore
VERBOSE: [Restore-DbaDatabase][13:42:08] Starting FileSet
VERBOSE: Test-DbaLsnChain - Testing LSN Chain
VERBOSE: Test-DbaLsnChain - Passed LSN Chain checks
VERBOSE: Test-DbaRestoreVersion - RestoreVersion is 11
VERBOSE: Test-DbaRestoreVersion - Closing smo connection
VERBOSE: [Restore-DBFromFilteredArray][13:42:09] Starting
[...]

Issue #2119, opened 2017-08-17
Restore-DbaDatabase with UseSourceDataDirectory


NeilWinder:
When I try to restore an existing database with the -UseSourceDataDirectory and -WithReplace options I get an error saying that the .mdf file already exists. When I don't use -UseSourceDataDirectory and specify an existing location the database is restored correctly.

For example, this works fine:
Restore-DbaDatabase -MaintenanceSolutionBackup: True -SqlInstance: MyInstance -Path: Y:\backups\MyDB -WithReplace: True -DestinationDataDirectory: e:\data\MyDB -DatabaseName: MyDB

But this gives an error:
Restore-DbaDatabase -MaintenanceSolutionBackup: True -SqlInstance: MyInstance -Path: Y:\backups\MyDB -WithReplace: True -ReuseSourceFolderStructure: True -DatabaseName: MyDB

WARNING: [Restore-DbaDatabase][16:25:54] Restore of MyDB failed, Destination File E:\data\MyDB\MyDB.mdf  exists on MyInstance

Stuart-Moore:
Hi,
does the file exist at the second location? (E:\data\MyDB\MyDB.mdf)

And could you let us have the verbose output of an example of failing
please. Just trying to work it out, but unfortunately have limited
resources as I'm on vacation atm so the more I've got to go on the easier
it is

Cheers
Stuart

NeilWinder:
Yes, the file is there. Basically I'm restoring a newer backup of MyDB to replace the current one. This is a DR server so every time the production server creates a backup I'm restoring it on the DR server.

If it helps, I believe the problem is in Restore-DBFromFilteredArray.ps1 around line 143. It creates the source folder structure if it doesn't exist, but if one of the files does exist then the function is exited. I think that's fine, but maybe at some point it should check for the -WithReplace option to allow the restore?

Attached is the verbose output.

[restore-verbose.txt](https://github.com/sqlcollaborative/dbatools/files/1235220/restore-verbose.txt)

Stuart-Moore:
Fixed in #2145

PR #2145, opened 2017-08-24
Restorebits

### LINKED ISSUE #2144: Restore-dbfromfilteredarray not account for filename with multiple periods

### Version Information
 - Operating System (Name|Version): WindowsServer2012R2
 - PowerShell Version: 5 

## Steps to Reproduce

create a database which has a filename that includes multiple periods. Such as NewTestV1.2.1.mdf and then copy that database with copy-dbadatabase. When the command is complete look at the new filename of the copied database.

### Problem to solve
We have a number of Databases which include additional periods (.) in the physical filenames. When using copy-dbadatabase with the -BackupRestore switch, the physical filenames are being set incorrectly at the new server. i.e.
NewDatabase_2017.1.mdf would become: "NewDatabase_2017.1 mdf"

### Additional information
 It appears that restore-dbfromfilteredarray is assuming the filename only contains a single period to denote the extension and is using the following to attempt to parse the value:
`$filename, $extension = (Split-Path $file.PhysicalName -leaf).split('.')`

some additional logic to test if the split returns more than two values would be helpful. In my local restore-dbfromfilteredarray I replaced the above line with the following which resolved things, though I'm sure there's a more eloquent solution:
```
$tempfname = (Split-path $file.PhysicalName -leaf).split('.')
if($tempfname.count -gt 2){
    $filename = $tempfname[0..($tempfname.length-2)] -join '.'
    $extension = $tempfname[$tempfname.length-1]
    }
else{
    $filename,$extension = $tempfname
}
```

Stuart-Moore:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #2144, #2119, #2126 )
 - [ ] Pester test is included

### Purpose
Fixing a couple of oddball occurences that weren't envisioned

### Approach
Handles the situations correctly

### Commands to test
See individual Issues

### DIFF
--- a/internal/Restore-DBFromFilteredArray.ps1
+++ b/internal/Restore-DBFromFilteredArray.ps1
@@ -205,7 +205,7 @@ Function Restore-DBFromFilteredArray {
                     $MoveFile = New-Object Microsoft.SqlServer.Management.Smo.RelocateFile
                     $MoveFile.LogicalFileName = $File.LogicalName
                     $Extension = ($file.PhysicalName.split('.'))[-1]
-                    $Filename  = $file.PhysicalName -replace ".$extension",""
+                    $Filename  = (split-path $file.PhysicalName -leaf) -replace ".$extension",""
                     if ($ReplaceDbNameInFile) {
                         $Filename = $filename -replace $OldDatabaseName, $dbname
                     }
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Thanks @Stuart-Moore ! Merging because failures were 100% DAC sucking.

PR #2162, opened 2017-08-28
Format and cleanup Get-DbaRoleMember

wsmelton:
Just cleaning up and format. Only minor code change to add the ComuterName, InstanceName and SqlInstance properties.

Also added a test for command.


wsmelton:
Closing this for now until I determine if we should split this command up.

PR #2169, opened 2017-08-29
Allow SMO RegisteredServer object to be passed as DbaInstanceParameter

galador:
## Type of Change

 - [ ] Pester test is included

### Purpose
First commit in splitting [PR 2109](https://github.com/sqlcollaborative/dbatools/pull/2109) into components. 

This PR is the first step, it allows the SMO object for a Registered Server to be passed into any function taking a `DbaInstanceParameter` object. For further information, please see the other PR.

### DIFF
--- a/bin/projects/dbatools/dbatools/Parameter/DbaInstanceParameter.cs
+++ b/bin/projects/dbatools/dbatools/Parameter/DbaInstanceParameter.cs
@@ -175,6 +175,8 @@ namespace Sqlcollaborative.Dbatools.Parameter
                             return DbaInstanceInputType.Server;
                         case "microsoft.sqlserver.management.smo.linkedserver":
                             return DbaInstanceInputType.Linked;
+                        case "microsoft.sqlserver.management.registeredservers.registeredserver":
+                            return DbaInstanceInputType.RegisteredServer;
                         default:
                             return DbaInstanceInputType.Default;
                     }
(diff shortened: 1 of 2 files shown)


wsmelton:
One note, that I'm not sure it matters as I don't claim to know all the C# stuff....but the parameter was renamed to `DbaInstance`. I know the `DbaInstanceParameter` still works but need to confirm when that is going to be dropped.

In case any of that depends on your code changes to function properly.

bin/projects/dbatools/dbatools/Parameter/DbaInstanceParameter.cs
```diff
@@ -450,6 +452,26 @@ public DbaInstanceParameter(object Input)
                         throw new PSArgumentException("Failed to interpret input as Instance: " + Input, e);
                     }
                     break;
+                case "microsoft.sqlserver.management.registeredservers.registeredserver":
```
zippy1981: Please create a C# unit tests in dbatools.test.csproj that covers at the very least this entire case statement in the `DbaInstanceParameter(object Input)` constructor overload.

potatoqualitee:
@zippy1981 / @wsmelton - this code has sample CMS entries https://github.com/sqlcollaborative/dbatools/blob/master/tests/Copy-DbaCentralManagementServer.Tests.ps1

galador:
This command also has some tests that build out a CMS... mostly the same commands (maybe those should be combined in a startup-type script?):  https://github.com/sqlcollaborative/dbatools/blob/master/tests/Get-DbaRegisteredServerName.Tests.ps1

zippy1981:
@potatoqualitee So wait what is actually changing with this PR? This is a new feature of SQL Server to me (can you tell I'm not a production DBA). 

I guess I just need to understand, do we expect that function to work exactly the same after this change, and are there more changes coming to use this changed code?

Yes playing a little dumb, but I'd like to understand the why.

galador:
@zippy1981 This is the first step in a multi-PR set of changes. They are being split from https://github.com/sqlcollaborative/dbatools/pull/2109

The simple version is that SQL Server has a feature to store a list of your instances, and organize them into groups. The function we have right now will just return the name of the instance, but my plan is to return the full SMO object (renaming `Get-DbaRegisteredServerName` to `Get-DbaRegisteredServer`).

The problem with that is that `DbaInstanceParameter` does not auto-convert the `RegisteredServer` SMO object (which it does for other class types). So, this PR adds the code that will do that conversion.

Instead of having to do something like `Get-DbaRegisteredServer ... | Select-Object -Unique -Property ServerName | Get-SomeOtherFunctionThatTakesSqlInstance ...`, you could just pass `Get-DbaRegisteredServer ... | Get-SomeOtherFunctionThatTakesSqlInstance ...` and it would just work 🌈 

I have another PR that I will do after this one that will do the function name change, and update the documentation.

zippy1981:
@potatoqualitee this can be merged. The unit tests are going to be hard. I made a note for me to figure them out in #2185.

PR #2177, opened 2017-08-31
Get-DbaUptime output & parameter updates

### LINKED ISSUE #2152: Get-DbaUptime output needs update
## Feature Request

The output of this command is controlled by two parameters: 
```powershell
.PARAMETER SqlOnly
Excludes the Windows server information

.PARAMETER WindowsOnly
Excludes the SQL server information
```

This does not really conform to our standard in other commands of having a default output and then users doing a `select-object` to return all of it.

### Problem to solve

I would say the last bit of code in this function that sets the output needs to be adjusted to be controlled by `$defaults` variable as we do with other functions. Then just pass that into `Select-DefaultView`. I would also suggest just having the SQL only output as default, or just outputting it all.

Could just work on adjusting it and see what makes the most sense.

alevyinroc:
## Type of Change

- [ ] Pester test is included
- [X] Spelling, grammar & other cleanup in Help and other text content

### Purpose
Fixes #2152. Eliminates the `-WindowsOnly` and `-SqlOnly` switches and always returns both Windows and SQL uptime, _unless_ Windows uptime cannot be obtained (in which case it's eliminated from the output). Uses `Set-DefaultView` for output.

### Commands to test

See help

### DIFF
--- a/functions/Get-DbaUptime.ps1
+++ b/functions/Get-DbaUptime.ps1
@@ -25,12 +25,6 @@ function Get-DbaUptime {
 
 			$wincred = Get-Credential, then pass $wincred object to the -WindowsCredential parameter.
 			
-		.PARAMETER SqlOnly
-			Excludes the Windows server information
-
-		.PARAMETER WindowsOnly
-			Excludes the SQL server information
-
 		.PARAMETER Silent
 			If this switch is enabled, the internal messaging functions will be silenced.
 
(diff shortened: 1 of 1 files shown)


functions/Get-DbaUptime.ps1
```diff
@@ -134,7 +134,7 @@ function Get-DbaUptime {
 				Select-DefaultView -InputObject $rtn -ExcludeProperty WindowsBootTime, WindowsUptime, SinceWindowsBoot
 			}
 			else {
-				Select-DefaultView -InputObject $rtn -Property ($rtn|get-member -MemberType NoteProperty|select-object -ExpandProperty name)
```
wsmelton: You can remove the whole if/else block since we dropped the parameters for the output.

Remove the variable `$rtn` and just have the custom object output straight. 
```powershell
			[PSCustomObject]@{
				ComputerName     = $WindowsServerName
				InstanceName     = $server.ServiceName
				SqlServer        = $server.Name
				SqlUptime        = $SQLUptime
				WindowsUptime    = $WindowsUptime
				SqlStartTime     = $SQLStartTime
				WindowsBootTime  = $WinBootTime
				SinceSqlStart    = $SQLUptimeString
				SinceWindowsBoot = $WindowsUptimeString
			}
```

alevyinroc:
Failed test was in a function not changed by this PR.

PR #2179, opened 2017-08-31
Testing/db instance paramater constructors

zippy1981:
## Type of Change

 - [ ] Pester test is included
 - [X] C# unit test is included

### Purpose
Add unit tests

### Approach
Its a unit test

### Notes

Two unit tests fail. I'm not sure if  var dbaInstanceParamater = new DbaInstanceParameter("[computerName\\sql2008r2sp2]"); is supposed to work or not. Please advise if it should.

On another PR my unit tests seem to break becasue AppVoyer seems to test agains `[host\instance]` instead of `host\instance`. Tests have passed in recent builds though. I'm not even sure if `[host\instance]` with the square brackets should be legit.

### DIFF
--- a/bin/projects/dbatools/dbatools.Tests/Parameter/DbaInstanceParamaterTest.cs
+++ b/bin/projects/dbatools/dbatools.Tests/Parameter/DbaInstanceParamaterTest.cs
@@ -29,30 +29,6 @@ namespace Sqlcollaborative.Dbatools.Parameter
         }
 
         /// <summary>
-        /// Checks that [localhost\instancename] is treated as a localhost connection
-        /// </summary>
-        [TestMethod]
-        public void TestBracketedName()
-        {
-            var dbaInstanceParamater = new DbaInstanceParameter("[computerName]");
-
-            Assert.AreEqual("computerName", dbaInstanceParamater.FullName);
-            Assert.IsTrue(dbaInstanceParamater.IsLocalHost);
-        }
-
-        /// <summary>
-        /// Checks that [localhost\instancename] is treated as a localhost connection
-        /// </summary>
-        [TestMethod]
-        public void TestBracketedNameWithInstance()
-        {
-            var dbaInstanceParamater = new DbaInstanceParameter("[computerName\\sql2008r2sp2]");
-
-            Assert.AreEqual("computerName\\sql2008r2sp2", dbaInstanceParamater.FullName);
-            Assert.IsTrue(dbaInstanceParamater.IsLocalHost);
-        }
-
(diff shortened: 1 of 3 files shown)


wsmelton:
Wouldn't have anything to do with the fact we use "localhost" as the host machine name would it?

And no with the brackets would not be a legit instance name.

Curious question though, what happened to `DbaInstance`? I have been able to use `DbaInstance` instead of `DbaInstanceParameter` before but see that is not out there anymore in dev branch.

zippy1981:
@wsmelton I don't think that was my doing. When did it break? Does it work on master currently?

Removed the failing tests. Things should work.

bin/projects/dbatools/dbatools.Tests/Parameter/DbaInstanceParamaterTest.cs
```diff
@@ -0,0 +1,67 @@
...
+        public void TestLocalhostIPConstructor()
+        {
+            var dbaInstanceParamater = new DbaInstanceParameter(IPAddress.Loopback);
+
+            Assert.AreEqual("127.0.0.1", dbaInstanceParamater.FullName);
+            Assert.IsTrue(dbaInstanceParamater.IsLocalHost);
+        }
+
+        /// <summary>
+        /// Checks that . is treated as a localhost connection
```
FriedrichWeinmann: Wrong description. You are testing ::1

bin/projects/dbatools/dbatools/Parameter/DbaInstanceParameter.cs
```diff
@@ -244,12 +244,12 @@ public DbaInstanceParameter(string Name)
             string tempString = Name;
 
             // Handle and clear protocols. Otherwise it'd make port detection unneccessarily messy
-            if (Regex.IsMatch(tempString, "^TCP:", RegexOptions.IgnoreCase))
+            if (Regex.IsMatch(tempString, "^TCP:", RegexOptions.IgnoreCase)) //TODO: Use case insinsitive String.BeginsWith()
```
FriedrichWeinmann: Why?

zippy1981: I'll write some code to prove it, but I'm pretty sure Regexes are a lot slower than simpple substring calls.

zippy1981: REgardless, I'll touch that later.

zippy1981:
Ok this can be merged!

PR #2186, opened 2017-09-02
DbaInstanceParameter - Added support for connection strings

FriedrichWeinmann:
## Type of Change

 - [ ] Pester test is included

## What it does

 - Adds support for connection strings as input to the `DbaInstanceParameter`. Must include a server target. The connection will use the connection string, other functions get their usual properties (ComputerName, InstanceName, Port, ...)
 - Adds support for sql notation server names (such as `[Server\Instance]`)
Braces used to fail interpretation.
 - Adds support for named pipe path notations (e.g.: `\\.\pipe\MSSQL$SQLEXPRESS\sql\query`)

### DIFF
--- a/bin/library.ps1
+++ b/bin/library.ps1
@@ -1,5 +1,5 @@
 ﻿# Current library Version the module expects
-$currentLibraryVersion = New-Object System.Version(0, 6, 1, 20)
+$currentLibraryVersion = New-Object System.Version(0, 6, 2, 21)
 
 <#
 Library Versioning 101:
--- a/bin/projects/dbatools/dbatools/Exceptions/BloodyHellGiveMeSomethingToWorkWithException.cs
+++ b/bin/projects/dbatools/dbatools/Exceptions/BloodyHellGiveMeSomethingToWorkWithException.cs
@@ -1,0 +1,71 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Sqlcollaborative.Dbatools.Exceptions
+{
+    /// <summary>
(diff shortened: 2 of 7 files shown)


wsmelton:
Verified brackets are parsed now...
image

Also successful with `[DbaInstance]"[server_01\instance1]"` and `[DbaInstance]"[server.01\instance1]"`. Confirmed same result if using the old parameter type `DbaInstanceParameter` as well.

image

wsmelton:
Testing the connection string...tried using `New-DbaSqlConnectionString` to actually generate the string for me but seems it has some extra stuff being added that (1) is not needed but (2) exceeds some character limit.

image

So stripped it down to only what I wanted and seemed to fail connection:

image

Maybe I'm misinterpreting what type of string you are supporting?

FriedrichWeinmann:
Actually, the parameterclass is working as designed ... just forgot to look into how it's implemented in `Connect-DbaSqlServer`
Run
`[DbaInstance]$string` to check out its interpretation.
Looking into where that function fails how.

FriedrichWeinmann:
Ah, mystery solved:
`Connect-DbaSqlServer` doesn't use `[DbaInstanceParameter]` so any refinements I add to the parameter class will not carry over to it. Gonna check out whether that's how it is supposed to be.

potatoqualitee:
Yes, you can add it back @FriedrichWeinmann - I took it out temporarily to make that command work with Named Pipes connections. When you add it back, that'll also give appveyor another chance to process the tests to ensure it all works.

potatoqualitee:
confirmed :( @FriedrichWeinmann whatever happened here made the tests infinitely wait/fail

image

potatoqualitee:
heck yah!

PR #2187, opened 2017-09-02
Find-DbaSimilarTable

SQLJana:
This is my first pull request. Please do be hard on me where I am not meeting team standards but also help me where it is necessary as this is my first online collaborative effort.


wsmelton:
@SQLJana there is no changes included in your PR (no files).

Just a general question though what is the purpose of the command you are submitting? What issue/problem is it trying to solve? Can you provide examples of how it would be used, with screenshots would also help?

SQLJana:
Shawn, Sorry that the PR did not have any files. I am having trouble making a PR to development branch from my local branch. Anyway, the function helps find tables/views that have a similar structure by matching column names as described in detail here - https://sqljana.wordpress.com/2017/03/31/sql-server-find-tables-with-similar-table-structure/

In the PowerShell version of the function, everything is optional except the InstanceName parameter in which case it shows all tables/views that have common column names with the matched column name percent. Given a table/view name, it helps identify other tables/views that have similar column names. This is helpful when you are trying to find the archive version of a table for example.

functions/Find-DbaSimilarTable.ps1
```diff
@@ -0,0 +1,247 @@
...
+
+            $totalcount = 0
+            $dbcount = $dbs.count
+            foreach ($db in $dbs) {
+                
+                Write-Message -Level Verbose -Message "Searching on database $db"
+                    
+                $props = 'ComputerName', 'InstanceName', 'SqlInstance', 'DatabaseName', 'MatchPercent', 'TableNameRankInDB', 'SchemaName', 'TableName', 'TableType', 'ColumnCount', 'MatchingDatabaseName', 'MatchingSchemaName', 'MatchingTableName', 'MatchingTableType', 'MatchingColumnCount'
+
+                $rows = $db.ExecuteWithResults($sql).Tables.Rows
```
wsmelton: We have a custom method added to `$server` and `$db` object so can be run as:

```powershell
$data = $db.Query($sql)
```
Your data object will have the data table returned so you can just reference each property as `$data.MatchPercent` and such.

functions/Find-DbaSimilarTable.ps1
```diff
@@ -0,0 +1,247 @@
...
+
+            Write-Message -Level Debug -Message $sql
+
+            $totalcount = 0
+            $dbcount = $dbs.count
+            foreach ($db in $dbs) {
+                
+                Write-Message -Level Verbose -Message "Searching on database $db"
+                    
+                $props = 'ComputerName', 'InstanceName', 'SqlInstance', 'DatabaseName', 'MatchPercent', 'TableNameRankInDB', 'SchemaName', 'TableName', 'TableType', 'ColumnCount', 'MatchingDatabaseName', 'MatchingSchemaName', 'MatchingTableName', 'MatchingTableType', 'MatchingColumnCount'
```
wsmelton: Just for readability please break this one up into multiple lines

e.g. 

```powershell
$props = 'ComputerName', 'InstanceName', 'SqlInstance', 'DatabaseName', 'MatchPercent', 'TableNameRankInDB', 'SchemaName', 'TableName',
'TableType', 'ColumnCount'.....
````

File: functions/Find-DbaSimilarTable.ps1 [hunk 3]
```diff
@@ -0,0 +1,247 @@
...
+
+                $rows = $db.ExecuteWithResults($sql).Tables.Rows
+
+                Select-DefaultView -InputObject $rows -Property $props
+
+                $vwcount = $vwcount + $rows.Count
+                $totalcount = $totalcount + $rows.Count
+                $everyservervwcount = $everyservervwcount + $rows.Count
```
wsmelton: For formatting purpose we use camelCase for variables in the functions. So `$everyservervwcount` would be written as `everyServerVwCount`.

SQLJana:
Shawn, thanks for the detailed review. I have incorporated all the recommendations. Can you please review again? It should work on every version from 2000 to date since the views existed since then. I have only tested on 2014 though.

Before, I had based most of my code on existing functions. Is there a good "reference" function that follows the group's best practices and has the most commonly used logic?

wsmelton:
We have a templates repo under the this org but I've had not a chance to a lot of examples of the various logic. One thing on my to do list.

SQLJana:
Ok.  Thank you. 

Please let me know if there is anything else you need. Hopefully, I this time around, I pushed the changes properly.

potatoqualitee:
Thank you for the PR @sqljana! And thank you for taking care of this, @wsmelton. Before merging, I'd like to confirm the name with the name master, @PowerDBAKlaas. Please allow me time to confirm.

wsmelton:
I would say something like `Find-DbaMatchingTable` or `Find-DbaIdenticalTable` would sound better than "Similar". I mean the synoposis and columns both reference `matching`.

wsmelton:
Identical may be too strong of a word to use...

SQLJana:
I agree. "Identical" is too strong. I like both "matching" and "similar". Also, should we say TableOrView instead of just Table?

PowerDBAKlaas:
Although ...TableOrView... may be more correct, I feel it will lead to strange names. It would also create a precedent. @SQLJana do you think this function is easily adapted to include nonclustered indexes or stored procedures? If not, I'd stick to ...Table...

@wsmelton maybe I don't grasp the semantics correctly, but 'matching' makes me think of regexp and meaning 'this and optionally more, but not less'

wsmelton:
So are you saying make it Find-DbaTable @powerdbaklaas?

PowerDBAKlaas:
That would be the nicest name, but not consistent with other `Find-` functions I believe. Don't de others have options for looking for keywords, or the owner or such things?
Also, will this function likely remain as is, or is there a possibility it will work on stored procedures or nonclustered indexes too?
For now, I think Find-DbaSimilarTable sounds as the best description to me.

potatoqualitee:
Agreed, I will create a test and merge 👍 with name Find-DbaSimilarTable

potatoqualitee:
Thanks again @SQLJana, love this command! I made one commit with 4 changes

1. added your website to the description 👍 
2. changed it so that -Database will work for system dbs without specifying an additional param
3. Added two columns that have the fully qualified table names, plus i moved the matchingpercent column up closer. This way, I find the info I need is immediately available.
4. Updated the column names to say "Original" for the columns that were not "Matching"

I also created a Pester test and will ping you on the PR 💃 This is not expected of new team members but thought you may be interested to see an integration test.

Welcome to the team! I'll be adding you to dbatools.ip/team shortly.

wsmelton:
@PowerDBAKlaas We can't call this command `Find-DbaTable` because it does not offer a search ability. It is comparing a base table to the rest of the tables in any database or all databases.

PowerDBAKlaas:
yes that too

BTW has there been any performance testing on this? Comparing all tables on the whole sqlinstance can take a few days, can't it?
Shouldn't there be a warning about that?

potatoqualitee:
It outputs as it goes along, so the person can just cancel if they get frustrated, eh?

SQLJana:
Thank you @wsmelton, @PowerDBAKlaas  and @potatoqualitee!

@PowerDBAKlaas, regarding "possibility it will work on stored procedures or nonclustered indexes too", it would be a difficult for stored procedures as what constitutes similarity could be different for different users (count by keywords, tables/views used, code similarity, return columns similarity etc). For NC indexes, using system views outside the INFORMATION_SCHEMA, it may not be hard.

Regarding performance testing, I did tests on small databases and it returns immediately. The one test that I did on a database with about 1000 tables and around 1000 views took 90 seconds for the raw SQL to return about 1 million rows in SSMS on an instance with 300 GB of memory (with no conditions from parameters). From PowerShell, the time taken on a client PC (SQL host is no a different server) for the same conditions is about the same time to start returning the rows..and the user can always cancel before it finishes returning everything as @potatoqualitee said (or) it may exhaust the memory.

PR #2189, opened 2017-09-03
New command Set-DbaAgHadr

### LINKED ISSUE #1712: Awesome Availability Groups Commands

wsmelton:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #1712 )
 - [x] Pester test is included

### Purpose
 
Stepping stone to getting additional AG commands around building an AG.
### Approach

Simply enables Hadr for the SQL Server instance. Nothing special.
### Commands to test

```powershell
# When Hadr is already disabled, should receive a warning as such.
Set-DbaAgHadr -SqlInstance Whatever\InstanceName -Enabled 0

# When Hadr is disabled and want to enable it, includes prompts to confirm
Set-DbaAgHadr -SqlInsance Whatever -Enabled 1

# Will have to manually stop/start services so...
Stop-DbaSqlService -SqlInstance Whatever -Instance mssqlserver -Type Engine, Agent
Stop-DbaSqlService -SqlInstance Whatever -Instance mssqlserver -Type Engine, Agent

# When Hadr is enabled and want to disable it, and also restart services to apply
Set-DbaAgHadr -SqlInstance Whatever\InstanceName -Enabled 0 -Force
```
### Screenshots

image

image

### Learning
- I was not able to make `Invoke-ManagedComputerComand` work for this instance. You can look at previous commits to see what I was doing wrong if you want it to use that command.
- Need to determine if I should output an object for this command, and if we do what should it include?


potatoqualitee:
hey @wsmelton - should this be Enable/Disable-DbaAgHadr? I think if a command *only* enables something, like with my Enable-DbaForceNetworkEncryption, that it should be an Enable/Disable command, along with a Get if it doesn't exist, because you will want to see its state.

@PowerDBAKlaas - what are your thoughts on my approach?

wsmelton:
You can enable it and disable by changing the state of the property `Enabled`, which the parameter name matches. So `Set-DbaAgHadr -Enabled 0` will disable it if not already and `Set-DbaAgHadr -Enabled 1` will enable it if not already.

wsmelton:
I don't agree with having two commands for enable/disable, otherwise what is the purpose of "set"? You are causing duplicate code that only has one minor difference between the two commands are you not?

wsmelton:
Also, the Get command already exist in the module as `Get-DbaAgHadr`.

potatoqualitee:
I'd be happy to see the Set command - can you Set anything else? If it's only Enable/Disable, that'd be harder to justify when the PS team approved the Enable/Disable verbs. (I'd prefer the Set but it has to do more>)

wsmelton:
Like what?

wsmelton:
I don't see how we can justify two commands that will literally be a copy of each other with exception of one character (0 or 1).

potatoqualitee:
I don't know enough about the technology to say. But I don't want a Set that can only enable/disable. The param is awkward without benefit. 

I imagine the Ps Team approved the verbs because ultimately, it would probably be switches for -Enable and -Disable instead of Enable $true. But that is also kind of awkward. 

Can you find any other Microsoft examples of Enable being used as a parameter instead of a command? I'd be interested in how others approached this because Enable disable will always be duplicated.

wsmelton:
The command that gave me the idea of using `-Enabled`

image

But does not matter one way or the other. I've split the command into two and created the test for each because I need this command to keep moving onto the other AG commands.

Side note, the Appveyor failure was because we hit the 1 hour mark so the testing stopped.

wsmelton:
To disable all the firewalls on Window the command would be: `Set-NetFirewallProfile -Profile Domain, Public, Private -Enabled False`

potatoqualitee:
LOL! Your commit message  ❤️ 

How weird that sometimes it just gets stuck! And on the same command - the Get Certificate. I'm gonna make sure there are no prompts with that one. 

Thank you for the commands and the tests! 🏆

PR #2198, opened 2017-09-05
[New Command] Get-DbaSqlPolicyManagement

Staggerlee011:
## Type of Change

 - [ ] Pester test is included

### Purpose
 
New function to return Polices from PBM

### DIFF
--- a/functions/Get-DbaPolicy.ps1
+++ b/functions/Get-DbaPolicy.ps1
@@ -1,109 +1,90 @@
 ﻿Function Get-DbaPolicy
 {
-<#
-.SYNOPSIS
-Returns polices from policy based management from an instance.
-
-.DESCRIPTION
-Returns details of policies with the option to filter on Category and SystemObjects.
-
-.PARAMETER SqlInstance
-Source SQL Server.You must have sysadmin access and server version must be SQL Server version 2008 or higher.
-
-.PARAMETER SqlCredential
-Allows you to login to servers using SQL Logins as opposed to Windows Auth/Integrated/Trusted. To use:
-
-$scred = Get-Credential, then pass $scred object to the -SourceSqlCredential parameter. 
-
-Windows Authentication will be used if DestinationSqlCredential is not specified. SQL Server does not accept Windows credentials being passed as credentials. 	
-To connect as a different Windows user, run PowerShell as that user.
-
-.PARAMETER Category
-Filters results to only show policies in the category selected
-
-.PARAMETER SystemObject
-By default system objects are filtered out. Use this parameter to INCLUDE them 
-
-.NOTES 
-Author: Chrissy LeMaire (@cl), netnerds.net
-Requires: sysadmin access on SQL Servers
-
-dbatools PowerShell module (https://dbatools.io, clemaire@gmail.com)
-Copyright (C) 2016 Chrissy LeMaire
-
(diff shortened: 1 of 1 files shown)


potatoqualitee:
🍳 cellent work, my friend. I will merge this after I create a test for it.

functions/Get-DbaPolicy.ps1
```diff
@@ -0,0 +1,100 @@
+﻿Function Get-DbaPolicy {
```
wsmelton: This one will need to be renamed to `Get-DbaSqlPolicyManagement` unless @potatoqualitee has another preference. This is the noun we have for the Copy command.

functions/Get-DbaPolicy.ps1
```diff
@@ -0,0 +1,100 @@
...
+		[switch]$IncludeSystemObject,
+		[switch]$Silent
+	)
+	
+	process {
+		foreach ($instance in $SqlInstance) {
+			Write-Message -Level Verbose -Message "Attempting to connect to $instance"
+			
+			try {
+				$server = Connect-SqlInstance -SqlInstance $instance -SqlCredential $SqlCredential
```
wsmelton: Need to add `-MinimumVersion` parameter to the connect since this is not available in all versions.

functions/Get-DbaPolicy.ps1
```diff
@@ -0,0 +1,100 @@
...
+			}
+			catch {
+				Stop-Function -Message "Failure" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
+			}
+			
+			$sqlStoreConnection = New-Object Microsoft.SqlServer.Management.Sdk.Sfc.SqlStoreConnection $server.ConnectionContext.SqlConnectionObject
+			
+			# DMF is the Declarative Management Framework, Policy Based Management's old name
+			$store = New-Object Microsoft.SqlServer.Management.DMF.PolicyStore $sqlStoreConnection
```
wsmelton: Likely would not hurt to wrap line 74 and 77 in a try/catch, just in the event SMO version does not support PBM. Maybe just have the `Stop-Function` output a message as such, or think we have a function that we can verify what version of SMO is loaded.

functions/Get-DbaPolicy.ps1
```diff
@@ -0,0 +1,100 @@
...
+			
+			$allpolicies = $store.Policies
+			
+			if (!$IncludeSystemObject) {
```
wsmelton: Please add a comment on this line to just note why you check for `-not` on the parameter...assuming because system objects are included by default?

functions/Get-DbaPolicy.ps1
potatoqualitee: I was wondering about this because the Copy copies over the whole suite but this is just the policy and returns `Microsoft.SqlServer.Management.Dmf.Policy` So i think we'll keep this name

functions/Get-DbaPolicy.ps1
potatoqualitee: Very nice and true! thanks, I'll add

functions/Get-DbaPolicy.ps1
potatoqualitee: Yeah, this is one of the included params. I think the comment would be unusual as we don't comment in other commands when we respond to parameters?

functions/Get-DbaPolicy.ps1
potatoqualitee: done-zo

PR #2212, opened 2017-09-09
Switching Tepp default mode to synchronous updates

FriedrichWeinmann:
## Type of Change

 - [ ] Pester test is included

### Purpose
Fixes complaints about persistent connections due to asynchronous update of TEPP data.
This is accomplished by making the synchronous mode the default mode.

### DIFF
--- a/internal/Connect-SqlInstance.ps1
+++ b/internal/Connect-SqlInstance.ps1
@@ -35,21 +35,11 @@ function Connect-SqlInstance {
 	
 	[CmdletBinding()]
 	param (
-		[Parameter(Mandatory = $true)]
-		[object]
-		$SqlInstance,
-		
-		[object]
-		$SqlCredential,
-		
-		[switch]
-		$ParameterConnection,
-		
-		[switch]
-		$RegularUser = $true,
-		
-		[int]
-		$MinimumVersion
+		[Parameter(Mandatory = $true)][object]$SqlInstance,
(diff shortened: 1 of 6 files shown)


potatoqualitee:
@FriedrichWeinmann thanks so much for the patch! Unfortunately, all tests failed hard.

FriedrichWeinmann:
Victory :trophy:

internal/Connect-SqlInstance.ps1
```diff
@@ -36,15 +36,51 @@ function Connect-SqlInstance {
...
+		$SqlCredential,
+		
+		[switch]
+		$ParameterConnection,
+		
+		[switch]
+		$RegularUser = $true,
+		
+		[int]
+		$MinimumVersion
```
wsmelton: I'm sure this the result of the editor utilized, but just noting the previous format is how we have modified commands to be setup as...meaning commands that had the type and parameter name on different lines is being put back to single lines in all the functions.

If we want to stay with two lines we need to do this across all of them. Just saying :man_shrugging:

FriedrichWeinmann: Heya,
indeed, it was the result of my editor and I didn't scroll high enough to catch it.
Global policy is to keep things as single-line as possible (Chrissy prefers that, especially for one of our mega-functions with zillions of parameters). Gonna fix that before merging.

Issue #2217, opened 2017-09-11
Replace use of Detailed or Simple parameters


wsmelton:
To follow standards of current commands any command that currently utilizes `-Detailed` and/or `-Simple` should be replaced to utilize `Select-DefaultView` internal function call. So this would mean the "simple" output would generally become the default properties to output. Then the detailed, if applicable would be obtained by piping to `select *`. In cases where a command only has a detailed and it already utilizes the default view function, just remove the code for the parameter.

Rough draft list of commands found to still have these parameters in use:
_search done via VS Code for text `PARAMETER Detailed`_
- [ ] Find-DbaAgentJob
- [x] Find-DbaDatabase
- [x] Get-DbaClusterActiveNode
- [x] Get-DbaDiskSpace
- [ ] Get-DbaRunningJob.ps1 
- [x] Get-DbaSpConfigure
- [ ] Get-DbaTcpPort
- [x] Repair-DbaServerName
- [x] Test-DbaConnectionAuthSchema
- [x] Test-DbaDatabaseCollation
- [x] Test-DbaDatabaseCompatibility
- [x] Test-DbaDatabaseOwner
- [x] Test-DbaDiskAlignment
- [x] Test-DbaDiskAllocation
- [x] Test-DbaFullRecoveryModel
- [x] Test-DbaJobOwner
- [x] Test-DbaMaxDop
- [x] Test-DbaPowerPlan
- [x] Test-DbaServerName
- [x] Test-DbaTempdbConfiguration
- [x] Test-DbaValidLogin
- [x] Test-DbaVirtualLogFile

Determined by @potatoqualitee and @wsmelton  that this parameter stays
_search done via VS Code for text `PARAMETER Simple`_
~~- [ ] Get-DbaMemoryUsage~~
~~- [ ] Get-DbaStartupParameter~~
~~- [ ] Read-DbaBackupHeader~~
~~- [ ] Test-DbaLogShippingStatus~~

### important [Updated]

The following block needs to be added to the `begin{}` block of the command, if there is not already a begin block then just add it.

```powershell
begin {
    Test-DbaDeprecation -DeprecatedOn 1.0.0 -Parameter Detailed
}
```

Due to how we handle test for parameters we need to make the following adjustments to the commands:
- Alias the Detailed parameter, if possible
- If you cannot alias the parameter then leave it in the param block.
- Update parameter in the help documents to this:
```
.PARAMETER Detailed
[...]

PR #2218, opened 2017-09-11
Update to Test-DbaTempDbConfiguration

### LINKED ISSUE #2165: Test-DbaTempDbConfiguration suggestion for change
## Feature Request

 Simply answer the question: What problem or scenario would you like to solve with dbatools?

I'm currently working on setting up a pester test for our SQL servers. In this I use Test-DbaTempDbConfiguration to figure out if we are following best practies. I have som cases where it has been decided to use fewer tempdb files than is best practies. Since the command returns a string as output when it's OK and a PScustomobject when its not its hard to write a good pester test. 

Suggestion is that also make the output on success a PSCustomObject that follow the same standard as the error output. The string that is the output today can be moved to the notes.

This way I can work on the pscustombject and check IsBestpractises, Recomended or currentsettings to see if I follow our setup on the sql server.

wsmelton:
## Type of Change

 - [x] Address #2217 
 - [x] Breaking change fixes #2165 
 - [ ] Pester test is included

### Purpose
To bring command up to the standard of other commands and output object no matter what, not just an output if IsBestPractice returns all true. As well have default output, adjusted to just require SELECT * to get that notes property.

### Approach
- Removed detailed parameter
- Updated help
- changed servername to instance on foreach
- added -MinimumVersion 9 to connect command, dropped version check
- updated ExecuteQueryWithResults to use Query() method
- Changed IsBestpractice to IsBestPractice
- Changed T-SQL to get file info for tempdb to use Get-DbaDatabaseFile
- Changed write-verbose to write-message
- Added Silent parameter to param block and help
- changed output to have basic properties, use select * to see the notes property
- removed code from end block so it outputs full object

### Screenshots

image

image

Previous version of the command only output those that were false for the `IsBestpractice`

image

### DIFF
--- a/functions/Test-DbaTempDbConfiguration.ps1
+++ b/functions/Test-DbaTempDbConfiguration.ps1
@@ -5,7 +5,7 @@ function Test-DbaTempDbConfiguration {
 
 		.DESCRIPTION
 			Evaluates tempdb against a set of rules to match best practices. The rules are:
-			
+
 			* TF 1118 enabled - Is Trace Flag 1118 enabled (See KB328551).
 			* File Count - Does the count of data files in tempdb match the number of logical cores, up to 8?
 			* File Growth - Are any files set to have percentage growth? Best practice is all files have an explicit growth value.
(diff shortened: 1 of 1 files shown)


potatoqualitee:
ouuu this is nice, i'll merge and write a test once tests have passed

wsmelton:
I was contemplating the test part while I was writing up the PR. Based on the other "tests" files for the Test commands there would be some mocking required for this one?

potatoqualitee:
I suppose, though I think there could be integration tests with real testing as well 👍

potatoqualitee:
added test :monkeydance:

PR #2245, opened 2017-09-18
Adjust Watch-DbaDbLogin

wsmelton:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [x] Documentation
 - [ ] Build system
 

### Purpose
 
Cleanup and standardize code to use commands available in module, among other things

- removed default Database name (this needs to be specified)
- removed SqlCmsGroups from help as it didn't exist in the code
- Updated default table name to be more descriptive on it being dbatools related, updated help as well on this
- changed Database parameter to a single object
- Updated message system
- Adjusted validation of SqlCms and ServersFromFile parameter using Test-Bound
- Replaced datatabe object  and bulkcopy object with Out-DbaDatabase and Write-DbaDataTable
- Changed enumprocess() to DMV, can expand data in the future now if we need to.
- Changed code for CMS pull to Get-DbaRegisteredServer
- Removed comment block code in end block
- Cleaned up logic where it will not try to run code if CMS or file is not found.

### Approach

Looked at adding more columns but only ended up adding a few extra to show pertinent information and prevent the need of having the table created first. So on the first run, this command will now create the table with the needed column names.

### Screenshots

image

The previous version would continue running code if CMS or file didn't exist, this has been fixed

image

image

image

Captured data:

image

### Learning
EnumProcess() runs DMVs that do not exist in 2000 so added `-MinimumVersion 9`.

One additional tidbit discovered by tracing the `EnumProcess()` method that the query uses the `is_user_process` column found in `sys.dm_exec_sessions` but actually reverses the meaning of the column. Default this will show a 1 if the session is a user session or 0 (zero) if system. The method calls it in this manner: `CAST(~s.is_user_process AS bit) AS [IsSystem]` which changes that to return 0 (zero) when it is a user session and a 1 if is not.

### DIFF
--- a/functions/Watch-DbaDbLogin.ps1
+++ b/functions/Watch-DbaDbLogin.ps1
@@ -78,12 +78,14 @@ function Watch-DbaDbLogin {
 		[string]$SqlCms,
 		# File with one server per line
 
-		[string]$ServersFromFile
+		[string]$ServersFromFile,
+		[switch]$Silent
 	)
 
 	process {
-		if ([string]::IsNullOrEmpty($SqlCms) -and [string]::IsNullOrEmpty($ServersFromFile)) {
-			throw "You must specify a server list source using -SqlCms or -ServersFromFile"
+		if (Test-Bound 'SqlCms','ServersFromFile' -Not) {
+			Stop-Function -Message "You must specify a server list source using -SqlCms or -ServersFromFile"
+			return
 		}
 
 		<#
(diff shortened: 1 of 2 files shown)


wsmelton:
Added test and adjusted command to also include the instance information, we were not including that in the table.

image

image

wsmelton:
@potatoqualitee that is an issue with Appveyor because this test runs fine against my 2008 and 2016 instance (see screenshot provided in previous comment). I can just remove the test if you'd like?

PR #2269, opened 2017-09-20
Buildreference, added latest releases

niphlod:
added CU8 for 2016 RTM and CU5 for 2016 SP1


potatoqualitee:
ur the best 🚎 thanks for keeping up with this

PR #2271, opened 2017-09-20
Close connection in Get-DbaRegisteredServer.

galador:
Because the connection is passed to the RegisteredServersStore object, it does not get closed after use, and stays alive on the SQL Server.

## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Clean up the connection in `Get-DbaRegisteredServer`. Because we pass it to another object (`RegisteredServersStore`), this prevents the connection from auto-closing and being released back to the pool.

You can verify on the current version by running the following (or comment out [line 151](https://github.com/sqlcollaborative/dbatools/pull/2271/files#diff-ddbe25c86fa572357b14f4bcc64435e9R151) if you pull in this change) in a fresh PowerShell window:

```powershell
for ($i=0; $i -lt 50; $i++) {
	$servers = Get-DbaRegisteredServer -SqlInstance sqldbatools -Group "Group Name"
}
```

Then run this query on the instance to see the sleeping threads:

```sql
SELECT c.session_id, c.connect_time, c.num_reads, c.num_writes, 
	s.host_name, s.program_name, s.cpu_time,
	S.last_request_start_time, s.last_request_end_time
FROM sys.dm_exec_connections c
INNER JOIN sys.dm_exec_sessions s
ON s.session_id = c.session_id
WHERE program_name LIKE 'dbatools PowerShell module%'
ORDER BY last_request_start_time DESC
```

You will see roughly 50 connections (one for each loop, and possibly a few for the Tab Expansion processes).

Then, running the same test harness with the fix in place, you should only see one thread (and maybe one for TEPP).

### DIFF
--- a/functions/Get-DbaRegisteredServer.ps1
+++ b/functions/Get-DbaRegisteredServer.ps1
@@ -146,6 +146,9 @@ function Get-DbaRegisteredServer {
 			if ($ExcludeCmsServer) {
 				$servers = ($servers | Where-Object { $_.ServerName -ne $instance})
 			}
+
+			# Close the connection, otherwise using it with the ServersStore will keep it open
+			$cmsStore.ServerConnection.Disconnect()
 		}
 		
 		foreach ($server in $servers) {


potatoqualitee:
@FriedrichWeinmann - is this a TEPP issue? @galador - I've actually seen .Close() not really close until the pool was emptied. Does this fa sho work?

FriedrichWeinmann:
TEPP will use a single connection per instance if ran asynchronously, which will auto-release after a short period. So, no, TEPP isn't the culprit here. Next update will explicitly release connections as far as the pool manager respects that.

galador:
Well, I'm not 100% sure that the TEPP threads I'm seeing aren't related to another bug.

BUT. Without this change, `Get-DbaRegisteredServer` definitely leaves a bunch of threads out on the SQL Server until you close the PowerShell window (or 💣 the pool).

potatoqualitee:
craazaaay! thank you for investigating and addressing, merged 🥇

PR #2275, opened 2017-09-21
New function Get-DbaSuspectPages

gbargsley:
## Type of Change

 - [X ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Screenshots

image


functions/Get-DbaSuspectPages.ps1
```diff
@@ -0,0 +1,98 @@
...
+					END as EventType, 
+					error_count, 
+					last_update_date
+					from msdb.dbo.suspect_pages"
+				
+			try {
+				$results = $server.Query($sql) 
+			}
+			catch {
+				Stop-Function -Message "Issue collecting data on $server" -Target $server -ErrorRecord $_
```
FriedrichWeinmann: When using Stop-Function inside the process block's foreach-loop, it's almost always necessary to specify `-Continue`
That will make it skip the current instance but continue on with the next.
Also `-Target $instance` would be a neat debugging enhancement::
`Stop-Function -Message "Issue collecting data on $server" -Target $server -ErrorRecord $_ -Target $instance -Continue`

potatoqualitee:
great lil command! 🏆 🏅 🌮 ! Please rename to the singular, both the command and the filename. Also, try another instance? whichever one isn't allowing inserts into msdb.

gbargsley:
@potatoqualitee The research in the issue that @wsmelton listed suggested keeping it plural as that is the way it is listed in msdb.

gbargsley:
Anyone have an idea why my test cannot complete?  I am not sure how to troubleshoot the tests.

potatoqualitee:
$server.Databases is plural as well but we keep it singular. It's wayyy rare that we'll use plural and this is not a valid case. I'll take a look at the test once the rename is complete.

gbargsley:
I made the changes.  And took out some parts of the test that are not needed and could be the cause of failure.

potatoqualitee:
thank you! looks like the query is still messing up :( I'll look more into it tomorrow

potatoqualitee:
oops wait, that was based off of an older commit! We'll see if the other ones work

potatoqualitee:
fixed the test and updated a column name - the reason the test was failing mayyyyyybe was because it was trying to count a column that didn't exist. just in case instance1's msdb was the issue, i also updated it to instance2

potatoqualitee:
Woo! Checks pass! Great pR!!

PR #2291, opened 2017-09-22
Fixing SSIS service check in Copy-DbaSsisCatalog

### LINKED ISSUE #2278: Copy-DbaSsisCatalog fails when only one SSIS service is running on destination server

- [x] Verified running latest release of dbatools?  (Downloaded and installed on 9/20/2017 via 
`
Invoke-Expression (Invoke-WebRequest -UseBasicParsing https://dbatools.io/in)
`

### Version Information
 - Operating System (Windows Server 2012 R2): 
 - PowerShell Version: 

```
PS C:\Windows\system32> $PSVersionTable.PSVersion

Major  Minor  Build  Revision
-----  -----  -----  --------
4      0      -1     -1
```

 - SQL Server (SQL 2016|13.0.4451): 

## Steps to Reproduce

I only have one SSIS instance running, it appears that the `Copy-DbaSsisCatalog` function calls the `Get-RemoteIntegrationService` function... and `Get-RemoteIntegrationService` calls `Get-DbaSqlService -ComputerName $Computer | Where-Object Servicename -like "MsDts*"` ... and it is erroring out when `$result.Count -gt 0` fails (line 113 on https://github.com/sqlcollaborative/dbatools/blob/master/functions/Copy-DbaSsisCatalog.ps1)

I think line 113 fails because only one object is returned, and thus it isn't a collection? (And thus the .Count property is empty?)

Changing line 113 solved it for me:

```
#Old Line 113
if ($result.Count -gt 0) {
#New Line 113:
if ($result) {
```

 - [x] Attach any screenshots (if possible/allowed)
Screenshot of error:
image

Screenshot of related lines from Copy-DbaSsisCatalog:
image

### Problem to solve
I think the `$result = Get-DbaSqlService -ComputerName $Computer | Where-Object Servicename -like "MsDts*"` call might need to be changed to `Get-DbaSqlService -Type SSIS` (however, my testing shows the same issue).

In the end, it appears that `$result = Get-DbaSqlService -ComputerName $Computer | Where-Object Servicename -like "MsDts*"` has a value, the the .count property is empty when $result is only one object.

nvarscar:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #2278 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
To fix incorrect handling of the array of objects in the code, which can sometimes be just an object

### Approach
removed .Count property

### Commands to test
Copy-DbaSsisCatalog -Source src -Destination dst

### DIFF
--- a/functions/Copy-DbaSsisCatalog.ps1
+++ b/functions/Copy-DbaSsisCatalog.ps1
@@ -109,8 +109,8 @@ Deploy entire SSIS catalog to an instance without a destination catalog.  Passin
             param (
                 [Object]$Computer
             )
-	    $result = Get-DbaSqlService -ComputerName $Computer | Where-Object Servicename -like "MsDts*"
-            if ($result.Count -gt 0) {
+	    $result = Get-DbaSqlService -ComputerName $Computer -Type SSIS
+            if ($result) {
                 $running = $false
                 foreach ($service in $result) {
                     if (!$service.State -eq "Running") {


potatoqualitee:
Thank you! 🍰

PR #2312, opened 2017-09-25
Get-DbaAgentJobOutputFile, kill NetBios

niphlod:
plus tests ^_^

## Type of Change

 - [x] Bug fix (non-breaking change, fixes "Every time a dba uses NetBios after 1998, there is a network admin somewhere that falls down dead")
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Adding tests to the function. Also, we should start addressing the fact that NO SMO properties return the FQDN of the server unless Resolve-DbaNetworkName is called. 
IMHO
- ComputerNamePhysicalNetBIOS should ALWAYS discarded
- NetName holds the machine name, which is more useful than Name if you're targeting a cluster
- all of them have no knowledge of FQDN, which, if you ask me, is a HUGE miss on SMO's side

That being said, NetName is surely more useful than ComputerNamePhysicalNetBIOS when you're trying to target the underlying server, if you don't wanna impose the "burden" of Resolve-DbaNetworkName

### DIFF
--- a/functions/Get-DbaAgentJobOutputFile.ps1
+++ b/functions/Get-DbaAgentJobOutputFile.ps1
@@ -127,7 +127,7 @@
                             Job                  = $j.Name
                             JobStep              = $Step.Name
                             OutputFileName       = $Step.OutputFileName
-                            RemoteOutputFileName = Join-AdminUNC $Server.ComputerNamePhysicalNetBIOS $Step.OutputFileName
+                            RemoteOutputFileName = Join-AdminUNC $Server.NetName $Step.OutputFileName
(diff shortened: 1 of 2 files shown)


potatoqualitee:
agreed, solid change thank you. will merge once tests pass, thanks for test!

niphlod:
I promised tests a long ago, they were long due. Going to fix them.... damn pester discrepancies in handling arrays....

potatoqualitee:
🍳 cellent! And yes, i need to add that .Tests to all of them 👎

PR #2332, opened 2017-09-27
New command - Update-DbaSqlServiceAccount

### LINKED ISSUE #2120: New command to change service account using SQL WMI
http://sqlity.net/en/2368/service-master-key/

"However, if you use the SQL Server Configuration Manager to change the service account, it will automatically take care of decrypting the service master key before the account change and re-encrypting it afterwards. This is another reason, why you always want to use that tool when changing SQL Server service level settings."

nvarscar:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #2120 )
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Configuring SQL Server service accounts using dbatools

### Approach
Utilizing SetServiceAccount and ChangePassword methods of the ManagedComputer.Services collection.

### Commands to test
See examples/pester

### Screenshots
image


potatoqualitee:
that detach test is bitchy - im going to disable it for now

potatoqualitee:
Thanks so much, btw! Looking forward to adding this - should be tonight, after I disable that test 🌮

nvarscar:
No prob, turns out that SetServiceAccount actually takes care of the service master key as well, so it was pretty easy to just wrap it with some digital noise and write a test 😄

potatoqualitee:
Service master key and permissions! And probably other stuff too. It's awesome!

nvarscar:
It still took around a minute or two to complete the test 🤢 Let me know if that's too much, I'll remove SQL Server engine from the test keeping only agent changes - it should cut the execution time at least in half - or even more!

potatoqualitee:
We can keep as is until we decide against it 👍 💯 Helllllllll yeahhhhh! thanks for another awesome command!

PR #2336, opened 2017-09-27
Dismount test - telling pester to skip this problematic test

potatoqualitee:


niphlod:
why is pester.groups so hard to use ?

PR #2337, opened 2017-09-27
tests, how to avoid run in appveyor

niphlod:
while everyone can still test on their own test rigs

## Type of Change
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [x] Build system
 

### Purpose
appveyor is useful, but sometimes not cooperative. Instead of fiddling with the source code of a completely working test OUTSIDE appveyor, slipping it in pester.groups.ps1 'appveyor_disabled' makes it not running just in appveyor (until an inspired soul gets it working again)
Let's try to think tests we write CAN (and should) be run in every dba test rig, not only on appveyor ... the big picture is that for getting coverage for all things we say we support we'd need more than 500 envs (factor sql releases, sql editions, x86/x64, AD/DNS infrastructure, OS versions, local/remote launch, clusters infrastructure, AGs, etc)... most of them won't ever be possible in appveyor ^_^ 

### Approach
when you -Skip, you disable the test for everyone. beforeall and afterall, and all the commands not in the skip "stanza" gets executed no matter what. 
If all tests are skipped, why should appveyor bother to run the entire test ? ù
I'll document better and improve appveyor env and tests "environments" (pinky promise), and I have plans for additional helpers. For now, let's start using this "on/off" switch when it's needed.


potatoqualitee:
perfect thank you 📟

PR #2367, opened 2017-10-02
lets see what happens if we remove setters

zippy1981:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
I removed these empty setters. I think @FriedrichWeinmann said they needed to be there, but all the unit tests passed so I'm not sure why.

### Approach
Deleting code is always awesome!

### Commands to test
Ok I'm not actually sure what uses Size. If someone whats to tell me what uses it I'll see about adding coverage.

### DIFF
--- a/bin/projects/dbatools/dbatools/Utility/Size.cs
+++ b/bin/projects/dbatools/dbatools/Utility/Size.cs
@@ -22,10 +22,6 @@ namespace Sqlcollaborative.Dbatools.Utility
             {
                 return (Byte / (double)1024);
             }
-            set
-            {
-
-            }
         }
 
         /// <summary>
@@ -36,10 +32,6 @@ namespace Sqlcollaborative.Dbatools.Utility
             get
             {
                 return (Byte / (double)1048576);
-            }
-            set
-            {
-
             }
         }
 
@@ -52,10 +44,6 @@ namespace Sqlcollaborative.Dbatools.Utility
             {
                 return (Byte / (double)1073741824);
             }
-            set
-            {
-
-            }
         }
 
         /// <summary>
@@ -66,10 +54,6 @@ namespace Sqlcollaborative.Dbatools.Utility
             get
             {
                 return (Byte / (double)1099511627776);
-            }
-            set
-            {
-
             }
         }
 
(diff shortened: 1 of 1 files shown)


FriedrichWeinmann:
Well ... they were needed for a future feature I'm planning on implementing.
It's not there yet, so obviously all tests will pass.

Specifically, they are needed for PowerShell-integrated automatic serialization. Setting that up fully would allow us to:
 - Retain object types across process boundaries (for example using jobs).
 - Store to filesystem and retrieve it from there while maintaining type fidelity.
 - Scale out operations across multiple computers in a powershell grid without loosing types in transit.

Serialization will fail if any public property can't be set without an exception (nevermind whether the process actually changes a byte).
That's why I keep them on any type I implement that is designed to be part of an output object (such as size, which will become a string using default serialization).
If it irritates you so much that you want to kill them - for example if your resharper complains about them - feel free to kill them though. Not many parts of the library are subject to that and I can add them back when it becomes necessary (or find a way around the requirement, which would be my preference anyway).
Besides, we mostly re-use SMO objects currently, which is a dead letter where serialization is concerned, no matter what we do.

potatoqualitee:
I will leave this to you @FriedrichWeinmann

bin/projects/dbatools/dbatools/Utility/Size.cs
```diff
@@ -118,7 +102,7 @@ public override string ToString()
             }
             if (Byte == -1)
                 return "Unlimited";
-            return "";
+            throw new ArgumentOutOfRangeException("Byte", string.Format("Invalid value of {0} for Byte", Byte));
```
FriedrichWeinmann: Don't want to flood `$error` with irrelevant errors, `""` was intentional

potatoqualitee:
@zippy1981 please discontinue the abandonment of your PRs or I shall close them with the heat of 1000 suns 🌞

zippy1981:
@potatoqualitee well actually, unless your running Visual Studio on the USS-Enteprise-D, your probably going to close it with 100-200 watts of electricity, which is way smaller than a sun.

FriedrichWeinmann:
And with that it has my blessings :)

PR #2377, opened 2017-10-03
Fix bug in Show-DbaDatabaseList.ps1 - use $SqlCredential instead of $…

eratolekov:
Use $SqlCredential instead of $SourceSqlCredential

## Type of Change
 
### Purpose
Show-DbaDatabaseList does not work properly with parameter -SqlCredential.


potatoqualitee:
Awesome catch! Thanks so much @eratolekov

PR #2385, opened 2017-10-04
Library - Added build configurations

FriedrichWeinmann:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Purpose & Change
Updated the way the library is being built-on-demand, as configuration specific issues have kept coming up:
 - Only appveyor will build the tests library (A dependency would cause issues)
 - Users will build the release configuration instead, which only contains the dbatools library itself
 - Users are specifically pointed at the configuration to build

Should fix: https://github.com/sqlcollaborative/dbatools/issues/2378#issuecomment-334176115

### DIFF
--- a/bin/library.ps1
+++ b/bin/library.ps1
@@ -90,10 +90,12 @@ if ($ImportLibrary) {
 				$start = Get-Date
 				$system = [Appdomain]::CurrentDomain.GetAssemblies() | Where-Object FullName -like "System, *"
 				$msbuild = (Resolve-Path "$(Split-Path $system.Location)\..\..\..\..\Framework$(if ([intptr]::Size -eq 8) { "64" })\$($system.ImageRuntimeVersion)\msbuild.exe").Path
-				$msbuildOptions = "/p:Configuration=Release"
+				$msbuildConfiguration = "/p:Configuration=Release"
+				$msbuildOptions = ""
 				if ($env:APPVEYOR -eq 'True') {
 					# This doesn't seem to work. Keep it here for now
-					$msbuildOptions = '/p:Configuration=Debug /logger:"C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"'
+					$msbuildOptions = '/logger:"C:\Program Files\AppVeyor\BuildAgent\Appveyor.MSBuildLogger.dll"'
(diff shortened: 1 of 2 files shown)


potatoqualitee:
heyyy it passed!

FriedrichWeinmann:
@zippy1981 would you mind going over these changes?
I've configured the two default configurations, having appveyor build debug and everybody else build release config, with release config not building the tests library (which was causing some trouble for some installations with resolving a dependency)

potatoqualitee:
thank you for the review @zippy1981 🙇

FriedrichWeinmann:
Heya @zippy1981 , care to state your case _why_ you would call the csproj directly?
I can't really see the benefit in that, compared to yet another adding to the complexity of this script.

potatoqualitee:
@zippy1981, my #1 🌻 . Please holla back.

zippy1981:
@FriedrichWeinmann Why would I? Reason zero, is because I remember a time when msbuild could only  build csproj files and not sln files. So that's my prejudice.

Reason #1, well quite frankly, that's the project we want to build, and as everything stands, its simpler to just call that from msbuild because it has no dependencies.  If it depended on other csproj files, I'd actually say make a third target in the SLN explicitly for being built as part of the module loading process.

Honestly, we're building a DLL to be called by powershell, and have to eventually solve the problem of compiling this in .net 4.5 instead of .net 4.0 on a server without Visual Studio on the windows SDK. Whatever voodoo solves that problem will mean revisiting this. But calling msbuild on the sln or the csproj is ultimately six of one or half a dozen of another.

FriedrichWeinmann:
Alright, in that case I'll merge and close this PR. It's no technical impediment for now and will be a non-issue in the close future. This way we get out of Chrissy's clean-slate plan for PASS for now.

The Voodoo is called TSFS and I'll finish my work on this once I'm done with my current hot potato issue. Basically, on all PRs to development it will automatically recompile the DLL in 4.5. Worked excellent in trial runs, but the devil lurks in the details. Somewhen probably still this month is the best I can give you for now.

zippy1981:
@FriedrichWeinmann You mentioned that strategy, and I think its an excellent one.

PR #2392, opened 2017-10-05
New-Command Invoke-DbaBalanceDataFiles

sanderstad:
Initial commit to enable the option to balance out data between multiple data files.

### DIFF
--- a/functions/Invoke-DbaBalanceDataFiles.ps1
+++ b/functions/Invoke-DbaBalanceDataFiles.ps1
@@ -8,12 +8,16 @@ function Invoke-DbaBalanceDataFiles {
 		When you have a large database with a single data file and add another file, SQL Server will only use the new file until it's about the same size.
 		You may want to balance the data between all the data files.
 
-		The command will check the server version and edition to see if the it allows for online index rebuilds.
+		The function will check the server version and edition to see if the it allows for online index rebuilds.
 		If the server does support it, it will try to rebuild the index online.
 		If the server doesn't support it, it will rebuild the index offline. Be carefull though, this can cause downtime
 
 		The tables must have a clustered index to be able to balance out the data.
-		The command does NOT yet support heaps.
+		The function does NOT yet support heaps.
+
+		The function will also check if the file groups are subject to balance out.
+		A file group whould have at least have 2 data files and should be writable.
+		If a table is within such a file group it will be subject for processing. If not the table will be skipped.
 
 	.PARAMETER SqlInstance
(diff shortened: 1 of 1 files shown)


sanderstad:
image

potatoqualitee:
hey sander, thanks for the PR! Please rely less on Write-Message Output (Turn those to verbose) and create a new [pscustomobject] that's basically

-- start of script --
$startingsizes = Get-DbaDatabaseFile whateverwhatever

-- do your fancy footwork --
$endingsize = Get-DbaDatabaseFile whateverwhatever

then output

````
[pscustomobject]@{
     ComputerName = $server.ComputerName ... (and so on)
     Database
     StartingX = $startingsizes.X
     StartingY = $startingsizes.Y
    EndingX = $endingsizes.X
     EndingX = $endingsizes.Y
}

Also, please add help for Confirm , Whatif and RebuildOffline

Stuart-Moore:
Really needs to check for multiple files in the same file group. Could have
multiple file groups with one file each which would currently pass the
check but fail with the rewrite.

Could ado with checking space in files, at least double the original table,
half of which needs to be in each file or you're back where you started
with unbalanced files

Also a warning that you're about to generate a lot of log. Probably alert
if table larger than x

Bare in mind people using this will probably be worried about large tables
rather than smaller ones. So there'll be a lot of IO.

potatoqualitee:
Thanks so much, @Stuart-Moore !

potatoqualitee:
This is looking so good, Sander! Did it address all of the points Start made?

Also, can you send me the code that you used to test (from creating the db and clustered indexes?) That way, I can write an integration test. Or if you would prefer, that is good too. I want it to be as easy as you need.

Stuart-Moore:
Yep, it does. We've been going over some bits on slack, and think we've covered all of them.

Just wanted to make sure people don't shoot themselves in both feet,  and being up front about limitations from the start (even doing it manually isn't going to get a pure 50% split without some tuning to each case)

potatoqualitee:
awesome, thanks stuart! i just dm'd sander, he'll add one more check and then I'll merge 👍

potatoqualitee:
Hello @sanderstad - can you please add your last change so we can get this bad boy merged? Danke!

sanderstad:
Added the free space check 
image

potatoqualitee:
@niphlod can you please add the two failing tests to the 2008 instance instead?

potatoqualitee:
Thanks @sanderstad for another fabulous command! <3

PR #2393, opened 2017-10-05
Get-DbaXEventSessionTarget and test

potatoqualitee:


functions/Get-DbaXEventSessionTarget.ps1
```diff
@@ -0,0 +1,143 @@
...
+		[string[]]$Session,
+		[string[]]$Target,
+		[parameter(ValueFromPipeline, ParameterSetName = "piped", Mandatory)]
+		[Microsoft.SqlServer.Management.XEvent.Session[]]$SessionObject,
+		[switch]$Silent
+	)
+	
+	begin {
+		if ([System.Reflection.Assembly]::LoadWithPartialName("Microsoft.SqlServer.Management.XEvent") -eq $null) {
+			Stop-Function -Message "SMO version is too old. To collect Extended Events, you must have SQL Server Management Studio 2012 or higher installed."
```
FriedrichWeinmann: should have a return on the next line.
Not a technical requirement for this function, but it is neater and avoids errors if someone later adds things to the begin block

functions/Get-DbaXEventSessionTarget.ps1
```diff
@@ -0,0 +1,143 @@
...
+		}
+		
+		function get-target ($Sessions, $server) {
```
FriedrichWeinmann: I know it's an internal function, but still, it should have a `[CmdletBinding()]` (and thus a param block) anyway
Also watch your casing (it's upper-case when you later call it)

functions/Get-DbaXEventSessionTarget.ps1
```diff
@@ -0,0 +1,143 @@
...
+			foreach ($xsession in $Sessions) {
+				
+				if ($null -eq $server) {
+					$server = $xsession.Parent
+				}
+				
+				if ($Session -and $xsession.Name -notin $Session) { continue }
```
FriedrichWeinmann: Scope boundary violation detected.

functions/Get-DbaXEventSessionTarget.ps1
potatoqualitee: fixed

functions/Get-DbaXEventSessionTarget.ps1
potatoqualitee: fixed

functions/Get-DbaXEventSessionTarget.ps1
potatoqualitee: wut how to fix

potatoqualitee: nm i see

FriedrichWeinmann: By adding a `$Session` parameter to `Get-Target` and passing the original `$Session` parameter of `Get-DbaXEventSessionTarget` though to it.
Basically:
 - Add `$Session`-Parameter to param block of  `Get-Target`
 - Append `-Session $Session` to all calls of Get-Target later in the function

potatoqualitee: AHHHHHH thank you 🙇 I also added Target

PR #2406, opened 2017-10-07
Get-FilteredRestoreFile - fixed time comparison issue

FriedrichWeinmann:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Purpose
Fixes a time comparison that was breaking Restore-DbaDatabase when receiving input from Get-DbaBackupHistory.

### DIFF
--- a/internal/Get-FilteredRestoreFile.ps1
+++ b/internal/Get-FilteredRestoreFile.ps1
@@ -173,7 +173,7 @@ function Get-FilteredRestoreFile {
             #If we're continuing a restore, then we aren't going to be needing a full backup....
             $TlogStartlsn = 0
             if (!($continue)) {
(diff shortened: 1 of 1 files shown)


Stuart-Moore:
Hi Fred. What's changed downstream that means that conditional needs reversing? Just wanting to check there's not more impact that this one

FriedrichWeinmann:
Nothing that I know of.
Actually, I'm surprised it acts that way, as line 372 in Restore-DbaDatabase should pretty clearly convert the type to DateTime, which already should prevent this issue. I recognized the error and worked around it, but cannot see where it retains its original DbaDateTime type.

Stuart-Moore:
I think the reported issue might be to do with the history pipeline. The
gentleman was having issues with the ordering of history output so
wondering if there's something elsewhere going wrong. Waiting for some test
data from him to work out what's happening.

PR #2409, opened 2017-10-08
fix get-dbavlf

potatoqualitee:


potatoqualitee:
@wsmelton - worked on my local machine and works now with intance2 instead of instance1

potatoqualitee:
@wsmelton  Thank you very much for your work and Tests on these two commands!

PR #2414, opened 2017-10-09
Touchup Test-DbaTempdbConfiguration

wsmelton:
Touchup to standard.

Also changed piping the `$value` object to `Add-Member` to utilize the `-InputObject`, less overhead using that method.

Adjusted the test for this command as well to include more items and code coverage:
image


wsmelton:
Failure not associated with this command or files modified.

potatoqualitee:
thank you! 🚂 Great stuff

PR #2415, opened 2017-10-09
Adjust Test-DbaSqlPath

wsmelton:
Adjust to standard.

Went ahead and added a test to this one as it is used heavily in the module.

image

🤞 this passes on first run ⛅️

### DIFF
--- a/functions/Test-DbaSqlPath.ps1
+++ b/functions/Test-DbaSqlPath.ps1
@@ -1,16 +1,16 @@
 #ValidationTags#Messaging,FlowControl,Pipeline,CodeStyle#
 function Test-DbaSqlPath {
 	<#
-        .SYNOPSIS
-            Tests if file or directory exists from the perspective of the SQL Server service account.
-        
-        .DESCRIPTION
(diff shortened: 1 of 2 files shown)


wsmelton:
🌞

functions/Test-DbaSqlPath.ps1
```diff
@@ -1,96 +1,99 @@
...
+			Write-Message -Level VeryVerbose -Message "Checking acces to $path for $instance."
+			$sql = "EXEC master.dbo.xp_fileexist '$path'"
+			try {
+				Write-Message -Level Debug -Message "Executing: $sql."
+				$fileExist = $server.Query($sql)
+			}
+
+			catch {
+				Stop-Function -Message "Failed to test the path $Path." -ErrorRecord $_ -Target $SqlInstance
+				return
```
FriedrichWeinmann: The target should be `$instance` and you'd need to use the `-Continue` parameter here, instead of using return on the next line

wsmelton: Done. Caught this same issue on other commands.

PR #2423, opened 2017-10-10
Remove .Test from test output

potatoqualitee:


potatoqualitee:
failure is that frikken test vlf. passed once.

Issue #289, opened 2016-10-27
New cmdlet request - Rename-SqlDatabase


Zhunya:
Change name of one or more properties of a database:
- change database name
- change logical names, may be with template
- change file name, may be with template

Template parameters:
- %PX prefix
- %DN database name
- %FG file group name
- %FT file type
- %DT datetime
- %SX sufix

SirCaptainMitch:
I think that this has also been asked for in Copy-SqlDatabase. By adding this as a separate cmdlet we could potentially just call it there instead of adding a bunch of new code to the Copy-SQlDatabase.

potatoqualitee:
Gorgeous design, @Zhunya ! I'be brought it back up again and hopefully someone can get on it.

niphlod:
Going to write something but either with *PX* or <PX>. "%" is a valid character for a file, while * and < aren't. ATM I feel <PX> is more pleasant but changing that before the final release won't be an issue.

niphlod:
this is now in development. Happy renaming!

PR #2030, opened 2017-08-07
Rename-DbaDatabase, first round

niphlod:
## Type of Change
 - [ ] Bug fix (non-breaking change, fixes #<enter issue number>)
 - [x] New feature (non-breaking change, adds functionality, see #289)
 - [ ] Breaking change (effects multiple commands or functionality)
 - [ ] Ran manual Pester test and has passed (`.\tests\manual.pester.ps1)
 - [ ] Pester test is included

### Purpose
#289 holds pretty much everything. Not sure if this needs to handle also "moves" other than renames, such as "move this datafile to another disk".

This is just a first draft, but it's a working one nonetheless.
Cringe points:
- the templating system (I'm not good at naming things)
- general code style and loops, I feel there's a better way but can't find it
- it tries to proceed a database at a time and stop if anything goes wrong as soon as possible. Unfortunately reverting to the previous state can surely be done but could incur in the same errors, so for the moment...let's hope to not break everything
- the "default recursive behaviour" (see latest examples)
- what to return to the user, as it's a process made of several steps and ATM I can't figure a proper "layout" for the resulting object

For testers: do a backup of everything (or even better, a system image). This can (and probably will) nuclearize everything (again, "new born code alert").

I'm open to reviews and/or comments (don't go too technical on me, let's do that on round 2 or 3) ^_^

PR #2442, opened 2017-10-11
Rename-DbaDatabase, new command

niphlod:
## Type of Change
 - [x] New feature (non-breaking change, adds functionality, see #289)
 - [ ] Pester test is included

### Purpose
#289 holds pretty much everything. Not sure if this needs to handle also "moves" other than renames, such as "move this datafile to another disk".

This is just a first draft, but it's a working one nonetheless.
Cringe points:
- the templating system (I'm not good at naming things)
- general code style and loops, I feel there's a better way but can't find it
- it tries to proceed a database at a time and stop if anything goes wrong as soon as possible. Unfortunately reverting to the previous state can surely be done but could incur in the same errors, so for the moment...let's hope to not break everything
- the "default recursive behaviour" (see latest examples)
- what to return to the user, as it's a process made of several steps and ATM I can't figure a proper "layout" for the resulting object

For testers: do a backup of everything (or even better, a system image). This can (and probably will) nuclearize everything (again, "new born code alert").

I'm open to reviews and/or comments (don't go too technical on me, let's do that on round 2 or 3) ^_^

### DIFF
--- a/functions/Rename-DbaDatabase.ps1
+++ b/functions/Rename-DbaDatabase.ps1
@@ -1,0 +1,809 @@
+Function Rename-DbaDatabase {
+<#
+	.SYNOPSIS
+		Changes database name, logical file names, file group names and physical file names (optionally handling the move).
+	
+	.DESCRIPTION
+		Can change every database metadata that can be renamed. 
+		The ultimate goal is choosing to have a default template to enforce in your environment
+		so your naming convention for every bit can be put in place in no time.
+		The process is as follows (it follows the hierarchy of the entities):
+			- database name is changed (optionally, forcing users out)
+			- filegroup name(s) are changed accordingly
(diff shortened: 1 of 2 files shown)


niphlod:
clean reproposition for https://github.com/sqlcollaborative/dbatools/pull/2030, which is stale as ..... something stale. @zippy1981 test-psremoting is out of the picture. @potatoqualitee I'd go ahead and merge this, surely there will be improvements in the future but IMHO it does what it says it does.

gbargsley:
First test shows issues with EnableException it looks like.

image

niphlod:
yeah, this comes from before "the big rename" and I'm not sure why it has been left out to dry. I'll revamp in a bit.

niphlod:
umpf. Need to open yet another PR.....wait for it.

PR #2443, opened 2017-10-11
Test-PSRemoting, did I smell a rogue -Continue ?

niphlod:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 - [x] Refinement
 
### Purpose
The title says it all. 
Refine Test-PSRemoting (and learn, in the process). Did I get this right @FriedrichWeinmann ?

### DIFF
--- a/internal/Test-PSRemoting.ps1
+++ b/internal/Test-PSRemoting.ps1
@@ -5,34 +5,34 @@ Function Test-PSRemoting {
 Jeff Hicks
 https://www.petri.com/test-network-connectivity-powershell-test-connection-cmdlet
 #>
-  [cmdletbinding()]
- 
-  Param(
-    [Parameter(Position=0,Mandatory,HelpMessage = "Enter a computername",ValueFromPipeline)]
-    [ValidateNotNullorEmpty()]
-    [string]$Computername,
-    $Credential = [System.Management.Automation.PSCredential]::Empty
-  )
- 
-  Begin {
-    Write-Message -Level Verbose -Message "Starting $($MyInvocation.Mycommand)"
-  } #begin
- 
-  Process {
-    Write-Message -Level Verbose -Message "Testing $computername"
-    Try {
-      $r = Test-WSMan -ComputerName $Computername -Credential $Credential -Authentication Default -ErrorAction Stop
-      $True 
-    }
-    Catch {
-      Stop-Function -Message "Remote testing failed for computer $ComputerName" -Target $ComputerName -ErrorRecord $_ -Continue
-      return $false
-    }
-    
-  } #Process
- 
-  End {
-    Write-Message -Level Verbose -Message "Ending $($MyInvocation.Mycommand)"
-  } #end
- 
+	[cmdletbinding()]
+	param(
+	[Parameter(Position=0,Mandatory,HelpMessage = "Enter a computername",ValueFromPipeline)]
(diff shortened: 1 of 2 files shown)


internal/Test-PSRemoting.ps1
```diff
@@ -5,34 +5,34 @@ Function Test-PSRemoting {
...
- 
-  End {
-    Write-Message -Level Verbose -Message "Ending $($MyInvocation.Mycommand)"
-  } #end
- 
+	[cmdletbinding()]
+	param(
+	[Parameter(Position=0,Mandatory,HelpMessage = "Enter a computername",ValueFromPipeline)]
+	[ValidateNotNullorEmpty()]
+	[string]$Computername,
```
FriedrichWeinmann: Go ahead and use our `[DbaInstance]` type here (and use the `.ComputerName` property later).
Makes it easier on contributors without breaking anything.
In that case, you also do not need that validate attribute.
The helpmessage is not much use in an internal command either

internal/Test-PSRemoting.ps1
```diff
@@ -5,34 +5,34 @@ Function Test-PSRemoting {
...
+
+	process {
+		Write-Message -Level VeryVerbose -Message "Testing $computername"
+		try {
+			$r = Test-WSMan -ComputerName $Computername -Credential $Credential -Authentication Default -ErrorAction Stop
+			$True 
+		}
+		catch {
+			Stop-Function -Message "Remote testing failed for computer $ComputerName" -Target $ComputerName -ErrorRecord $_
+			return $false
```
FriedrichWeinmann: Consider whether this is how you want to use it. It's a significant improvement over `-Continue`, but do you want to throw on `-Silent`
It's a test command, so it may be appropriate to just return false.
In that case, you can use Write-Message to log the error record instead.
In order to find the right answer here, you'd have to check how this command is used by our other commands and whether they are designed to handle exceptions or boolean returns.

FriedrichWeinmann:
Alright, just looked it up:
This command is used only by `Get-DbaMsdtc`, `Get-DbaPrivilege` and `Set-DbaPrivilege`

In all cases it's used in a manner expecting a boolean result.

 - Drop Stop-Function entirely and use Write-Message instead (Level Verbose plus adding the error record should do the trick. Internal tests shouldn't write warnings to the user screen I'd say - that's the responsibility of the calling commands, if they deem it appropriate).

niphlod:
yeah, definitely this is intended to work as a boolean. I'll update the PR and add tests soon. Thanks, as always, for the nice review.

potatoqualitee:
Thank you all 🌈 Can you please tend to the changes so we can get this closed today?

niphlod:
the function is roasted since mid-August (0.9.34), it has been reported originally on mid-August and never fixed ..... this PR is just 6 days old... we *could* get a bit of a stretch if there's a review and a regression test in the middle .... 

Anyway, I'll finish this up probably today (i.e. in 6 or 7 hours, tops.)

niphlod:
voilà, @potatoqualitee and @FriedrichWeinmann

potatoqualitee:
yall the bestttttt! thank you 🏆

potatoqualitee:
@FriedrichWeinmann please merge at your convenience

PR #2452, opened 2017-10-16
Touchup Test-DbaServerName

wsmelton:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Making 1.0 standard but also cleaned up a few things in the logic so this needs more formal review.

Not sure I used the Silent parameter correctly when calling `Get-DbaSqlService`, Fred or Simone may need to check that part for me.

### Screenshots

Previous output for default view (not using Detailed)

image

new version replaces the reference to the instance name and changes them to our standard initial properties for the objects

image

The `NoWarning` parameter didn't sit well so changes this to just `SkipSsrs`, probably could be named something better. Logic either way should not even bother checking for SSRS if they don't care about the warning, so adjusted that logic to not even bother validating SSRS exist. If this is wrong I can put it back.

image
image

Then doing select * has the additional two properties showing

image

### DIFF
--- a/functions/Test-DbaServerName.ps1
+++ b/functions/Test-DbaServerName.ps1
@@ -35,6 +35,7 @@ function Test-DbaServerName {
 
 		.NOTES
 			Tags: SPN, ServerName
+
 			dbatools PowerShell module (https://dbatools.io, clemaire@gmail.com)
 			Copyright (C) 2016 Chrissy LeMaire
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Can you make it `ExcludeSsrs` instead please, thank you

wsmelton:
You know I had typed it as that the first time and changed it 👍 

All done.

PR #2472, opened 2017-10-19
Balance patch

potatoqualitee:
Please do not merge yet


functions/Invoke-DbaBalanceDataFiles.ps1
```diff
@@ -84,113 +92,109 @@ function Invoke-DbaBalanceDataFiles {
 	By supplying this parameter you give permission to do the rebuilds offline if the edition does not support it.
 
 #>
-	[CmdletBinding(DefaultParameterSetName = "Default", SupportsShouldProcess = $true)]
+	[CmdletBinding(SupportsShouldProcess = $true, ConfirmImpact = 'High', DefaultParameterSetName = "Default")]
 	param (
-		[parameter(ParameterSetName = "Pipe", Mandatory = $true)]
+		[parameter(Mandatory, ParameterSetName = "instance")]
+		[Alias("ServerInstance", "SqlServer")]
 		[DbaInstanceParameter[]]$SqlInstance,
```
wsmelton: This should just be `[DbaInstance]`. I've been removing the `DbaInstanceParameter` as I touch commands so we can get rid of it :tada:

functions/Invoke-DbaBalanceDataFiles.ps1
```diff
@@ -84,113 +92,109 @@ function Invoke-DbaBalanceDataFiles {
...
-					
-					[bool]$supportOnlineRebuild = $false
-				}
-				1 {
-					Stop-Function -Message "You chose to not allow offline rebuilds of indexes. Use -RebuildOffline" -Target $SqlInstance
-					return
+			# Check the database parameter
+			if ($Database) {
+				if ($Database -notin $server.Databases.Name) {
+					Stop-Function -Message "One or more databases cannot be found on instance on instance $instance" -Target $instance -Continue
```
wsmelton: OH! I like this check...we should add this to the other commands...it will clean up errors of databases not existing.

Need to update the template with this example.

functions/Invoke-DbaBalanceDataFiles.ps1
```diff
@@ -84,113 +92,109 @@ function Invoke-DbaBalanceDataFiles {
...
 				}
-			} # switch
-		}
-		elseif ($serverVersion -ge 9 -and (([string]$Server.Edition -like "Developer*") -or ($Server.Edition -like "Enterprise*"))) {
-			[bool]$supportOnlineRebuild = $true
+				
+				$DatabaseCollection += $server.Databases | Where-Object { $_.Name -in $Database }
```
wsmelton: I may be missing it but I do not see any check to see if the databases are even accessible.

functions/Invoke-DbaBalanceDataFiles.ps1
```diff
@@ -84,113 +92,109 @@ function Invoke-DbaBalanceDataFiles {
...
+			# Get the server version
+			$serverVersion = $server.Version.Major
+			
+			# Check edition of the sql instance
+			if ($RebuildOffline) {
+				Write-Message -Message "Continuing with offline rebuild." -Level Verbose
+			}
+			elseif (-not $RebuildOffline -and ($serverVersion -lt 9 -or (([string]$Server.Edition -notmatch "Developer") -and ($Server.Edition -notmatch "Enterprise")))) {
+				# Set up the confirm part
+				$message = "The server does not support online rebuilds of indexes. `nDo you want to rebuild the indexes offline?"
```
wsmelton: Do we need to validate that the databases are involved in mirroring or an AG as well?

Doing this to a database in an AG can cause serious issues on the secondaries if the (1) index is large or (2) it is just a large database in general.

wsmelton: Be something like "This database is involved in an AG or mirror, are you sure you want to continue?"

wsmelton: Check for Log Shipping as well...big time issue in some client servers with the standby server getting behind because of index maintenance.

potatoqualitee:
@sanderstad will be returning to this when he returns home, btw

potatoqualitee:
Sup @sanderstad

PR #2476, opened 2017-10-19
version on top to override settings in appveyor UI

niphlod:
(do Identity)

version was just added as an environmental variable (maybe is useful in c# tests ?)

However [appveyor docs](https://www.appveyor.com/docs/build-configuration/#appveyoryml-and-ui-coexistence) dictate that to override the build number set in appveyor UI, "version" should be at the root of the setting "tree"


potatoqualitee:
looks good 😎

PR #2481, opened 2017-10-20
Fixing Get-FilteredRestoreFile: Using comma in SqlInstance causes function to fail

nvarscar:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes issue reported in slack: Using comma in SqlInstance causes function to fail due to incorrect parsing of the Group-Object Name property)
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Removing .Name.Split(',').Trim() in favor of .Values

### Approach
Stop using string Name, start using grouped objects

### Commands to test
included in pester

### Screenshots
Before:
image
After:
image


Stuart-Moore:
Also fixed in 2794

Root cause of error is in restore-dbadatabase, so fixed it there.

Just need to fix fred induced test fails. All passed before the great
rename.

nvarscar:
I just realized that it would also affect databases with trailing spaces in their names and checked for other trim() statements. Fixing this in a moment.

nvarscar:
Well, the test passed, but one of the other scenarios failed due to Agent not being able to startup in time lol

PR #2492, opened 2017-10-22
#2046 Performance improvements

### LINKED ISSUE #2046: Out-DbaDataTable seems slow
@jpomfret reported some slowness that I can repro. Jess, did it get slower recently?

Here's my sample

image

7429 with 26 columns (Kawww!) when piped takes 1 min, 47 seconds. When used as `InputObject` then 22 seconds. Not so bad I suppose, considering how wide it is.

JohnRoos:
## Type of Change
 - [x] Bug fix (non-breaking change, fixes #2046 )
 - [x] Pester test is included

No functionality has changed. Out-DbaDataTable should be a bit faster, especially with large objects.

### DIFF
--- a/functions/Out-DbaDataTable.ps1
+++ b/functions/Out-DbaDataTable.ps1
@@ -219,7 +219,8 @@ Function Out-DbaDataTable {
 				}
 				else {
 					$datarow = $datatable.NewRow()
-					foreach ($property in $object.PsObject.get_properties()) {
+					$objectProperties = $object.PsObject.get_properties()
+					foreach ($property in $objectProperties) {
 						# The converted variable will get the result from the ConvertType function and used for type and value conversion when adding to the datatable.
 						$converted = @{ }
 						if ($ShouldCreateCollumns) {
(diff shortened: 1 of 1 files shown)


wsmelton:
Do you have a comparison of the timing that shows this improved performance?

I can't say I have observed improvements by changing a one if/else into multiple nested if/else statements.

JohnRoos:
Small test samples probably doesnt show much difference, but with a larger amount of fairly coplex objects there is a difference in my tests.
Try this:
`$samples = @()
1..10 | % { $samples += ls -Path C:\git\dbatools -Recurse -File }
$samples.count # in my environment this is 8240 objects
1..3 | % { measure-command { $datatable = Out-DbaDataTable -InputObject $samples } | select totalseconds }`

My results with the current version:
TotalSeconds
------------
  26.8969817
  25.7828032
  25.0991068

With this version:
TotalSeconds
------------
  18.0843163
  16.6114867
   17.205913

Please let me know if you still dont notice any difference.

potatoqualitee:
Ouuuuu thank youu! test failed but its unrelated

PR #2494, opened 2017-10-23
#2044: Get-DbaLogin: Add -IncludeFilter and -ExcludeFilter

### LINKED ISSUE #2044: Get-DbaLogin:  add parameter ExcludeFilter to allow filtering logins like '##*','NT *'
## Feature Request
  Add an additional parameter to Get-DbaLogin to exclude system/built-in logins.

### Problem to solve
Currently only following logins can be excluded:

- [object[]]$ExcludeLogin,
- [switch]$HasAccess,
- [switch]$Locked,
- [switch]$Disabled

I want to have an additional paramter to exclude $NoSystem logins too.

A simple new parameter can exlude these accounts based on their names:

` if ($NoSystem) 
{ $serverLogins = $serverLogins | Where-Object { !(($_.Name.StartsWith('##') -or $_.Name -eq 'sa' -or $_.Name.StartsWith('NT AUTHORITY\') -or $_.Name.StartsWith('NT SERVICE\'))) }
}`

### Additional information
I've already tested the new paramter for SQL versions 2008-2012.

rjcorrig:
## Type of Change

 - [X] New feature (non-breaking change, adds functionality, resolves #2044 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [X] Documentation
 - [ ] Build system
 

### Purpose
Allows wildcard include and exclude filters on the results of Get-DbaLogin, per #2044. The original suggested use case was to filter out system logins such as sa or those beginning with NT, later expanded to a more general wildcard filter.

### Approach
Pass a list of wildcarded strings as the value of either -IncludeFilter or -ExcludeFilter. The strings are processed according to the -Like and -NotLike PowerShell operators. The -IncludeFilter is processed before the -ExcludeFilter to support "these logins here except those logins over there" semantics.

### Commands to test
(From the revised documentation)
Get-DbaLogin -SqlInstance sql2016 -IncludeFilter '##\*','NT \*'
Get-DbaLogin -SqlInstance sql2016 -ExcludeFilter '##\*','NT \*'

### DIFF
--- a/functions/Get-DbaLogin.ps1
+++ b/functions/Get-DbaLogin.ps1
@@ -18,6 +18,12 @@ function Get-DbaLogin {
 		.PARAMETER ExcludeLogin
 			The login(s) to exclude - this list is auto-populated from the server
 
+		.PARAMETER IncludeFilter
+			A list of logins to include - accepts wildcard patterns
+
+		.PARAMETER ExcludeFilter
+			A list of logins to exclude - accepts wildcard patterns
+		
 		.PARAMETER Locked
 			Filters on the SMO property to return locked Logins.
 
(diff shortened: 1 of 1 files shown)


potatoqualitee:
fanastic! Thank you @rjcorrig 💯

PR #2501, opened 2017-10-24
Add parameter to Select-Object when finding dbname

### LINKED ISSUE #2500: Restore-DBADatabase failing on wrong case
## Bug Report
- [ X ] Verified running latest release of dbatools?
- [ X ] Verified errors are not related to permissions?
- [ X ] Can duplicate in new/fresh PowerShell.exe session?

### Version Information
 - Operating System Windows 2012
 - PowerShell Version: 4.0
 - SQL Server SQL 2014 (N/A)

## Steps to Reproduce

1. Have backups from Ola's scripts.
2. Take a manual Ola backup e.g. EXEC usp_BackupDB_Full_1_File @DBName = "Different case than your DB" ex. AdventureWorks becomes adventureworks
3. Have log backups as well.
4. Try restoring database

This will error out Restore-DBADatabase on line 580 as it is trying to pass a list at that point to Restore-DBFromFilteredArray -OldDatabaseName where it expects a String. I fixed this by modifying line 574 on my local copy from $OldDatabaseName = ($FilteredFiles | Select-Object -Property DatabaseName -unique).DatabaseName to $OldDatabaseName = ($FilteredFiles | Select-Object -Property DatabaseName -unique- First 1).DatabaseName. I'll submit the code change soon as I don't believe it to break anything else.

josh-simar:
Select-Object needs the -First 1 parameter added due to the fact that Group-Object above is case-insensitive but it is case-sensitive so it's passing multiple values to a parameter that should be a string. Ties to #2500

## Type of Change

 - [ X ] Bug fix #2500 
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose

Fix a possible array so that the string that the function expects is created

### Approach
- First 1 only takes the first of the array (can only be an array if above didn't catch it as Group-Object is case-insensitive but Select-Object is not.

### Commands to test
Just the one modified in my opinion.

### DIFF
--- a/functions/Restore-DbaDatabase.ps1
+++ b/functions/Restore-DbaDatabase.ps1
@@ -569,9 +569,9 @@ function Restore-DbaDatabase {
 				Stop-Function -Target $FilteredFiles -Message "We can only handle 1 Database at a time - $dbs"
                 return
             }
-			$OldDatabaseName = ($FilteredFiles | Select-Object -Property DatabaseName -unique).DatabaseName
+			$OldDatabaseName = ($FilteredFiles | Select-Object -Property DatabaseName -unique -First 1).DatabaseName
             IF ($DatabaseName -eq '') {
-                $DatabaseName = $RestoredDatababaseNamePrefix + ($FilteredFiles | Select-Object -Property DatabaseName -unique).DatabaseName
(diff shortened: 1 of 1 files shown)


josh-simar:
Gotcha. Will do.

PR #2509, opened 2017-10-24
Add case-sensitivity to GroupObject for CS collations

josh-simar:
Ran into an issue where I had manually created a backup of database and it had a different case. Because of this Restore did not work. Group-Object was failing to detect multiple values because by default it is case insensitive but Select-Object below is not. Added case-sensitivity to group-object to avoid this error unknown and instead give a proper error message.

## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system


potatoqualitee:
aw, what happened, @josh-simar ? that looked cool

PR #2510, opened 2017-10-24
Add case-sensitivity to GroupObject for CS collations

josh-simar:
Ran into an issue where I had manually created a backup of database and it had a different case. Because of this Restore did not work. Group-Object was failing to detect multiple values because by default it is case insensitive but Select-Object below is not. Added case-sensitivity to group-object to avoid this error unknown and instead give a proper error message.

## Type of Change

 - [ X ] Bug fix (non-breaking change, fixes #<enter issue number>)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Allow CaseSensitivity on group object for case sensitive collations

### Approach
Uses switch already there but not used by default.

### Screenshots

image

### Learning
I did this by mistake by running an out-of-band backup manually and not noticing the case I was using when calling the database name.

### DIFF
--- a/functions/Restore-DbaDatabase.ps1
+++ b/functions/Restore-DbaDatabase.ps1
@@ -655,7 +655,7 @@ function Restore-DbaDatabase {
                 
                 
                 Write-Message -Level Verbose -Message "Starting FileSet"
-                if (($FilteredFiles.DatabaseName | Group-Object | Measure-Object).count -gt 1) {
+                if (($FilteredFiles.DatabaseName | Group-Object -CaseSensitive | Measure-Object).count -gt 1) {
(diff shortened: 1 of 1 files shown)


potatoqualitee:
ahh, you moved it here! k, will merge once tests pass.

potatoqualitee:
excellent, all checks pass!

potatoqualitee:
Thanks much, @josh-simar! We really need to get a CS edition in our lab mix.

josh-simar:
I'm honestly glad I can help. Most of the code is over my head but when I can find something I'm glad I'm able to help the community.

PR #2523, opened 2017-10-25
Adjust Get-DbaDatabaseFile - Add "friendly" growth size

### LINKED ISSUE #2471: Get-DbaDatabaseFile returning wrong 'Growth'
Hello,

Using the command 'Get-DbaDatabaseFile' doesn't return the correct 'Growth'. Screenshots below.

- [x ] Verified running latest release of dbatools?
- [x ] Verified errors are not related to permissions?
- [x ] Can duplicate in new/fresh PowerShell.exe session?

### Version Information
 - Operating System (Name|Version): Windows 10 
 - PowerShell Version: 5.1
 - SQL Server (Edition|Version): Microsoft SQL Server 2017 (RTM) - 14.0.1000.169 (X64)

### Additional information
sql_growth_1

sql_growth_2

josh-simar:
Added friendly growth size calculation to output. Classifying for both Percent and kb it is named "NextGrowthEventSize" which should be static for kb but will change based on filesize for percent growth.

## Type of Change

 - [ X ] New feature (non-breaking change, adds functionality)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Fixes  #2471

### Approach
Calculates how much space the next growth event will take with different calculations for kb vs percent.

### Commands to test
Get-DbaDatabaseFile

### DIFF
--- a/functions/Get-DbaDatabaseFile.ps1
+++ b/functions/Get-DbaDatabaseFile.ps1
@@ -186,7 +186,12 @@ function Get-DbaDatabaseFile {
 						$free = $disks | Where-Object { $_.drive -eq $result.PhysicalName.Substring(0, 1) } | Select-Object 'MB Free' -ExpandProperty 'MB Free'
 						$VolumeFreeSpace = [dbasize]($free * 1024 * 1024)
 					}
-					
+					if($result.GrowthType -eq "Percent") {
+						$nextgrowtheventadd = [dbasize]$size * ($result.Growth * 0.01)
+					}
+					else {
+						$nextgrowtheventadd = $result.Growth * 8
+					}
 					
 					[PSCustomObject]@{
 						ComputerName             = $server.NetName
@@ -203,6 +208,7 @@ function Get-DbaDatabaseFile {
 						MaxSize                  = $maxsize
 						Growth                   = $result.Growth
 						GrowthType               = $result.GrowthType
+						NextGrowthEventSize	 = $nextgrowtheventadd
 						Size                     = $size
 						UsedSpace                = $usedspace
 						AvailableSpace           = $AvailableSpace


wsmelton:
This does not address that issue. The issue is noting that the growth value is not showing a friendly value. The property you added is more suited to the `Get-DbaDatabaseFreeSpace` command, as it is more on topic for that data.

To address #2471 would mean making the Growth property a friendly value based on whether it is a percent or kb growth rate.

josh-simar:
@wsmelton I can do that if wanted on this command. @ClaudioESSilva, care to comment as well?

potatoqualitee:
My brain is too muddled up with a cold :( And claudio is out at a conference. @PowerDBAKlaas can you decide?

PowerDBAKlaas:
right, I'm gonna make this a bit more complicated

I believe growth is useful both for `Get-DbaDbSpace' and `Get-DbaDbFile`: the setting is defined on every file, but whether a DB can grow is a sum of the settings in all files in a filegroup. There it will require some calculations.
While we're doing that, I think this is a nice opportunity to compare the next growth with the available space: if it's bigger, there will be no growth at all. When autogrow is disabled neither. In that case the `NextGrowthEventSize` should be 0, or we could add a 'info' property, if that's still on our style guide.

Regarding units: we should always keep the raw, original number and add calculated sizes if useful. In this case it seems SSMS GUI already does show another unit, so it may be wise to choose the same one, as DBA's will be most familiar with it.
If we judge that there are better options, we can either add more, or offer the user a parameter to choose the units himself.
Select-Defaultview could contain only the 'friendly named' ones.
We will have to include the unit in the propertyname if the value is numeric.

I'm surprised this is still a PSCustomObject. I thought this was changed to a file object a long time ago. Would this be the time to rewrite?

wsmelton:
@PowerDBAKlaas A file object?

PowerDBAKlaas:
@wsmelton that's what I assumed it was. I thought we did that a long time ago. Maybe it was only for backup file?

potatoqualitee:
can this PR get some lovin' plz @josh-simar, @wsmelton

wsmelton:
Well if we stick to the basics of this PR should address the issue referenced, the only thing needed is to match up the output of `Growth` property to the unit shown by SSMS. 

So, a calculated value would be added to show `Growth` as either a percent or MB value (since that is the unit SSMS shows). Then another property that shows the raw value as noted by @PowerDBAKlaas . 

The other items can be addressed in a different PR as this command is due for a rewrite to address other issues.

@josh-simar if you can adjust the command with the above items we can get this PR completed.

potatoqualitee:
hey @josh-simar - can you please update this PR and we can get it merged? Thanks much

josh-simar:
Yes. I’ve got it on my plan for Monday. Sorry for the delay.

josh-simar:
At this time I'm unsure what should be done. reading through what has ultimately been proposed, growth should show as it does in SSMS which it does, it shows the percent growth if percent and KB if fixed growth as it shows in SSMS. Growth type then shows kb or percent. The added "Next Growth Event" shows exactly how much that will be using the dbasize variable created (I'm guessing in the dll). I really want to get this fixed but I'm just not sure how at this point.

potatoqualitee:
I understand, I don't either ;) Can you help clarify, @wsmelton ?

PowerDBAKlaas:
Although this would not be a final version of this function, the change @josh-simar submitted is a good addition.
However, there are 2 issues I'd like to see handled before accepting:
+ the name of the property should be something like  `NextGrowthKB` instead of `..Size`
+ `if ($_.NextGrowthKB -gt $_.AvailableSpace ) { NextGrowthKB = 0}`

We can think about more improvements later.
Agreed @wsmelton @cl ?

potatoqualitee:
I dig it, thank you @PowerDBAKlaas

josh-simar:
Just want to make sure. Since we are using [dbasize] I have a feeling that could be in a number of different size variations (KB vs MB vs GB) and that being the case would we not want to call the property NextGrowth instead of NextGrowthKB? Also, knowing that will it still work with $_.AvailableSpace or will extra code be needed for that?

Thank you.

PowerDBAKlaas:
@josh-simar you're right about the `[dbasize]` type.
They're very handy for presentation, but require some attention when calculating or comparing. That's why I like them as additional properties, but not to replace the original values. Since there are a lot of properties in this function cast like this, we can indeed add the `NextGrowth` without `KB` or `MB`.
We'll see afterwards what we're gonna do with this.

The calculation with `availablespace` will have to be done with the raw values before the casting;
something like `( size * growth / 100 ) -lt availablespace` for the percent growth,
or else with a size property;
something like `NextGrowth.Megabyte -lt AvailableSpace.Megabyte`

PowerDBAKlaas:
Commits from this PR are reprised in new branch DBFileNextgrowth.
thanks @josh-simar

PR #2524, opened 2017-10-25
Solving Module Import Woes

FriedrichWeinmann:
# Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
## Fixes

 - Fixed configuration system issues with processing user profiles, causing module import to hang
 - Fixed configuration system issues changing settings using `Set-DbaConfig`
 - Fixed error preventing PowerShell version 3 or 4 from using the web install command

### Configuration System issues
Scriptblocks can only be executed on the runspace they were created on. Moving configuration processing to another runspace - which was done to improve import performance - caused the system to ...

 - Try to execute the profile configuration settings (which were created in the main runspace) in a foreign runspace
 - Create the validation and handler scriptblocks to be created in a foreign runspace, causing `Set-DbaConfig` to fail in the main runspace.

This was resolved by having both kinds of scriptblock cloned before invoking them, which has the clones created in the runspace that executes them.

### Web Install issues
Since our web installation deployment does not ship the library, it tries to build it from the project. The recent change to name runspaces however requires a feature only present in PS5. This was noted during development and handled for dlls built on PS5, so that they could run on older systems. However I didn't consider that we also build on older systems through said web installer.

I solved this issue by setting up project configurations per PowerShell versions and excluding the offending code, using a property only PS5, from compilation on systems with PS4 or older. Given that it is a cosmetic feature, this will not cause any compatibility issues.

### DIFF
--- a/bin/library.ps1
+++ b/bin/library.ps1
@@ -1,5 +1,5 @@
 ﻿# Current library Version the module expects
-$currentLibraryVersion = New-Object System.Version(0, 9, 0, 32)
+$currentLibraryVersion = New-Object System.Version(0, 9, 0, 33)
 
 <#
 Library Versioning 101:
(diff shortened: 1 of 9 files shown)


potatoqualitee:
Thanks!

PR #2527, opened 2017-10-26
Get-DbaAgentOperator Repair

PowerDBAKlaas:
## Type of Change
 - [X ] Bug fix (non-breaking change, fixes #<enter issue number>)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Fix the misbehaviour: this function did return only the operators of the first sql instance. 

### Approach
A parameter "Operator" was added, but this was already used in a foreach loop. After the first iteration through the sql instances, this parameter would contain the result of the last iteration through the operators of the first instance. For all consecutive sqlinstances, the check for operators being equal to the $operator would return nothing.
Renaming the variable in the 'foreach operator' loop, solved this.

### DIFF
--- a/functions/Get-DbaAgentOperator.ps1
+++ b/functions/Get-DbaAgentOperator.ps1
@@ -95,17 +95,17 @@ function Get-DbaAgentOperator {
 				$operators = $server.JobServer.Operators
 			}
 			
-            foreach ($operator in $operators) {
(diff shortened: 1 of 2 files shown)


potatoqualitee:
Merci, my friend 🌈

PR #2536, opened 2017-10-26
[New Command] Start-DbaXESession

dgmyrs:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Add function Start-DbaXESession to start Extended Event Sessions.

### Commands to test

Examples are in the help.

### DIFF
--- a/functions/Start-DbaXESession.ps1
+++ b/functions/Start-DbaXESession.ps1
@@ -1,0 +1,133 @@
+function Start-DbaXESession {
+<#
+	.SYNOPSIS
+	Starts Extended Events sessions.
+
+	.DESCRIPTION
+	This script starts Extended Events sessions on a SQL Server instance.
+
+	.PARAMETER SqlInstance
+	The SQL Instances that you're connecting to.
+
+	.PARAMETER SqlCredential
+	Credential object used to connect to the SQL Server as a different user
+
+	.PARAMETER Session
+	Only start specific Extended Events sessions.
+
+	.PARAMETER AllSessions
+	Start all Extended Events sessions on an instance, ignoring the packaged sessions: AlwaysOn_health, system_health, telemetry_xevents.
+
+	.PARAMETER Silent
+	If this switch is enabled, the internal messaging functions will be silenced.
+
+	.NOTES
+	Tags: Memory
+	Author: Doug Meyers
+	Website: https://dbatools.io
+	Copyright: (C) Chrissy LeMaire, clemaire@gmail.com
+	License: GNU GPL v3 https://opensource.org/licenses/GPL-3.0
+
+	.LINK
+	https://dbatools.io/Start-DbaXESession
+
+	.EXAMPLE
+	Start-DbaXESession -SqlInstance sqlserver2012 -AllSessions
+
+	Starts all Extended Event Session on the sqlserver2014 instance.
+
+	.EXAMPLE
+	Start-DbaXESession -SqlInstance sqlserver2012 -Session xesession1,xesession2
+
+	Starts the xesession1 and xesession2 Extended Event sessions.
+
+	.EXAMPLE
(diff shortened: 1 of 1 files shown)


potatoqualitee:
ohh snap@ looking forward to getting this one merged! I'll take a closer look once im done with practice

functions/Start-DbaXESession.ps1
```diff
@@ -0,0 +1,133 @@
...
+	Only start specific Extended Events sessions.
+
+	.PARAMETER AllSessions
+	Start all Extended Events sessions on an instance, ignoring the packaged sessions: AlwaysOn_health, system_health, telemetry_xevents.
+
+	.PARAMETER Silent
+	If this switch is enabled, the internal messaging functions will be silenced.
+
+	.NOTES
+	Tags: Memory
```
wsmelton: This should adjusted to maybe `Tags: XE, ExtendedEvents`

functions/Start-DbaXESession.ps1
```diff
@@ -0,0 +1,133 @@
...
+		[parameter(ParameterSetName='Session')]
+		[parameter(ParameterSetName='All')]
+		[PSCredential]$SqlCredential,
+
+		[parameter(Mandatory, ParameterSetName='Session')]
+		[Alias("Sessions")]
+		[object[]]$Session,
+
+		[parameter(Mandatory, ParameterSetName='All')]
+		[switch]$AllSessions,
```
wsmelton: This is a dangerous parameter to add when it comes to XE and I would disagree with adding it. Unless @potatoqualitee has some overwhelming evidence to think it needs to be there.

Starting up EVERY session on a given instance should never be done in the real world because the DBA does not know what those sessions are doing. An example, you have a one-off session configured that was capturing execution plans on an OLTP system. A session that was capturing AG related information (see examples from TigerTeam) that is recommended by Microsoft to never be left running for an extended period of time.

I don't think this should be offered...but that is just my opinion.

functions/Start-DbaXESession.ps1
```diff
@@ -0,0 +1,133 @@
...
+		if ([System.Reflection.Assembly]::LoadWithPartialName("Microsoft.SqlServer.Management.XEvent") -eq $null) {
+			Stop-Function -Message "SMO version is too old. To collect Extended Events, you must have SQL Server Management Studio 2012 or higher installed."
+		}
+
+		# Start each XESession
+		function Start-XESessions {
+			[CmdletBinding()]
+			param ([object]$xesessions)
+
+			foreach ($x in $xesessions) {
```
wsmelton: Please adjust this to be `$xe`....a bit more understandable compared to just "x".

As well, on variables that are two words we have been using camelCase. So `$xesessions` would be `$xeSessions`...we can clean this up after but if you can do it now would save us some time 💯

functions/Start-DbaXESession.ps1
```diff
@@ -0,0 +1,133 @@
...
+				
+				# Filter xesessions based on parameters
+				if ($Session) {
+					$xesessions = $xesessions | Where-Object { $_.Name -in $Session }
+				} elseif ($AllSessions) {
+					$systemSessions = @('AlwaysOn_health', 'system_health', 'telemetry_xevents')
+					$xesessions = $xesessions | Where-Object { $_.Name -notin $systemSessions }
+				}
+				
+				Start-XESessions $xesessions
```
wsmelton: This needs to be wrapped in a try/catch, or do that in the function itself.

functions/Start-DbaXESession.ps1
potatoqualitee: I think there are cases when it'd be used. We could add some additional prompts to if they use it

potatoqualitee:
Thanks for the review @wsmelton ! I will merge, patch and add a test 💃 !

PR #2542, opened 2017-10-27
Return Plan Warnings with Get-DbaExecutionPlan

### LINKED ISSUE #1973: Search for implicit conversions in execution plans
## Feature Request
Provide the ability to search and/or highlight implicit conversions in plans from the plan cache so we can identify them for potential performance tuning opportunities.

### Problem to solve
Implicit conversions are a performance inhibitors, mainly due to the way that they force index scans on queries that could benefit from index seeks.

They are also not highlighted unless you look in the execution plans (boo!), but this also means that these implicit conversions are stored in the plan cache (yay!).

If we highlight these plans with implicit conversions in them then, potentially, the user can go back to the queries, choose the correct data type (if they can), and that should help with query performance in turn.

### Additional information
This was originally from a blog post by Jonathan Kehayias of SQLskills:
https://www.sqlskills.com/blogs/jonathan/finding-implicit-column-conversions-in-the-plan-cache/

Thinking that it could be a possible add-on to `Get-DbaExecutionPlan` but will need to see if it's feasible or worthwhile.

shaneis:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #1973 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Added functionality to Get-DbaExecutionPlan to return warnings in Query Plans.
Started out just implicit conversions but it was slightly beyond my level so settled for every warning, as given by the plan.

### Approach

Navigates down the plan tree to the Warnings.PlanAffectingConvert section, if there is one, and returns what's there.

### Commands to test

Haven't changed help yet as can more than likely be improved.
```
Get-DbaExecutionPlan -ServerInstance 'servername' -Database 'databasename' |
  Where-Object {$_.PlanWarnings -ne $null} |
  Select-Object -ExpandProperty PlanWarnings;
```

### Screenshots

implicitconversions

Setup


wsmelton:
What about a command that inspects the execution plan specifically output from this command? Something like 'Test-DbaExecutionPlan` ? Then you could expand to check for various key areas of a plan that should be checked or inspected further by the user. In this change that would then allow you to pull out that property and then go through to test for various areas/warnings.

As well you have a conflict in this PR so you may need to update your branch and then resolve the merge issues.

shaneis:
Sounds good. 
It would help break it out into the different types of warnings then too.
Probably won't be as good as [How's My Plan](http://www.howsmyplan.com/) but we can hopefully get the main points then.

Yeah, couldn't figure out the conflicts.
Tried to resolve them but the only conflicts it showed where the changes I did...
Suppose that's what I get for stopping practicing git when it annoys me 😞

potatoqualitee:
Thanks much, @shaneis! i updated it to 1.0 while i was at it!

PR #2547, opened 2017-10-27
Rewriting for new format

ConstantineK:
## Type of Change

 - [x] Bug fix (no issue filed)

### Purpose
 
Fixes the change to the post format that Glenn was using.

### Approach

Changes approach to just find the first post that matches the existing format, grabbing relevant links, and returning all the ones that are dropbox related with diagnostic in the name (given he might add new versions, as with the Azure database.

I was a bit quick on the file handling, so if there's a better way to do that, I would love someone to take a look at line 94.

### Commands to test

Save-DbaDiagnosticQueryScript -Path .

### DIFF
--- a/functions/Save-DbaDiagnosticQueryScript.ps1
+++ b/functions/Save-DbaDiagnosticQueryScript.ps1
@@ -1,6 +1,6 @@
 function Save-DbaDiagnosticQueryScript {
   <#
-.SYNOPSIS 
+.SYNOPSIS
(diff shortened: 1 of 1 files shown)


ConstantineK:
The test fail is unrelated apparently, but I noticed that I changed the output format of the files, working on that, please do not merge.

potatoqualitee:
this is perfect 😸  thank you!

Issue #2116, opened 2017-08-16
Copy-DbaAudit bug reported via email


potatoqualitee:
> Hey there,
> You have a bug in your Copy-SqlAudit function.  It reports "Cannot bind argument to parameter 'Path' because it is an empty string." when you are trying to copy an Audit that writes to the application log.  It's the check on the file path that's failing so I wrote a fix for it if you're interested.
> 
> My new line is simply:
> `if (($audit.Filepath) -ne $null -AND $(Test-SqlPath -SqlServer $destserver -Path $audit.Filepath) -eq $false)`
> 
> 

That's all the old variables so would need to be updated to  using test-dbasqlpath -sqlilnstance

dhtaylor:
With the merge of this fix, does this close the issue?

garethnewman:
Hmmm is this fixed already?

potatoqualitee:
looks like it! Please let us know if either of you have issues.

PR #2553, opened 2017-10-27
#2116: Copy-DbaServerAudit: Test Current Audit Filepath for Null

dhtaylor:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Check the current audit filepath for null before testing the path to avoid error. This is a resolution to issue #2116 where this function is reporting "Cannot bind argument to parameter 'Path' because it is an empty string" when you try to copy an Audit that writes to the application log.

### Approach
Changes will first test that the value of $currentAudit.Filepath is not null before testing the path for existence.


potatoqualitee:
fantastic! Thank you both 🥇

PR #2564, opened 2017-10-30
Touchup Test-DbaMaxDop

wsmelton:
Touchup for 1.0 standard.

Addressed detailed switch and added in select default view command.


potatoqualitee:
heck yah 💃

PR #2568, opened 2017-10-31
Get-DbaDatabaseFile - Unified output type

FriedrichWeinmann:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Changes

 - Enforced `[DbaSize]` as output type for `.MaxSize` property in all instances

### Reason

Piping to `Out-DbaDataTable` would fail if the first file had no limit (and thus was represented with a an integer `-1` as max size), since the column types are defined by the first object. When later max sizes too large for int32 came along, this would fail in blood & error.

### Breaking Change

 - This will change the way -1 will show up on screen and in exports to csv (in which cases it will be marked as `Unlimited`). The latter may break some scripts that transfer data via file and then process the file.

### Do not merge

This shouldn't be merged before @potatoqualitee had a chance to look at it and decide whether that amount of breakage is alright.

### DIFF
--- a/functions/Get-DbaDatabaseFile.ps1
+++ b/functions/Get-DbaDatabaseFile.ps1
@@ -177,6 +177,9 @@ function Get-DbaDatabaseFile {
 					if ($maxsize -gt -1) {
 						$maxsize = [dbasize]($result.MaxSize * 8192)
 					}
+					else {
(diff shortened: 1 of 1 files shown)


potatoqualitee:
passes all tests, thank you!

PR #2569, opened 2017-10-31
Moved special character impacting variable usage

### LINKED ISSUE #2558: ? impacting $NetworkShare variable
https://github.com/sqlcollaborative/dbatools/blob/81a6f3f8711b92702db1b28e64d58ecec9dcd92e/functions/Copy-DbaDatabase.ps1#L1064

The ? on the above line appears to be breaking the $NetworkShare variable. Messages write without the network share name.

**_WARNING: [Copy-DbaDatabase][16:14:50] Backup Failed. Does SQL Server account DOMAIN\ACCOUNT have access to  Aborting routine for this database._**

cporteou:
$NetworkShare was not displaying due to a question mark impacting it.

## Type of Change

 - [x] Bug fix (non-breaking change, fixes #2558)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Minor edit to a variable in a Write command to allow it to show.

### Approach
Added a space to move special character

### Commands to test
Copy-DbaDatabase
Force a failure

### Screenshots
image

### DIFF
--- a/functions/Copy-DbaDatabase.ps1
+++ b/functions/Copy-DbaDatabase.ps1
@@ -1061,7 +1061,7 @@ function Copy-DbaDatabase {
 						$backupResult = $BackupTmpResult.BackupComplete
 						if ($backupResult -eq $false) {
 							$serviceAccount = $sourceServer.ServiceAccount
-							Write-Message -Level Verbose -Message "Backup Failed. Does SQL Server account $serviceAccount have access to $NetworkShare? Aborting routine for this database."
+							Write-Message -Level Verbose -Message "Backup Failed. Does SQL Server account $serviceAccount have access to $NetworkShare ? Aborting routine for this database."
 							
 							$copyDatabaseStatus.Status = "Failed"
 							$copyDatabaseStatus.Notes = "Backup failed. Verify service account access to $NetworkShare"


wsmelton:
Placement is everything with punctuation. :party:

gwojan:
@wsmelton. @cporteou. Guys, rather than add a space and make things look funny you could have surrounded the variable like ```${NetworkShare}```. So the warning message could be something like:

```PowerShell
"Backup Failed. Does SQL Server account '${serviceAccount}' have access to '${NetworkShare}'? Aborting routine for this database."
```

Notice I've even put the variables in single quotes to call the values out a little more.

cporteou:
Ah, that's perfect @gwojan. That's new to me, thank you

Would make sense to update all variables in messages then.

Shall I re-submit or you got this?

gwojan:
@cporteou, I'm not able to make a PR at this time. If you could make the change(s) that would be great. Thanks.

niphlod:
I dunno about curly, I'd have it gone with simple parenthesis for variable expansion

gwojan:
@niphlod Why? Please explain.

```${NetworkShare}``` looks way better than ```$($NetworkShare)``` especially if you live in bash or zsh where this is a very common idiom... Am I missing something?

**Comment by niphlod** [comment]
no real problem, I just said "I'd go for that", not "you need to go for that". In this particular case it works out.
I guess I simply didn't know the alternative existed without "-f", but at this point I'd say the "point" stands for "common usage". 
The `$()` notation is a full blown evaluation, and it works for everything (`$($foo.bar)`), the `${}` instead is only for strings ?

**Comment by cporteou** [comment]
Using parentheses won't work here though.

```
$NetworkShare = "\\share\share"

Write-Host "Some text $(NetworkShare)?"
```

results in an error.

```
$NetworkShare = "\\share\share"

Write-Host "Some text ($NetworkShare)?"
```

results in a needless set of parentheses around the variable. 

Unless I'm using it wrong here?

**Comment by gwojan** [comment]
@niphlod _"The $() notation is a full blown evaluation, and it works for everything ($($foo.bar)), the ${} instead is only for strings ?"_

No, it's an alternate form of variable declaration syntax. See this comment:

**Curly braces in PowerShell variable names allow for arbitrary characters in the name of the variable. If there are no "pathological" characters in the variable name, then the braces are not needed and have no effect.**

**You'll find that "generated" code will sometimes use curly braces because they guarantee that the variable name is valid.**

from [Do curly braces mean anything in PowerShell variable declaration?](https://stackoverflow.com/questions/36773800/do-curly-braces-mean-anything-in-powershell-variable-declaration)

Note you can also create variable name with spaces:

```
4199> ${My Name} = 'Bubba Gump'
PS C:\temp
4200> ${My Name}
Bubba Gump
```

The shell will even do tab completion for you...

niphlod:
ok, this I knew of. It talks about variable declaration, but `$NetworkShare` has not pathological characters ^_^
The `${}` solution here is supposed to close another deal, which is to evaluate the `$NetworkShare` variable within a string if it's followed by a character (in this case a `?`)  without a space in between.
In that deal, `${NetworkShare}?` and `$($NetworkShare)?` are good to use.

PR #2570, opened 2017-10-31
TypeExtension - Only update types the first time

FriedrichWeinmann:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Changes

 - Added conditional to the type extension (`.Query()` and `.Invoke()`) so it only runs on the first import.

Type extensions persist across the entire process until it terminates. Duplication errors were previously ignored (and thus wouldn't show up). However invoking dbatools in runspaces would report those errors back anyway. Thus this change prevents execution on consecutive imports.


potatoqualitee:
thanks to you both!

PR #2583, opened 2017-11-03
Fix for new database sizing not being reported when -LogsOnly is used

dgmyrs:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
When -LogsOnly is used, the returned object contains the old $db.Size and $db.SpaceAvailable because the $db object does not get a Refresh() run against it.

### Approach
Changed $db.LogFiles.Refresh() to $db.Refresh() which will refresh everything needed.


potatoqualitee:
Thanks so much for the catch! 👍

PR #2584, opened 2017-11-03
Added braces to variable to protect from special character

cporteou:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #2569 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Added braces to variable to protect from special character. Also made message punctuation consistent

### Commands to test
Copy-DbaDatabase

### Learning
@gwojan mentioned that use of curly braces protects the variable from special characters.

### DIFF
--- a/functions/Copy-DbaDatabase.ps1
+++ b/functions/Copy-DbaDatabase.ps1
@@ -282,7 +282,7 @@ function Copy-DbaDatabase {
 			$databaseProgressbar = 0
 			
 			foreach ($db in $databaseList) {
-				Write-Progress -Id 1 -Activity "Processing database file structure" -PercentComplete ($databaseProgressbar / $dbCount * 100) -Status "Processing $databaseProgressbar of $dbCount"
+				Write-Progress -Id 1 -Activity "Processing database file structure" -PercentComplete ($databaseProgressbar / $dbCount * 100) -Status "Processing $databaseProgressbar of $dbCount."
 				$dbName = $db.Name
 				Write-Message -Level Verbose -Message $dbName
 				
@@ -819,13 +819,13 @@ function Copy-DbaDatabase {
 			$fsWarning = $true
 		}
 		
-		Write-Message -Level Verbose -Message "Writing warning about filestream being enabled"
+		Write-Message -Level Verbose -Message "Writing warning about filestream being enabled."
 		if ($fsWarning) {
 			Write-Message -Level Verbose -Message "FILESTREAM enabled on $source but not $destination. Databases that use FILESTREAM will be skipped."
 		}
 		
 		if ($DetachAttach -eq $true) {
-			Write-Message -Level Verbose -Message "Checking access to remote directories"
+			Write-Message -Level Verbose -Message "Checking access to remote directories."
 			$remoteSourcePath = Join-AdminUNC $sourceNetBios (Get-SqlDefaultPaths -SqlInstance $sourceServer -filetype data)
 			
 			if ((Test-Path $remoteSourcePath) -ne $true -and $DetachAttach) {
@@ -1013,7 +1013,7 @@ function Copy-DbaDatabase {
 						$dropResult = Remove-SqlDatabase $destServer $dbName
 						
 						if ($dropResult -eq $false) {
-							Write-Message -Level Verbose -Message "Database could not be dropped. Aborting routine for this database"
+							Write-Message -Level Verbose -Message "Database could not be dropped. Aborting routine for this database'"
 							
 							$copyDatabaseStatus.Status = "Failed"
 							$copyDatabaseStatus.Notes = "Could not drop database"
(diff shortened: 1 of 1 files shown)


niphlod:
mmmh, special how ? why simple parenthesis are enough ? (newbie here, but curly are normally used with placeholder and -f )

gwojan:
@niphlod If you look at the code there was something like ```"... access to $ShareName?"``` and the ```?``` was being interpreted as part of the variable name resulting in a null value.

Using ```${...}``` syntax for a variable name is perfectly valid and many times necessary if you have special characters as part of the variable name. It actually allows you to do perverse things like ```${$variableName}``` where the inner **$** is part of the name. 😉 

This idiom is also very common in *nix shells -- I use it every day.

cporteou:
Hi @niphlod ,

The original problem was a trailing question mark (special character) causing the variable not to appear in the message. ie.

`"Backup Failed. Does SQL Server account $serviceAccount have access to '$NetworkShare? Aborting routine for this database."`

I fixed this by simply adding a space but @gwojan suggested the use of curly braces to allow the question mark to remain in place & thus not add an unexpected space in the message.

`... ${NetworkShare}?"`

Adding parenthesis wouldn't help this as it would invalidate the variable or add a needless set around the message
eg: 
`... ($NetworkShare)?"`

would result in:
**Backup Failed. Does SQL Server account USERNAME have access to (\\Share\Folder\)? Aborting routine for this database.**

But there are other methods to fix the issue. Single quotes would have also worked on their own & were only added for better readability of the network share's name. The curly brackets also allow you to use spaces and special characters inside a variable name which isn't relevant here though. 

`... '$NetworkShare'?"`

niphlod:
yeah, sorry, I just posted a comment over there. I'm not stopping the merge nor not approving or requiring changes .... It's a matter of personal preference coupled with a not-so-deep knowledge of all ways to string-interpolate-expand powershell offers. 
I got the point of the error, and I see this solves it, I'm just arguing that it's probably the first occurrence of the `${}` notation in the module and that a "commonly used" `$($NetworkShare)` notation would have sufficed.

functions/Copy-DbaDatabase.ps1
```diff
@@ -1214,7 +1214,7 @@ function Copy-DbaDatabase {
 								Write-Message -Level Verbose -Message "Successfully updated DatabaseOwnershipChaining for $sourceDbOwnerChaining on $dbName on $destination."
 							}
 							catch {
-								$copyDatabaseStatus.Status = "Successful - failed to apply DatabaseOwnershipChaining."
+								$copyDatabaseStatus.Status = "Successful - failed to apply DatabaseOwnershipChaining"
```
wsmelton: The period needs to be put back, not sure why it was removed.

functions/Copy-DbaDatabase.ps1
```diff
@@ -1061,15 +1061,15 @@ function Copy-DbaDatabase {
 						$backupResult = $BackupTmpResult.BackupComplete
 						if ($backupResult -eq $false) {
 							$serviceAccount = $sourceServer.ServiceAccount
-							Write-Message -Level Verbose -Message "Backup Failed. Does SQL Server account $serviceAccount have access to $NetworkShare ? Aborting routine for this database."
+							Write-Message -Level Verbose -Message "Backup Failed. Does SQL Server account $serviceAccount have access to '${NetworkShare}'? Aborting routine for this database."
 							
 							$copyDatabaseStatus.Status = "Failed"
-							$copyDatabaseStatus.Notes = "Backup failed. Verify service account access to $NetworkShare"
+							$copyDatabaseStatus.Notes = "Backup failed. Verify service account access to '${NetworkShare}'."
```
wsmelton: In these cases we have been using `$($variable)` and not curly braces. If you can use that same pattern it would help keep to the standard.

functions/Copy-DbaDatabase.ps1
```diff
@@ -1013,7 +1013,7 @@ function Copy-DbaDatabase {
 						$dropResult = Remove-SqlDatabase $destServer $dbName
 						
 						if ($dropResult -eq $false) {
-							Write-Message -Level Verbose -Message "Database could not be dropped. Aborting routine for this database"
+							Write-Message -Level Verbose -Message "Database could not be dropped. Aborting routine for this database'"
```
wsmelton: This should have `this database."` instead of `this database'"`

cporteou: Yes, that should be a period. Sorry

functions/Copy-DbaDatabase.ps1
cporteou: Perfect, no problem.

functions/Copy-DbaDatabase.ps1
cporteou: I think other status messages lacked the period. I'll add back though. thanks

PR #2602, opened 2017-11-07
Update Export-DbaScript to support servers with creds

ConstantineK:
Changed server name to server object

## Type of Change

### Purpose
 
Allows servers with credentials to export data

### Approach

Updates the scripter object to use the server object instead of the server name.


potatoqualitee:
awesome, thanks, works!

PR #2618, opened 2017-11-09
Changed save method to work

ConstantineK:
Updated to use io.file to write a bytestream instead of out-file, tested, worked great.

## Type of Change

### Purpose
 
Update files to show data instead of a series of numbers

### Approach

Change t io.write because I forget how to do that in out-file


potatoqualitee:
thank youuuu!

PR #2633, opened 2017-11-11
[New Command] Get-DbaWaitingTask

wsmelton:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Provides method for DBAs to get information on waiting task via script shared by Paul Randal, querying `sys.dm_os_waiting_tasks`.
### Approach

Just the query shared by Paul (referenced in help for blog post). I added a few filters similar to what `Get-DbaWaitStatistics` offered.
### Commands to test
The test for this was a pain to get working right. Per recommendation from Fred it uses `Start-Job` so I can have a known session running, something I know will be returned by the command. 

The big fix to get the test to run was adding `Start-Sleep` and I had to find the right value in seconds for the job to load the module and then actually run the command to have a session.

Since the test requires that I import the psm1 file, since this command is not part of the manifest yet, I had to hard code what I think is the path to the file. **I need to know the preferred method we have for pulling this path, or have it added to the constants file so I can reference it.**

image

### DIFF
--- a/functions/Get-DbaWaitingTask.ps1
+++ b/functions/Get-DbaWaitingTask.ps1
@@ -10,7 +10,7 @@ function Get-DbaWaitingTask {
 		.PARAMETER SqlInstance
 			The SQL Server instance. Server version must be SQL Server version XXXX or higher.
 
-        .PARAMETER SqlCredential
+		.PARAMETER SqlCredential
 			Allows you to login to servers using SQL Logins instead of Windows Authentication (AKA Integrated or Trusted). To use:
 
 			$scred = Get-Credential, then pass $scred object to the -SqlCredential parameter.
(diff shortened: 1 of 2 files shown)


potatoqualitee:
Yessssssss!

potatoqualitee:
thanks so much shawn! I'm super excited about this command, thanks. My only issue with the test is that it uses jobs and will add at least 20 extra seconds (to start a whole new process then import dbatools). Is there a way to avoid the use of jobs or no? (I'll likely merge but if you can update the test if it is possible, that'd be appreciated)

wsmelton:
No, I talked to @FriedrichWeinmann and that was his recommendation on how to deal with it. I don't know if runspaces would do any better, or if they would be handled properly by Pester.

FriedrichWeinmann:
Runspaces should work for this and cut down on time needed. The only issue with doing it in runspaces if you run too many connections against the same instance from the same process (which is ... unlikely for your test :) ), due to the way the connection pool is managed by SMO.

You just have to kick the query call out of the main runspace, as the runspace that executes it will be hanging, waiting for an answer ... so, either job or new runspace will do.

PR #2636, opened 2017-11-11
[New Command] Remove-DbaDbUser

dgmyrs:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
New Command Remove-DbaDbUser

### Approach
Users without any owned objects will be removed.  Users with a schema of the same name as the user that doesn't own objects will also be removed.  Users that own a schema of the same name as the user that also has objects owned by the schema will not be removed.  Users that own a schema of another name that has owned objects will only be removed after the schema owner is changed to dbo, and only if the -Force param is used.

### DIFF
--- a/functions/Remove-DbaDbUser.ps1
+++ b/functions/Remove-DbaDbUser.ps1
@@ -1,0 +1,230 @@
+function Remove-DbaDbUser {
+<#
+    .SYNOPSIS
+    Drop database user
+
+    .DESCRIPTION
+    If user is the owner of a schema with the same name and if if the schema does not have any underlying objects the schema will be
+    dropped.  If user owns more than one schema, the owner of the schemas that does not have the same name as the user, will be
+    changed to 'dbo'. If schemas have underlying objects, you must specify the -Force parameter so the user can be dropped.
+       
+    .PARAMETER SqlInstance
+    The SQL Instances that you're connecting to.
+
+    .PARAMETER SqlCredential
+    Credential object used to connect to the SQL Server as a different user.
+
+    .PARAMETER Database
+    Specifies the database(s) to process. Options for this list are auto-populated from the server. If unspecified, all databases will be processed.
+
+    .PARAMETER ExcludeDatabase
+    Specifies the database(s) to exclude from processing. Options for this list are auto-populated from the server.
+
+    .PARAMETER User
+    Specifies the list of users to remove.
+        
+    .PARAMETER UserCollection
+    Internal parameter to support piping from Get-DbaDatabaseUser.
+
+    .PARAMETER Force
+    If enabled this will force the change of the owner to 'dbo' for any schema which owner is the User.
+
+    .PARAMETER WhatIf
+    Shows what would happen if the command were to run. No actions are actually performed.
+
+    .PARAMETER Confirm
+    If this switch is enabled, you will be prompted for confirmation before executing any operations that change state.
+
+    .PARAMETER EnableException
+    By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
+    This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
+    Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
+        
+    .NOTES
+    Tags: Databases, User
+    Author: Doug Meyers (@dgmyrs)
+
+    Website: https://dbatools.io
+    Copyright: (C) Chrissy LeMaire, clemaire@gmail.com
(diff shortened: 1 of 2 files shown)


potatoqualitee:
fantastic Test!!!! excellent work, great command as well, thank you. You even added pipeline support! so good. @ClaudioESSilva - should we modify Remove-DbaOrphanUser - make it Find, return SMO objects then we can just pipe it to this command?

functions/Remove-DbaDbUser.ps1
```diff
@@ -0,0 +1,230 @@
...
+                            }
+                        }
+                    }
+
+                    # Perform the changes with the owner alters first so function can exit w/o a drop if -Force wasn't used
+                    foreach ($schema in $alterSchemas) {
+                        Write-Message -Level Verbose -Message "Owner of Schema $schema will be changed to [dbo]."
+                        if ($PSCmdlet.ShouldProcess($server, "Change the owner of Schema $schema to [dbo].")) {
+                            $schema.Owner = "dbo"
+                            $schema.Alter()
```
wsmelton: This needs to be in a try/catch block

functions/Remove-DbaDbUser.ps1
```diff
@@ -0,0 +1,230 @@
...
+                        }
+                    }
+
+                    foreach ($schema in $dropSchemas) {
+                        if ($PSCmdlet.ShouldProcess($server, "Drop Schema $schema from Database $db.")) {
+                            $schema.Drop()
```
wsmelton: Try/catch block

functions/Remove-DbaDbUser.ps1
```diff
@@ -0,0 +1,230 @@
...
+
+                        # Change the owner of any schema not the same name as the user
+                        if ($schema.Name -NE $user.Name) {
+                            # Check for owned objects early so we can exit before any changes are made
+                            $ownedUrns = $schema.EnumOwnedObjects()
+                            if (($ownedUrns -And $Force) -Or (-Not $ownedUrns)) {
+                                $alterSchemas += $schema
+                            } else {
+                                Write-Message -Level Warning -Message "User $user owns the Schema $schema, which owns $($ownedUrns.Count) Object(s).  If you want to change the schemas' owner to [dbo] and drop the user anyway, use -Force parameter.  User $user will not be removed."
+                                return  #exit the function
```
wsmelton: I would believe this should be a Stop-Function call with a continue and not a return. If more than one schema is passed in wouldn't this exit the function even if there were additional schemas in the collection?

functions/Remove-DbaDbUser.ps1
```diff
@@ -0,0 +1,230 @@
...
+                        $user.Drop()
+                        [pscustomobject]@{
+                            SqlInstance = $server.name
+                            Database    = $db.name
+                            User        = $user
+                            Status      = "Dropped"
+                        }
+                    } catch {
+                        Write-Message -Level Verbose -Message "Could not drop $user from Database $db on target $server"
+                        [pscustomobject]@{
```
wsmelton: Need to adjust the `SQLInstance` value and then also add the Computer and Instance name:

```powershell
ComputerName             = $server.NetName
InstanceName             = $server.ServiceName
SqlInstance              = $server.DomainInstanceName
```

functions/Remove-DbaDbUser.ps1
```diff
@@ -0,0 +1,230 @@
...
+                
+                if ($Database) {
+                    $databases = $databases | Where-Object Name -In $Database
+                }
+                if ($ExcludeDatabase) {
+                    $databases = $databases | Where-Object Name -NotIn $ExcludeDatabase
+                }
+                
+                foreach ($db in $databases) {
+                    Write-Message -Level Verbose -Message "Get users in Database $db on target $server"
```
wsmelton: Need to just filter out databases that are not accessible on line 211 or validate if it is accessible here. I think our pattern is to filter them out when we build the `$databases` variable in most commands.

functions/Remove-DbaDbUser.ps1
dgmyrs: Yes, it would exit the function.  My thinking was to do as little change as possible if it wasn't going to be able to drop the user anyway.  Stop-Function is new to me, so if that's the standard, can certainly switch to it.

dgmyrs: OK, I see there was still an issue in that it would exit even if it had more than one user to go through.  I refactored the code, see PR 2647: https://github.com/sqlcollaborative/dbatools/pull/2647

Issue #2638, opened 2017-11-12
Dbatools Config - New feature request for debugging mode


wsmelton:
## Feature Request

Noticed in the `Get-DbaConfig` output there is now a developer module configuration. The only one listed is `developer.mode.enable` which appears to be around the logging/message system.

### Problem to solve
We currently have the variable `$dbatools_dotsourcemodule` to utilize for having the code dot sourced into our current session. This was added some time back to appease those of that want to debug the module using PowerShell debug engine.

Can we have this variable added into the configuration? Either one of two options would suite me:

1. Just have it be part of the current `developer.mode.enable` so when that is enabled it adds the advanced logging **AND** ensures code is dot sourced.
2. Add another configuration under the developer module as something like `developer.mode.dotsource` maybe?

FriedrichWeinmann:
Heya, it's technically doable, but might be a bit unintuitive.
You see, module import happens _before_ the configuration system is ready to rock.

I can work around that though and it won't affect performance.

FriedrichWeinmann:
And there you go: https://github.com/sqlcollaborative/dbatools/pull/2639
Added a now configuration scope though, since those are the import options after all (even though some of them are useful to developers only)

PR #2639, opened 2017-11-12
Added configuration for module import options

FriedrichWeinmann:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Changes

 - Bugfix: `Set-DbaConfig` was running the handler before applying changes, rather than after. This was causing issues with registering changes as part of the handler.
 - New feature: Added configurations for the import options dbatools offers

### Notes

Those new configuration settings automatically register themselves in registry (local user scope) so they persist across sessions. This is necessary, as the configuration system is loaded during the import itself obviously, thus setting this will only affect the next import after that.

Implements https://github.com/sqlcollaborative/dbatools/issues/2638

### DIFF
--- a/functions/Set-DbaConfig.ps1
+++ b/functions/Set-DbaConfig.ps1
@@ -278,21 +278,20 @@ function Set-DbaConfig
 				}
 				$Value = $testResult.Value
 			}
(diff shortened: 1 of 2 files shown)


potatoqualitee:
thank youuuuuuu

PR #2644, opened 2017-11-13
New Function - Get-DbaSqlDump

gbargsley:
## Type of Change

 - [X] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Find a memory dump on one or multiple SQL instances.

### Screenshots

image

### DIFF
--- a/functions/Get-DbaSqlDump.ps1
+++ b/functions/Get-DbaSqlDump.ps1
@@ -1,0 +1,116 @@
+function Get-DbaSqlDump {
+	<#
+		.SYNOPSIS
+			Locate a SQL Server that has generated any memory dump files.
+  
+		.DESCRIPTION
+			The type of dump included in the search include minidump, all-thread dump, or a full dump.  The files have an extendion of .mdmp.
+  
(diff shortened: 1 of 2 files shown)


potatoqualitee:
this looks fabulous! thank you for the test too <3

Can you remove the warning - we can just return null. Also, can you please change the name of size_in_bytes to Size then do `[dbasize]($result.size_in_bytes)` - that makes it pretty

potatoqualitee:
in addition, we're starting to move away from -Detailed. Can you remove that Param and just return all of the detailed by default? 🙇

gbargsley:
@potatoqualitee 
I added the size conversion.

You want me to take the warning out about no memory dumps found?

Also, if you want me to remove detailed param, then I will just show the output only and not worry about a count?

potatoqualitee:
Correct! We've got some folks excited about this command, so I'll merge and patch. Thanks so much! We'll be adding this to a new project we're working on as well 💃

PR #2685, opened 2017-11-17
Remove (comment out) call to Connect() method

wsmelton:
Discussed in Slack channel on issues with commands like `Get-DbaProcess` and others in the module causing two issues:

1. Multiple connections being generated (on average you get two for one call to the command)
2. The connections associated with the command itself (not including TEPP connection) does not "close" until the PowerShell session is closed.

Various areas of SMO docs point to two things that we do that will cause the above issues:

- Specifically calling `Connect()` will not allow that connection to be put back into the connection pool when it is done.
- Specifically calling `Connect()` will not automagically close the connection for you unless you explicitly call `Close()` or `Dispose()` on the `$server` object.

_Note each test below was done using a fresh PowerShell session._

## Testing 1

Just running `Get-DbaProcess` causes two connections to be made:
1. Is connection associated with call to `Connect-SqlInstance` by the command itself
2. Is from TEPP implementation and is known non-issue connection.

Leaving the above running the connection associated with populating the TEPP cache will be killed in 5-6 minutes. The connection associated with the call to the connect command will not be killed until your PowerShell session exits.

## Testing 2

Running:

```powershell
for ($i = 0; $i -le 50; $i++) { Write-Output "Iteration #: $i"; $null = Get-DbaProcess -SqlInstance manatarms\sql12}
```

Causes 18 connections to be generated, and all those connections don't go away until I close the PowerShell session.

## Testing 3

Implement changes from this PR and run command from test 1 again. Result: Only two connections are generated and both are killed within a 6 minute period.

## Testing 4

Implement changes from this PR and run command from test 2 again. Result: Only two connections are generated and both are killed within a 6 minute period.

🎉 🎈 🐎 🐎

### DIFF
--- a/internal/Connect-SqlInstance.ps1
+++ b/internal/Connect-SqlInstance.ps1
@@ -109,9 +109,9 @@ function Connect-SqlInstance {
 	#region Input Object was a server object
 	if ($ConvertedSqlInstance.InputObject.GetType() -eq [Microsoft.SqlServer.Management.Smo.Server]) {
 		$server = $ConvertedSqlInstance.InputObject
-		# if ($server.ConnectionContext.IsOpen -eq $false) {
-		# 	$server.ConnectionContext.Connect()
-		# }
+		 if ($server.ConnectionContext.IsOpen -eq $false) {
+		 	$null = $server.Name
+		 }
 
 		# Register the connected instance, so that the TEPP updater knows it's been connected to and starts building the cache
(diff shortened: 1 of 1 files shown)


potatoqualitee:
I would *LOVE* to merge this. I believe it will solve some problems, just gotta look into these couple failures.

wsmelton:
The puzzle is why are they not the same error for both branch and PR?

potatoqualitee:
frikken bon, ppl gone love this thanks @wsmelton !

PR #2687, opened 2017-11-17
Deprecate -LogsOnly and replace with -FileType

### LINKED ISSUE #2590: Change that -LogsOnly to Type = All, Data, Log
For Invoke-DbaDatabaseShrink and deprecate LogsOnly

dgmyrs:
## Type of Change

 - [X] Bug fix (non-breaking change, fixes #2590)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Deprecate -LogsOnly, and replace with -FileType of either All, Data, or Log

### Approach

All and Log use existing functionality.  I added a section for Data that shrinks every data file in the database.

### Commands to test

Included in the test file.

### DIFF
--- a/functions/Invoke-DbaDatabaseShrink.ps1
+++ b/functions/Invoke-DbaDatabaseShrink.ps1
@@ -55,7 +55,7 @@ Specifies the method that is used to shrink the database
 Timeout in minutes. Defaults to infinity (shrinks can take a while.)
 
 .PARAMETER LogsOnly
-Only shrink the log file, not the entire database
+Deprecated. Use FileType instead
 
 .PARAMETER FileType
 Specifies the files types that will be shrunk
@@ -137,7 +137,7 @@ Shrinks all databases on SQL2012 (not ideal for production)
 
 	begin {
 		if ($LogsOnly) {
-			Write-Message -Level Warning -Message "LogsOnly is deprecated and will be removed in dbatools 1.0."
+			Test-DbaDeprecation -DeprecatedOn "1.0.0" -Parameter "LogsOnly"
 			$FileType = 'Log'
 		}
 
--- a/functions/Invoke-DbaDatabaseShrink.ps1
+++ b/functions/Invoke-DbaDatabaseShrink.ps1
@@ -57,6 +57,16 @@ Timeout in minutes. Defaults to infinity (shrinks can take a while.)
 .PARAMETER LogsOnly
 Only shrink the log file, not the entire database
 
+.PARAMETER FileType
+Specifies the files types that will be shrunk
+
+		All
+			All Data and Log files are shrunk, using database shrink (Default)
+		Data
+			Just the Data files are shrunk using file shrink
+		Log
+			Just the Log files are shrunk using file shrink
+
 .PARAMETER ExcludeIndexStats
 Exclude statistics about fragmentation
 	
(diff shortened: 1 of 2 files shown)


potatoqualitee:
OUU YESSS! I've been wanting this, thank you. Gonna merge when it passes.

functions/Invoke-DbaDatabaseShrink.ps1
```diff
@@ -117,13 +127,20 @@ Shrinks all databases on SQL2012 (not ideal for production)
 		[int]$PercentFreeSpace = 0,
 		[ValidateSet('Default', 'EmptyFile', 'NoTruncate', 'TruncateOnly')]
 		[string]$ShrinkMethod = "Default",
+		[ValidateSet('All', 'Data', 'Log')]
+		[string]$FileType = "All",
```
wsmelton: Need to add `LogsOnly` as an alias to this parameter if we are removing it. This prevents commands from failing and does not require us to keep the parameter on the list and help for it.

potatoqualitee:
You are on 🔥 ! Thanks for updating the Test too.

functions/Invoke-DbaDatabaseShrink.ps1
dgmyrs: @wsmelton wouldn't that be a bit confusing since LogsOnly is a switch, and FileType is a string?

potatoqualitee: ahh yeah, good point, Shawn. It's our deprecation feature, @dgmyrs where it doesnt throw an error and it just says that its deprecated. brb ill put a lil patch

potatoqualitee: oh, @dgmyrs already has most of it :D agreed with Doug in his approach beecause of the switch vs string and he takes care of it within the code.

dgmyrs:
Ah, so that's how that works :) .  As an FYI, I used Test-DbaDiskAllocation as and example, so looks like there are a few out there using an old method.  The ones I found in a search are Test-DbaDiskAllocation, Test-DbaPowerPlan, and Test-DbaValidLogin.

functions/Invoke-DbaDatabaseShrink.ps1
wsmelton: It won't matter in usage @dgmyrs because we just want the name alias so it "points" to the new parameter, the type is irrelevant in this case.

wsmelton: Using the alias method allows us to go ahead and update the help documentation now. It is one less thing to deal with when we have to go through all the commands and remove the deprecated messages.

functions/Invoke-DbaDatabaseShrink.ps1
```diff
@@ -55,7 +55,7 @@ Specifies the method that is used to shrink the database
 Timeout in minutes. Defaults to infinity (shrinks can take a while.)
 
 .PARAMETER LogsOnly
-Only shrink the log file, not the entire database
+Deprecated. Use FileType instead
```
wsmelton: Just note, we are not doing this with any other command 😸

potatoqualitee: One that i updated the other day ;D It's required for params since you cant remove it

PR #2688, opened 2017-11-17
Login to SMO Output

potatoqualitee:


nvarscar:
👍 
PS There is $newLogin variable that contains the SMO object already, no need to call Get-DbaLogin.
```
$newLogin.Refresh()
$newLogin
```

potatoqualitee:
Get-DbaLogin makes it pretty 💅

potatoqualitee:
Can you plz help with my failing tests 😂

nvarscar:
Looking into it 👀 
For some reason it fails to return objects in the same order, it seems. I wonder if it's "by design".

potatoqualitee:
thank you 💃  about to publicize it on twitter if we can get it merged within 2h rs

nvarscar:
Hopefully it should be good now. There is a renaming pattern in that test that swap login names - `-LoginRenameHashtable @{claudio = 'port'; port = 'claudio'}`, it expected the logins to appear in a different order.

potatoqualitee:
passed!

PR #2698, opened 2017-11-18
Get-DbaDatabaseFile add next growth

PowerDBAKlaas:
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Purpose
show the size of the next growth event in a friendly format

### Approach
@josh-simar added the calculations
@powerdbaklaas added a check to compare with available space

### DIFF
--- a/functions/Get-DbaDatabaseFile.ps1
+++ b/functions/Get-DbaDatabaseFile.ps1
@@ -203,7 +203,7 @@ function Get-DbaDatabaseFile {
  					else {
  						$nextgrowtheventadd = [dbasize]($result.Growth * 8 * 1024)
  					}
-                    if ( $nextgrowtheventadd.Byte -lt $AvailableSpace.Byte ) { [dbasize]$nextgrowtheventadd = 0 }
+                    if ( ($nextgrowtheventadd.Byte -gt ($MaxSize.Byte - $size.Byte)) -and $maxsize -gt 0 ) { [dbasize]$nextgrowtheventadd = 0 }
 					
 					[PSCustomObject]@{
 						ComputerName             = $server.NetName
--- a/functions/Get-DbaDatabaseFile.ps1
+++ b/functions/Get-DbaDatabaseFile.ps1
@@ -203,6 +203,7 @@ function Get-DbaDatabaseFile {
  					else {
  						$nextgrowtheventadd = [dbasize]($result.Growth * 8 * 1024)
  					}
+                    if ( $nextgrowtheventadd.Byte -lt $AvailableSpace.Byte ) { [dbasize]$nextgrowtheventadd = 0 }
 					
 					[PSCustomObject]@{
 						ComputerName             = $server.NetName


josh-simar:
@PowerDBAKlaas  Should that growth greater check not be greater than the way it is written?

wsmelton:
It does not return an accurate value...At most by default if growth is set to a static value, the next growth event should be showing that size value.

image

It would only be where growth is set to no value, or auto growth is disabled that the next growth event would show zero.

image

If interested this could be more easily done via adjusting the T-SQL than trying to get the right login in PowerShell:
[_Database_NextGrowthIncrement - Copy.txt](https://github.com/sqlcollaborative/dbatools/files/1485353/_Database_NextGrowthIncrement.-.Copy.txt)

PowerDBAKlaas:
oops
I made a lot of mistakes in one line:
+ I assumed availablespace was up to maxsize, but it is actually the unusedspace. I believe we named this differently in other functions.
+ I forgot to check for unlimited maxsize
+ I checked with `-lt` while it should obviously be `-gt`, you're right @josh-simar 

This should all be corrected now.
@wsmelton I'll look at the T-SQL version tomorrow, at work.

PowerDBAKlaas:
After a bit of examination, I think we can indeed move all calculations inside a T-SQL statement.
I've also had a look at `Get-DbaDatabaseSpace`. If we include the last 3 properties from that function in here, that one becomes totally redundant.
Should that be a goal for 1.0? @wsmelton @potatoqualitee @ck @Stuart-Moore ?

potatoqualitee:
hey @PowerDBAKlaas, thanks so much for all of your work! I see what you're saying about the redundancy but these are two things that people will look for (Space perhaps moreso than File) so I'd like to keep them, even with the overlap, as the name is just intuitive for both. I'd like to merge these fixes then get another PR for moving to T-SQL, does that work for everyone?

wsmelton:
Yeah, been thinking about that a bit the last day or so. Compare with T-SQL folks match up `Get-DbaDbSpace` with running something like `exec sp_spaceused` and running something different to get file info on a database, even though the latter can give you both.

I'd say in this specific situation it would be best to keep them both.

PowerDBAKlaas:
Well, in the beginning there was overlap, but after all modifications both functions are exactly the same.
If we keep `Get-DbaDbSpace` it could better return the DB totals instead of files.

potatoqualitee:
Agreed there! Let's change both the name and the output - returning totals instead of files (most of the time, I care about totals, not files, and wouldnt like to have to sum manually)

potatoqualitee:
Would you be up for doing that bit this week @PowerDBAKlaas ?

PowerDBAKlaas:
Actually, 'Size', 'Dataspaceusage','IndexSpaceUsage','AvailableSpace' and 'MaxSize' are in `Get-DbaDatabase` too. So I'm back at "`Get-DbaDatabaseSpace` is completely redundant.

Anyway, that has nothing to do with this PR: this is improving the DBFile function. This one can be merged, can't it?

potatoqualitee:
Oh, oops! You're right. My apologies.

potatoqualitee:
merged 😌

PR #2699, opened 2017-11-18
Added option for monthly interval

sanderstad:
The monthly interval wasn't possible.
Updated the help with the other values for the parameters and another
example.


sanderstad:
Proof that it works 
image

Issue #2485, opened 2017-10-22
Adjusting Get-DbaRoleMember


wsmelton:
## Feature Request

[This command](https://github.com/sqlcollaborative/dbatools/blob/development/functions/Get-DbaRoleMember.ps1) does not allow you to **only** pull server-level role members. Is there any reason why we created it in this manner?

If I just want sysadmin role members it still has to run through all the databases and their role members. On a server with 1 database this may work but servers that have 100+ where I just want to verify who is in sysadmin it does not.

### Problem to solve

I want to audit an instance's server-level role(s) or even just test existence of a login in a specific server-level role(s).

Option 1:

- Add additional parameters to the current command to `-IgnoreDatabaseRoles` or a `-ServerRole` that if that is server-level role is provided **it does not** parse the roles in a database.

Option 2:

- Rename the current command to `Get-DbaDbRoleMember`
- Create a new command `Get-DbaServerRoleMember`

PowerDBAKlaas:
@wsmelton I can't think of a valid reason. It probably got lost in the -Include... vs -No... vs -Exclude... vs -...Only vs ... strategy changes.
I think it's best to add a switch to the existing function.

wsmelton:
I guess my next question would be then why do we have one command that touches two different objects?

If we build out a command such as `Get-DbaRole` then it would be assumed that should be allowed to pipe to `Get-DbaRoleMember`, which in the current state would seem impossible to do since that role member command is focused on database roles.

PowerDBAKlaas:
Well, I didn't consider the members to be objects. It's the result of the `Enummember()` and `EnumServerRoleMembers()` methods and PSCustomObject at both levels. Since I didn't see any other use for those than display, I thought it most practical to put them together.

I don't know if I misjudged that and it would be possible to create 'useful' objects after all?

Feeding the roles through the pipeline would indeed change the strategy, unless those could be accepted as strings and handled in an `-IncludeRole` kind of way.

wsmelton:
Only thing that has stirred my interest is this issue #2441 but also with the AG commands I need to be able to get server-level roles and adjust them to build an AG.

So I guess we just need to decide how we want to move forward with it.

PowerDBAKlaas:
The current method could be `(Get-DbaRoleMember -... -IncludeServerLevel).Where({$_.Database -eq $Null})`
[...]

PR #2700, opened 2017-11-19
New Command: Get-DbaServerRole (initial version)

wsmelton:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system

Addresses issues noted in #2485
 

### Purpose
 
Provide method of pulling server-level role information, including members.
### Commands to test

image
image

### DIFF
--- a/functions/Get-DbaServerRole.ps1
+++ b/functions/Get-DbaServerRole.ps1
@@ -1,0 +1,119 @@
+function Get-DbaServerRole {
+	<#
+		.SYNOPSIS
+			Gets list of server-level roles and members
+
+		.DESCRIPTION
+			The SQL Server instance. Server version must be SQL Server version 2005 or higher.
+
+		.PARAMETER SqlInstance
+			SQL Server name or SMO object representing the SQL Server to connect to. This can be a collection and recieve pipeline input to allow  the function to be executed against multiple SQL Server instances.
+
+		.PARAMETER SqlCredential
+			Allows you to login to servers using SQL Logins instead of Windows Authentication (AKA Integrated or Trusted). To use:
+
+			$scred = Get-Credential, then pass $scred object to the -SqlCredential parameter.
+
+			Windows Authentication will be used if SqlCredential is not specified. SQL Server does not accept Windows credentials being passed as credentials.
+
+			To connect as a different Windows user, run PowerShell as that user.
+
+		.PARAMETER EnableException
+			By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message. This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting. Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
+
+		.NOTES
+			Tags: ServerRole, Security
+			Original Author: Shawn Melton (@wsmelton)
+
+			Website: https: //dbatools.io
+			Copyright: (C) Chrissy LeMaire, clemaire@gmail.com
+			License: GNU GPL v3 https: //opensource.org/licenses/GPL-3.0
+
+		.LINK
+			https://dbatools.io/Get-DbaServerRole
+
+		.EXAMPLE
+			Get-DbaServerRole -SqlInstance sql2016a
+
+			Outputs list of server-level roles and logins that are members for sql2016a instance.
+
+		.EXAMPLE
+			Get-DbaServerRole -SqlInstance sql2016a -Role sysadmin
+
+			Outputs members of sysadmin server-level role on sql2016a instance.
+	#>
+	[CmdletBinding()]
+	param (
+		[Parameter(Position= 0, Mandatory= $true, ValueFromPipeline= $true)]
+		[DbaInstance[]]$SqlInstance,
+		[PSCredential]$SqlCredential,
+		[object[]]$ServerRole,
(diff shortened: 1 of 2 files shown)


functions/Get-DbaServerRole.ps1
```diff
@@ -0,0 +1,139 @@
+function Get-DbaServerRole {
+	<#
+		.SYNOPSIS
+			Gets the list of server-level roles with the logins that are members of that role.
+
+		.DESCRIPTION
+			Gets the list of server-level role for SQL Server instance. Output will include the logins that are members of the server-level role(s).
```
PowerDBAKlaas: should be plural: roles

functions/Get-DbaServerRole.ps1
```diff
@@ -0,0 +1,139 @@
...
+
+		.EXAMPLE
+			Get-DbaServerRole -SqlInstance sql2016a -Role sysadmin
+
+			Outputs members of sysadmin server-level role on sql2016a instance.
+
+		.EXAMPLE
+			Get-DbaServerRole -SqlInstance sql2017a -ExcludeFixedRole
+
+			Outputs the user server-level role(s) on sql2017a instance. A SQL Server 2017 supported feature only.
```
PowerDBAKlaas: I'm no sure 'user  server-level role' is the right expression. IIRC 'fixed roles' is not the same as 'system objects'.

functions/Get-DbaServerRole.ps1
```diff
@@ -0,0 +1,139 @@
...
+			}
+
+			foreach ($role in $roles) {
+				$members = $role.EnumMemberNames()
+				if ($Login) {
+					$members = $members | Where-Object {$_ -in $Login}
+				}
+
+				if ($members) {
+					foreach ($member in $members) {
```
PowerDBAKlaas: ??
if the role has 7 members, you add ComputerName, InstanceName, SqlInstance and Login 7 times?

functions/Get-DbaServerRole.ps1
wsmelton: 2017 allows you to create custom roles. You only know that they are custom roles based on the IsFixedRole being false.

functions/Get-DbaServerRole.ps1
wsmelton: Correct. It will generate 7 objects.

wsmelton: Screen shot shows what this looks like for sysadmin role.

functions/Get-DbaServerRole.ps1
PowerDBAKlaas: yes, but [public] is not a fixed role, yet I wouldn't call that a 'user role'

functions/Get-DbaServerRole.ps1
PowerDBAKlaas: That definitely belongs in another function:
from `Get-DbaServerRole` I expect 1 object for each role.
This output is `RoleMember`

wsmelton: Even though your output of role member function will not be an SMO object? Why would we need an extra function to just output strings?

functions/Get-DbaServerRole.ps1
wsmelton: Public is a fixed role.

wsmelton: https://docs.microsoft.com/en-us/sql/relational-databases/security/authentication-access/server-level-roles#fixed-server-level-roles

functions/Get-DbaServerRole.ps1
wsmelton: Enumerate for members of server level roles is part of the Roles object. It is not a seperate object in SMO. So not sure I follow the logic of making it a seperate function.

wsmelton: As well, how do you see use of two commands? It would be extra overhead to pipe the same object to pull the members when they are already in the role object. I have no reason to call a RoleMember function when I can just as easily call the enumerate method that is returned from the server role command.

wsmelton: How about adding it as a collection property?

image

wsmelton: Just adding the Login as a collect equates to doing the same thing with the `EnumMemberNames()`

image

functions/Get-DbaServerRole.ps1
PowerDBAKlaas: strange:

image

functions/Get-DbaServerRole.ps1
PowerDBAKlaas: that is acceptable, 
although this is also possible without adding the property:
`$r.EnumServerRoleMembers()` is already available.

PowerDBAKlaas: The motivation for having `Get-DbaRoleMember` as a separate function is
1. It's not for you and me, but for DBAs who don't know how to enumerate members and consequently handle that output.
2. It's not correct to return members as multiplied Role objects.

We certainly need to examine the `RoleMember` function(s). The existing one can be improved, no doubt. We have to decide whether we combine or split the Database and Server Role members. Also if this could include group memberships and so on. Functionality that can make it more useful, even for you and me.
But that's for later.

PR #2704, opened 2017-11-20
Add try/catch to Start and Stop DbaXESession functions

dgmyrs:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Add try/catch to Stop/Start XESession

### Approach

Write-Error on catch, and return Get-DbaXESession in either case.

### Commands to test

Included in Tests file

### DIFF
--- a/functions/Start-DbaXESession.ps1
+++ b/functions/Start-DbaXESession.ps1
@@ -89,7 +89,7 @@ function Start-DbaXESession {
 					try {
 						$xe.Start()
 					} catch {
-						Write-Error -Message "Could not start XEvent Session $session on $instance"
+						Stop-Function -Message "Could not start XEvent Session on $instance" -Target $session -ErrorRecord $_
 					}
 				} else {
 					Write-Message -Level Warning -Message "$session on $instance is already running"
--- a/functions/Start-DbaXESession.ps1
+++ b/functions/Start-DbaXESession.ps1
@@ -89,7 +89,7 @@ function Start-DbaXESession {
 					try {
 						$xe.Start()
 					} catch {
-						Stop-Function -Message "Could not start XEvent Session on $instance" -Target $session -ErrorRecord $_
+						Stop-Function -Message "Could not start XEvent Session on $instance" -Target $session -ErrorRecord $_ -Continue
 					}
 				} else {
 					Write-Message -Level Warning -Message "$session on $instance is already running"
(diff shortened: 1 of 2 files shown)


functions/Start-DbaXESession.ps1
```diff
@@ -86,11 +86,15 @@ function Start-DbaXESession {
 				$session = $xe.Name
 				if (-Not $xe.isRunning) {
 					Write-Message -Level Verbose -Message "Starting XEvent Session $session on $instance."
-					$xe.Start()
-					Get-DbaXESession -SqlInstance $xe.Parent -Session $session
+					try {
+						$xe.Start()
+					} catch {
+						Write-Error -Message "Could not start XEvent Session $session on $instance"
```
wsmelton: This should use `Stop-Function` this is part of the message system in the module. Supports not showing the sea of read all the time.

So in this case you would have something like `Stop-Function -Message "Could not start XEvent Session on $instance" -Target $session -ErrorRecord $_`

functions/Stop-DbaXESession.ps1
```diff
@@ -86,11 +86,15 @@ function Stop-DbaXESession {
 				$session = $xe.Name
 				if ($xe.isRunning) {
 					Write-Message -Level Verbose -Message "Stopping XEvent Session $session on $instance."
-					$xe.Stop()
-					Get-DbaXESession -SqlInstance $xe.Parent -Session $session
+					try {
+						$xe.Stop()
+					} catch {
+						Write-Error -Message "Could not stop XEvent Session $session on $instance"
```
wsmelton: Same as previous, need to use `Stop-Function`.

functions/Start-DbaXESession.ps1
dgmyrs: OK.  I added a commit with the changes.

functions/Stop-DbaXESession.ps1
```diff
@@ -89,7 +89,7 @@ function Stop-DbaXESession {
 					try {
 						$xe.Stop()
 					} catch {
-						Write-Error -Message "Could not stop XEvent Session $session on $instance"
+						Stop-Function -Message "Could not stop XEvent Session on $instance" -Target $session -ErrorRecord $_
```
wsmelton: One more minor adjustment I didn't catch until I looked at the whole command. We need to add `-Continue` switch to `Stop-Function` call so it won't stop the foreach loop and keep going to the next one in the collection. If you can do that on each one I'll get this merged.

dgmyrs: ah, yup, makes sense.  Done in the latest commit

PR #2710, opened 2017-11-21
Get-XpDirTreeRestoreFile, use SMO for fixed queries

niphlod:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Purpose
For fixed queries, now that the annoying SMO connection bug is fixed, better to use .Query(). This fixed a bit of issues where the "smo" server object gets passed to invoke-sqlcmd2 (which autoconverts to a string, which drops the dns suffix, ending in a big pool of red blood). I was playing with Get-DbaBackupInformation ^_^ . Incidentally, backup-*/restore-* related cmdlets which make a lot of function cooperate now seems to be a bit speedier (probably because a proper reuse of the connection pool).
tl;dr : whenever the query is fixed (i.e. you don't need parameters) stick with SMO, else, go with invoke-sqlcmd2

### DIFF
--- a/internal/Get-XpDirTreeRestoreFile.ps1
+++ b/internal/Get-XpDirTreeRestoreFile.ps1
@@ -39,7 +39,7 @@ function Get-XpDirTreeRestoreFile {
 	Write-Message -Level InternalComment -Message "Starting"
 	
 	Write-Message -Level Verbose -Message "Connecting to $SqlInstance"
-	$Server = Connect-SqlInstance -SqlInstance $SqlInstance -SqlCredential $SqlCredential
+	$server = Connect-SqlInstance -SqlInstance $SqlInstance -SqlCredential $SqlCredential
 	
 	if (($path -like '*.bak') -or ($path -like '*trn')){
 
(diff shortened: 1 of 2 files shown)


internal/Get-XpDirTreeRestoreFile.ps1
```diff
@@ -63,7 +63,7 @@ function Get-XpDirTreeRestoreFile {
 	ForEach ($d in $dirs) {
 		$fullpath = "$path$($d.Subdirectory)"
 		Write-Message -Level Verbose -Message "Enumerating subdirectory '$fullpath'"
-		$Results += Get-XpDirTreeRestoreFile -path $fullpath -SqlInstance $SqlInstance
+		$Results += Get-XpDirTreeRestoreFile -path $fullpath -SqlInstance $Server -SqlCredential $SqlCredential
```
Stuart-Moore: Does this need the credential as we're passing through a server object?

niphlod: technically no but why not ?

niphlod:
submitted.

internal/Get-XpDirTreeRestoreFile.ps1
Stuart-Moore: More me checking I'd done the right thing when I took the credential passes out when I streamlined the connection stuff (your comment reminded me), and wanted to make sure I wasn't going to cause problems later one.

niphlod:
No probls but being explicit is better than explicit, more so in an
internal function that relies on feature x. There's nothing scary in being
verbose

Il 21 nov 2017 4:18 PM, "Stuart Moore" <notifications@github.com> ha
scritto:

internal/Get-XpDirTreeRestoreFile.ps1
wsmelton: It should actually ignore the credential in this case. The Connect-SqlInstance command validates whether an SMO object was passed in and if it was it goes with that...have not specifically debugged it to watch but recall it does not acknowledge the credential.

wsmelton: It would be standard among other commands that we do not re-pass in that credential object if you are using the server object. So would be cleaner to remove it so it is understood that it is not in use.

niphlod:
ok, last commit removes the redundant $sqlcredential. Let's see if this fails, too.

niphlod:
aha! passed. Thanks for the pointer @Stuart-Moore , I've been a bad boy submitting without running tests before.

internal/Get-XpDirTreeRestoreFile.ps1
```diff
@@ -48,13 +48,16 @@ function Get-XpDirTreeRestoreFile {
 		$Path = $Path + "\"
 	}
 	
-	If (!(Test-DbaSqlPath -SqlInstance $server -path $path)) {
+	If (!(Test-DbaSqlPath -SqlInstance $Server -path $path)) {
```
wsmelton: That would be lowercase `if` and `$server`.

internal/Get-XpDirTreeRestoreFile.ps1
```diff
@@ -48,13 +48,16 @@ function Get-XpDirTreeRestoreFile {
...
 		Stop-Function -Message "SqlInstance $SqlInstance cannot access $path" -EnableException $true
 	}
-	
-	$query = "EXEC master.sys.xp_dirtree '$Path',1,1;"
-	$queryResult = Invoke-Sqlcmd2 -ServerInstance $SqlInstance -Credential $SqlCredential -Database tempdb -Query $query
-	#$queryResult = $Server.Query($query)
+	if ($Server.VersionMajor -lt 9) {
+		$query = "EXEC master..xp_dirtree '$Path',1,1;"
```
wsmelton: Not sure if this matters, not familiar with all the code in this command, but we have been using `$sql` for queries in most commands. Unless there reason to split that up to be more specific.

internal/Get-XpDirTreeRestoreFile.ps1
```diff
@@ -48,13 +48,16 @@ function Get-XpDirTreeRestoreFile {
...
+	} else {
+		$query = "EXEC master.sys.xp_dirtree '$Path',1,1;"
+	}
+	#$queryResult = Invoke-Sqlcmd2 -ServerInstance $SqlInstance -Credential $SqlCredential -Database tempdb -Query $query
+	$queryResult = $Server.Query($query)
```
wsmelton: Do we want to add a `Write-Message -Level Debug` to this so we can output that query execution?

niphlod:
@wsmelton : now the capitalization is consistent function-wise as per your specs

PR #2715, opened 2017-11-21
restore speed improvements

niphlod:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Purpose
Make restore faster. I played a bit with Get-DbaBackupInformation and piping to Restore-DbaDatabase. The current method showed that with a bit of latency (150/200ms) to the backend Test-DbaSqlPath (and it accepting one test at a time) introduces toooooo much slowness. 
Let's try to be the "Operations Studio" rather than the full bloated "Management Studio". 
Please note that this addresses speed without introducing ANY shortcuts, so the restore process remains as sound as it is right now. 

### Approach
Test-DbaSqlPath can be made test multiple paths at the same time. Piping blows, foreaching too. This PR introduces a new Test-DbaSqlPath, whose rationale is:
- the original version accepted multiple instances, while returning at the first result returned. As such, it now just accepts one sql instance, which is the 100% usecase scenario around dbatools
- the original version just returns a bool. While it *could* have been rigged to return an array of bools, it was then left to the caller to rebuild what path was correctly existing and what not. This returns a bool if the passed path is one (to avoid breaking things that rely on that spec) BUT returns a pscustomobject with the list of paths checked and the result

At this point, feel free to destroy this version and ask me to implement Test-DbaSqlPath**s** and/or change the name of the properties returned BUT... this serves as an example, and there are several places around that could benefit from this new "algorithm". Changing Test-DbaBackupInformation to use the new and improved method resulted in a drop from 394 to 211 seconds (almost 100% speedup) while restoring a chain of  1256 files.

tl;dr: I hope all cogs around backup/restore start sitting down a bit to be able to do a proper review and introduce speedups where needed. This is truly where dbatools can shine against "competitors"... 
Right now, e.g., Restore-DbaDatabase spends 5 or 6 seconds just enumerating database files via Get-DbaDatabaseFile (to be sure the path where the file *would* be restored doesn't exist yet). 
This approach is as sound as it can be but for a destination instance with 100 dbs it introduces a 10 minutes delay. The approach seen on Find-DbaOrphanedFiles (or, even, just xp_dirtreeing or Get-DbaFileing the paths where Restore-DbaDatabase already knows it'll drop the files) would reduce it to 5 seconds at most. 

On a related note ( yes @FriedrichWeinmann, I'm back bashing ^_^) messaging really does need to address a function that with a single switch (-Verbose) produces a transcript of 4 MB instead of nothing. I know these are among the most complicated functions we have but in each case (no verbose, *the shell sits around for 4 minutes without giving any indication of the progress* **vs** verbose, *the shell spitting out infos about each db's compatibility level going 3 or 4 levels deep in callers*) the objective of messaging, which is "informing the user" and/or "give a proper feedback", here "blows".

### DIFF
--- a/functions/Test-DbaSqlPath.ps1
+++ b/functions/Test-DbaSqlPath.ps1
@@ -56,41 +56,53 @@ function Test-DbaSqlPath {
 
 	#>
 	[CmdletBinding()]
-	[OutputType([bool])]
 	param (
 		[Parameter(Mandatory = $true)]
 		[Alias("ServerInstance", "SqlServer")]
-		[DbaInstance[]]$SqlInstance,
+		[DbaInstance]$SqlInstance,
 		[PSCredential]$SqlCredential,
 		[Parameter(Mandatory = $true)]
-		[string]$Path,
+		[string[]]$Path,
 		[switch][Alias('Silent')]$EnableException
 	)
-
+	begin {
+		try {
+			$server = Connect-SqlInstance -SqlInstance $SqlInstance -SqlCredential $SqlCredential
+		}
+		catch {
+			Stop-Function -Message "Failure" -Category ConnectionError -ErrorRecord $_ -Target $SqlInstance -Continue
+		}
+	}
 	process {
-		foreach ($instance in $SqlInstance) {
-			try {
-				$server = Connect-SqlInstance -SqlInstance $instance -SqlCredential $SqlCredential
+		if (Test-FunctionInterrupt) { return }
+		
+		$counter = [pscustomobject] @{ Value = 0 }
+		$groupSize = 100
+		$groups = $Path | Group-Object -Property { [math]::Floor($counter.Value++ / $groupSize) }
+		foreach($g in $groups) {
+			$paths_batch = $g.Group
+			$q = @()
+			foreach($p in $paths_batch) {
+				$q += "EXEC master.dbo.xp_fileexist '$p'"
 			}
-			catch {
-				Stop-Function -Message "Failure" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
-			}
-
-			Write-Message -Level VeryVerbose -Message "Checking access to $path for $instance."
-			$sql = "EXEC master.dbo.xp_fileexist '$path'"
-			try {
-				Write-Message -Level Debug -Message "Executing: $sql."
-				$fileExist = $server.Query($sql)
-			}
-
-			catch {
-				Stop-Function -Message "Failed to test the path $Path." -ErrorRecord $_ -Target $instance -Continue
-			}
-			if ($fileExist[0] -eq 1 -or $fileExist[1] -eq 1) {
-				return $true
-			}
-			else {
-				return $false
+			$sql = $q -join ';'
+			$batchresult = $server.ConnectionContext.ExecuteWithResults($sql)
+			if ($Path.Count -eq 1) {
+				if ($batchresult.Tables.rows[0] -eq $true -or $batchresult.Tables.rows[1] -eq $true) {
+					return $true
+				} else {
+					return $false
+				}
+			} else {
+				$i = 0
+				foreach($r in $batchresult.tables.rows) {
+					$doespass = $r[0] -eq $true -or $r[1] -eq $true
+					[pscustomobject]@{
+						filepath = $paths_batch[$i]
+						fileexists = $doespass
+					}
+					$i += 1
+				}
 			}
 		}
 	}
--- a/functions/Test-DbaBackupInformation.ps1
+++ b/functions/Test-DbaBackupInformation.ps1
@@ -142,13 +142,14 @@ Function Test-DbaBackupInformation {
             }
 
             #Test all backups readable
-            Foreach ($path in ($DbHistory | Select-Object -ExpandProperty FullName)){
-                if(!(Test-DbaSqlPath -SqlInstance $RestoreInstance -Path $path)){
+            $allpaths = $DbHistory | Select-Object -ExpandProperty FullName
+            $allpaths_validity = Test-DbaSqlPath -SqlInstance $RestoreInstance -Path $allpaths
+            Foreach($path in $allpaths_validity) {
+                if ($path.fileexists -eq $false) {
                     Write-Message -Message "Backup File $path cannot be read" -Level Warning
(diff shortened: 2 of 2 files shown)


Stuart-Moore:
The failing test is another unit test with a mock. Having a look at it now, and not completely convinced it was actually doing what I though it was doing anyway!

Stuart-Moore:
This mock should work better:
```
            $TestPath = [PSCustomObject]@{Path = 'c:\null.bak'; fileexists = $false}
            Mock Test-DbaSqlPath { return $TestPath } 
```

replaces L30

niphlod:
I'm waiting for the "name and style" guys to come back for checking
test-sqlpath changed as it is right now before changing anything else: this
was just an example implementation (although I'm really happy you approve)

Il 22 nov 2017 9:13 AM, "Stuart Moore" <notifications@github.com> ha
scritto:

wsmelton:
Would we want to make a `Test-SqlPath` and put it back in as internal function. For our use in the module we would not really need all the messaging parts, and would then speed it up further. 

We could keep all the output messaging on the public version because a lot of folks are likely using that after they discovered it.

niphlod:
I'd not duplicate just to "fix" messaging. Besides test-dbasqlpath is not
the most verbose, and it'll be less than the usual with this patch

Il 23 nov 2017 12:37 AM, "Shawn Melton" <notifications@github.com> ha
scritto:

functions/Test-DbaSqlPath.ps1
```diff
@@ -56,41 +56,53 @@ function Test-DbaSqlPath {
...
+		foreach($g in $groups) {
+			$paths_batch = $g.Group
+			$q = @()
+			foreach($p in $paths_batch) {
+				$q += "EXEC master.dbo.xp_fileexist '$p'"
 			}
-			else {
-				return $false
+			$sql = $q -join ';'
+			$batchresult = $server.ConnectionContext.ExecuteWithResults($sql)
```
wsmelton: This should be `$batchResult = $server.Query($sql)`

functions/Test-DbaSqlPath.ps1
```diff
@@ -56,41 +56,53 @@ function Test-DbaSqlPath {
...
-			if ($fileExist[0] -eq 1 -or $fileExist[1] -eq 1) {
-				return $true
+		if (Test-FunctionInterrupt) { return }
+		
+		$counter = [pscustomobject] @{ Value = 0 }
+		$groupSize = 100
+		$groups = $Path | Group-Object -Property { [math]::Floor($counter.Value++ / $groupSize) }
+		foreach($g in $groups) {
+			$paths_batch = $g.Group
+			$q = @()
```
wsmelton: I'd kind of rather see this named `$query = @()` instead of just `$q` to make it more readable.

functions/Test-DbaSqlPath.ps1
```diff
@@ -56,41 +56,53 @@ function Test-DbaSqlPath {
...
+				} else {
+					return $false
+				}
+			} else {
+				$i = 0
+				foreach($r in $batchresult.tables.rows) {
+					$doespass = $r[0] -eq $true -or $r[1] -eq $true
+					[pscustomobject]@{
+						filepath = $paths_batch[$i]
+						fileexists = $doespass
```
wsmelton: This should have property names cased as `FilePath` and `FileExists`

Also why the underscore in the variable name (`$paths_batch`)?

niphlod:
I'll comply soon enough but there is not a single way to code and
styleguide is not that clear. Can I assume it's good to have
test-dbasqlpath change its current behaviour and return a bool when a
single path is passed and a pscustomobject when $path is an array?

Il 24 nov 2017 8:35 PM, "Shawn Melton" <notifications@github.com> ha
scritto:

wsmelton:
There are 44 references showing for the command, 2 of those are the test for the command itself.

I checked a few of the commands and tested. I see no issue making the changes noted.

potatoqualitee:
I'm really excited to hear that we've got a boost in speed! Thanks so much for all of your work @niphlod :it:

potatoqualitee:
ah, i see "soon enough", k gonna merge

potatoqualitee:
Please submit a new PR with required changes. can you foreach instead of changing the behavior of test-dbasqlpath, @niphlod ? it's kinda scary since it's a base command

niphlod:
the whole deal is NOT foreaching. Merging as it is changes test-dbasqlpath behaviour.

potatoqualitee:
aw damn, i didnt realize there was a change to test-dbasqlpath D: does it break other ones? i didnt see failures.

niphlod:
right now test-dbasqlpath is coded wrong. it accepts multiple sqlinstances but stops at the first result. that needs fixing, indipendently of this PR.
Next: test-dbasqlpath is rigged to return a single bool. This forces higher-level functions to foreach, which exhibit wait times that *can* be sidestepped changing how test-dbasqlpath works.
This PR was a POC to show how test-dbasqlpath needs to be changed (with a small change on the caller function) to improve performances, a lot.
I *specifically* asked for directions on how to treat the current feature-set of test-dbasqlpath. My change affected the behaviour only when *multiple* paths are passed in, while retaining the original behaviour of returning a bool when a single path is returned.
This ends up in not breaking any preexisting functionality, but introduces maybe for the first time a function that does return different objects when parameters are passed in differently.
If we wanna leave test-dbasqlpath as it *was*, we need to copy the functionality with my fixes in another "test-dbasqlmultiplepaths", so caller functions can use the improved version and benefit from performance improvements.

potatoqualitee:
ah ha! i like that change. this sounds great.

niphlod:
so what we do ? 
to avoid code repetition, IMHO we should:

- create internals/test-dbasqlmultiplepaths with the main and improved logic
- modify functions/test-dbasqlpath that calls internals/test-dbasqlmultiplepaths for a corner-case (single path passed)
- modify higher level functions, e.g. Test-DbaBackupInformation, to use internals/test-dbasqlmultiplepaths
?

potatoqualitee:
Sorry to confuse. I like your change as is. Just reutrn true/false for one server and multiple for multiple. Since everything seems to pass, I'm happy to close as resolved.

PR #2733, opened 2017-11-27
restore speed improvements, round 2

niphlod:
added modifications previously required on https://github.com/sqlcollaborative/dbatools/pull/2715 .

### DIFF
--- a/functions/Test-DbaBackupInformation.ps1
+++ b/functions/Test-DbaBackupInformation.ps1
@@ -1,170 +1,171 @@
 Function Test-DbaBackupInformation {
-<#
-    .SYNOPSIS
-        Tests a dbatools backuphistory object is correct for restoring
+	<#
+	.SYNOPSIS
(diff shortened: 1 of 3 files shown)


niphlod:
till this is merged, Test-DbaBackupInformation tests will always fail... and right now development is failing.

functions/Test-DbaBackupInformation.ps1
```diff
@@ -1,170 +1,171 @@
...
+		[DbaInstanceParameter]$SqlInstance,
+		[PSCredential]$SqlCredential,
+		[switch]$Withreplace,
+		[switch]$Continue,
+		[switch]$EnableException
+	)
+
+	begin {
+		try {
+			$RestoreInstance = Connect-SqlInstance -SqlInstance $SqlInstance -SqlCredential $SqlCredential
```
wsmelton: We can fix this in another PR but this should be `$server`, that is the pattern we use in every command and not sure why it was allowed in this one.

PR #2739, opened 2017-11-28
Adjusting Detailed parameter

wsmelton:
Addressing #2217 and reverting some things that inadvertently got changed.

Going with updates added to 2217 for keeping the Detailed parameter in the param block and help until 1.0.0 release. Adding the deprecated command as just notification. The parameter will no longer do anything but just adding it so it does not break scripts.


potatoqualitee:
thanks, bud! will merge once it passes.

potatoqualitee:
merci beaucoup 🍰

PR #2740, opened 2017-11-28
Adjust Install-DbaMaintenanceSolution add disconnect

wsmelton:
Reference need to have a non-pooled connection via #2738 

Added a explicit `Disconnect()` to prevent left over connections.

### DIFF
--- a/functions/Install-DbaMaintenanceSolution.ps1
+++ b/functions/Install-DbaMaintenanceSolution.ps1
@@ -1,33 +1,33 @@
 function Install-DbaMaintenanceSolution {
-    <#
+	<#
     .SYNOPSIS
         Download and Install SQL Server Maintenance Solution created by Ola Hallengren (https://ola.hallengren.com)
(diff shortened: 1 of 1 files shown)


functions/Install-DbaMaintenanceSolution.ps1
```diff
@@ -277,47 +277,55 @@ function Install-DbaMaintenanceSolution {
...
+		}
+
+		if ((Test-Path $path)) {
+			Remove-Item -Path $temp\sql-server-maintenance-solution-master -Recurse -Force -ErrorAction SilentlyContinue
+		}
+		Write-Message -Level Output -Message "Installation complete"
+	}
+	end {
+		# Only here due to need for non-pooled connection in this command
+		try {
```
potatoqualitee: This will only close the last server - so it should be moved up to be within the `foreach ($instance in $sqlinstance)`

potatoqualitee:
Thanks baw!

PR #2746, opened 2017-11-29
Adjust test for Find-DbaOrphanedFile

wsmelton:
- Standardized test to be against one instance.
- Added unit test for parameters


potatoqualitee:
i poked around this one for a while and found the combination in there worked the best, though not always 100% 👎  but it did have the best odds

wsmelton:
I'm not sure I see what the need is to just validate a database is not on instance1, but all the work of creating and detaching are being done for testing the command against instance2?

wsmelton:
Right now I'm having the issue Appveyor is having on this test. It fails on loading the library, so the Query() method is not available.

wsmelton:
This is all I'm getting on this test now:

image

If I go run another command that uses that Query() method get same result:
image

niphlod:
uhm, lemme check if that happens for me too.

wsmelton:
This is because of VS Code and the integrated terminal @niphlod

niphlod:
to be fair, I can't use development in a fresh PS console 2 out of 3 times.... and why would appveyor fail ?

niphlod:
hum. dotsource is maybe the answer ....

niphlod:
nope. it's the fact we import the psm1 in appveyor (which enables us testing internals, I *think*) and the dll module not being loaded correctly.

potatoqualitee:
looks like you guys got it to work 👍 thank you!

PR #2759, opened 2017-12-01
Copy-DbaTable, first draft

niphlod:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Purpose
We don't have a way to migrate a single table from one database to another. This "keeps it simple" without incurring in lots of RAM usage and *could* be used as a good substitute for snapshot replication scenario (without fiddling with agents, monitoring, articles, etc)

### Approach
Streaming reads piped to BCP. Unfortunately there's no easy way to pipe into Write-DbaDatatable an ExecuteReader but if you happen to be able to do it, Get-DbaTableData | Write-DbaDatatable could be beneficial. As always, major cringe points for this iteration of Copy-DbaTable for the sheer number of parameters, I tried to be balanced about it (and that's why the CBH isn't complete). 

As always, I'm open to suggestions. Code review for later ^_^

A biig shootout goes to @potatoqualitee for enduring one or two evenings ironing out things.

### DIFF
--- a/functions/Copy-DbaData.ps1
+++ b/functions/Copy-DbaData.ps1
@@ -1,4 +1,4 @@
-function Copy-DbaTable {
+function Copy-DbaData {
 	<#
 		.SYNOPSIS
 			Copies data between SQL Server tables.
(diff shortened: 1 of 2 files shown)


wsmelton:
Maybe for version 3 of this command, but had this issue today for a client. I have a database that is tied to the monitoring tool we use (Pythian) for clients. I needed to sync the table on one server to two other servers.

The method used in this copy command would not have worked because it was dealing with SQL Agent jobs, so I had a column for the `job_id` that I couldn't overwrite. I ended up using a merge statement that would updated those values that matched the job name.

Not sure off the top how we could do this in a clean manner, would require some lookup of columns and mapping and blah blah blah....

niphlod:
Lemme split this according to what has been discussed in #dbatools-dev,
then we'll see. I think with -Query you can cheat pretty much everything...

Il 04 dic 2017 6:56 PM, "Shawn Melton" <notifications@github.com> ha
scritto:

niphlod:
okay, this now copies only data, as discussed in #dbatools-dev

potatoqualitee:
Looks fabulous, just gonna change the name to Copy-DbaTableData, then we can also have Copy-DbaTableSchema ?

niphlod:
I was merely following what has been discussed on dbatools-dev about the
naming

Il 06 dic 2017 9:56 AM, "Chrissy LeMaire" <notifications@github.com> ha
scritto:

PR #2765, opened 2017-12-03
DacPac Patches

potatoqualitee:
@RichieBzzzt check out the changes that I made to Export and let me know what you think

image

I made the most commonly used params PowerShell params, then allowed for additional params like so

```powershell
.EXAMPLE
Export-DbaDacpac -SqlInstance sql2016 -Database SharePoint_Config
Exports the dacpac for SharePoint_Config on sql2016 to $home\Documents\SharePoint_Config.dacpac

.EXAMPLE
$moreprops = "/p:VerifyExtraction=$true /p:CommandTimeOut=10"
Export-DbaDacpac -SqlInstance sql2016 -Database SharePoint_Config -Directory C:\temp -ExtendedProperties $moreprops

Sets the CommandTimeout to 10 then extracts the dacpac for SharePoint_Config on sql2016 to C:\temp\SharePoint_Config.dacpac then verifies extraction.
```

Do you see any room for improvement or additional "default" Parameters?


functions/Export-DbaDacpac.ps1
```diff
@@ -0,0 +1,179 @@
...
+				$dbs = $dbs | Where-Object Name -notin $ExcludeDatabase
+			}
+			
+			foreach ($db in $dbs) {
+				$dbname = $db.name
+				$connstring = $server.ConnectionContext.ConnectionString.Replace('"', "'")
+				if ($connstring -notmatch 'Database=') {
+					$connstring = "$connstring;Database=$dbname"
+				}
+				$path = "$Directory\$dbname.dacpac"
```
PowerDBAKlaas: Since we can suply multiple instances, there's a chance we'll have DB's with the same name.
Won't it be safer to include the '\computername\instancename' in the path?

functions/Export-DbaDacpac.ps1
```diff
@@ -0,0 +1,179 @@
...
+		[object[]]$ExcludeDatabase,
+		[switch]$AllUserDatabases,
+		[string]$Directory = "$home\Documents",
+		[string]$ExtendedParameters,
+		[string]$ExtendedProperties,
+		[switch]$EnableException
+	)
+	
+	process {
+		if (!$Database -and !$ExcludeDatabase -and !$AllUserDatabases) {
```
PowerDBAKlaas: Didn't we replace such checks with `Test-Bound` or `$PSBoundParameters`?

functions/Export-DbaDacpac.ps1
```diff
@@ -0,0 +1,179 @@
...
+			
+			try {
+				Write-Message -Level Verbose -Message "Connecting to $instance."
+				$server = Connect-SqlInstance -SqlInstance $instance -SqlCredential $sqlcredential -MinimumVersion 9
+			}
+			catch {
+				Stop-Function -Message "Failure" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
+			}
+			
+			$dbs = $server.Databases | Where-Object { $_.IsSystemObject -eq $false }
```
PowerDBAKlaas: Aren't there additional requirements, like status online or normal?

functions/Export-DbaDacpac.ps1
```diff
@@ -0,0 +1,179 @@
...
+					$startprocess.UseShellExecute = $false
+					$process = New-Object System.Diagnostics.Process
+					$process.StartInfo = $startprocess
+					$process.Start() | Out-Null
+					$process.WaitForExit()
+					$stdout = $process.StandardOutput.ReadToEnd()
+					$stderr = $process.StandardError.ReadToEnd()
+					Write-Message -level Verbose -Message "StandardOutput: $stdout"
+					
+					[pscustomobject]@{
```
PowerDBAKlaas: It would be cleaner to just set the `$status` in the `try{} catch{}` and define the `PSCustomObject` and DefaultView only once after that block.

functions/Export-DbaDacpac.ps1
```diff
@@ -0,0 +1,179 @@
...
+					} | Select-DefaultView -ExcludeProperty ComputerName, InstanceName
+				}
+				catch {
+					[pscustomobject]@{
+						ComputerName	 = $server.NetName
+						InstanceName	 = $server.ServiceName
+						SqlInstance	     = $server.DomainInstanceName
+						Database		 = $dbname
+						FileName	   = $path
+						Elapsed	       = [prettytimespan]($elapsed)
```
PowerDBAKlaas: I don't see `$elapsed` defined.
Could the `$resultstime = [diagnostics.stopwatch]::StartNew()` be moved outside the `try{} catch{}` so we have a value for 'elapsed' whether it fails or succeeds?

functions/Export-DbaDacpac.ps1
potatoqualitee: Yes, I forgot, it is Test-Bound, will update thanks

functions/Export-DbaDacpac.ps1
potatoqualitee: It's IsAccessible! Will add

functions/Export-DbaDacpac.ps1
potatoqualitee: good call! ill do instance with \ replaced with $

functions/Export-DbaDacpac.ps1
potatoqualitee: I was surprised (and disappointed) to find out that it does keep the variable, no matter what. at least that was the behavior last time I tested it. but I'll move it above try.

functions/Export-DbaDacpac.ps1
potatoqualitee: thank you, i fixed the $elapsed but the Stop-Function makes it so that there wouldnt be an output of the object if it failed. Which is alright if that's the way we wanna go.

RichieBzzzt:
Hi,

Just fetched and had a look. All good.

Export seems fine. I did consider adding commonly params myself (not surewhy I didn't TBH), but this makes sense.

Thanks for bringing all functions in-line with DBATools standards!

functions/Publish-DbaDacpac.ps1
```diff
@@ -0,0 +1,255 @@
...
+		
+		if (-not (Test-Path -Path $Path)) {
+			Stop-Function -Message "$Path not found!"
+		}
+		
+		if (-not (Test-Path -Path $PublishXml)) {
+			Stop-Function -Message "$PublishXml not found!"
+		}
+		
+		if ($PipedPublishedProfile.ConnectionString) {
```
RichieBzzzt: Should check be removed now that the $pipedPublishPreview variable has is not part of function?

potatoqualitee: oh good point, thank you!

PR #2766, opened 2017-12-03
Library import offload

FriedrichWeinmann:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Purpose

 - Resolve library-related import issues
 - Resolve library-related update issues

### Changes

 - New: By default, all smo libraries will be copied to temp and imported from temp. This avoids locking module files and enables update from website on older systems.
 - Fix: dbatools library is no longer imported as part of the manifest. This enables dbatools to unblock itself, resolving "library or dependencies not found" exceptions on import
 - Updated: dbatools library is also copied to that temporary folder before importing, same as smo (used to be copied to temp straight)
 - Updated: Maintenance task that cleans up in temp now cleans up the folders with all libraries, instead of just dbatools library

### Notes

 - The import behavior for the smo libraries also supports the dbatools configuration for strict security mode, enabling dbas to disable the copy

### DIFF
--- a/bin/library.ps1
+++ b/bin/library.ps1
@@ -68,17 +68,8 @@ if ($ImportLibrary) {
 				$start = Get-Date
 				try {
 					Write-Verbose -Message "Found library, trying to copy & import"
-					$libraryTempPath = "$($env:TEMP)\dbatools-$(Get-Random -Minimum 1000000 -Maximum 9999999).dll"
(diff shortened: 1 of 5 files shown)


potatoqualitee:
thanks for the update and approval!

Issue #2764, opened 2017-12-02
enhancement to Restore-DbaFromDatabaseSnapshot


lostsole:
## Bug Report
when the Restore-DbaFromDatabaseSnapshot will fail to refresh if there are open users process having on the database that will be refreshed from the Database snapshot.
### General Troubleshooting steps

Run the 
Restore-DbaFromDatabaseSnapshot -SqlServer sqlserver2012 -Snapshots Perf_test_snap
If I have an open connection to the database: Perf_test, the snapshot will not restore because of the open process.

### Version Information
 - Operating System (Name|Version): Windows 10 and windows 7
 - PowerShell Version: 5.1
 - SQL Server (Edition|Version): SQL 2012 standard

## Steps to Reproduce
Restore-DbaFromDatabaseSnapshot -SqlServer sqlserver2012 -Snapshots Perf_test_snap

## Feature Request
Please add the  Stop-DbaProcess  to the Restore-DbaFromDatabaseSnapshot.  Just like the restore-DBADatabase has to kill the open process before restoring.
 
This way when I automate a job for the end user to Refresh back to snapshot the any open process on the database being restored will be killed.

### Problem to solve
Kill any open user process locking the database from being refresh

niphlod:
That's what -Force is for. Or do you have issues even with -Force?

Il 02 dic 2017 5:14 AM, "lostsole" <notifications@github.com> ha scritto:

wsmelton:
@lostsole we already have equivalent to what `Stop-DbaProcess` does in this command at [line202](https://github.com/sqlcollaborative/dbatools/blob/development/functions/Restore-DbaFromDatabaseSnapshot.ps1#L202). We have tied the `-Force` parameter to killing active connections to a snapshot.

So in your example, you would call it as such:
```powershell
Restore-DbaFromDatabaseSnapshot -SqlServer sqlserver2012 -Snapshots Perf_test_snap -Force
```

We do not wish to add a default behavior that kills all connections, which is why the action is tied to that parameter. If you are having direct problems with the use of the Force parameter please reopen this issue and let us know the details.

I've got a PR submitted that adds an example for clarity on use of the Force parameter as noted above.

lostsole:
Thanks for the FYI, I missed the -Force when I tested.  But works like a charm.  All is good.

Shawn Melton closed the item. 

Sent from Mail for Windows 10

lostsole:
[...]

PR #2767, opened 2017-12-04
Add example and format

wsmelton:
Add additional example to show the use of `-Force` to kill active sessions. Reference to #2764
Format help and adjust for camelCasing.


potatoqualitee:
hey shawn, can you check out whats up with these particular tests?

Issue #2754, opened 2017-11-30
[Repair-DbaOrphanUser/Remove-DbaOrphanUser] Processes correctly but ends with an error.


MakwaWes:
### General Troubleshooting steps

- [ ] Verified running latest release of dbatools?
0.9.108
- [ ] Can duplicate in new/fresh PowerShell.exe session?
Yes

### Version Information
 - Operating System (Name|Version): Windows Server 2012
 - PowerShell Version: 5.1
 - SQL Server (Edition|Version): Microsoft SQL Server Standard (64-bit) 11.0.5678.0

## Steps to Reproduce

SharePoint Content Databases migrated and having a bunch of orphan users.
While running the following cmdlet, the orphan users are removed but the cmdlet ends up with an error.
`Remove-DbaOrphanUser -SqlInstance $sqlInstance -Database $myCDB -EnableException -Force`

Outputs correctly the orphan users deleted but ends with this error message from [Write-Message] function :
```
Instance          : mySqlServer\myNamedInstance
Database          : myContentDBName-01
SchemaName        : myADDomain\svc_acct_orphan01
Action            : DROP
SchemaOwnerBefore : myADDomain\svc_acct_orphan01
SchemaOwnerAfter  : N/A

Instance          : mySqlServer\myNamedInstance
Database          : myContentDBName-02
SchemaName        : myADDomain\svc_acct_orphan02
Action            : DROP
SchemaOwnerBefore : myADDomain\svc_acct_orphan02
SchemaOwnerAfter  : N/A

Write-Message : Cannot process argument transformation on parameter 'Level'. Cannot convert value "Error" to type "Sqlcollaborative.Dbatools.dbaSystem.MessageLevel".
Error: "Unable to match the identifier name Error to a valid enumerator name. Specify one of the following enumerator names and try again:
Critical, Important, Output, Significant, VeryVerbose, Verbose, SomewhatVerbose, System, Debug, InternalComment, Warning"
At line:333 char:28
+                         Write-Message -Level Error -Message $_
+                                              ~~~~~
    + CategoryInfo          : InvalidData: (:) [Write-Message], ParameterBindingArgumentTransformationException
    + FullyQualifiedErrorId : ParameterArgumentTransformationError,Write-Message
```

### Problem to solve
Could it be possible to fix this cosmetic error message please ?

### Additional information
Thanks for your hard work, one of the best PowerShell modules. Using it daily and for automation scripts !
🥇

[...]

PR #2768, opened 2017-12-04
Fix Write-Message -Level Error

ClaudioESSilva:
This no longer exists need to replace by the Stop-Function

This also fix the issue #2754

### DIFF
--- a/functions/Remove-DbaDatabaseSafely.ps1
+++ b/functions/Remove-DbaDatabaseSafely.ps1
@@ -10,7 +10,7 @@ function Remove-DbaDatabaseSafely {
 
 		.PARAMETER SqlInstance
 			The SQL Server instance holding the databases to be removed. You must have sysadmin access and server version must be SQL Server version 2000 or higher.
-			
+
 		.PARAMETER SqlCredential
 			Allows you to login to servers using SQL Logins instead of Windows Authentication (AKA Integrated or Trusted). To use:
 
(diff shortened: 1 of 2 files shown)


functions/Remove-DbaDatabaseSafely.ps1
```diff
@@ -384,7 +384,7 @@ function Remove-DbaDatabaseSafely {
 				}
 			}
 			catch {
-				Write-Message -Level Error -Message "Restore failed: $($_.Exception)."
+				Stop-Function -Message "Restore failed: $($_.Exception)."
```
potatoqualitee: Thanks, Claudio! The ErrorRecord actually outputs the text you're looking for. Try this instead `Stop-Function -Message "Restore failed" -ErrorRecord $_ -Target $whateveristhetarget` 

but replace target with whatever you need to. unsure why? Take a look at  `Get-DbatoolsLog | OGV`

functions/Remove-DbaOrphanUser.ps1
```diff
@@ -330,7 +330,7 @@ function Remove-DbaOrphanUser {
 						$User = $null
 					}
 					catch {
-						Write-Message -Level Error -Message $_
+						Stop-Function -Message $_ -Target $db -Continue
```
potatoqualitee: Appending `-ErrorRecord $_` would be good here and removing $_ as the message. You could just say "Failure" if you like

functions/Remove-DbaDatabaseSafely.ps1
ClaudioESSilva: Thanks @potatoqualitee ! done.

functions/Remove-DbaOrphanUser.ps1
ClaudioESSilva: Thanks @potatoqualitee ! done.

wsmelton: You can actually get that inner exception expanded to a point with Stop-Function by using the `-Exception` parameter. Fred added that months back so you can do:

```
Stop-Function -Message "Whatever failed" -Target <your target variable> -ErrorRecord $_ -Exception $_.Exception -Continue
```

I don't know if you could just pass `$_.Exception.InnerException` as well but not sure how that effects it.

potatoqualitee:
merci! 🌻

PR #2770, opened 2017-12-04
Restore Suspect Pages

Stuart-Moore:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [x] Documentation
 - [ ] Build system
 

### Purpose
As we had a command Get-DbaSuspectPages thought it'd be nice to have something to fix them as well

### Approach
Gets suspect pages with Get-DbaSuspect pages, then restores those from the last full backup and rolls forward all modification using the log backups. Takes a tail log backup to bring the page(s) right up to date.

If the target system does it online, if not then it's offline.

### Commands to test
```
Get-DbaBackupHistory -SqlInstance localhost\sqlexpress2016 -Database pagerestore -last  | Restore-DbaDatabase -SqlInstance localhost\sqlexpress2016 -PageRestore (Get-DbaSuspectPage -SqlInstance localhost\sqlexpress2016 -Database PageRestore) -TrustDbBackupHistory -AllowContinue -DatabaseName PageRestore -PageRestoreTailFolder c:\temp

### DIFF
--- a/functions/Backup-DbaDatabase.ps1
+++ b/functions/Backup-DbaDatabase.ps1
@@ -281,12 +281,13 @@ function Backup-DbaDatabase {
 				$backup.Incremental = $true
 				$outputType = 'Differential'
 			}
-			
+			$Backup.NoRecovery = $False
 			if ($Type -eq "Log") {
 				Write-Message -Level VeryVerbose -Message "Creating log backup"
 				$Suffix = "trn"
 				$OutputType = 'Log'
 				$SMOBackupType = 'Log'
+				$Backup.NoRecovery = $NoRecovery
 			}
 			
 			if ($type -in 'Full', 'Database') {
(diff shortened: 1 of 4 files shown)


potatoqualitee:
holy crap!! HOW COOL

gbargsley:
That is cool.  Never thought my little function would help so much...

Stuart-Moore:
Good news is that none of the Pester Fails appear to be mine :+1:

Seemed an obvious hole to fill once you'd put that in @gbargsley 

Think we should have a seperate Clear-SuspectPage cmdlet to remove the pages fixed. Then I can call if from this to clear RESTORED (type 4) ones, and leave users to clear the other ones as they get fixed.

gbargsley:
I am going to have some dedicated work time tomorrow.  Maybe I can do the clear- command.  So do you think we should have a param to clear specific item db, fileid, pageid and eventtype?

potatoqualitee:
How are they cleared by SQL Server? can we clear them using SQL Server natively, perhaps using some kind of dbcc page command? or verify using dbcc page first?

gbargsley:
When I was doing the research, DBCC makes the entry, but it does not go back and do any pruning to the table.  So if we fix a page and want to clear it we would have to manage.

potatoqualitee:
@Stuart-Moore - can I  merge?

Stuart-Moore:
Yes. A blanket clean up would be a bad thing. The more flexible it is the
more use we can make of it internally

PR #2778, opened 2017-12-04
Whoisactive - added extra try/catch

potatoqualitee:


potatoqualitee:
This is a weird branch

PR #2807, opened 2017-12-07
Touchup and address messaging issues - Read-DbaBackupHeader

wsmelton:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Addressing use of `Write-Exception` and updating the messaging system in this command. Use of `Write-Message` in some areas replaced with proper call to `Stop-Function`. I hopefully have this where the actual error that you hit on T-SQL executions is showing properly now. I added `$_.Exception.InnerException.InnerException` that should now show the actual error you hit.

I also removed the `return` calls as this command accepts multiple paths and will stop running on the first file that hits an error, or hits the catch blocks. It will now continue on trying the other paths passed into the command.

image

### Commands to test

This command is referenced by three commands in the module, I ran each one locally just to make sure and each test passed with flying green colors:

image

image

image

So if this fails in Appveyor...screw you Appveyor!

### DIFF
--- a/functions/Read-DbaBackupHeader.ps1
+++ b/functions/Read-DbaBackupHeader.ps1
@@ -84,7 +84,7 @@ function Read-DbaBackupHeader {
 
             Gets the backup header information from the SQL Server backup file stored at https://dbatoolsaz.blob.core.windows.net/azbackups/restoretime/restoretime_201705131850.bak on Azure
     #>
-	[Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidUsingPlainTextForPassword")]
+	[Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidUsingPlainTextForPassword",'')]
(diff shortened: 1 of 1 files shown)


wsmelton:
I'm not seeing any indentation issues, do you have an example?

If there are some then it is settings that VS Code is not picking up.

The `-Simple` in this section likely needs to be removed as well...

```			if ($Simple) {
				$dataTable | Select-Object DatabaseName, BackupFinishDate, RecoveryModel, BackupSizeMB, CompressedBackupSizeMB, DatabaseCreationDate, UserName, ServerName, SqlVersion, BackupPath
			}
			elseif ($FileList) {
				$dataTable.filelist
			}
			else {
				$dataTable
			}```

wsmelton:
Failure on test related to DACPAC command. Going to go ahead and merge this.

PR #2808, opened 2017-12-07
Added description column - Get-DbaAgentSchedule fixes #2559 

### LINKED ISSUE #2559: Get-DbaAgentSchedule, include "plain English" description in output
## Feature Request
It would be very handy if `GetDbaAgentSchedule` could include the "plain English" version of the schedule, similar to what is displayed in SSMS. I don't believe it does this at the moment.

### Problem to solve
I need to create a report, for consumption by others who might not necessarily have access to SQL instances, so they can see what the backup schedules are. Currently I can achieve this for myself by using a Central Management Server and running the following code against all msdb databases:

`sp_help_schedule @include_description = 1`

I can interpret what "Every 1 week(s) on Sunday at 1500" means, which is the kind of output that SP gives us, but it's not exactly what I would call plain English ;) For example, "Sunday at 1500" is not 3pm on a Sunday, it's actually 00:15:00am. Very different!

### Additional information

This information appears to be available, ready made, if the following is anything to go by:  https://dba.stackexchange.com/questions/10018/job-schedule-view

They're creating a new object with `Microsoft.SqlServer.Management.SqlMgmt.JobScheduleData`, then getting the `Description` property from that. It looks exactly the same, to me, as shown in SSMS.

If this property were added, I could automate code which inserts the results into a database table, then write an SSRS report against the data, thus my manager will be happy.

Name|Schedules
-------|-----------
ASPState_Job_DeleteExpiredSessions|Occurs every day every 1 minute(s) between 00:00:00 and 23:59:59. Schedule will be used starting on 16/10/2000.
syspolicy_purge_history|Occurs every day at 02:00:00. Schedule will be used starting on 1/01/2008.

Keep up the great work, dudes, you're killin' it!

sanderstad:
The schedule can be kind of cryptic when it comes to the schedule. SSMS
shows a nice description with each schedule and so should this function.

### DIFF
--- a/functions/Get-DbaAgentSchedule.ps1
+++ b/functions/Get-DbaAgentSchedule.ps1
@@ -48,6 +48,181 @@ function Get-DbaAgentSchedule {
 		[switch][Alias('Silent')]$EnableException
 	)
 
+	begin {
+		function Get-ScheduleDescription {
+			[CmdletBinding()]
+			param (
+				[Parameter(Mandatory = $true)]
+				[ValidateNotNullOrEmpty()]
+				[object]$Schedule
+			
+			)
+			
+			# Get the culture to make sure the right date and time format is displayed
+			$datetimeFormat = (Get-culture).DateTimeFormat
+
+			# Set the intial description
+			$description = ""
+
+			# Get the date and time values
+			$startDate = Get-Date $Schedule.ActiveStartDate -format $datetimeFormat.ShortDatePattern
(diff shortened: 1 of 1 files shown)


sanderstad:
Command to get the below result
`Get-DbaAgentSchedule -SqlInstance yourserver | Select-Object Name, Description | Out-GridView`

image

potatoqualitee:
bon merci!

potatoqualitee:
looks very pretty 💅

PR #2820, opened 2017-12-09
Add filters to Find-DbaDatabaseGrowthEvent

wsmelton:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [x] Documentation
 - [ ] Build system
 

### Purpose
 
If you are querying a heavy server that has large number of databases, use of PowerShell filtering requires the command to pull all the data out of the default trace before you can filter it. Added parameters that let you filter the data as it is pulled from the default trace. Improves performance on large systems.

Provides the ability to only pull variations of growth or shrink only, and then only evens for data files or log files.
### Approach

I didn't want to sit there and put in a bunch of `if/else` statements so went with using `switch`. However how doing that where it would cover any sequence of `log,growth` or `data,shrink` required use of a collection on the event classes. So the logic in the command is backwards, in that I start the collection with all the events in it, and as the filter is tested it removes each event that is not desired.

I'm fairly sure I got the logic right but do please test and break if able.
### Commands to test
image

### DIFF
--- a/functions/Find-DbaDatabaseGrowthEvent.ps1
+++ b/functions/Find-DbaDatabaseGrowthEvent.ps1
@@ -94,12 +94,44 @@ function Find-DbaDatabaseGrowthEvent {
 		[object[]]$ExcludeDatabase,
 		[ValidateSet('Growth', 'Shrink')]
 		[string]$EventType,
-		[ValidateSet('Data','Log')]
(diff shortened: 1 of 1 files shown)


gbargsley:
Love it.  Ran it on a couple servers and does as you showed in your suggested tests.

potatoqualitee:
thanks to you both!

PR #2823, opened 2017-12-11
Set-DbaStartParameter now works with offline instances

### LINKED ISSUE #2822: Set-DbaStartupParameter requires Service to be running
### Problem to solve
Many of the common Scenarios for changing Startup Parameters for a SQL Server instance are to correct a Service that is not running. As written the Set-DbaStartupParameter function cannot address any of these issues.

### Additional information
The Set-DbaStartupParameter function has option to set the -m switch to allow Server to be started in Single User Mode. The most common reason for this is to restore master database.
However after setting the database into Single-User mode and restoring the database the master database the Service is immediately stopped. As Set-DbaStartupParameter will not work against a stopped service the command cannot be used to revert the change

Other examples like fixing incorrect memory allocations or moving location of Master or Error logs after a disk failure also require modifying the Startup parameters of a stopped instance

Stuart-Moore:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #2822)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Fixes #2822

@sqllensman can you give this a spin just so we've an independent test. Thanks

### Approach
If it can't connect to the specified instance will the proceed using just WMI.  By default it won't apply path changes as they can't be tested.

if TrustPath set, we'll trust the paths and push them in.

using the Offline switch shortcuts the connection test, so you can have 30s of your life for other uses! Needs TrustPath for path changes as well.

### Commands to test
```
Set-DbaStartupParameter -SqlInstance localhost\sqlexpress2016 -Verbose -MinimalStart -Offline
```

### DIFF
--- a/functions/Set-DbaStartupParameter.ps1
+++ b/functions/Set-DbaStartupParameter.ps1
@@ -13,6 +13,8 @@ For full details of what each parameter does, please refer to this MSDN article 
 .PARAMETER SqlInstance
 The SQL Server instance to be modified
 
+If the Sql Instance is offline path parameters will be ignored as we cannot test the instance's access to the path. If you want to force this to work then please see the TrustPath switch
+
 .PARAMETER SqlCredential
 Windows or Sql Login Credential with permission to log into the SQL instance
 
@@ -79,6 +81,9 @@ Overrides the default behaviour and replaces any existing trace flags. If not tr
 .PARAMETER StartUpConfig
 Pass in a previously saved SQL Instance startUpconfig
 using this parameter will set TraceFlagsOverride to true, so existing Trace Flags will be overridden
+
+.PARAMETER TrustPath
+This skips testing the paths passed in. This is useful if you are working with an offline SQL Server instance
 
 .PARAMETER WhatIf 
 Shows what would happen if the command were to run. No actions are actually performed. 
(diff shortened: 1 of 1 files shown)


functions/Set-DbaStartupParameter.ps1
```diff
@@ -80,6 +82,14 @@ Overrides the default behaviour and replaces any existing trace flags. If not tr
 Pass in a previously saved SQL Instance startUpconfig
 using this parameter will set TraceFlagsOverride to true, so existing Trace Flags will be overridden
 
+.PARAMETER Offline
+Setting this switch will try perform the requested actions without conntect to the SQL Server Instance, this will speed things up if you know the Instance is offline.
+
+Note, that with this switch Paths will be ignored unless you also speficy TrustPath
+
+.PARAMETER TrustPath
```
wsmelton: I'd just use a `-Force` parameter as we do with other commands. Same rule applies for TrusthPath as it would Force.

Stuart-Moore: Switched Force for TrustPath in latest push.

potatoqualitee:
thanks to you both!

Issue #2826, opened 2017-12-11
Test-DbaVirtualLogFile - Column "TotalCount" is empty


ClaudioESSilva:
## Bug Report
`Test-DbaVirtualLogFile` returns empty "TotalCount".
I will submit a PR to fix this.

Before the last change (https://github.com/sqlcollaborative/dbatools/commit/2aa93586e79c4a369d4a82f7a9ebe586fc4c594c#diff-5725fea5cee257ddcc233e60e64e5fa2) , "Count" was the column name, and that is the actual name of the output object.

### General Troubleshooting steps

- [x] Verified running latest release of dbatools?
- [x] Verified errors are not related to permissions?
- [x] Can duplicate in new/fresh PowerShell.exe session?

### Version Information
 - Operating System (Name|Version): Windows Server 2012 R2 Standard (6.3.9600 N/A Build 9600)
 - PowerShell Version: 4.0
 - SQL Server (Edition|Version): Microsoft SQL Server 2016 (SP1-CU5) (KB4040714) - 13.0.4451.0

## Steps to Reproduce
Test-DbaVirtualLogFile -SqlInstance <instance>

 - [x] Attach any screenshots (if possible/allowed)
![image](https://user-images.githubusercontent.com/19521315/33842091-76da30ca-de91-11e7-95c9-aa646c650f74.png)

PR #2832, opened 2017-12-12
Change Test-DbaVirtualLogFile Property name to "Total"

ClaudioESSilva:
Instead of Count.

This is the correct fix for issue #2826

Thanks @wsmelton !


potatoqualitee:
hello, lookin good! please also update the test 👍 then we should be set.

potatoqualitee:
Yesssssssssssssssssss 🥇 Thank you, bud 🌻

PR #2836, opened 2017-12-12
Able to use -ScriptAnalyzer without -Coverage

ClaudioESSilva:
We were not able to use `-ScriptAnalyzer` switch without using `-Coverage` too.


potatoqualitee:
added request for @niphlod to review

potatoqualitee:
awesome, thanks! will merge once we get some passes

PR #2841, opened 2017-12-13
Fix Test-dbaConnectionAuthSchema detailed #2217

joshcorr:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [x] Documentation
 - [ ] Build system
 

### Purpose
 
- Update the output of the command to use Select-DefaultView
- Remove any use of the -Detailed parameter
- Update Documentation and Examples

### Approach

- Replaced the if/else block at the end with Select-DefaultView
- Tested the Command both with Select-Object * and -Detailed
- Executed Pester Test and Validated it is not being called with the old syntax in other cmdlets

### Screenshots

image


potatoqualitee:
this is fabulous! thank you 🌮

PR #2842, opened 2017-12-13
Fixed the multi database option

sanderstad:
There was a problem with assigning multiple databases to the function. This is now fixed.
Also fi a database does not exist on the primary instance the function will
now stop.


sanderstad:
Databases before
image

Databases after
image

During the setup
image

Error if database doesn't exist
image

potatoqualitee:
fabulous! thank you for your continued work on these commands.

PR #2848, opened 2017-12-13
fixed handling of query text and query plan field in DbaDiagnosticQuery and Export

andrekamman:
## Type of Change

 - [X] Bug fix (non-breaking change, fixes (no issue nr available)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Purpose
Glenn Berry uses different names for the query text columns and query plan columns in the various sql statements. (sql text, sql_text, sqltext) etc. I'm now catching all options.
This fixes the export to excel. It now no longer tries to put the query text and query plans in Excel but saves them as separate .sql and .sqlplan files. 

I've also changed the parsing of the scripts a bit. Glenn commented out the columns mentioned above in some queries (not all, some had a comment that they should be commented out). I'm uncommenting the columns while parsing so they are alwasys included in the result set. When exporting to Excel I'm making sure they are saved in their own files.
I've also included switches to skip exporting to .sql and / or .sqlplan files
And I've created a switch which will not parse those columns in case someone has specific wishes regarding the result sets and want's me te not touch that.

### Commands to test

 Invoke-DbaDiagnosticQuery -SqlInstance localhost\sql2016 -QueryName "Ad hoc Queries" | Export-DbaDiagnosticQuery -ConvertTo Excel -Path C:\temp

(This breaks without this fix because the column name for the query text is not what I expected)

-NoQueryExport should skip creation of .sql files

(The export to Excel relies on the importexcel module!)

### DIFF
--- a/functions/Export-DbaDiagnosticQuery.ps1
+++ b/functions/Export-DbaDiagnosticQuery.ps1
@@ -63,7 +63,7 @@ function Export-DbaDiagnosticQuery {
 		[string]$Suffix = "$(Get-Date -format 'yyyyMMddHHmmssms')",
         [switch]$NoPlanExport,
         [switch]$NoQueryExport,
-		[switch][Alias('Silent')]$EnableException = $false
+		[switch][Alias('Silent')]$EnableException
 	)
 	
 	begin {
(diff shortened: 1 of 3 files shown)


potatoqualitee:
this is great! Thank you so much @andrekamman 🙇🙇🙇

Issue #2252, opened 2017-09-19
Make Write-Message -WhatIf aware


potatoqualitee:
User had an issue where Write-Message was messing up with -WhatIf. Turns out it was trying to write a string that couldn't exist yet. Can we please make Write-Message whatif aware so that it doesn't log stuff.. or is there a better way to go about it like prepending WhatIf to the log?

IISResetMe:
@potatoqualitee @FriedrichWeinmann mind if I grab this?

potatoqualitee:
That'd be great! thank youuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuuu!

FriedrichWeinmann:
Closing this issue, as likely fixed and not enough information to hunt it down if it isn't

PR #2866, opened 2017-12-14
Add Write-Message -WhatIf support

IISResetMe:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose

This PR addresses https://github.com/sqlcollaborative/dbatools/issues/2252
 

### Approach

Using PowerShell's builtin ShouldProcess facility to void calls to WriteLogEntry, WriteErrorEntry and Set-DbaConfig when -WhatIf is specified

### DIFF
--- a/internal/Write-Message.ps1
+++ b/internal/Write-Message.ps1
@@ -107,7 +107,7 @@ function Write-Message {
 	#>
 	[Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidUsingWriteHost", "")]
 	[Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidDefaultValueForMandatoryParameter", "")]
-	[CmdletBinding(DefaultParameterSetName = 'Level')]
+	[CmdletBinding(DefaultParameterSetName = 'Level', SupportsShouldProcess = $true)]
 	param (
 		[Parameter(Mandatory = $true)]
 		[string]
--- a/internal/Write-Message.ps1
+++ b/internal/Write-Message.ps1
@@ -214,7 +214,9 @@ function Write-Message {
 		}
 	}
 	if ($ErrorRecord) {
-		[Sqlcollaborative.Dbatools.dbaSystem.DebugHost]::WriteErrorEntry($ErrorRecord, $FunctionName, $timestamp, $baseMessage, $Host.InstanceId)
+		if($PSCmdlet.ShouldProcess("Error: $ErrorRecord", 'WriteErrorEntry')){
+			[Sqlcollaborative.Dbatools.dbaSystem.DebugHost]::WriteErrorEntry($ErrorRecord, $FunctionName, $timestamp, $baseMessage, $Host.InstanceId)
+		}
 	}
 	#endregion Handle Errors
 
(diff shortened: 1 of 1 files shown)


wsmelton:
Only problem with doing this is you will end up hiding errors even when you call a command...`Stop-Function` calls this command as well. So to properly handle this I would expect we need to make sure if `Stop-Function` is calling it that it still errors?

Example from using this PR code, I pass in a file that does not even exist...you can't tell it will error or not.

image

IISResetMe:
> The  triggering issue lacks the necessary information to identify the root cause

`-WhatIf` is not a debugging facility - it's intended as a way to consistently _dry-run_ your statements. 
From the issue (#2252), it sounds like the user needs to be able to test/run dbatools with _zero side-effects_. That sounds like the definition of a "dry-run" to me.

> [...], the function does not need to support ShouldProcess, rather it should report when it is called from a command under `-WhatIf`

It _could_, I'm not sure if it _should_. Per-empting/bypassing/escaping or otherwise _simulating_ support for `ShouldProcess` seems like a fundamentally bad idea

> [...] it is assumed to be triggered by a bad call of `Write-Message`

No - these change _only_ assume that a caller **might** want to pass `-WhatIf` in order to conduct a "dry-run" of a script/pipeline/application

potatoqualitee:
fantastic discussion, thanks to you both - looking forward to seeing whats decided 👍

FriedrichWeinmann:
Heya Mathias,

well, let's split the discussion into two parts, in order to keep them separate:
- The original issue
- Whether and how to react to WhatIf when it is used by the calling command (and thus inherited to `Write-Message`

### The original issue
From the vague description, my assumption is that the calling function used a message that was empty if WhatIf was specified, causing an error. Since `Write-Message` is completely indifferent to `ShouldProcess` at the moment, it is very unlikely that the issue was caused by code within `Write-Message`

### How Write-Message should react
I disagree - strongly - with supporting `ShouldProcess` in `Write-Message` directly, for several reasons:

 - It would tell the user, that theoretically it would be writing a message now. The user will have no understanding for this process, but it may very well flood his screen in scenarios where `Write-Message` was used more frequently.
 - It would force the user to confirm every single message (potentially multiple times) when running a command with `-Confirm` active (for example by increasing the ConfirmPreference)
 - It would force the user to deal with a command he doesn't know exists and which doesn't contribute to the functionality s/he is trying to execute.

That said, there _is_ however the issue, that if the messages don't change, depending on whether it is running under `-WhatIf` or not, they _may_ confuse the user (Imagine calling a command with `-WhatIf` and it tells you "Dropping database ..."). Which is why I think that flagging messages as coming from a dry-run command as such may be a good idea.

potatoqualitee:
ay baw, you gave up on this one or no? if this gets merged, you'll be #100 :D

FriedrichWeinmann:
Naw, this one would have too problematic consequences to merge it.
Want to have a discussion on it with Mathias before I go and kill it (unless he manages to convince me otherwise). Hasn't happened yet, hope it will tonight.

potatoqualitee:
k thank you!

IISResetMe:
Yeah, @FriedrichWeinmann has a point about the limited usefulness of having internal routines support ShouldProcess without the user-facing functions supporting it. 

I suggest we identify all user-facing functions with side-effects and start there instead

PR #2884, opened 2017-12-16
Test-DbaLogShippingStatus returning one result

sanderstad:
Fixed the issue that the function only returned the last result


sanderstad:
Proof that it works now.
Detailed
image


Simple
image

potatoqualitee:
Thanks bud! Bon weekend 👍

Issue #2859, opened 2017-12-14
Restore-DbaDatabase Try to put FileStream files on the source path


ClaudioESSilva:
## Bug Report
While implementing a FileStream check on `Test-DbaMigrationConstraint` I found that FileStream files will try to restore on the **source** default path. If the path does not exists will fail.

### General Troubleshooting steps

- [x] Verified running latest release of dbatools?
- [x] Verified errors are not related to permissions?
- [x] Can duplicate in new/fresh PowerShell.exe session?

## Steps to Reproduce
 - [x] Attach any screenshots (if possible/allowed)
`Backup-DbaDatabase -SqlInstance sql2012 -Database db3 -BackupDirectory "\\dc\sql\SQL2012"  | Restore-DbaDatabase -SqlInstance sql2016 -DestinationDataDirectory "C:\temp" -DestinationLogDirectory "C:\temp"

![image](https://user-images.githubusercontent.com/19521315/33999158-a221aa86-e0e1-11e7-8d86-dfa4d94b3560.png)

 - [x] Attach logs generated by dbatools (`Get-DbatoolsLog`)
[DbatoolsLog_Backup_Restore_one_db.txt](https://github.com/sqlcollaborative/dbatools/files/1559728/DbatoolsLog_Backup_Restore_one_db.txt)
Here I can see the `Test-DbaDatabaseCompatibility` being called for every single DB even that I just passed one.
 - [X] Attach output from PowerShell console (if possible/allowed)
See screenshot
`
## Feature Request
Implement a new parameter like `-DestinationDataDirectory` and `-DestinationLogDirectory` for FileStream filegroups.

### Problem to solve
Being able to restore a database that uses FileStream to a specific directory

niphlod:
closed via #2892

PR #2888, opened 2017-12-18
FileStream cmdlets

Stuart-Moore:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
We don't have a means of getting FileStream status or setting that status. Need it for #2859 as we need to check/enable before restoring a FileStream DB

### Approach
Get/Set pair of commands 

### Commands to test
Get-DbaFilestream -SqlInstance server1\instance1

Set-DbaFileStream -SqlInstance server1\istance1 -FileStreamLevel 2
Above will prompt for confirmation for the change and the server restart

Set-DbaFileStream -SqlInstance server1\istance1 -FileStreamLevel 2 -Force
Same as previous by will just do the job

Get-DbaFilestream -SqlInstance server1\instance1, server\instance, prod\hr | Where-Object {$_.FileStreamStateId -gt 0}  | Set-DbaFileStream -FileStreamLevel 0 -force
Queries a bunch of servers, and disables Filestream on any of them with it enabled.


ClaudioESSilva:
Hi @Stuart-Moore,

we already have the command to do this, I mean, in this case, is to set up inside SQL Server so we can use the `Get/Set-DbaSpConfigure` commands.
Example: `Get-DbaSpConfigure -SqlInstance $sourceServer -ConfigName FilestreamAccessLevel` (I have used on the MigrationConstraint command https://github.com/sqlcollaborative/dbatools/blob/master/functions/Test-DbaMigrationConstraint.ps1#L198). The main difference in comparison with yours is the service restart, on this commands it only informs if you need to do it.

What we do not have is a command to set it at the SQL Server Configuration Manager level outside the instance. And I just found that without it enabled... we can't restore the database.

potatoqualitee:
That's a good point, @ClaudioESSilva - I added the psd1 changes w/o looking more into what the commands were. If we were to use the commands, it should probably be Get/Enable/Disable

Stuart-Moore:
This does the Config manager level change. Not the Sql config level change

potatoqualitee:
Should an `-IncludeSpConfigure` or some better named param be added so that it's a thorough change?

ClaudioESSilva:
Unless I'm missing something, I disagree @Stuart-Moore 

image

Stuart-Moore:
I asked on slack and was told to write something.

enable/disable is too binary for something with at least 3 states (there's internal SMO code for state 4, but that barfs atm when you set it)

wsmelton:
@Stuart-Moore Working under `$server.Configuration` is going to be the instance-level configuration.

To handle the service-level configuration you will need to do something similar to [this](https://haydenhancock.wordpress.com/2013/03/12/enable-filestream-via-powershell/), but using our CIM commands.

potatoqualitee:
Can you add a required `-Type` param and ValidateSet with the 3 different types (using human words, then transform them to 0, 1, 2, etc)

wsmelton:
We also need to clarify one item on the command names. Are we going to differentiate between service-level and instance-level configuration in set/disable?

Service-level configuration there are three level of settings to consider:
 - Enable Filestream for T-SQL Access
 - Enable Filestream for I/O Access
 - Windows share name
 - Allow remote clients access to Filestream data

At the instance-level if I recall setting it to 2 requires that you have I/O access enabled at the service-level, but has been a while since I had to set this up on a database.

Stuart-Moore:
OK. If anyone wants to run with this feel free. I know almost nothing about Filestream and just wanted this to get @ClaudioESSilva restore request sorted.

potatoqualitee:
@ClaudioESSilva - will you take this one over? He even included tests 🤗

wsmelton:
What will the command names be for the service level config?

potatoqualitee:
I think it should be all in one with an IncludeServiceLevel

potatoqualitee:
How’s this going? Anyone taking it on?

wsmelton:
What I have written so far...

image

Working on test file now.

wsmelton:
@ClaudioESSilva So I've found out testing that the AccessLevel values for the instance and service match up with one exception. Service level goes from 0 - 3, with 3 being full access. So I added two hash tables to the code in order to do a lookup on the value, make it more dynamic.

```powershell
		$idServiceFS =[ordered]@{
			0 = 'Disabled'
			1 = 'Transact-SQL access'
			2 = 'Transact-SQL and I/O access'
			3 = 'Transact-SQL, I/O and remote client access'
		}
		$idInstanceFS =[ordered]@{
			0 = 'Disabled'
			1 = 'Transact-SQL access enabled'
			2 = 'Full access enabled'
		}
```

So the interesting part is if you have the service only set to level 1 and then try and set the instance to level 2 (Full Access), then the configured value on the instance (what is returned by `Get-DbaSpConfigure -SqlInstance $server -ConfigName FilestreamAccessLevel | Select-Object ConfiguredValue, RunningValue`) will never update. I have restarted the SQL14 instance multiple times and the configured value never updates:

image

The purpose in all this, I'd like to add a flag `IsConfigured` that we can use with the restore process. I'm not sure if in this setup a restore would work (don't have a FS database on me).

@potatoqualitee @Stuart-Moore does this sound accurate/usable?

ClaudioESSilva:
Thanks, @wsmelton Nice finding.
If you set the service level to other value and restart the instance level change?
If yes, maybe we should add some warning to that combination.

PS: Just to be clear, this is not a real-life problem to me (right now). I was implementing an improvement to Test-DbaMigrationConstraint and found that the Restore-DbaDatabase didn't handle this filegroups. And now we are here! The beauty of implementing a new feature 😄 
I don't have any database with filestream (in my work environment), I just used the lab to create one even without filestream data (just filegroup/file).

wsmelton:
A restart makes no difference. If you set the service to a lower level to the instance it will not update the instance configuration.

potatoqualitee:
I'm gonna close this PR but leave the branch open. It's kinda lingering :/ Feel free to continue commenting.

sheldonhull:
I came across this and not sure what's remaining on it. I could use this as I deploy out a readyroll project requiring configuration of filestream. What work would help finish this?

wsmelton:
SMO has to be fixed 😁 

there is a limitation in SMO itself to properly handling filestream configuration...at least from my findings.

sheldonhull:
Such a pain! Ok. Do you have a SMO issue opened already?

potatoqualitee:
I read on a forum that

````
There are actually two places you need to enable FILESTREAM support, in the engine and in the server. Enabling FILESTREAM support in the engine tells SQL Server to load the drivers that are required to support FILESTREAM, while enabling it in the server is what activates and deactivates it. So you can't activate it unless the engine was loaded with support for it.
````

So I'll try engine first then WMI, tho everyone is saying SMO is read-only

wsmelton:
Some of the code was fixed to enable it at the service level. It is just changing it is where the bug is.

potatoqualitee:
yah, gonna do a combo 💪

PR #2920, opened 2017-12-22
made dbowner 1.0 compliant and warned when dbs were inacessible

potatoqualitee:
/cc @gbargsley @SQLDBAWithABeard

### DIFF
--- a/functions/Test-DbaDatabaseOwner.ps1
+++ b/functions/Test-DbaDatabaseOwner.ps1
@@ -13,15 +13,15 @@ function Test-DbaDatabaseOwner {
 			Best Practice reference: http://weblogs.sqlteam.com/dang/archive/2008/01/13/Database-Owner-Troubles.aspx
 
 		.NOTES
-			Tags: 
+			Tags:
 			Author: Michael Fal (@Mike_Fal), http://mikefal.net
 			Website: https://dbatools.io
 			Copyright: (C) Chrissy LeMaire, clemaire@gmail.com
 			License: GNU GPL v3 https://opensource.org/licenses/GPL-3.0
 
-		.PARAMETER SqlInstance 
+		.PARAMETER SqlInstance
 			Specifies the SQL Server instance(s) to scan.
-			
+
 		.PARAMETER SqlCredential
 			Allows you to login to servers using SQL Logins instead of Windows Authentication (AKA Integrated or Trusted). To use:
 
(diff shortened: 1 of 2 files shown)


wsmelton:
Working on the test for you @potatoqualitee I have good bit of fixed, just checking one area that is still pestering me (see what I did there 🌝 )

wsmelton:
Ok @potatoqualitee this should pass 🤞 

A few things to note on both the function and the test:

- Testing on the wrong login was actually failing for a few reasons, one of which was a legit issue in the function. When you call `Test-Bound` to see if `TargetLogin` was being passed in, and if it is not then we should check for "sa" this was validating the wrong way. This particular section:

```powershell
			# dynamic sa name for orgs who have changed their sa name
			if (Test-Bound -ParameterName TargetLogin) {
				$TargetLogin = ($server.logins | Where-Object { $_.id -eq 1 }).Name
			}
```

Will always return true because you passed in a TargetLogin. In this case we only want to set it to the sa when TargetLogin is not passed in, so I just added `-Not` to that call, this makes Test-Bound return a $true "if the parameter was not passed".

```powershell
			# dynamic sa name for orgs who have changed their sa name
			if (Test-Bound -ParameterName TargetLogin -Not) {
				$TargetLogin = ($server.logins | Where-Object { $_.id -eq 1 }).Name
			}
```

- Testing for the login I changed to "NotMatch" instead of "NotContains", it is generally safer to use for comparing a single string to an array that I've seen. It may not really matter though as I tried it both ways and it still works. ??? 🤷‍♂️ 

- On the test I have pretty pictures that follow to explain, but your mock on the connection command has the instance name set to `Name` property. In your test you check for `SqlInstance` which we set using the `DomainInstanceName` and not the `Name` property from the server object. I simply updated your mock to have the new property name and all is good....now pretty pictures:

Before change:
image

After change:
image

- Only other thing I adjusted in the test was referencing the first object of the array, instead of the whole thing.

Now we see if I got it right on Appveyor grounds.

wsmelton:
Oh, forgot one pretty big change that was required.

We can't test for "throw" when we use `Stop-Function` because we are not throwing it anymore. So I updated this to a similar thing we do in other tests, think @niphlod added this. I mocked `Stop-Function` and then just asserted that it was called properly.

```
                Mock Stop-Function {}

                $null = Test-DbaDatabaseOwner -SqlInstance 'SQLServerName' -TargetLogin 'WrongLogin'
                $assertMockParams = @{
                    'CommandName' = 'Stop-Function'
                    'Times' = 1
                    'Exactly' = $true
                }
                Assert-MockCalled @assertMockParams
```

wsmelton:
Failure is not for these changes.

potatoqualitee:
YESSSSS thank you, @wsmelton !

PR #2926, opened 2017-12-24
Last review cleanups.

wsmelton:
1.0 standards and adding validation tags for Bill Of Health.

The `New-DbaSqlConnectionStringBuilder.ps1` has been getting flagged in the gallery for failing rules on the PSSA scan they perform. Which apparently they enable almost every rule compared to what we check. So we might need to adjust our testing to follow their standard, possibly...something to discuss in Slack dev channel after the holidays.


potatoqualitee:
Thanks, bud!

PR #2930, opened 2017-12-25
Allow overriding of the uptime check

Stuart-Moore:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Allow Find-DbaUnusedIndex to be used on an instance that's been online for less than 7 days.

I needed to check that an index had been preferred  on an overnight run with an instance that had been restarted 2 days before, so scratching a personal need really

### Approach
Added IgnoreUptime switch. Default is as now and function will exit, set the switch and it'll return any data, though still printing a warning to remind the user that the output may not be completely useful


potatoqualitee:
One pass 💯

PR #2944, opened 2017-12-28
fixes for backup info output and a test with quetions

potatoqualitee:
hey @Stuart-Moore - can you take a look? i had to update two columns. and also fixed the test but have questions about it


potatoqualitee:
hey @Stuart-Moore - can you take a look? i had to update two columns. and also fixed the test but have questions about it

potatoqualitee:
one pass

PR #2946, opened 2017-12-28
Invoke-DbaAdvancedRestore, fixes #2940

### LINKED ISSUE #2940: Restore-DbaDatabase with -OutputScriptOnly sets database to RESTRICTED_USER

- [x] Verified running the latest release of dbatools?
- [x] Can duplicate in new/clean PowerShell session (clean = `powershell -NoProfile`)?

### Version Information
 - Operating System (Name|Version):  Support package attached.
 - PowerShell Version:  Support package attached.
 - SQL Server (Edition|Version):  Support package attached.

## Steps to Reproduce
Support package attached.
 - [ ] Attach any screenshots (if possible/allowed)
 - [ ] Attach logs generated by dbatools (`Get-DbatoolsLog`)
 - [ ] Attach output from PowerShell console (if possible/allowed)

## Feature Request
When using the -OutputScriptOnly option, the targeted database should not be put into RESTRICTED_USER mode.

### Problem to solve
Same as above.

### Additional information
[dbatools_support_pack_2017_12_28-11_49_11.zip](https://github.com/sqlcollaborative/dbatools/files/1591663/dbatools_support_pack_2017_12_28-11_49_11.zip)

niphlod:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #2940)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Approach
Address the boolean logic needed to avoid setting the db to restricted_user if not needed

### DIFF
--- a/functions/Invoke-DbaAdvancedRestore.ps1
+++ b/functions/Invoke-DbaAdvancedRestore.ps1
@@ -1,340 +1,340 @@
-Function Invoke-DbaAdvancedRestore{
-    <#
-    .SYNOPSIS
-        Allows the restore of modified BackupHistory Objects
-        For 90% of users Restore-DbaDatabase should be your point of access to this function. The other 10% use it at their own risk
+function Invoke-DbaAdvancedRestore {
+	<#
+	.SYNOPSIS
+		Allows the restore of modified BackupHistory Objects
+		For 90% of users Restore-DbaDatabase should be your point of access to this function. The other 10% use it at their own risk
 
-    .DESCRIPTION
-        This is the final piece in the Restore-DbaDatabase Stack. Uusally a BackupHistory object will arrive here from Restore-DbaDatabse via the following pipeline:
-        Get-DbaBackupInformation  | Select-DbaBackupInformation | Format-DbaBackupInformation | Test-DbaBackupInformation | Invoke-DbaAdvancedRestore
+	.DESCRIPTION
+		This is the final piece in the Restore-DbaDatabase Stack. Uusally a BackupHistory object will arrive here from Restore-DbaDatabse via the following pipeline:
+		Get-DbaBackupInformation  | Select-DbaBackupInformation | Format-DbaBackupInformation | Test-DbaBackupInformation | Invoke-DbaAdvancedRestore
 
-        We have exposed these functions publically to allow advanced users to perform operations that we don't support, or won't add as they would make things too complex for the majority of our users
+		We have exposed these functions publically to allow advanced users to perform operations that we don't support, or won't add as they would make things too complex for the majority of our users
 
-        For example if you wanted to do some very complex redirection during a migration, then doing the rewrite of destinations may be better done with your own custom scripts rather than via Format-DbaBackupInformation
+		For example if you wanted to do some very complex redirection during a migration, then doing the rewrite of destinations may be better done with your own custom scripts rather than via Format-DbaBackupInformation
 
-        We would recommend ALWAYS pushing your input through Test-DbaBackupInformation just to make sure that it makes sense to us.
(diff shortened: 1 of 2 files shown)


niphlod:
ok. Adding regressiontests, then we should be oki.

niphlod:
passed! @Stuart-Moore , sorry, I took a stab at VSCode reformat without splitting the commits.

potatoqualitee:
I just looked with `?w=1` and saw your changes were minimal and as discussed. Thank you very much for the test, too! I will merge and test since the tests pass and the code looks sane (and im brushing my teeth as promised ;)).

niphlod:
Most of those issues would be solved if - against every other core devs recommendation - we convert to using spaces instead of a tab. Apparently I didn't stress it enough in the past, but once every while lemme stress again....
dbatools's convention requires specific indentation settings that are not common on all freeish/generally available/major editors.
Powershell didn't come with a linter and is ending up looking like perl. ScriptAnalyzer is slowly getting to a point where a linter is due, and part of it has been integrated inside VSCode.
I expect SA to enforce more and more rules (see Invoke-Formatter/CodeFormatting) about linting a script because the ps community is growing and all the scripts and module are harder and harder to read. 
There isn't already a rule for tabs/spaces indentation, but if we go for 4 spaces instead of tabs, irregardless of the editor, code would look the same (also on gh)... and editors (np++, st3, vscode, atom, ise, etc) won't need special settings to accomodate dbatools's own PS coding conventions.
If instead we want to continue using the tab convention.... I hope everybody contributing code with proper settings in VSCode or a better tool, else as soon as @niphlod sees:
- bad indentation
- trailing spaces
- camelcase for-if-else-foreach
- unaligned hashmaps
- etc
this will happen again and again.

Zhunya:
oh, going to fix my code before submitting

Stuart-Moore:
Maybe stress is somewhere obvious? Appears that people don't know that this is a 'thing' (I contribute a bit and don't, so that's probably a sign). not everyone can keep up with Slack. And I've not been picked up on previous PRs, so guessing it's luck of the draw on who does the review.

Just checked the dbatools style pages I know about and can't see anything about this. 
https://github.com/sqlcollaborative/dbatools/wiki/Do-And-Do-Not
https://github.com/sqlcollaborative/dbatools/wiki/Style-Guide

don't appear to reference all of the issues above.

Write it up, shove it in the wiki, and link it from the front readme and people will  take notice.

PS, my VS code reckons it's using spaces and not tabs:
image
(That's the restore tests file prior to this PR)

Thanks for sorting this though :+1:

niphlod:
@Stuart-Moore that's what I'm telling. The majority of the code in dbatools is formatted using standards that rely more on "the first contributors used editor x so you should all use that" rather than a specific need/common usage, and it's not clearly documented. 
 
Now that linters are arising maybe we should go with them ("going with the flow") with a bit more logic to it rather than "assuming" common rules set by "the past". 
If you use VSCode already a simple "format document" before saving at least fixes most of the things (common indentation (all spaces or all tabs), braces, hashmaps, unused vars, most of the piping, etc) by itself, just leaving the spaces/tabs "war" as it is. 
While the spaces/tabs "war" is still "undocumented", I see once in a while files with:
- unaligned braces
- mixed indentation (half tabs, half spaces)
- unaligned hashmaps
- unused variables
I then proceed to edit those in a few seconds via the editor, usually making a commit on its own (in this case I forgot, sorry), in order to isolate "cosmetics" changes.

tl;dr: I strongly vouched against using tabs since the beginning (I still am ^_^), but have been asked to adhere to tabs for code I originally committed using spaces...so, unless told otherwise, I'm going with the "all dbatools code should look alike" rule rather than my personal preference.

PR #2959, opened 2017-12-31
Replace leading tabs with 4 spaces

IISResetMe:
Replace leading tabs with 4 spaces in all script files with no here-strings

This is an experimental PR to judge and discuss the outcome of converging on 4 space (`"$([char]0x32)" * 4`) tabwidth throughout the project.

**DON'T MERGE THIS PR YET**, current state only covers ~2/3rds of the code base


IISResetMe:
The initial attempt had some unintended breaking consequences, so I changed strategy to let VS Code auto-format all the scripts and then manually go through each and remove trailing tabs, and then finally replace leading tabs with 4 spaces everywhere but in multiline strings

potatoqualitee:
merging to local branch so i can fix a couple of my new files 👍

PR #2960, opened 2017-12-31
Get-DbaCpuUsage - initial commit

potatoqualitee:
image

image

### DIFF
--- a/functions/Get-DbaCpuUsage.ps1
+++ b/functions/Get-DbaCpuUsage.ps1
@@ -24,7 +24,10 @@
 
 	.PARAMETER Credential
 		Allows you to login to the Windows Server using alternative credentials.
-		
+	
+	.PARAMETER Threshold
+		CPU threshold.
+	
 	.PARAMETER EnableException
 		By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
 		This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
--- a/functions/Get-DbaCpuUsage.ps1
+++ b/functions/Get-DbaCpuUsage.ps1
@@ -1,0 +1,141 @@
+﻿Function Get-DbaCpuUsage {
+<# 
+	.SYNOPSIS 
+		Provides detailed CPU usage information about a SQL Server's process
+		
+	.DESCRIPTION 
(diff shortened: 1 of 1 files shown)


functions/Get-DbaCpuUsage.ps1
```diff
@@ -0,0 +1,144 @@
...
+				Write-Message -Level Verbose -Message "Connecting to $instance"
+				$server = Connect-SqlInstance -SqlInstance $instance -SqlCredential $sqlcredential
+			}
+			catch {
+				Stop-Function -Message "Failure" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
+			}
+			
+			$processes = Get-DbaProcess -SqlInstance $server
+			$threads = Get-DbaCmObject -ComputerName $instance.ComputerName -ClassName Win32_PerfFormattedData_PerfProc_Thread -Credential $Credential | Where-Object { $_.Name -like 'sql*' -and $_.PercentProcessorTime -ge $Threshold }
+			$spidcollection = $server.Query("select spid, kpid from sysprocesses")
```
wsmelton: I would disagree with using a view that is documented to be on the depreciation list. This should only be used to cover older versions of SQL Server.

potatoqualitee: i saw it was deprecated but i'm not convinced they'll actually remove it ;)

wsmelton: It won't go away because Activity Monitor is SSMS utilizes it. However, what is wrong with using current views that give the same information?

wsmelton: It is not really the point that it is not going away but we are using views on the depreciation list, which by Microsoft terms is a bad practice. First thing folks see is that we are using SQL Server 2000 views.

Something like this should provide you the same information...

```sql
select s.host_name, s.host_process_id, t.os_thread_id, s.session_id
from sys.dm_exec_sessions s
   join sys.dm_exec_requests er on s.session_id = er.session_id
   join sys.dm_os_workers w on er.task_address = w.task_address
   join sys.dm_os_threads t on w.thread_address = t.thread_address;
```

potatoqualitee:
yay thanks for the code, @wsmelton ! Added

PR #2962, opened 2018-01-01
Test-DbaBackupInformation, fix and speed

niphlod:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Purpose
Restore-DbaDatabase for small restores started suffering from O(n) restore times because it uses Get-DbaDatabaseFile some time ago, with n being the number of databases present on the destination instance. 
Finally had the time to review the code to introduce fixes that overcome the problem issuing one query per instance, instead of one query per database, to fetch the exact same informations needed.
This is probably due to the fact Get-DbaDatabaseFile in the beginning was speedier, but for the kind of info Test-DbaBackupInformation is using them it is now _clearly_ an overkill. _Clearly_ of course is subjective, but if you tried to use Restore-DbaDatabase on a 30ish databases instance, you may have noticed already. If you tried against a 150ish databases instances, you may have started asking "why is my instance so slow?" :-P

Moreover, Restore-DbaDatabase hit some problems (not due to the recent "isAccessible" patches, but similar): Get-DbaDatabaseFile skips over unaccessible databases but that skip means Test-DbaBackupInformation fails to know if a file is used by another database, when it is offline. Again, Get-DbaDatabaseFile was more and more a limitation.

### Approach
Instead of two calls to Get-DbaDatabaseFile (which in turn triggers a query for each hosted database) we just use sys.master_files or sysaltfiles to know which files are tied to registered databases, being offline or not. 

This reduces the time to O(1). 

Doing that, I saw another corner-case: it may happen that a file is present on the destination which is not tied to any database, not even offline ones (one that could be inspected beforehand with find-dbaorphanedfile). 
So, we fire one additional query (via the recent feature of Test-DbaSqlPath being able to process more than one path) to know if the "programmed destination" indeed exists or not.

Reorganizing this call also reduces the time from O(n) to O(1) where n was the number of physical files of the database to be restored (yup, "prematury optimization is the root of all evil", but since a reorg was needed in any case, why not?)

In case dest files already exists and are not tied to any database, it won't register as an issue ONLY if -WithReplace is passed. The WITHREPLACE option passed down the line to the restore is able to re-use the same destination (as we all know, for "registered" databases) but it's needed also in case an existing file is already in place (for any other file....and I gained this knowledge just now ^_^).


niphlod:
okay, let's see why tests fail :-(

potatoqualitee:
Tests have been running clean! Should be related to your updates.

niphlod:
yeah I know. it's mocks that need to be updated accordingly. I'm updating them, my fault.

niphlod:
okay. unittests for test-dbabackupinformation just needed a little mock love. Funny enough, I get the same fail on restore-dbabackup even from the development branch. Need to investigate a little bit more on that.

niphlod:
okay, this passed but I need to raise an hand and request help/guidance.

See [line 158](https://github.com/sqlcollaborative/dbatools/pull/2962/files#diff-f265cc4b262d77e9d9afad3bce2e2e88R158) on Test-DbaBackupInformation.
It's the additional line that checks for a target filename already present at destination that is not tied to any database.

If that check is added, the test "Page level restores" on "Restore-DbaDatabase" [fails](https://ci.appveyor.com/project/sqlcollaborative/dbatools/build/0.9.10142/job/ameldiln33cqqs17).
Formally, from the synopsis on Test-DbaBackupInformation, I'd go with adding the check. I see Test-DbaBackupInformation as the tool that tells me "beforehand" if a set of backupfiles will end in an error or not (and so it seems to be used within higher level functions).
Without it, a restore could go wrong because of an orphaned file lying around.
If one doesn't care about orphaned files (i.e. no detach-attach are a daily operation), -WithReplace works perfectly.
 
That error on Restore-DbaDatabase tests brought to my attention the "AllowContinue" parameter, that seems to exist to "cope with existing files". IMHO it just correlate to WithReplace...but I may very well be wrong.
If you go in Restore-DbaDatabase's code, AllowContinue basically trumps checks done in Test-DbaBackupInformation.
Now, without reading all of the backup's functions code, I humbly ask to core devs:
 * the failing test on Restore-DbaDatabase should be rewritten to just use WithReplace (and the code from Restore-DbaDatabase adapted to handle it) ?
  * Is AllowContinue really needed given WithReplace is already there? What's the usecase ?
  * Should Test-DbaBackupInformation raise a flag or not a restore chain that will end in an error because the target file already exists at destination, but is not tied to any database ?

potatoqualitee:
@Stuart-Moore, @wsmelton, @ClaudioESSilva, @PowerDBAKlaas - if you guys have time, can you take a look at niph's question.

Stuart-Moore:
Happy New Year

Just some general comments as I'm out and about  (not back in office till tomorrow)

If we're dropping the SMO fundamentalism then going to the tables was always going to be the better way. (Personally I'd suggest that Get-DbaDatabaseFile was rewritten to do that, and cope with being called here, so we've one good function (that was knocked back last time I tried it)). Can it at least be dropped out as Internal function, (personally don't like functions within functions as maintenance and duplication becomes an issue)

AllowContinue and WithReplace are very seperate use cases. WithReplace says replace anything preexisting, but only if it passes all tests. AllowContinue is bascially saying just give it a go (for example it will attempt to restore as far along a tree as it can, ie; if you've got an LSN break it'll just plough on till it gets the restore error at the missing file).

Page Level restores DO NOT want to have WithReplace set. It's a different use case. It wants the files to exist as we're just replacing a couple of pages, and you don't want WITH REPLACE in the eventual restore string either. They're in seperate ParameterSets at the Restore-DbaDatabase level for that reason.

Yes, it should complain. We (dbatools) don't know what that file is, and we can't make that decision for the user. All we can do is flag it up and let them make the decision (always the way to go in general).

potatoqualitee:
@niphlod  can you repaste your Test-DbaBackupInformation.ps1 with your spaces and mark as conflict resolved?

niphlod:
btw, we can make "yet another function" instead of having the one included in this, ala get-dbadatabaseFile, the problem is that soon enough "it's a gift that keeps on giving". Get-DbaDatabaseFile originally started with "enumeration in SMO is soooo slow"....but that soonly reverted to one query per db to attach "yet another property useful for something/someone else".
I guess we can create a Get-DbaDatabasePhysicalPaths in internal and put heavy comments in there that it must remain speedy and return results both for online and offline databases.

niphlod:
> Page Level restores DO NOT want to have WithReplace set. It's a different use case. It wants the files to exist as we're just replacing a couple of pages, and you don't want WITH REPLACE in the eventual restore string either. They're in seperate ParameterSets at the Restore-DbaDatabase level for that reason.

> Yes, it should complain. We (dbatools) don't know what that file is, and we can't make that decision for the user. All we can do is flag it up and let them make the decision (always the way to go in general).

Ookay then, those files need to be flagged. 
But, if I flag it as a file that exists already at destination, Page Level restore test fails. 
Is that suppose to happen ?

Stuart-Moore:
Thought those were 2 independent events you were talking about

Anything we can't prove what it's owned by we shouldn't touch, because we don't know what owns it (could be a 2nd instance on the box). So that shouldn't be an issue with Page restores, as we can't page restore into a data file not owned by the instance we're running against here.

The test failing you reference above, is it because you're not excluding the files owned by the target database which you've already checked for on L148 which handles WithReplace?  

Now had a look at the code (and T-SQL), and I was wrong. We force WithReplace if the ParameterSet is RestorePage, so it's coming through that way in the pipeline

On functions, use an Internal function. Should be small and fast, no added extras, if people want slow and bloated we've public functions for that :) Compare Connect-SqlInstance and Connect-DbaInstance for instance. Path tests should be very common, so there should be a standard internal way of testing (validatiing copy destinations, testing before a move/rename of a file, etc, etc)

niphlod:
- okay, I'll craft a PR that extrapolates the internally used function in a separate "internal" one. 
- dunno what is going on, I didn't fiddle with all code paths there, that was why I raised an hand. That additional check flags as "unverified" the chain to be restored in one corner-case I *thought* it should be flagged has the repercussion of making the test for page restore fail

Stuart-Moore:
I'm in the middle of building a new test environment (got by MBP back and
trying to move off of work owned kit), once i've got that done I'll step
through it and see what's going on

potatoqualitee:
@Stuart-Moore and @niphlod - i believe you both have merge capabilities. Please merge as desired.

niphlod:
okay, lemme extract the sub-function and do a rebase here, then @Stuart-Moore can inspect in more depth if the added check needs to be decommented or is completely wrong.

PR #2965, opened 2018-01-02
tabs to spaces

potatoqualitee:
From original PR:
Replace leading tabs with 4 spaces in all script files with no here-strings

This is an experimental PR to judge and discuss the outcome of converging on 4 space (`"$([char]0x32)" * 4`) tabwidth throughout the project.

**DON'T MERGE THIS PR YET**, current state only covers ~2/3rds of the code base
________________________________

The initial attempt had some unintended breaking consequences, so I changed strategy to let VS Code auto-format all the scripts and then manually go through each and remove trailing tabs, and then finally replace leading tabs with 4 spaces everywhere but in multiline strings.

Using VS Code to format the scripts also had these benefits:
1. Consistent K&R-style bracing
2. Alignment of value expressions in hashtable literals
3. Consistent space after commas in array literals
4. Space around operators in expressions


potatoqualitee:
updated PR name from cute name to regular name since 600+ files will be tagged with it

potatoqualitee:
thanks all!

PR #2968, opened 2018-01-03
Test-DbaBackupInformation, need for speed.

niphlod:
followup of https://github.com/sqlcollaborative/dbatools/pull/2962 .

@Stuart-Moore : relevant lines are 138-139

### DIFF
--- a/functions/Test-DbaBackupInformation.ps1
+++ b/functions/Test-DbaBackupInformation.ps1
@@ -125,19 +125,20 @@ function Test-DbaBackupInformation {
                 $DBHistoryPhysicalPathsExists = ($DBHistoryPhysicalPathsTest | Where-Object FileExists -eq $True).FilePath
                 foreach ($path in $DBHistoryPhysicalPaths) {
                     if (($DBHistoryPhysicalPathsTest | Where-Object FilePath -eq $path).FileExists) {
-                        if (($path -in $DBFileCheck) -and ($WithReplace -ne $True -and $Continue -ne $True)) {
-                            Write-Message -Message "File $path already exists on $SqlInstance and WithReplace not specified, cannot restore" -Level Warning
-                            $VerificationErrors++
+                        if ($path -in $DBFileCheck) {
+                            #If the Files are owned by the db we're restoring check for Continue or WithReplace. If not, then report error otherwise just carry on 
+                            if  ($WithReplace -ne $True -and $Continue -ne $True) {
+                                Write-Message -Message "File $path already exists on $SqlInstance and WithReplace not specified, cannot restore" -Level Warning
+                                $VerificationErrors++
+                            }
                         }
                         elseif ($path -in $OtherFileCheck) {
                             Write-Message -Message "File $path already exists on $SqlInstance and owned by another database, cannot restore" -Level Warning
                             $VerificationErrors++
                         }
                         elseif ($path -in $DBHistoryPhysicalPathsExists) {
-                            if (-not $WithReplace) {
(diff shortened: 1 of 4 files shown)


Stuart-Moore:
You're still 'double dipping' on the path tests. A non optimised version which works:
```
                        elseif ($path -in $DBHistoryPhysicalPathsExists -and $path -notin ($RegisteredFileCheck).PhysicalName) {
                                Write-Message -Message "File $path already exists on $($SqlInstance.NetName), not owned by any database in $SqlInstance, will not overwrite." -Level Warning
                                $VerificationErrors++
                        }
                      
```
That final check (If I've followed your logic properly) should be checking for things that exist but aren't "owned" by the Instance. You check those in the previous 2 conditions which check for WithReplace and Continue, so you don't want to check for them again without those so we need to filter them out.

Also, those are files that just happen to be on the Server. as they're not owned by the instance you're restoring to we can't make any assumptions about them, or if we can overwrite them (could be an active mdf attached to another instance) so should be a warning and a fail for that restore.

niphlod:
doh... I can't understand.
`$DBFileCheck` are files that are registered to the same database we're trying to restore.
`$OtherFileCheck` are files that are registered to databases we're not trying to restore.
`$DBHistoryPhysicalPaths` are files which are targeted by the restore process.
`$DBHistoryPhysicalPathsExists` are files which are targeted by the restore process which exist already at destination.
The pseudo code is as follows:
```
foreach $path in $DBHistoryPhysicalPaths {
    if $path -in $DBFileCheck {
        the file we're trying to target is already one of the database we'e trying to restore, so we warn if continue or replace is not passed
    } elseif $path -in $OtherFileCheck {
        the file we're trying to target belongs to another database on the instance, so it'll fail, plain and simple
    } elseif $path -in $DBHistoryPhysicalPathsExists {
         you end up here ONLY if both are not true: $path is not one of the target database, nor any other database. 
         the file we're trying to target doen't belong to any database, could be an orphan that doesn't belong to anyone OR owned by another instance. if you don't pass REPLACE to the underlying command, the restore would fail. if you pass REPLACE, if the file is not owned by another instance the replace can go fine. If it's owned by another instance ..... we can't really detect it without hindering the restore times.
    }
}
```
as such, the third condition, the one I added, covers a case that is not covered in any of the previous ones.... except one culprit (I realize it now): the first condition has more logic, so you end up here even if the restore targets the same database but you didn't pass restore or continue, and MAYBE this is the issue ?!

Stuart-Moore:
:+1: That's it.

It's a fall through problem. You're first condition is very restrictive so in this case it's not being matched, which was dropping it to a more open condition further down the tree.

The first one would be better as a nested if, as it's not really an AND condition, you want to match there even if WithReplace is set as you've handled that file for this cycle through the loop.  

The other option would be to start with 
```
if (($path -in $DBFileCheck) -and ($WithReplace -eq $True -or $Continue -eq $True)) 
```
to shortcut when everything is good.

This is why the restore stuff is riddle with nested IF statements. Not pretty, but it keeps the logic seperate and obvious (or I forget what I was trying to do when I go back to it :wink:)

niphlod:
that part I didn't change, I just sidestepped Get-DbaDatabaseFile. Feel free to edit, I'm busy for one or two days :-(

Stuart-Moore:
Whereas I'm back at my desk, so need something to keep me occupied whilst swearing at idiot providers ;)

You'd added the general path test, which was an improvement on mine as I'd not considered checking for non Instance owned files.

Just pushed an edit up.

PR #2975, opened 2018-01-03
Find-DbaUnusedIndex Remove Out-GridView

js0505:

Purpose is to remove Out-GridView pipe command.

This change gives us the warranty of being able to utilize this command in other modules such as Pester.

find-dbaunusedindex


potatoqualitee:
almost! The columns need to be PowerShelled - so remove all underscores and replace with caps.

potatoqualitee:
thanks bud!

PR #2980, opened 2018-01-04
Find-DbaDatabaseGrowth - hot fix and rename

wsmelton:
- Rename `Find-DbaDatabaseGrowthEvent` to `Find-DbaDbGrowthEvent`
- Adjust local and UTC time output to be via parameter `-UseLocalTime`, the default will be UTC time as it was but if the parameter is passed it will output the time as-is in default trace (which is local time of the instance).

Updated psd1 and psm1 for rename of the command.


potatoqualitee:
sweet thanks - failure was due to another command

PR #2984, opened 2018-01-04
don't detect empty lines

niphlod:
our bfdl asked for it.


potatoqualitee:
@FriedrichWeinmann and i appreciate it! If PSS ever supports this, we can remove ;) Will merge once tests pass

Issue #2985, opened 2018-01-04
SQL Server Patch Management / Early Warning System


sql-williamd:
## Feature Request

We are seeing the storm brewing around the Intel CPU scandal. Microsoft has already issued emergency patches for SQL Server 2016 and 2017, with more to come.

There is already some version checking code, along with the build information table here: https://sqlcollaborative.github.io/builds

These are all great to identify if your systems are patched, but with this Spectre/Meltdown being such a specific emergency patch (whether warranted or not), it would be great to have the ability for dbatools to offer a new command/feature allowing for a patch that is "ultra-mega-super critical" to be marked as such and cause dbatools to throw a huge, unmissable tantrum about the patch missing on any machines.

It could be argued that SQL 2016 SP1 would have a similar warning, because of the amazing extras unlocked by it, although that isn't a security alert. Maybe this is something to put into different categories, but certainly allow a more "verbal" warning system.

Of course, this may be something for a certain side project too.

niphlod:
haha, "certain"... guess we can put a prio on https://github.com/sqlcollaborative/dbatools/issues/2375 . I'll cook something up right away.

sql-williamd:
I was asked to word it that way 😄 

Yeah, that is similar to/in conjunction with #2375 for sure.

wsmelton:
Overall I don't see a need to have special commands that bring attention to the patching for the CPU bugs. From a security standpoint, these are no different or more critical than other MS bulletins put out for security fixes. It is just the amount of systems involved in the CPU bugs has brought more attention to it.

I'd say what we have in `Test-DbaSqlBuild` is sufficient.

sql-williamd:
Hi @wsmelton - I didn't mean for a special cpu bug based issue system. I was meaning for the current command to somehow provide a clearer warning about critical patches becoming available. As it is, Test-DbaSqlBuild informs if a server is behind patching. However, many companies don't necessarily see these patches as important. A clear sign that a patch is an emergency/critical fix requiring immediate action would help make this clear.

wsmelton:
To do that on a continuous basis would require that we manually flag them as such. There is no way based on a KB number or build number to know anything more than it is an SP, CU, or out of cycle patch (which on average now are security patches). Which the out of cycle patches are generally all categorized as critical by Windows Update, WSUS, and Microsoft. If we start manually flagging them, who makes the discussion that it is or is not?

[...]

PR #2988, opened 2018-01-04
Test-DbaSqlBuild, fixes #2375

### LINKED ISSUE #2375: [New Command] Test-DbaSqlBuild
### Problem to solve
Get-DbaSqlBuildReference provides build information and the closest match:
image

Discussion with the team at Pythian and we'd like to have a command that we can utilize to determine the path to get up to the latest build. 

So my local instance is SQL Server 2014 SP1 + some unknown patch since it does not match what is in our build reference file. The test command would have similar output but include some additional information:

```
SqlInstance: MANATARMS\SQL12
Current Build: 12.0.4237
Current SP: SP1
Current CU: 
Recommended SP: <latest SP number>
Recommended CU: <latest CU to SP, or lastest CU to RTM>
Recommended KB: < could just be comma separated list of KB #s >
```

niphlod:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality, see #2375 and #2985)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Purpose
 see #2375 and #2985

### Approach
Builds up on Get-DbaSqlBuildReference and @FriedrichWeinmann speedy [version] type to compare versions. Granted, doing the same thing MinimumBuild does was only a matter of knowing that `[version]` existed, but the added logic for MaxBehind makes checking compliance really swift.
For #2985 all it takes is a nice blogpost with the MinimumBuild for each version ... or at most a wrapper around this one. 

### Commands to test
See ALL examples. Then show it to your compliance auditor. Then go take care of his/her goosebumps ^_^.

### Screenshots
image

### DIFF
--- a/functions/Test-DbaSqlBuild.ps1
+++ b/functions/Test-DbaSqlBuild.ps1
@@ -27,6 +27,12 @@ function Test-DbaSqlBuild {
     .PARAMETER Update
         Looks online for the most up to date reference, replacing the local one.
 
+    .PARAMETER Quiet
+        Makes the function just return $true/$false. It's useful if you use Test-DbaSqlBuild in your own scripts, like
+            if (Test-DbaSqlBuild -Build "12.0.5540" -MaxBehind "0CU" -Quiet) {
+                Do-Something
+            }
+
     .PARAMETER EnableException
         By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
         This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
@@ -101,11 +107,14 @@ function Test-DbaSqlBuild {
         $SqlInstance,
 
         [Alias("Credential")]
-        [PsCredential]
+        [PSCredential]
         $SqlCredential,
 
         [switch]
         $Update,
+
+        [switch]
+        $Quiet,
 
         [switch]
         [Alias('Silent')]$EnableException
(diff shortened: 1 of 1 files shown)


potatoqualitee:
nice!

potatoqualitee:
mannnnnn tests have been great all day and now it has to fail for you :(

functions/Test-DbaSqlBuild.ps1
```diff
@@ -0,0 +1,246 @@
...
+            catch {
+                Stop-Function -Message "Error parsing MaxBehind" -ErrorRecord $_
+                return
+            }
+        }
+    }
+    process {
+        if (Test-FunctionInterrupt) { return }
+        try {
+            $IdxRef = Get-DbaSqlBuildReferenceIndex -Moduledirectory $moduledirectory -Update $Update -EnableException $EnableException
```
FriedrichWeinmann: This looks like something that should run in `begin`
The parameters aren't on the helper function defined in begin.
When calling one of our commands in try/catch, you shouldn't pass through `-EnableException $EnableException`, rather set it to `$true`. Also keep in mind that it is a switch parameter.

niphlod: blunt copy/paste mistake. Thanks for the catch

niphlod:
@FriedrichWeinmann : sure we want to introduce Quiet ? We have no Test- functions with Quiet, not even Test-DbaConnection ...

FriedrichWeinmann:
@niphlod it's an accepted standard in PowerShell, so I'm fine with it (and in favor of retrofitting other commands for it). Probably want to run it past @potatoqualitee first though, before declaring it policy ^^

Example why:
```powershell
if (Test-Something) {
    Do-Something
}
```
If the test returns an object, it will always be true, even if the test failed. Having a graceful way to handle this that is already accepted in the PS world sounds like a winner to me :sunglasses:

niphlod:
Sounds good and easy to engineer in. Will do all the needed mods in a bit.

potatoqualitee:
🔥 ! thanks to all who participated

PR #2992, opened 2018-01-05
Test-DbaSqlBuild, fix when no CU released after SP

niphlod:
followup of https://github.com/sqlcollaborative/dbatools/pull/2988

### DIFF
--- a/functions/Test-DbaSqlBuild.ps1
+++ b/functions/Test-DbaSqlBuild.ps1
@@ -217,7 +217,7 @@ function Test-DbaSqlBuild {
                 Stop-Function -Message "$($BuildVersion.Build) is not recognized as a correct version" -ErrorRecord $_ -Continue
             }
             if ($MinimumBuild) {
-                Write-Message -Level verbose -Message "Comparing $MinimumBuild to $inputbuild"
+                Write-Message -Level Debug -Message "Comparing $MinimumBuild to $inputbuild"
                 if ($inputbuild -ge $MinimumBuild) {
                     $compliant = $true
                 }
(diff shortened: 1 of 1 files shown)


functions/Test-DbaSqlBuild.ps1
```diff
@@ -254,19 +254,21 @@ function Test-DbaSqlBuild {
                             $targetSP = 0
                         }
                         $targetSPName = $AllSPs[$targetSP]
-                        Write-Message -Level verbose -Message "Target SP is $targetSPName - $targetSP on $($AllSPs.Length)"
+                        Write-Message -Level Debug -Message "Target SP is $targetSPName - $targetSP on $($AllSPs.Length)"
```
ClaudioESSilva: Needs to be `$($AllSPs.Length -1)`

functions/Test-DbaSqlBuild.ps1
```diff
@@ -254,19 +254,21 @@ function Test-DbaSqlBuild {
...
-                        if ($targetCU -lt 0) {
-                            $targetCU = 0
+                        if ($AllCUs.Length -gt 0) {
+                            #CU after the targeted build available
+                            $targetCU = $AllCUs.Length - $ParsedMaxBehind['CU'] - 1
+                            if ($targetCU -lt 0) {
+                                $targetCU = 0
+                            }
+                            $targetCUName = $AllCUs[$targetCU]
+                            Write-Message -Level Debug -Message "Target CU is $targetCUName - $targetCU on $($AllCUs.Length)"
```
ClaudioESSilva: Needs to be `$($AllCUs.Length -1)`

functions/Test-DbaSqlBuild.ps1
```diff
@@ -254,19 +254,21 @@ function Test-DbaSqlBuild {
...
-                        $targetCU = $AllCUs.Length - $ParsedMaxBehind['CU'] - 1
-                        if ($targetCU -lt 0) {
-                            $targetCU = 0
+                        if ($AllCUs.Length -gt 0) {
+                            #CU after the targeted build available
+                            $targetCU = $AllCUs.Length - $ParsedMaxBehind['CU'] - 1
+                            if ($targetCU -lt 0) {
+                                $targetCU = 0
+                            }
+                            $targetCUName = $AllCUs[$targetCU]
```
ClaudioESSilva: `$targetCUName` must be clear before each execution.

functions/Test-DbaSqlBuild.ps1
niphlod: hum.... it's a debug message in "computer language", using 0-indexed numbers... in the end we want $AllSPs[$targetSP], so I don't see the point in changing the message

functions/Test-DbaSqlBuild.ps1
niphlod: same as previous

functions/Test-DbaSqlBuild.ps1
niphlod: right, better safeguarding also targetspname

functions/Test-DbaSqlBuild.ps1
ClaudioESSilva: The "problem" here is that RTM is being seen as "SP". That's why the output is 5.
For me the message `Target SP is SP4 - 4 on 5` indicates that I have an SP to apply and it's not true.
If we can have the message written in a way that doesn't offer doubts why not? 😄

functions/Test-DbaSqlBuild.ps1
ClaudioESSilva: Forget this comment. Here this is right!

functions/Test-DbaSqlBuild.ps1
niphlod: you need to look at the result, not at debug messages! `Target SP is SP4 - 4th element on 5 long array` would please you ?

ClaudioESSilva: Not really. For that just leave it as it is right now.

niphlod: we can also do `Target SP is SP4` :-P

potatoqualitee:
YESSSSS !

PR #3003, opened 2018-01-09
New Function Find-DbaDisabledIndex

js0505:
## Type of Change

### Purpose
Find Disabled Indexes and list them out.

### Approach

Same approach as Find-DbaDuplicateIndex and Find-DbaUnusedIndex

### Commands to test
Find-DbaDisabledIndex

### Screenshots
find-dbadisabledindex

### DIFF
--- a/functions/Find-DbaDisabledIndex.ps1
+++ b/functions/Find-DbaDisabledIndex.ps1
@@ -1,0 +1,180 @@
+﻿function Find-DbaDisabledIndex {
+    <#
+        .SYNOPSIS
+            Find Disabled indexes
+
+        .DESCRIPTION
+            This command will help you to find disabled indexes on a database or a list of databases
+
+        .PARAMETER SqlInstance
+            The SQL Server you want to check for disabled indexes.
+
+        .PARAMETER SqlCredential
+            Allows you to login to servers using SQL Logins instead of Windows Authentication (AKA Integrated or Trusted). To use:
+
+            $cred = Get-Credential, then pass $cred object to the -SqlCredential parameter.
+
+            Windows Authentication will be used if SqlCredential is not specified. SQL Server does not accept Windows credentials being passed as credentials.
+
+            To connect as a different Windows user, run PowerShell as that user.
+
+        .PARAMETER Database
+            The database(s) to process. Options for this list are auto-populated from the server. If unspecified, all databases will be processed.
+
+        .PARAMETER ExcludeDatabase
+            Specifies the database(s) to exclude from processing. Options for this list are auto-populated from the server.
+
+        .PARAMETER NoClobber
+            If this switch is enabled, the output file will not be overwritten.
+
+        .PARAMETER Append
+            If this switch is enabled, content will be appended to the output file.
+
+            .PARAMETER WhatIf
+            If this switch is enabled, no actions are performed but informational messages will be displayed that explain what would happen if the command were to run.
+
+        .PARAMETER Confirm
+            If this switch is enabled, you will be prompted for confirmation before executing any operations that change state.
+
+        .PARAMETER EnableException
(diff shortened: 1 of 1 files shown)


functions/Find-DbaDisabledIndex.ps1
```diff
@@ -0,0 +1,180 @@
...
+        $diffDays = (New-TimeSpan -Start $endDate -End (Get-Date)).Days
+
+        <#
+            Validate if server version is:
+                - sql 2012 and if have SP3 CU3 (Build 6537) or higher
+                - sql 2014 and if have SP2 (Build 5000) or higher
+            If the major version is the same but the build is lower, throws the message
+        #>
+        if ($pipedatabase.Length -gt 0) {
+            $database = $pipedatabase.name
```
wsmelton: This variable would be best called `$databases` to differentiate it from the parameter on the command.

wsmelton: Don't know that we have in any particular guide but I try to not overwrite the parameter for the command, so in this case `$Database` should only ever exist as what is passed in.

functions/Find-DbaDisabledIndex.ps1
```diff
@@ -0,0 +1,180 @@
...
+        }
+
+        if ($database.Count -gt 0) {
+            foreach ($db in $database) {
+                
+                if ($ExcludeDatabase -contains $db -or $null -eq $server.Databases[$db]) {
+                    continue
+                }
+                if ($server.Databases[$db].IsAccessible -eq $false) {
+                    Write-Message -Level Warning -Message "Database [$db] is not accessible."
```
wsmelton: This would seem redundant since we already removed databases that were not accessible. We are not refreshing the server object so even if a database was taken offline before it hits this line we would not know about it.

functions/Find-DbaDisabledIndex.ps1
```diff
@@ -0,0 +1,180 @@
...
+                    continue
+                }
+                
+                try {
+                    Write-Message -Level Output -Message "Getting indexes from database '$db'."
+
```
wsmelton: Need to add a debug so we output the query before executing it.

js0505:
Thanks for the comments, I will update.

potatoqualitee:
Thanks much, @js0505 and @wsmelton !

functions/Find-DbaDisabledIndex.ps1
```diff
@@ -0,0 +1,162 @@
...
+        FROM sys.tables t
+        JOIN sys.schemas s
+            ON t.schema_id = s.schema_id
+        JOIN sys.indexes i
+            ON i.object_id = t.object_id
+        WHERE i.is_disabled = 1"
+    }
+    process {
+        Write-Message -Level Output -Message "Attempting to connect to Sql Server."
+        $server = Connect-SqlInstance -SqlInstance $SqlInstance -SqlCredential $SqlCredential -MinimumVersion 9
```
wsmelton: Still need to address a `foreach` loop to support multiple instances being passed in, or piped.

Also wrap the connect attempt in try/catch. This is the pattern we use for the other commands:

```powershell
            try {
                Write-Message -Level Verbose -Message "Connecting to $instance."
                $server = Connect-SqlInstance -SqlInstance $instance -SqlCredential $sqlcredential
            }
            catch {
            }
```

potatoqualitee:
will merge once @wsmelton  approves

js0505:
I do love his feedback, makes me want to look at other commands to get things cleaned up. So I am going to try to start working on some of those when I have the chance. Reminder, once this function gets applied and I write the dbachecks function, the version of dbatools will be required. You are all the best! Love contributing to all of this.

potatoqualitee:
Thanks to you both! :us:

potatoqualitee:
👍

PR #3012, opened 2018-01-10
Created a Parameterset for destination parameter.

### LINKED ISSUE #2939: Copy-DbaLogin requires Destination even when OutFile specified

- [x] Verified running the latest release of dbatools?
- [x] Can duplicate in new/clean PowerShell session (clean = `powershell -NoProfile`)?

## Steps to Reproduce

Calling the Copy-DbaLogin command and specifying a value for the OutFile Parameter will still prompt for a Destination.
eg
Copy-DbaLogin -Source InstanceName -Login 'Login' -OutFile 'C:\temp\login.sql'

The help states:
 -OutFile <String>
        Calls Export-SqlLogin and exports all logins to a T-SQL formatted file. This does not perform a copy, so no destination is required.

However the parameter Destination is declared as Mandetory so must be specified even when not required.
Setting the Destination same as Source will produce a correct result so this may just need better documentation

tengen10:
## Type of Change

 - [x ] Bug fix (non-breaking change, fixes #2939 )
 

### Purpose
 
When sending the output to a file the function no longer prompts for a destination.

### Approach

By creating a ParameterSet for the destination it is now mutually exclusive of the OutFile parameter which was already configured as a ParameterSet. 

### Commands to test

```
Get-Command Copy-DbaLogin -syntax
Copy-DbaLogin -Source myserver -OutFile c:\temp\myserver_logins.sql

```
### Learning

[](https://blogs.msdn.microsoft.com/powershell/2008/12/23/powershell-v2-parametersets/)
[](http://blog.simonw.se/powershell-functions-and-parameter-sets/)

### DIFF
--- a/functions/Copy-DbaLogin.ps1
+++ b/functions/Copy-DbaLogin.ps1
@@ -131,7 +131,7 @@ function Copy-DbaLogin {
         [DbaInstanceParameter]$Source,
         [PSCredential]
         $SourceSqlCredential,
-        [parameter(Mandatory = $true)]
+        [parameter(ParameterSetName = "Destination", Mandatory = $true)]
         [DbaInstanceParameter]$Destination,
         [PSCredential]
         $DestinationSqlCredential,


potatoqualitee:
plz approve if you get a moment, @wsmelton ! Thank you, @tengen10!

wsmelton:
Add some discussion in Slack but will add here as well for documentation.

I would have to disagree with this change because you are changing the purpose/use of the Copy command. The purpose of the Copy commands is for migration between a source and destination. The change in this PR is changing that use case for one copy command. You are basically duplicating functionality between `Copy-DbaLogin` and `Export-DbaLogin`.

If this PR stands there is then no difference between this:

```powershell
Copy-DbaLogin -Source InstanceName -Login 'Login' -OutFile 'C:\login.sql'
```
and this:
```powershell
Export-DbaLogin -SqlInstance InstanceName -Login 'Login' -FilePath 'C:\login.sql'
```

This is basically making the copy login command conform to another purpose and match what the export login command should be used for.

wsmelton:
I do see the copy login command for the OutFile parameter the help states destination is not required:

> Calls Export-SqlLogin and exports all logins to a T-SQL formatted file. This does not perform a copy, so no destination is required.

So I guess it can go back to what purpose was there in adding `-OutFile` to the copy login command in the first place?

potatoqualitee:
thank you!

PR #3014, opened 2018-01-10
0.9.155

potatoqualitee:


potatoqualitee:
big ol bunch of new XE and Profile commands. I need to revisit the docs/examples but will do that tomorrow. @gbargsley I finally pushed those new commands to master 👍  I'm not trying, but it's cool to see we now have 357 commands!

PR #3019, opened 2018-01-11
Add parent AD Group Name to Find-ADLoginInGroup

LowlyDBA:
## Type of Change

### Purpose
By default, return the "parent" AD group for users. I.e. the top level AD group login by which the user is granted access.

 When there are many sub-groups, the info returned requires manual work to ferret out how a user or sub-group inherited access - this removes the need for extra querying/investigation to track down that info.

### Approach
Adds a parameter to the `Get-AllLogins` function to optionally pass the parent AD group. This is used when the function is called on recursive groups. If not passed, assume the working AD group is the parent and use that value instead. 

### Commands to test
`Find-DbaLoginInGroup`

### Screenshots
Proposed output:
image


potatoqualitee:
Thanks so much, @LowlyDBA ! I'm going to ask @niphlod to take a look (also, niph, the tests didn't run, is that because this command is excluded? Merci!)

niphlod:
@potatoqualitee : I can only manage magic(tm) INSIDE the build, I cannot make a build not start.... guess it's a hiccup on appveyor's side

potatoqualitee:
tried to remove Export-DbaSpConfigure but it left a lil cr as expected. will merge if tests pass 👍

LowlyDBA:
@potatoqualitee This may not be correctly returning the top-level AD group, but instead just one level up when there are > 2 nested groups. Please hold off on any merging while I toy with it some more.

potatoqualitee:
@niphlod please approve and I'll merge then fix the tabs. ty!

@LowlyDBA thanks so much!

potatoqualitee:
Thanks to you both! I'll clean up the tabs

PR #3042, opened 2018-01-13
added all columns to watcha and read

potatoqualitee:


potatoqualitee:
one pass. appveyor is blowing today.

PR #3061, opened 2018-01-15
Find dba disabled index

js0505:
## Type of Change

 Cleaned up the foreach DB wrapper. It was piping all databases even if listed one. Cleaned that up and have tested multiple scenarios. Sorry about that.


potatoqualitee:
Thanks baw!

PR #3074, opened 2018-01-16
DbaInstanceParameter - Added Support for Azure Instances

FriedrichWeinmann:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Changes
 - Fixed: SMO Server Instances of Azure SQL Instances can now be understood and passed through by the `[DbaInstance]` parameter class


potatoqualitee:
unfortunately, both tests failed :(

wsmelton:
Curious how are they not already?
image

The above was with 0.9.152 version.

I don't see much difference with this PR version:
image

FriedrichWeinmann:
Dang it, I'll look into it.
@wsmelton : Strings work just fine. Try running `Connect-DbaInstance` and using the resulting server object to connect.

wsmelton:
Works for me:

image

image

wsmelton:
With Azure SQL you won't be able to access much of anything at the server-level with SMO (e.g. `$server.Logins`, etc) it is just not supported.

FriedrichWeinmann:
And now that you have the `$server` object, try this:
`[DbaInstance]$server`

wsmelton:
Ah, ok see your point. However, you will never be able to get instance level information from that `$server` object with Azure SQL...it is just not supported. 

The properties like InstanceName, ComptuerNamePhysicalNetBIOS or NetName will not have any value. Actually, not much of anything is populated on that top level server object in SMO when you hit an Azure SQL instance.

FriedrichWeinmann:
Oh, that's totally fine with me. My priority is to have it accept the connection object, so pure SQL connections can be reused, rather than always opening a new one.

The parameter class only requires me to get a computer name (a dns resolveable name, which you need to connect anyway), and that's good enough for me. `Connect-SqlInstance` will ignore it all anyway if it's a live connection.

potatoqualitee:
thanks fred! 💯

PR #3084, opened 2018-01-17
Clean up of MaxMemory commands (Get, Set, and Test)

wsmelton:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
A bit of a bug found in Set-DbaMaxMemory in how connections were managed. However, in the end, this required touching all three commands to clean up things.

Conversation in Slack but overall the commands were not following the standard we use for making connections to a given instance, and then use that connection in any sub call we have. This PR address that among other things.

The test for `Set-DbaMaxMemory` I also added a revert with BeforeAll and AfterAll so if you run this against your local instance it will put the values back the way they were found.

### DIFF
--- a/functions/Get-DbaMaxMemory.ps1
+++ b/functions/Get-DbaMaxMemory.ps1
@@ -60,17 +60,16 @@ function Get-DbaMaxMemory {
 				Stop-Function -Message "Failure" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
 			}
 
-			$totalmemory = $server.PhysicalMemory
+			$totalMemory = $server.PhysicalMemory
 
 			# Some servers under-report by 1MB.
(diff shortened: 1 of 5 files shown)


wsmelton:
We do need to likely expand the test on the set command...just don't have time to put in on it today....best to get fix out on issue with the command.

potatoqualitee:
awesome, thanks so much shawn!

potatoqualitee:
Tests pass except for tabs(?)

wsmelton:
Unsure what converted it...

wsmelton:
@potatoqualitee @niphlod so yall will have to explain what that test is looking for because from VS Code the file is set to spaces. So why does your test show it is not?

wsmelton:
Finally

potatoqualitee:
oh la la!

potatoqualitee:
Thank youuuuuuuuu 🙇

PR #3086, opened 2018-01-17
Get-DbaUpTime - added tests, cleaned up a couple of things in function

jpomfret:
## Type of Change

 - [X] Pester test is included
 - [ ] Nunit test is included
 - [X] Documentation
 - [ ] Build system
 

### Purpose
 
- Added tests for the function.
- Changed parameter from 'WindowsCredential' to 'Credential' in help and function to match standard params
- forced use of dbadatetime for variables, previously two variables that I think should be the same type were coming back one as datetime and one as dbadatetime.
- removed $SqlOnly variable as it was never used.

### Commands to test

image

image

### DIFF
--- a/functions/Get-DbaUptime.ps1
+++ b/functions/Get-DbaUptime.ps1
@@ -20,6 +20,9 @@ function Get-DbaUptime {
 
             To connect to SQL Server as a different Windows user, run PowerShell as that user.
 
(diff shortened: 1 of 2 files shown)


potatoqualitee:
yessss thanks jess!

PR #3103, opened 2018-01-19
Get-DbaSqlInstanceProperty tests

jpomfret:
## Type of Change

 - [X] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Added tests for Get-DbaSqlInstanceProperty, included unit and integration tests.

### Screenshots

image

### DIFF
--- a/tests/Get-DbaSqlInstanceProperty.Tests.ps1
+++ b/tests/Get-DbaSqlInstanceProperty.Tests.ps1
@@ -37,7 +37,7 @@ Describe "$CommandName Integration Tests" -Tags "IntegrationTests" {
     }
     Context "Command can handle multiple instances" {
         It "Should have results for 2 instances" {
-            $(Get-DbaSqlInstanceProperty -SqlInstance $script:instance1, $script:instance2 | Select-Object -unique computername).count | Should Be 2
+            $(Get-DbaSqlInstanceProperty -SqlInstance $script:instance1, $script:instance2 | Select-Object -unique SqlInstance).count | Should Be 2
         }
     }
     Context "Command stops when can't connect" {
--- a/tests/Get-DbaSqlInstanceProperty.Tests.ps1
+++ b/tests/Get-DbaSqlInstanceProperty.Tests.ps1
@@ -1,0 +1,38 @@
+$CommandName = $MyInvocation.MyCommand.Name.Replace(".Tests.ps1", "")
+Write-Host -Object "Running $PSCommandPath" -ForegroundColor Cyan
+. "$PSScriptRoot\constants.ps1"
+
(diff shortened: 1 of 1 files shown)


tests/Get-DbaSqlInstanceProperty.Tests.ps1
```diff
@@ -0,0 +1,44 @@
...
+
+Describe "$CommandName Integration Tests" -Tags "IntegrationTests" {
+    Context "Command actually works" {
+        $results = Get-DbaSqlInstanceProperty -SqlInstance $script:instance2
+        It "Should have correct properties" {
+            $ExpectedProps = 'ComputerName,InstanceName,PropertyType,SqlInstance'.Split(',')
+            (($results | Get-Member -MemberType NoteProperty).name | Sort-Object) | Should Be ($ExpectedProps | Sort-Object)
+        }
+        It "Should return that instance2 is running developer edition" {
+            ($results | Where-Object {$_.name -eq 'Edition'}).Value | Should Be "Developer Edition (64-bit)"
```
wsmelton: It would be ideal to make more generic test...values that would/could be the same for both local environments and AppVeyory.

tests/Get-DbaSqlInstanceProperty.Tests.ps1
```diff
@@ -0,0 +1,44 @@
...
+        }
+        It "Should get the correct DefaultFile location" {
+            $defaultFiles = Get-DbaDefaultPath -SqlInstance $script:instance2
+            ($results | Where-Object {$_.name -eq 'DefaultFile'}).Value | Should BeLike "$($defaultFiles.Data)*"
+        }
+    }
+    Context "Command warns when can't connect to instance" {
+        $null = Get-DbaSqlInstanceProperty -SqlInstance MadeUpServer -WarningVariable warn -WarningAction SilentlyContinue
+        It "Should warn cannot connect to MadeUpServer" {
+            $warn | Should Match "Can't connect to MadeUpServer"
```
wsmelton: A few different ways you can actually do this one. You would need to check other test but I think we use the `-EnableException` and then just check that it throws.

tests/Get-DbaSqlInstanceProperty.Tests.ps1
jpomfret: Thanks for the feedback, would you suggest picking a different property, maybe IsSingleUser which would more than likely be False on all systems, or  picking ResourceVersionString which returns the build number and running that through `Get-DbaSqlBuildReference` to check it's valid?

potatoqualitee:
thanks much!

PR #3119, opened 2018-01-21
Improved information display Test-DbaLogShippingStatus

sanderstad:
Imroved the way the information is presented when not using a monitor.


sanderstad:
Before when no monitor would be selected the information would not be accurate because the function assumes you're using a monitor. Therefore some information would be show incorrectly like below:
image

The new version checks if it needs to return results for a primary or secondary instance and will display "N/A" if it's not able to return information like below:
image

potatoqualitee:
👍 💯

PR #3137, opened 2018-01-22
Fixes PSUseBOMForUnicodeEncodedFile ScriptAnalyzer rule

maddhatter:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Fix files that violate `ScriptAnalyzer`'s  `PSUseBOMForUnicodeEncodedFile` rule.

### Approach

Re-save files in VS Code as "UTF-8 with BOM"

### DIFF
--- a/functions/Add-DbaComputerCertificate.ps1
+++ b/functions/Add-DbaComputerCertificate.ps1
@@ -1,4 +1,4 @@
-#ValidationTags#Messaging,FlowControl,Pipeline,CodeStyle#
+﻿#ValidationTags#Messaging,FlowControl,Pipeline,CodeStyle#
 function Add-DbaComputerCertificate {
     <#
         .SYNOPSIS
--- a/functions/Find-DbaTrigger.ps1
+++ b/functions/Find-DbaTrigger.ps1
@@ -1,4 +1,4 @@
-function Find-DbaTrigger {
+﻿function Find-DbaTrigger {
     <#
 .SYNOPSIS
 Returns all triggers that contain a specific case-insensitive string or regex pattern.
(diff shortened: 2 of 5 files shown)


niphlod:
good and bad, frankly having half of files with the BOM and half without isn't smarty........ like the check. 
If for some reason within the code or the docs there's something out of "normal chars" (e.g. happened with @ClaudioESSilva signature, for some accented letters), the check is triggered.
The solution by PSSCriptAnalyzer is including the BOM, mine would be to hunt down the "special character" and replace it with something not outside the range of normal chars.

wsmelton:
@maddhatter This fix is irrelevant because we do not check for this rule with PSSA.

However, it is also a bug in that check for PSSA and it is documented in their repository. There are files that this check won't warn even though the file is only saved as UTF-8 and not UTF-8 w/ BOM. So it is an unstable check in my opinion.

maddhatter:
I don't have a strong opinion either way, I just was going to submit a couple ScriptAnalyzer PRs from the Bill of Health post while I had some downtime during meetings, etc (multitasking, ftw).

I ran `.\tests\manual.pester.ps1 -Path .\functions\Add-DbaComputerCertificate.ps1 -ScriptAnalyzer` since it was the very first file in the Bill of Health list, and it returned that `PSUseBOMForUnicodeEncodedFile` rule for that file.

wsmelton:
Slack discussion with @niphlod about some of this and from the standpoint of PSSA rules we do not have any that are showing errors, which would be the ones we really needed to fix.

The only thing being thrown when the manual test is run are warnings and informational rules. So I'm going to go ahead and close this PR out.

PR #3147, opened 2018-01-24
Remove Distribution from list of excluded DBs

bajmccormack:
Line 122 lists system databases which are to be excluded from process. I have removed distribution as it is perfectly valid to set this to any recovery model, replication is not affected by this. Also, distribution is a valid name for a user database, it is not reserved for replication.

I believe this could also be done for SSIS but I wasn't sure the reason this was listed so have left it for now.

## Type of Change

 - [ x ] Bug fix (non-breaking change, fixes #<No bug listed>)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Remove Distribution from list of excluded DBs

### Approach

Distribution is no longer added to $systemdbs

### Commands to test

Set-DbaDbRecoveryModel -SqlInstance SQL2016 -RecoveryModel Simple -Database Distribution

### DIFF
--- a/functions/Set-DbaDbRecoveryModel.ps1
+++ b/functions/Set-DbaDbRecoveryModel.ps1
@@ -119,7 +119,7 @@ function Set-DbaDbRecoveryModel {
             }
 
             # We need to be able to change the RecoveryModel for model database
-            $systemdbs = @("master", "msdb", "tempdb", "SSIS", "distribution")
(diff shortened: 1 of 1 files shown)


potatoqualitee:
I love that, thank you all. Full Recovery on all dbs except tempdb is actually a requirement on many systems, so that final commit is a 💯 solution

PR #3161, opened 2018-01-26
Changes to Remove-DbaAgentJob and Remove-DbaAgentJobStep and added tests for Remove-DbaAgentJob

jpomfret:
## Type of Change

 - [X] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Getting Remove-DbaAgentJob and Remove-DbaAgentJobStep up to standards. Added tests for Remove-DbaAgentJob.

### Approach

tidying up formatting, fixed messaging, tidying up CBH, added unit and integration tests.

### Issues
- Added tests for -KeepHistory but its failing, I'm not convinced that scenario works, if anyone agrees I'll remove that from the cmdlet and the tests. I think history is automatically removed when a job is.
- Also any more suggestions for getting all boh items met.

### DIFF
--- a/functions/Remove-DbaAgentJob.ps1
+++ b/functions/Remove-DbaAgentJob.ps1
@@ -91,10 +91,10 @@ function Remove-DbaAgentJob {
         [switch]$KeepHistory,
         [Parameter(Mandatory = $false)]
         [switch]$KeepUnusedSchedule,
-        [DbaMode]
-        $Mode = (Get-DbaConfigValue -Name 'message.mode.default' -Fallback "Strict"),
+        [DbaMode]$Mode = (Get-DbaConfigValue -Name 'message.mode.default' -Fallback "Strict"),
         [Parameter(Mandatory = $false)]
-        [switch][Alias('Silent')]$EnableException
+        [Alias('Silent')]
+        [switch]$EnableException
     )
     process {
         foreach ($instance in $SqlInstance) {
@@ -125,7 +125,7 @@ function Remove-DbaAgentJob {
                         try {
                             $currentJob = $Server.JobServer.Jobs[$j]
                             # Delete the history
-                            if (-not $KeepHistory) {
+                            if (Test-Bound -ParameterName KeepHistory -Not) {
                                 Write-Message  -Level SomewhatVerbose -Message "Purging job history"
                                 $currentJob.PurgeHistory()
                             }
--- a/functions/Remove-DbaAgentJob.ps1
+++ b/functions/Remove-DbaAgentJob.ps1
@@ -1,103 +1,105 @@
 #ValidationTags#Messaging,FlowControl,Pipeline,CodeStyle#
 function Remove-DbaAgentJob {
     <#
-.SYNOPSIS
-Remove-DbaAgentJob removes a job.
+        .SYNOPSIS
+            Remove-DbaAgentJob removes a job.
 
-.DESCRIPTION
-Remove-DbaAgentJob removes a a job in the SQL Server Agent.
+        .DESCRIPTION
+            Remove-DbaAgentJob removes a a job in the SQL Server Agent.
 
-.PARAMETER SqlInstance
-SQL Server instance. You must have sysadmin access and server version must be SQL Server version 2000 or greater.
+        .PARAMETER SqlInstance
+            SQL Server name or SMO object representing the SQL Server to connect to. This can be a collection and receive pipeline input to allow the function to be executed against multiple SQL Server instances.
 
-.PARAMETER SqlCredential
-Allows you to login to servers using SQL Logins as opposed to Windows Auth/Integrated/Trusted. To use:
-$scred = Get-Credential, then pass $scred object to the -SqlCredential parameter.
-To connect as a different Windows user, run PowerShell as that user.
+        .PARAMETER SqlCredential
+            SqlCredential object to connect as. If not specified, current Windows login will be used.
 
-.PARAMETER Job
-The name of the job. Can be null if the the job id is being used.
+        .PARAMETER Job
+            The name of the job. Can be null if the the job id is being used.
 
-.PARAMETER KeepHistory
-Specifies to keep the history for the job. By default is history is deleted.
+        .PARAMETER KeepHistory
+            Specifies to keep the history for the job. By default history is deleted.
 
-.PARAMETER KeepUnusedSchedule
-Specifies to keep the schedules attached to this job if they are not attached to any other job.
-By default the unused schedule is deleted.
+        .PARAMETER KeepUnusedSchedule
+            Specifies to keep the schedules attached to this job if they are not attached to any other job.
+            By default the unused schedule is deleted.
 
-.PARAMETER WhatIf
-Shows what would happen if the command were to run. No actions are actually performed.
+        .PARAMETER Mode
+            Default: Strict
+            How strict does the command take lesser issues?
+            Strict: Interrupt if the job specified doesn't exist.
+            Lazy:   Silently skip over jobs that don't exist.
+        
+        .PARAMETER WhatIf
+            Shows what would happen if the command were to run. No actions are actually performed.
 
-.PARAMETER Confirm
-Prompts you for confirmation before executing any changing operations within the command.
+        .PARAMETER Confirm
+            Prompts you for confirmation before executing any changing operations within the command.
 
-.PARAMETER EnableException
-        By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
-        This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
-        Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
+        .PARAMETER EnableException
+            By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
+            This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
+            Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
 
-.NOTES
-Author: Sander Stad (@sqlstad, sqlstad.nl)
-Tags: Agent, Job
+        .NOTES
+            Author: Sander Stad (@sqlstad, sqlstad.nl)
+            Tags: Agent, Job
 
-Website: https://dbatools.io
-Copyright: (C) Chrissy LeMaire, clemaire@gmail.com
-License: GNU GPL v3 https://opensource.org/licenses/GPL-3.0
+            Website: https://dbatools.io
+            Copyright: (C) Chrissy LeMaire, clemaire@gmail.com
+            License: GNU GPL v3 https://opensource.org/licenses/GPL-3.0
 
-.LINK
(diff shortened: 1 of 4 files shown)


jpomfret:
Also the cmdlets don't return any object (except verbose) that they did anything. Talked about making a -mode report but should they return something as standard?

functions/Remove-DbaAgentJob.ps1
```diff
@@ -1,103 +1,105 @@
...
-
         [Parameter(Mandatory = $false)]
         [switch]$KeepUnusedSchedule,
-
+        [DbaMode]
+        $Mode = (Get-DbaConfigValue -Name 'message.mode.default' -Fallback "Strict"),
         [Parameter(Mandatory = $false)]
         [switch][Alias('Silent')]$EnableException
     )
-
```
wsmelton: Can you go ahead and update the format of EnableException parameter to match

```        
[Alias('Silent')]
[switch]$EnableException```

File: functions/Remove-DbaAgentJob.ps1 [hunk 2]
```diff
@@ -1,103 +1,105 @@
...
         [object[]]$Job,
-
         [Parameter(Mandatory = $false)]
         [switch]$KeepHistory,
-
         [Parameter(Mandatory = $false)]
         [switch]$KeepUnusedSchedule,
-
+        [DbaMode]
+        $Mode = (Get-DbaConfigValue -Name 'message.mode.default' -Fallback "Strict"),
```
wsmelton: This can be on one line as well to match formatting.

functions/Remove-DbaAgentJobStep.ps1
```diff
@@ -81,61 +80,66 @@ Remove the job step from the job on multiple servers using pipeline
         [Parameter(Mandatory = $true)]
         [ValidateNotNullOrEmpty()]
         [string]$StepName,
-        [Parameter(Mandatory = $false)]
-        [switch][Alias('Silent')]$EnableException
+        [DbaMode]
+        $Mode = (Get-DbaConfigValue -Name 'message.mode.default' -Fallback "Strict"),
```
wsmelton: This can be one line.

functions/Remove-DbaAgentJob.ps1
```diff
@@ -106,53 +108,48 @@ Removes the job from multiple servers using pipe line
...
-                    }
-
-                    # Execute
-                    if ($PSCmdlet.ShouldProcess($instance, "Removing the job on $instance")) {
+                    if ($PSCmdlet.ShouldProcess($instance, "Removing the job on $currentJob")) {
                         try {
-                            Write-Message -Message "Removing the job" -Level Verbose
+                            $currentJob = $Server.JobServer.Jobs[$j]
+                            # Delete the history
+                            if (-not $KeepHistory) {
```
wsmelton: I would update this to use `Test-Bound`, an internal function that will handle testing whether the parameter was passed or not in a more stable manner.

So if you want to test for "false" meaning it was not passed in you use `Test-Bound 'KeepHistory -Not' this will return `$true` if the KeepHistory was not used or provided.

tests/Remove-DbaAgentJob.Tests.ps1
```diff
@@ -0,0 +1,74 @@
...
+            if (Get-DbaAgentSchedule -SqlInstance $script:instance2 -Schedule dbatoolsci_weekly) { Remove-DbaAgentSchedule -SqlInstance $script:instance2 -Schedule dbatoolsci_weekly }
+        }
+        $null = Remove-DbaAgentJob -SqlInstance $script:instance2 -Job dbatoolsci_testjob_schedule -KeepUnusedSchedule
+        It "Should have deleted job: dbatoolsci_testjob_schedule" {
+            (Get-DbaAgentJob -SqlInstance $script:instance2 -Job dbatoolsci_testjob_schedule) | Should BeNullOrEmpty
+        }
+        It "Should not have deleted schedule: dbatoolsci_daily" {
+            (Get-DbaAgentSchedule -SqlInstance $script:instance2 -Schedule dbatoolsci_weekly) | Should Not BeNullOrEmpty
+        }
+        Context "Command removes job but not history" {
```
wsmelton: This needs to be a new context script block. As-is this is within the context above "Command removes job but not schedule". Just missing a closing curly brace.

wsmelton: Not sure if that is why the test is failing but worth updating to just make sure.

jpomfret:
Thanks for the comments @wsmelton, I believe I have addressed them all apart from the failing test for not purging history. I think the default for job.Drop() is to purge history as well.

wsmelton:
Well I tested it last night and believe the history stayed but let me double check. I just copied what you had in your test and ran it manually.

jpomfret:
Let me know what you find, I just retested on dbatools 0.9.176 and 0.9.136 and get the same behaviour where -KeepHistory does not work.

wsmelton:
Ok, no it clears history. However checking SSMS it will do the same thing, there is no option to keep job history when you use the UI.

You can via T-SQL so we may have to add that instead of the `Drop()` method. The `sp_delete_job` has `@delete_history` parameter, which defaults to 1 (one) to clear history.

jpomfret:
@wsmelton the delete_history bit works perfect on sp_delete_job, however delete_unused_schedule doesn't seem to work. I believe setting it to 1 should keep the unused schedule intact, however in my testing on both SQL Server 2008 and 2016 the schedules are always deleted. Can you see if it works for you? Maybe I haven't had enough coffee yet (lets hope).

jpomfret:
Seems to work from Powershell just fine actually, I need to rewrite my history test since the job is now gone `Get-DbaAgentJobHistory -SqlInstance server -Job dbatoolsci_testjob_history` doesn't find any history even though it is there.

functions/Remove-DbaAgentJob.ps1
```diff
@@ -1,103 +1,105 @@
...
         [ValidateNotNullOrEmpty()]
         [object[]]$Job,
-
         [Parameter(Mandatory = $false)]
         [switch]$KeepHistory,
-
         [Parameter(Mandatory = $false)]
         [switch]$KeepUnusedSchedule,
-
+        [DbaMode]$Mode = (Get-DbaConfigValue -Name 'message.mode.default' -Fallback "Strict"),
```
wsmelton: Curious to know where this is coming from?

Can't recall other commands having this added.

functions/Remove-DbaAgentJob.ps1
```diff
@@ -106,53 +108,44 @@ Removes the job from multiple servers using pipe line
...
-
-                    # Delete the history
-                    if (-not $KeepHistory) {
-                        Write-Message -Message "Purging job history" -Level Verbose
-                        $currentjob.PurgeHistory()
-                    }
-
-                    # Execute
-                    if ($PSCmdlet.ShouldProcess($instance, "Removing the job on $instance")) {
+                    if ($PSCmdlet.ShouldProcess($instance, "Removing the job on $currentJob")) {
```
wsmelton: Believe this should say `Removing the $currentJob job on $instance` ?

functions/Remove-DbaAgentJobStep.ps1
```diff
@@ -81,61 +80,65 @@ Remove the job step from the job on multiple servers using pipeline
...
-                            Stop-Function -Message "Something went wrong creating the job step" -Target $JobStep -Continue -ErrorRecord $_
-                        }
-
                         # Execute
                         if ($PSCmdlet.ShouldProcess($instance, "Removing the job step $StepName for job $j")) {
                             try {
-                                Write-Message -Message "Removing the job step $StepName for job $j" -Level Verbose
-
+                                $JobStep = $Server.JobServer.Jobs[$j].JobSteps[$StepName]
+                                Write-Message -Level SomewhatVerbose -Message "Removing the job step $StepName for job $j."
```
wsmelton: Should we go ahead and have this one call `Remove-DbaAgentJobStep` ?

tests/Remove-DbaAgentJob.Tests.ps1
```diff
@@ -0,0 +1,77 @@
...
+            $null = New-DbaAgentJobStep -SqlInstance $script:instance2 -Job dbatoolsci_testjob_history -StepId 1 -StepName dbatoolsci_step1 -Subsystem TransactSql -Command 'select 1'
+            $null = Start-DbaAgentJob -SqlInstance $script:instance2 -Job dbatoolsci_testjob_history
+        }
+    
+        $null = Remove-DbaAgentJob -SqlInstance $script:instance2 -Job dbatoolsci_testjob_history -KeepHistory
+        It "Should have deleted job: dbatoolsci_testjob_history" {
+            (Get-DbaAgentJob -SqlInstance $script:instance2 -Job dbatoolsci_testjob_history) | Should BeNullOrEmpty
+        }
+        It "Should not have deleted history: dbatoolsci_testjob_history" {
+            (Invoke-SqlCmd2 -ServerInstance $script:instance2 -Database msdb -Query "select 1 from sysjobhistory where job_id = '$($job.jobid)'") | Should Not BeNullOrEmpty
```
wsmelton: I don't know other folk's thinking on this but I try to stay clear of using this command in test, because it is one that can change and end up causing the test to break.

I would recommend just using the `$server.Databases['msdb'].Query()` method or just `$server.Query($sql,"msdb")`.

tests/Remove-DbaAgentJob.Tests.ps1
```diff
@@ -0,0 +1,77 @@
...
+        }
+        It "Should only contain $paramCount parameters" {
+            $params.Count - $defaultParamCount | Should Be $paramCount
+        }
+    }
+}
+Describe "$CommandName Integration Tests" -Tags "IntegrationTests" {
+    Context "Command removes jobs" {
+        BeforeAll {
+            $null = New-DbaAgentSchedule -SqlInstance $script:instance2 -Schedule dbatoolsci_daily -FrequencyType Daily -FrequencyInterval Everyday -Force
```
wsmelton: This just hit me but might be a good idea for time to just create the connection to your test instance using `Connect-DbaInstance`...can help shave off some time since you are calling multiple commands. Just have that connection and pass it to each command.

Would also give you access to the `Query()` method when you need to execute SQL statements.

wsmelton: You would call the connect within each context block to prevent conflicts.

functions/Remove-DbaAgentJobStep.ps1
jpomfret: Not sure what you mean on this one, this is within the Remove-DbaAgentJobStep function.

jpomfret:
Had trouble getting the tests to work with Connect-DbaInstance, it was reporting that the job still existed when it didn't. Using $server.refresh() didn't help either. Any thoughts?

functions/Remove-DbaAgentJobStep.ps1
wsmelton: Ah read file name wrong

potatoqualitee:
@jpomfret - I avoid Connect and do a Get and if it get returns nothing, then you're set.

PR #3174, opened 2018-01-30
Fixing Copy-DbaTableData wrong DestinationDatabase #3171

### LINKED ISSUE #3171: Copy-DbaTableData problem with DestinationDatabase
`Copy-DbaTableData -SqlInstance ds1 -Destination ds2 -Database db1 -DestinationDatabase db2 -Table table
`
If you start this you get a WARNING `[Copy-DbaTableData][16:42:40] table does not exist on destination` (but the table is there), the reason for this is, you wrote $Database instead of $DestinationTable

`$desttable = Get-DbaTable -SqlInstance $destServer -Table $DestinationTable -Database $Database -EnableException -Verbose:$false | Select-Object -First 1`
this should fix it:
`$desttable = Get-DbaTable -SqlInstance $destServer -Table $DestinationTable -Database $DestinationDatabase -EnableException -Verbose:$false | Select-Object -First 1`

Second small thing: https://dbatools.io/functions/copy-dbatabledata/ has a misstake "DatabaseDest" should be "DestinationDatabase"

General Troubleshooting steps
 Verified running the latest release of dbatools?
 Verified errors are not related to permissions?
 Can duplicate in new/clean PowerShell session (clean = powershell -NoProfile)?
Version Information
Operating System (Name|Version): Microsoft Windows 10 Pro 10.0.15063
PowerShell Version: 5.1.15063.413
SQL Server (Edition|Version): Microsoft SQL Server Standard 13.0.4466.4

road42:
## Type of Change

 - [X] Bug fix (non-breaking change, fixes #3171)
 - [X] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
BugFix
### Commands to test

Copy-DbaTableData

### DIFF
--- a/tests/Copy-DbaTableData.Tests.ps1
+++ b/tests/Copy-DbaTableData.Tests.ps1
@@ -5,21 +5,31 @@ Write-Host -Object "Running $PSCommandpath" -ForegroundColor Cyan
 Describe "$commandname Integration Tests" -Tags "IntegrationTests" {
     BeforeAll {
         $db = Get-DbaDatabase -SqlInstance $script:instance1 -Database tempdb
+        $db2 = Get-DbaDatabase -SqlInstance $script:instance2 -Database tempdb
         $null = $db.Query("CREATE TABLE dbo.dbatoolsci_example (id int);
             INSERT dbo.dbatoolsci_example
             SELECT top 100 1
             FROM sys.objects")
         $null = $db.Query("CREATE TABLE dbo.dbatoolsci_example2 (id int)")
+        $null = $db2.Query("CREATE TABLE dbo.dbatoolsci_example3 (id int)")
     }
     AfterAll {
         $null = $db.Query("DROP TABLE dbo.dbatoolsci_example")
         $null = $db.Query("DROP TABLE dbo.dbatoolsci_example2")
+        $null = $db2.Query("DROP TABLE dbo.dbatoolsci_example3")
     }
 
     It "copies the table data" {
         $null = Copy-DbaTableData -SqlInstance $script:instance1 -Database tempdb -Table dbatoolsci_example -DestinationTable dbatoolsci_example2
         $table1count = $db.Query("select id from dbo.dbatoolsci_example")
         $table2count = $db.Query("select id from dbo.dbatoolsci_example2")
+        $table1count.Count -eq $table2count.Count
+    }
+
+    It "copies the table data to another instance" {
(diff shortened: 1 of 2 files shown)


tests/Copy-DbaTableData.Tests.ps1
```diff
@@ -5,15 +5,18 @@ Write-Host -Object "Running $PSCommandpath" -ForegroundColor Cyan
 Describe "$commandname Integration Tests" -Tags "IntegrationTests" {
     BeforeAll {
         $db = Get-DbaDatabase -SqlInstance $script:instance1 -Database tempdb
+        $db2 = Get-DbaDatabase -SqlInstance $script:instance2 -Database tempdb
```
wsmelton: @road42 we are not allowed to do this with Appveyor. We cannot connect between two VMs in that environment.

wsmelton: This is not allowed so will have to be removed. We cannot connect between two VMs in Appveyor environment. As well, I would think even doing it would cause instability in our testing process.

wsmelton: If you want to add a test for the parameter you fixed simply create another database on the same Appveyor instance. Then that would provide code coverage for the parameter you fixed.

wsmelton: Ok...I have that wrong this is apparently allowed.

potatoqualitee: Sorry, my bad. I misunderstood! We allow instance to instance but not VM to VM.

niphlod:
just to clarify: whatever test uses both $script:instance2 and $script:instance1 ends up in the slowest scenario that has both instances alive. if a test can live WITHOUT it, it'd be better to. if a test NEEDS it, it'll run. Details on https://github.com/sqlcollaborative/dbatools/wiki#appveyor-setup

potatoqualitee:
Thank you for the clarification! This particular Test is for a `Copy-Dba*` which I believe justifies two instances. While you _can_ do local instance copies, I think it'd be better to ensure the Destination server is actually used. Otherwise, we could accidentally write to source both times.

potatoqualitee:
Thanks so much, @road42! Should I wait to merge?

road42:
No I think you can merge this PR. 

The other one is not so easy to fix: If you use a single SourceTable and you don‘t add a single DestinationTable Parameter the script starts copying the SourceTable to the first table found in Target. I‘ll do some investigation in this first :)

PR #3178, opened 2018-01-31
Measure-DbaDiskSpaceRequirement

pollusb:
A function to calculate disk space requirement when you plan to migrate databases

## Type of Change
  - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
You have a large list of databases to migrate and you want to make sure you will not hit a disk full.

### Approach
If you copy database in an empty server, you don't really need this, even if it should do the trick. This is more for refreshing database for DEV or TEST directly from PROD. Databases are sometime there, sometime not ! This will calculate disk space before the migration.

### Commands to test
Check out the examples in help

### Learning
It's not easy to FULL OUTER JOIN objects. I use a nested loops then Compare-Object for the other part.
I'm happy with the consolidation using the Group-Object and using Sqlcollaborative.Dbatools.Utility.Size

### DIFF
--- a/functions/Measure-DbaDiskSpaceRequirement.ps1
+++ b/functions/Measure-DbaDiskSpaceRequirement.ps1
@@ -1,0 +1,309 @@
+Function Measure-DbaDiskSpaceRequirement {
+    <#
+        .SYNOPSIS
+            Calculate the space needed to copy and possibly replace a database from one SQL server to another.
+
+        .DESCRIPTION
+            Returns a file list from source and destination where source file may overwrite destination. Complex scenarios where a new file may exist is taken 
+            into account. This procedure will accept an object in pipeline as long as it as provide these required properties: Source, SourceDatabase, Destination.
+            Using this method will provide a way to prepare before a complex migration with lots of databases from different sources and destinations.
+            
+        .PARAMETER Source
+            The source SQL Server instance.
+
+        .PARAMETER SourceSqlCredential
+            Allows you to login to servers using SQL Logins instead of Windows Authentication (AKA Integrated or Trusted). To use:
+
+            $scred = Get-Credential, then pass $scred object to the -SourceSqlCredential parameter.
+
+            Windows Authentication will be used if SourceSqlCredential is not specified. SQL Server does not accept Windows credentials being passed as credentials.
+
+            To connect as a different Windows user, run PowerShell as that user.
+
+        .PARAMETER SourceDatabase
+            The database to copy. It MUST exist.
+        
+        .PARAMETER Destination
+            The destination SQL Server instance.
+
+        .PARAMETER DestinationSqlCredential
+            Same as SourceSqlCredential.
+
+        .PARAMETER DestinationDatabase
+            The database name at destination. May or may not be present. Unspecified name will assume database name be the same as source.
+
+        .PARAMETER Credential
+            Windows credentials to connect via CIM/WMI/PowerShell remoting for MountPoint definition.
+
+        .PARAMETER Consolidate
+            Will summarize space by ComputerName and MountPoints.
+
+        .PARAMETER NbItems
+            Will show a progress bar if specified.
+
+        .NOTES
+            Tags: Migration
+
+            Website: https://dbatools.io
+            Copyright: (C) Chrissy LeMaire, clemaire@gmail.com
+            License: GNU GPL v3 https://opensource.org/licenses/GPL-3.0
+
+        .LINK
+            https://dbatools.io/Measure-DbaDiskSpaceRequirement
+
+        .EXAMPLE
+            Measure-DbaDiskSpaceRequirement -SI INSTANCE1 -SD DB1 -DI INSTANCE2 -Consolidate
+
+            Calculate a simple migration with one database with then same name at destination. In this scenario,
+            no space is required since we will recover 448 KB on destination disk.
+
+            DatabaseName1   DatabaseName2   Name1     Name2   SizeKB1 SizeKB2 DiffKB ComputerName MountPoint
+            -------------   -------------   -----     -----   ------- ------- ------ ------------ ----------
+            DB1             DB1             DB1       DB1        3072   -3000     72 INSTANCE2    D:\
+            DB1             DB1             DB1_log   DB1_log     504   -1024   -520 INSTANCE2    D:\
+
+            ComputerName MountPoint RequiredSpaceKB
+            ------------ ---------- ---------------
+            INSTANCE2    D:\                   -448
+
+        .EXAMPLE
+            @([PSCustomObject]@{Source='SQL1';Destination='SQL2';Database='DB1'}, 
+              [PSCustomObject]@{Source='SQL1';Destination='SQL2';Database='DB2'}
+            ) | Measure-DbaDiskSpaceRequirement -Consolidate -NbItems 2
+
+            Using a PSCustomObject with 2 databases to migrate with a progress bar
+
+        .EXAMPLE
+            ($CSV = Import-Csv -Path .\migration.csv -Delimiter "`t") | Measure-DbaDiskSpaceRequirement -Consolidate -NbItems $CSV.Count
+
+            Using a CSV file. You will need a header in migration.csv "Source<tab>Destination<tab>Database"
+
+        .EXAMPLE
+            ($RST = Invoke-DbaSqlCmd -SqlInstance DBA -Database Migrations -Query 'select Source,Destination,DatabaseName from refresh.Migrations' `
+                | Measure-DbaDiskSpaceRequirement -Consolidate -Verbose -NbItems $RST.Count
+
+            Using a SQL table. We are DBA after all!
+    #>
+    
+    [CmdletBinding()]
+    Param(
+        [Parameter(Mandatory,ValueFromPipelineByPropertyName)]
+        [Alias('SI','From')]
+        [string]$Source,
+
+        [Parameter(Mandatory,ValueFromPipelineByPropertyName)]
+        [Alias('SD','Database','DatabaseName')]
+        [string]$SourceDatabase,
+
+        [Alias('SC')]
+        [PSCredential]$SourceSqlCredential,
+
+        [Parameter(Mandatory,ValueFromPipelineByPropertyName)]
+        [Alias('DI','To')]
+        [string]$Destination,
+
+        [Parameter(Mandatory=$false,ValueFromPipelineByPropertyName)]
+        [Alias('DD')]
+        [string]$DestinationDatabase,
+
+        [Alias('DC')]
+        [PSCredential]$DestinationSqlCredential,
+
+        [PSCredential]$Credential, # For Windows access to MountPoints
+        [switch]$Consolidate, # Default would be detail only.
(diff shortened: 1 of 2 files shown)


wsmelton:
Concept of the command I agree with but a few items to address...

We have command that pull file info for a given database. Did you consider using those before using the SMO code?

I don't quite agree with this parameter, NbItems. I see no reason to give the user an option to not show the progress bar. However there are very few command we even include a progress bar on.

The format of the command in general. Your internal functions need to be in the begin block of your function and not on the end. 

Please use camel casing on your variable names inside the function. The parameters of the primary function get a capital for the first letter.

A few areas you use F1 as a variable name but it will make it more clear if you just use file1 and such.

wsmelton:
Yeah, quick check and please remove the parameter on the progress bar. There is a preference variable in PS that folks can use if they want to turn that off or leave it on.

pollusb:
Thanks for all the comments.

About the progress bar, how do you get the number of items in the pipeline ?

Thanks

wsmelton:
Input from the pipeline is streamed so you cannot really get the total count that I'm aware of. In those cases, though the progress bar I don't recall will even show regardless.

Examples you can check are the Backup and Restore command. I know when I pipe to those there is never a progress bar shown in my environments.

pollusb:
I can't use Get-DbaDatabaseFile because it doesn't work with a French version of SQL Server AND
It assume that files are on MountPoint of 1 character in length which is wrong.

For the progress bar, I will simply removed it for now.
I moved the private functions inside the begin block.
I will renamed the variables for CamelCasing.

functions/Measure-DbaDiskSpaceRequirement.ps1
```diff
@@ -0,0 +1,355 @@
...
+        if(!$DestinationDatabase) {
+            $DestinationDatabase = $SourceDatabase
+        }
+        Write-Verbose "$Source.[$SourceDatabase] -> $Destination.[$DestinationDatabase]"
+
+        $DB1 = Get-DbaDatabase -SqlInstance $Source -Database $SourceDatabase -SqlCredential $SourceSqlCredential
+        if(!$DB1) {
+            Stop-Function -Message "Database [$SourceDatabase] MUST exist on Source Instance $Source." -ErrorRecord $_
+        }
+<<<<<<< HEAD
```
wsmelton: Need to clean up the merge conflict references in the file.

functions/Measure-DbaDiskSpaceRequirement.ps1
```diff
@@ -0,0 +1,355 @@
...
+        [Alias('SD','Database','DatabaseName')]
+        [string]$SourceDatabase,
+        [Alias('SC')]
+        [PSCredential]$SourceSqlCredential,
+        [Parameter(Mandatory,ValueFromPipelineByPropertyName)]
+        [Alias('DI')]
+        [string]$Destination, 
+        [Parameter(Mandatory=$false,ValueFromPipelineByPropertyName)]
+        [Alias('DD')]
+        [string]$DestinationDatabase,
```
wsmelton: This may need a longer discussion in Slack but I'm not sure I follow the need to have two parameters for the database. 

In my use of this command, it would be to determine if a server has enough space to hold the database from my source server onto the destination server. This is what I get from your synopsis. If that is the case then it does not matter if the destination database already exists or not (in all reality it likely does not). If it did happen to exist I would not care because I could be refreshing a development environment from production, and I'm going overwrite whatever is there.

All that to say in the end you should really only need to have a parameter for `[object[]]$Database` and iterate through each one to see if there is enough space on the destination to hold it.

functions/Measure-DbaDiskSpaceRequirement.ps1
```diff
@@ -0,0 +1,355 @@
...
+        [Alias('SI','SqlInstance')]
+        [string]$Source,
+        [Parameter(Mandatory,ValueFromPipelineByPropertyName)]
+        [Alias('SD','Database','DatabaseName')]
+        [string]$SourceDatabase,
+        [Alias('SC')]
+        [PSCredential]$SourceSqlCredential,
+        [Parameter(Mandatory,ValueFromPipelineByPropertyName)]
+        [Alias('DI')]
+        [string]$Destination,
```
wsmelton: This should be `[DbaInstanceParameter]`

wsmelton: Also please remove the alias.

functions/Measure-DbaDiskSpaceRequirement.ps1
```diff
@@ -0,0 +1,355 @@
...
+                | Measure-DbaDiskSpaceRequirement -Consolidate -Verbose
+
+            Using a SQL table. We are DBA after all!
+    #>
+
+    [CmdletBinding()]
+    Param(
+        [Parameter(Mandatory,ValueFromPipelineByPropertyName)]
+        [Alias('SI','SqlInstance')]
+        [string]$Source,
```
wsmelton: This should be `DbaInstanceParameter`, and the aliases can be removed.

functions/Measure-DbaDiskSpaceRequirement.ps1
```diff
@@ -0,0 +1,355 @@
...
+        [Parameter(Mandatory,ValueFromPipelineByPropertyName)]
+        [Alias('SD','Database','DatabaseName')]
+        [string]$SourceDatabase,
+        [Alias('SC')]
+        [PSCredential]$SourceSqlCredential,
```
wsmelton: We do not set an alias on parameters so those can be removed from this command.

functions/Measure-DbaDiskSpaceRequirement.ps1
```diff
@@ -0,0 +1,355 @@
...
+        [Parameter(Mandatory,ValueFromPipelineByPropertyName)]
+        [Alias('DI')]
+        [string]$Destination, 
+        [Parameter(Mandatory=$false,ValueFromPipelineByPropertyName)]
+        [Alias('DD')]
+        [string]$DestinationDatabase,
+        [Alias('DC')]
+        [PSCredential]$DestinationSqlCredential,
+        [PSCredential]$Credential, # For Windows access to MountPoints
+        [switch]$Consolidate # Default would be detail only.
```
wsmelton: Inline comments can be removed.

Consolidate sounds like a details parameter we used to have and we have opted to stop using this method of controlling what information is output. We set the default view by only allowing a handful of properties to be output, and then if the user wants to see all of it they would pipe the command to `| select *`.

functions/Measure-DbaDiskSpaceRequirement.ps1
```diff
@@ -0,0 +1,355 @@
...
+                [Parameter(Mandatory)]
+                $ComputerName,
+                [PSCredential]$Credential
+            )
+            if(!$CacheMP[$ComputerName]) {
+                try {
+                    $CacheMP.Add($ComputerName, (Get-DbaDiskSpace -ComputerName $ComputerName -Credential $Credential -EnableException))
+                    Write-Verbose "CacheMP[$ComputerName] is in cache"
+                } catch {
+                    Write-Warning "Can't connect to $ComputerName. CacheMP[$ComputerName] = ?"
```
wsmelton: This can error for other reasons than connection issues.

wsmelton: As well, catch blocks need to utilize `Stop-Function` instead of a write command. You can view other commands in the module to see how it is used.

functions/Measure-DbaDiskSpaceRequirement.ps1
```diff
@@ -0,0 +1,355 @@
...
+            }
+            if($CacheMP[$ComputerName] -eq '?') {
+                return '?'
+            }
+            foreach($M in $CacheMP[$ComputerName]) {
+                if($Path -like "$($M.Name)*") {
+                    return $M.Name
+                }
+            }
+            Write-Warning "Path $Path can't be found in any MountPoints of $ComputerName"
```
wsmelton: This should be using `Write-Message` in place of any `-Warning, -Verbose, or -Output` commands.

functions/Measure-DbaDiskSpaceRequirement.ps1
```diff
@@ -0,0 +1,355 @@
...
+                    return $M.Name
+                }
+            }
+        }
+    }
+    process {
+        if(!$DestinationDatabase) {
+            $DestinationDatabase = $SourceDatabase
+        }
+        Write-Verbose "$Source.[$SourceDatabase] -> $Destination.[$DestinationDatabase]"
```
wsmelton: This can be `Write-Message -Level VeryVerbose -Message "$($Source.[$SourceDatabase]) -> $($Destination.[$DestinationDatabase])"`

functions/Measure-DbaDiskSpaceRequirement.ps1
pollusb: Another use for this command, would be to refresh already existing database in destination environment. Database could be with the same name or not. In both scenario (refresh or new copy), using a database name for destination can be useful as a database with the same name may exist but I really want to create a NEW copy of it.

functions/Measure-DbaDiskSpaceRequirement.ps1
pollusb: Consolidate generate a second output. I don't understand what you ask of me here. But the only viable option would be to ALWAYS return the consolidation with details or with no details. What is your tought on returning 2 tables?

functions/Measure-DbaDiskSpaceRequirement.ps1
pollusb: You want me to remove alias but I find alias in 416 files out of 752 which is more than half. They provide interface with the pipeline. I can remove the shortest one because they are really meant to shorten the call. But SqlInstance is THE way to call an instance in dbatools.

functions/Measure-DbaDiskSpaceRequirement.ps1
pollusb: In this case, I don't want to stop-function but to continue with a warning. So I guess Write-Message -Level Warning -Message should do?

functions/Measure-DbaDiskSpaceRequirement.ps1
wsmelton: You should be returning one object.

The only new column you have in the example for that second object is `RequiredSpaceKB`. This can just be added to the main object, which already has the other properties.

functions/Measure-DbaDiskSpaceRequirement.ps1
wsmelton: `Stop-Function .... -Continue`

functions/Measure-DbaDiskSpaceRequirement.ps1
wsmelton: `SqlInstance` is used only when you are working with one server. This server utilizes parameters for source and destination, which the commands we have with those parameters do not include aliases.

pollusb:
Hi, I made the changes requested.
Thanks

potatoqualitee:
Thanks so much! Will add to next release 💯

PR #3181, opened 2018-01-31
Wait stats

### LINKED ISSUE #3169: Get-DbaWaitStatistic not correctly filtering Ignorable Waits

- [Y ] Verified running the latest release of dbatools?
- [Y ] Verified errors are not related to permissions?
- [ Y] Can duplicate in new/clean PowerShell session (clean = `powershell -NoProfile`)?

## Steps to Reproduce

The Get-DbaWaitStatistic function, when called without the IncludeIgnorable switch does not return same information as the original query from Paul Randal.

In the original query the Ignorable Waits are filtered out of result set prior to calculation of the Percentage Figures while in the function the filter is only applied during output. This results in Ignorable Waits being included as part of the total percentage of waits even when they are not displayed. This results in displayed percentages being too low and some waits being incorrectly excluded. 

The results should match those of original query

sqllensman:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #3169)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Fix query to return same results as obtained from the Paul Randal source query

### Approach
Moves filter for ignorable Waits back into the actual SQL Query. This gives correct Percentage calculations

### Commands to test
Get-DbaWaitStatistic

### DIFF
--- a/functions/Get-DbaWaitStatistic.ps1
+++ b/functions/Get-DbaWaitStatistic.ps1
@@ -866,10 +866,10 @@
     }
     process {
         foreach ($instance in $SqlInstance) {
-            Write-Message -Level Verbose -Message "Attempting to connect to $instance"
+            Write-Message -Level Verbose -Message "Connecting to $instance"
 
             try {
-                $server = Connect-SqlInstance -SqlInstance $instance -SqlCredential $SqlCredential #-MinimumVersion 9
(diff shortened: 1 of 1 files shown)


potatoqualitee:
thanks so much! 👍 👍

PR #3201, opened 2018-02-05
new way to load the module

potatoqualitee:
Loads a big combined file using a dotload if .git is not found in the directory. This facilitates both end-users and devs who are adding new commands.


potatoqualitee:
thanks, it's created from scratch and signed!

PR #3202, opened 2018-02-06
Invoke-DbaLogshipping Add Prefix Functionality fixes #3109 

### LINKED ISSUE #3109: Enhancement to Invoke-DbaLogShipping
Since there is already a parameter of SecondaryDatabaseSuffix, it would be nice to have one for SecondaryDatabasePrefix as some environments might have them reversed.

sanderstad:
Added parameter -SecondaryDatabasePrefix
Changed logic to support both the prefix and the suffix
Changed error messages to support the new parameter
Changed restore to support both the prefix and suffix correctly


sanderstad:
Workings of the new parameter
image

potatoqualitee:
thanks so much!

PR #3206, opened 2018-02-06
Restore-DbaDatabase - Address param order

wsmelton:
Standard in commands is to have `SqlInstance` and `SqlCredential` as first parameters. The restore command has path as the first parameter. As well the examples in the command show providing these first and then the path.

Adjusting that to place those parameters as first two and then `Path` is third.


potatoqualitee:
👍

PR #3214, opened 2018-02-09
Fix and Update Test-DBAJobOwner(#2217)

joshcorr:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [x] Documentation
 - [ ] Build system
 

### Purpose
 
Update the output of the command to use Select-DefaultView
Remove any use of the -Detailed parameter
Update Documentation and Examples
Fix the existing output of the to reflect what the description says it should do
### Approach

Replaced the 'Return' with Setlect-DefaultView
Aliased the "Detailed" parameter for the new "All" parameter as the default functionality is to only return non matching jobs.
Included a Test-DbaDeprecation for the "Detailed" parameter
Included new Examples and updated descriptions
Corrected a filter for ExcludeJob which returned no results
Added an If block to set the default login to SA if the Login parameter is not present.
### Screenshots

Before:
image

After:
image


wsmelton:
@joshcorr 
I would say the `-All` switch should actually be removed as that functionality is outside of the purpose of the command. There is no reason to have the command return all jobs when they are testing for a specific login, it should only test and not just return everything.

If they want everything that is what the `get` command would be for. I do not believe the other test commands return all objects with a parameter.

So to follow the standard we have for other test commands I would say remove that parameter.

@potatoqualitee  @niphlod

joshcorr:
@wsmelton 
Okay, I can do that. I didn't want this to be breaking change, so I added the -All parameter. 
 I was trying to keep the functionality as best as I could from the previous state. The -Detailed parameter in the previous state was used for returning all results whether the owner of the job matched or not.

By removing this parameter the cmdlet will only return results where the Owners do not match, unless the -Job parameter is given then it will return the specific job regardless of the OwnerMatch status.

potatoqualitee:
awesome, thanks so much @joshcorr ! Will merge when @wsmelton approves

potatoqualitee:
Fantastic :D Thank you both!

PR #3217, opened 2018-02-11
Added new function New-DbaXESmartCsvWriter

spaghettidba:


functions/New-DbaXESmartCsvWriter.ps1
```diff
@@ -0,0 +1,95 @@
...
+        .SYNOPSIS
+            This Response type is used to write Extended Events to a CSV file.
+
+        .DESCRIPTION
+            This Response type is used to write Extended Events to a CSV file.
+
+        .PARAMETER OutputFile
+            Specifies the path to the output CSV file.
+
+        .PARAMETER OverWrite
```
wsmelton: Minor case update. The parameter name is showing as `Overwrite` in param block, just need to update this one to match.

functions/New-DbaXESmartCsvWriter.ps1
```diff
@@ -0,0 +1,95 @@
...
+
+            Example: duration > 10000 AND cpu_time > 10000
+
+        .PARAMETER EnableException
+            By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
+            This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
+            Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
+
+        .NOTES
+            Website: https://dbatools.io
```
wsmelton: Not sure if we did this on the other commands, can add `Tags: XE, XESmartTarget` or something to group all of your commands together. We use the tag with `Find-DbaCommand`.

functions/New-DbaXESmartCsvWriter.ps1
```diff
@@ -0,0 +1,95 @@
...
+        [string]$OutputFile,
+        [switch]$Overwrite,
+        [string[]]$Event,
+        [string[]]$OutputColumn,
+        [string]$Filter,
+        [switch]$EnableException
+    )
+    begin {
+        try {
+            Write-Host "$script:PSModuleRoot\bin\XESmartTarget\XESmartTarget.Core.dll"
```
wsmelton: Please update this one to `Write-Message -Level Verbose -Message "$script:PSModuleRoot\bin\XESmartTarget\XESmartTarget.Core.dll"`

functions/New-DbaXESmartCsvWriter.ps1
```diff
@@ -0,0 +1,95 @@
...
+            Add-Type -Path "$script:PSModuleRoot\bin\XESmartTarget\XESmartTarget.Core.dll" -ErrorAction Stop
+        }
+        catch {
+            Write-Host "$script:PSModuleRoot\bin\XESmartTarget\XESmartTarget.Core.dll"
```
wsmelton: Not sure we need another `Write-Message` on the catch block?

functions/New-DbaXESmartCsvWriter.ps1
```diff
@@ -0,0 +1,95 @@
...
+            Stop-Function -Message "Could not load XESmartTarget.Core.dll" -ErrorRecord $_ -Target "XESmartTarget"
```
wsmelton: I'd probably update this one to be the full path, that way it is captured in the logging system.

So something like `-Message "Could not load $($script:PSModuleRoot\bin\XESmartTarget\XESmartTarget.Core.dll)"...`

functions/New-DbaXESmartCsvWriter.ps1
potatoqualitee: Thanks, I added those tags to all the other commands and will soon commit to this command with the same

functions/New-DbaXESmartCsvWriter.ps1
potatoqualitee: This will be replaced with the same routine as the other commands

potatoqualitee:
Thanks to you both!

functions/New-DbaXESmartCsvWriter.ps1
spaghettidba: Sorry, my bad. I left the Write-Host in there, but I should have deleted it. It was there for my own debugging purposes. Same goes for the one in the catch block.

PR #3224, opened 2018-02-12
Get/Set-DbaDatabaseState revamp

niphlod:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Purpose
Being able to use the commands. SMO ultra-wonky enumeration on not-so-accessible databases is .... suboptimal. Skipping over inaccessible made half of the features go away (damn mass-rewrites...) 
This has been a long-standing issue and I hated it, and hated it, and hated it, and tried to find out what property triggered whatever enumeration on each version... and came out empty/without a consistent root-cause. 
So.... I circumvented it. Granted, if SMO somehow fixes the bug we can revert to the "old way", but since dbatools exists *also* to avoid SMO nightmares, this is a good case for being smart the way SMO isn't.

### Approach
We had tests that did not work for some time. Part of the blame was on:
- typos 
- hurry
- get-dbadatabase being as wonky as SMO enumeration
- ...

Given we just need to traverse props which are in sys.databases .... the new function(s) ask there for data instead of db properties. Nonetheless, every previous feature has been kept, and it also works with Case Sensitive instances, with a little bit of added love. Tests have been rewritten to avoid wonky Get-DbaDatabase calls, and now it *just works &trade;*

### Commands to test
Included pester tests are pretty extensive


functions/Set-DbaDatabaseState.ps1
```diff
@@ -285,9 +279,14 @@ Gets the databases from Get-DbaDatabase, and sets them as SINGLE_USER, dropping
                 Write-Message -Level Warning -Message "Database $db is a system one, skipping"
                 Continue
             }
+            $stateCache = @{}
```
wsmelton: Why is this variable called "cache"?

niphlod: because we may get out-of-order instances when piped, and we only need one call per instance.

wsmelton: Oh, I understand what the variable is for...just curious to know why it is named in that manner. Meaning we don't have variables in most commands with "cache". Why not just call it `$dbState` like it is in the other command in this PR?

niphlod: it's a cache. $dbState is for a single instance. $dbStates sounded funny. let's settle for $dbStatuses ?

niphlod:
groan. will rebase and reissue a new one.

PR #3240, opened 2018-02-15
Messaging update for the most part

wsmelton:
Updating commands to use the messaging system.

A few areas of formatting also updated, but the focus is on messaging updates.

Hopefully, I didn't break the commands 😁


potatoqualitee:
yay! thank you 🙇

PR #3248, opened 2018-02-16
Get/Set-DbaDatabaseState, performance improvements

niphlod:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Purpose
Being able to use the commands. SMO ultra-wonky enumeration on not-so-accessible databases is .... suboptimal. Skipping over inaccessible made half of the features go away (damn mass-rewrites...) 
This has been a long-standing issue and I hated it, and hated it, and hated it, and tried to find out what property triggered whatever enumeration on each version... and came out empty/without a consistent root-cause. 
So.... I circumvented it. Granted, if SMO somehow fixes the bug we can revert to the "old way", but since dbatools exists *also* to avoid SMO nightmares, this is a good case for being smart the way SMO isn't.

### Approach
We had tests that did not work for some time. Part of the blame was on:
- typos 
- hurry
- get-dbadatabase being as wonky as SMO enumeration
- ...

Given we just need to traverse props which are in sys.databases .... the new function(s) ask there for data instead of db properties. Nonetheless, every previous feature has been kept, and it also works with Case Sensitive instances, with a little bit of added love. Tests have been rewritten to avoid wonky Get-DbaDatabase calls, and now it *just works &trade;*

### Commands to test
Included pester tests are pretty extensive

reissue of #3224

### DIFF
--- a/functions/Get-DbaDatabaseState.ps1
+++ b/functions/Get-DbaDatabaseState.ps1
@@ -7,7 +7,7 @@ Gets various options for databases, hereby called "states"
 .DESCRIPTION
 Gets some common "states" on databases:
  - "RW" options : READ_ONLY or READ_WRITE
- - "Status" options : ONLINE, OFFLINE, EMERGENCY
+ - "Status" options : ONLINE, OFFLINE, EMERGENCY, RESTORING
  - "Access" options : SINGLE_USER, RESTRICTED_USER, MULTI_USER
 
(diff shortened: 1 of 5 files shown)


potatoqualitee:
thansk for the awesome tests! If I were to test, I'd probably just merge it right on in admittedly, bits prep is overwhelming :/. @wsmelton would you be able to give approval when you get a moment?

PR #3288, opened 2018-02-22
Invoke-DbaSqlQuery, fixes #3164

### LINKED ISSUE #3164: Invoke-DbaSqlCmd should accept databases through pipeline
## Feature Request

Let `Invoke-DbaSqlCmd` accept databases as input object.

### Problem to solve
The description of `Invoke-DbaSqlCmd` says it was created to align with our functions and accept pipeline input. Yet, since every query is executed against a database, the objects piped in should actually be databases instead of instances.

### Additional information

Actually, any object that has both a `SQLInstance` and a `database` property should do.

niphlod:
## Type of Change
 - [x] Bug fix (non-breaking change, fixes #3164)
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Purpose
Fixing #3164 . Not sure if there's a better way to do it, but I'm always ready for suggestions

### DIFF
--- a/functions/Invoke-DbaSqlQuery.ps1
+++ b/functions/Invoke-DbaSqlQuery.ps1
@@ -41,6 +41,9 @@ function Invoke-DbaSqlQuery {
         .PARAMETER AppendServerInstance
             If this switch is enabled, the SQL Server instance will be appended to PSObject and DataRow output.
 
+        .PARAMETER DatabaseCollection
+            A collection of databases (such as returned by Get-DbaDatabase)
+
         .PARAMETER EnableException
             By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
             This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
@@ -71,7 +74,7 @@ function Invoke-DbaSqlQuery {
             "server1", "server1\nordwind", "server2" | Invoke-DbaSqlQuery -File "C:\scripts\sql\rebuild.sql"
 
             Runs the sql commands stored in rebuild.sql against the instances "server1", "server1\nordwind" and "server2"
-        
+
         .EXAMPLE
             Get-DbaDatabase -SqlInstance "server1", "server1\nordwind", "server2" | Invoke-DbaSqlQuery -File "C:\scripts\sql\rebuild.sql"
 
@@ -112,12 +115,12 @@ function Invoke-DbaSqlQuery {
         [switch]
         $AppendServerInstance,
 
+        [parameter(ValueFromPipeline = $true)]
+        [Microsoft.SqlServer.Management.Smo.Database[]]$DatabaseCollection,
+
         [Alias('Silent')]
         [switch]
-        $EnableException,
-
-        [parameter(ValueFromPipeline = $true)]
-        [Microsoft.SqlServer.Management.Smo.Database[]]$DatabaseCollection
+        $EnableException
 
     )
 
--- a/functions/Invoke-DbaSqlQuery.ps1
+++ b/functions/Invoke-DbaSqlQuery.ps1
@@ -18,9 +18,9 @@ function Invoke-DbaSqlQuery {
         The database to select before running the query. This list is auto-populated from the server.
 
         .PARAMETER Query
-            Specifies one or more queries to be run. The queries can be Transact-SQL, XQuery statements, or sqlcmd commands. Multiple queries in a single batch may be separated by a semicolon.
+            Specifies one or more queries to be run. The queries can be Transact-SQL, XQuery statements, or sqlcmd commands. Multiple queries in a single batch may be separated by a semicolon or a GO
 
-            Do not specify the sqlcmd GO separator. Escape any double quotation marks included in the string.
+            Escape any double quotation marks included in the string.
 
             Consider using bracketed identifiers such as [MyTable] instead of quoted identifiers such as "MyTable".
 
@@ -74,7 +74,7 @@ function Invoke-DbaSqlQuery {
     #>
     [CmdletBinding(DefaultParameterSetName = "Query")]
     Param (
-        [parameter(Mandatory = $true, ValueFromPipeline = $true)]
+        [parameter(ValueFromPipeline = $true)]
         [Alias("ServerInstance", "SqlServer")]
         [DbaInstance[]]
         $SqlInstance,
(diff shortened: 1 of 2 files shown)


niphlod:
whoppsie, one or two things are missing. regarding "better ways to do it", is it better to enforce a less stricter "databasecollection" and then inspect if the passed object is either a full blown SMO db or a "loose" db representation with sqlinstance and database ?

functions/Invoke-DbaSqlQuery.ps1
```diff
@@ -233,7 +240,43 @@ function Invoke-DbaSqlQuery {
...
+        foreach($el in $DatabaseCollectionLoose) {
+            if ($el.SQLInstance -and $el.Database) {
+                $DatabaseCollection = $el.Database
+            }
+        }
+        if ((Test-Bound -ParameterName "Database") -and ((Test-Bound -ParameterName "DatabaseCollection") -or $DatabaseCollection)) {
+            Stop-Function -Category InvalidArgument -Message "You can't use -Database with piped databases"
+            return
+        }
+        if ((Test-Bound -ParameterName "SqlInstance") -and (Test-Bound -ParameterName "DatabaseCollection")) {
```
FriedrichWeinmann:
```powershell
(Test-Bound -ParameterName "SqlInstance") -and (Test-Bound -ParameterName "DatabaseCollection")
```
Can be simplified to:
```powershell
Test-Bound -ParameterName "SqlInstance", "DatabaseCollection" -And
```

functions/Invoke-DbaSqlQuery.ps1
```diff
@@ -233,7 +240,43 @@ function Invoke-DbaSqlQuery {
...
+            Stop-Function -Category InvalidArgument -Message "You can't use -SqlInstance with piped databases"
+            return
```
FriedrichWeinmann: Wouldn't parametersets be more helpful here?
Then the legal combinations will show up in help...
On the other hand, that would conflict with DatabaseCollectionLoose ...

functions/Invoke-DbaSqlQuery.ps1
```diff
@@ -233,7 +240,43 @@ function Invoke-DbaSqlQuery {
 
     process {
         if (Test-FunctionInterrupt) { return }
+        foreach($el in $DatabaseCollectionLoose) {
+            if ($el.SQLInstance -and $el.Database) {
+                $DatabaseCollection = $el.Database
+            }
+        }
+        if ((Test-Bound -ParameterName "Database") -and ((Test-Bound -ParameterName "DatabaseCollection") -or $DatabaseCollection)) {
```
FriedrichWeinmann: Can be reduced to:
```powershell
(Test-Bound -ParameterName "Database") -and $DatabaseCollection
```

niphlod:
Yeah, as told before pieces are missing, but I wanted to pick brains on the
actual implementation. That piece of logic was added as for request on the
original issue, and could be extended to all other functions that ATM only
accept piped SMO db objects...it makes sense that we improve "chainability"
between our functions if most of the properties match. Theorerically all
that output a sqlinstance and a db SMO under the database properties could
be piped here and everywhere needed.

Il 23 feb 2018 11:43 AM, "Friedrich Weinmann" <notifications@github.com> ha
scritto:

FriedrichWeinmann:
Heya, in that case, I'd opt for the following approach:
 - Design a `LimitedDatabaseParameter` parameter class for use when both Instance and Database are pipeable (that only accepts full SMO objects, whether straight or from a Database property)
 - Design a `DatabaseParameter` parameter class for use when only a database may be piped, or when the Database parameter does not support piping.

That would allow using parameter sets here and could be used almost anywhere in dbatools where we support specifying a database.

niphlod:
okay. lemme recap a bit

Technically every function acting on databases needs to be able to have a pointer to it.... but:

- SQLInstance and databasename as strings do not make the cut (e.g. SQLInstance drops the domain part)... in hindsight, we should have adopted a convention to pass along [dbainstance[]]
- I'm not aware of any "widespread" usage of a property that holds the SMO instance object in our functions.
- SMO databases fortunately have Parent property, so they do make the cut
- "Recent/adhering to codestyle/etc" functions that act on databases output a SQLInstance as string, "Name" or "DatabaseName" as strings but a "Database" (either hidden or not) that is the full SMO db object, so they do make the cut

Unfortunately piping just discerns "strong types", but we just do output a [pscustomobject] (granted, with the aforementioned conventions). 

So, design-wise, is there a necessity for two distinct parameters, that'd be "internal usage" only ? 

Can't we just accept a less strongly typed object and check for it to be a SMO db or look for a Database property that is a SMO db, and discard anything else ?

FriedrichWeinmann:
> Unfortunately piping just discerns "strong types", but we just do output a [pscustomobject] (granted, with the aforementioned conventions).
Who said it only accepts strong types?
Heck, `[DbaInstance]` does _not know_ the SMO datatypes, and can handle SMO Server objects all the same! Yes, I can build a wrapper that handles PSObjects.

niphlod:
okay, so we need an additional "type" that proxyies out a SMO db OR searches for a Database property with a SMO db and passes that along. Then we can spruce that over $DatabaseCollection and be cool kids.

potatoqualitee:
Hey niph - Microsoft allowed us to include sqlcmd. Do you want to wrap that? Then again, it'll be a string probably :( Nevermind - will wait for you and fred's agreement. Fred can you merge when it's ready?

niphlod:
@potatoqualitee : it's entangled with https://github.com/sqlcollaborative/dbatools/pull/3302 . A few days will probably pass before this can be in usable state, let's see where it goes. 
This is basically a first poc also to help consolidating all database-related functions ("coalescing" to a consistent output) and have all of them chainable one to the other. 
_And pleeeease, we also renamed this function: sqlcmd and invoke-dbasqlquery do not share a few -important- things. The main and ultra-big deal (as with invoke-sqlcmd2, which is misleadingly named) is SqlParameters, given there's no other way in "core SMO" (nor in sqlcmd) to use prepared statements._

niphlod:
we spawned #3314 for a more general approach to the original issue. 
This PR solves it "partially", as it takes as input a full SMO object like many other functions at this very moment.
Once the RFC settles, we'll probably spruce the new solution around every other command, this one included.
That being said, no need to drag this PR around.

potatoqualitee:
will be relying on you and @FriedrichWeinmann for this one, @niphlod

niphlod:
yeah, given the RFC will probably take some time IMHO it's better to merge this, then we'll pick it up when design-wise we'll have a better agreement.

potatoqualitee:
So merge now, @niphlod ? 🙇

PR #3292, opened 2018-02-23
Replace Get-DbaClusterActiveNode with Get-DbaClusterNode

joshcorr:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [x] Documentation
 - [ ] Build system
 

### Purpose
 
Originally it was to merely mark `-Detailed` as deprecated, but found that it was being used in another cmdlet so I set out to re-write it while trying to keep the same general functionality.
Ensure that the cmdlet is using the desired syntax and formatting as represented in other cmdlets.
Include a basic pester test to validate the parameters.

### Approach

Renamed the cmdlet to remove the 'Active'. By Default this cmdlet will return all Nodes in a cluster.
Using the `-ActiveNode` returns only the name of the node where the cluster is currently active.
Also removed reference to the ShouldProcess as it doesn't appear to be needed by this cmdlet.

NOTE: I have not updated the psd1 for the cmdlet name, and since this is a change in the name I assume there will need to be an alias for the old cmdlet, but I do not know how to do this.

### Commands to test

`Get-DbaClusterNode -SqlInstance mytestcluster`
Will return all of the nodes in the computer with properties for:
ComputerName, InstanceName, SqlInstance, StatusDescription , CurrentOwner 

`Get-DbaClusterNode -SqlInstance mytestcluster -ActiveNode`
Will return the property ComputerName for the active owner of the cluster

`Get-DbaClusterNode -SqlInstance mytestcluster -Detailed`
Will Return all of the nodes and a warning that the Detailed parameter is deprecated 

### Screen Shots
image


functions/Get-DbaClusterNode.ps1
```diff
@@ -0,0 +1,116 @@
...
+        [PSCredential]$SqlCredential,
+        [switch]$ActiveNode,
+        [switch]$Detailed,
+        [alias('Silent')]
+        [switch]$EnableException
+    )
+
+    begin {
+        Test-DbaDeprecation -DeprecatedOn 1.0.0 -Parameter Detailed
+        $server = Connect-SqlInstance -SqlInstance $SqlInstance -SqlCredential $SqlCredential -RegularUser
```
wsmelton: You can remove the `-RegularUser` parameter...it is default on the internal command.

functions/Get-DbaClusterNode.ps1
```diff
@@ -0,0 +1,116 @@
...
+    process {
+        if ($server.IsClustered -eq $false) {
+            Stop-Function -Message "Not a clusterd instance." -Category ConnectionError -ErrorRecord $_ -Target $SqlInstance -Continue
+        }
+
+        if ($ActiveNode) {
+            try{
+                if($server.VersionMajor -ge 11) {
+                    $sql = "SELECT * FROM sys.dm_os_cluster_nodes where is_current_owner = 1"
+                    $datatable = $server.ConnectionContext.ExecuteWithResults($sql).Tables
```
wsmelton: You can just use `$server.Query($sql)` and it will return the datatable object.

functions/Get-DbaClusterNode.ps1
```diff
@@ -0,0 +1,116 @@
...
+                            CurrentOwner      = $datatable.is_current_owner
+                        } | Select-DefaultView -Property ComputerName
+            }
+            catch{
+                Stop-Function -Message "Unable to query sys.dm_os_cluster_nodes on $server." -ErrorRecord $_ -Target $SqlInstance -Continue
+            }
+        }
+        else {
+            try{
+                $sql = "SELECT * FROM sys.dm_os_cluster_nodes"
```
wsmelton: I don't see the reason behind calling this if the version is less than 11? This looks redundant to me as-is.

functions/Get-DbaClusterNode.ps1
wsmelton: Need to add `-MinimumVersion 10` because the DMV used only applies to 2008 or higher.

potatoqualitee:
Great PR, thank you! Will merge once questions have been addressed.

joshcorr:
@potatoqualitee made the requested changes. I tested on a 2016 FCI I built for this cmdlet. Would like to write a better integration test, but don't believe we have a FCI in the appveyor environment. Also know that this cmdlet needs to be aliased as it is changing. Curious how this is done.

@wsmelton Thanks for pointing out that Connect-SqlInstance has a `-MinimumVersion` I hadn't read up on this one yet since most of my changes have been Documentation, or Output changes.

functions/Get-DbaClusterNode.ps1
```diff
@@ -0,0 +1,116 @@
...
+        [DbaInstanceParameter]$SqlInstance,
+        [PSCredential]$SqlCredential,
+        [switch]$ActiveNode,
+        [switch]$Detailed,
+        [alias('Silent')]
+        [switch]$EnableException
+    )
+
+    begin {
+        Test-DbaDeprecation -DeprecatedOn 1.0.0 -Parameter Detailed
```
wsmelton: Ah, we need to also add a deprecation call for the alias so it spits out the warning as well.

It would just be: `Test-DbaDeprecation -DeprecatedOn "1.0.0" -Alias Get-DbaClusterActiveNode`

PR #3298, opened 2018-02-24
Test-DbaTempDbConfiguration Minor Refactor and Test Coverage

awickham10:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Improve Test-DbaTempDbConfiguration test coverage.

### Approach

Integration tests use both the 2008 R2 and 2016 servers in Appveyor now. Refactored some of the Test-DbaTempDbConfiguration in order to minimize duplicate code and increase test coverage.

### Commands to test

See test file.

### DIFF
--- a/tests/Test-DbaTempDbConfiguration.Tests.ps1
+++ b/tests/Test-DbaTempDbConfiguration.Tests.ps1
@@ -27,27 +27,33 @@ Describe "$CommandName Unit Tests" -Tag 'UnitTests' {
         It "Should only contain $paramCount parameters" {
             $params.Count - $defaultParamCount | Should Be $paramCount
         }
+        It "Should throw on an invalid SQL Connection" {
+            Mock -ModuleName 'dbatools' Connect-SqlInstance { throw }
+            { Test-DbaTempDbConfiguration -SqlInstance 'MadeUpServer' -EnableException } | Should Throw
+        }
     }
 }
 
 Describe "$CommandName Integration Tests" -Tags "IntegrationTests" {
-    Context "Command actually works" {
-        $results = Test-DbaTempDbConfiguration -SqlInstance $script:instance2
-        It "Should have correct properties" {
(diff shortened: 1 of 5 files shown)


awickham10:
I have an erroneous file in here... working on getting it reverted.

tests/Test-DbaTempDbConfiguration.Tests.ps1
```diff
@@ -27,27 +27,33 @@ Describe "$CommandName Unit Tests" -Tag 'UnitTests' {
         It "Should only contain $paramCount parameters" {
             $params.Count - $defaultParamCount | Should Be $paramCount
         }
+        It "Should throw on an invalid SQL Connection" {
+            Mock -ModuleName 'dbatools' Connect-SqlInstance { throw }
+            { Test-DbaTempDbConfiguration -SqlInstance 'MadeUpServer' -EnableException } | Should Throw
```
potatoqualitee: Thanks so much for your updates! This one in particular is no longer needed as we test the internal command. It also adds 15 seconds to each test. Will merge upon removal 💯

awickham10: With the mock Connect-SqlInstance in there it shouldn't add any time to the tests. I did some comparisons on doing try/catch around the Connect-SqlInstance and the behavior of the function did change if you don't do the try/catch. I feel like it's a valid test to include to make sure that we are getting consistent behavior between cmdlets if an instance is unreachable.

potatoqualitee: i suck with mocks - can you test the timing of the two and see? 

genuine question and up for a change of mind: if the connect-sqlinstance code block is universal, shouldn't that cover consistency across the board? did this particular command vary from the template?

wsmelton: @potatoqualitee we do this in other test as well, but this is one area that we will be changing because we will no longer need to test the `Connect-SqlInstance` call in every test. Once we get the test for Connect-SqlInstance at 💯 we can remove mocks and all against it in all test.

PR #3301, opened 2018-02-26
Refactor Test-DbaFullRecoveryModel with Test-DbaRecoveryModel

### LINKED ISSUE #3222: Refactor Test-DbaFullRecoveryModel to be Test-DbaRecoveryModel
## Feature Request

Refactor `Test-DbaFullRecoveryModel` to be `Test-DbaRecoveryModel` in order that a type of recovery model may be tested on an instance. The default would still be to test the full recovery model.

Example: Test-DbaRecoveryModule -SqlInstance MyServer -RecoveryModel Simple

This change was purposed by @wsmelton  in PR #3218

joshcorr:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #3222 )
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [x] Documentation
 - [ ] Build system
 

### Purpose
 
Refactor Test-DbaFullRecoveryModel to be more modular as Test-DbaRecoveryModel.
Include Integration Tests for the cmdlet
Rename Tests, Functions, and any References to the old files.
### Approach

Included a parameter for `-RecoveryModel` (Default Execution without it is Full)
Provided validation for the `-RecoveryModel` inputs
Wrote Integration tests for default, and scoped executions.
Replaced the Existing files with the new cmdlet.

Note: Did not replace in the dbatools.psd1 file, but did include a Test-DbaDeprecation for the alias.
### Screenshots

Initial Change Output:
test-dbarecoverymodel

Revised Output:
image


wsmelton:
@potatoqualitee I'm not to fond of the `ActualRecoveryModel` property. It seems like we should have a bit/boolean property to output since it is a test command.

Looking over the other test commands we don't really have one common property name between them all. The valid login command has `found`, PowerPlan uses `isBestPractice`, and the OptimizeForAdhoc just has `RecommendedOptimizeAdHoc`. That last one is just an integer value.

From my thought if I want to test for a recovery model it should return a true/false property that can be easily checked against. Whether we call it `IsCompliant` or whatever.

wsmelton:
The main beef with this particular command is the property names seem to treat this like an sp_configure option (actual/configured), which is not possible with the recovery model itself. It is either set to full or it's not. So really having a "configured" and "actual" for this command is redundant.

joshcorr:
I agree the property is a bit strange. When it was just testing for Full recovery it makes sense, because you could see if a backup had been taken under that new model, but on Simple and BulkLogged it doesn't make sense. I also did not want to transform the output properties to be different for Full vs. the other two. Dropping this property would make it somewhat simpler. It should return what the recovery is configured regardless of if it has acted upon that yet or not.

potatoqualitee:
Naming is hard and that has always felt awkward. Great suggestions. Please make the suggested changes 👍 And thank you SOOO much for all of your contributions, Josh!

And your reviews, as always, :shawn:

joshcorr:
Absolutely @potatoqualitee! 😄 

But just to clarify which suggested change should be made dropping that property or replacing it with `IsCompliant`?

potatoqualitee:
i think so, what you think @ClaudioESSilva

ClaudioESSilva:
IMHO, for FULL RECOVERY MODEL we should have a way to say `pseudo-simple` when it is not compliant.
I would say `IsCompliant` with `TRUE`/`FALSE` and when `FALSE` shows something like `You need to take a full backup in order to get your database on FULL recovery model until then it is at pseudo-simple state` on a `Notes` field? (don't remember if we want to get rid of `Notes` like properties 😅 )

For this command, I really think is important to alert the user by giving a more descriptive message and not only with `FALSE`

joshcorr:
image

Saw there was a Conflicted file and decided to rebase on development. Didn't expect it to pull all of those commits into this merge request. 

@ClaudioESSilva  The notes property is still present, but not in the default view.

wsmelton:
I don't believe a more description message is required. I don't see the purpose of this command as being one to provide a consultant's answer (if that makes sense). It is more for utilizing checks (cough: dbachecks) or I know it will be used when I start working on the AG commands.

And yes...overall many of us don't want the notes property to stay around 😀 .

I see the command for more use like this:
```if (Test-DbaDbRecoveryModel -SqlInstance $server -Database DbToJoinInAg -RecoveryModule Full)...```

The issue with showing a message it is psuedo simple is also a false and myth just because they have not taken another full backup. If a database is in FULL recovery and is switched to SIMPLE, and then switched back to FULL you can start log backups again after taking a FULL **OR** a DIFF backup. SO in that since if neither of those have been done maybe we just show the recovery model as SIMPLE. Because logically it still is.

And we will need to fix the merge conflict @joshcorr so we can keep this PR intact.

joshcorr:
Okay, so I am purposing the following changes...
- Leave `ConfiguredRecoveryModel` alone
- Update `ActualRecoveryModel` to be either the Configured or Simple (removing psudo-simple)
- Remove the `Notes` property entirely (But will move the note to the `ActualRecoveryModel` help so people understand why this exists)
- Add back the conflicted file I removed. 

Sound good? @wsmelton @ClaudioESSilva @potatoqualitee

wsmelton:
Please provide an example of the new output.

joshcorr:
Had to fix how I was using Select-DefaultView, but included a revised screen shot above, and if everything checks out on appveyor it should be good to merge. @wsmelton

potatoqualitee:
thank you!

PR #3302, opened 2018-02-26
DbaDatabase & DbaDatabaseSmo parameter classes

FriedrichWeinmann:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Changes

 - Implemented `[DbaDatabase]` parameter class (Accepts SMO objects & String)
 - Implemented `[DbaDatabseSmo]` parameter class (Accepts SMO objects)

### Compatibility warning!
`[DbaDatabase]` conflicts with `[DbaInstance]` in cases where _both accept from pipeline_ !
*Do not, do not, do not use both at the same time with both accepting from pipeline!!!*


niphlod:
ahum, can't review anything, is above my knowledge. 
what does
`[DbaDatabase] conflicts with [DbaInstance] in cases where both accept from pipeline`
mean ? 
We kinda need either one of those piped in into invoke-dbasqlQuery

FriedrichWeinmann:
`[DbaDatabase]` also accepts strings
`[DbaDatabaseSmo]` does not (that's the sole difference between the two parameter classes)

If you use the former together with `[DbaInstance]` and pipe string to it (which both of them accept) _both_ parameters will be bound, so if you pipe the name of an instance, it'll also look for a database of the same name as the instance.

In those use-cases you have two options:

1. Use `[DbaDatabaseSmo]` and do not accept string for databases (easy to implement in script)
2. Use `[DbaDatabase]` and do some reasonably complex checking for whether both were received from pipeline and give one or the other precedence (not too much effort but significantly reduces readability and precludes use of parameter-sets)

> ahum, can't review anything, is above my knowledge.

Need you to give it a field test, not a C# crossreading 😄 
Specifically, whether it correctly interprets those custom objects with a Database property. I gave it a manual test, but since I don't know what of our commands produce them I can't guarantee it will properly work with them without some testing. Sooo ... just see whether their output can be cast to `[DbaDatabase]` and `[DbaDatabaseSmo]`.

niphlod:
Looking at the description, dbadatabase accepts "one-bit-too-much" (as discussed in the original PR, we cannot really use strings to identify databases reliably). 
Will give it a good run tomorrow for sure and report back.

FriedrichWeinmann:
I preliminarily merged this in order to avoid conflicts with the library down the road.
It is not officially released for use until/unless the RFC https://github.com/sqlcollaborative/dbatools/issues/3314 is confirmed.

PR #3303, opened 2018-02-26
Get-DbaDatabase, speedups

niphlod:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Speeding up enumeration for some usecases. The "framework" is there if anyone wants to fiddle with additional properties. The "game" is discerning from a raw sys.databases query the filtered properties, in order to use a pre-filtered collection to be passed down to additional filters.
Added here a new switch (I'm sure someone will come up with better naming) to pre-filter on isaccessible databases. Doing so down the line in most of the commands is ... non-optimal, so we'd better do it here.

### Approach
See the code

----

Hopefully nothing breaks, this is just a "rapid poc".

PS: the databases.refresh() at the end was chewing up resources as well. After a bit of discussion with @potatoqualitee, it seems that the only impacted usecase would be the one reported on piping to Remove-DbaDatabase, in which case .... it'd better fail than slowdown any other function.

### DIFF
--- a/functions/Get-DbaDatabase.ps1
+++ b/functions/Get-DbaDatabase.ps1
@@ -265,8 +265,10 @@ function Get-DbaDatabase {
                 $true { @($true) }
                 default { @($true, $false, $null) }
             }
-
-            $backed_info = $server.Query("SELECT *, SUSER_NAME(owner_sid) AS [Owner] FROM sys.databases")
+            function Invoke-QueryRawDatabases {
+                $server.Query("SELECT *, SUSER_NAME(owner_sid) AS [Owner] FROM sys.databases")
+            }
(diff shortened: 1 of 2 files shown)


niphlod:
yup, try passing in an instance with offline dbs and use -OnlyAccessible vs where isAccessible, or try selecting/excluding databases on an instance with lots of databases....

wsmelton:
I dumped this file over the one on a client server and I can't even use the OnlyAccessible switch, the parameter does not exist even though I can see it in the file. Restart PowerShell session makes no difference.

potatoqualitee:
ahhhhhhhhhhhhh fabulous!!

Issue #3275, opened 2018-02-19
Copy-DbaLinkedServer -UpgradeSqlClient assumes incorrect version based on Sql Major Version


sqlslinger:
### General Troubleshooting steps

- [X] Verified running the latest release of dbatools?
- [X] Verified errors are not related to permissions?
- [X] Can duplicate in new/clean PowerShell session (clean = `powershell -NoProfile`)?

### Version Information
 - Operating System (Name|Version): PS C:\Users\viacoboni2> Microsoft Windows Server 2012 R2 Standard|6.3.9600 N/A Build 9600
 - PowerShell Version: PSVersion                      4.0
 - SQL Server (Edition|Version): 
Microsoft SQL Server 2014 - 12.0.4422.0 (X64) 
	Jul 27 2015 16:56:19 
	Copyright (c) Microsoft Corporation
	Standard Edition (64-bit) on Windows NT 6.3 <X64> (Build 9600: ) (Hypervisor)

## Steps to Reproduce

Run Copy-DbaLinkedServer -UpgradeSqlClient where the source server is less than SQL 2014 and the destination server is SQL 2014.  The cmdlet will finish successfully.

Now script the copied linked server and you will see the provider is set to SQLNCLI12.  This doesn't exist, and should be SQLNCLI11 instead.

 - [ ] Attach any screenshots (if possible/allowed)
 - [ ] Attach logs generated by dbatools (`Get-DbatoolsLog`)
 - [ ] Attach output from PowerShell console (if possible/allowed)

### Problem to solve

The code for UpgradeSqlClient looks like this:

if ($UpgradeSqlClient -and $sql -match "sqlncli") {	
 	$newstring = "sqlncli$($destServer.VersionMajor)"	
 	Write-Message -Level Verbose -Message "Changing sqlncli to $newstring"	
 	$sql = $sql -replace ("sqlncli[0-9]+", $newstring)	
 	}

The assumption is that Sql version xx will have a SQLNCLIxx provider.  That is not true for SQL 2014 as the I get the following from @@version:
Microsoft SQL Server 2014 - 12.0.4422.0 (X64) 
	Jul 27 2015 16:56:19 
	Copyright (c) Microsoft Corporation
	Standard Edition (64-bit) on Windows NT 6.3 <X64> (Build 9600: ) (Hypervisor)
But I get this from EXEC sp_MSset_oledb_prop:
provider_name
SQLOLEDB
SQLNCLI11
ADsDSOObject
[...]

PR #3304, opened 2018-02-26
Issue 3275 - Copy-DbaLinkedServer Providers Incorrect

awickham10:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Fix issue #3275. Linked servers for SQL Server 2014 are incorrect (using SQLNCLI12 instead of SQLNCLI11).

### Approach

Pick the latest SQLNCLIXX registered instead of building it based on the SQL Server major version.

While I was in there I also changed the verbage for when a provider does not exist on the destinations erver from "Already exists" to "Provider missing." 

### Commands to test

See test case.

### Screenshots

N/A

### Learning

Funny enough, when I was looking for how to find the registered providers through SMO, I came across a bug that [Chrissy had a post about](https://blog.netnerds.net/2015/01/workaround-for-smo-bug-urn-could-not-be-resolved-at-level-oledbprovidersetting/).


potatoqualitee:
thank you so very much! 🍰

sqlslinger:
@awickham10, thanks for addressing this bug so quickly.

PR #3310, opened 2018-02-27
Add-DbaComputerCertificate - Changed Write-Verbose to Write-Message and UTF-8 BOM Encoding

awickham10:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Reduce deprecated verbosity commands.

### Approach

Changed Write-Verbose to Write-Message -Level Verbose. Saved file as UTF-8 with BOM encoding.

### DIFF
--- a/functions/Add-DbaComputerCertificate.ps1
+++ b/functions/Add-DbaComputerCertificate.ps1
@@ -106,13 +106,13 @@ function Add-DbaComputerCertificate {
 
             $cert = New-Object System.Security.Cryptography.X509Certificates.X509Certificate2
             $cert.Import($CertificateData, $Password, [System.Security.Cryptography.X509Certificates.X509KeyStorageFlags]::DefaultKeySet)
-            Write-Verbose "Importing cert to $Folder\$Store"
+            Write-Message -Level Verbose -Message "Importing cert to $Folder\$Store"
             $tempStore = New-Object System.Security.Cryptography.X509Certificates.X509Store($Folder, $Store)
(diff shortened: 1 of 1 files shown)


potatoqualitee:
SO GOOOOD!

wsmelton:
The encoding no this file needs to be reverted back to UTF-8 @potatoqualitee @awickham10

awickham10:
Why does it need reverted? The reason for the change was for ScriptAnalyzer's recommendation. If it's causing issues, is there anyway to make ScriptAnalyzer ignore this rule for that file?

`Missing BOM encoding for non-ASCII encoded file 'Add-DbaComputerCertificate.ps1'`

wsmelton:
Based on what editor you use you can configure it to ignore this rule, but overall we ignore most every info and warning rule in PSSA...at least right now. The PS Gallery runs PSSA against all modules that are submitted and only flag those that have errors thrown, so we only check against those rules right now to just make sure we are in compliance.

That particular rule has an open issue on PSSA repo if I recall. The reason we are ignoring it is because majority of our files are encoding in UTF-8 and that rule will flag a file needing the BOM encoding because one particular character or something.

awickham10:
Got ya. I thought it was kind of strange, but another file I had opened also had BOM so I didn't think much of it.

Should we need to modify manual.pester.ps1 to include PSUseBOMForUnicodeEncodedFile in $ScriptAnalyzerRulesExclude?

PR #3329, opened 2018-03-01
issue3328_js0505

js0505:
changed the url to eliminate the failure to install because of the change in TLS rules


potatoqualitee:
tks!

PR #3341, opened 2018-03-03
Add last query to collection

andrekamman:
Repeated a small section to make sure the last query in the script gets added as well.

## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Purpose
The parser for the diagnosticquery scripts had a bug, it didn't return the last query. The result was always one query short.
### Approach
Repeated a small section to make sure the last query is included
### Commands to test
Invoke-DbaDiagnosticQuery -SqlInstance localhost\sql2017 -UseSelectionHelper 
Should show 86 queries, the last one is called "Recent Full Backups"
Previously it would show "Automatic Tuning Options", which is wrong, that's not the last one in the script.

### DIFF
--- a/internal/functions/Invoke-DbaDiagnosticQueryScriptParser.ps1
+++ b/internal/functions/Invoke-DbaDiagnosticQueryScriptParser.ps1
@@ -66,5 +66,8 @@
         }
     }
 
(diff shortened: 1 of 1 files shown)


potatoqualitee:
awesome, thank you Andre!

PR #3348, opened 2018-03-06
Get-DbaBackupHistory, hash is faster than PS pipe

niphlod:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Purpose
processing groups was hindered if the list of backupsets was long. 

### Approach
go with an hash rather than PS + pipe + where

-----

hopefully doesn't break. Thanks to Tony for indirectly spotting the issue

### DIFF
--- a/functions/Get-DbaBackupHistory.ps1
+++ b/functions/Get-DbaBackupHistory.ps1
@@ -503,52 +503,61 @@ function Get-DbaBackupHistory {
                 Write-Message -Level SomewhatVerbose -Message "$($GroupedResults.Count) result-groups found."
                 $groupResults = @()
                 $BackupSetIds = $GroupedResults.Name
-                $BackupSetIds_List = $BackupSetIds -Join "','"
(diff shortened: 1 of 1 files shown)


niphlod:
thank stuart for the massive test suite, it makes so easy to work !
 
PS: if someone has a faster method (group-object also on filelist?!) please clock it when, e.g., you pull history -since 2 months ago ^_^

potatoqualitee:
Thanks so much, @niphlod ! @Stuart-Moore, would you like to do the honors? The code LGTM

Stuart-Moore:
Only query I've got is on the names of `$Start_` et al just look a bit strange to me, on first scan I thought there was some PS internal trickery going on. It's the `_`, may just `$Start` etc?

niphlod:
I'm accustomed to `var_` being "internal" variable names, no problem in changing.

Stuart-Moore:
Did wonder if it was a work practice. Will let the tests pass and then merge

PR #3361, opened 2018-03-12
add dacFxpath variable to publish-DbDacpac

RichieBzzzt:
- Allows a version of dacfx other than the one packaged with dbatools to be used.

## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Allow different versions of dacFx to be used other than the one packaged with dbatools.

### Approach
Adding variable and see if it is indluded by running Test-Bound, which is basically a wrapper for PSBoundParameters

### Commands to test
See 3rd example in help

### DIFF
--- a/functions/Publish-DbaDacpac.ps1
+++ b/functions/Publish-DbaDacpac.ps1
@@ -50,6 +50,9 @@ function Publish-DbaDacpac {
             This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
             Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
 
+        .PARAMETER DacFxPath
+            Path to the dac dll. If this is ommited, then the version of dac dll which is packaged with dbatools is used. 
+
         .NOTES
             Tags: Migration, Database, Dacpac
             Author: Richie lee (@bzzzt_io)
@@ -92,7 +95,8 @@ function Publish-DbaDacpac {
         [Switch]$ScriptOnly,
         [string]$OutputPath = "$home\Documents",
         [switch]$IncludeSqlCmdVars,
-        [switch]$EnableException
+        [switch]$EnableException,
+        [String]$DacFxPath
     )
 
     begin {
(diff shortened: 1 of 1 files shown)


RichieBzzzt:
I see the build failed on a rogue empty space. Have updated.

wsmelton:
What benefit does it server to use other versions?

RichieBzzzt:
Teams may be on older versions of SSDT or .NET in their developments environments. So instead of teams having to upgrade, which may be a slow process, or not even possible in some cases, we support different versions of SSDT by simply enabling teams to pass in a path to the dac dll that is already installed on whatever box they are running the "Publish-DbaDacpac" instead of relying on the one that is packaged with dbatools.

wsmelton:
Why don't we just include the possible versions of the DLL file and set a parameter for version they want to utilize? 

There are only two versions of SSDT supported by Microsoft right now, 2017 and 2015. Anything prior to 2015 is considered out of support by Microsoft. Just pull the 2015 DLL file and we add that to the module.

RichieBzzzt:
There are more than 2 versions as currently all these versions are supported/available in Nuget - 

https://www.nuget.org/packages/Microsoft.Data.Tools.Msbuild/

And aeven if we cherry-picked which ones to include,  48mb for each one unpacked, dbatools could get a bit bloated :-)

wsmelton:
Bloating will be cleaned up when we get the CI configured as all the ps1 and test files will no longer he shipped to the gallery.

So the latest version of the file does not support backward compatibility like most other development tools?

As well, if older versions are required we can just include option to download the Nugget file.

RichieBzzzt:
OK thanks for sharing your thoughts.

potatoqualitee:
Thank you, @RichieBzzzt for the PR and thank you @wsmelton for the review! I do see the benefit of this functionality. Initially, I was going to comment that it needed a `Test-Path` but see that it exists 👍 I'm unsure if a mix and match of SMO + dac versions is problematic but it doesn't seem like they are related. Can you please confirm that you have tested this and it works? I noticed a test failed but it was not related to your command (darn Appveyor)

RichieBzzzt:
Hello!

I have tested and it does work, both in terms of including and omitting the new variable. There was one failed build that was a result of my check-in (trailing space in the comment) but I've fixed that. The current failure is due to a previous PR.

potatoqualitee:
Thank you very much, Richie!

Issue #3330, opened 2018-03-01
Get-DbaDiskSpace IsSqlDisk is not populated when value cannot be deteremined


wsmelton:
This command utilizes CM object to grab disk information but when it fails to gather some bit of information it is not able populate the `IsSqlDisk` with a true/false value. It is just returning null.

Not sure if we won't to update this to return an "unknown" value.

I am able to still pull disk information on a given server so not sure I completely follow the error that is silently skipped:

```PSMessageDetails      :
Exception             : System.Management.Automation.MethodInvocationException: Exception calling "QueryCimRMInstance"
                        with "4" argument(s): "WinRM cannot process the request. The following error occurred while
                        using Kerberos authentication: Cannot find the computer <redacted>. Verify that the computer
                        exists on the network and that the name provided is spelled correctly." --->
                        Microsoft.Management.Infrastructure.CimException: WinRM cannot process the request. The
                        following error occurred while using Kerberos authentication: Cannot find the computer
                        <redacted>. Verify that the computer exists on the network and that the name provided is
                        spelled correctly.
                           at
                        Microsoft.Management.Infrastructure.Internal.Operations.CimSyncEnumeratorBase`1.MoveNext()
                           at
                        Sqlcollaborative.Dbatools.Connection.ManagementConnection.QueryCimRMInstance(PSCredential
                        Credential, String Query, String Dialect, String Namespace) in
                        D:\Code\Github\dbatools\bin\projects\dbatools\dbatools\Connection\ManagementConnection.cs:line
                        907
                           at CallSite.Target(Closure , CallSite , Object , Object , String , String , String )
                           --- End of inner exception stack trace ---
                           at System.Management.Automation.ExceptionHandlingOps.CheckActionPreference(FunctionContext
                        funcContext, Exception exception)
                           at System.Management.Automation.Interpreter.ActionCallInstruction`2.Run(InterpretedFrame
                        frame)
                           at
                        System.Management.Automation.Interpreter.EnterTryCatchFinallyInstruction.Run(InterpretedFrame
                        frame)
                           at
                        System.Management.Automation.Interpreter.EnterTryCatchFinallyInstruction.Run(InterpretedFrame
                        frame)
TargetObject          :
CategoryInfo          : NotSpecified: (:) [], MethodInvocationException
FullyQualifiedErrorId : CimException
ErrorDetails          :
InvocationInfo        : System.Management.Automation.InvocationInfo
[...]

Issue #2424, opened 2017-10-10
Make Get-DbaDiskSpace 1.0 compliant


potatoqualitee:
Including [dbainstance[]]

FriedrichWeinmann:
It is compliant by now, thank you to all who contributed

PR #3366, opened 2018-03-15
Get-DbaDiskSpace - 1.0 Compliance check and CheckForSql improvement

### LINKED ISSUE #3141: Get-DbaDiskSpace with CheckForSql switch only works for Default instance
## Feature Request

When Get-DbaDiskSpace is called with CheckForSql switch all instances on Computer should be checked so that accurate usage is provided

### Problem to solve
If a Server has a named instance, multiple instances or is in a Cluster then drives containing SQL Server files for these instances will not be necessarily flagged. 

### Additional information

For Clusters where instance name and Computer name are different this will always be an issue.

wsmelton:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #3141 #2424 #3330 )
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Add support for named instances and where it will actually validate if a disk is part of SQL Server for any instance on a given computer.

Although not likely needed added support for SQL Server 2000 just for the heck of it.

Updated test to include validating the CheckForSql parameter works, this should work fine on Appveyor.


wsmelton:
I'm not sure if this is just from the workflow in Appveyor but when this test is running none of the instances are running. So I updated the test on IsSqlDisk to just check for false, because there should be a drive on there that is always false. It may break local test when you run it but I guess we could add check for appveyor environment like we do in some other test.

potatoqualitee:
checks passed 👍

potatoqualitee:
Thanks so much!

PR #3401, opened 2018-03-20
Test-DbaOptimizeForAdHoc - added tests

jpomfret:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Added tests for Test-DbaOptimizeAdHoc

### Screenshots

image

### DIFF
--- a/tests/Test-DbaOptimizeForAdHoc.Tests.ps1
+++ b/tests/Test-DbaOptimizeForAdHoc.Tests.ps1
@@ -19,9 +19,15 @@ Describe "$CommandName Unit Tests" -Tag 'UnitTests' {
 
 Describe "$CommandName Integration Tests" -Tags "IntegrationTests" {
     Context "Command actually works" {
+        $results = Test-DbaOptimizeForAdHoc -SqlInstance $script:instance2
         It "Should return result for the server" {
-            $results = Test-DbaOptimizeForAdHoc -SqlInstance $script:instance2
             $results | Should Not Be Null
+        }
+        It "Should return 'CurrentOptimizeAdHoc' property as int" {
+            $results.CurrentOptimizeAdHoc | Should BeOfType System.Int32
+        }
+        It "Should return 'RecommendedOptimizeAdHoc' property as int" {
+            $results.RecommendedOptimizeAdHoc  | Should BeOfType System.Int32
         }
     }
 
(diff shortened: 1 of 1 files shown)


niphlod:
hi, every contribution, especially on tests, is welcome .... 
these are better than nothing but when _the evil contributor_ will introduce a bug all that _the evil_ must comply to is to return something. 
The added value of this test is close to zero (granted, now syntax can't be broken and parameter names can't change).
Especially if this comes from adding tests for dbachecks, the most important part is probably that the output doesn't change ^_^

jpomfret:
Suggestions are more than welcome, was struggling to decide how to test this. Should I do a before all set it to optimize for advice and test that it gets the right result?

niphlod:
Since you're dabbling in writing test, I'll explain my POV here in more detail, hopefully is "logic" even if it's not "the logic".... YMMV. 

I always try to describe the output too in unittests, so the casual (or advanced) contributor can be sure its modifications didn't break some "contract" if, e.g., the function is used in a higher-level tool (like dbachecks).
Mind that this doesn't mean that further modifications are not allowed or that function stops evolving: it just means that an _evil_ commit doesn't pass by without raising flags.
When a new feature is requested and/or the output needs to be tailored to the "next level", a failing test is a good alarm bell that, e.g., can initiate a review (or later spur an impact analysis) and (we'll get there sooner) promote an entry in the CHANGELOG.md . 

In this specific case the core of the function is syntactic sugar over SMO's `Configuration.OptimizeAdhocWorkloads.ConfigValue`, so there's no need to test what _hopefully_ MS tests before releasing SMO (i.e. no need to alter the instance, set forcefully the value and verify that is read correctly, also because users may run tests on *their* instances and this is a pretty high impact change).

Checking for the correctness of the `CurrentOptimizeAdHoc` property is then inherently "wasting resources" on something SMO grants by itself, so IMHO here is worth testing at least:
- for the presence of a `CurrentOptimizeAdHoc` property, of the correct type (int) 
- for the presence of a `RecommendedOptimizeAdHoc` property, of the correct type (int)

At the end, it'll assure that higher level tools that _probably_ use/compare those values continue to work in case of an _evil_ commit

jpomfret:
Thanks for the ideas, I didn't think about testing type of parameters, that would make sense here for sure if things like dbachecks were relying on them.

potatoqualitee:
thank you both 👍

PR #3438, opened 2018-03-27
Restore from snapshot  - cleaned up db loop

potatoqualitee:
for @niphlod - dont merge yet

### DIFF
--- a/functions/Restore-DbaFromDatabaseSnapshot.ps1
+++ b/functions/Restore-DbaFromDatabaseSnapshot.ps1
@@ -81,12 +81,12 @@ function Restore-DbaFromDatabaseSnapshot {
         [Alias('Silent')]
         [switch]$EnableException
     )
-
+    
     process {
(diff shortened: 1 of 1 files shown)


niphlod:
unfortunately still 1 minute spent before hitting line 132. The real monster - at least here - is line 99. Probably because no props are selected, it goes around and enums all the things. 
Commenting out line 99 and searching/replacing `$allDbs` with `$server.Databases` closes the deal.
Still slower than a tailored select but is more "SMOish".

niphlod:
scratch that. was testing passing a connect-dbainstance as sqlinstance. 
when passed a string as sqlinstance, even that mod results in no improvements.

potatoqualitee:
i cannot find a smoish fix and would be happy for you to fix it with your spruced select 👍

niphlod:
Okay!
PS: is there a place to shout out for SMO lovers that we have a problem?

Il mer 28 mar 2018, 10:25 Chrissy LeMaire <notifications@github.com> ha
scritto:

PR #3439, opened 2018-03-27
Get-DbaBackupHistory, fix for same-named databases

niphlod:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Purpose
Getting correct last backups in a recurring scenario when database with the same name get dropped and recreated. We continue to trust LSNs, which ultimately is what makes a restore possible.

### Approach
I slapped in the most annoying comment ever, because we revisited bits and pieces of the logic over and over without finding a proper fix (till now). 
Long story short (still long), LSNs are ordered correctly always. The whole point of LSNs is to be comparable and sortable to build a chain. Opposed to most of the scripts out there, we don't rely on backup_finish_date (even if that's what SMO does) because msdb holds dates with the timezone of the server. DST happens rarely, but they happen everywhere. That's why we go for last_lsn being "truthful" always. A small hiccup in this scenario happens when msdb holds the history for a database that has been backed up with the same name. LSNs are guaranteed to be ordered for the same database, but when you drop and recreate a database with the same name it's not the same database. 
database_guid, on the other end, changes. For the same database_guid, last_lsn is ordered correctly. So, to circumvent the hiccup, before trusting last_lsn, we filter out the history seeking the database that has the database_guid which has the last backup_finish_date (which translates to selecting the history that is relative to the last database backed up on the server with the same name).  

Short story: we trust last_lsn whenever possible. When it's not possible we filter out those rows whose database_guid has the oldest backup_finish_date. We don't go for backup_finish_date for DST. We don't go for backup_set_id because there's no explicit documentation stating that is ordered the same as LSNs.

### DIFF
--- a/functions/Get-DbaBackupHistory.ps1
+++ b/functions/Get-DbaBackupHistory.ps1
@@ -295,6 +295,17 @@ function Get-DbaBackupHistory {
                     if ($deviceTypeFilter) {
                         $devTypeFilterWhere = "AND mediafamily.device_type $deviceTypeFilterRight"
                     }
+                    # recap for future editors (as this has been discussed over and over):
+                    #   - original editors (from hereon referred as "we") rank over backupset.last_lsn desc, backupset.backup_finish_date desc for a good reason: DST
+                    #     all times are recorded with the timezone of the server
+                    #   - we thought about ranking over backupset.backup_set_id desc, backupset.last_lsn desc, backupset.backup_finish_date desc
+                    #     but there is no explicit documentation about "when" a row gets inserted into backupset. Theoretically it _could_
+                    #     happen that backup_set_id for the same database has not the same order of last_lsn.
+                    #   - given ultimately to restore something lsn IS the source of truth, we decided to trust that and only that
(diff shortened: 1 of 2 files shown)


potatoqualitee:
haha enjoyed the comment. thank you for the fix. I'll let @Stuart-Moore take a look and merge /cc @Zhunya

niphlod:
@Zhunya : went and added the same filtering in the subquery that is applied in the outer

@cviorel : added a fix for your finding and a regr test, thanks for the catch

niphlod:
okay, could replicate. Working on a fix right now

potatoqualitee:
yay all passes!

niphlod:
yup. was due to backup concluding in the same second. seems that "second" is the default precision for msdb backupset-ty data.

potatoqualitee:
is that a 👍 to merge?

niphlod:
@Stuart-Moore approved when I discussed the main gist of the idea, @Zhunya notes have been incorporated and @cviorel fix is in. 
Tests pass, so I think we're good.

potatoqualitee:
yessssssss!

Issue #3462, opened 2018-04-03
New function - Test-DbaLoginPassword


PsPsam:
## Feature Request

  Simply answer the question: What problem or scenario would you like to solve with dbatools?
https://github.com/sqlcollaborative/dbachecks/issues/413

To be able to make that test in dbachecks I created a test function for dba tools. I looked but was unable to find it.

This function will by default test SQL Logins for null or same password as username. The function can take in other password by parameter -SQLPass as an array or by a csv file.

The function will return information about the user that will get caught by this check.

### Problem to solve
Bad password policy!

potatoqualitee:
@scinobe had something like this - brandon, did you hash the values then compare?

PsPsam:
Added the PR New function - Test-DbaSqlLoginWeakPassword #3462 #3464

PsPsam:
Missed to remove the comment out the first time in the first post

scinobe:
There's a function :)
This query returns the matches.


$pwdhashquery = "select serverproperty('machinename') as 'ServerName',
isnull(serverproperty('instancename'),serverproperty('machinename')) as
'InstanceName',
 name as 'Login',
 is_disabled,
 create_date,
 IS_SRVROLEMEMBER('sysadmin',name) as is_sa,
 IS_SRVROLEMEMBER('dbcreator',name) as is_dbcreator,
 IS_SRVROLEMEMBER('bulkadmin',name) as is_bulkadmin,
 IS_SRVROLEMEMBER('diskadmin',name) as is_diskadmin,
 IS_SRVROLEMEMBER('processadmin',name) as is_processadmin,
 IS_SRVROLEMEMBER('securityadmin',name) as is_securityadmin,
 IS_SRVROLEMEMBER('serveradmin',name) as is_serveradmin,
 IS_SRVROLEMEMBER('setupadmin',name) as is_setupadmin
from master.sys.sql_logins
where pwdcompare(name,password_hash) = 1
order by name
option (maxdop 1);"

PsPsam:
@scinobe @potatoqualitee i don't have the role part in the query else it looks like the same query. 
But @wsmelton and @zippy1981 had some other feedback on the PR itself that might be usefull.

scinobe:
I think I had added the roles myself so that I could tell how big of a risk
the account posed...

PR #1523, opened 2017-06-30
Added password hashing C# code

zippy1981:
Changes proposed in this pull request:
 - C# code to create passworh hash objects

How to test this code: 
- [ ] `New-Object Sqlcollaborative.Dbatools.Utility.DbaPasswordHash hashes from sql server`
- [ ] `[Sqlcollaborative.Dbatools.Utility.DbaPasswordHash]::GenerateHash()`

Has been tested on minimum requirements:
- [ ]  Powershell 3
- [ ]  Windows 7
- [ ]  SQL Server 2000

Has been tested on maximum requirements:
- [ ]  SQL Server vNext
- [ ]  Windows 10
- [ ]  Azure Database

Tests for tester:
- [ ] Working/useful help content, including link to command on dbatools web site
- [ ] All examples work as advertised
- [ ] Does not contain template content
- [ ] Does not contain excessive/unnecessary amounts of comments
- [ ] Works remotely
- [ ] Works locally
- [ ] Works on lower versions or throws error specifying version not supported
- [ ] Works with named instances
- [ ] Works with clustered instances
- [ ] Handles offline/read only databases
- [ ] Supports multiple servers (at the command line or piped from Get-SqlRegisteredServerName)
- [ ] No un-handled errors which stop the command working with multiple servers

PR #3464, opened 2018-04-03
New function - Test-DbaSqlLoginWeakPassword #3462 

PsPsam:
## Type of Change

 - [/] Pester test is included
 - [ ] Nunit test is included
 - [X] Documentation
 - [ ] Build system
 

### Purpose
To test SQL Logins for null/same as username/user provided passwords.

### Approach
Tests SQL Logins for null/same as username/user provided passwords.

### Screenshots

image

### Learning

My first PR to DBAtools so please give me feedback :) I used a current dbatools function as template don't remember though which one it was.
I added a pester test also but I'm not the best at that :) so it is also a copy but it's something :)


functions/Test-DbaSqlLoginWeakPassword.ps1
```diff
@@ -0,0 +1,131 @@
...
+
+            Windows Authentication will be used if SqlCredential is not specified. SQL Server does not accept Windows credentials being passed as credentials.
+
+            To connect as a different Windows user, run PowerShell as that user.
+
+        .PARAMETER SqlPass
+            Specifies a list of passwords to include in the test for weak passwords.
+
+        .PARAMETER Path
+            Specifies a csv file path of passwords to include in the test for weak passwords.
```
wsmelton: This should be removed and simply have the command handle multiple logins. The user would pass the data to the function via whatever method they use to get data out of the CSV file, or other sources.

If that makes sense?

functions/Test-DbaSqlLoginWeakPassword.ps1
```diff
@@ -0,0 +1,131 @@
...
+            Test all SQL logins that the password is null, same as username and defined in the csv file on SQL server instance Dev01
+
+    #>
+    [CmdletBinding()]
+    Param (
+        [parameter(Position = 0, Mandatory = $true, ValueFromPipeline = $true)]
+        [Alias("ServerInstance", "SqlServer", "SqlServers")]
+        [DbaInstanceParameter[]]$SqlInstance,
+        [PSCredential]$SqlCredential,
+        [String[]]$SQLPass,
```
wsmelton: Need to rename this parameter....specifying pass or password does not sound right. Have to think a bit on this one.

zippy1981:
A while back I wrote [a crap ton of beautiful C# code that implemented all versions SQL Server password hash. Not the actual code encryption algorithim](https://github.com/sqlcollaborative/dbatools/pull/1523), which is built into .NET, but just code to actually parse and verify the password hashes. One idea was to do a rainbowtable password check.

We have a stupid problem that makes us stuck with .net 4.0 so it never got merged. I don't think we should hold up this cmdlet until that works. However, if we can shave the .net 4.0 yak, this function could be modified to take a rainbow table.

functions/Test-DbaSqlLoginWeakPassword.ps1
wsmelton: Having a parameter named "pass" or "password" could also get flagged in the gallery for security issue.

functions/Test-DbaSqlLoginWeakPassword.ps1
```diff
@@ -0,0 +1,131 @@
...
+        [String]$Path,
+        [Alias('Silent')]
```
wsmelton: You can remove the alias, this was only needed for old command...new ones do not require it.

functions/Test-DbaSqlLoginWeakPassword.ps1
```diff
@@ -0,0 +1,131 @@
...
+        [switch]$EnableException
+    )
+
+    begin {
+        Test-DbaDeprecation -DeprecatedOn 1.0.0 -Parameter Detailed
```
wsmelton: This line can be removed.

functions/Test-DbaSqlLoginWeakPassword.ps1
```diff
@@ -0,0 +1,131 @@
...
+            } catch {
+                Stop-Function -Message "Failure" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
+            }
+            if ($Server.VersionMajor -lt 10) {
+                throw "PWDCOMPARE is only supported on sql 2008 and up. Quitting."
+            }
+            Write-Message -Level Debug -Message "Executing $sql"
+            Write-Message -Level Verbose -Message "Testing: same username as Password"
+            Write-Message -Level Verbose -Message "Testing: the following Passwords $CheckPasses"
+            $server.Query("$sql")
```
wsmelton: The output should include ComputerName, InstanceName, and SqlInstance (see other commands line Get-DbaLogin).

functions/Test-DbaSqlLoginWeakPassword.ps1
niphlod: `Weaklist` ?!

potatoqualitee: `Dictionary`?

functions/Test-DbaSqlLoginWeakPassword.ps1
potatoqualitee: `DictionaryPath`?

wsmelton: Oh, I think the path itself and functionality should be removed...not just renaming the parameter.

functions/Test-DbaSqlLoginWeakPassword.ps1
wsmelton: Dictionary sounds good [to me](https://en.wikipedia.org/wiki/Dictionary_attack).

functions/Test-DbaSqlLoginWeakPassword.ps1
PsPsam: So if I understand correct. The command should be able to handle that you do import-csv file.csv | test-dbasqlloginweakpassword ? 
this instead of having the function beeing able to read the csv file?

It will make the function more what it should do. I'm just not sure on how to make it accept through the pipeline when you have that option allreary on one parameter?

functions/Test-DbaSqlLoginWeakPassword.ps1
PsPsam: password was wrong and didnt want to use Pass so sqlPass was the next thing I thougt off that made sence to me. Dictonary I think is better!!

functions/Test-DbaSqlLoginWeakPassword.ps1
PsPsam: Added ComputerName, InstanceName and SqlInstance. Did it from the tsql since I can get multiple results on one server. 

I do not know if that makes any difference? I don't call Select-DefaultView either do I need to?

PsPsam:
Should we rename the function as suggested by @wsmelton from Test-DbaSqlLoginWeakPassword  to Test-DbaLogin @potatoqualitee // @niphlod 

And add the flag for the different test -valid and -weakpassword?

At the same time I like a function to do one thing and not multiple things depending on what switch I write.

potatoqualitee:
I like the two diff functions, too. One of the problems with Test-DbaLogin is that..what would it default to? The output would be drastically different depending on the switch too.

Please rename to `Test-DbaLoginPassword`

I know it's only for SqlLogins but I Test-DbaValidLogin is only for Windows and we didn't name it AD or whatever. I'd rather have the command tell them that only SQL Logins are supported if they try a Windows login.

PsPsam:
Change the name. I feel its better also. 

The command doesn't take a login as parameter, and it only looks in sys.sql_logins that only holds sql logins.

functions/Test-DbaLoginPassword.ps1
```diff
@@ -0,0 +1,113 @@
...
+                SysLogins.modify_date as ModifiedDate,
+                SysLogins.default_database_name as DefaultDatabase
+            FROM sys.sql_logins SysLogins
+            INNER JOIN @WeakPwdList WeakPassword ON (PWDCOMPARE(WeakPassword.WeakPwd, password_hash) = 1
+                OR PWDCOMPARE(REPLACE(WeakPassword.WeakPwd,'@@Name',SysLogins.name),password_hash) = 1)"
+    }
+    process {
+        foreach ($instance in $SqlInstance) {
+            try {
+                $server = Connect-SqlInstance -SqlInstance $instance -SqlCredential $sqlcredential
```
wsmelton: This needs to include -MinimumVersion 10 unless you tested and know the view in use works against 2005?

functions/Test-DbaLoginPassword.ps1
```diff
@@ -0,0 +1,113 @@
...
+                Write-Message -Message "Connected to: $instance." -Level Verbose
+            } catch {
+                Stop-Function -Message "Failure" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
+            }
+            if ($Server.VersionMajor -lt 10) {
+                throw "PWDCOMPARE is only supported on sql 2008 and up. Quitting."
```
wsmelton: Didn't read far enough.

You can remove this check and just use the MinimumVersion parameter for the connection.

functions/Test-DbaLoginPassword.ps1
```diff
@@ -0,0 +1,113 @@
...
+
+    [CmdletBinding()]
+    Param (
+        [parameter(Position = 0, Mandatory = $true, ValueFromPipeline = $true)]
+        [Alias("ServerInstance", "SqlServer", "SqlServers")]
+        [DbaInstanceParameter[]]$SqlInstance,
+        [PSCredential]$SqlCredential,
+        [String[]]$Dictionary,
+        [switch]$EnableException
+    )
```
wsmelton: This definitely needs a filter for Login and ExcludeLogin. I have client systems with 100 plus logins on it that I don't want to wait for it to check all just to get one.

functions/Test-DbaLoginPassword.ps1
PsPsam: Ahhh didn't know that. Change the code

functions/Test-DbaLoginPassword.ps1
PsPsam: Changed this. Was this in the function I worked from. But Changed to the correct way

nvarscar:
> A while back I wrote a crap ton of beautiful C# code that implemented all versions SQL Server password hash. Not the actual code encryption algorithim, which is built into .NET, but just code to actually parse and verify the password hashes. One idea was to do a rainbowtable password check.

There is [Get-PasswordHash](https://github.com/sqlcollaborative/dbatools/blob/development/internal/functions/Get-PasswordHash.ps1) which can be used to get the hash string for the known password, not sure if it can help with rainbowtable checks though (which to me sounds like rocket science) 🚀

wsmelton:
Just noticed one thing about the output on your example. How do we now the password is actually weak?

I would say this could be called Test-DbaLoginPassword and then have the property weak just be true or false.

potatoqualitee:
Yes, my apologies. I haven't been around much. I'll merge and take a closer look.

PsPsam:
Sorry to all, been to busy at work to do anything about this. And to tired when I got home. 

Thanks @potatoqualitee

PR #3466, opened 2018-04-03
manual.pester, do not consider utf-8 without bom

niphlod:
(do state)

Since the rule has often incorrect checking, better to leave it out of the picture till gets properly fixed.


potatoqualitee:
thanks niph!

PR #3470, opened 2018-04-04
Adjusting default output and test for Get-DbaOperatingSystem

wsmelton:
Just adjusting default output of Get-DbaOperatingSystem, and added the `Caption` property that gives you the full user-friendly OS Edition and Version (not sure why I didn't have this in there before).

Updating test to also include the standard unti testing on parameters and such.

Found cool way to get the common parameters based on commit on PowerShell repo:

```powershell
#this will output the "default" parameter that functions get with cmdletbinding()
[System.Management.Automation.PSCmdlet]::CommonParameters

#this outputs the additional parameters that come with ShouldProcess being set to true
[System.Management.Automation.PSCmdlet]::OptionalCommonParameters
```
image


potatoqualitee:
@wsmelton - could you give it a look?

PR #3472, opened 2018-04-04
Get-DbaDatabaseFile - refix array issue

potatoqualitee:


potatoqualitee:
actually, will merge since it works and we can rework if you see an issue

ashdar:
The changes to the cmdlet are what I expect. I didn't write the those tests, but I am sure that you wrote them and/or they work. In short, I think this is OK to merge.

PR #3482, opened 2018-04-04
appveyor timings and speedups

niphlod:
## Type of Change
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [x] Build system
 

### Purpose
Since we're on the verge/in the middle of testing spree, having a way to "monitor" timings regains some value. As everyone can watch times explicitely for each step, most of the time is spent on setting up the needed services. 
I'll tackle startup scripts in the future (todolist: remove specific actions that make tests go nuclear on "private" testing setups and port them over inside specific tests that need those preparation, see if we can avoid loading dbatools that takes some time too, etc)

Main deal for now is shaving off precious time by:
- fetching the repo via zip rather than by git (we ship DLLs in source control and cloning the repo is a PITA) 
- install early all deps for every scenario to enable caching deps when possible 

NB: cache will not be preserved on PRs by default, looking at Appveyor's docs, and for a good reason

NB2: once in a while we may want to touch appveyor.yml which will trigger cache invalidation (e.g. when dependencies needed for tests get updated)

Future deals: 
- incorporating DLL tests in the main runner so it if fails (as it is for some time now) build fails too
- avoid setting up services when no tests need to be ran (e.g. `(do something)`)

### Approach
Pester is able to generate full blown Nunit blorbs that appveyor happily accepts, although with the zillions test we have it makes it not useful at all. 
We just use Appveyor's facility to store times for main steps of each scenario run, so we all get a specific idea on what takes time and what doesn't.
As for the cache, we use choco to install, and cache choco dirs. Also, we cache directories used to store PS modules needed which are not available via choco.
Choco dir on appveyor is kinda big, but for the time being I could not find an easier and succint method to install "manually" and be able to cache our dependencies.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,6 +3,14 @@
 build: false
 version: 0.9.{build}
 
+cache:
+  - C:\ProgramData\chocolatey\bin -> appveyor.yml
+  - C:\ProgramData\chocolatey\lib -> appveyor.yml
+  - C:\Program Files\WindowsPowerShell\Modules\PSScriptAnalyzer -> appveyor.yml
+  - C:\Program Files\WindowsPowerShell\Modules\Pester -> appveyor.yml
+
+shallow_clone: true
+
 # Set build info
 environment: 
   environment: development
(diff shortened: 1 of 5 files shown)


potatoqualitee:
Thanks so much! I’m on mi mobile but saw it passed. What about for our dependencies like Pester and codecov? Can that be sped up with zip downloads?

potatoqualitee:
Ignore my dumb comment! I saw you addressed it thanks

PR #3488, opened 2018-04-05
Small Text and Punctuation Fixes found during web page builds

gbargsley:
Small text and punctuation fixes.


potatoqualitee:
fails cuz of a diff test that i'll look into, thanks!

PR #3492, opened 2018-04-06
Fix restoring simple model dbs with diffs

### LINKED ISSUE #3469: Restore-DbaDatabase fails if source DB was in SIMPLE RecoveryMode and you specify a RecoveryTime
Tested in 0.9.318

- [X] Verified running the latest release of dbatools?
- [X] Verified errors are not related to permissions?
- [X] Can duplicate in new/clean PowerShell session (clean = `powershell -NoProfile`)?

### Version Information
 - Operating System (Windows 10|Windows7): 
 - PowerShell Version: 5.1.14409.1012 
 - SQL Server (Developer|10.50.6220): 

## Steps to Reproduce

In a situation where you are restoring a FULL backup plus one out of several DIFF backups, Restore-DbaDatabase fails if the source DB was in SIMPLE RecoveryMode and you specify a RecoveryTime in order to filter out DIFF backups that are "too new". 

To sum up: The restore command fails with a syntax error because dbatools includes STOPAT in the RESTORE database command, even though the DB is in simple mode.

**Here is the scenario:**
1. Database ABC is running in SIMPLE mode.
2. We are using Hallengren-style backups
3. A full backup occurs Sunday mornings at 6 AM. Let's call that 2/4/2018 06:00.
4. A DIFF backup occurs every other day, Monday through Saturday, at 6 AM. We have backups from 2/5/2018 through 2/10/2018, always at 6 AM.
5. (Pretend that) today is 2/13/2018. We want to restore the ABC database to a DEV server, but only up to Wednesday (2/7/2018) morning
6. Restore-DbaDatabase -SqlInstance TestSrv -database ABC -MaintenanceSolution -path \\backupsvr\backups -RestoreTime '2/7/2018 9:00 AM'
7. This command finds the FULL backup from 2/4. **Good**. This command also finds the most recent DIFF backup that is older than our RestoreTime. That would be the DIFF from 2/7/2018 06:00. **Good**.
9. **Problem**: The RESTORE command for the full backup file is incorrect because it includes a STOPAT value. You can easily see this by using -OutputScriptOnly. STOPAT is not legal syntax for a restore of a full backup file for a database running in simple recovery mode. STOPAT is legal syntax for a restore of a full backup of a database running in bulk logged or fully logged recovery modes. 

**My proposed fix:**
1. BackupHistory.cs needs to be enhanced to add the RecoveryModel attribute to the Sqlcollaborative.Dbatools.Database.BackupHistory class.
2. Get-DbaBackupInformation needs to be enhanced to output the RecoveryModel in addition to the attributes it already provides
3. Invoke-DbaAdvancedRestore needs to check if backup type (which is already available) is full and RecoveryModel (which requires the first change) is simple, if so then do not set a ToPointInTime on the $restore object. 

### Additional information
I've got code changed and ready to go in my repo, I am looking for someone to reassure me that this is a sane thing to do and I'll submit a PR.

Stuart-Moore:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #3469)
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Cope with restore chains from dbs in Simple Recovery mode
### Approach
Added RecoveryModel to the BackupHistory object, populated, and RestoreToPoint set to $null if it's simple.


niphlod:
hum, at this point $CanDoStopAt can be removed alltogether, altough I'm still not sure why this whole shenanigan is needed if I got the point right on the original issue. 

PS: get-dbabackuphistory needs a new column ^_^

Stuart-Moore:
Think it’s just down to how we want to handle it.

Will sort those bits out later this evening when I’m not impersonating a
network engineer.

Stuart-Moore:
Last commit is from me. Just didn't set the new machine up properly.

Both those issues fixed

PR #3506, opened 2018-04-09
Read-DbaBackupHeader: now multi-threaded

nvarscar:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Got tired of restores running indefinitely for any folder that has >100 files (read: any production backup folder) reading every single header of every backup file one by one. This change enables the command to submit up to 10 tasks simultaneously using runspaces and run header checks in parallel.
With this approach, restores from folders can be sped up up to 6x, which I'm more than happy about.

### Approach
* Headers are now being read as a runspace task
* All the relevant commands are trying their best to send the files as an array instead of pipelining them one by one
* Runspace is configured to run max 10 threads due to SQL Server behavior: apparently it has an internal queue and allows only ~6 tasks to run simultaneously anyways (still researching on this one), no reason to bloat it up any higher
* File access is checked beforehand now - again, to pass all the files as an array
* Implemented grouping - to avoid scanning the same file twice

### Commands to test
I really hope everything would work as expected, as I had to revisit some parts in other commands to ensure that arrays are passed properly everywhere. My manual tests are throwing errors because apparently I have an outdated branch and materials from appveyor repo are returning different numbers now. Let's see if PR test succeeds.
List of modified commands:
Read-DbaBackupHeader
Restore-DbaDatabase
Get-DbaBackupInformation

### Screenshots

Before:

After:

Beautiful progress bar:

### Learning
Surprisingly, 50 threads ran more slowly than 10 threads due to Powershell overheads on top of internal SQL Server queuing of RESTORE HEADERONLY commands - sessions just sit there waiting on PARALLEL_BACKUP_QUEUE - that's the first time I personally see it. It allowed no more than 6 sessions running at the same time.

### DIFF
--- a/functions/Read-DbaBackupHeader.ps1
+++ b/functions/Read-DbaBackupHeader.ps1
@@ -1,7 +1,7 @@
 #ValidationTags#Messaging,FlowControl,Pipeline,CodeStyle#
 
 function Read-DbaBackupHeader {
-    <#
+	<#
         .SYNOPSIS
             Reads and displays detailed information about a SQL Server backup.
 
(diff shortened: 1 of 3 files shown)


Zhunya:
Lately I'm using (for most restore operations) `Get-DbaBackupHistory -Last* | Restore-DbaDatabase -TrustDbBackupHistory` - blazing fast =)
But you did save us lot of time with this change!

Stuart-Moore:
All looks good to me from a restore point of view :+1:

Tagging in Shawn and Chrissie to make sure it fits with the overall way they see the module moving and that there's no obvious issues with the backwards PS support

potatoqualitee:
looks great to me and runspace are backwards compat to 2.0, i believe but for sure 3!

potatoqualitee:
Thanks so much 💃

functions/Read-DbaBackupHeader.ps1
```diff
@@ -101,99 +101,166 @@ function Read-DbaBackupHeader {
     )
 
     begin {
-        $loopCnt = 1
+        Write-Message -Level InternalComment -Message "Starting reading headers"
         try {
             $server = Connect-SqlInstance -SqlInstance $SqlInstance -SqlCredential $SqlCredential
```
Zhunya: This `$server` never used later, especially why its not used instead of `$SqlInstance` at line 120

Zhunya:
* If its possible remove progress, maybe show big increments in progress - say every 10 files, or every 50 files read

functions/Read-DbaBackupHeader.ps1
wsmelton: This is because of the runspace @Zhunya at line 210 you can see the transfer of the variables so the runspace threads are being passed that `$server` object.

nvarscar:
> If its possible remove progress, maybe show big increments in progress - say every 10 files, or every 50 files read

It is possible of course, may I ask why? It basically progresses every 50 ms and calculates progress depending on how many threads have completed since last check. Is it bad to send progress updates that often? If it is I've got no idea 😄 Also, number of files can be lower than 10, which would make the progress bar pretty much non-existent. What I could do though is to adjust Start-Sleep to let's say 500 ms and add one more check to eliminate the very last Sleep when all the threads have finished running.

nvarscar:
> Lately I'm using (for most restore operations) Get-DbaBackupHistory -Last* | Restore-DbaDatabase -TrustDbBackupHistory - blazing fast =)
> But you did save us lot of time with this change!

That's what I'm usually using, but a couple of days ago I had to restore a bunch of databases without access to the source server in a totally isolated network... Had to wait ~4 hours to restore 25 small-ish databases (700+ backup files each)

PR #3526, opened 2018-04-12
better unit tests

zippy1981:
## Type of Change

 - [ ] Pester test is included
 - [X] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
More code coverage for Size.cs and I am using the newer MSTests package so I can take advantage of `DataRowAttribute` to run the same test with multiple inputs.

### Commands to test
Well lets not jinx ourselves and test the crap out of the CopyDb command.

### Learning
NCrunch is way awesome as an alternative to dotcover.


zippy1981:
I presume @FriedrichWeinmann  will merge when he has time to thoughtfully review. 

@potatoqualitee banged out some more tests. I might duplicate some of these tests in pester, with not as much detail. It would be a little more "accessible." On thee other hand it might just lack redundancy.  

I guess its just weird this was broken for like a day, everyone knew where in the C# it was broken and no one figured out how to write a Powershell test that didn't require connecting to a database. 

I'm just thinking "how do we systemically improve?" Maybe the answer is just be more careful, and if shit breaks add tests after the fact. Maybe the answer is do something differently.

niphlod:
to say my piece, I know that the bug was in c# but reaaally didn't have any other answer than "restore to the last working version". To me c# is obscure (and there's just a bunch of people that know the build process enough to make sure dll/whatever gets rebuilt properly). 
That being said, tests in PS weren't catching it because it revealed with a fairly high "number" that we cannot replicate easily in "live" tests. In addition to that, I'm really not a fan of mocking because the time taken to properly mock intricated SMO objects can cover those kind of glitches but won't cover "most" of them, which is why you find so little mocks and so much "live" tests.

Last but not least, once this gets merged, let's make sure tests are run on appveyor's CI for DLLs at every commit, just like it "was".

zippy1981:
@niphlod SMO stuff is hard to mock I agree. The only way around that is encapsulating or FakeItEasy. Both are fragile solutions.

In this case the bug was in Size.cs which doesn't depend on anything., Its pure math and string concating. I was able to replicate the bug in Size.cs that lead to the error in the backup cmdlets  through one line of PowerShell that didn't require a database connection or anything. 

However, I guess you are right, for most of the contributors the C# code is as opaque as inline assembly in a C++ file would be to me. It would be harder for me to write pester tests than C# tests for the C# code. I should have written more tests and we need to get appveyor running them.

niphlod:
it's probably the same line @potatoqualitee added as a regr test and a similar one to the one I proposed while debugging in slack. 
From my point there has been this pseudo-brain:
- I know there's a bug 
- we trusted too much c#, we missed it in pester
- I can easily replicate it in PS (and know how to add a failing test)
- we'll surely add it
- I don't know where/how to fix it

Full blame on me not catching up with c# code, which I _guess_ could be easy if I got the time to introduce my brain into it.
Sadly, the same deal goes for adding other (similar?) tests: as I can't read c# code, I can't foresee where it _could_ fail (read: "it feels like writing tests for the PS operator `-EQ`")

zippy1981:
@potatoqualitee I mean I still have to get the unit test running in appveyor, but I can do that in a separate PR. 

I didn't mess with any of the app code, just build scripts and test code. So yeah lets merge it!

potatoqualitee:
😁

PR #3563, opened 2018-04-25
New Command Get-DbaDbPageInfo fixes #3455 

### LINKED ISSUE #3455: New Function - Get-DbaDbPageCount
## Feature Request
Return page usage information by table (total pages, used pages, unused pages). This command will be used to support the command the gets written in the new command in issue #3450.

### Problem to solve
Get number of pages by database/table.

### Additional information
Name Suggestions: 
- Get-DbaDbPageCount
- Get-DbaDbTblPageCount

Parameters: 
- SqlInstance
- SqlCredential (optional) 
- Database (optional)
- Schema (optional)
- Table (optional)
- ExcludeDatabase (optional)
- IncludeSystemDBs (optional)
- EnableException (optional)

sanderstad:
New command to count pages in a database
Test to test the new command


wsmelton:
I would say it would be best to have the command output the raw data and then we can use ps to aggregate or sum whatever we want. So something like Get-DbaDbPageInfo.

sanderstad:
Thanks, I'll look into that. It shouldn't be that big of a problem. I just have to make sure I get the same content for different versions of SQL Server.

wsmelton:
Well at most this command will only support 2008 and higher so you can set that in the connect call using `-MinimumVersion 10`.

potatoqualitee:
ty!

PR #3569, opened 2018-04-27
Get-DbaWaitResource - Looks up the underlying wait resource

Stuart-Moore:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
To help people decipher the handy wait_resource column in sys.dm_exec_requests, these sort of things:
'KEY: 7:35457594073541168 (de21f92a1572)'
'PAGE: 10:1:9180084'

### Approach
Use regex to split the resource into it's constituent parts, then query the system tables to work out what owns the resources.

on KEY types will look up the actual data being locked using %%lockres%% if `-row` specified. Note, this is the data at the time the comand is run, as we don't have the dbatools time machine yet.

### Commands to test
Examples in the CBH

### Screenshots

image


potatoqualitee:
cool, thank you stuart! Would you be able to make all output uniform? So auto expand that data row and fill in the information it provides and give all columns, no matter the output, standard names? I see col1 and col2 is.. not possible to standardize. perhaps that could be an array in a Detail column?

Stuart-Moore:
col1 and col2 are the names of the columns in my test table. With real
tables it'll use the real column names.
As it pulls everything from the row, we could end with something ugly if
someone has a really wide table

Stuart-Moore:
Have added a fuller example to the CBH of working with the ObjectData ouput. Will try to think of a nicer way of displaying it.

Would we want to keep it in a datatable? or maybe drop it to a hash table?

Stuart-Moore:
expanded the data row object during output, (and just to show it does real column names ;) ):
image

Though about hiding the ObjectData property, but thought it stands out as a nice obvious break between metadata and actual data in the output

potatoqualitee:
ty!

PR #3579, opened 2018-05-02
Get-DbaOrphanUser, fixes #3573

### LINKED ISSUE #3573: Get-DbaOrphanUser only returns a single row per database
## Issue
Get-DbaOrphanUser only returns a single row per database.  So if a single DB has multiple orphan users then it only counts as 1.  Question is whether this should return the full count of orphans per DB by returning @($results.user).count instead of @($results).count

- Checked latest version of Database.Tests.ps1 file has the same return value.
- Manually ran the command to show that the object can produce the exact number of orphan users per DB

### Version Information
Although code reacts the same on multiple versions of SQL Server
 - Windows 10
 - PowerShell Version: 5.1
 - SQL Server SQL Server 2008

### Steps to Reproduce
<--  
PS C:\Users\userID> $results = Get-DbaOrphanUser -SqlInstance tst-sqlserver.domain.com

PS C:\Users\userID> $results

SqlInstance              DatabaseName   User                            
-----------              ------------   ----                            
tst-sqlserver.domain.com DB1        {user1, user2, user3, user4...}
tst-sqlserver.domain.com DB2          user1
tst-sqlserver.domain.com DB3        {user1, user2, user3, user4...}

PS C:\Users\userID> @($results).count
3

PS C:\Users\userID> @($results.user).count
32

## Feature Request
I would like it to return the full number of orphan users rather than just the number of databases with orphan users

niphlod:
## Type of Change
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Main thing was to get rid of the unusual behaviour described in #3573 but since functions needed to be adapted no matter what, this includes more fixes.
Let this be a review of all code relating to orphan users (hopefully it won't take weeks though ^_^). Code was "old-ish" and get/remove were actually fast copy-paste from the original repair-. Now everything is more "1.0" ready but we're not there yet completely. 

### Approach
Fixed a few things, removed code paths never executed, added tests stubs
Standardized:
  - params
  - output properties

### Tests
Added proper unittest without relying on the database with just one orphan user in appveyorlabrepo. That is probably the reason behind slipping in a change in behaviour (when multiple orphans are found within the same database). Now the main behaviour is set, and less experienced users/lazier can chime in adding tests for uncovered code branches.

### Discuss
Naming conventions (see #3573): once everyone is on board, I'd feel safer to merge this and then approach just the rename in another PR

Remove-DbaOrphanUser: 
  - doesn't return anything (like Remove-Item) but I'm not sure if that's something dbatools wants. We *may* output the usual properties plus status = 'Dropped' easily
  - Not sure if the output properties should change: in fact, when a schema or owner needs dropping/altering, remove-dbaorphanuser DOES output something
 - again, when everybody is on board, I'll create the relevant test suite

### DIFF
--- a/functions/Get-DbaOrphanUser.ps1
+++ b/functions/Get-DbaOrphanUser.ps1
@@ -1,41 +1,59 @@
+#ValidationTags#Messaging,FlowControl,Pipeline,CodeStyle#
 function Get-DbaOrphanUser {
     <#
         .SYNOPSIS
             Get orphaned users.
+
         .DESCRIPTION
             An orphan user is defined by a user that does not have their matching login. (Login property = "").
+
         .PARAMETER SqlInstance
             The SQL Server Instance to connect to.
+
         .PARAMETER SqlCredential
             Login to the target instance using alternative credentials. Windows and SQL Authentication supported. Accepts credential objects (Get-Credential)
+
         .PARAMETER Database
             Specifies the database(s) to process. Options for this list are auto-populated from the server. If unspecified, all databases will be processed.
+
+        .PARAMETER ExcludeDatabase
+            Specifies the database(s) to exclude from processing. Options for this list are auto-populated from the server
+
         .PARAMETER WhatIf
             If this switch is enabled, no actions are performed but informational messages will be displayed that explain what would happen if the command were to run.
+
         .PARAMETER Confirm
             If this switch is enabled, you will be prompted for confirmation before executing any operations that change state.
+
         .PARAMETER EnableException
             By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
             This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
             Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
+
         .NOTES
             Tags: Orphan, Databases
             Author: Claudio Silva (@ClaudioESSilva)
             Author: Garry Bargsley (@gbargsley)
+            Editor: Simone Bizzotto (@niphlod)
             Website: https://dbatools.io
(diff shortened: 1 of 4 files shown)


niphlod:
@gbargsley : sure you fetched the correct branch ? that's output from current development

gbargsley:
You are right.  I read it as the development branch.  Let me try again.

gbargsley:
How about this.
image

niphlod:
that'd point to the right output ^_^

potatoqualitee:
Thanks much, Niph 💞 

for remove: "doesn't return anything (like Remove-Item) but I'm not sure if that's something dbatools wants. We may output the usual properties plus status = 'Dropped' easily"

Yes please - something similar to Remove-DbaDatabase

niphlod:
I'd go also as far as saying that we have a bit of a pickle to sort out with remove-dbaorphanuser: schema ownership / -Force right now return a pscustomobject to signal changes on schema. Should we just do it without outputting anything and consider actions taken as "required" by the user calling remove-dbaorphanuser or what ?

potatoqualitee:
I'll take a c loser look this weekend, sorry for the delay. btw, anything that is "Collection" i've been renaming to InputObject. Took me forever to figure that out 😂

niphlod:
that's not an input object, it's an internal variable holding the list of targeted databases. no problems though, if we wanna strike out "collection", but I'd go for inputobject only if it's an input nonetheless.

potatoqualitee:
> Should we just do it without outputting anything and consider actions taken as "required" 

You mean should we perform the action and not output a Note field letting the user know? btw, since this is Get, I'm gonna merge and we can do Remove next up.

Issue #3548, opened 2018-04-18
New Command: Get-DbaResourceGovernorClassiferFunction


suxstellino:
## Feature request
Provide the ability to get the Resource Governor custom classifier function, if exists. 

As described in issue #3528 (thanks to @wsmelton), `Get-DbaResourceGovernorClassiferFunction` would be the command, then the copy command would call that [...] and all of that code can just be put in RG copy command.

I see:
- [x] add `Get-DbaResourceGovernorClassiferFunction`
- [x] check rules/styles of the new command
- [x] document the new command
- [ ] add unit tests for the new command
- [x] try the `Get-Help`for the new command
- [x] change `Copy-DbaResourceGovernor`
- [ ] ~edit integration tests after the change on the copy command~

### Additional info
When an environment needs to balance custom workloads, it's possible to setup the Resource Governor with a custom classifier function which dispatches the sessions toward certain workload groups. 

### Problem to solve
The `Copy-DbaResourceGovernor` command is not considering the optional custom classifier function (a dependency object) since is copying the default workload group based architecture. The command could be improved getting the definition of that function, if exists, in order to copy between instances also the custom object which the Resource Governor is dependent on when defining more than the default workload group.

suxstellino:
I've a set of questions about this issue I'm working on:

**scenario: source server has a classifier function and destination hasn't it**
If -Force parameter is true, I will drop and create _A function_, which potentially have the same name of the source classifier one, then I'll rebind to the resource governor. Is that right in your opinion?
If -Force is false, what do you suggest to do? I'd like to "do anything", in order to avoid any change/risk with names.

**scenario: source server hasn't a classifier function and destination has it**
How to behave in case of -Force true/false? 

**scenario: both servers has a classifier function**
How to behave in case of -Force true/false?

any suggestion will be appreciated

niphlod:
source has, dest has, no force --> warn and exit
source has, dest has, force --> overwrite dest
source has not, dest has --> nothing to copy ;-) 
source has, dest has not --> copy irregardless of force

suxstellino:
In 2 and 4 (when I've to create or edit the dest classifier function) another dest function can have the same name of the source function (the one to copy). 
Indeed, in 2, I've to set to NULL che classifier function property of the resource governor in order to create or alter (if already exists) the dest function and in 4, I have to create or alter (if already exists) the function based on the source name. So I'm concerned about dropping or altering a previously created function, which could be unrelated to resource governor.
[...]

PR #3580, opened 2018-05-03
New Command - Get-DbaResourceGovernorClassifierFunction

suxstellino:
## Type of Change

 - [ ] Bug fix (non-breaking change, fixes #)
 - [x] New feature (`Get-DbaResourceGovernorClassifierFunction` #3548)
 - [x] Breaking change (on `Copy-DbaResourceGovernor` #3548)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [x] Documentation
 - [ ] Build system
 
### Purpose
Managing the status of the resource governor (ignored before this change) and also the optional classifier function related to the resource governor itself. In order to get the classifier function, a new cmdlet has been added (`Get-DbaResourceGovernorClassifierFunction`) as described on the issue #3548.

### Approach
 - Added a check of the status at the end of `Copy-DbaResourceGovernor` cmdlet 
 - Added a new command `Get-DbaResourceGovernorClassifierFunction`
 - Added the logic for managing the classifier function when updating and reconfiguring resource governor

### Commands to test
See help. 
About unit tests, no implementation has been made as described in issue #3548.


potatoqualitee:
This looks cool, thanks so much! I'll take a closer look this weekend. Sorry about the test failures - I've modified the failing test for now and will update your branch.

.gitignore
```diff
@@ -20,3 +20,4 @@ msbuild.log
 # Local constant file
 tests/constants.local.ps1
 
+functions/Copy-DbaResourceGovernor_TEST.ps1
```
wsmelton: @suxstellino you need to remove this update from the PR

functions/Copy-DbaResourceGovernor.ps1
```diff
@@ -89,13 +89,17 @@ function Copy-DbaResourceGovernor {
         $sourceServer = Connect-SqlInstance -SqlInstance $Source -SqlCredential $SourceSqlCredential
         $destServer = Connect-SqlInstance -SqlInstance $Destination -SqlCredential $DestinationSqlCredential
 
-        $source = $sourceServer.DomainInstanceName
-        $destination = $destServer.DomainInstanceName
-
         if ($sourceServer.VersionMajor -lt 10 -or $destServer.VersionMajor -lt 10) {
             Stop-Function -Message "Resource Governor is only supported in SQL Server 2008 and above. Quitting."
```
wsmelton: This can actually be removed if you want. Update the `Connect-SqlInstance` calls to include `-MinimumVersion 10`. This handles validating the version of each instance more gracefully.

functions/Copy-DbaResourceGovernor.ps1
```diff
@@ -109,26 +113,93 @@ function Copy-DbaResourceGovernor {
...
+                        $copyResourceGovClassifierFunc | Select-DefaultView -Property DateTime, SourceServer, DestinationServer, Name, Type, Status, Notes -TypeName MigrationObject
+                    }
+                    else {
+                        $fullyQualifiedFunctionName = $sourceClassifierFunction.Schema+"."+$sourceClassifierFunction.Name
+                                  
+                        if (!$destClassifierFunction) {
+                            $destFunction = $destServer.Databases["master"].UserDefinedFunctions[$sourceClassifierFunction.Name]
+
+                            if ($destFunction) {
+                                Write-Message -Level Verbose -Message "Dropping the function with the source classifier function name."
```
wsmelton: Anything where you are modifying the object (or dropping it) needs to be wrapped in a ShouldProcess block. Not sure if the other copy commands do one blanket block but since we are adding the features for pulling the functions it would be good to add the whatif support as well.

functions/Copy-DbaResourceGovernor.ps1
```diff
@@ -109,26 +113,93 @@ function Copy-DbaResourceGovernor {
...
                 Write-Message -Level Verbose -Message "The resource governor is not available in this edition of SQL Server. You can manipulate resource governor metadata but you will not be able to apply resource governor configuration. Only Enterprise edition of SQL Server supports resource governor."
             }
             else {
                 try {
-                    $sql = $sourceServer.ResourceGovernor.Script() | Out-String
-                    Write-Message -Level Debug -Message $sql
-                    Write-Message -Level Verbose -Message "Updating Resource Governor settings."
-                    $destServer.Query($sql)
+                    Write-Message -Level Verbose -Message "Managing classifier function."
+                    if (!$sourceClassifierFunction) {
```
wsmelton: Do you have a potential of ever having more than one classifier function?

If so this should be adjusted to handle multiple being found.

functions/Get-DbaResourceGovernorClassiferFunction.ps1
```diff
@@ -0,0 +1,88 @@
...
+Gets the classifier function object on Sql1 and Sql2/sqlexpress instances
+
+#>
+    [CmdletBinding()]
+    param (
+        [parameter(Mandatory, ValueFromPipeline)]
+        [Alias("ServerInstance", "SqlServer")]
+        [DbaInstanceParameter[]]$SqlInstance,
+        [PSCredential]$SqlCredential,
+        [Alias('Silent')]
```
wsmelton: You can remove the alias, that is added for current/older commands. New ones do not have to comply with having that added.

functions/Copy-DbaResourceGovernor.ps1
```diff
@@ -109,26 +113,93 @@ function Copy-DbaResourceGovernor {
...
-                    $copyResourceGovSetting.Status = "Successful"
-                    $copyResourceGovSetting | Select-DefaultView -Property DateTime, SourceServer, DestinationServer, Name, Type, Status, Notes -TypeName MigrationObject
+                            Write-Message -Level Verbose -Message "Creating function."
+                            $destServer.Query($sourceClassifierFunction.Script())
+                            
+                            $sql = "ALTER RESOURCE GOVERNOR WITH (CLASSIFIER_FUNCTION = $fullyQualifiedFunctionName);"
+                            $sql
+                            Write-Message -Level Debug -Message $sql
+                            Write-Message -Level Verbose -Message "Mapping Resource Governor classifier function."
+                            $destServer.Query($sql)
```
wsmelton: The calls to execute a query really should be wrapped in a try/catch block so a proper error can be returned when one of them fails and the user is aware of what area failed.

functions/Get-DbaResourceGovernorClassiferFunction.ps1
wsmelton: Can we also add a filter for the function? Standard parameter names we use would be `Classifier` and then a `ExcludeClassifier`...don't think we have to make it long like `ClassifierFunction` do you @potatoqualitee ?

functions/Copy-DbaResourceGovernor.ps1
suxstellino: The resource governor can have only one classifier function bound.

functions/Get-DbaResourceGovernorClassiferFunction.ps1
suxstellino: What this filter should do @wsmelton?

suxstellino: done, pushing

functions/Copy-DbaResourceGovernor.ps1
suxstellino: done, pushing

.gitignore
suxstellino: done, pushing

functions/Copy-DbaResourceGovernor.ps1
suxstellino: so you're suggesting me to indent the try/catch inside the outer one? I've copied the previous version which did a single block. Would you change fragmenting the commands into different try catch blocks?

wsmelton: If any one of those queries fails does the user need to know which one?

suxstellino: I think that the IF part and the ELSE part will get just a query to know for in case of error. Queries like `ALTER RESOURCE GOVERNOR WITH (CLASSIFIER_FUNCTION = NULL);`  and `ALTER RESOURCE GOVERNOR RECONFIGURE;` will not fail (error could be connection). Just the final query (in the IF and ELSE parts) needs to be checked.

wsmelton: Those queries can fail if permissions do not exist.

suxstellino: I'd like to know if I've to change the original code (the one before my changes) splitting in more than an outer try/catch block or indeting two more inside it.

potatoqualitee:
gonna merge and update, thank you very much!

potatoqualitee:
And my apologies for the delay

PR #3595, opened 2018-05-09
Invoke-DbaSqlQuery, enhance messaging

niphlod:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Tinker easily with queries where messages (PRINT and RAISERROR) are as important as the result
With stream redirection it has almost been possible to do this but it's always kinda be a PITA. This has been discussed in slack for a while, sorry for the delay

### Approach
Basically stripped down the code we need from invoke-sqlcmd2 and adapted to use dbatools' facilities, coupled with async execution and a good tip from @FriedrichWeinmann. It's safe to assume we can now move invoke-sqlcmd2 to the optional folder, as this, coupled with connect-dbainstance to establish "funny" connections, has all the needed features.

### Commands to test
Added basic tests to make sure output is not buffered (as it is in invoke-sqlcmd and sqlcmd -ge 2012). Think reading the output of ola's maintenance procedures while it is outputted without waiting for the procedure to end.

### DIFF
--- a/functions/Invoke-DbaSqlQuery.ps1
+++ b/functions/Invoke-DbaSqlQuery.ps1
@@ -5,7 +5,7 @@ function Invoke-DbaSqlQuery {
             A command to run explicit T-SQL commands or files.
 
         .DESCRIPTION
-            This function is a wrapper command around Invoke-SqlCmd2.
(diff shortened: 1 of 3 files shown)


potatoqualitee:
oh la la! ty will look when i have time and a brain.

PR #3599, opened 2018-05-09
Database Specific Improvements For Invoke-DbaDiagnosticQuery

sheldonhull:
### fixes/improvements
- Progress ID uses random id instead of hardcoded value, and ensures marked as completed when done to avoid orphaned progress bar. Additionally avoids progress bar conflict with other calling scripts due to hardcoded id. 
- Aliased DatabaseName with database for consistency with other dbatools magic.
- Aliased SQLCredential with Credential for consistency with other dbatools magic.

### new Features
 - [x] New feature Database Specific: Added database specific filtering to allow database specific to provide databasename which then will filter the database level queries to only run against provided [string], or [string[]]. 
 - [x] New feature ExportQueries: Added export functionality to allow user to preview all the queries that would be run on the matching version of SQL Server and output to disk. Did not include any parameter sets at this time as didn't see them currently in use. Can enhance further with that if requested.

### other support changes and improvements
- Output inserted explicitly into Collections.Generic.List[pscustomobject] for improved performance over array += syntax and to define explicitly the results to return. This is to improve the handling for returning the parsed queries as returned [pscustomobject[]] to allow further manipulation or exporting if desired.
- Added documentation examples of how to use
- Added Pester Tests for all my various usage cases and verified passed... #workedonmymachine

## Purpose
improve diagnostic query functionality.
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [x] Documentation
 - [ ] Build system

### Approach
adds awesome stuff (i think, YMMV) to query specific database, export queries instead of running them, and supporting tests that #workedonmymachine. First time contributing so hoping this helps some others!

### Commands to test
in help

### DIFF
--- a/functions/Invoke-DbaDiagnosticQuery.ps1
+++ b/functions/Invoke-DbaDiagnosticQuery.ps1
@@ -149,10 +149,12 @@ function Invoke-DbaDiagnosticQuery {
         [Switch]$NoQueryTextColumn,
         [Switch]$NoPlanColumn,
         [Switch]$NoColumnParsing,
-        [switch][Alias('Silent')]
-        [switch]$EnableException,
+
         [string]$OutputPath,
-        [switch]$ExportQueries
+        [switch]$ExportQueries,
+
+        [switch][Alias('Silent')]
+        [switch]$EnableException
     )
 
     begin {
--- a/functions/Invoke-DbaDiagnosticQuery.ps1
+++ b/functions/Invoke-DbaDiagnosticQuery.ps1
@@ -21,7 +21,7 @@ function Invoke-DbaDiagnosticQuery {
     .PARAMETER Path
     Alternate path for the diagnostic scripts
 
-    .PARAMETER DatabaseName
+    .PARAMETER Database
     The database(s) to process. If unspecified, all databases will be processed
 
     .PARAMETER ExcludeDatabase
--- a/functions/Invoke-DbaDiagnosticQuery.ps1
+++ b/functions/Invoke-DbaDiagnosticQuery.ps1
@@ -135,7 +135,9 @@ function Invoke-DbaDiagnosticQuery {
         [DbaInstanceParameter[]]$SqlInstance,
 
         [Alias('Database')]
-        [string[]]$DatabaseName,
+        [object[]]$DatabaseName,
+
+        [object[]]$ExcludeDatabase,
 
         [Alias('Credential')]
         [PSCredential]$SqlCredential,
@@ -243,12 +245,10 @@ function Invoke-DbaDiagnosticQuery {
 
             if (!$instanceOnly) {
                 if (-not $DatabaseName) {
-                    $databases = (Get-DbaDatabase -SqlInstance $server -ExcludeAllSystemDb).Name
-                    Write-Message -Level Debug -Message  "No Database Name Filtered: Pulled $(@($Databases).count)" -Verbose
+                    $databases = (Get-DbaDatabase -SqlInstance $server -ExcludeAllSystemDb -ExcludeDatabase $ExcludeDatabase).Name
                 }
                 else {
-                    $databases = (Get-DbaDatabase -SqlInstance $server -ExcludeAllSystemDb -Database $DatabaseName).Name
-                    Write-Message -Level Debug -Message  "Filtered to databases: $(($Databases.Name) -join ',')" -Verbose
+                    $databases = (Get-DbaDatabase -SqlInstance $server -ExcludeAllSystemDb -Database $DatabaseName -ExcludeDatabase $ExcludeDatabase).Name
                 }
             }
 
@@ -260,31 +260,31 @@ function Invoke-DbaDiagnosticQuery {
                 $first = $false
             }
             #since some database level queries can take longer (such as fragmentation) calculate progress with database specific queries * count of databases to run against into context
-            $CountOfDatabases = ($databases).count
+            $CountOfDatabases = ($databases).Count
 
 
             if ($QueryName.Count -ne 0) {
                 #if running all queries, then calculate total to run by instance queries count + (db specific count * databases to run each against)
-                $CountOfDatabaseSpecific = @($parsedscript | Where-Object {$_.QueryName -in $QueryName -and $_.DBSpecific -eq $true}).count
-                $CountOfInstanceSpecific = @($parsedscript | Where-Object {$_.QueryName -in $QueryName -and $_.DBSpecific -eq $false}).count
+                $countDBSpecific = @($parsedscript | Where-Object {$_.QueryName -in $QueryName -and $_.DBSpecific -eq $true}).Count
+                $countInstanceSpecific = @($parsedscript | Where-Object {$_.QueryName -in $QueryName -and $_.DBSpecific -eq $false}).Count
             }
             else {
                 #if narrowing queries to database specific, calculate total to process based on instance queries count + (db specific count * databases to run each against)
-                $CountOfDatabaseSpecific = @($parsedscript | Where-Object DBSpecific).count
-                $CountOfInstanceSpecific = @($parsedscript | Where-Object DBSpecific -eq $false).count
+                $countDBSpecific = @($parsedscript | Where-Object DBSpecific).Count
+                $countInstanceSpecific = @($parsedscript | Where-Object DBSpecific -eq $false).Count
 
             }
             if (!$instanceonly -and !$DatabaseSpecific -and !$QueryName) {
-                $scriptcount = $CountOfInstanceSpecific + ($CountOfDatabaseSpecific * $CountOfDatabases )
+                $scriptcount = $countInstanceSpecific + ($countDBSpecific * $CountOfDatabases )
             }
             elseif ($instanceOnly) {
-                $scriptcount = $CountOfInstanceSpecific
+                $scriptcount = $countInstanceSpecific
             }
             elseif ($DatabaseSpecific) {
-                $scriptcount = $CountOfDatabaseSpecific * $CountOfDatabases
+                $scriptcount = $countDBSpecific * $CountOfDatabases
             }
             elseif ($QueryName.Count -ne 0) {
-                $scriptcount = $CountOfInstanceSpecific + ($CountOfDatabaseSpecific * $CountOfDatabases )
+                $scriptcount = $countInstanceSpecific + ($countDBSpecific * $CountOfDatabases )
 
 
             }
@@ -342,7 +342,7 @@ function Invoke-DbaDiagnosticQuery {
                                 DatabaseSpecific = $scriptpart.DBSpecific
                                 DatabaseName     = $null
                                 Notes            = $null
-                                Result           = $result | Select-Object * -ExcludeProperty RowError, RowState, Table, ItemArray, HasErrors
+                                Result           = Select-DefaultView -InputObject $result -Property *
                             }
 
                         }
@@ -418,7 +418,7 @@ function Invoke-DbaDiagnosticQuery {
                                 DatabaseSpecific = $scriptpart.DBSpecific
                                 DatabaseName     = $currentDb
                                 Notes            = $null
-                                Result           = $result | Select-Object * -ExcludeProperty RowError, RowState, Table, ItemArray, HasErrors
+                                Result           = Select-DefaultView -InputObject $result -Property *
                             }
                         }
                         else {
--- a/functions/Invoke-DbaDiagnosticQuery.ps1
+++ b/functions/Invoke-DbaDiagnosticQuery.ps1
@@ -372,7 +372,7 @@ function Invoke-DbaDiagnosticQuery {
                     foreach ($currentdb in $databases) {
                         if ($ExportQueries) {
                             $null = New-Item -Path $OutputPath -ItemType Directory -Force
-                            $FileName = Remove-InvalidFileNameChars ('{0}-{1}-{2}.sql' -f $server.DomainInstanceName, $currentDb.Name, $Scriptpart.QueryName)
+                            $FileName = Remove-InvalidFileNameChars ('{0}-{1}-{2}.sql' -f $server.DomainInstanceName, $currentDb, $Scriptpart.QueryName)
                             $FullName = Join-Path $OutputPath $FileName
                             Write-Message -Level Verbose -Message  "Creating file: $FullName"
                             $scriptPart.Text | out-file -FilePath $FullName -encoding UTF8 -force
@@ -380,16 +380,16 @@ function Invoke-DbaDiagnosticQuery {
                         }
 
 
-                        if ($PSCmdlet.ShouldProcess(('{0} ({1})' -f $instance, $currentDb.Name), $scriptpart.QueryName)) {
+                        if ($PSCmdlet.ShouldProcess(('{0} ({1})' -f $instance, $currentDb), $scriptpart.QueryName)) {
 
                             if (-not $EnableException) {
                                 $Counter++
-                                Write-Progress -Id $ProgressId -ParentId 0 -Activity "Collecting diagnostic query data from $($currentDb.Name) on $instance" -Status ('Processing {0} of {1}' -f $counter, $scriptcount) -CurrentOperation $scriptpart.QueryName -PercentComplete (($Counter / $scriptcount) * 100)
+                                Write-Progress -Id $ProgressId -ParentId 0 -Activity "Collecting diagnostic query data from $($currentDb) on $instance" -Status ('Processing {0} of {1}' -f $counter, $scriptcount) -CurrentOperation $scriptpart.QueryName -PercentComplete (($Counter / $scriptcount) * 100)
                             }
 
-                            Write-Message -Level Verbose -Message "Collecting diagnostic query data from $($currentDb.Name) for $($scriptpart.QueryName) on $instance"
+                            Write-Message -Level Verbose -Message "Collecting diagnostic query data from $($currentDb) for $($scriptpart.QueryName) on $instance"
                             try {
-                                $result = $server.Query($scriptpart.Text, $currentDb.Name)
+                                $result = $server.Query($scriptpart.Text, $currentDb)
                                 if (!$result) {
                                     [pscustomobject]@{
                                         ComputerName     = $server.NetName
@@ -407,7 +407,7 @@ function Invoke-DbaDiagnosticQuery {
                                 }
                             }
                             catch {
-                                Write-Message -Level Verbose -Message ('Some error has occured on Server: {0} - Script: {1} - Database: {2}, result will not be saved' -f $instance, $scriptpart.QueryName, $currentDb.Name) -Target $currentdb -ErrorRecord $_
+                                Write-Message -Level Verbose -Message ('Some error has occured on Server: {0} - Script: {1} - Database: {2}, result will not be saved' -f $instance, $scriptpart.QueryName, $currentDb) -Target $currentdb -ErrorRecord $_
                             }
 
                             [pscustomobject]@{
@@ -418,7 +418,7 @@ function Invoke-DbaDiagnosticQuery {
                                 Name             = $scriptpart.QueryName
                                 Description      = $scriptpart.Description
                                 DatabaseSpecific = $scriptpart.DBSpecific
-                                Database         = $currentDb.Name
+                                Database         = $currentDb
                                 Notes            = $null
                                 Result           = Select-DefaultView -InputObject $result -Property *
                             }
--- a/functions/Invoke-DbaDiagnosticQuery.ps1
+++ b/functions/Invoke-DbaDiagnosticQuery.ps1
@@ -1,425 +1,425 @@
 function Invoke-DbaDiagnosticQuery {
-	<#
-	.SYNOPSIS
-	Invoke-DbaDiagnosticQuery runs the scripts provided by Glenn Berry's DMV scripts on specified servers
+    <#
+    .SYNOPSIS
+    Invoke-DbaDiagnosticQuery runs the scripts provided by Glenn Berry's DMV scripts on specified servers
 
-	.DESCRIPTION
-	This is the main function of the Sql Server Diagnostic Queries related functions in dbatools.
-	The diagnostic queries are developed and maintained by Glenn Berry and they can be found here along with a lot of documentation:
-	http://www.sqlskills.com/blogs/glenn/category/dmv-queries/
+    .DESCRIPTION
+    This is the main function of the Sql Server Diagnostic Queries related functions in dbatools.
+    The diagnostic queries are developed and maintained by Glenn Berry and they can be found here along with a lot of documentation:
+    http://www.sqlskills.com/blogs/glenn/category/dmv-queries/
 
-	The most recent version of the diagnostic queries are included in the dbatools module.
-	But it is possible to download a newer set or a specific version to an alternative location and parse and run those scripts.
-	It will run all or a selection of those scripts on one or multiple servers and return the result as a PowerShell Object
+    The most recent version of the diagnostic queries are included in the dbatools module.
+    But it is possible to download a newer set or a specific version to an alternative location and parse and run those scripts.
+    It will run all or a selection of those scripts on one or multiple servers and return the result as a PowerShell Object
 
-	.PARAMETER SqlInstance
-	The target SQL Server. Can be either a string or SMO server
+    .PARAMETER SqlInstance
+    The target SQL Server. Can be either a string or SMO server
 
-	.PARAMETER SqlCredential
-	Allows alternative Windows or SQL login credentials to be used
+    .PARAMETER SqlCredential
+    Allows alternative Windows or SQL login credentials to be used
 
-	.PARAMETER Path
-	Alternate path for the diagnostic scripts
+    .PARAMETER Path
+    Alternate path for the diagnostic scripts
 
-	.PARAMETER DatabaseName
-	The database(s) to process. If unspecified, all databases will be processed
+    .PARAMETER DatabaseName
+    The database(s) to process. If unspecified, all databases will be processed
 
-	.PARAMETER ExcludeDatabase
-	The database(s) to exclude
+    .PARAMETER ExcludeDatabase
+    The database(s) to exclude
 
-	.PARAMETER UseSelectionHelper
-	Provides a gridview with all the queries to choose from and will run the selection made by the user on the Sql Server instance specified.
+    .PARAMETER UseSelectionHelper
+    Provides a gridview with all the queries to choose from and will run the selection made by the user on the Sql Server instance specified.
 
-	.PARAMETER QueryName
-	Only run specific query
+    .PARAMETER QueryName
+    Only run specific query
 
-	.PARAMETER InstanceOnly
-	Run only instance level queries
+    .PARAMETER InstanceOnly
+    Run only instance level queries
 
-	.PARAMETER DatabaseSpecific
-	Run only database level queries
+    .PARAMETER DatabaseSpecific
+    Run only database level queries
 
-	.PARAMETER NoQueryTextColumn
-	Use this switch to exclude the [Complete Query Text] column from relevant queries
+    .PARAMETER NoQueryTextColumn
+    Use this switch to exclude the [Complete Query Text] column from relevant queries
 
-	.PARAMETER NoPlanColumn
-	Use this switch to exclude the [Query Plan] column from relevant queries
+    .PARAMETER NoPlanColumn
+    Use this switch to exclude the [Query Plan] column from relevant queries
 
-	.PARAMETER NoColumnParsing
-	Does not parse the [Complete Query Text] and [Query Plan] columns and disregards the NoQueryTextColumn and NoColumnParsing switches
+    .PARAMETER NoColumnParsing
+    Does not parse the [Complete Query Text] and [Query Plan] columns and disregards the NoQueryTextColumn and NoColumnParsing switches
 
-	.PARAMETER EnableException
-	By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
-	This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
-	Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
+    .PARAMETER EnableException
+    By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
+    This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
+    Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
 
-	.PARAMETER Confirm
-	Prompts to confirm certain actions
+    .PARAMETER Confirm
+    Prompts to confirm certain actions
 
-	.PARAMETER WhatIf
-	Shows what would happen if the command would execute, but does not actually perform the command
+    .PARAMETER WhatIf
+    Shows what would happen if the command would execute, but does not actually perform the command
 
-	.PARAMETER OutputPath
-	Directory to parsed diagnostict queries to. This will split them based on server, databasename, and query.
+    .PARAMETER OutputPath
+    Directory to parsed diagnostict queries to. This will split them based on server, databasename, and query.
 
-	.PARAMETER ExportQueries
-	Use this switch to export the diagnostic queries to sql files. I
-	nstead of running the queries, the server will be evaluated to find the appropriate queries to run based on SQL Version.
-	These sql files will then be created in the OutputDirectory.
+    .PARAMETER ExportQueries
+    Use this switch to export the diagnostic queries to sql files. I
+    nstead of running the queries, the server will be evaluated to find the appropriate queries to run based on SQL Version.
+    These sql files will then be created in the OutputDirectory.
 
 
 
-	.NOTES
-	Tags: Database, DMV
-	Author: André Kamman (@AndreKamman), http://clouddba.io
-	Website: https://dbatools.io
-	Copyright: (C) Chrissy LeMaire, clemaire@gmail.com
-	License: MIT https://opensource.org/licenses/MIT
+    .NOTES
+    Tags: Database, DMV
+    Author: André Kamman (@AndreKamman), http://clouddba.io
+    Website: https://dbatools.io
+    Copyright: (C) Chrissy LeMaire, clemaire@gmail.com
+    License: MIT https://opensource.org/licenses/MIT
 
-	.LINK
-	https://dbatools.io/Invoke-DbaDiagnosticQuery
+    .LINK
+    https://dbatools.io/Invoke-DbaDiagnosticQuery
 
-	.EXAMPLE
-	Invoke-DbaDiagnosticQuery -SqlInstance sql2016
+    .EXAMPLE
+    Invoke-DbaDiagnosticQuery -SqlInstance sql2016
 
-	Run the selection made by the user on the Sql Server instance specified.
+    Run the selection made by the user on the Sql Server instance specified.
 
-	.EXAMPLE
-	Invoke-DbaDiagnosticQuery -SqlInstance sql2016 -UseSelectionHelper | Export-DbaDiagnosticQuery -Path C:\temp\gboutput
+    .EXAMPLE
+    Invoke-DbaDiagnosticQuery -SqlInstance sql2016 -UseSelectionHelper | Export-DbaDiagnosticQuery -Path C:\temp\gboutput
 
-	Provides a gridview with all the queries to choose from and will run the selection made by the user on the SQL Server instance specified.
-	Then it will export the results to Export-DbaDiagnosticQuery.
+    Provides a gridview with all the queries to choose from and will run the selection made by the user on the SQL Server instance specified.
+    Then it will export the results to Export-DbaDiagnosticQuery.
 
-	.Example
-	# Exporting Queries To SQL Files
-	- Parse the appropriate diagnostic queries by connecting to server to get version matched queries
-	- Instead of running the diagnostic queries, export them to SQL files
+    .Example
+    # Exporting Queries To SQL Files
+    - Parse the appropriate diagnostic queries by connecting to server to get version matched queries
+    - Instead of running the diagnostic queries, export them to SQL files
 
-	# Export All Queries to Disk
-	Invoke-DbaDiagnosticQuery -sqlinstance localhost -ExportQueries -outputpath "C:\temp\DiagnosticQueries"
+    # Export All Queries to Disk
+    Invoke-DbaDiagnosticQuery -sqlinstance localhost -ExportQueries -outputpath "C:\temp\DiagnosticQueries"
 
-	# Export Database Specific Queries for all User Dbs
-	Invoke-DbaDiagnosticQuery -sqlinstance localhost -DatabaseSpecific -DatabaseName 'tempdb' -ExportQueries -outputpath "C:\temp\DiagnosticQueries"
+    # Export Database Specific Queries for all User Dbs
+    Invoke-DbaDiagnosticQuery -sqlinstance localhost -DatabaseSpecific -DatabaseName 'tempdb' -ExportQueries -outputpath "C:\temp\DiagnosticQueries"
 
-	# Export Database Specific Queries For One Target Database
-	Invoke-DbaDiagnosticQuery -sqlinstance localhost -DatabaseSpecific -DatabaseName 'tempdb' -ExportQueries -outputpath "C:\temp\DiagnosticQueries"
+    # Export Database Specific Queries For One Target Database
+    Invoke-DbaDiagnosticQuery -sqlinstance localhost -DatabaseSpecific -DatabaseName 'tempdb' -ExportQueries -outputpath "C:\temp\DiagnosticQueries"
 
-	# Export Database Specific Queries For One Target Database and One Specific Query
-	Invoke-DbaDiagnosticQuery -sqlinstance localhost -DatabaseSpecific -DatabaseName 'tempdb' -ExportQueries -outputpath "C:\temp\DiagnosticQueries" -queryname 'Database-scoped Configurations'
+    # Export Database Specific Queries For One Target Database and One Specific Query
+    Invoke-DbaDiagnosticQuery -sqlinstance localhost -DatabaseSpecific -DatabaseName 'tempdb' -ExportQueries -outputpath "C:\temp\DiagnosticQueries" -queryname 'Database-scoped Configurations'
 
-	# Choose Queries To Export
-	Invoke-DbaDiagnosticQuery -sqlinstance localhost -UseSelectionHelper
+    # Choose Queries To Export
+    Invoke-DbaDiagnosticQuery -sqlinstance localhost -UseSelectionHelper
 
-	This will export with SqlInstance, DatabaseName, and QueryName as appropriate based on query.
+    This will export with SqlInstance, DatabaseName, and QueryName as appropriate based on query.
 
-	.Example
-	# Export Queries (not run) as returned object
+    .Example
+    # Export Queries (not run) as returned object
 
-	[System.Management.Automation.PSObject[]]$results = Invoke-DbaDiagnosticQuery -SqlInstance localhost -whatif
+    [System.Management.Automation.PSObject[]]$results = Invoke-DbaDiagnosticQuery -SqlInstance localhost -whatif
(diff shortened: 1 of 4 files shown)


functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -79,235 +88,341 @@ Run the selection made by the user on the Sql Server instance specified.
...
+					11 { "2012" }
+					12 { "2014" }
+					13 { "2016" }
+					14 { "2017" }
+				}
+			}
+
+			if (!$instanceOnly) {
+				if (@($DatabaseName).count -eq 0) {
+					$databases = $server.Query("Select Name from sys.databases where name not in ('master', 'model', 'msdb', 'tempdb')")
```
niphlod: CS_AS instances would fail here, use name instead of Name

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -79,235 +88,341 @@ Run the selection made by the user on the Sql Server instance specified.
...
+				}
+				else {
+					[string]$FilteredList = "'" + ($DatabaseName -join "','") + "'"
+					write-verbose "Database Specific Filtered List: $FilteredList"
```
niphlod: watch out for verbose here, use write-message instead

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -79,235 +88,341 @@ Run the selection made by the user on the Sql Server instance specified.
...
+					$databases = $server.Query("Select Name from sys.databases where name in ($FilteredList)")
```
niphlod: CS_AS instances would fail here, use name instead of Name

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -79,235 +88,341 @@ Run the selection made by the user on the Sql Server instance specified.
...
+
+			foreach ($scriptpart in $parsedscript) {
+
+				if (($QueryName.Count -ne 0) -and ($QueryName -notcontains $scriptpart.QueryName)) { continue }
+				if (!$scriptpart.DBSpecific -and !$DatabaseSpecific) {
+					if ($ExportQueries) {
+						[io.directory]::CreateDirectory($OutputPath)>$null
+						$FileName = Remove-InvalidFileNameChars ('{0}.sql' -f $Scriptpart.QueryName)
+						$FullName = ([io.path]::combine($OutputPath, $FileName))
+						write-verbose "Creating file: $FullName" -Verbose
```
niphlod: watch out for write-verbose

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -79,235 +88,341 @@ Run the selection made by the user on the Sql Server instance specified.
...
+				}
+				elseif ($scriptpart.DBSpecific -and !$instanceOnly) {
+					foreach ($currentdb in $databases) {
+						$dbname = $currentdb.name
+
+						if ($ExportQueries) {
+							[io.directory]::CreateDirectory($OutputPath)>$null
+							$FileName = Remove-InvalidFileNameChars ('{0}-{1}-{2}.sql' -f $server.DomainInstanceName, $dbname, $Scriptpart.QueryName)
+							$FullName = ([io.path]::combine($OutputPath, $FileName))
+							write-verbose "Creating file: $FullName" -Verbose
```
niphlod: watch out for write-verbose

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
```diff
@@ -1,16 +1,102 @@
...
+		$script:PesterOutputPath = Join-Path ([System.IO.Path]::GetTempPath()) ([System.IO.Path]::GetRandomFileName())
+		[io.directory]::CreateDirectory($script:PesterOutputPath) > $null
+		write-verbose "Temporary output location for test: $script:PesterOutputPath"
+
+		$database = "dbatoolsci_frk_$(Get-Random)"
+		$database2 = "dbatoolsci_frk_$(Get-Random)"
+        $server = Connect-DbaInstance -SqlInstance $script:instance2
+        $server.Query("CREATE DATABASE $database")
+        $server.Query("CREATE DATABASE $database2")
+        write-verbose "Created Test Databases: $database , $database2 on $script:instance2"
```
niphlod: do not clutter pester tests with verbosity

functions/Invoke-DbaDiagnosticQuery.ps1
sheldonhull: Case sensitivity strikes! Nice catch.

functions/Invoke-DbaDiagnosticQuery.ps1
niphlod: You can see the results directly on appveyor (see "All checks have failed" --> "Appveyor build failed" right on this page)

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
niphlod: you debug when you code, but you don't code to debug. you code to avoid ANY users debugging.
if you watch at ANY other test, no write-verbose is involved. If you need, use write-debug instead but, again, there's no place for any "write" in pester, as it reports already pretty much everything you need already

functions/Invoke-DbaDiagnosticQuery.ps1
niphlod: as for buffering, you can see other commands *usually* just return pscustomobject to the pipeline as soon as they come, without "witholding" them until the end of the function execution. In fact, such usage is discouraged but allowed **if needed**. Is there a need to return the output only in the end block ?

sheldonhull: Ok. I see. I normally do that to try and be more explicit about what is being returned as I cast the return object. If that is not desired it should be an easy change. Thanks for the feedback

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
niphlod: 259 test files are not enough ?! ;-) jokes aside, I think the only pointer would be to read, digest and understand https://github.com/sqlcollaborative/dbatools/blob/development/contributing.md#tests . (think "all code, and to be properly ___serious___ test code too, will be run on users' environment"). Debugging just pleases your daemons when you're building the solution, but it will quickly escalate to "nagging" when:
- niphlod reads the appveyor log to see why a build failed
- the early (but __serious__ ) adopter shows its colleagues dbatools works in their environment, too, running all or part of the test suite

sheldonhull: Ok, trying to learn here, appreciate the help. I get that the other test files are great examples, I meant more on the design philosophy of what you said, "you debug when you code, but you don't code to debug. you code to avoid ANY users debugging.". I figured I'd learn from this first contribution :-) Working on changes now

niphlod: everyone has his/her own method, the helpful bit is to grasp the "agenda"/"goals".

sheldonhull:
Vscode user fail... I see it accidentally included the changes to installing first responder kit, which I checked have already been updated in other work. 

Can you reject that individual file, or is there a way I can withdraw that item from the pull request? My apologies!

potatoqualitee:
I can do that for you, @sheldonhull! One moment.

potatoqualitee:
(where is that other updated work, btw? that'd be a great new PR, usebasicparsing is not in the development branch)

potatoqualitee:
also, it's been removed 👍

potatoqualitee:
1. Yes, I do see all changes to your branch, even after the PR is created. Yes, my apologies, I meant the first responder kit work.
2. Yes, I would love another PR that adds the UseBasicParsing, thank you. It's sometimes forgotten but required for some environments. That's about my understanding as well.

sheldonhull:
Is there still further failures? I'm a little unclear on the status as I'm guessing appveyor has to be triggered by reviewer. 

I also noticed the code snippet for creating pester test database is 2016 syntax. Is that acceptable (was copied from another pester test). Would using the traditional check be better?

niphlod:
Every commit starts a new build on appveyor

Il sab 12 mag 2018, 20:17 SheldonH <notifications@github.com> ha scritto:

potatoqualitee:
These are legit failures so I can't merge yet :(

sheldonhull:
Thank you. I haven't forgotten this, just have to dive into it a little more to fix. Still plan on doing this asap

sheldonhull:
Do you have a pointer on why I'd get an error that shows: 
```
A positional parameter cannot be found that accepts argument 'C:\Users\appveyor\AppData\Local\Temp\1\0w3ygqxe.hmd'
```
Is there some basic goof I have in the design of my test? Doesn't have a lot of debug output so just checking. I'll dive back in monday and see if I can fix this sucker up, any help appreciated so I can better understand.

sheldonhull:
Figured it out. Had accidental argument that was at end of new-item command, syntax issue being hidden by the assumption by powershell it was a new argument. Submitting fix now, along with fix to remove SQL 2016 specific syntax for creating and dropping databases in this and the installation of first responder tests.

sheldonhull:
Not sure why this is still failing. @niphlod  anyidea? 
1 - I committed change to convert indenting to tabs, not sure if since whitespace cache didn't get this change. Still failing for indenting without tabs. I used vscode settings on dbatools site, and applied, after converting everything to tabs. 
2 - It says can't find path but right above it it says the files were created. Additionally the same test works on the other machines. 

Thinking I might change over to the testdrive pester uses but that's not the core issue. Still should be matching the files it says it just created

sheldonhull:
Fixes in place thanks to help by simone bizzotto in slack. This finally passed all my tests. Lets see if the magic happens this time :-)

sheldonhull:
Can we bypass the indentation failure? I didn't do it I promise :-)

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -100,9 +149,12 @@ Then it will export the results to Export-DbaDiagnosticQuery.
         [Switch]$NoColumnParsing,
         [switch][Alias('Silent')]
         $EnableException
+        , [string]$OutputPath
+        , [switch]$ExportQueries
     )
 
     begin {
+        [int]$ProgressId = Get-Random
```
niphlod: cohercing is really not needed here

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -189,7 +241,14 @@ Then it will export the results to Export-DbaDiagnosticQuery.
             }
 
             if (!$instanceOnly) {
-                $databases = $server.Query("Select Name from sys.databases where name not in ('master', 'model', 'msdb', 'tempdb')")
+                if (-not $DatabaseName) {
+                    $databases = Get-DbaDatabase -SqlInstance $SqlInstance -ExcludeAllSystemDb  | select-object Name
```
niphlod: if you just need names, use (get-foo).Name

niphlod: also, reuse $instance instead of $SqlInstance (will be faster, reuse properly opened connection and doesn't need -sqlcredentials to be passed after the first connection attempt). Same thing on line 249

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -189,7 +241,14 @@ Then it will export the results to Export-DbaDiagnosticQuery.
...
+                    Write-Message -Level Debug -Message  "No Database Name Filtered: Pulled $(@($Databases).count)" -Verbose
+                }
+                else {
+                    $databases = Get-DbaDatabase -SqlInstance $SqlInstance -ExcludeAllSystemDb -Database $DatabaseName | select-object Name
+                    Write-Message -Level Debug -Message  "Filtered to databases: $(($Databases | select-object -ExpandProperty Name) -join ',')" -Verbose
+                }
             }
 
             $parsedscript = $scriptversions | Where-Object -Property Version -eq $version | Select-Object -ExpandProperty Script
```
niphlod: if you only need Script, use `($object).property` notation instead of -ExpandProperty

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -217,18 +276,26 @@ Then it will export the results to Export-DbaDiagnosticQuery.
 
                 if (($QueryName.Count -ne 0) -and ($QueryName -notcontains $scriptpart.QueryName)) { continue }
                 if (!$scriptpart.DBSpecific -and !$DatabaseSpecific) {
+                    if ($ExportQueries) {
+                        New-Item -Path $OutputPath -ItemType Directory -Force *> $null
```
niphlod: $null = or [void] is more readable here

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -305,9 +398,28 @@ Then it will export the results to Export-DbaDiagnosticQuery.
                                 Result           = $result | Select-Object * -ExcludeProperty RowError, RowState, Table, ItemArray, HasErrors
```
niphlod: use select-defaultview

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
```diff
@@ -1,16 +1,93 @@
-$CommandName = $MyInvocation.MyCommand.Name.Replace(".Tests.ps1", "")
+# test ouput directory to confirm creation of test files
+$commandname = $MyInvocation.MyCommand.Name.Replace(".Tests.ps1", "")
 Write-Host -Object "Running $PSCommandpath" -ForegroundColor Cyan
-. "$PSScriptRoot\constants.ps1"
+. (Join-path (($PSScriptRoot, '.\' -ne '')[0]) "constants.ps1")  #if running in interactive console, cd to folder and you can run this without error
```
niphlod: revert to the previous line

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
```diff
@@ -1,16 +1,93 @@
...
+write-host "loaded constants"
```
niphlod: delete this

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
```diff
@@ -1,16 +1,93 @@
...
-            $results = Invoke-DbaDiagnosticQuery -SqlInstance $script:instance2 -DatabaseSpecific *>&1
-            $results.Name.Count -gt 10 | Should Be $true
+            $results = Invoke-DbaDiagnosticQuery -SqlInstance $script:instance2 -DatabaseSpecific
+            @($results).count | Should -BeGreaterThan 10
         }
     }
-}
\ No newline at end of file
+
+    context "verifying output when exporting queries as files instead of running" {
```
niphlod: use `Context` for consistency

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
```diff
@@ -1,16 +1,93 @@
...
+
+        It "exports queries to sql files without running" {
+            Invoke-DbaDiagnosticQuery -SqlInstance $script:instance2 -ExportQueries -QueryName 'Memory Clerk Usage' -OutputPath $script:PesterOutputPath
+            @(Get-ChildItem -path $script:PesterOutputPath -filter *.sql).Count | Should -be 1
```
niphlod: `-Be` too for consistency

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
```diff
@@ -1,16 +1,93 @@
...
+        }
+
+        It "returns pscustomobject[] containing parsed queries" {
+            [System.Management.Automation.PSObject[]]$results = Invoke-DbaDiagnosticQuery -SqlInstance $script:instance2 -whatif
```
niphlod: why this type cohercition ? also, `-WhatIf`

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
```diff
@@ -1,16 +1,93 @@
...
+            @($results).count | Should -BeGreaterThan 10
+
+            #verifying the data types returned are consistent with an array of pscustom customobjects to allow working with the queries further if needed
+            #this also ensures that any "whatif" is not polluting the output stream with objects
+            # for better understanding on why I'm using write-output to validate  type, see: https://github.com/pester/Pester/issues/38
+
+            write-debug "Note that to match , results have to be declared as above, explicitly psobject, otherwise just Object[] array"
```
niphlod: use a `#comment` here, no need for this debug line

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
```diff
@@ -1,16 +1,93 @@
...
+            Write-Output -NoEnumerate $results | Should -BeOfType [System.Management.Automation.PSObject[]]
+            if ($DebugPreference -ne 'silentlycontinue') {  $results | foreach-object { write-debug "foreach-object in `$results - UnderlyingSystemType $($_.GetType().UnderlyingSystemType)" } }
```
niphlod: remove the entire line

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
```diff
@@ -1,16 +1,93 @@
...
+            $results | foreach-object {
+                Write-Output -NoEnumerate $_ | Should -BeOfType  [System.Management.Automation.PSCustomObject]
+            }
+        }
+
+        It "exports single database specific query against single database" {
+            Invoke-DbaDiagnosticQuery -SqlInstance $script:instance2  -ExportQueries  -DatabaseSpecific -queryname 'Database-scoped Configurations' -databasename $database -OutputPath $script:PesterOutputPath
+            @(Get-ChildItem -path $script:PesterOutputPath -filter *.sql | where {$_.FullName -match "($database)"}).Count | should -be 1
```
niphlod: use `Where-Object` (not only here)

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
```diff
@@ -1,16 +1,93 @@
...
+        It "exports database specific query when multiple specific databases are referenced" {
+            Invoke-DbaDiagnosticQuery -SqlInstance $script:instance2 -ExportQueries -DatabaseSpecific -queryname 'Database-scoped Configurations' -OutputPath $script:PesterOutputPath
+            @(Get-ChildItem -path $script:PesterOutputPath -filter *.sql | where {$_.FullName -match "($database)|($database2)"}).Count | should -Be 2
+        }
+
+    }
+
+    context "verifying output when running database specific queries" {
+        It "runs database specific queries against single database only when providing database name" {
+            $results = Invoke-DbaDiagnosticQuery -SqlInstance $script:instance2 -DatabaseSpecific -queryname 'Database-scoped Configurations' -databasename $database
```
niphlod: `-QueryName` (and `Should` and `-Be`, you got the gist)

functions/Invoke-DbaDiagnosticQuery.ps1
sheldonhull: Ok, if it's preferred to use `$null =` I will change that. I agree that's probably easier to read.

functions/Invoke-DbaDiagnosticQuery.ps1
sheldonhull: Since I in the past used strict-mode, using the syntax you mention above fails, while ExpandProperty handles property not existing without issues. 
Can you educate me on 2 things: 
1- wouldn't it be @($Object).property to ensure it recognizes it's an object?
2- Is it considered ok to use .property, even though property doesn't exist? I'm willing to change my approach overall, just want to understand why a little better.

functions/Invoke-DbaDiagnosticQuery.ps1
sheldonhull: I normally try to type any definition on first declaration to ensure it's correct and not implicit. Is that a bad practice?

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
sheldonhull: I did this check to ensure that objects were coming back with only pscustom object array. This was to ensure that no unbuffered output was creeping into the final result collection and impacting the returned array by having mixed types. For example, if someone put write-output type logic, this would break, so I thought I'd want to ensure the results came back without issue. Does this make sense?

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
sheldonhull: will do, but see comment above please, as had reason for doing, to ensure pure `[pscustomobject[]]` returned.

functions/Invoke-DbaDiagnosticQuery.ps1
niphlod: sorry, typo. use `$object.property`.

forget strict-mode. dbatools doesn't run with strict mode.

straight from https://github.com/sqlcollaborative/dbatools/wiki/Do-And-Do-Not . 
> Prefer dot notation over Select -ExpandProperty.


as for point 2)
> Is it considered ok to use .property, even though property doesn't exist? I'm willing to change my approach overall, just want to understand why a little better.

```
PS > $x.b | gm
gm : You must specify an object for the Get-Member cmdlet.
At line:1 char:8
+ $x.b | gm
+        ~~
    + CategoryInfo          : CloseError: (:) [Get-Member], InvalidOperationException
    + FullyQualifiedErrorId : NoObjectInGetMember,Microsoft.PowerShell.Commands.GetMemberCommand

PS > ($x | select-object -ExpandProperty b) | gm
select-object : Property "b" cannot be found.
At line:1 char:7
+ ($x | select-object -ExpandProperty b) | gm
+       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (@{a=1}:PSObject) [Select-Object], PSArgumentException
    + FullyQualifiedErrorId : ExpandPropertyNotFound,Microsoft.PowerShell.Commands.SelectObjectCommand

gm : You must specify an object for the Get-Member cmdlet.
At line:1 char:42
+ ($x | select-object -ExpandProperty b) | gm
+                                          ~~
    + CategoryInfo          : CloseError: (:) [Get-Member], InvalidOperationException
    + FullyQualifiedErrorId : NoObjectInGetMember,Microsoft.PowerShell.Commands.GetMemberCommand
```
both will end in null. -expandproperty will raise ..... but no code here is actually checking (the exception or the being null). 
So go for the dot notation which is:
- shorter 
- easier to read 
- for arrays of objects is a ton faster than passing down the dreaded `|`
- doesn't raise, so you don't need additional code for try/catch
- in the even you need to assure result is there, just check for $null

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
niphlod: not really. got that you seem to like forcing types, but you're missing the big picture.
if someone puts write-output [pscustomobject] the function is screwed either way, isn't it ?
if you wanna check for creeping mistakes, check that every returned "thing" is an object with all the correct properties you expect

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
niphlod: comment was on line 57. we don't need CI to run any debug. debugging is for humans hunting bugs, not computers.

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
sheldonhull: Ok so you major difference in approach is that instead of worrying about the pipeline returning the exact same object type, you'd be more concerned with do all the properties I want exist? I might be just trying to be *too strongly typed* with a c# approach, and instead should design for properties existing, but not worry about object vs pscustomobject. Does that make sense?

functions/Invoke-DbaDiagnosticQuery.ps1
sheldonhull: This is great. Thanks for this. 
I'm assuming that for most dev work you find not using strict-mode and this type of approach is a much easier/cleaner to work with?

Thanks again for the help and resources, will use to improve my next pulls.

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -1,94 +1,143 @@
...
 
     [CmdletBinding(SupportsShouldProcess)]
+    [outputtype([pscustomobject[]])]
     param (
         [parameter(Mandatory = $true, ValueFromPipeline = $true, Position = 0)]
         [Alias("ServerInstance", "SqlServer")]
         [DbaInstanceParameter[]]$SqlInstance,
+
+        [Alias('Database')]
+        [string[]]$DatabaseName,
```
wsmelton: This needs to be changed to `[object[]]$Database` so the TEPP will kick in properly. You can remove the alias.

wsmelton: Our standard is to also include `[object[]]$ExcludeDatabase` to allow folks to have the opposite filter.

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -189,7 +242,14 @@ Then it will export the results to Export-DbaDiagnosticQuery.
             }
 
             if (!$instanceOnly) {
-                $databases = $server.Query("Select Name from sys.databases where name not in ('master', 'model', 'msdb', 'tempdb')")
+                if (-not $DatabaseName) {
+                    $databases = (Get-DbaDatabase -SqlInstance $server -ExcludeAllSystemDb).Name
+                    Write-Message -Level Debug -Message  "No Database Name Filtered: Pulled $(@($Databases).count)" -Verbose
```
wsmelton: Why are we including `-Verbose` parameter on this command?

We reserve `-Level Debug` for the purpose of T-SQL statements. You might put this as `-Level VeryVerbose` or `-Level System` just so the logging system picks it up. I don't see any reason to output this type of information as debug.

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -199,36 +259,61 @@ Then it will export the results to Export-DbaDiagnosticQuery.
                 $QueryName = Invoke-DiagnosticQuerySelectionHelper $parsedscript
                 $first = $false
             }
+            #since some database level queries can take longer (such as fragmentation) calculate progress with database specific queries * count of databases to run against into context
+            $CountOfDatabases = ($databases).count
```
wsmelton: You used this method earlier to pull this value, why not use the same method here?

`@($Databases).Count`

wsmelton: Count is also capitalized because it is a property name.

functions/Invoke-DbaDiagnosticQuery.ps1
wsmelton: Count property also needs to be capitalized.

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -199,36 +259,61 @@ Then it will export the results to Export-DbaDiagnosticQuery.
...
+
 
+            if ($QueryName.Count -ne 0) {
+                #if running all queries, then calculate total to run by instance queries count + (db specific count * databases to run each against)
+                $CountOfDatabaseSpecific = @($parsedscript | Where-Object {$_.QueryName -in $QueryName -and $_.DBSpecific -eq $true}).count
```
wsmelton: We use camel casing for variables that are not part of the parameter block of the function... plus you could also use shorthand for database references (e.g. `$countDbSpecific`).

functions/Invoke-DbaDiagnosticQuery.ps1
wsmelton: Also should be done as `Select-DefaultView -InputObject $result...` instead of piping it for better performance.

tests/Install-DbaFirstResponderKit.Tests.ps1
```diff
@@ -10,8 +10,7 @@ Describe "$CommandName Integration Tests" -Tags "IntegrationTests" {
             $server.Query("CREATE DATABASE $database")
         }
         AfterAll {
-            $server.Query("ALTER DATABASE $database SET OFFLINE WITH ROLLBACK IMMEDIATE")
-            $server.Query("DROP DATABASE IF EXISTS $database")
+            $null = Get-DbaDatabase -SqlInstance $server -Database  $database | Remove-DbaDatabase -Confirm:$false
```
wsmelton: This should be reverted back to the original use of `Query()` method because we do not want changes to public functions breaking test on a different command.

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
```diff
@@ -1,16 +1,93 @@
...
+
+        $database = "dbatoolsci_frk_$(Get-Random)"
+        $database2 = "dbatoolsci_frk_$(Get-Random)"
+        $server = Connect-DbaInstance -SqlInstance $script:instance2
+        $server.Query("CREATE DATABASE $database")
+        $server.Query("CREATE DATABASE $database2")
+
+    }
+    AfterAll {
+        $null = Get-DbaDatabase -SqlInstance $server -Databases @($database, $database2) | Remove-DbaDatabase -Confirm:$false
```
wsmelton: Utilize similar pattern as the other test with `Query()` method for consistence and not having a dependency on other public functions.

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
```diff
@@ -1,16 +1,93 @@
...
+        Remove-Item $script:PesterOutputPath -Recurse -erroraction silentlycontinue
+    }
+
+
+    Context "verifying output when running queries" {
         It "runs a specific query" {
-            $results = Invoke-DbaDiagnosticQuery -SqlInstance $script:instance2 -QueryName 'Memory Clerk Usage' *>&1
-            $results.Name.Count | Should Be 1
+            $results = Invoke-DbaDiagnosticQuery -SqlInstance $script:instance2 -QueryName 'Memory Clerk Usage'
+            @($results).Count | Should -Be 1
```
wsmelton: Please use `....Should Be 1`. We do not reference the parameters for Should with `-` prefix so keeps code looking consistent.

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
niphlod: this is pretty much on every test....... let's not overcomplicate things, get-dbadatabase and remove-dbadatabase are not going to suddenly disappear

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
niphlod: consistency over anything else but please do mind that pester (from 4.0) is going on with all dashes `-` rather than no dashes at all. New tests should definitely use the `-Be` syntax

functions/Invoke-DbaDiagnosticQuery.ps1
sheldonhull: I didn't remove Database alias, as that was the previous name, wanted to prevent breaking existing scripts, so aligned name with standard, but left alias to allow database to be used.
The exclude is a good idea, so I'll include in the check-in.

niphlod: go the other way around. TEPP will kick in only on `Database` and `ExcludeDatabase`. Every existing dbatools function with `DatabaseName` as a parameter will be changed to use `Database` (optionally using `DatabaseName` as the alias)

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -1,94 +1,145 @@
...
+    #>
 
     [CmdletBinding(SupportsShouldProcess)]
+    [outputtype([pscustomobject[]])]
     param (
         [parameter(Mandatory = $true, ValueFromPipeline = $true, Position = 0)]
         [Alias("ServerInstance", "SqlServer")]
         [DbaInstanceParameter[]]$SqlInstance,
+
+        [Alias('Database')]
```
niphlod: switch this .

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -99,10 +150,13 @@ Then it will export the results to Export-DbaDiagnosticQuery.
         [Switch]$NoPlanColumn,
         [Switch]$NoColumnParsing,
         [switch][Alias('Silent')]
```
niphlod: last param should always be 
```
        [Alias('Silent')]
        [switch]$EnableException
```
for functions coming before the great-silent-rewrite. New functions can just use
```
[switch]$EnableException
```

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -189,7 +244,12 @@ Then it will export the results to Export-DbaDiagnosticQuery.
             }
 
             if (!$instanceOnly) {
-                $databases = $server.Query("Select Name from sys.databases where name not in ('master', 'model', 'msdb', 'tempdb')")
+                if (-not $DatabaseName) {
```
niphlod: changing the parameter at the top, remember to change this too

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
```diff
@@ -1,16 +1,98 @@
...
+            $results = Invoke-DbaDiagnosticQuery -SqlInstance $script:instance2 -DatabaseSpecific -ExcludeDatabase $database2
+            @($results | Where-Object {$_.DatabaseName -eq $Database1}).Count | Should -BeGreaterThan 1
+            @($results | Where-Object {$_.DatabaseName -eq $Database2}).Count | Should -Be 0
+        }
+    }
+
+    context "verifying output when exporting queries as files instead of running" {
+
+        It "exports queries to sql files without running" {
+            Invoke-DbaDiagnosticQuery -SqlInstance $script:instance2 -ExportQueries -QueryName 'Memory Clerk Usage' -OutputPath $script:PesterOutputPath
```
niphlod: assign this to a result so output won't be cluttered

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
```diff
@@ -1,16 +1,98 @@
...
+            @(Get-ChildItem -path $script:PesterOutputPath -filter *.sql).Count | Should -Be 1
+        }
+
+        It "verifies returns object with required properties when running with whatif" {
+            [System.Management.Automation.PSObject[]]$results = Invoke-DbaDiagnosticQuery -SqlInstance $script:instance2 -WhatIf
+
+            @($results).count                                                             | Should -BeGreaterThan 10
+            (($results).ComputerName     | Where-Object {$_ -eq $null}).count             | Should -Be 0
```
niphlod: why all those parenthesis ? `$results.ComputerName` isn't the same of `($results).ComputerName`

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
```diff
@@ -1,16 +1,98 @@
...
+            (($results).InstanceName     | Where-Object {$_ -eq $null}).count             | Should -Be 0
+            (($results).SqlInstance      | Where-Object {$_ -eq $null}).count             | Should -Be 0
+            (($results).Number           | Where-Object {$_ -eq $null}).count             | Should -Be 0
+            (($results).Name             | Where-Object {$_ -eq $null}).count             | Should -Be 0
+            (($results).Description      | Where-Object {$_ -eq $null}).count             | Should -Be 0
+            (($results).DatabaseSpecific | Where-Object {$_ -eq $null}).count             | Should -Be 0
+            (($results).DatabaseName).count                                               | Should -Be @($results).count
+            (($results).Notes).count                                                      | Should -Be @($results).count
```
niphlod: Consistency! Either `count` or `Count` (preferred the latter) but definitely not both

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
```diff
@@ -1,16 +1,98 @@
...
+            (($results).Result).Count                                                     | Should -Be @($results).count
+        }
+
+        It "exports single database specific query against single database" {
+            Invoke-DbaDiagnosticQuery -SqlInstance $script:instance2  -ExportQueries  -DatabaseSpecific -QueryName 'Database-scoped Configurations' -databasename $database -OutputPath $script:PesterOutputPath
```
niphlod: assign this to a result so output won't be cluttered (here and following)

niphlod:
I just use autocomplete. I feel learning to code in python taught me "code
is read 10x times it is written, so it should be nice".
I agree that with PS or c# the lack of a "good" editor/IDE results in coder
bored and code ugly.

Kinda like ssms in not autocomplete days for tsql.

Il gio 31 mag 2018, 01:25 SheldonH <notifications@github.com> ha scritto:

tests/Invoke-DbaDiagnosticQuery.Tests.ps1
sheldonhull: Sorry, I'm unclear. I will address if my next pull request gets this wrong. I adjusted. I also couldn't be sure I was clear on why it impacted result. I got the same results when reviewing

```powershell
($results).Name.GetType() 
$results.Name.GetTYpe() 
($null).Count
$null.Count
```

wsmelton:
@sheldonhull you just commit to your remote of `sheldonhull:development` and this PR will automatically update once you push the changes to GitHub.

sheldonhull:
Ok, changes pushed. Let me know if this looks good or if I'm missing anything else. Thanks for all the feedback!

potatoqualitee:
so close 😇 will merge upon approval

sheldonhull:
awesome. Let me know if anything else I need to do. I'm assuming it's just pending someone's time for now, so will wait. Cheers

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -1,94 +1,145 @@
...
     [CmdletBinding(SupportsShouldProcess)]
+    [outputtype([pscustomobject[]])]
     param (
         [parameter(Mandatory = $true, ValueFromPipeline = $true, Position = 0)]
         [Alias("ServerInstance", "SqlServer")]
         [DbaInstanceParameter[]]$SqlInstance,
+
+        [Alias('Database')]
+        [object[]]$DatabaseName,
+
```
wsmelton: The above parameter needs to be changed to:
```[Alias('DatabaseName')
[object[]]$Database,
```

functions/Invoke-DbaDiagnosticQuery.ps1
wsmelton: Update this to the `$Database` parameter

functions/Invoke-DbaDiagnosticQuery.ps1
wsmelton: Need to also add `[object[]]$ExcludeDatabase`

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -189,7 +246,12 @@ Then it will export the results to Export-DbaDiagnosticQuery.
             }
 
             if (!$instanceOnly) {
-                $databases = $server.Query("Select Name from sys.databases where name not in ('master', 'model', 'msdb', 'tempdb')")
+                if (-not $DatabaseName) {
+                    $databases = (Get-DbaDatabase -SqlInstance $server -ExcludeAllSystemDb -ExcludeDatabase $ExcludeDatabase).Name
+                }
+                else {
+                    $databases = (Get-DbaDatabase -SqlInstance $server -ExcludeAllSystemDb -Database $DatabaseName -ExcludeDatabase $ExcludeDatabase).Name
+                }
```
wsmelton: Need to add the filter for the `$Database` and `$ExcludeDatabase` parameter

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -258,22 +344,54 @@ Then it will export the results to Export-DbaDiagnosticQuery.
...
+
+                        [pscustomobject]@{
+                            ComputerName     = $server.NetName
+                            InstanceName     = $server.ServiceName
+                            SqlInstance      = $server.DomainInstanceName
+                            Number           = $scriptpart.QueryNr
+                            Name             = $scriptpart.QueryName
+                            Description      = $scriptpart.Description
+                            DatabaseSpecific = $scriptpart.DBSpecific
+                            DatabaseName     = $null
```
wsmelton: Change this to just `Database`

wsmelton: Change other occurrences of the `DatabaseName` property to `Database` as well for consistency.

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -300,14 +418,33 @@ Then it will export the results to Export-DbaDiagnosticQuery.
                                 Name             = $scriptpart.QueryName
                                 Description      = $scriptpart.Description
                                 DatabaseSpecific = $scriptpart.DBSpecific
-                                DatabaseName     = $currentdb.name
+                                DatabaseName     = $currentDb
```
wsmelton: This should be reverted back to `$currentDb.Name` and property updated to just `Database`

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -258,22 +344,54 @@ Then it will export the results to Export-DbaDiagnosticQuery.
...
+                                Write-Progress -Id $ProgressId -ParentId 0 -Activity "Collecting diagnostic query data from $($currentDb) on $instance" -Status ('Processing {0} of {1}' -f $counter, $scriptcount) -CurrentOperation $scriptpart.QueryName -PercentComplete (($Counter / $scriptcount) * 100)
+                            }
 
-                            if (-not $EnableException) { Write-Progress -Id 1 -ParentId 0 -Activity "Collecting diagnostic query data from $dbname on $instance" -Status ('Processing {0} of {1}' -f $counter, $scriptcount) -CurrentOperation $scriptpart.QueryName -PercentComplete (($Counter / $scriptcount) * 100) }
-                            Write-Message -Level Output -Message "Collecting diagnostic query data from $dbname for $($scriptpart.QueryName) on $instance"
+                            Write-Message -Level Verbose -Message "Collecting diagnostic query data from $($currentDb) for $($scriptpart.QueryName) on $instance"
                             try {
-                                $result = $server.Query($scriptpart.Text, $currentdb.Name)
+                                $result = $server.Query($scriptpart.Text, $currentDb)
```
wsmelton: The string value of the database name has to be passed into the query method. Revert this back to `$currentDb.Name`

functions/Invoke-DbaDiagnosticQuery.ps1
```diff
@@ -300,14 +418,33 @@ Then it will export the results to Export-DbaDiagnosticQuery.
...
+
+                            [pscustomobject]@{
+                                ComputerName     = $server.NetName
+                                InstanceName     = $server.ServiceName
+                                SqlInstance      = $server.DomainInstanceName
+                                Number           = $scriptpart.QueryNr
+                                Name             = $scriptpart.QueryName
+                                Description      = $scriptpart.Description
+                                DatabaseSpecific = $scriptpart.DBSpecific
+                                DatabaseName     = $null
```
wsmelton: Update to `Database`

tests/Install-DbaFirstResponderKit.Tests.ps1
wsmelton: I would still like to see this reverted back to `Query()` method, regardless of whether the commands will stay around it ensures the test runs at a consistent time.

sheldonhull:
The last failure had nothing to do with my checkin. Something else seems to be wrong

```
Describe : Export-DbaDacpac Integration Tests
Context  : 
Name     : It exports a dacpac
Result   : Failed
Message  : Cannot bind argument to parameter 'Path' because it is null.
```

Can someone trigger off a test without that included to make sure I've passed all the checks, or mark as passed from the the last test?

potatoqualitee:
That test failure is indeed unrelated. Sometimes appveyor craps out. Your changes have been merged! :bananadance:

PR #3602, opened 2018-05-10
Text fixes glb0509

gbargsley:
Punctuation and text fixes found during web site creations.


potatoqualitee:
thanks so much to you both!

PR #3605, opened 2018-05-12
Set-DbaDbCompression - Workaround for what seems like SMO bug

jpomfret:
## Type of Change

 - [X] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
- Workaround so that setting CompressionType to 'None' for all objects works. There seems to be a bug in SMO that doesn't generate the DATA_COMPRESSION keyword in the rebuild statement at the index level (works fine if you set the object (heap/CL) to None.
- Adds tests to confirm all objects in a database get row/page/none compression
- Now supports ShouldProcess to resolve ScriptAnalyzer issue

SMO bug report logged: https://feedback.azure.com/forums/908035-sql-server/suggestions/34080112-data-compression-smo-bug

### Approach

- Added logic so that at the index level if setting DataCompression to None it uses T-SQL instead of SMO.

### Commands to test

`Set-DbaDbCompression -SqlInstance Server -Database DatabaseName -CompressionType None`

### Screenshots
image

### DIFF
--- a/functions/Set-DbaDbCompression.ps1
+++ b/functions/Set-DbaDbCompression.ps1
@@ -146,7 +146,8 @@ function Set-DbaDbCompression {
                     if ($CompressionType -eq "Recommended") {
                         if (Test-Bound "InputObject") {
                             Write-Message -Level Verbose -Message "Using passed in compression suggestions"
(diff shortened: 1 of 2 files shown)


potatoqualitee:
passes tests 👍 thank you!

PR #3611, opened 2018-05-16
Test-DbaDbCompression - exclude memory optimized tables

jpomfret:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
If you run this against a database with in-memory tables it will fail

### Approach

Changed to use sys.tables instead of sys.objects and then only include where is_memoryoptimized = 0


jpomfret:
I've found another bug, fails if there is an encrypted column... I'll try and resolve that asap. Looks like I need to add some test coverage also

PR #3612, opened 2018-05-16
Fix usage of ForEach in Get-DbaDbPageInfo

HouzuoGuo:
Replace the usage of ForEach in Get-DbaDbPageInfo with ForEach-Object so that the function will import successfully in PowerShell version 3.0.

## Type of Change

### Purpose
Fix the usage of ForEach in function Get-DbaDbPageInfo so that it may be imported successfully into PowerShell v3.0

### Approach
Use ForEach-Object instead.

### Commands to test
Simply run `Import-Module` in PowerShell v3.0


potatoqualitee:
thanks again 👍 💯  Totally overlooked this. I will merge then update to `ForEach-Object` in place of the alias. Should be in master shortly 🍦

PR #3613, opened 2018-05-17
Restore-DbaDatabase: use SqlCredential to call Get-DbaBackupInformation

HouzuoGuo:
## Type of Change

### Purpose
Pass user-specified SqlCredential to the call of Get-DbaBackupInformation made by Restore-DbaDatabase, so that the function Restore-DbaDatabase will not spawn a credential input dialog unnecessarily. Plus, the dialog will not spawn in non-interactive session, causing Get-DbaBackupInformation to hang.

### Commands to test
Run `Restore-DbaDatabase` with a constructed `PSCredential` passed into parameter `-SqlCredential`.


Zhunya:
Thank you for fix!

potatoqualitee:
thanks so much!

PR #3615, opened 2018-05-17
Get-DbaConnection - up for discusion

potatoqualitee:
Created this because I needed more information than what Test-DbaConnectionAuthScheme was offering.

After I made the command and the test, I realized that Get-DbaProcess has most, but not all, of this information.


wsmelton:
` Get-DbaProcess` is very limited in what it returns because it uses that enumerator. For connections, we could also validate if a listener exists on the instance and if it does also include that information.

Would we want to include say number of request/sessions? You can do a left join to see if it has any activity.

potatoqualitee:
sweet, glad you see value. i'll merge. we can include the number, do you ahve the SQLs?

PR #3631, opened 2018-05-23
Backup-DbaDatabase, path fixes

niphlod:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Purpose
Fix path building logic for some scenarios. Multiple issues have been raised and - potentially- this solves most. Dunno exactly where, but we started building backup file paths without extensions (e.g. ...\foo-1-of-3bak) and totally wrong names too (e.g. \foo-1-of-33-of-3bak). That wouldn't have been that much of a deal if higher level function didn't use striping by default, but since they do, this raised the number of issues and brought this to my attention. 
I'd then say it's urgently needed to be merged in dev, then master, and incorporated in a new release ASAP.

### Approach
Logic has been wherever possible shifted from string concatenation to .net path methods. It has been refactored and some code paths have been removed.
In addition:
- path checking is now done just once per path (before it was once per file, even if the directory was the same)
- disk path are now normalized (think `c:\\temp\\foo\bar\\baz.bak` )
- createfolder wasn't working as advertised when multiple backupfolder were passed in
- refactored some camelCase and brought every bit of syntax to be more consistent
- slapped in diagnostic exclusion to make pester happy
- progress has not a fixed id (works when multithreaded)
- regression tests for each fix have been added to the unittests

### DIFF
--- a/functions/Backup-DbaDatabase.ps1
+++ b/functions/Backup-DbaDatabase.ps1
@@ -84,7 +84,7 @@ function Backup-DbaDatabase {
 
             .PARAMETER BuildPath
                 By default this command won't attempt to create missing paths, this switch will change the behavious so that it wll
-            
+
             .PARAMETER IgnoreFileChecks
                 This switch stops the function from checking for the validity of paths. This can be useful if SQL Server only has read access to the backup area.
                 Note, that as we can't check the path you may well end up with errors.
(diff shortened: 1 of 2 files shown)


potatoqualitee:
thanks so very much!

wsuhoey:
For posterity, I was one of the users that reported this issue :) Thanks @js0505 for filing it for me!

Issue #3641, opened 2018-05-30
Backup-DbaDatabase double the dot before the extension


pollusb:
### Before submitting a bug report:

- [x] Ensure you are able to reproduce it on the latest released version (_we release often_)
- [x] Verified this bug is not already reported in an issue
- [x] Verified errors are not related to permissions
- [x] Can reproduce in a clean PowerShell session (_clean = `powershell -NoProfile`_)

## Steps to Reproduce
```powershell
Backup-DbaDatabase @XPS1 -Database MyDB -BackupDirectory D:\SQL -BackupFileName MyDB.bak -CopyOnly -FileCount 3
```

## Expected Behavior
I expect to get the 3 following files:
MyDB-1-of-3.bak
MyDB-2-of-3.bak
MyDB-3-of-3.bak

## Actual Behavior
**It doubles the dot before the extension only if I provide the BackupFileName and FileCount > 1**
MyDB-1-of-3..bak
MyDB-2-of-3..bak
MyDB-3-of-3..bak

## Environmental data

 - PowerShell: PSVersion                      5.1.17134.48
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.17134.48
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
 - - SQL Server: SQL 2012

PR #3645, opened 2018-05-30
Fixed Extra dot in multi filename #3641

gbargsley:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Fix Issue 3641.  Not sure if best method, but it works.
### Screenshots

image

### DIFF
--- a/functions/Backup-DbaDatabase.ps1
+++ b/functions/Backup-DbaDatabase.ps1
@@ -333,7 +333,7 @@ function Backup-DbaDatabase {
             elseif ('' -ne $BackupFileName) {
                 $File = New-Object System.IO.FileInfo($BackupFileName)
                 $BackupFinalName = $file.Name
-                $suffix = $file.extension
+                $suffix = $file.extension -Replace '^\.',''
                 if ( '' -ne (Split-Path $BackupFileName)) {
                     Write-Message -Level Verbose -Message "Fully qualified path passed in"
                     $FinalBackupPath += [IO.Path]::GetFullPath($file.DirectoryName)
@@ -359,7 +359,7 @@ function Backup-DbaDatabase {
             if ($FinalBackupPath.Count -gt 1) {
                 $File = New-Object System.IO.FileInfo($BackupFinalName)
                 for ($i = 0; $i -lt $FinalBackupPath.Count; $i++) {
-                    $FinalBackupPath[$i] = $FinalBackupPath[$i] + "\" + $($File.BaseName) + "-$($i+1)-of-$FileCount$suffix"
+                    $FinalBackupPath[$i] = $FinalBackupPath[$i] + "\" + $($File.BaseName) + "-$($i+1)-of-$FileCount.$suffix"
                 }
             }
             elseif ($FinalBackupPath[0] -ne 'NUL:') {
--- a/functions/Backup-DbaDatabase.ps1
+++ b/functions/Backup-DbaDatabase.ps1
@@ -359,7 +359,7 @@ function Backup-DbaDatabase {
             if ($FinalBackupPath.Count -gt 1) {
                 $File = New-Object System.IO.FileInfo($BackupFinalName)
                 for ($i = 0; $i -lt $FinalBackupPath.Count; $i++) {
-                    $FinalBackupPath[$i] = $FinalBackupPath[$i] + "\" + $($File.BaseName) + "-$($i+1)-of-$FileCount.$suffix"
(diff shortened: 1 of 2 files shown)


niphlod:
wan't the problem an EXTRA dot ?

niphlod:
(BTW, if it fixes a bug, put the number in the description...... @wsmelton spent a lot of time in templating GH issues)

niphlod:
How can adding a dot in code solve the problem of having an extra dot in
the filename? Ghggggh, I'm prolly missing something

Il mer 30 mag 2018, 15:25 Garry Bargsley <notifications@github.com> ha
scritto:

gbargsley:
I removed a dot in the source code that was between the filename and extention.  I think somehow the $suffix was getting an a dot in it, and made it have two dots.  So I just removed the hardcoded dot as I could not tell where $suffix was getting a dot added.

niphlod:
okay, misread in mobile the diff, sorry.

niphlod:
something else seems to be the culprit, let's see if I can replicate

niphlod:
line 336 is the culprit...... a fileinfo extension property contains the actual dot.

gbargsley:
I saw that, but did not think the .extension would add a dot.  So should we strip it from that call to make for better code? Then add the dot back to where it was in the line 362?

gbargsley:
I did this to line 336: `$suffix = $file.extension.Replace(".","")`
And added the period back to 362: `$FinalBackupPath[$i] = $FinalBackupPath[$i] + "\" + $($File.BaseName) + "-$($i+1)-of-$FileCount.$suffix"`

And it seems to work.

niphlod:
yup, fix would just be changing that line to 
`$suffix = $file.Extension  -Replace '^\.', ''`

BTW: tests failed correctly with your solution, maybe slap in a regr test to make sure #3461 doesn't happen again

gbargsley:
I changed the code to your way.  Unfortunately, I do not know what you mean about a regression test.  I would love to learn.

niphlod:
I recently added, e.g., https://github.com/sqlcollaborative/dbatools/blob/development/tests/Backup-DbaDatabase.Tests.ps1#L126 to make sure all backups are done with extension (till recently there were cases where extension was not used). A regression test is a test that makes sure a problem won't pop up ever again. In this case, the side-case of running a backup with:
- a single backuppath
- a single backupfilename 
- filecount greater than 1 
showed that files with more than one dot were created.
The regression test in this case would be to create lines in the pester test that make sure that when you run backup-dbadatabase with:
- a single backuppath
- a single backupfilename 
- filecount greater than 1 
created files have no the double-dot in them.

gbargsley:
Ok, let me try.

gbargsley:
So all green.  Does that mean the test I added worked?

niphlod:
seems good . BTW: isn't 
`[IO.Path]::GetExtension($path) | Should -Not -BeLike '*..*'` more readable ?

gbargsley:
But what if 3 or anything other than 1 is used in the filename/extension?

niphlod:
3 or 1 of what ? tests are not "dynamic". 
You ask for a backup with a filename with .bak extension and you expect the resulting files to not have more than one dot (i.e. "it doesn't change the extension passed in"). 
It'd be cool to have genetic test cases (i.e. computers checking random inputs and verifying all things obey to rules) but, alas, we're not there yet.

gbargsley:
I meant if it gets ... or .... or more. Your solution seems to look for only ..

niphlod:
My solution looks for `*..*`. which we deemed wrong. 
Coincidentally, `*..*` does not only match two consecutive periods but also also `...` (3 consecutive periods) and `....` (4 consecutive periods)  and so on till the end of path limitations in PS so ..... I don't see where your doubts arise

gbargsley:
It is not doubt as I would never attempt to do that. I am just trying to understand.

niphlod:
Match is "two or more" so all cases are covered.

Il mer 30 mag 2018, 19:00 Garry Bargsley <notifications@github.com> ha
scritto:

potatoqualitee:
thank you both!

wsmelton:
@gbargsley when you do your commits if you include specific term/phrase as:

fixes #3641 

It will cause the issue to be closed once Chrissy merges it. I should also do the same thing if you put that phrase in the PR content too.

gbargsley:
So not in the title and only in the description if I use the #. 

Trying to understand.

niphlod:
https://help.github.com/articles/closing-issues-using-keywords/ . 
tl;dr: fill the template sqlcollaborative has already

PR #3652, opened 2018-05-31
Fixed #3667 - Removed unneeded line causing issue

gbargsley:
Found the line that was causing the addition of traceflags to not apply correctly.

image

Was not sure how to code a regression test to check for this type issue.


potatoqualitee:
woo thanks!

PR #3667, opened 2018-06-07
0.9.343

potatoqualitee:


potatoqualitee:
hey mario, yeah, master is good enough. we have tons of pester tests which increases my confidence in the releases. I'll make one today as soon as I find a good name

PR #3673, opened 2018-06-10
Get-DbaResourceGovernorClassiferFunction updates

potatoqualitee:
@suxstellino - i made some updates using sample code from microsoft. unfortunately, if the udf has dependencies such as a table, those aren't copied and it fails :( But i dont know if a good way to trace that. `Get-DbaDependency` failed me here.


suxstellino:
Thank you @potatoqualitee, I'm sorry for the late replies. Hope to push suggested changes (the ones received two weeks ago) next week. About the object (dependency) how can we manage that?

PR #3681, opened 2018-06-13
Remove CredentialAttribute Type from left overs

wsmelton:
Cleaning out commands that still had the CredentialAttribute() type added to credential parameters.

This does not allow double-hop situations, particularly when you have remote sessions in use.

It should not break anything but we will see 🤞


wsmelton:
Why does that stinking CodeCov keep showing up? // @niphlod that is the only thing that caused this to fail.

niphlod:
@wsmelton : from "my" management panel on codecov no hooks are defined. I'll send a mail to their wonderful support to see what's going on

PR #3688, opened 2018-06-17
[New Commands] Get-DbaSqlLogConfig (400th) and Set-DbaSqlLog

wsmelton:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
So a few big things changed in this PR as the commands created require it.

- SMO updated to the latest version released from the `sqlserver` module
- SQL Server 2012 added the ability to [limit the size of the errorlog file](https://support.microsoft.com/en-us/help/2199578/how-to-manage-the-sql-server-error-log). Microsoft added support for this in SMO but only in the latest version of SMO.
- In the process of updating SMO it was discovered we were not actually importing the version of SMO packaged with the module. Worked with Fred on this and we discovered it was due to the order we load the SMO dll files. I checked how the `sqlserver` module was doing this and simply adjusted our order to match it.
- The new commands include the ability to pull the number of log files and error log size (for 2012+). The error log size value required the newer version of SMO.

### Commands to test
See pester test. They all pass on my local system.


functions/Get-DbaSqlLogConfig.ps1
```diff
@@ -0,0 +1,70 @@
...
+            catch {
+                Stop-Function -Message "Failure" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
+            }
+
+            $numLogs = $server.NumberOfLogFiles
+            if ($server.VersionMajor -ge 11) {
+                $logSize = $server.ErrorLogSizeKb
+            }
+            else {
+                $null
```
FriedrichWeinmann: This will return an empty null object to the pipeline.
Probably should be
```
$logSize = $null
```

functions/Set-DbaSqlLog.ps1
```diff
@@ -0,0 +1,162 @@
...
+                catch {
+                    $collectionNumberLog.Status = "Failed collection"
+                    $collectionNumberLog
+                    Stop-Function -Message "Issue collecting current value for number of error logs" -Target $server -ErrorRecord $_ -Exception $_.Exception.InnerException.InnerException.InnerException -Continue
+                }
+
+                if ($NumberOfLog -eq $currentNumLogs) {
+                    if ($PSCmdlet.ShouldProcess($server, "Provide warning that NumberOfLog value already matches the configured value")) {
+                        Write-Message -Level Warning -Message "The provided value for NumberOfLog is already set" -Continue
+                    }
```
FriedrichWeinmann: Looks good in `-WhatIf`, but makes little sense in Production use (asking for a confirm to do nothing)
Instead could use `Test-Bound 'WhatIf'` to check for WhatIf use and change messages accordingly

functions/Set-DbaSqlLog.ps1
```diff
@@ -0,0 +1,162 @@
...
+                catch {
+                    $collectionErrorLogSize.Status = "Failed collection"
+                    $collectionErrorLogSize
+                    Stop-Function -Message "Issue collecting current value for number of error logs" -Target $server -ErrorRecord $_ -Exception $_.Exception.InnerException.InnerException.InnerException -Continue
+                }
+
+                if ($SizeInKb -eq $currentSizeInKb) {
+                    if ($PSCmdlet.ShouldProcess($server, "Provide warning that SizeInKb value already matches the configured value")) {
+                        Write-Message -Level Warning -Message "The provided value for SizeInKb is already set" -Continue
+                    }
```
FriedrichWeinmann: Looks good in -WhatIf, but makes little sense in Production use (asking for a confirm to do nothing)
Instead could use `Test-Bound 'WhatIf'` to check for WhatIf use and change messages accordingly

wsmelton:
Some reason `Read-DbaAuditFile` tests and `Publish-DbaDacpac` test both failed. The dll both of these use was including in updating SMO.

functions/Set-DbaSqlLog.ps1
wsmelton: Think I had some other idea writing that part but took it out or something...went and just remove the shouldprocess block.

wsmelton:
I'm not sure why these broke around the XE and DAC dll files, but I had to revert those to the older versions for some reason. I have not worked with the commands enough to really troubleshoot those at this time.

potatoqualitee:
awwwwwwwwwwwwwesome!! thank you :D 

one thing to note, I run Find-DbaCommand before every release, so you don't have to worry about that. 

Also, I noticed you no longer have to manually load XEVents. How did that happen and was there any impact on the loading time of the module?

Annnnnd, I'd like to prevent the sprawl of DbaSql commands. Perhaps we can consider Dba being the equivalent of Sql, and only have limited exceptions (SqlBuild comes to mind). Can you please change the name of this command to remove Sql or propose an alternative?

🙇

wsmelton:
The new commands do have DBA prefix on them. There is no alternative to calling it `Get-DbaLogConfig` and `Set-DbaLogConfig` but that sounds to close to being the log for the module. We already have `Get-DbaSqlLog` to pull the error log for an instance so I simply followed that naming convention.

Only other option is to make "Sql" related to "Instance" and then have `Get-DbaInstanceLog', 'Get-DbaInstanceLogConfig' and `Set-DbaInstanceLog`.

The no longer requiring load of SMO files came from actually correcting the order that we loaded them. The original method actually had the module using what was registered in the GAC (something Fred determined). Since that is corrected now I'm guessing it does not require having to load them explicitly. From a timing standpoint I don't notice because I have not generated the `allcommands` file to see if it mattered. Loading the module with `.git` folder has always take a while for me and I have not noticed it getting worse or better from dev environment setup.

wsmelton:
A lot of commands to change if you don't want to use `DbaSql`, don't think we have been limiting the use of it: 

image

wsmelton:
I'd be good with taking these:

```
Get-DbaSqlLog
Get-DbaSqlLogConfig
Set-DbaSqlLog
```

And renaming them all to this:

```
Get-DbaErrorLog
Get-DbaErrorLogConfig
Set-DbaErrorLog
```

I think that naming wouldn't confuse to many folks.

potatoqualitee:
ohhhhh is that what fred offered? i actually wondered about the loading - get-dbasqlmanagementobject (valid use of sql there) always showed gac but i dont understand how things work so i figured it was my issue.

lemme try this on my win7 x64

potatoqualitee:
I agree, @wsmelton - i believe ive even looked up errorlog before and never found it

potatoqualitee:
so the ErrorLog naming makes even more sense

wsmelton:
Name changes made for error log commands from `DbaSqlLog` to `DbaErrorLog`

potatoqualitee:
sweeeeeeet baw! thank you thank you thank you

potatoqualitee:
oh nice, even has an updated manifest! yesssssss thank you

PR #3697, opened 2018-06-19
Updated Set-DbaErrorLog

potatoqualitee:
made some updates to the command and the tests. simplified output and names. updated help to be more accurate. required sqlinstance. added dbasize

note that the parameter validation did not work so i used the native

image

### DIFF
--- a/functions/Get-DbaErrorLogConfig.ps1
+++ b/functions/Get-DbaErrorLogConfig.ps1
@@ -4,7 +4,8 @@ function Get-DbaErrorLogConfig {
             Pulls the configuration for the ErrorLog on a given SQL Server instance
         .DESCRIPTION
             Pulls the configuration for the ErrorLog on a given SQL Server instance.
-            Includes number of log files configured and size in KB (SQL Server 2012+ only)
+
+            Includes number of log files configured and size (SQL Server 2012+ only)
 
         .PARAMETER SqlInstance
             The target SQL Server instance(s)
@@ -34,7 +35,7 @@ function Get-DbaErrorLogConfig {
             Returns error log configuration for server2017 and server2014
     #>
     [cmdletbinding()]
-    param(
+    param (
         [Parameter(ValueFromPipelineByPropertyName)]
         [DbaInstanceParameter[]]$SqlInstance,
         [PSCredential]$SqlCredential,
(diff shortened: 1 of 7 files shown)


potatoqualitee:
Simplified output

image

potatoqualitee:
🤔 is there a way to modify the path or is that just sqlwmi?

functions/Set-DbaErrorLog.ps1
```diff
@@ -35,130 +35,98 @@ function Set-DbaErrorLog {
...
+        [int]$LogCount,
+        [int]$LogSize,
         [switch]$EnableException
     )
     process {
         foreach ($instance in $SqlInstance) {
             Write-Message -Level Verbose -Message "Connecting to $instance"
             try {
-                $server = Connect-SqlInstance -SqlInstance $instance -SqlCredential $SqlCredential
+                $server = Connect-SqlInstance -SqlInstance $instance -SqlCredential $SqlCredential -MinimumVersion 11
```
wsmelton: You cannot use this.

wsmelton: Setting the log file count is supported for any version of SQL Server (2000+). The only thing restricted to 2012+ is the size.

functions/Set-DbaErrorLog.ps1
```diff
@@ -35,130 +35,98 @@ function Set-DbaErrorLog {
...
+            
+            $currentNumLogs = $server.NumberOfLogFiles
+            $currentLogSize = $server.ErrorLogSizeKb
+            
+            $collection = [PSCustomObject]@{
+                ComputerName              = $server.NetName
+                InstanceName              = $server.ServiceName
+                SqlInstance               = $server.DomainInstanceName
+                LogCount                  = $currentNumLogs
+                LogSize                   = [dbasize]($currentLogSize * 1024)
```
wsmelton: This change would also need to be done to the Get-DbaErrorLogConfig so the output matches.

wsmelton:
The path for errorlog is changed by using `Set-DbaStartupParameter`, that is the only way to change it without going directly to the registry.

functions/Get-DbaErrorLogConfig.ps1
```diff
@@ -25,7 +25,7 @@ function Get-DbaErrorLogConfig {
 
             Website: https://dbatools.io
             Copyright: (C) Chrissy LeMaire, clemaire@gmail.com
-            License: GNU GPL v3 https://opensource.org/licenses/GPL-3.0
+            License: MIT https://opensource.org/licenses/MIT
```
wsmelton: My bad, got this from template and didn't catch it was old. I'll go update the templates.

potatoqualitee: ty!

potatoqualitee:
got all passes!

PR #3701, opened 2018-06-19
Stop-Function  - more informative messages

potatoqualitee:
way more informative error messages

@FriedrichWeinmann - i know you weren't a fan of this approach, and admittedly it's hacky, but it really seems to work. 

image

I'd like to ensure this or an equivalent stays when the new messaging system is in place 🏹

### DIFF
--- a/functions/Copy-DbaAgentJob.ps1
+++ b/functions/Copy-DbaAgentJob.ps1
@@ -209,7 +209,7 @@ function Copy-DbaAgentJob {
                         }
                         catch {
                             $copyJobStatus.Status = "Failed"
-                            $copyJobStatus.Notes = $_.Exception.Message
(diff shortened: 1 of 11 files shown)


potatoqualitee:
It's so useful! 😍 

This last commit took a slightly different approach by adding an internal function /cc @FriedrichWeinmann 

image

FriedrichWeinmann:
Btw, don't worry about the new message system breaking this: The mechanism is already in place!

wsmelton:
So on command where we pass in the 3 or 4 deep inner exception to stop function this won't duplicate or will it wipe that out?

potatoqualitee:
@wsmelton - this only looks at $errorrecord and not $exception so i dont think it'll impact it? nevertheless, the "finally" part of this command defaults to what it's been this whole time

PR #3711, opened 2018-06-20
Snapshot command updates

potatoqualitee:
All kinds of pipability, along with SMO objects being returned instead of custom objects. Need to add examples and will be writing a blog post to explain.


potatoqualitee:
Thanks for the review, @niphlod ! I have not tested perf but can add it back if there's an issue. Can you propose a way to test?

We need to kill existing connections on the server cuz appveyor has a habit of holding on to msdb and CREATE DATABASE fails. I removed code that actually restarted the server (:O) and replaced it with what other commands have.

Thanks for noticing the if not appveyor, I just removed it, but it's boning my own tests now, so I expect my next commit to fail.

potatoqualitee:
Notes for me from slack

```
ake a server which is preferrably not local, throw in 50 dbs, make 4 of them restricted_access and 4 offline, then take a snap on one of the remaining online
end game is having  the command to take a snap on one database not dependant of any other database(s) around
```

niphlod:
no prob. as we have very few people actually writing test code, it's better to bash the head around till everything works as expected rather than spending time remembering why a particular test got excluded. Bonus points for avoiding server restarts, but even msdb "locally locked" is something showing up in appveyor coupled with dbatools rather than an exquisitly specific problem on appveyor. Prolly previous tests do not release properly connections...

potatoqualitee:
that change worked :D no more need to skip appveyor

potatoqualitee:
BONNNNNNNNNNN 🎊 🔥 🚒 👩‍🚒

PR #3720, opened 2018-06-21
Rename Test-DbaValidLogin, fixes #437

### LINKED ISSUE #437: Test-DbaValidLogin - output properties
I like this function a lot, it's very handy. And it is very fast too in my environment. Great work.

But, the object properties should be renamed to be clearer, and consistent with other functions.

My suggestion:

ComputerName
SqlInstance (split in ComputerName and SqlInstance)
Domain
LoginName (or SAMAccountName if the AD property used)
LoginType
ExistsinAD
EnabledinAD
EnabledinSqlServer

maybe 'PasswordIsExpired' could be of interest too

I would definitely choose for two [enabled] or two [disabled] properties, even if the original is one time enabled = true, and the other one disabled = false.
This is exactly where dbatools can add value: a weak point in the MS methodology that we can make better and consistent for users.
Another option is to have [ADstatus] and [SqlStatus] as 'enabled' or 'disabled' instead of true and false

I don't know what the ADLoginDetails column is for, what I see is like a concatenation of  other properties? If they are useful there, than they can be more useful as separate properties, no?

The values of [LoginType] could remain what they are: 'Windows User', 'Windows Group',... That would be what DBA's are used to. So there's no need to manipulate those.

A general  way to go is to replace all -Detailed switches with -Simple. Default we'll always show all we have, and with -Simple we reduce the output to the essentials
 ( as a shortcut to the most probable ' | Select-Object a, b, c')

- [ ] Rename output properties
- [ ] Replace -Detailed with -Simple
- [ ] Add PasswordIsExpired
- [ ] Change Login Type values to original name

niphlod:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #437)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
This rename was long due.
/cc @gbargsley to rename the relevant page on dbatools.io
/cc @potatoqualitee to update properly manifest

### DIFF
--- a/functions/Test-DbaWindowsLogin.ps1
+++ b/functions/Test-DbaWindowsLogin.ps1
@@ -1,7 +1,8 @@
-function Test-DbaValidLogin {
+#ValidationTags#Messaging,FlowControl,Pipeline,CodeStyle#
+function Test-DbaWindowsLogin {
     <#
         .SYNOPSIS
-            Test-DbaValidLogin finds any logins on SQL instance that are AD logins with either disabled AD user accounts or ones that no longer exist
+            Test-DbaWindowsLogin finds any logins on SQL instance that are AD logins with either disabled AD user accounts or ones that no longer exist
 
         .DESCRIPTION
             The purpose of this function is to find SQL Server logins that are used by active directory users that are either disabled or removed from the domain. It allows you to keep your logins accurate and up to date by removing accounts that are no longer needed.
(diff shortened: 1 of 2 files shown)


potatoqualitee:
thanks @niphlod ! Should this be named anything else like DomainLogin? thinking xplat.

niphlod:
This is what was ruled a long time ago,  see #437. I'm unaware of further
"developments" on the naming side of things

Il lun 25 giu 2018, 12:28 Chrissy LeMaire <notifications@github.com> ha
scritto:

gbargsley:
The dbatools.psd1 does not have the name updated?

gbargsley:
Or what branch should I be pulling from?

niphlod:
This is the branch but modifications to manifest and module core are only
allowed to miss potatoqualitee ;-)

Il lun 25 giu 2018, 19:57 Garry Bargsley <notifications@github.com> ha
scritto:

potatoqualitee:
makes sense since it's doing a bunch of ad queries, thanks @niphlod.

@gbargsley - it'll be there in the next release, i'll ping you 👍

PR #3727, opened 2018-06-22
Adding time filtering before scanning for Maintenance Plan backups

### LINKED ISSUE #3704: Restore-DbaDatabase, smarter checks when -MaintenanceSolution is passed
---
name: Feature request/idea :bomb:
about: Suggest a new feature or enhancement

---

### Summary of new feature

* When there are multiple full backup files in the directory there should exist a flag so that the most recent is the one picked for reading headers.  This is using the -MaintenanceSolution flag.  

We've run into the error: 

BackupDiskFile::OpenMedia: Backup device '\\a\storage\path\adb\LOG\adb_LOG_20180529_073009.trn' failed to open. Operating system error 2(The system cannot find the file specified.).

The file does not exist, because it is outside the retention window for whatever backup file is being read.  If the most recent file was read log backups would still exist.

* In recent testing we found that multiple full backup files will cause some issues where all the files are read.  For example restoring the transaction logs could fail if your retention time is not encompassing the entirety of the headers on each backup file?  (still a little foggy on this)

If a new flag could be introduced that gets the most recent backup file based on date created that might be useful to avoid this situation.

* latest

Stuart-Moore:
## Type of Change

 - [ ] Bug fix (non-breaking change, fixes #3704)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Stop the whining :wink:

### Approach
Use the one thing we can trust to filter the files (don't have file creation for XpDirTree so that's out) which is pattern matching on the filename. Only works if people haven't played with Ola's scripts as it relys on it being yyyyMMdd_hhmmss

### Commands to test
Example in Restore-DbaDatabase

### DIFF
--- a/functions/Get-DbaBackupInformation.ps1
+++ b/functions/Get-DbaBackupInformation.ps1
@@ -57,9 +57,10 @@ function Get-DbaBackupInformation {
         .PARAMETER MaintenanceSolution
             This switch tells the function that the folder is the root of a Ola Hallengren backup folder
 
-        .PARAMETER RestoreSince
+        .PARAMETER RestoreTime
             This parameter ONLY works with the MaintenanceSolution switch
             If both are specified we will pre filter the files before reading the headers. This is not guaranteed to always work if you've moved files, or they contain multiple backups or you've altered any of the default naming rules in Ola's scripts
+        
         .PARAMETER IgnoreLogBackup
             This switch only works with the MaintenanceSolution switch. With an Ola Hallengren style backup we can be sure that the LOG folder contains only log backups and skip it.
             For all other scenarios we need to read the file headers to be sure.
@@ -146,7 +147,7 @@ function Get-DbaBackupInformation {
         [switch]$DirectoryRecurse,
         [switch]$EnableException,
         [switch]$MaintenanceSolution,
-        [DateTime]$RestoreSince,
+        [DateTime]$RestoreTime,
         [switch]$IgnoreLogBackup,
         [string]$ExportPath,
         [string]$AzureCredential,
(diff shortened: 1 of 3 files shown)


niphlod:
thanks for the whiner but .... I need to reiterate the bad behaviour ;-) 
ahum..... this is not quite what I'd have expected....
We all agreed no new switch were needed.... to filter files -since, is not a solution.

What you have is the list of names when called remotely, coupled with the creation time extrapolated from the filename....And if you have local files, you can have the LastWriteTime which is the end date.

Given *some* level of trust has been "granted", we can assume creation time / filename is the start date of the backup and LastWriteTime, if feasible, is the end date.

If so....
- for files contained in the FULL, you can select the latest whose creation date is not after the requested PITR
- for files contained in DIFF, same thing
- for files contained in LOG..... if you had consistently the end date you could go and skip anything older than the lastwritetime of the DIFF/FULL, and if you just have the creation date .... maybe we can safely skip whatever has been created BEFORE the FULL was created ?

Stuart-Moore:
Just me being curmudgeonly over the use of non-LSNs, I've done enough whining on the other side of that argument

This starts putting Select-DbaBackupInformation work into Get-DbaBackupInformation which pollutes the design somewhat. And we'd probably need to duplicate the testing backwards as well to make sure we catch the issues we catch elsewhere

As we've got to scan the files at some point to make sure we've got everything, and then contain exactly what we expect, allowing people to at least shave off the bottom set  will save a lot of header reading for people who only clean out every week

Thinking aloud, to do it the other way
Would need to read in all the fileinfo
Go through the object and add a property with a proper date in it.
Then pick out the last full before
Go back and pick out the other stripe members
Then find the diff
Stripe members
Logs
Stripe members
Then scan that lot anyway to make sure we've got a full tree 

The logic in pinching them between a lower and upper bound seems much simpler and less likely to cause issues with the oddball scenarios we come across.

I'll know something up, but already typing it up feels like repeating already done and working code.

Zhunya:
At this point, I recommend write custom script to gather all needed files for restore by user, and then supply it to Get-DbaBackupInformation.

niphlod:
The fact that the current code and function were made with "trust only the LSNs" is clear. 
The fact that having a way to avoid scanning the whole set of files is clear too.
The fact that the two are difficult to integrate at this point, is probably clear to you and everyone who dealt as a developer on those functions.
The fact that Restore-DbaDatabase is the highest level a user uses is clear too, so whatever needs to be done, Restore-DbaDatabase is the function that should "have the feature", even if imported from somewhere else in "internal".

BUT (because a whiner can't finish a sentence without it)

All of the above do not count for yet another quick and dirty shortcut/parameter which is gonna confuse users even more.

Stuart-Moore:
Last push was meant to go to personal repo, but it's the route I'm looking at for this if you want a peak. 

But it's not properly plumbed into Restore yet so just try Get-DbaBackupInformation and image RestoreSince is RestoreTime (i'll rename it when I polish it up.).

And it's not pretty as it's in the debug stage.

potatoqualitee:
ohh snap, new changes! once these get approved by @niphlod or @Zhunya, I'll merge 👍

potatoqualitee:
shitttttt i should have built upon this branch.

Stuart-Moore:
This wasn't complete and is a potential breaking change

Due to the way I was asked to implement this EVERYONE who uses maintenance
solution and restore time is immediately using a new work flow without
being able to opt out of it.

Sorry been busy and trying to belt and brace it before submitting

Stuart-Moore:
Yep. That was what I was trying to belt and brace around, but it was taking to much effort to catch everything and test it. 

 Removal is the easiest option, should be easy to deprecate. First commit just throws a non terminating error that the switch will be removed and we'll drop to standard. Later commit removes the warning.

Other option is to implement some way for users to add their templates and we parse those. Can't see that being pretty, easy or unlikley to cause a lot of 'issues' to be raised when they find the one instance in their estate that's different.

Getting really silly  we could scan history and reverse engineer the templating. Would be cool, but would be very fragile

An inital filtering step (yet more complexity) that does what @Zhunya says he does might be an option, puts the emphasis on the user to understand their own setup. But, as most of that can be done with Get-ChildItem and Where-Object or other PS standards, do we want to reinvent that wheel?

If people are worried about the time taken to read headers we can push the idea of using Get-DbaBackupInformation to do a regular scan and then store it in clixml ready for import (doesn't do header reads in that case as it's a BackupHistory object)

PR #3739, opened 2018-06-27
Add Select-DbaObject, Teach strings to DateTime, update instance parsing

FriedrichWeinmann:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Updates

 - Adds command `Select-DbaObject`, which works like `Select-Object`, only it can use some simple SQL Statements, rather than having to go back to complex hashtables each time.
 - All dbatools DateTime and Timestamp types now understand string input
 - `[DbaInstance]` now understands instance-qualified `.`-instance notation (e.g. `.\name`)

### Select-DbaObject

A few examples:

```powershell
# Rename property
$obj | Select-DbaObject Name, "Length as Size"

# Change property type
Import-Csv .\file.csv | Select-DbaObject Name, "Length as Size to DbaSize"

# Selecting values from other objects
$obj = [PSCustomObject]@{ Name = "Foo" }
$obj2 | Select-DbaObject FullName, Length, "Name from obj"

# Select values from the right object in another list of objects
$list = @()
$list += [PSCustomObject]@{ Type = "Foo"; ID = 1 }
$list += [PSCustomObject]@{ Type = "Bar"; ID = 2 }
$obj | Select-DbaObject Type, "ID as ID from list WHERE Type = Name"
```

### DbaInstanceParameter

Can now take things like this:
```powershell
[DbaInstance]".\foo"
```
but still loathes this:
```powershell
[DbaInstance]".\select"
```


potatoqualitee:
tyvm!

PR #3741, opened 2018-06-28
Install-DbaMaintenanceSolution - Fixing issue with backup location not being replaced correctly

Splaxi:
## Type of Change

### Purpose
 
Fix issue with the current cmdlet that doesn't replace the backup location variable correctly when being run. When the SQL Agent jobs are being created the backup location is set to NULL instead of the supplied value from the cmdlet.

### Approach

Using the same approach as other parts of the cmdlet to replace the built in variables in the script

### Commands to test

Install-DbaMaintenanceSolution -SqlInstance . -Database master -BackupLocation "J:\MSSQL_BACKUP" -InstallJobs -ReplaceExisting -Solution All

### Screenshots

image

### DIFF
--- a/functions/Install-DbaMaintenanceSolution.ps1
+++ b/functions/Install-DbaMaintenanceSolution.ps1
@@ -216,8 +216,8 @@ function Install-DbaMaintenanceSolution {
 
             # Backup location
             if ($BackupLocation) {
-                $findBKP = 'C:\Backup'
-                $replaceBKP = $BackupLocation
+                $findBKP = 'SET @BackupDirectory     = NULL'
+                $replaceBKP = 'SET @BackupDirectory     = N''' + $BackupLocation + ''''
                 foreach ($file in $listOfFiles) {
                     (Get-Content -Path $file -Raw).Replace($findBKP, $replaceBKP) | Set-Content -Path $file
                 }


potatoqualitee:
Ohhhh whaddap!

allcommands.ps1
```diff
@@ -46001,8 +46001,8 @@ function Install-DbaMaintenanceSolution {
 
             # Backup location
             if ($BackupLocation) {
-                $findBKP = 'C:\Backup'
-                $replaceBKP = $BackupLocation
+                $findBKP = 'SET @BackupDirectory     = NULL'
```
potatoqualitee: Thanks so much for the PR! 😍 

I combine all changes to the allcommands.ps1 during publish so you do not have to worry about this particular file.

functions/Install-DbaMaintenanceSolution.ps1
```diff
@@ -216,8 +216,8 @@ function Install-DbaMaintenanceSolution {
 
             # Backup location
             if ($BackupLocation) {
-                $findBKP = 'C:\Backup'
-                $replaceBKP = $BackupLocation
+                $findBKP = 'SET @BackupDirectory     = NULL'
+                $replaceBKP = 'SET @BackupDirectory     = N''' + $BackupLocation + ''''
```
potatoqualitee: Mmm what if BackupLocation is empty? I'd do a 

`if ((Test-Bound -ParameterName BackupLocation))`

Splaxi: Already part of the code, earlier up. ;)

if ((Test-Bound -Parameter BackupLocation -Not)) {
                $BackupLocation = (Get-DbaDefaultPath -SqlInstance $server).Backup
            }

allcommands.ps1
Splaxi: Roger - just wanted to make sure. Will try and remember that ;)

functions/Install-DbaMaintenanceSolution.ps1
potatoqualitee: aw, man. my bad. i looked at the code at the top but didnt look within the body!

potatoqualitee:
bon, merci! 💃

PR #3757, opened 2018-07-03
Make Get-DbaDefaultPath skip Azure SQLDB Instances

johnmart82:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Experienced an issue with Install-DbaMaintenanceSolution on Azure SQLDB, issue was due to allcommands implementation of Get-DbaDefaultPath. Azure SQLDB does not return paths so there are errors.

install-dbamaintenancesolutionred

This makes the errors go away.

### Approach
Implemented a check on $server.EngineEdition to find if the target is SqlDatabase and to not perform checks if true. This will still work with Managed Instance as that returns paths that can be parsed by the function.

### Learning
Tested revised function against Local SQL Server, Managed Instance, and Azure SQLDB manually.
get-dbadefaultpath-sqltest

get-dbadefaultpath-mitest

get-dbadefaultpath-dbtest


potatoqualitee:
yessss, looks fab. thank you! 🌮 🌯

PR #3762, opened 2018-07-04
Azure enhancements for Backup & Restore

potatoqualitee:
Added some encrypted values so that we can test against Azure within Appveyor. During the course of this, I also fixed some bugs and added more switches as required.


potatoqualitee:
I hate code cov :( This is a test-centric PR. Codecov is out of its mind.

niphlod:
if there's no env variable for credentials the test should be skipped, else this test cannot be run, e.g., on my test rig...

functions/Invoke-DbaAdvancedRestore.ps1
```diff
@@ -234,13 +234,17 @@ function Invoke-DbaAdvancedRestore {
                     Write-Message -Message "Adding device $file" -Level Debug
                     $Device = New-Object -TypeName Microsoft.SqlServer.Management.Smo.BackupDeviceItem
                     $Device.Name = $file
-                    if ($file -like "http*") {
+                    if ($file -match 'http') {
```
niphlod: and if my database name is I_love_HTTP_appname ..... this breaks because it searches for http throughout all the string ... before was strictly more correct, searching just at begin

functions/Backup-DbaDatabase.ps1
```diff
@@ -416,6 +428,16 @@ function Backup-DbaDatabase {
                     else {
                         $device.DeviceType = "File"
                     }
+                    
+                    if ($WithFormat) {
+                        Write-Message -Message "WithFormat specified. Ensuring Initialize and SkipTapeHeader are set to true." -Level Verbose
```
niphlod: more relevant for a debug level rather than verbose ....

potatoqualitee: it seems important tho - cuz it changes people's settings even though they chose not to. I felt like it was either a warning / prompt or verbose and decided on verbose. You dont think it's important?

functions/Invoke-DbaAdvancedRestore.ps1
potatoqualitee: LOL k ty

potatoqualitee: changed it to StartsWith

functions/Backup-DbaDatabase.ps1
niphlod: I think we should log by default to a MINIMUM since we don't have any "knob" to tune logging verbosity. if it's so important because it changes people's settings it should warn and stop. if it's a requirement (as it seems written in code), either you go ahead and "do the right thing" 'cause not doing it results always in an error (but doing so, there's no reason to provide more than one additional switch) or you relinquish responsibilities stopping if the user isn't specifically asking for the combination that do NOT result in an error.

tests/Backup-DbaDatabase.Tests.ps1
```diff
@@ -23,13 +23,22 @@ Describe "$commandname Integration Tests" -Tags "IntegrationTests" {
         }
         Get-DbaDatabase -SqlInstance $script:instance1 -Database "dbatoolsci_singlerestore" | Remove-DbaDatabase -Confirm:$false
         Get-DbaDatabase -SqlInstance $script:instance2 -Database $DestDbRandom | Remove-DbaDatabase -Confirm:$false
+        $server = Connect-DbaInstance -SqlInstance $script:instance2
+        $sql = "CREATE CREDENTIAL [https://dbatools.blob.core.windows.net/sql] WITH IDENTITY = N'SHARED ACCESS SIGNATURE', SECRET = N'$env:azurepasswd'"
```
niphlod: here container is fixed....it'd be good to make it a variable . Also, if there are no details about a container and the relative credentials, test should be skipped as already done in the other test file

functions/Backup-DbaDatabase.ps1
potatoqualitee: we do have knobs via `Set-DbaConfig` 👯  (tho i dont recall which one but fred showed me before)

tests/Backup-DbaDatabase.Tests.ps1
potatoqualitee: this testdoes have `if ($env:azurepasswd) {` right above. good suggestion about making it flexible. I'd like to save that for a later PR 😴

functions/Backup-DbaDatabase.ps1
niphlod: yeah but not on the function itself, so logging anything to verbose and expecting users to "ingest" loads of not-so-important-messages via a single -Verbose switch isn't feasible

tests/Backup-DbaDatabase.Tests.ps1
niphlod: nope ... beforeall is not concerned about $env:azurepasswd

potatoqualitee: oh good call, ty, iwll fix

functions/Backup-DbaDatabase.ps1
potatoqualitee: i tried to change it to debug but my brain insisted it was fasho verbose 🤪

PR #3779, opened 2018-07-11
Write-DbaDataTable - Removed -RegularUser requirement

potatoqualitee:
Fsck that requirement to heck.


JanMulkens:
I fully agree to this change 👍 
Thanks Chrissy!

niphlod:
sounds good to remove everywhere if we catch it, a tandem blogpost on how to handle connections via connect-dbainstance is long due :P

PR #3781, opened 2018-07-11
PlanCache updates, renames and tests

potatoqualitee:
/cc @tboggiano for fysa

image

image

### DIFF
--- a/functions/Get-DbaPlanCache.ps1
+++ b/functions/Get-DbaPlanCache.ps1
@@ -56,9 +56,11 @@
             [switch]$EnableException
         )
         begin {
-            $Sql = "SELECT MB = sum(cast((CASE WHEN usecounts = 1 AND objtype IN ('Adhoc', 'Prepared') THEN size_in_bytes ELSE 0 END) as decimal(12, 2))) / 1024 / 1024,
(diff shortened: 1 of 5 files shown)


functions/Get-DbaPlanCache.ps1
```diff
@@ -72,14 +74,10 @@
                 }
 
                 $results = $server.Query($sql)
+                $size = [dbasize]($results.MB*1024*1024)
```
wsmelton: Why not just return the raw value of bytes instead of pulling it as MB (which converted it from bytes) and then converting it back?

wsmelton: Also I would remove the decimal cast in the query.

potatoqualitee: easier in the refactor ;)

wsmelton: Don't think I follow... You mean that type expects MB as input to allow it to refactor to the other units?

potatoqualitee: yah the clear plancache relies on it so i just kept it and hid it

potatoqualitee:
this has a file i need to have fixed in dev so i merged 👎

PR #3782, opened 2018-07-11
Test mods + Export-DbaLogin now outputs Get-ChildItem

potatoqualitee:
NBD test updates that silence warnings, but Fred will fix that. Ultimately, I like this PR cuz it changes the behavior of Export-DbaLogin.


potatoqualitee:
unrelated failures

PR #3794, opened 2018-07-13
set a high threshold for codecov

niphlod:
this should make the codecov test always pass (it fails just if coverage drops by 50%, in which case, I'd say there's a problem).


potatoqualitee:
cool, ill reenable it too, thank you!

PR #3812, opened 2018-07-17
NetName -> ComputerName for Azure Support

potatoqualitee:
Azure does not support NetName, which we rely on for ComputerName.

This adds a new property to $server in both Connect-SqlInstance and Connect-DbaInstance. It checks to see if NetName exists and if it does, then that will be used, otherwise, it'll use `[dbainstance]`'s ComputerName property.


potatoqualitee:
@PowerDBAKlaas you will appreciate 👍

potatoqualitee:
works on prem 
image

works in azure
image

PR #3815, opened 2018-07-17
fix from codecov support

niphlod:
straight out of codecov support to avoid codecov posting back to github


potatoqualitee:
tytyyttytytytytytty

PR #3825, opened 2018-07-18
Get-DbaSsisExecutionHistory update

potatoqualitee:
will start a review to show what i changed and why, one moment

### DIFF
--- a/functions/Get-DbaSsisExecutionHistory.ps1
+++ b/functions/Get-DbaSsisExecutionHistory.ps1
@@ -6,6 +6,15 @@ function Get-DbaSsisExecutionHistory {
 
         .DESCRIPTION
             This command gets execution history for SSIS executison given one or more instances and can be filtered by Project, Environment,Folder or Status.
+        
+        .PARAMETER SqlInstance
+            SQL Server name or SMO object representing the SQL Server to connect to.
+            This can be a collection and receive pipeline input to allow the function
+            to be executed against multiple SQL Server instances.
+
+        .PARAMETER SqlCredential
+            Login to the target instance using alternative credentials. Windows and SQL Authentication supported. Accepts credential objects (Get-Credential)
+
         .PARAMETER Project
             Specifies a filter by project
 
--- a/functions/Get-DbaSsisExecutionHistory.ps1
+++ b/functions/Get-DbaSsisExecutionHistory.ps1
@@ -59,7 +59,6 @@ function Get-DbaSsisExecutionHistory {
         [String[]]$Project,
         [String[]]$Folder,
         [String[]]$Environment,
-        [Switch]$Force,
         [Alias('Silent')]
         [switch]$EnableException
     )
--- a/functions/Get-DbaSsisExecutionHistory.ps1
+++ b/functions/Get-DbaSsisExecutionHistory.ps1
@@ -49,12 +49,12 @@ function Get-DbaSsisExecutionHistory {
 
             Shows what would happen if the command were executed and would return the SQL statement that would be executed per instance.
     #>
-    [CmdletBinding(DefaultParameterSetName = "Default", SupportsShouldProcess = $true)]
+    [CmdletBinding()]
     param (
         [parameter(Mandatory = $true)]
-        [DbaInstanceParameter]$SQLInstance,
+        [DbaInstanceParameter]$SqlInstance,
         [PSCredential]$SqlCredential,
-        [ValidateSet("Created", "Running", "Cancelled", "Failed", "Pending", "Halted", "Succeeded", "Stopping","Completed")]
+        [ValidateSet("Created", "Running", "Cancelled", "Failed", "Pending", "Halted", "Succeeded", "Stopping", "Completed")]
         [String[]]$Status,
         [String[]]$Project,
         [String[]]$Folder,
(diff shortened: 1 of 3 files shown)


functions/Get-DbaSsisExecutionHistory.ps1
```diff
@@ -49,12 +49,12 @@ function Get-DbaSsisExecutionHistory {
 
             Shows what would happen if the command were executed and would return the SQL statement that would be executed per instance.
     #>
-    [CmdletBinding(DefaultParameterSetName = "Default", SupportsShouldProcess = $true)]
```
potatoqualitee: parametersets are only required when you have conflicting params, like..

"If you run Copy-DbaDatabase, you can run EITHER backuprestore OR detachattach"  or "database is ONLY mandatory when..."

SupportsShoudlProcess is for whatif and is only used when there's modification to settings/data/etc.

functions/Get-DbaSsisExecutionHistory.ps1
```diff
@@ -49,12 +49,12 @@ function Get-DbaSsisExecutionHistory {
...
+    [CmdletBinding()]
     param (
         [parameter(Mandatory = $true)]
-        [DbaInstanceParameter]$SQLInstance,
+        [DbaInstanceParameter]$SqlInstance,
         [PSCredential]$SqlCredential,
-        [ValidateSet("Created", "Running", "Cancelled", "Failed", "Pending", "Halted", "Succeeded", "Stopping","Completed")]
+        [ValidateSet("Created", "Running", "Cancelled", "Failed", "Pending", "Halted", "Succeeded", "Stopping", "Completed")]
```
potatoqualitee: My editor did that ;)

functions/Get-DbaSsisExecutionHistory.ps1
```diff
@@ -49,12 +49,12 @@ function Get-DbaSsisExecutionHistory {
 
             Shows what would happen if the command were executed and would return the SQL statement that would be executed per instance.
     #>
-    [CmdletBinding(DefaultParameterSetName = "Default", SupportsShouldProcess = $true)]
+    [CmdletBinding()]
     param (
         [parameter(Mandatory = $true)]
-        [DbaInstanceParameter]$SQLInstance,
+        [DbaInstanceParameter]$SqlInstance,
```
potatoqualitee: our standard, like Microsoft's is to use Sql and not SQL within PS vars

functions/Get-DbaSsisExecutionHistory.ps1
ChrisTuckerNM: Cute

functions/Get-DbaSsisExecutionHistory.ps1
```diff
@@ -65,50 +65,54 @@ function Get-DbaSsisExecutionHistory {
...
-
-            if ($Environment) {
-                $csv = "`'" + ($Environment -join"'","'") + "`'"
-                $environmentq = "AND e.[environment] in ($csv)"
-            } else {
-                $environmentq = ''
-            }
-
-            $q = @"
+            'Created'   = 1
```
potatoqualitee: caps cuz that's what we usually see from PowerShell results.

functions/Get-DbaSsisExecutionHistory.ps1
```diff
@@ -65,50 +65,54 @@ function Get-DbaSsisExecutionHistory {
...
+        else {
+            $environmentq = ''
+        }
+        
+        $sql = "
             WITH
             cteLoglevel as (
                 SELECT
-                    execution_id,
-                    cast(parameter_value AS INT) AS logging_level
```
potatoqualitee: logging_level is SQLy, while LoggingLevel is PowerShelly so I changed the output to look more like PS and less like SQL

potatoqualitee:
@ChrisTuckerNM - can you test this against a server that has some executions on it, making sure to include various params like status, etc? ty

niphlod:
good now, except for no parametrized queries. Lemme know when I can chime in, so everyone can see how easy it is

ChrisTuckerNM:
I tested all parameters except for credentials and multiple instances... Is there an issue?

niphlod:
yeah, even if theoretically a user dabbling with dbatools is a "sane" user, constructing the query via string interpolation is susceptible to SQL Injection......

ChrisTuckerNM:
Oops, @potatoqualitee IS there an issue? I tested with all parameters against an instance with tons of executions, including failures and cancels.

ChrisTuckerNM:
@niphlod yeah, that last comment was meant for CL, not you, I agree on the injection, I was gonna have a crack at it but fire away..

niphlod:
I use this snippet, it's easily sprinkable throughout this module. If you have questions about implementing it ask away

```
function Get-SqlOrIfied([object[]]$element, [string]$keyprefix) {
            $stmt = ''
            $kv = [ordered]@{}
            for ($i=0; $i -lt $element.length; $i++) {
	            $kv["@$($keyprefix)_$($i)"] = $element[$i]
            }
            $stmt = '(' + ($kv.Keys -Join ', ') + ')'
            [pscustomobject]@{
                stmt = $stmt
                params = $kv
            }
        }
        ......
        $query_params = @{}

        if ($Status) {
            $piece = Get-SqlOrIfied -element $statuses[$Status] -keyprefix 'status'
            $query_params += $piece.params 
            $statusq = "AND e.[Status] IN $($piece.stmt)"
        }
        else {
            $statusq = ''
        }
```

niphlod:
Note for later, we definitely need a -Since parameter (and maybe with a week old default that trims at a sane level the retrieved history)

potatoqualitee:
this is breakin all my tests so imma merge

wsmelton:
@niphlod unless you have been going through updating commands the method used in this command for dynamic T-SQL is how it is done in every command we have...and I mean every one literally.

niphlod:
@wsmelton : and that's exactly why we should change. BTW, it's not "safe by default" but others supposedly "interpolate" parameters that come from autocomplete, and things such databasename . Here there are several parameters "in free text" so it's more "urgent"

PR #3828, opened 2018-07-19
Get-DbaDiskSpace CheckForSql Mount Points

awickham10:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
When using CheckForSql mount points cause errors. 

### Approach

Change drive letter checking to check paths. Given a path D:\SQL\DataFiles\ and a mount point D:\SQL\ it is on a mount point.

### Commands to test

Create a mount point for a drive and execute Get-DbaDiskSpace using CheckForSql.

### Screenshots

Before
image

After
image

### DIFF
--- a/functions/Get-DbaDiskSpace.ps1
+++ b/functions/Get-DbaDiskSpace.ps1
@@ -161,14 +161,18 @@ function Get-DbaDiskSpace {
                         Write-Message -Level VeryVerbose -Message "Processing instance $($instanceName)"
                         try {
                             $server = Connect-SqlInstance -SqlInstance $instanceName -SqlCredential $SqlCredential
-
-                            $files = @()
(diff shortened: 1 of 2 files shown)


awickham10:
I think the "CheckForSql returns IsSqlDisk property with a value (likely false)" test is failing now because it was never properly working. I believe all of the disks on the Appveyor box have a SQL data or log file on them (C:\Program Files for system DBs, C:\Mounts probably as well). I'm not 100% on that though... can someone else confirm?

If there is SQL data or log on all drives how about mocking the Get-DbaCmObject function and returning a fake drive?

potatoqualitee:
thanks, going to merge then skip that portion of the test. I'd love a mock, would you be able to sync the dev branch and create it? 🎖

potatoqualitee:
also, this worked well for me (it always has, even with mounts, so im not sure what the diff is but appreciate your fix)

image

potatoqualitee:
Oh, I see now. I was missing the param. Good catch, thanks again!

awickham10:
Opened up PR #3859 with mocking to fix the test

Issue #3820, opened 2018-07-18
Invoke-DbaDatabaseShrink - Math for PercentFreeSpace doesn't work per fileType


jpomfret:
### Before submitting a bug report:

- [X] Ensure you are able to reproduce it on the latest released version (_we release often_)
- [X] Verified this bug is not already reported in an issue
- [X] Verified errors are not related to permissions
- [X] Can reproduce in a clean PowerShell session (_clean = `powershell -NoProfile`_)

## Steps to Reproduce

```powershell
    Invoke-DbaDatabaseShrink -SqlInstance server -Database db-PercentFreeSpace 50 -FileType Log
```

## Expected Behavior
I would expect the 50% freespace to apply to just the log, not the total database size. When I run the command above (log is currently 14GB and 99% free) I would expect the log to be resized to 50% free.

## Actual Behavior
No shrink since freespace is just reported for the data files and that doesn't have 50% free.
![image](https://user-images.githubusercontent.com/981370/42882965-333e6e7a-8a68-11e8-82e9-f918617c9941.png)

## Environmental data

 - PowerShell: 
![image](https://user-images.githubusercontent.com/981370/42883078-899f2188-8a68-11e8-88b4-6a1f3f3521c8.png)

 - SQL Server: 
Microsoft SQL Server 2014 (SP2-CU11) (KB4077063) - 12.0.5579.0 (X64)   Feb 21 2018 12:19:47   Copyright (c) Microsoft Corporation  Developer Edition (64-bit) on Windows NT 6.3 <X64> (Build 9600: ) (Hypervisor)

Zhunya:
This can be closed?

potatoqualitee:
totally thank you

PR #3834, opened 2018-07-20
Invoke-DbaDbShrink - couple of issues resolved, and march towards 1.0

### LINKED ISSUE #3516: Invoke-DbaDbShrink - add incremental shrink for VLDBs
### Problem to solve
Database shrinkage has to occur at times. I'm involved in a client project now that we are going to be NULL'n out an NTEXT column of a log table that exist in 1400+ databases on a single instance. Some databases are medium size, in MB range, while others are well into GB range. Overall total space recovered from this is estimated to be close to a TB.

Now shrinking a file that is only 100GB down to 10% free space is not a big deal, as long as the IO can keep up. However when I get to a data file that is 800GB+ a single shrink operation will take forever since the DBCC operation is single-threaded.

I came across this tip on actually looping through and incrementally shrinking a given file. So proposal is to implement a parameter, say `SizeIncrement` maybe, that will have the shrink operation utilize until it reaches the final size/percentage desired.

### Additional information

MSSQLTip: https://www.mssqltips.com/sqlservertip/3178/incrementally-shrinking-a-large-sql-server-data-file-using-powershell

jpomfret:
## Type of Change

 - [X] Bug fix (non-breaking change, fixes #3516 #3820)
 - [X] Pester test is included
 - [ ] Nunit test is included
 - [X] Documentation
 - [ ] Build system
 

### Purpose
 
Breaking Changes:
- Changed function name to match naming standard of DbaDb instead of DbaDatabase
  - (I did not include .psm1 or .psd1 files in PR but I can if you want, added alias)
- changed output to include name of file for shrink, also output per file instead of one row of output for database

Other Changes:
- Fix issue #3820 - math issues with calculating the size to shrink to
- changed logic so it loops through files and calculates per file
- added UnitTests
- cleaned up CBH
- New parameter StepSizeMB added to address issue #3516 

### Commands to test

Examples are in the help

### DIFF
--- a/functions/Invoke-DbaDbShrink.ps1
+++ b/functions/Invoke-DbaDbShrink.ps1
@@ -153,6 +153,8 @@ function Invoke-DbaDbShrink {
             $FileType = 'Log'
         }
 
+        $StepSizeKB = ($stepSizeMB * 1024)
+
         $StatementTimeoutSeconds = $StatementTimeout * 60
 
         $sql = "SELECT
@@ -177,16 +179,10 @@ function Invoke-DbaDbShrink {
                 Stop-Function -Message "Failure" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
             }
 
-            # changing statement timeout to $StatementTimeout
-            if ($StatementTimeout -eq 0) {
-                Write-Message -Level Verbose -Message "Changing statement timeout to infinity"
-            }
-            else {
-                Write-Message -Level Verbose -Message "Changing statement timeout to $StatementTimeout minutes"
-            }
             $server.ConnectionContext.StatementTimeout = $StatementTimeoutSeconds
+            Write-Message -Level Verbose -Message "Connection timeout set to $StatementTimeout"
 
-            $dbs = $server.Databases | Where-Object { $_.IsSystemObject -eq $false -and $_.IsAccessible }
+            $dbs = $server.Databases | Where-Object {$_.IsAccessible}
 
             if ($Database) {
                 $dbs = $dbs | Where-Object Name -In $Database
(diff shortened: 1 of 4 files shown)


potatoqualitee:
fantastic :D!! thank you!

potatoqualitee:
would it be feasible to change the columns with MB to [dbadbsize] ?

potatoqualitee:
then remove MB from the column name

functions/Invoke-DbaDbShrink.ps1
```diff
@@ -0,0 +1,345 @@
...
+            }
+
+            # changing statement timeout to $StatementTimeout
+            if ($StatementTimeout -eq 0) {
+                Write-Message -Level Verbose -Message "Changing statement timeout to infinity"
+            }
+            else {
+                Write-Message -Level Verbose -Message "Changing statement timeout to $StatementTimeout minutes"
+            }
+            $server.ConnectionContext.StatementTimeout = $StatementTimeoutSeconds
```
wsmelton: I don't think this if/else block is really needed just for different verbose messages...I'd update this to just two statements, unless someone sees an issue:

```powershell
$server.ConnectionContext.StatementTimeout = $StatementTimeout
Write-Mesage -Level Verbose -Message "Connection timeout set to $StatementTimeout"
```
Just keep it simple since we have a default value set on the parameter.

functions/Invoke-DbaDbShrink.ps1
```diff
@@ -0,0 +1,345 @@
...
+
+            $dbs = $server.Databases | Where-Object { $_.IsSystemObject -eq $false -and $_.IsAccessible }
```
wsmelton: This prevents the ability of providing `tempdb` or `msdb` which are common system databases that tend to be shrunk in certain circumstances (e.g. clearing out years of backup history, or space issue on tempdb drive).

I would just remove this line all together.

functions/Invoke-DbaDbShrink.ps1
```diff
@@ -0,0 +1,345 @@
...
+                        Write-Message -Level Verbose -Message "File: $($file.Name)"
+                        Write-Message -Level Verbose -Message "Starting Size (MB): $([int]$startingSize)"
+                        Write-Message -Level Verbose -Message "Space Used (MB): $([int]$spaceUsed)"
+                        Write-Message -Level Verbose -Message "Starting Freespace (MB): $([int]$spaceAvailableMB)"
+                        Write-Message -Level Verbose -Message "Desired Freespace (MB): $([int]$desiredSpaceAvailable)"
+                        Write-Message -Level Verbose -Message "Desired FileSize (MB): $([int]$desiredFileSize)"
+                    }
+                    catch {
+                        $success = $false
+                        Stop-Function -message "Shrink Failed: $($_.Exception.InnerException)" -EnableException $EnableException -ErrorRecord $_ -Continue
```
wsmelton: This is wrong and should be updated...

Something like `Stop-Function -Message "Calculation failure" -Target $file -ErrorRecord $_`

1. All this seems to be doing is calculating values to be used in the shrink operation. 
2. If that actually does fail for some reason should not be allowed to continue the command. It could cause unintentional consequences.

functions/Invoke-DbaDbShrink.ps1
```diff
@@ -0,0 +1,345 @@
...
+                        continue
```
wsmelton: This line should be removed.

functions/Invoke-DbaDbShrink.ps1
```diff
@@ -0,0 +1,345 @@
...
+        [object[]]$Database,
+        [object[]]$ExcludeDatabase,
+        [switch]$AllUserDatabases,
+        [ValidateRange(0, 99)]
+        [int]$PercentFreeSpace = 0,
+        [ValidateSet('Default', 'EmptyFile', 'NoTruncate', 'TruncateOnly')]
+        [string]$ShrinkMethod = "Default",
+        [ValidateSet('All', 'Data', 'Log')]
+        [string]$FileType = "All",
+        [int]$StepSizeMB,
```
wsmelton: I agree with @potatoqualitee on this one to drop the MB

functions/Invoke-DbaDbShrink.ps1
jpomfret: I would agree, thanks!

functions/Invoke-DbaDbShrink.ps1
jpomfret: removed catch around calculating file sizes

functions/Invoke-DbaDbShrink.ps1
jpomfret: done 👍

functions/Invoke-DbaDbShrink.ps1
```diff
@@ -0,0 +1,334 @@
...
+                , max(avg_fragmentation_in_percent) as [max_fragmentation_in_percent]
+                FROM sys.dm_db_index_physical_stats (DB_ID(), NULL, NULL, NULL, NULL) AS indexstats
+                WHERE indexstats.avg_fragmentation_in_percent > 0 AND indexstats.page_count > 100
+                GROUP BY indexstats.database_id"
+    }
+
+    process {
+        if (!$Database -and !$ExcludeDatabase -and !$AllUserDatabases) {
+            Stop-Function -Message "You must specify databases to execute against using either -Database, -Exclude or -AllUserDatabases" -Continue
+        }
```
FriedrichWeinmann: This is broken code:
- Drop the `-Continue` parameter
- Add a `return` on the next line
- Move to `begin` block
- add the following line as the first line in process instead:

```powershell
```

Otherwise it can be a danger to the calling function

potatoqualitee:
awesome work, @jpomfret ! thank you. i'll update psm1 and psd1 prior to publish.

PR #3835, opened 2018-07-20
Connect commands and alt Windows Credential fix

potatoqualitee:
It sometimes said "untrusted domain" failure.

image

Now it doesn't. And I used some extended events to confirm I logged in as that other user.

image


potatoqualitee:
outdated branch caused failures, otherwise, got the thumbs up from 3 of us

PR #3902, opened 2018-08-18
Fixed QueryName list bugs

andrekamman:
Query name list was empty when not provided which caused an error when providing ExcludeQuery parameter.

And Selectionhelper would run everything when you press cancel, it now gives a message and stops running

## Type of Change

 - [X] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Query name list was empty when not provided which caused an error when providing ExcludeQuery parameter.
And Selectionhelper would run everything when you press cancel, it now gives a message and stops running

### Commands to test
Invoke-DbaDiagnosticQuery -SqlIntance localhost -UseSelectionhelper
Pressing cancel should give a message and should not run any query

Invoke-DbaDiagnosticQuery -SqlInstance localhost -ExcludeQuery "Missing Index Warnings"
Should run everything except the excluded query without giving an error message
(Before fix it would give error and it would not properly exclude the query)
A pester test is included for this in this pull request

### DIFF
--- a/functions/Invoke-DbaDiagnosticQuery.ps1
+++ b/functions/Invoke-DbaDiagnosticQuery.ps1
@@ -273,13 +273,22 @@ function Invoke-DbaDiagnosticQuery {
             if ($UseSelectionHelper -and $first) {
                 $QueryName = Invoke-DiagnosticQuerySelectionHelper $parsedscript
                 $first = $false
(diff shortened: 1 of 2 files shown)


potatoqualitee:
thanks so much, andre!

Issue #3886, opened 2018-08-10
Restore-DbaDatabase does not work on SQL 2000


pollusb:
### Before submitting a bug report:

- [x] Ensure you are able to reproduce it on the latest released version (_we release often_)
- [ ] Verified this bug is not already reported in an issue
- [x] Verified errors are not related to permissions
- [x] Can reproduce in a clean PowerShell session (_clean = `powershell -NoProfile`_)

## Steps to Reproduce

```powershell
Restore-DbaDatabase -SqlInstance SQL2000 -Path c:\temp\ABC.bak -DatabaseName ABC

Exception setting "DatabaseBackupLsn": "Cannot convert null to type "System.Numerics.BigInteger"."
At C:\Users\###\Modules\dbatools\0.9.380\allcommands.ps1:23941 char:17
+ ...             $historyObject.DatabaseBackupLsn = $group.Group[0].Databa ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], SetValueInvocationException
    + FullyQualifiedErrorId : ExceptionWhenSetting

The property 'Fullname' cannot be found on this object. Verify that the property exists and can be set.
At C:\Users\###\Modules\dbatools\0.9.380\allcommands.ps1:70268 char:17
+ ...             $full.Fullname = ($DatabaseHistory | Where-Object {$_.Typ ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
    + FullyQualifiedErrorId : PropertyNotFound

You cannot call a method on a null-valued expression.
At C:\Users\###\Modules\dbatools\0.9.380\allcommands.ps1:70287 char:17
+ ...             [bigint]$LogBaseLsn = ($dbHistory | Sort-Object -Property ...
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
    + FullyQualifiedErrorId : InvokeMethodOnNull

WARNING: [09:38:29][Restore-DbaDatabase] Failure | Error enumerating files
```

## Expected Behavior
restore database

[...]

PR #3918, opened 2018-08-23
Backup and restore with sql 2000

### LINKED ISSUE #3889: Backup-DbaDatabase not working with SQL Server 2000
image

I added the usual backup/restore suspects. If anyone doesn't want to be assigned, please DM me and I'll keep you off in the future 🌮

Stuart-Moore:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #3889 and #3886 )
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Backups weren't working with SQL 2k without a full path being passed in as 2k didn't support default path in the same way. So we're now providing it for it.

Restores were failing as RESTORE HEADERONLY doesn't read all the headers we need in 2k. There is a work around using a newer sql version to read the headers and then pass in the results as a backupHistory object to the restore process. This is documented in Help, and will be presented to teh user if they try to restore to 2k without passing in TrustDbBackupHistory

Copy-DbaDatabase works differently as we pull the info from msbd..backuphistory* which is there in 2k (even if we have another workaround for that as well.....)


potatoqualitee:
awesome! thanks for the new test and fixing the command itself 💯 💯 💯 💯

PR #3952, opened 2018-08-30
Backupcompress

### LINKED ISSUE #3941: Backup-DbaDatabase, add NO_COMPRESS for encrypted dbs
Had a comment:

> Whew!!  Found the problem.  The database that could not be read was encrypted.  It was backed up using Backup-dbaDatabase .... -Compress.  since it is the only encrypted DB we have, we specify Compress='Y' in Ole's backup job and it knows to switch to NO_COMPRESS for an encrypted DB.  Backup-dbaDatabase will successfully compress an encrypted DB, but it will be unusable.  That's usually but a good thing   :slightly_smiling_face:   You might consider putting some logic into Backup-dbaDatabase to switch to NO_COMPRESS no matter what the switch or server default is.  THAT was a mind bender.

Basically, compression will never work with encrypted dbs, so we should warn and stop if -Compress is used or pull an Ola and do NO_COMPRESS no matter what (and warn).

Stuart-Moore:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #3941 )
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Don't compress backups from encrypted databases as requested

### Approach
If db is encrypted, set CompressionOption to 2 (NO_COMPRESS) and alerts the user that we're ignoring their compression option

### DIFF
--- a/functions/Backup-DbaDatabase.ps1
+++ b/functions/Backup-DbaDatabase.ps1
@@ -296,7 +296,10 @@ function Backup-DbaDatabase {
             $Suffix = "bak"
 
             if ($CompressBackup) {
-                if ($server.Edition -like 'Express*' -or ($server.VersionMajor -eq 10 -and $server.VersionMinor -eq 0 -and $server.Edition -notlike '*enterprise*') -or $server.VersionMajor -lt 10) {
+                if ($database.EncryptionEnabled) {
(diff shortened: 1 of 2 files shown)


potatoqualitee:
awesomeee! 🔥  tytyty

PR #3967, opened 2018-09-04
Adding rename and prefixing to Copy-DbaDatabase

### LINKED ISSUE #3925: Copy-DbaDatabase - Add "-OutputDatabaseName" variable
---
name: Copy-DbaDatabase - name your destination database
---

### Summary of new feature
To be able to make a copy of the database on the same instance, it would be useful to be able to rename the destination database. 

Add "-OutputDatabase" or "-DestinationDatabase" variable - if used, this overwrites default of same name.

0.9.392

Stuart-Moore:
## Type of Change

 - [ ] Bug fix (non-breaking change, fixes #3925)
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Someone asked about allowing a different name for the output of Copy-DbaDatabase, so I added the option.

### Approach
2 new parameters NewName and Prefix, they are mutually exclusive

NewName can only be used when copying a single db, and copies the db under the newname and replaces the old db name in the Physical filename

prefix works with as many dbs as you like. Prefixes the copied database name and the physical file name with the supplied string

Both work with -BackupRestore and -DetachAttach

### Commands to test
`Copy-DbaDatabase -Source localhost\sql2016 -Destination localhost\sql2017 -Database copy1,copy2 -DetachAttach -Reattach -prefix outputtest6`

will copy database copy1 as outputtest6copy1 and prefix it's datafiles with outputtest16, and do the same for database copy2. Using the Detach, copy and attach path

`Copy-DbaDatabase -Source localhost\sql2016 -Destination localhost\sql2017 -Database copy1 -BackupRestore -NetworkShare \\store\sql -NewName copieddb`

will copy database copy1 as copieddb, and replace copy1 with copieddb in the physical filename (if it's in there) using the BackupRestore path

BTW, I'm now aware that this Issue/Request has been knocked back before so I won't be hugely offended if this isn't wanted/needed. But as I've written it I might as well submit it for review and discussion :)

As far as integration with Start-DbaMigration, this shouldn't make any difference. To get the new features the parameters have to be called explicitly which they aren't in SDM so won't impact on that.

### DIFF
--- a/functions/Copy-DbaDatabase.ps1
+++ b/functions/Copy-DbaDatabase.ps1
@@ -302,13 +302,13 @@ function Copy-DbaDatabase {
 
                         if ($null -eq $d.physical) {
                             $directory = Get-SqlDefaultPaths $destServer data
-                            $fileName = Split-Path $file.filename -Leaf
+                            $fileName = (Split-Path $file.filename -Leaf).replace($dbname,$destinationdbname)
                             $d.physical = "$directory\$fileName"
                         }
                     }
                     else {
                         $directory = Get-SqlDefaultPaths $destServer data
-                        $fileName = Split-Path $file.filename -Leaf
+                        $fileName = (Split-Path $file.filename -Leaf).replace($dbname,$destinationdbname)
                         $d.physical = "$directory\$fileName"
                     }
                     $d.logical = $file.Name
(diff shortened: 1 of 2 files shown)


potatoqualitee:
fab! will take a closer look tomorrow, but I like 🌮 thank yoU!

potatoqualitee:
love it, thank you for this. very useful 👍

PR #3970, opened 2018-09-04
Bunch of new commands revolving around exports

potatoqualitee:
Export-DbaInstance needs a smidge more work, then will be ready


wsmelton:
We not going to add the warning on deprecated parameters (e.g. FilePath to just Path)?

potatoqualitee:
naw, just gonna leave the alias in forever. it's useful for gci piping

potatoqualitee:
thank you for looking

potatoqualitee:
also wrt get-dbatrigger, just gonna make an announcement or something because the new proper commands are so drastically different. there's no new replacement.

PR #3990, opened 2018-09-07
rename internal command

potatoqualitee:
i cannot find any references to this command, so i just renamed it


potatoqualitee:
k tested and it works, also all checks pass

Issue #4017, opened 2018-09-12
Set-DbaPrivilege does not set the privilege if no users are currently assigned that privilege


matt-darwin:
## Steps to Reproduce

On a server where there is a right with no users assigned (Lock Pages In Memory in my example):-
![image](https://user-images.githubusercontent.com/29404748/45436636-2c147080-b6ab-11e8-9bae-49a7c72dc6b3.png)

`Set-DbaPrivilege -Computer [Yourcomputer] -type LPIM`

## Expected Behavior

Expect the right to be assigned.

## Actual Behavior

The function appears to work, but no rights are assigned.

This is because the secedit export only contains entries for security policies that have assignments.  As such when the code from the function below runs, it finds nothing to replace:-

```
(Get-Content $tempfile) -replace "SeLockMemoryPrivilege = ", "SeLockMemoryPrivilege = *$SID," |
Set-Content $tempfile
```

In this case, for the first account being added, you might want to do something like this instead:-

```
$NewContent = @"
[Privilege Rights]
SeLockMemoryPrivilege = *$SID
"@
(Get-Content $tempfile) -replace "[Privilege Rights]", $NewContent | Set-Content $tempfile
```

niphlod:
closed via https://github.com/sqlcollaborative/dbatools/pull/4019

PR #4019, opened 2018-09-12
Set dba privilege fix unassigned privilege

matt-darwin:
## Type of Change

 - [ X] Bug fix (non-breaking change, fixes #<enter issue number>)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Fix https://github.com/sqlcollaborative/dbatools/issues/4017

### Approach
Adds the right assignment if it does not exist for the first Account being passed in.

### Commands to test
Delete all users from a privilege.

Set-DbaPrivilege -ComputerName [yourcomputer] -type LPIM


functions/Set-DbaPrivilege.ps1
```diff
@@ -92,45 +92,80 @@ function Convert-UserNameToSID ([string] `$Acc ) {
...
+                                        }
+                                    }    
                                     else {
-                                        Write-Warning "$acc already has Instant File Initialization Privilege on $env:ComputerName"
+                                    
+                                        #here-strings require the tabbing to get all messed up.
+                                        $NewContent = @"
+[Privilege Rights]
+SeManageVolumePrivilege = *$SID
+"@
```
wsmelton: We can't stand here strings....you can just put this as a regular string so it formats properly

wsmelton:
The only section that needs that update will be for LPIM, the others do not require adding content.

functions/Set-DbaPrivilege.ps1
```diff
@@ -92,45 +92,80 @@ function Convert-UserNameToSID ([string] `$Acc ) {
...
-                                    }
+                                        #here-strings require the tabbing to get all messed up.
+                                        $NewContent = @"
+[Privilege Rights]
+SeLockMemoryPrivilege = *$SID
+"@
+                                        (Get-Content $tempfile) -replace "[Privilege Rights]", $NewContent | Set-Content $tempfile
+                                        Write-Verbose "Added new line for Lock Pages In Memory Privileges and granted $acc the privilege on $env:ComputerName" 
+                                    }       
                                 }
```
wsmelton: Tested in my lab and this does not update the LPIM.

I played with this last night for a while as well and never could get it to work. The file won't ouput properly with the update.

matt-darwin: Let me check today if I get chance. The original work I did was outside this function and worked using the same method (and disgusting here-string!); I'm thinking I didn't remove the privilege before going back and testing this so it looked like it worked but hadn't done anything.

functions/Set-DbaPrivilege.ps1
matt-darwin: I'm using the here-string for the carriage return between the section title in the file and the entries.  I don't like them either, but seemed the simplest way to to do this.

matt-darwin:
For the volume maintenance and batch job privs, yes these should always have administrator assigned by default so shouldn't be missing from the file. Unless someone (for who knows what reason) unassigns administrator? In the solution I used for my purpose, I'm actually collecting all the privileges together in a variable and looping through them; so one code change for all.  Could do that here, which would also make it easier to add new privileges in the future?

functions/Set-DbaPrivilege.ps1
potatoqualitee: thank you both so much 🍕🍔!

functions/Set-DbaPrivilege.ps1
wsmelton: You can use "`n" for cariage return in a normal string.

So this would serve the same purpose `"[Privilege Rights]`nSeManageVolumePrivilege = *$SID"

potatoqualitee:
hey @wsmelton - im not sure of the answer, would you be able to help out? I'd like to get this PR resolved and merge 🌮

wsmelton:
It does not appear to be possible the way we execute remote commands. I've tried various ways and none will update the privilege of they do not exists already.

wsmelton:
@matt-darwin does this code change actually work against your test domain environment?

potatoqualitee:
hey matt, do you have any screenshots that confirm this works for you? I'd like to work together to ensure we merge your code but we still need a bit more work.

matt-darwin:
Apologies for the delay; been a bit caught up with stuff.  The code I used to update is attached; I've not used the Set-DbaPrivilege command from dbatools directly.  

[Grant-SQLServerPolicyPrivileges.txt](https://github.com/sqlcollaborative/dbatools/files/2442754/Grant-SQLServerPolicyPrivileges.txt)

Before and after running this (sorry for the obfuscations but you can see there's an account there) :- 
before
after

The servers I was configuring this on didn't have ps remoting enabled for _(not very valid) reasons_ so that definitely might be an issue and the above is done locally.

PR #4022, opened 2018-09-12
Get-DbaProductKey, fixes #4020

### LINKED ISSUE #4020: Get-DbaSqlProductKey does not return a product key for SQL 2016 instances

## Steps to Reproduce

```powershell

PS HKLM:\software\microsoft\Microsoft SQL Server\130\tools> Get-DbaSqlProductKey -SqlInstances . | format-list


SQL Instance : .
SQL Version  : SQL Server 2016 SP2
SQL Edition  : Enterprise Edition: Core-based Licensing (64-bit)
Product Key  :
```

## Expected Behavior

The product key will be populated.

## Actual Behavior

The product key is absent.

## Environmental data

 - PowerShell:
```
PSVersion                      5.1.14393.2312
BuildVersion                   10.0.14393.2312
```
 - - SQL Server:
```
Microsoft SQL Server 2016 (SP2) (KB4052908) - 13.0.5026.0 (X64)
        Mar 18 2018 09:11:49
        Copyright (c) Microsoft Corporation
        Enterprise Edition: Core-based Licensing (64-bit) on Windows Server 2016 Standard 10.0 <X64> (Build 14393: )
```

niphlod:
Do not merge yet, I still need to test on a broader range of setups (but feel free to try).

First and foremost, this function IMHO needs to either target a live instance, or a computer. If for each passed instance every other instance details are fetched, then it should target ComputerName, not SqlInstance.

Fixed things ATM:
 - the original failed to reset product key when multiple instances were found (didn't get properly reset within the loop)
 - it seems that "recent" sql installs (e.g. on WS2016) the regpath is ClientSetup rather than Setup. So for every installable version on "WS2016" now it tries both Setup and ClientSetup
 - it fetched for -gt 2012 the productid of the ClientTools package rather than the instance itself, so I switched the base search to the instance regroot
 - given we have Get-DbaRegistryRoot already, to avoid code duplication, now we use that to get available regroots
 - Get-DbaregistryRoot returned bad instances names due to wrong concatenation, so I fixed that as well

### DIFF
--- a/functions/Get-DbaProductKey.ps1
+++ b/functions/Get-DbaProductKey.ps1
@@ -16,7 +16,7 @@ Login to the target instance using alternative credentials. Windows and SQL Auth
 
 .PARAMETER SqlCms
 Deprecated, pipe in from Get-DbaRegisteredServer
-    
+
 .PARAMETER ServersFromFile
 Deprecated, pipe in from Get-Content
 
@@ -51,7 +51,7 @@ Gets SQL Server versions, editions and product keys for all instances within eac
         [string]$ServersFromFile,
         [switch]$EnableException
     )
-    
+
     begin {
         Function Unlock-SqlInstanceKey {
             [CmdletBinding()]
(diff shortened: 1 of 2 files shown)


niphlod:
okay, I tried on OSes (2008R2 to 2016) and SQLs (2012 to 2017), multiple instances, clusters, it all seems to work.
Someone wants to give it a try on a 2005 ? I think it's the only combination not covered.

PR #4025, opened 2018-09-13
Internal override for Join-Path

potatoqualitee:
does not require the path to exist


potatoqualitee:
tested and works with no complaints when a drive does not exist

potatoqualitee:
fysa @Stuart-Moore

PR #4049, opened 2018-09-20
fixing path check/create for script only

### LINKED ISSUE #4046: Test-DbaBackupInformation tries to create directories for read-only calls
I recall looking at the logic before and couldn't see why but this should not happen when exporting

image

Stuart-Moore:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #4046 )
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
OutputScriptOnly wasn't shorting out the folder tests.

### Approach
Now it does skip them. An info message about the missing path is raised in the verbose stream


potatoqualitee:
thanks so very much!

PR #4051, opened 2018-09-20
Test-DbLastBackup Max size now working as expected

### LINKED ISSUE #3968: test-dbalastbackup -copyfile -maxmb 2 issues
Test-DbaLastBackup -SqlServer $s.Instance -Destination $TestServer -copyfile -MaxMB 500 

1) the files aren't cleaned up for the databases that are skipped due to size.  I think that section of code could be moved down 6 lines
2) if a database is skipped due to size, all the rest of the databases are skipped.  I think it's because the $restoreresult variable isn't initialized in the loop.

Stuart-Moore:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #3968)
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Fix the user requirements

### Approach
reset the results parameter properly on each loop
Move the cleanup outside of the success section so it works when things fail

The trailing space fails are because someone committed them into dev.

### DIFF
--- a/functions/Test-DbaLastBackup.ps1
+++ b/functions/Test-DbaLastBackup.ps1
@@ -447,19 +447,19 @@ function Test-DbaLastBackup {
                             }
 
                             #Cleanup BackupFiles if -CopyFile and backup was moved to destination
-                            if ($CopyFile) {
-                                Write-Message -Level Verbose -Message "Removing copied backup file from $destination."
-                                try {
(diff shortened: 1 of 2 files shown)


potatoqualitee:
awesome, thank you! looks like the test failure is my fault. I'll merge and clean! thanks again 💯 🍵

PR #4059, opened 2018-09-22
Fixing up backup rebase functionality

Stuart-Moore:
## Type of Change

 - [x] Pester test is included (well, it works now).
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Reported in slack that the rebase functionality on Format-DbaBackupInformation wasn't working

### Approach
Made it work.

With the pass by copy a foreach loop wasn't working so switched to a for loop which allowed me to poke directly back into the array (can have multiple paths so needs to loop). 

Also fixed the pester test that was testing the opposite of what should have (ie it would fail if the backups were rebased)

### Learning
Check the test before assuming you can use it to prove things are working.


potatoqualitee:
thanks so much - test failures due to appveyor insanity

PR #4061, opened 2018-09-23
DbaDeprecatedFeature

potatoqualitee:
Added two commands to Get and Test for deprecated features. @ClaudioESSilva - if you'd like to add any claudio magic, please do. For anyone else reviewing this PR:

I did Get-DbaDeprecatedFeature which is instance level but then alsoTest-DbaDeprecatedFeature which could potentially be Test-DbaDbDeprecatedFeature but that didn't feel right because of the name of the Get.


potatoqualitee:
i'm way behind on my tests 👎

PR #4091, opened 2018-09-28
enhanced the error message for Restore-DbaDatabase

potatoqualitee:
would it be ideal to have a try/catch with mroe details instead?


potatoqualitee:
failure due to other dumb thing

PR #4093, opened 2018-09-28
RegisteredServer - CmsRegServer

potatoqualitee:
this makes sense and is a good distinction


potatoqualitee:
an unrelated thing failed

PR #4108, opened 2018-10-02
Restore-DbaDbCertificate - add parameter, fix some formatting

jpomfret:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Added the -EncryptionPassword parameter, previously certificates were being restored to the server and then encrypted with the Decryption password. Now If you specify an encryption password it will use that, however if you don't it will be encrypted using the master key which I believe would be preferable behaviour (at least it is in my shop).

### Approach

Adds `-EncryptionPassword` parameter

### Commands to test

```
#encrypts using master key
Restore-DbaDbCertificate -SqlInstance $server -Path $path -Password $password

#encrypts using password
Restore-DbaDbCertificate -SqlInstance $server -Path $path -Password $password -EncryptionPassword $password
```

### DIFF
--- a/functions/Restore-DbaDbCertificate.ps1
+++ b/functions/Restore-DbaDbCertificate.ps1
@@ -18,6 +18,9 @@ function Restore-DbaDbCertificate {
         .PARAMETER Password
             Secure string used to decrypt the private key.
 
+        .PARAMETER EncryptionPassword
+            If specified this will be used to encrypt the private key.
+
         .PARAMETER Database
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Excellent,t hank you!

jpomfret:
@potatoqualitee just a heads up, I don't think the tests are working so I'm going to look at those today.

PR #4117, opened 2018-10-03
Set-DbaLogin Test Coverage + A Couple Bugs

awickham10:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Added additional test coverage and fixed a few bugs I came across while writing the tests.

### Approach
Fixed bug when renaming login. Username was not validated against a list of existing logins before renaming.

Fixed bug where PSCredential was not accepted for the Password parameter. 

Performance improvement for getting server roles - passed 'master' database to Get-DbaRoleMember to prevent scanning all user databases.

### Commands to test
Pester tests are included.

### Learning
When trying to test against "Couldn't rename login" or "Couldn't disable login" or "Couldn't enable login" I learned that logins can be renamed/enabled/disabled while the login is connected. Pretty dangerous! I'm glad I haven't found that out the hard way.


functions/Set-DbaLogin.ps1
```diff
@@ -179,8 +179,10 @@ function Set-DbaLogin {
                 "PSCredential" { $newPassword = $Password.Password }
                 "SecureString" { $newPassword = $Password }
             }
-        }
-        else {
+
+            if ($Password.GetType().Name -notin @('PSCredential', 'SecureString')) {
+                Stop-Function -Message "Password must be a PSCredential or SecureString" -Target $Login
```
wsmelton: I would just throw the Stop-Function call into the switch statement, it would work either way. Just seems backward for us to go through and validate the type name twice.

Something like this:
```
        if (Test-Bound -ParameterName Password) {
            switch ($Password.GetType().Name) {
                "PSCredential" { $newPassword = $Password.Password }
                "SecureString" { $newPassword = $Password }
                default {
                    Stop-Function -Message "Password must be a PSCredential or SecureString object" -Target $Login
                }
            }
        }
```

wsmelton: Just a bit easier to read to me.

potatoqualitee:
thanks andrew! im eager to address the bugs. will merge. please update with shawn's code when you get a moment.

functions/Set-DbaLogin.ps1
awickham10: Makes sense, will do. Thanks!

PR #4132, opened 2018-10-05
Mirroring command set

potatoqualitee:
I still have some tests to finish & fix but wanted to submit for review.


potatoqualitee:
Alright, I'm done for now. I can't create mirroring on appveyor, so not every command has a test. If I end up adding certs for auth, that may help but as far as this batch goes, please someone give a look over.

I named the big ol automated command `Invoke-DbaDbMirroring` because it does way more than just a `New-DbaDbMirror`. Something with the name doesn't sit right tho.

Here's what it does

* Verifies that a mirror is possible
 * Sets the recovery model to Full if needed
* If the database does not exist on mirror or witness, a backup/restore is performed
* Sets up endpoints if necessary
* Creates a login and grants permissions to service accounts if needed
* Starts endpoints if needed
* Sets up partner for mirror
* Sets up partner for primary
* Sets up witness if one is specified

PR #4139, opened 2018-10-06
Test fixes

potatoqualitee:
from newly changed command names


functions/Get-DbaDbRoleMember.ps1
```diff
@@ -144,14 +144,14 @@ function Get-DbaDbRoleMember {
...
                         if ($user) {
-                            Add-Member -Force -InputObject $user -MemberType 'NoteProperty' -Name 'ComputerName' -Value $server.ComputerName
-                            Add-Member -Force -InputObject $user -MemberType 'NoteProperty' -Name 'InstanceName' -Value $server.ServiceName
-                            Add-Member -Force -InputObject $user -MemberType 'NoteProperty' -Name 'SqlInstance' -Value $server.DomainInstanceName
-                            Add-Member -Force -InputObject $user -MemberType 'NoteProperty' -Name 'Database' -Value $db.Name
-                            Add-Member -Force -InputObject $user -MemberType 'NoteProperty' -Name 'Role' -Value $dbRole.Name
-                            Add-Member -Force -InputObject $user -MemberType 'NoteProperty' -Name 'UserName' -Value $user.Name
-
-                            Select-DefaultView -InputObject $user -Property 'ComputerName', 'InstanceName', 'SqlInstance', 'Database', 'Role', 'UserName', 'Login'
+                            $user | Add-Member -Force -MemberType 'NoteProperty' -Name 'ComputerName' -Value $server.ComputerName
```
wsmelton: We don't use the piping option in any of the commands because it is slower. Any reason you chose to do it on this one?

functions/Get-DbaServerRoleMember.ps1
```diff
@@ -145,12 +145,12 @@ function Get-DbaServerRoleMember {
                     $l = $server.Logins | Where-Object { $_.Name -eq $member }
 
                     if ($l) {
-                        Add-Member -Force -InputObject $l -MemberType 'NoteProperty' -Name 'ComputerName' -Value $server.ComputerName
-                        Add-Member -Force -InputObject $l -MemberType 'NoteProperty' -Name 'InstanceName' -Value $server.ServiceName
-                        Add-Member -Force -InputObject $l -MemberType 'NoteProperty' -Name 'SqlInstance' -Value $server.DomainInstanceName
-                        Add-Member -Force -InputObject $l -MemberType 'NoteProperty' -Name 'Role' -Value $role
-
-                        Select-DefaultView -InputObject $l -Property 'ComputerName', 'InstanceName', 'SqlInstance', 'Role', 'Name'
+                        $l | Add-Member -Force -MemberType 'NoteProperty' -Name 'ComputerName' -Value $server.ComputerName
```
wsmelton: Same question as previous comment, the Set command does not use piping why would the Get need it?

functions/Test-DbaDbCompression.ps1
```diff
@@ -570,84 +581,69 @@ IF OBJECT_ID('tempdb..##tmpEstimatePage', 'U') IS NOT NULL
...
+                            PercentUpdate = $row.PercentUpdate
+                            RowEstimatePercentOriginal = $row.RowEstimatePercentOriginal
+                            PageEstimatePercentOriginal = $row.PageEstimatePercentOriginal
                             CompressionTypeRecommendation = $row.CompressionTypeRecommendation
-                            SizeCurrent                   = [dbasize]($row.SizeCurrentKB * 1024)
-                            SizeRequested                 = [dbasize]($row.SizeRequestedKB * 1024)
-                            PercentCompression            = $row.PercentCompression
+                            SizeCurrent  = [dbasize]($row.SizeCurrentKB * 1024)
+                            SizeRequested = [dbasize]($row.SizeRequestedKB * 1024)
+                            PercentCompression = $row.PercentCompression
```
wsmelton: Just an FYI, the first time someone opens this up in VS Code and commits a change the formatting will go back the way it was 😀 🔢 ☣️ 

A ton of PRs I have seen commands have one minor change from the contributor and the remainder are formatting changes caused by the editor of choice. I don't really know how we can fix that unless we enforce rules and catch things like this.

functions/Get-DbaDbRoleMember.ps1
potatoqualitee: yes, when there is a missing foreach () loop. This has multiple $user so it needs to foreach in the pipe. I think. I'm seeing REALLY weird stuff.

Look at this

image

wsmelton: Your `OutVariable roles` needs to be prior to the pipe and filter.

potatoqualitee:
finally got some passes 😅

PR #4141, opened 2018-10-06
Get-DbaSpDatabase

potatoqualitee:
This is something I've needed a bunch of times, also will be useful for AutoSPAg

### DIFF
--- a/dbatools.psd1
+++ b/dbatools.psd1
@@ -11,7 +11,7 @@
     RootModule             = 'dbatools.psm1'
 
     # Version number of this module.
-    ModuleVersion          = '0.9.458'
+    ModuleVersion          = '0.9.459'
 
     # ID used to uniquely identify this module
     GUID                   = '9d139310-ce45-41ce-8e8b-d76335aa1789'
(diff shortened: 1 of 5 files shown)


potatoqualitee:
I cannot get a compressed backup of this database to restore, so I used a dacpac instead. Not sure what's up - I even backed up / restored different/old SharePoint_Config dbs. This works but takes 15 seconds. I've asked stuart to take a look.

wsmelton:
Are going to start adding in commands for different products that may be hosted on SQL Server?

potatoqualitee:
we can potentially with justification, but sharepoint is the reason dbatools exists, i'm a huge user of it and when mixed with ags, this will be another magic feature people lose their minds over.

potatoqualitee:
all tests finally pass 👍

potatoqualitee:
my only issue is that..is this the best name? Would Get-DbaDbSharePoint be better?

wsmelton:
Product name should be full, because "sp" is short for stored procedure with SQL Server world.

potatoqualitee:
i dig, ty, will rename

PR #4150, opened 2018-10-08
Centralize verbose messages

potatoqualitee:
This centralizes the Verbose messages to Connect-SqlInstance and Invoke-ManagedComputerCommand. It will only report that it's connecting when an explicit connection is being made.


potatoqualitee:
one pass

PR #4158, opened 2018-10-09
Install-SqlServer

reitse:
Hi all,

As mentioned weeks ago on twitter i'm working on a SQL Installer script. This is what i've created so far.

I'm sure this will come back to me for improvement, but i'm really curious what you think. 
It's my first pull request so it's exiting ;)


potatoqualitee:
hey @reitse, thanks so much! I'll take a look and provide a review momentarily 👍💯

SqlInstall.ps1
```diff
@@ -0,0 +1,556 @@
+﻿function Install-SqlServer ([string]$AppVolume, [string]$DataVolume, [string]$LogVolume, [string]$TempVolume, [string]$BackupVolume, [int]$Ctp)
```
potatoqualitee: All of our commands start with Dba, so this would be `Install-DbaInstance`. The params are also laid out in a specific way. Take a look at `Get-DbaDatabase` to see what I'm talking about, down near [cmdletbinding()]

SqlInstall.ps1
```diff
@@ -0,0 +1,556 @@
...
+    The perform volume maintenance right will be granted to the SQL Server account. If you run this script in an environment where you are not allowed to do this,
+    please revert that operation by removing the right from the local security policy (secpol.msc).
+
+    The autogrowth of file will be set to 64MB, the Cost Threshold for Parallelism will be set to 40. 
+
+    Note that the dowloaded installation file must be unzipped, an ISO has to be mounted. This will not be executed from this script.
+    
+    .PARAMETER Name
+
+    $AppVolume will hold the volume letter of the application disc. If left empty, it will default to C, unless there is a drive named like App
```
potatoqualitee: Doc content looks good! Just need to format it properly, like `.PARAMETER AppVolume`

SqlInstall.ps1
```diff
@@ -0,0 +1,556 @@
...
+    This will run the installation with the default settings
+
+    C:\PS> Install-SqlServer -AppVolume "G"
+
+    This will run the installation with default setting apart from the application volume, this will be redirected to the G drive.
+
+
+    #>
+
+    $startScript = @'
```
potatoqualitee: We always try to accommodate as many scenarios as possible. Looks like this one is just for SQL Server 2014. I'd recommend finding the config files for various sqls (SPADE has some), placing them in the bin\installtemplate directory, then doing find/replace on them as needed (`$configini = Get-Content "$script:PSModuleRoot\bin\installtemplate\$version.ini"` then `$configini = $configini.Replace("--placeholder-role--","DatabaseEngine_only")`

SqlInstall.ps1
```diff
@@ -0,0 +1,556 @@
...
+
+;SQL Server 2014 Configuration File
+[OPTIONS]
+; Specifies a Setup work flow, like INSTALL, UNINSTALL, or UPGRADE. This is a required parameter. 
+ACTION="Install"
+; Detailed help for command line argument ROLE has not been defined yet. 
+ROLE="AllFeatures_WithDefaults"
```
potatoqualitee: AllFeatures_WithDefaults is a heavy default. I'd place $Role as a parameter in the script and set it to whatever database engine only is.  So someone can do `Install-DbaInstance -Role AllFeaturesWithDefaults` if they'd like.

SqlInstall.ps1
```diff
@@ -0,0 +1,556 @@
...
+    $NumberOfCores = Get-WmiObject -Class Win32_processor |  
+        Measure-Object NumberOfLogicalProcessors -Sum | 
+        Select-Object -ExpandProperty sum
+
+    IF ($NumberOfCores -gt 8)
+    { $NumberOfCores = 8 }
+
+    #Get the amount of available memory. If it's more than 40 GB, give the server 10% of the memory, else reserve 4 GB.
+
+    $ServerMemory = Get-WmiObject -Class win32_physicalmemory |
```
potatoqualitee: Get-WmiObject is slower than the new Get-CimInstance, but in our toolset, we use `Get-DbaCmObject` which has the same syntax. Nevertheless, we do have a command that calculates this that you'd want to take advantage of after the install is perofrmed - `Test-DbaMaxMemory`

SqlInstall.ps1
```diff
@@ -0,0 +1,556 @@
...
+        $AppVolume = 'C'
+    }
+
+    IF ( $null -eq $BackupVolume -or $BackupVolume -eq '') {
+        $BackupVolume = $DataVolume
+    }
+
+
+
+    Clear-Host
```
potatoqualitee: dbatools does not use any -Host commands. Generally, we try to output objects to the user.

SqlInstall.ps1
```diff
@@ -0,0 +1,556 @@
...
+
+    Write-Host 'Your datadrive:' $DataVolume
```
potatoqualitee: This would be better as a verbose comment. Generally, that's `Write-Verbose` but in our toolset, we use `Write-Message -Level Verbose -Message "Your datadrive:' $DataVolume"`

SqlInstall.ps1
```diff
@@ -0,0 +1,556 @@
...
+    Write-Host 'Your logdrive:' $LogVolume
+    Write-Host 'Your TempDB drive:' $TempVolume
+    Write-Host 'Your applicationdrive:' $AppVolume
+    Write-Host 'Your Backup Drive:' $BackupVolume
+    Write-Host 'Number of cores for your Database:' $NumberOfCores
+
+    Write-Host  'Do you agree on the drives?'
```
potatoqualitee: We try not to use non-native interactivity. To prompt a user, we'd use SupportsShouldProcess. You can see this in `New-DbaEndpoint` and `Remove-DbaDatabase`, among others.

SqlInstall.ps1
```diff
@@ -0,0 +1,556 @@
...
+
+    $CheckLastTwoChar = ":\"
+    $CheckLastChar = "\"
+
+    Switch ($AlterDir) {
+        Y {Write-Host "Yes, drives agreed, continuing"; }
+        N {
+            Write-Host "Datadrive: " $DataVolume
+            $NewDataVolume = Read-Host "Your datavolume: "
+            If ($NewDataVolume.Substring($NewDataVolume.Length - 2 -eq $CheckLastTwoChar) -and $NewDataVolume.Length -gt 2) {
```
potatoqualitee: Replaces are a lot easier than SubStrings, I'd use that instead.

SqlInstall.ps1
```diff
@@ -0,0 +1,556 @@
...
+                Write-Host "BackupVolume remains on " $BackupVolume
+            }
+        }
+        Default {Write-Host "Drives agreed, continuing"; }
+    }
+
+    $CheckLastTwoChar = ":\"
+    $CheckLastChar = "\"
+
+    $SetupFile = Read-Host -Prompt 'Please enter the root location for Setup.exe'
```
potatoqualitee: This would be a required command parameter as opposed to a prompt (tho I'd likely try to determine the path for them. Detect removable drives, and search for the path to sql server setup)

SqlInstall.ps1
```diff
@@ -0,0 +1,556 @@
...
+
+    if ( -Not (Test-Path -Path $ConfigFile ) ) {
+        New-Item -ItemType directory -Path $ConfigFile
+    }
+
+    Out-File -FilePath C:\Temp\ConfigurationFile2.ini -InputObject $startScript
+
+    $FileLocation2 = $ConfigFile + 'ConfigurationFile2.ini'
+
+    (Get-Content -Path $FileLocation2).Replace('SQLBACKUPDIR="E:\Program Files\Microsoft SQL Server\MSSQL12.AXIANSDB01\MSSQL\Backup"', 'SQLBACKUPDIR="' + $BackupVolume + ':\Program Files\Microsoft SQL Server\MSSQL12.AXIANSDB01\MSSQL\Backup"') | Out-File $FileLocation2
```
potatoqualitee: Yes! nice!

SqlInstall.ps1
```diff
@@ -0,0 +1,556 @@
...
+    }
+
+
+    $ConfigFile = 'c:\temp\'
+
+    if ( -Not (Test-Path -Path $ConfigFile ) ) {
+        New-Item -ItemType directory -Path $ConfigFile
+    }
+
+    Out-File -FilePath C:\Temp\ConfigurationFile2.ini -InputObject $startScript
```
potatoqualitee: C:\temp may not exist. Since `$env:temp` is formatted in DOS syntax, use `[System.IO.Path]::GetTempPath()`

SqlInstall.ps1
```diff
@@ -0,0 +1,556 @@
...
+    (Get-Content -Path $FileLocation2).Replace('SQLTEMPDBDIR="E:\Program Files\Microsoft SQL Server\MSSQL12.AXIANSDB01\MSSQL\Data"', 'SQLTEMPDBDIR="' + $TempVolume + ':\Program Files\Microsoft SQL Server\MSSQL12.AXIANSDB01\MSSQL\Data"') | Out-File $FileLocation2
+
+    (Get-Content -Path $FileLocation2).Replace('SQLUSERDBLOGDIR="E:\Program Files\Microsoft SQL Server\MSSQL12.AXIANSDB01\MSSQL\Log"', 'SQLUSERDBLOGDIR="' + $LogVolume + ':\Program Files\Microsoft SQL Server\MSSQL12.AXIANSDB01\MSSQL\Log"') | Out-File $FileLocation2
+
+    (Get-Content -Path $FileLocation2).Replace('SQLSYSADMINACCOUNTS="WIN-NAJQHOBU8QD\Administrator"', 'SQLSYSADMINACCOUNTS="' + $env:COMPUTERNAME + '\Administrator"')| Out-File $FileLocation2
+
+    #$SetupFile = 'C:\Users\Administrator\Downloads\SQLEXPR_x64_ENU\Setup.exe'
+    #$ConfigFile = 'C:\temp\ConfigurationFile2.ini'
+
+    $SAPassW = '[InsertPasswordHere]'
```
potatoqualitee: We should try to avoid plain text when possible. Use a [PsCredential] (Get-Credential) and then `$credential.GetNetworkCredential().password`

SqlInstall.ps1
```diff
@@ -0,0 +1,556 @@
...
+
+    & $SetupFile /ConfigurationFile=$FileLocation2 /Q /IACCEPTSQLSERVERLICENSETERMS /SAPWD=$SAPassW
+
+    # Grant service account the right to perform volume maintenance
+    # code found at https://social.technet.microsoft.com/Forums/windows/en-US/5f293595-772e-4d0c-88af-f54e55814223/adding-domain-account-to-the-local-policy-user-rights-assignment-perform-volume-maintenance?forum=winserverpowershell
+
+    ## <--- Configure here
+    $accountToAdd = 'NT Service\MSSQL$AXIANSDB01'
```
potatoqualitee: This should also be a parameter

SqlInstall.ps1
```diff
@@ -0,0 +1,556 @@
...
+        $ntprincipal = new-object System.Security.Principal.NTAccount "$accountToAdd"
+        $sid = $ntprincipal.Translate([System.Security.Principal.SecurityIdentifier])
+        $sidstr = $sid.Value.ToString()
+    }
+    catch {
+        $sidstr = $null
+    }
+    Write-Host "Account: $($accountToAdd)" -ForegroundColor DarkCyan
+    if ( [string]::IsNullOrEmpty($sidstr) ) {
+        Write-Host "Account not found!" -ForegroundColor Red
```
potatoqualitee: I'd do validation earlier, and exit out of the command. To exit, we use our internal function 

```powershell
Stop-Function -Message "$accountToAdd could not be found"
return
```

SqlInstall.ps1
```diff
@@ -0,0 +1,556 @@
...
+        #exit -1
+    }
+
+    Write-Host "Account SID: $($sidstr)" -ForegroundColor DarkCyan
+    $tmp = ""
+    $tmp = [System.IO.Path]::GetTempFileName()
```
potatoqualitee: yeahhh! this is a better temp

SqlInstall.ps1
```diff
@@ -0,0 +1,556 @@
...
+    Write-Host "Export current Local Security Policy" -ForegroundColor DarkCyan
+    secedit.exe /export /cfg "$($tmp)" 
+    $c = ""
+    $c = Get-Content -Path $tmp
+    $currentSetting = ""
+    foreach ($s in $c) {
+        if ( $s -like "SeManageVolumePrivilege*") {
+            $x = $s.split("=", [System.StringSplitOptions]::RemoveEmptyEntries)
```
potatoqualitee: We have dbatools commands that can take care of this. Check out `Set-DbaPrivilege`

SqlInstall.ps1
```diff
@@ -0,0 +1,556 @@
...
+        Write-Host "NO ACTIONS REQUIRED! Account already in ""Perform Volume Maintenance Task""" -ForegroundColor DarkCyan
+    }
+    Write-Host "Done." -ForegroundColor DarkCyan 
+
+
+
+    # Now for the fun part, alter the database settings
+    # First we need to install (if necessary) the SQL commandlets
+
+    Install-PackageProvider -Name NuGet -Force
```
potatoqualitee: We rarely use external modules. In this case,  calling out to Nuget and getting SqlServer would not be allowed.

SqlInstall.ps1
```diff
@@ -0,0 +1,556 @@
...
+    #most queries work best when you're around databases. 
+    Set-Location .\SQL\$env:COMPUTERNAME\AXIANSDB01\databases
+
+    #Let's setup the Max Dop, CTfP and max server memory.
+    if($null -eq $Ctp -or $Ctp -eq '')
+    {
+        $Ctp = 40
+    }
+
+    $sql1 = @'
```
potatoqualitee: This would be out of scope for the Installer.

potatoqualitee:
Also, for clarity, what we'd like is to focus on is the Installer, not post-config. If it's in the config ini/options, it's in scope. If it's not, that can be highlighted in a blog post on dbatools.io 👍 Also, I've attached a zip of the ini files for 2005-2012. I can test those. We still need 2014 which you've got, then 2016 and 2017 which I can find if need be.

[config_ini.zip](https://github.com/sqlcollaborative/dbatools/files/2459969/config_ini.zip)

SqlInstall.ps1
reitse: I've changed the name of the function. 
Parameters are now in a seperate Param block. 
Further development will show if i've made the right choices :)

SqlInstall.ps1
reitse: Formatting changed. I'll keep the content up to date with all the additions and changes.

SqlInstall.ps1
reitse: The script has been tested with SQL2014 and 2017, but that's just one test. 
I agree that using the ini files from each version is a better way to go. 
Do you want support voor 2008/2008R2? Or do we start from 2012 up to 2017 or even 2019?
This because i want to be able to validate the user input and prevent an error on input like '2015'

SqlInstall.ps1
reitse: Missed that one... From an Express perspective it won't be too bad, but from an Enterprise edition some weird stuff could be happening.

SqlInstall.ps1
reitse: command removed. This was a left over from the basics when it was just a ps script.

SqlInstall.ps1
reitse: All 49 occurences changed to Write-Message. 

The main difference between these two is the pipeline input possibility. Correct or am i missing something?

potatoqualitee: It's part of our "messaging system" that allows us to write to file and stuff. Check this out:

`Get-DbatoolsConfigValue -Name path.dbatoolslogpath | Invoke-Item` 

That should open your logging folder and you can take a peek inside.

SqlInstall.ps1
potatoqualitee: if 2008 causes too much strife, I'd be happy with 2012+. We can use `[ValidateSet("2012", "2014", "2016","2017","2019")]` on that parameter and it'll allow auto-complete 💯

SqlInstall.ps1
potatoqualitee: Excellent

SqlInstall.ps1
potatoqualitee: awesome :D don't forget, if you commit to your same branch, it should show here and we can work through it as it happens.

reitse: I've committed to this branch and so you should some progress.

Nowhere near a new pullrequest though ;)

SqlInstall.ps1
reitse: ValidateSet is cool!

SqlInstall.ps1
reitse: Code removed

SqlInstall.ps1
reitse: Don't you just love scope-creep ;)

Code removed.

potatoqualitee:
going to close this for now but can reopen once the changes have been made. Looking forward to them! 🎉

SqlInstall.ps1
reitse: Looking into that right now. Replacing a ':\' or  '\' is really easy with -replace, but i have to find a way to check only for the last part of the string. If someone enters  E:\DBLOGS\DATA\ as a directory, it should only replace the last '\', not alle the occurences for it would mess up de directory.

potatoqualitee: ah ha! very true

reitse:
Hi Chrissy,

I think there's some progress here :) 
It's nowhere done yet, but some feedback would be appreciated. What do i have to do when you reopen this PR?

SqlInstall.ps1
```diff
@@ -75,6 +77,7 @@
     Param  (
         [ValidateSet("2008", "2008R2", "2012", "2014", "2016", "2017", "2019")][string]$Version, 
         [ValidateSet("Express", "Standard", "Enterprise", "Developer")][string]$Edition,
+        [ValidateSet("AllFeaturesWithDefaults", "Custom")][string]$Role,
```
potatoqualitee: Liking the new additions! Please place validate set on a sep line.

````powershell
[ValidateSet("AllFeaturesWithDefaults", "Custom")]
[string]$Role,
````

File: SqlInstall.ps1 [hunk 21]
```diff
@@ -93,6 +96,18 @@
         Stop-Function -Message "You need to specify a SQL Server Version to run this function." -Continue -EnableException $EnableException
     }
 
+    If($Null -eq $Edition -or $Edition -eq '')
+    {
+        $Edition = "Express"
```
potatoqualitee: You can actually take care of this in the parameters - so do `$Edition = "Express"` and it'll auto fill express if they didn't pass an Edition 👍

potatoqualitee:
oh la la, 6 files! lemme take a look

bin/installtemplates/2008/Configuration_2008.ini
```diff
@@ -0,0 +1,48 @@
...
+FEATURES="$ProductString"
+HELP="False"
+INDICATEPROGRESS="True"
+QUIET="True"
+QUIETSIMPLE="False"
+X86="$X86"
+PCUSOURCE="$SPSource"
+CUSOURCE="$CUSource"
+ERRORREPORTING="False"
+INSTALLSHAREDDIR="C:\Program Files\Microsoft SQL Server"
```
potatoqualitee: Can you make these dynamic as well? So add it as a parameter and say

````powershell
[string]$InstallSharedDirectory = "C:\Program Files\Microsoft SQL Server"
````

Then 
````powershell
INSTALLSHAREDDIR=$InstallSharedDirectory
````

**Comment by potatoqualitee** [comment]
k it finally synced. that was interesting, never did this before. btw, i updated your branch to be current, so you may notice that when you sync. the branch was hudnreds of commits behind

File: SqlInstall.ps1 [hunk 23]
```diff
@@ -0,0 +1,433 @@
+﻿function Install-DbaInstance {
```
potatoqualitee: Please rename the file as well to Install-DbaInstance.ps1 in the functions directory

SqlInstall.ps1
```diff
@@ -0,0 +1,433 @@
...
+    The perform volume maintenance right can be granted to the SQL Server account. If you happen to activate this in an environment where you are not allowed to do this,
+    please revert that operation by removing the right from the local security policy (secpol.msc).
+
+    You will see a screen with the users available on your machine. There you can choose the user that will act as Service Account for your SQL Server Install. This
+    implies that the user has been created beforehand. 
+
+    Note that the dowloaded installation file must be unzipped or an ISO has to be mounted. This will not be executed from this script. This function offers the possibility
+    to execute an autosearch for the installation files. But you can just browse to the correct file if you like.
+
+    .PARAMETER Version will hold the SQL Server version you wish to install. The variable will support autocomplete
```
potatoqualitee: Close! This should be

````
.PARAMETER Version
     Version will hold....
````

File: SqlInstall.ps1 [hunk 25]
```diff
@@ -0,0 +1,433 @@
...
+    .PARAMETER SaveFile will prompt you for a file location to save the new config file. Otherwise it will only be saved in the PowerShell bin directory.
+
+    .PARAMETER Authentication will prompt you if you want mixed mode authentication or just Windows AD authentication. With Mixed Mode, you will be prompted for the SA password.
+
+    .PARAMETER EnableException
+            By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
+            This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
+            Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
+
+    .Inputs
```
potatoqualitee: This can be removed, as well as outputs. Many projects use them, ours generally do not.

SqlInstall.ps1
```diff
@@ -0,0 +1,433 @@
...
+        [ValidateSet("Express", "Standard", "Enterprise", "Developer")][string]$Edition,
+        [ValidateSet("AllFeaturesWithDefaults", "Custom")][string]$Role,
+        [ValidateSet("Python", "R", "Python and R")][string]$StatsAndMl,
+        [string]$AppVolume, 
+        [string]$DataVolume, 
+        [string]$LogVolume, 
+        [string]$TempVolume, 
+        [string]$BackupVolume,
+        [string]$InstallFolder,
+        [ValidateSet("Yes", "No")][string]$PerformVolumeMaintenance,
```
potatoqualitee: Booleans are "switches", which allow you to do `-PerformVolumeMaintenance` for $true, or don't pass the param for false. This would need to be

`[switch]$PerformVolumeMaintenance`

SqlInstall.ps1
```diff
@@ -0,0 +1,433 @@
...
+        [ValidateSet("Yes")][string]$SaveFile,
+        [ValidateSet("Windows", "Mixed Mode")][string]$Authentication,
+        [switch]$EnableException
+    )
+
+
+    if ($Null -eq $Version -or $Version -eq '') {
+        Stop-Function -Message "You need to specify a SQL Server Version to run this function." -Continue -EnableException $EnableException
```
potatoqualitee: You can make a parameter mandatory in params, that'd be the best way

````powershell
[parameter(Mandatory)]
[ValidateSet("2008", "2008R2", "2012", "2014", "2016", "2017", "2019")]
[string]$Version,
````

You may be tempted to do `[parameter(Mandatory = $true)]` after some research but that is PowerShell v2 syntax and we support v3 and up, so we can keep the prettier syntax without `= $true`

File: SqlInstall.ps1 [hunk 28]
```diff
@@ -0,0 +1,433 @@
...
+        }
+    }
+
+    # Check if the edition of SQL Server supports Python and R. Introduced in SQL 2016, it should not be allowed in earlier installations.
+
+    IF ( $null -eq $StatsAndMl -or $StatsAndMl -ne '' ) {
+        $Array = "2016", "2017", "2019"
+
+        IF ($Version -notin $Array ) {
+            $StatsAndMl = '';
```
potatoqualitee: What if they pass StatsAndMl and 2008? This should be a check with a `Stop-Function`

Also, you'll love `-not`, it's super convenient. Takes care of nulls and empties.

`if (-not $StatsAndMl -or -not $StatsAndMl) {`

Also make sure you use lower case for `begin`, `process`, `end`, `if`, `else`, `foreach` and so on.

SqlInstall.ps1
```diff
@@ -0,0 +1,433 @@
...
+    # Let the user set the Service Account for SQL Server. This does imply that the user has been created.
+
+    $SqlServerAccount = Get-CimInstance -ClassName Win32_UserAccount  | Out-GridView -title 'Please select the Service Account for your Sql Server instance.' -PassThru | Select-Object -ExpandProperty Name
+    
+    # Get the installation folder of SQL Server. If the user didn't choose a specific folder, the autosearch will commence. It will take some time!
+    #To limit the number of results, the search exludes the Windows
+
+    IF ($InstallFolder::IsNullOrEmpty()) {
+        
+        Write-Message -level Verbose -Message "No Setup directory found. Switching to autosearch"
```
potatoqualitee: SO COOL!

SqlInstall.ps1
```diff
@@ -0,0 +1,433 @@
...
+    IF ( $null -eq $AppVolume -or $AppVolume -eq '') {
+        $AppVolume = 'C'
+    }
+
+    IF ( $null -eq $BackupVolume -or $BackupVolume -eq '') {
+        $BackupVolume = $DataVolume
+    }
+
+
+    Write-Message -Level Verbose -Message 'Your datadrive:' $DataVolume
```
potatoqualitee: Lookin good 👍

SqlInstall.ps1
```diff
@@ -0,0 +1,433 @@
...
+    Write-Message -Level Verbose -Message 'Your logdrive:' $LogVolume
+    Write-Message -Level Verbose -Message 'Your TempDB drive:' $TempVolume
+    Write-Message -Level Verbose -Message 'Your applicationdrive:' $AppVolume
+    Write-Message -Level Verbose -Message 'Your Backup Drive:' $BackupVolume
+    Write-Message -Level Verbose -Message 'Number of cores for your Database:' $NumberOfCores
+
+    Write-Message -Level Verbose -Message  'Do you agree on the drives?'
+    $AlterDir = Read-Host " ( Y / N )"
```
potatoqualitee: This will be taken care of by something kind of advanced called `SupportsShouldProcess`. Please remove it for now and we'll revisit once you are closer to your goal. It took me a while to understand `SupportsShouldProcess` and I want to make sure I have the time to make it approachable.

SqlInstall.ps1
```diff
@@ -0,0 +1,433 @@
...
+
+    Switch ($AlterDir) {
```
potatoqualitee: I love switches! But let's remove this for now and revisit.

SqlInstall.ps1
```diff
@@ -0,0 +1,433 @@
...
+                Write-Message -Level Verbose -Message "BackupVolume moved to " $BackupVolume
+            }
+        }
+        Default {Write-Message -Level Verbose -Message "Drives agreed, continuing"; }
+    }
+
+
+    # Out-File -FilePath C:\Temp\ConfigurationFile2.ini -InputObject $startScript
+
+    (Get-Content -Path $configini).Replace('SQLBACKUPDIR="E:\Program Files\Microsoft SQL Server\MSSQL12.AXIANSDB01\MSSQL\Backup"', 'SQLBACKUPDIR="' + $BackupVolume + ':\Program Files\Microsoft SQL Server\MSSQL12.AXIANSDB01\MSSQL\Backup"') | Out-File $configini
```
potatoqualitee: It looks like this will overwrite the config file in the module directory. Please use a temporary directory instead.

````
$temp = ([System.IO.Path]::GetTempPath()).TrimEnd("\")
$tempini = "$temp\config.ini"
````

File: SqlInstall.ps1 [hunk 34]
```diff
@@ -0,0 +1,433 @@
...
+
+    (Get-Content -Path $configini).Replace('SQLSYSADMINACCOUNTS="WIN-NAJQHOBU8QD\Administrator"', 'SQLSYSADMINACCOUNTS="' + $SqlServerAccount)| Out-File $configini
+
+    If ($SaveFile -eq "Yes") {
+        $SaveFileLocation = Read-Host "Please enter your preferred directory for saving a copy of the configuration file: "
+        Copy-Item $configini -Destination $SaveFileLocation
+    }
+
+    IF ($Authentication -eq "Mixed Mode") {
+        $SAPassW = [PsCredential](Get-Credential -UserName "SA"  -Message "Please Enter the SA Password.")
```
potatoqualitee: I'd do this check closer to the top and use that `Credential` parameter. Unless that is being used for remote stuff, then do like `[pscredential]$SaCredential` 

so that we can reduce interruption to automation as much as possible

SqlInstall.ps1
```diff
@@ -0,0 +1,433 @@
...
+    }
+
+    
+
+    # Grant service account the right to perform volume maintenance
+    # code found at https://social.technet.microsoft.com/Forums/windows/en-US/5f293595-772e-4d0c-88af-f54e55814223/adding-domain-account-to-the-local-policy-user-rights-assignment-perform-volume-maintenance?forum=winserverpowershell
+
+    if ($PerformVolumeMaintenance) {
+        ## <--- Configure here
+        $accountToAdd = 'NT Service\MSSQL$AXIANSDB01'
```
potatoqualitee: This would be `"NT Service\MSSQL`$$env:ComputerName"` (or whatever syntax works ;))

SqlInstall.ps1
```diff
@@ -0,0 +1,433 @@
...
+        catch {
+            $sidstr = $null
+        }
+        Write-Message -Level Verbose -Message "Account: $($accountToAdd)" -ForegroundColor DarkCyan
+        if ( [string]::IsNullOrEmpty($sidstr) ) {
+            Write-Message -Level Verbose -Message "Account not found!" -ForegroundColor Red
+            #exit -1
+        }
+
+        Write-Message -Level Verbose -Message "Account SID: $($sidstr)" -ForegroundColor DarkCyan
```
potatoqualitee: The colors are controlled by the config system, please remove the colors. At the end, we'll add some pizzaz with a progress bar if you'd like (I'd like!)

It would look something like this

image

SqlInstall.ps1
```diff
@@ -0,0 +1,433 @@
...
+        $tmp = ""
+        $tmp = [System.IO.Path]::GetTempFileName()
+        Write-Message -Level Verbose -Message "Export current Local Security Policy" -ForegroundColor DarkCyan
+        secedit.exe /export /cfg "$($tmp)" 
+        $c = ""
```
potatoqualitee: Please use the other dbatools commands such as `Set-DbaPrivilege`

SqlInstall.ps1
```diff
@@ -0,0 +1,433 @@
...
+        Write-Message -Level Verbose -Message "Done." -ForegroundColor DarkCyan 
+    }
+
+    
+
+    #Now configure the right amount of TempDB files.
+
+    $val = 1
+
+    WHILE ($val -ne $NumberOfCores) {
```
potatoqualitee: lower case too, just for PowerShell, the SQL is good

SqlInstall.ps1
```diff
@@ -0,0 +1,433 @@
...
+        $sqlM = 'ALTER DATABASE tempdb ADD FILE ( NAME = N''tempdev' + $val + ''', FILENAME = N''' + $TempVolume + ':\Program Files\Microsoft SQL Server\MSSQL12.AXIANSDB01\MSSQL\DATA\tempdev' + $val + '.ndf'' , SIZE = 64MB , FILEGROWTH = 64MB)'
```
potatoqualitee: We may have a command for this as well. Please check `Set-DbaTempdbConfig`

SqlInstall.ps1
```diff
@@ -0,0 +1,433 @@
...
+(NAME = tempdev,  
+SIZE = 64MB, FILEGROWTH = 64MB);  
+GO  
+'@
+
+    Invoke-Sqlcmd -Database TempDB -Query $sql
+
+    #Turn off SA, primary break-in point of the naughty users
+
+    $sql = 'ALTER LOGIN sa DISABLE'
```
potatoqualitee: I know it's best practice, but this should be left up to the user. Also note that Invoke-Sqlcmd is a command that comes from SqlServer. Our equivalent is `Invoke-DbaQuery`

SqlInstall.ps1
potatoqualitee: We'll have to discuss if this is in scope, however.

SqlInstall.ps1
potatoqualitee: Now that I think about it, this is likely not in scope. If the SQL Server binaries don't support it, we shouldn't either. If they do, we should rely on the binaries.

SqlInstall.ps1
reitse: Can i change filename and location without messing up this thread?

SqlInstall.ps1
reitse: It shouldn't overwrite it, earlier in the code, the chosen ini file gets copied to a different location (around line 127 this process takes place :))

SqlInstall.ps1
reitse: I think it is in scope. Starting with SQL 2016 you can set a check on perform volume maintenance in the script. So that part has to be in there for 2016 and up. I think it would be a nice service to the users to enable the same functionality for earlier versions. At this point i might need to use other code (Set-DbaPrivilige as you suggest).

SqlInstall.ps1
reitse: As this is a config option from SQL 2016 up as well, i believe it is in scope, for the same reasons the perform volume maintenance setting should be in scope. At least for 2016 and up and again as a service for the earlier versions.

SqlInstall.ps1
reitse: I've been thinking about this one for a while before. Should it be included or not, because in the settings i'm asking for a password and at the end i'm disabeling the thing. I can't really make a good case for it, so it's removed.

SqlInstall.ps1
reitse: I hope I've got all of them...

SqlInstall.ps1
reitse: Yeah that's really cool!

reitse:
I think we need to differentiate between editions. But, does the installer ini require a key? Or will that question come up during installation? I haven't got any keys i can use so testing that one is a bit hard...

SqlInstall.ps1
reitse: There are more combinations that shouldn't work. Even with more parameters to come...

SqlInstall.ps1
reitse: 👯‍♂️

potatoqualitee:
Thanks so much, super fast, awesome! I'll look again in a bit.

Wrt the key question, looks like it may - with that 2008 config file, it has `PID=$Key`

A few things I noticed when poking around the configs:

1. your info is in the inis - perhaps left over from when it was writing to the config files, but that needs to be replaced
2. If someone is using a named instance, please enable the browser by default (`BROWSERSVCSTARTUPTYPE="Enabled"`
3. Please allow the user to change their default collation. Right now it's static.

SqlInstall.ps1
potatoqualitee: great q, im unsure. you can make all of the changes you'd like and then rename, then ping me and ill give it a fresh look

SqlInstall.ps1
potatoqualitee: great point, that'd provide backwards compat for things that are offered in newer binaries. sold

SqlInstall.ps1
potatoqualitee: agreed, great points

potatoqualitee:
actually, once you do the rename, please create a new PR, i'd like to start fresh. The outdated bit isn't updating like i'd hoped.

PR #4163, opened 2018-10-09
 Export-DbaDacPackage example updatd

Splaxi:
## Type of Change

 - [ ] Bug fix (non-breaking change, fixes #<enter issue number>)
 - [ ] New feature (non-breaking change, adds functionality)
 - [ ] Breaking change (effects multiple commands or functionality)
 - [ ] Ran manual Pester test and has passed (`.\tests\manual.pester.ps1)
 - [ ] Adding code coverage to existing functionality
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [X] Documentation
 - [ ] Build system
 

### Purpose
Better example

PR #4166, opened 2018-10-09
Fix Clear-DbaPlanCache  + example added

Splaxi:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [X] Documentation
 - [ ] Build system
 

### Purpose
To fix an issue with the $server variable not being initialized correctly.
Provide pipeline example for the users to see how to use it.

### Error description
If you use the cmdlet / function without having connected to any SQL Server instances earlier, it fails.
When piping in cache plans using the `Get-DbaPlanCache -SqlInstance localhost | Clear-DbaPlanCache -Threshold 200` it would also fail.

**The error reported is:**
```
PS C:\windows\system32> Get-DbaPlanCache -SqlInstance localhost | Clear-DbaPlanCache -Threshold 200
You cannot call a method on a null-valued expression.
At C:\Program Files\WindowsPowerShell\Modules\dbatools\0.9.472\allcommands.ps1:1809 char:21
+ ...                    $server.Query("DBCC FREESYSTEMCACHE('SQL Plans')")
+                        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
    + FullyQualifiedErrorId : InvokeMethodOnNull
```

### Approach
The foreach loop that handles the $InputObject variable, doesn't take into account that there should be a $server variable declared, initialized and connected, before running the `$server.Query("DBCC FREESYSTEMCACHE('SQL Plans')")` code line.

### Screenshots

image

### DIFF
--- a/functions/New-DbaLogin.ps1
+++ b/functions/New-DbaLogin.ps1
@@ -103,6 +103,11 @@
 
         Copies logins [Login1] and [Login2] from instance sql1 to instance sql2, but enforces password and expiration policies for the new logins. New logins will also have a default database set to [tempdb] and will be created in a disabled state.
 
+    .EXAMPLE
+        PS C:\> New-DbaLogin -SqlInstance sql1 -Login domain\user
+
+        Creates a new Windows Authentication backed login on sql1. The login will be part of the public server role.
+        
 #>
     [CmdletBinding(SupportsShouldProcess = $true, DefaultParameterSetName = "Password")]
     param (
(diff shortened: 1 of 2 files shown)


functions/New-DbaLogin.ps1
```diff
@@ -103,6 +103,11 @@
 
         Copies logins [Login1] and [Login2] from instance sql1 to instance sql2, but enforces password and expiration policies for the new logins. New logins will also have a default database set to [tempdb] and will be created in a disabled state.
 
+    .EXAMPLE
```
wsmelton: You need to revert this file back to the version in the current development branch. You included this update in #4163

potatoqualitee:
ditto as previous, thank you! ☀️

Issue #3350, opened 2018-03-07
Create Pester Tests for Existing Functions 


joshcorr:
There are already a substantial amount of cmdlets that do not currently have any associated pester tests. As stated in the post "Path to 1.0" https://dbatools.io/the-path-to-1-0/: 

> 'The most important things you can fix even without being a skilled developer: ScriptAnalyzer and Coverage... if it’s tested, tests describe the way the function behaves, so the more coverage a function has, the more stable it is.

 All functions need to have coverage tests to ensure a high level of stability for this module. This issue was created to put all of these tests in one location, without the need for a new issue on each test. 
This issue is intended to be worked on by multiple people. Please just be sure to assign yourself to the issue, comment on which test you are currently working on, and keep pull requests to 1-2 tests.

If your new to writing tests use some of these links to get started:
https://dbatools.io/improving-tests/
https://github.com/sqlcollaborative/dbatools/wiki#appveyor-setup
https://github.com/sqlcollaborative/dbatools-templates/blob/master/tests/Verb-Noun.Tests.ps1
https://mva.microsoft.com/en-US/training-courses/testing-powershell-with-pester-17650

This list was gathered using the following methodology. 
```
PS C:\PowerShell\Dev>  $functions = Get-ChildItem .\Dbatools\functions\
PS C:\PowerShell\Dev> $tests = Get-ChildItem .\Dbatools\tests\
PS C:\PowerShell\Dev> (diff $functions.basename $tests.basename.replace('.Tests','')).Where({$_.Si
deIndicator -eq '<='})
```

**NOTE:** This is not a comprehensive list of all the tests that need to be updated, only of new ones.
**NOTE:** Some of these may not be able to be tested fully depending on what they do
If a Test is missing please feel free to add it in alphabetical order.

List Updated **10/11/18**

- [ ] ConvertTo-DbaTimeline
- [x] Copy-DbaAgentAlert              
- [x] Copy-DbaAgentCategory           
- [x] Copy-DbaAgentJob                
- [x] Copy-DbaAgentOperator           
- [ ] Copy-DbaAgentProxyAccount    
- [ ] Copy-DbaAgentServer   
- [ ] Copy-DbaAgentSharedSchedule     
- [x] Copy-DbaCustomError             
- [ ] Copy-DbaDatabaseAssembly        
- [ ] Copy-DbaDatabaseMail     
- [ ] Copy-DbaDataCollector 
- [ ] Copy-DbaDbMail      
[...]

PR #4181, opened 2018-10-11
Replace -Detailed in Get-DbaTcpPort (#2217)

joshcorr:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #2217 #3350 )
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [x] Documentation
 - [ ] Build system
 

### Purpose
Refactor Get-DbaTcpPort to not use `-Detailed` parameter
Add new descriptive parameter called `-AllTcpPort` to replace `-Detailed`
Include Test for cmdlet where one had not previously existed

### Approach
Tried to completely refactor the cmdlet, but realized it was being used by `Test-DbaConnection` in it's default form. Came up with a more suited parameter name to replace `-Detailed` (although I am completely up to changing it if it doesn't fit). `Select-DefaultView` returns the same properties regardless of if this new switch is used or not. 

### Commands to test
Get-DbaTcpPort -SqlInstance sqltestbed
Get-DbaTcpPort -SqlInstance sqltestbed -AllTcpPort

### Screenshots
Using Default Form
tcp1

Piping to Select *
tcp2

### DIFF
--- a/functions/Get-DbaTcpPort.ps1
+++ b/functions/Get-DbaTcpPort.ps1
@@ -66,148 +66,148 @@
         Returns an object with server name, IPAddress (just ipv4), port and static ($true/$false) for every server listed in the Central Management Server on sql2014.
 
 #>
-        [CmdletBinding()]
-        param (
-            [parameter(Mandatory, ValueFromPipeline)]
-            [Alias("ServerInstance", "SqlServer")]
-            [DbaInstanceParameter[]]$SqlInstance,
-            [Alias("Credential")]
-            [PSCredential]$SqlCredential,
-            [switch]$Detailed,
-            [switch]$AllTcpPort,
-            [Alias("Ipv4")]
-            [switch]$ExcludeIpv6,
-            [Alias('Silent')]
-            [switch]$EnableException
-        )
-        begin {
-            Test-DbaDeprecation -DeprecatedOn 1.0.0 -Parameter Detailed
-        }
-        process {
-            foreach ($instance in $SqlInstance) {
-                if ($AllTcpPort -eq $true) {
-                    try {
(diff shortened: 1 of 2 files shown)


potatoqualitee:
Thanks so much! And thank you for the new test! The test failures are unrelated (appveyor died)

Unfortunately, the help was formatted back to the old style and our updated notes were reverted as well. Please resubmit with properly formatted help and I'll be happy to merge 👍

potatoqualitee:
As for the name, yours is clearer, but it doesn't sound right. Maybe it needs to be plural? Perhaps the detailed output should just be included by default?

potatoqualitee:
This command was written when we were using Detailed a lot. Since then, we decided on the other way. Showing a lot by default and `-Simple` for simplified output. Perhaps we should go that way.

joshcorr:
Okay, Indents are fixed. Should be clearer to review the changes now. Sorry about that.

I was thinking plural (`-AllTcpPorts`) makes more sense, but was trying to stick with the parameter guide of singular. I considered having the fuctionality flipped, but didn't want to make that decision by myself. I only know of one internal dependency that would need to change as well.

potatoqualitee:
Looks good, thank you. I understand about the singular vs plural for sure. What I've done in other commands is `-Exclude Databases` so i get to have both 😅  I say go for `-Simple`, what you say, @wsmelton ?

potatoqualitee:
okay this is super strange - it keeps failing on the same tests. @joshcorr - can you try creating a new branch and copying your changes into there. I'm a bit confused as to what's going on.

joshcorr:
Okay will do. I will close this PR and open a new one.

PR #4196, opened 2018-10-12
Speedup Copy-DbaDatabase

alevyinroc:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
On instances with large numbers of databases, the call to `$destServer.databases.refresh()` can take a long time but only the new database needs to be refreshed (for comparison against properties of the source DB)

### Approach
Replaced refreshing all databases on the instance with getting a reference to just the new database that was restored.

On my test instance with 300+ databases, this was approx. a 10% speed improvement.

### Commands to test
`Copy-DbaDatabase`

### DIFF
--- a/functions/Copy-DbaDatabase.ps1
+++ b/functions/Copy-DbaDatabase.ps1
@@ -1292,15 +1292,15 @@
                             }
                         }
                     }
-                    $destServer.Databases.Refresh()
+                    $NewDatabase = Get-DbaDatabase -SqlInstance $destServer -database $DestinationdbName
 
                     # restore potentially lost settings
                     if ($destServer.VersionMajor -ge 9 -and $NoRecovery -eq $false) {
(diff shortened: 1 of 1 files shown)


potatoqualitee:
excellent, thank you so much, Andy!

wsmelton:
Just noting that the test did not actually run on your commit @alevyinroc . Appveyor seems to have 🐩 out.

I'm curious to know if you tested this against your larger instance and what the time difference really is?

`Get-DbaDatabase` does good bit more than just doing `$destServer.Databases[$currentDB].Refresh()`

alevyinroc:
@wsmelton I didn't have an opportunity to test it against the big one on Friday but I will give it a shot in the next couple days.

alevyinroc:
@wsmelton on my big instance, I got 4 minutes with the new version and 7:50 with the original. Th database I copied is about 230MB.

PR #4211, opened 2018-10-14
CI, parallelize tests

niphlod:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [x] Build system
 

### Purpose
Make an appropriate usage of the capabilities of dbatools's appveyor concurrent builds 

### Approach
Added "part" env variable. For now it's just a mere split by two for every scenario but as it can be easily figured out, its assignment capabilities are dynamic (should we ever find ourselves in the need of directing more "power" towards a scenario rather than another one)

----

If this works at the very first commit, I'd be amazed, so refrain from merging "as fast as one can"

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,7 +5,7 @@ configuration: "Debug"
 
 before_build:
   - nuget restore bin\projects\dbatools\dbatools.sln
-  
+
 build:
   project: bin\projects\dbatools\dbatools.sln
 
@@ -13,7 +13,7 @@ after_build:
     # Removed to prevent credential exposure
     # - ps: Push-AppveyorArtifact msbuild.binlog
     # - ps: Push-AppveyorArtifact msbuild.log
-    
+
 version: 0.9.{build}
 
 cache:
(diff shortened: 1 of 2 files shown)


niphlod:
first round, 8 min 40 sec. Which was good but ... left me somewhat unsatisfied.

I followed a hunch .... given tests are splitted in "parts" considering an array lexicographically ordered, possibly all `Clear-` go in the very first part and all `Write-` go in the very last.

It could be a beneficial exercise to burn quickly a fresh mind... FYI mine did:
 - "we have two parts"
 - "`Get` tests are averagely faster than `Set` ones"
 - "do we have more Get or Set tests now?"
 - "will there be more Get or Set tests in a year ?" 
 - "but we could have more" 
 - "the issue is exacerbated by most of `Get` and `Set` being apart" 
 - "but  `Copy` goes with `Get`."
 - "please stop overthinking"

Following the last point, I threw a maybe-stupid-but-successfull ordering by a "more consistent algorithm for our usecase", which cut down times to 6 min 20 sec.

Hence I declare such "algo for our usecase" whavever is after "Verb-Dba" (see https://github.com/sqlcollaborative/dbatools/pull/4211/files#diff-8088af7e427ef1ac9c2c5232ab624d68R297 ). 
This seem to shuffle tests among parts more evenly. 

Also, this is now good to merge by me.

PR #4216, opened 2018-10-15
Show-DbaServerFileSystem Cred fixes #4215

### LINKED ISSUE #4215: Show-DbaServerFileSystem not using SqlCredential parameter
I'm trying to use the "Show-DbaServerFileSystem" with the -SqlCredential parameter. No matter if I use a variable with a valid credential object or if I ask for if within the commandlet with "(Get-Credential)", I alwas get an error message that the user who is running the script cannot connect (because he has no rights). Obviously the commandlet doesn't use the credentials provided by the parameter.

potatoqualitee:
also fixes the command in general

### DIFF
--- a/functions/Show-DbaServerFileSystem.ps1
+++ b/functions/Show-DbaServerFileSystem.ps1
@@ -1,4 +1,5 @@
-﻿function Show-DbaServerFileSystem {
+﻿#ValidationTags#CodeStyle,Messaging,FlowControl,Pipeline#
+function Show-DbaServerFileSystem {
 <#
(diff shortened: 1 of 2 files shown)


potatoqualitee:
@niphlod - is the encoding being rewritten here?

niphlod:
doesn't seem so but no worry, @wsmelton explained the reason behind utf8bom and we'll address CI checks (and eventually a mass-update) ASAP.

potatoqualitee:
ty 🙇

PR #4231, opened 2018-10-16
Issue4106

### LINKED ISSUE #4106: Copy-DbaDbTableData doesn't copy to correct table if >1 table with same name in different schemas

- [X] Ensure you are able to reproduce it on the latest released version (0.9.450)
- [X] Verified this bug is not already reported in an issue
- [X] Verified errors are not related to permissions
- [X] Can reproduce in a clean PowerShell session (_clean = `powershell -NoProfile`_)

## Steps to Reproduce

Copy-DbaDbTableData doesn't copy to correct table if there are >1 schema with the same table name, and you are trying to copy a table that is not the first schema alphabetically.

image

## Expected Behavior

Both the top and bottom call would write to ww schema version of table.

## Actual Behavior

If you do not specify a destinationTable explicitly, the code sets (246-248)

```
if ((Test-Bound -Not -ParameterName DestinationTable)) {
   $DestinationTable = $sqltable.Name
}
```
Then it calls this: (230)

```
$desttable = Get-DbaDbTable -SqlInstance 'useomdbn12048,1113' -Table $DestinationTable -Database $Database -EnableException -Verbose:$false | Select-Object -First 1
```
What happens is that since ```$sqltable.Name``` is just the table name, and does not include the schema, the ``` | Select-Object -First 1``` ends up pulling the first schema alphabetically with this table name.

I believe the fix would be to change the ```$DestinationTable = $sqltable.Name``` line to append the schem name from the ```$sqltable``` object as well:

so line 247 would become this:
```
$DestinationTable = $sqltable.Schema + '.' + $sqltable.Name;
```
At this point you can likely remove the ``` | Select-Object -First 1``` also

## Environmental data

PowerShell:
PSVersion                      5.1.14393.2430
BuildVersion                   10.0.14393.2430

SQL Server:
Microsoft SQL Server 2012 (SP1) - 11.0.3156.0 (X64)   May  4 2015 18:48:09   Copyright (c) Microsoft Corporation  Developer Edition (64-bit) on Windows NT 6.2 <X64> (Build 9200: )

sqllensman:
## Type of Change

 - [X] Bug fix (non-breaking change, fixes #4106 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [X] Documentation
 - [ ] Build system
 

### Purpose
Fix issues in Copy-DbaDbTableData
1. Fixing issues with Schema's not being honoured
2. Fixing issues with same table existing in multiple schema
3. Fixing issues with table being copied onto itself

### Commands to test
Copy-DbaDbTableData


sqllensman:
Code will not allow a table to be copied into itself (especially with Truncate option as this would actually just be truncating the table. Tests required some modification to 
(1) Update test that was doing this
(2) Include test for this behaviour

potatoqualitee:
Thanks so much! To confirm, this is not a breaking change? I saw the docs update from three part to two part.

sqllensman:
The "change" to two-part is actually just clarification of how command works and not a change in functionality.  The use of three part names is not really supported by the code

In the current code there is a test at line 211 to require a Database when using a SqlInstance rather than piped input. 

To resolve the Table there is a call at line 230 to
Get-DbaDbTable -SqlInstance $server -Table $Table -Database $Database 

A three-part name can be passed but the database name in table must be same as the Database parameter. Using something like Database AdventureWorks2014 and Table AdventureWorks2014.[dbo].[DatabaseLog] will still work but the Database part of the Tablename does not add any value.
A call with something like 
-Database AdventureWorks2014 -Table AdventureWorks2016.[dbo].[DatabaseLog]
will cause an error of table not found.

Strictly there is functionality change as 
(a) It will not allow copying a table into itself
(b) If a table exists in multiple schemas and the Schema name is not included it will not "guess" and pick first table found.

However, I think both of these were Bugs and not intended features.

potatoqualitee:
excellent, thank you for the explanation! merged and should be in master soon.

PR #4235, opened 2018-10-17
New command get and set db compat level

gbargsley:
Here is a go at a Get and Set command for database compatibility level.  I feel like the Set command might be missing some functionality, but it works for my use case.

Also, I worked on building a pester tests, but was not sure what to test.

### DIFF
--- a/functions/Get-DbaDbCompatLevel.ps1
+++ b/functions/Get-DbaDbCompatLevel.ps1
@@ -1,5 +1,5 @@
 ﻿function Get-DbaDbCompatLevel {
-<#
+	<#
     .SYNOPSIS
         Displays the compatibility level for database(s).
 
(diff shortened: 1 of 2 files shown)


niphlod:
doesn't this clash partly with Invoke-DbaDbUpgrade ?

gbargsley:
From what @wsmelton and @potatoqualitee  said since Invoke-DbaDbUpgrade sets to what SQL Version is on the box, this command would give flexibility to set the compat level you want not take the SQL version value.

niphlod:
happy to reply to them too ... that doesn't vouch as a good reason to have triple the code for no reason if not flexibility ... also, this set- doesn't do anything to check if db is ok after up/downgrading the compat level, which is a lot less safe than what Invoke-DbaDbUpgrade does.

gbargsley:
Ok, no problem.  It works for me.  You can close the PR if you would like.

Sorry about the problems.

niphlod:
I'm not closing, I'm reviewing the PR.... 
- `get-` is basically `Get-DbaDatabase - .... | Select-Object ComputerName, InstanceName, SqlInstance, Name, Compatibility` .... no more, no less. If we really need something like that....
- `set-` at most should be a wrapper for Invoke-DbaDbUpgrade, whose only missing feature is being able to specify a target version (so it'll just need a new param there). Again, if strictly needed, the best deal would be to:
- rename Invoke-DbaDbUpgrade to be Update-DbaDbCompat, and move it to internals
- add the -TargetCompatibility param
- let Set-DbaDbCompat be a wrapper of Update-DbaDbCompat
- let Invoke-DbaDbUpgrade to reuse the code of Update-DbaDbCompat, fixing -TargetCompatibility to the highest value (mapped by $server.Version)

wsmelton:
This won't clash with invoke command. Unless someone is going to refactor it and change the name so I can also downgrade a database?

wsmelton:
Also the purpose of the invoke command was for migrations that you upgrade a database from major version to another. It is not useful when a user just wants to adjust the compatibility of a database that has already been migrated. 

And before anyone ask I do not go and update stats every time I change the compatibility level of a database....it is not required.

niphlod:
@wsmelton : see comments from one hour ago ^_^

niphlod:
> I see now reason to have anything named..

I see "now" or "not" ?

I'll rephrase the steps from my comment above:
- `Invoke-DbaDbUpgrade` stays as Invoke
- the entirety of its code that gets "moved" to internals (at which point, I think either `Update-` or `Set-` are as good as they can be)..... but at that point, that internal function (with a single param added) will be reused by both the "newly-created" `Set-DbaDbCompat...` and the "old" `Invoke-DbaDbUpgrade`.

potatoqualitee:
i like these commands (and also would love one for size / auto-growth) and will review more in-depth tomorrow 💯

potatoqualitee:
can @gbargsley just wrap `Invoke-DbaDbUpgrade` instead of internalizing code, @niphlod ?

niphlod:
Sure but what changes? What I'd like to avoid is code duplication, and "my
goto place" place for reusable tools/function/whatnot is internals.
Maybe the only pro of reusing the public function is avoiding copy-pasting
the parameter, but there's the huge con of ending up with an
invoke-dbadbupgrade that has -targetcompatibility, which doesn't make sense
at all. Jm2c

Il gio 18 ott 2018, 20:34 Chrissy LeMaire <notifications@github.com> ha
scritto:

potatoqualitee:
that works - @gbargsley would you be able to make those changes?

gbargsley:
So let me understand. Remove my new get/set commands. And instead add a parameter to Invoke-DbaDbUpgrade to allow the setting of the compat version instead of it forcing the SQL version?

potatoqualitee:
my understanding is that you have your commands plus the additional parameter.

niphlod:
- rewrite get as a get-dbadatabase wrapper
- move the common code of invoke-dbadbupgrade to internals, adding the param for targetcompatibility
- rewrite invoke-dbadbupgrade to use code from internals
- rewrite your set to use the newly added function to internals, using the new parameter targetcompatibility

potatoqualitee:
ty niph

gbargsley:
I am not sure that is in my skill set to accomplish.

potatoqualitee:
k ill do it, merging now

potatoqualitee:
thanks to everyone for their feedback!

PR #4245, opened 2018-10-18
Shush removal failure on Remove-Item in test

potatoqualitee:
the red will appear elsewhere if needed


potatoqualitee:
@nvarscar - can you look into making the DbaService commands faster and more reliable? Appveyor fails a bit on this command, but that may be on appveyor. Ultimately, though, I'm thinking that it should be single threaded by default with a -Multithread param that enables multithreading. The commands are super fast on one of my virtual machines, but unusably slow on my laptop VM.

PR #4299, opened 2018-10-21
ShouldProcess updates

potatoqualitee:
small changes.

After looking at the docs, Find used to return some drop statements to be executed. We removed that but forgot to update the docs and shouldprocess. This whole command is read-only so I removed it. Not sure if whoisactive requires shouldprocess so I set its impact to low.


potatoqualitee:
fixed failing

PR #4308, opened 2018-10-22
Copy-DbaDatabase fake Join-Path : fix ugly output, 

potatoqualitee:
it threw but worked anyway (??), this makes it prettier


potatoqualitee:
🔥 🔥 🔥 🔥 🚒🚒🚒🚒

PR #4335, opened 2018-10-24
Fix mirror output

potatoqualitee:


potatoqualitee:
appveyor is 💩 the bed lately, man.

PR #4347, opened 2018-10-25
stalebot settings

niphlod:
As discussed on #dbatools-dev : 
 - 60 days till it gets marked as stale
 - 30 more days to get closed automatically
 - only applied to issues (just relabelled all bugs as such, so questions, features, discussions, documentation issues for the moment won't be touched)

Once this is in, we can add the bot. Once we get acquainted, I feel we should turn it on also on features... but that's for another discussion (and another time)

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,20 @@
+issues:
+  daysUntilStale: 60
+  daysUntilClose: 30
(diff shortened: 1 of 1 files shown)


potatoqualitee:
looks awesome! perfect setup.

PR #4348, opened 2018-10-25
Restore-DbaDatabase SQL2000 fixes

potatoqualitee:
/cc @pollusb  

image

Works and no need to do workaround -we do it for you. Not so sure how much that `break` will impact things.


potatoqualitee:
merging cuz all the tests pass 🎉

PR #4359, opened 2018-10-29
Import-DbaCsvToSql: Added path quoting when calling x86 shell

NReilingh:
When Import-DbaCsvToSql calls itself in the x86 shell, it has to build its own parameter string. The `-Csv` parameter was not being passed in in quoted form, so this call would break if the CSV file path has odd characters. This change wraps each path in single quotes, and also escapes single quotes internal to the path.

## Type of Change

 - [X] Bug fix (non-breaking change)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Fixes breakage that occurs with weird file names and -Safe mode in Import-DbaCsvToSql

### Commands to test
`Import-DbaCsvToSql -Safe`

### Screenshots

filename breakage

### DIFF
--- a/functions/Import-DbaCsvToSql.ps1
+++ b/functions/Import-DbaCsvToSql.ps1
@@ -676,12 +676,13 @@ function Import-DbaCsvToSql {
 
                 # Perform the actual switch, which removes any registered Import-DbaCsvToSql modules
                 # Then imports, and finally re-executes the command.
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Thank you so very much! 🎉

PR #4374, opened 2018-10-30
New AG commands

potatoqualitee:
Bunch of sets and some removes, also a bug fix or two. Tests coming next.


potatoqualitee:
right on!

PR #4426, opened 2018-11-06
Connect-*Instance, fix errors with Windows logins

niphlod:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Discussed at lenght in dbatools-dev . 
All starts when a patient dbatools user reports some weird errors when running backup-dbadatabase in a powershell session started with an unprivileged user but passing the correct credentials.
Turns out that Connect-DbaInstance and Connect-SqlInstance behaved differently. While Connect-DbaInstance at least bombed out when trying to setinitfields, Connect-SqlInstance was quiet.
That, in conjunction with dbatools general - and good - reuse of the $server variable, make the weird error pop up: lower-level functions (such as Test-DbaPath or Get-DbaBackupInformation, used within backup-dbadatabase) were reporting a different user than the one specified via -SqlCredential.

Long story short, this fixes it, but I really don't know why `$null = $server.IsMemberOfWsfcCluster` was there in the first place. What I'm afraid of (and I need the review for, mostly) is that this new piece of code will affect how the connectionpool is handled. 

Sorry I can't be more precise, but the issue is reallly subtle. To replicate, start a powershell session with an unprivileged user (such as a windows user that can't connect to the target instance) and try to issue a Backup-DbaDatabase -SqlInstance foo  ..... -SqlCredential $validcredentials , where $validcredentials are for a WINDOWS user.

Before this PR, windows users and sql users differed in behaviour. Now they match but.... did I screw something else in the process ?

### DIFF
--- a/functions/Connect-DbaInstance.ps1
+++ b/functions/Connect-DbaInstance.ps1
@@ -320,7 +320,7 @@ function Connect-DbaInstance {
                         $server.ConnectionContext.Connect()
                     } elseif ($authtype -eq "Windows Authentication with Credential") {
                         # Make it connect in a natural way, hard to explain.
-                        $null = $server.IsMemberOfWsfcCluster
+                        $null = $server.Information.Version
                         if ($server.ConnectionContext.IsOpen -eq $false) {
                             $server.ConnectionContext.Connect()
                         }
--- a/functions/Connect-DbaInstance.ps1
+++ b/functions/Connect-DbaInstance.ps1
@@ -298,8 +298,8 @@ function Connect-DbaInstance {
                 }
 
                 try {
-                    if ($null -ne $Credential.username) {
-                        $username = ($Credential.username).TrimStart("\")
+                    if ($null -ne $Credential.UserName) {
+                        $username = ($Credential.UserName).TrimStart("\")
 
                         if ($username -like "*\*") {
                             $username = $username.Split("\")[1]
@@ -321,12 +321,19 @@ function Connect-DbaInstance {
                     } elseif ($authtype -eq "Windows Authentication with Credential") {
                         # Make it connect in a natural way, hard to explain.
                         $null = $server.IsMemberOfWsfcCluster
+                        if ($server.ConnectionContext.IsOpen -eq $false) {
+                            $server.ConnectionContext.Connect()
+                        }
                     } else {
                         $server.ConnectionContext.SqlConnectionObject.Open()
                     }
                 } catch {
-                    $message = $_.Exception.InnerException.InnerException
-                    $message = $message.ToString()
+                    $originalException = $_.Exception
+                    try {
+                        $message = $originalException.InnerException.InnerException.ToString()
+                    } catch {
+                        $message = $originalException.ToString()
+                    }
                     $message = ($message -Split '-->')[0]
                     $message = ($message -Split 'at System.Data.SqlClient')[0]
                     $message = ($message -Split 'at System.Data.ProviderBase')[0]
--- a/internal/functions/Connect-SqlInstance.ps1
+++ b/internal/functions/Connect-SqlInstance.ps1
@@ -117,7 +117,7 @@ function Connect-SqlInstance {
                 $server.ConnectionContext.Connect()
             } elseif ($authtypeSMO -eq "Windows Authentication with Credential") {
                 # Make it connect in a natural way, hard to explain.
-                $null = $server.IsMemberOfWsfcCluster
+                $null = $server.Information.Version
                 if ($server.ConnectionContext.IsOpen -eq $false) {
                     $server.ConnectionContext.Connect()
                 }
--- a/internal/functions/Connect-SqlInstance.ps1
+++ b/internal/functions/Connect-SqlInstance.ps1
@@ -186,7 +186,7 @@ function Connect-SqlInstance {
             $server.ConnectionContext.Connect()
         } elseif ($authtype -eq "Windows Authentication with Credential") {
             # Make it connect in a natural way, hard to explain.
-            $null = $server.IsMemberOfWsfcCluster
+            $null = $server.Information.Version
             if ($server.ConnectionContext.IsOpen -eq $false) {
                 $server.ConnectionContext.Connect()
             }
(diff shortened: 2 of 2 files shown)


potatoqualitee:
It is an incredibly subtle and tough issue. `$null = $server.IsMemberOfWsfcCluster` has a comment above it that says `# Make it connect in a natural way, hard to explain.` When you call a property (it can be any property that isn't initially loaded, I think) it connects in a way that helps the connection pool.

Do you have two domains you can test with to ensure cross domain connects still work? You can see my PR for adding that line here #3835 (man, github is awesome)

niphlod:
nope. 
But....instead of being bitten by sudden SMO changes (see the slowness when you do now $server.Databases[dbname] that was hella faster and now makes me cry, I'll do a mass PR later on),  instead of relying of a somewhat undocumented behaviour of SMO , can't we just use the method that someone uses to pick a connection from the pool, else open it ? 
Because it seems that what you originally tested clashes bad with what is happening now... and it's quite easy to replicate.
And no, I don't have cross-domain, but is it really that different from a windows auth ? The code doesn't seem to make much difference, I guess it's the underlying methods that do the magic of kerberos vs ntlmv2

potatoqualitee:
> instead of relying of a somewhat undocumented behaviour of SMO , can't we just use the method that someone uses to pick a connection from the pool, else open it ?

Can you please explain in code? 😓 

> but is it really that different from a windows auth ?

Yes, but I don't know why.

@Zhunya are you around to test Niph's PR? It needs to be tested cross-domain.

niphlod:
my "problem" is indeed code (and why I asked for a review). 
Is there some documentation on how to handle the connectionpool "explicitely" ? Relying on SMO trying to connect when evaluating the lazy property IsMemberOfWsfcCluster works JUST for SQL logins, now. And fails for Windows ones.
Let's imagine we take a different shortcut ........ what happens tomorrow ?!

Not being a .NET developer is a hard limitation, I know, but is it possible that there's no place where there is the "howto" of managing a connection pool within SMO ?

potatoqualitee:
>  Relying on SMO trying to connect when evaluating the lazy property IsMemberOfWsfcCluster works JUST for SQL logins, now. And fails for Windows ones.

Oh wow, now as in dev branch or now as in this branch?

> but is it possible that there's no place where there is the "howto" of managing a connection pool within SMO ?

When I looked up explicit info, it was inaccurate or non existent. No docs that I found ever described the type of stuff we're seeing. I'd love to know how SSMS handles connections (Maybe Azure Data Studio can provide insight since it's open source) because it always does it right. Connect-SqlInstance is a hack, but I'm unsure how to get it right without reverse engineering some of MSFT's stuff.

potatoqualitee:
Perhaps we can explore this https://github.com/Microsoft/azuredatastudio/tree/master/src/sql/parts/connection

potatoqualitee:
That's too manual. So I went back to Connecting with SMO, and they basically do the same, except they call another variable.

image

https://docs.microsoft.com/en-us/sql/relational-databases/server-management-objects-smo/create-program/connecting-to-an-instance-of-sql-server?view=sql-server-2017

potatoqualitee:
Perhaps we can look at the SqlServer module, though we'd have to test if they do it right, too. They may have made the same assumptions.

niphlod:
> Oh wow, now as in dev branch or now as in this branch?

now as in development. this branch *seems* to solve the issue, but I don't know how it affects the pool itself, although it's kinda "crazy" worrying because in development we do $server.ConnectionContext.Connect() already for SQL Logins and $server.IsMemberOfWsfcCluster for windows ones.

> When I looked up explicit info, it was inaccurate or non existent. No docs that I found ever described the type of stuff we're seeing.

sounds like someone should speak up 💃 

> Connect-SqlInstance is a hack
that bit (	` $server.IsMemberOfWsfcCluster `) for sure , but it's the same in Connect-DbaInstance and since we use it pretty much everywhere, we should stop calling it a hack

potatoqualitee:
> this branch seems to solve the issue

Excellent, @sanderstad was having issues previously and has a multi domain instance to test, he's firing up a laptop now. If it works for him, we should be set.

> sounds like someone should speak up

Yes indeed. If anyone does, please let us know here. My issue is the time and in-depth understanding of what it is we need to report and request.

> we should stop calling it a hack

True, I should be more accurate. Due to the lack of official documentation, Connect-SqlInstance is a culmination of codified experiences that is not yet as reliable as SSMS' connect dialog, but is getting closer every day.

I'm starting to wonder if SSMS even uses SMO to connect. I'll ask Microsoft.

potatoqualitee:
I just updated it to use `$server.Information.Version` instead of `$server.IsMemberOfWsfcCluster`

niphlod:
okay, just to be clear ..... what we need is:
- to establish a connection to a target with the provided credentials (or use integrated auth without)
- if possible, use a connection pool, which means
  - if it's the "first time", establish the connection and, if successfull, put that connection into a pool of "usable ones"
  - if it's the "second time", reuse "usable ones" or open another one (till pool exhaustion), cloning the properties of the "original one"
  - do not keep 3120998312098 open in standby, which is, use a pool that is actually a pool, and with sane timeouts (except commands that may need longer ones, but those already specify a different timeout value)


If the "hack" of calling `.Information.Version` is establishing a connection and using a connection pool, it apparently does what 

`if ($server.ConnectionContext.IsOpen -eq $false) { $server.ConnectionContext.Connect() }`

do (again, not documented, but it's happening in development already for any non-windows auth)

So, for the sake of avoiding possible "problems", I'd say we do the same thing for both SQL and Windows logins, but it _should_ be either (in boolean terms, OR):
- the call to `.information.version` 
- `if ($server.ConnectionContext.IsOpen -eq $false) { $server.ConnectionContext.Connect() }`

potatoqualitee:
We have so many Issues around this, but I swear, `$server.ConnectionContext.Connect()` is distinctly different from `.Information.Version`. 

Maybe it was related to #3809 

image

I'd really prefer to keep things as is because we keep getting unexpected situations with Connect-*Instance. Perhaps what we could do is document the whys a little better so that we have more info next time.

Btw, @sanderstad has failures BUT they are 

* failures that occur in dev
* but they are failing better this time

It's the cross-domain issue that I still haven't figured out. I'm happy to document and also happy to merge at this point.

potatoqualitee:
will merge and then now working on documentation with prs and stuff in the code

PR #4456, opened 2018-11-09
Restore-DbaDatabase and Backup-DbaDatabase - improve messaging

potatoqualitee:
This should be a temporary workaround so I won't close that other one.


potatoqualitee:
image

Now looks like this 👍

PR #4475, opened 2018-11-10
Fixing "Failure | You cannot call a method on a null-valued expression."

Splaxi:
## Type of Change

### Purpose
 
Fixing missing output when one of the three checks is empty.

PS C:\Users\administrator.LAB> Get-DbaPrivilege -Verbose
VERBOSE: [22:35:03][Test-PSRemoting] Testing GJ-AX2012CU13
VERBOSE: [22:35:03][Get-DbaPrivilege] Getting Privileges on GJ-AX2012CU13
VERBOSE: [22:35:03][Get-DbaPrivilege] Getting Batch Logon Privileges on GJ-AX2012CU13
VERBOSE: [22:35:03][Get-DbaPrivilege] Getting Instant File Initialization Privileges on GJ-AX2012CU13
VERBOSE: [22:35:03][Get-DbaPrivilege] Getting Lock Pages in Memory Privileges on GJ-AX2012CU13
WARNING: [22:35:03][Get-DbaPrivilege] Failure | You cannot call a method on a null-valued expression.

In my case it was that there was no entries at all in the Lock Pages in Memory.

It seems that you expect at least one entry in either and you cannot be entirely sure on that.

### Approach

Fetching the output from each check into a local variable and testing if it is $null or not.

### DIFF
--- a/functions/Get-DbaPrivilege.ps1
+++ b/functions/Get-DbaPrivilege.ps1
@@ -1,4 +1,4 @@
-function Get-DbaPrivilege {
+﻿function Get-DbaPrivilege {
     <#
     .SYNOPSIS
         Gets the users with local privileges on one or more computers.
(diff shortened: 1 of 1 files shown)


potatoqualitee:
awesome, thanks so much! works beautifully 🙇

PR #4477, opened 2018-11-11
Recoveryfork2

Stuart-Moore:
## Type of Change

 - [x] Bug fix (non-breaking change,)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Warnings in Get-DbaBackupHistory for multiple recovery forks now only happen if Get-DbaBackupHistory is the first command called. So won't trigger with Backup-DbaDatabase or Copy-DbaDatabase for example.

Piping will trigger it as PSCallstack doesn't catch that.

Removed leading 0s fom Numeric LSN output and fixed up the regex as well.


potatoqualitee:
code makes sense, and all tests pass, thank you!

PR #4502, opened 2018-11-13
fix for Invoke-DbatoolsFormatter

niphlod:
fix for weird corner-cases. This logic should be sounder.


potatoqualitee:
ty, it worked on my test cases!

PR #4509, opened 2018-11-14
Attempt to fix #4388

### LINKED ISSUE #4388: Copy-DbaDbDataTable no longer supports array of tables

- [ X] Ensure you are able to reproduce it on the latest released version (_we release often_)
- [ X] Verified this bug is not already reported in an issue
- [ X] Verified errors are not related to permissions
- [ X] Can reproduce in a clean PowerShell session (_clean = `powershell -NoProfile`_)

The check on 246:

```
if ($dbTable.Count -eq 1) {
                    $InputObject += $dbTable
                } else {
                    Write-Message -Level Warning -Message "The table $Table matches $($dbTable.Count) objects. Unable to determine which object to copy"
                    continue
}
```
Works for >1 table being piped into the function, but this is what breaks the string array of tables from being provided to the function directly from succeeding.

## Steps to Reproduce

image

## Expected Behavior

Each of the 7 tables would be copied from $prodInstance to $uatInstance

## Actual Behavior

Error catches this and will not let it proceed

## Environmental data

PSVersion                      5.1.14393.2248
BuildVersion                   10.0.14393.2248

Splaxi:
The $Table variable is an array and we need to loop through it.

## Type of Change

 - [X] Bug fix (non-breaking change, fixes #4388)

### DIFF
--- a/functions/Copy-DbaDbTableData.ps1
+++ b/functions/Copy-DbaDbTableData.ps1
@@ -242,13 +242,16 @@ function Copy-DbaDbTableData {
             }
 
             try {
-                $dbTable = Get-DbaDbTable -SqlInstance $server -Table $Table -Database $Database -EnableException -Verbose:$false
-                if ($dbTable.Count -eq 1) {
-                    $InputObject += $dbTable
-                } else {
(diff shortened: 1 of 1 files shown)


potatoqualitee:
aww yisss! 🎉  thank you, Mötz

PR #4563, opened 2018-11-17
fixes #2776

### LINKED ISSUE #2776: Update-SqlDbReadOnly, drop in favour of Set-DbaDatabaseState
Update-SqlDbReadOnly should be replaced with Set-DbaDatabaseState

potatoqualitee:
also formats

changes the logic too. man, i used to make programming so much harder than it should be with all that true/false business

### DIFF
--- a/functions/Copy-DbaDatabase.ps1
+++ b/functions/Copy-DbaDatabase.ps1
@@ -240,7 +240,7 @@ function Copy-DbaDatabase {
             Stop-Function -Message "-Continue cannot be used without -UseLastBackup"
             return
         }
-        
+
         function Join-Path {
             <#
         An internal command that does not require the local path to exist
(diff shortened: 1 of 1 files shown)


Latupho:
Using Pipeline is a good thing to learn fast, and to build into one's cmdlets / Advanced Functions.
The last one is not my strong side... but it's all ways on my todo list to do later...

potatoqualitee:
Stop-Function is available in PSFramework as `Stop-PSFunction`. It helps turn all that red into warnings, unless `-EnableException` is used. PSFramework is a great stating point for new modules because it gives you the instant ability to log, set configs and handle exceptions differently. I use it in all of my other modules.

Invoke-DbatoolsFormatter rewrites the format to abide by PSSA (PowerShellScriptAnalyzer). Frikken cool. No idea how @niphlod did it but bravo there.

PR #4573, opened 2018-11-17
Fixes #3443- Exclude 'user mapped to a certificate' from OrphanUsers

### LINKED ISSUE #3443: Remove-DbaRemoveOrphan not excluding valid user ModuleSigner in SSISDB
Hello,
I see now with the current upgrades that the remove-DbaRemoveOrphan users is not being utilized.
It seems with this function the SSISDB permissions used to execute SSIS packages.
So I thought it might be a good idea to log it with you guys so that it can be mentioned for other people in case they, too remove the SSIS permissions needed to execute jobs (woopsy) I got a lot of interesting emails afterwards.

Thank you so much!

/cc @ClaudioESSilva

ClaudioESSilva:
Added exclusion for "User mapped to a certificate" login type

## Type of Change

 - [x] Bug fix (non-breaking change, fixes #3443)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Exclude the user type `User mapped to a certificate` from the validations like `SQL user without login`.

### Commands to test
`Get-DbaOrphanUser`
`Remove-DbaOrphanUser`
`Repair-DbaOrphanUser`

@wsmelton can you please try it? Thanks!

### DIFF
--- a/functions/Remove-DbaOrphanUser.ps1
+++ b/functions/Remove-DbaOrphanUser.ps1
@@ -148,7 +148,7 @@ function Remove-DbaOrphanUser {
                         } else {
                             Write-Message -Level Verbose -Message "Validating users on database $db."
 
-                            if ($User.Count -eq 0) {
+                            if ($User.Count -ge 1) {
                                 #the third validation will remove from list sql users without login  or mapped to certificate. The rule here is Sid with length higher than 16
                                 $User = $db.Users | Where-Object { $_.Login -eq "" -and ($_.ID -gt 4) -and (($_.Sid.Length -gt 16 -and $_.LoginType -in @([Microsoft.SqlServer.Management.Smo.LoginType]::SqlLogin, [Microsoft.SqlServer.Management.Smo.LoginType]::Certificate)) -eq $false) }
                             } else {
--- a/functions/Remove-DbaOrphanUser.ps1
+++ b/functions/Remove-DbaOrphanUser.ps1
@@ -149,12 +149,12 @@ function Remove-DbaOrphanUser {
                             Write-Message -Level Verbose -Message "Validating users on database $db."
 
                             if ($User.Count -eq 0) {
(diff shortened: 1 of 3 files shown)


functions/Remove-DbaOrphanUser.ps1
```diff
@@ -149,13 +149,11 @@ function Remove-DbaOrphanUser {
                             Write-Message -Level Verbose -Message "Validating users on database $db."
 
                             if ($User.Count -eq 0) {
-                                #the third validation will remove from list sql users without login. The rule here is Sid with length higher than 16
-                                $User = $db.Users | Where-Object { $_.Login -eq "" -and ($_.ID -gt 4) -and (($_.Sid.Length -gt 16 -and $_.LoginType -eq [Microsoft.SqlServer.Management.Smo.LoginType]::SqlLogin) -eq $false) }
+                                #the third validation will remove from list sql users without login  or mapped to certificate. The rule here is Sid with length higher than 16
```
wsmelton: This block of code will always be called, you will never hit the else statement. So all you are doing is increasing the need to maintain the code every time a new login type is created in SQL Server or Azure SQL.

There is no reason to validate a login type because it never has a login, so this will always filter it out: `($User -contains $_.Name)`

wsmelton: Correction:

There is no reason to validate if a user has a login and what login type it is because if your user count is greater than `0` you hit your else statement and this validation in your filter is what excludes the user properly: `($User -contains $_.Name)`.

A user in a database that is authenticated with a certificate will never have a login so it will never match the name properly.

potatoqualitee:
hey @wsmelton - can you share the code to fix? I'm unsure if claudio will be available and I can update his code.

wsmelton:
Line 151 just needs to be changed to "-ge 1"

potatoqualitee:
yesssssss ty, passes, merging!

PR #4576, opened 2018-11-19
Connect-DbaInstance - enable other domain credentials

potatoqualitee:
fixed a bug. Turns out, if you log into a trusted domain, you need to use the user@domain.local syntax. this was previously fixed in connect-sqlinstance


potatoqualitee:
closing and starting again

Issue #4556, opened 2018-11-16
New command(s) Test-DbaAvailabilityGroup	(like Test-SqlAvailabilityGroup)


ijeb:
---
name: Feature request
about: Suggest a new feature or enhancement

---

### Summary of new feature

When I use PowerShell to manage Availability Groups, I use Test-SqlAvailabilityGroup a lot. But I don't like the -Path parameters (it has limitations). So adding these commands is helpful:
Test-DbaAvailabilityGroup	
Test-DbaAvailabilityReplica	
Test-DbaDatabaseReplicaState	

Maybe just a wrapper for Test-SqlAvailabilityGroup

This does not seem to work:
Get-DbaAvailabilityGroup -SqlInstance sqlserver-0 -AvailabilityGroup test-ag | Test-SqlAvailabilityGroup 

I can (try to) create a PR for this.

Looks like #543, but more simple.

Test-SqlAvailabilityGroup: Evaluates the health of an availability group.
Test-SqlAvailabilityReplica:Evaluates the health of availability replicas.
Test-SqlDatabaseReplicaState: Evaluates the health of an availability database.

PR #4599, opened 2018-11-20
Create Test-DbaAvailabilityGroup.ps1

ijeb:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality) #4556
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
New command #4556

### Approach
Using Test-SqlAvailabilityGroup, seems not to support -SqlCredential.

### Commands to test
Examples in the help

### DIFF
--- a/functions/Test-DbaAvailabilityGroup.ps1
+++ b/functions/Test-DbaAvailabilityGroup.ps1
@@ -15,15 +15,19 @@ function Test-DbaAvailabilityGroup {
     .PARAMETER AvailabilityGroup
         Return only specific availability groups.
 
-    .PARAMETER AllowUserPolicies
+    .PARAMETER IncludeUserPolicy
 
         Indicates that this function tests user policies found in the policy categories of Always On Availability Groups.
 
-    .PARAMETER ShowPolicyDetails
+    .PARAMETER IncludePolicyDetail
 
         Indicates that this function displays the result of each policy evaluation that it performs. The function returns one object per policy evaluation. Each policy object includes the results of evaluation. This information includes
         whether the policy passed or not, the policy name, and policy category.
 
+    .PARAMETER InputObject
+
+        Enables piping from Get-DbaAvailabilityGroup
+    
     .PARAMETER EnableException
         By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
         This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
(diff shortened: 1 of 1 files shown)


functions/Test-DbaAvailabilityGroup.ps1
```diff
@@ -0,0 +1,103 @@
...
+
+    #>
+    [CmdletBinding()]
+    param (
+        [DbaInstanceParameter[]]$SqlInstance,
+        [PSCredential]$SqlCredential,
+        [string[]]$AvailabilityGroup,
+        [parameter(ValueFromPipeline)]
+        [Microsoft.SqlServer.Management.Smo.AvailabilityGroup[]]$InputObject,
+        [switch]$IncludeUserPolicy,
```
potatoqualitee: I changed these to singular. I'm surprised that the SqlServer module does plural params.

functions/Test-DbaAvailabilityGroup.ps1
```diff
@@ -0,0 +1,103 @@
...
+        [switch]$EnableException
+    )
+    process {
+
+        if ($SqlInstance) {
+            $InputObject += Get-DbaAvailabilityGroup -SqlInstance $SqlInstance -SqlCredential $SqlCredential -AvailabilityGroup $AvailabilityGroup
+        }
+
+        foreach ($ag in $InputObject) {
+            #$results = Test-SqlAvailabilityGroup -Path ($ag.Urn | Convert-UrnToPath) -IncludeUserPolicy:$IncludeUserPolicy -IncludePolicyDetail:$IncludePolicyDetail
```
potatoqualitee: I'm so sorry I didn't mention this earlier, we try not to rely on 3rd party modules. basically, if you can recreate whatever `Test-SqlAvailabilityGroup` does, that would be good. Though there are a number of ways to test an AG. Do you see any room for improvement over what `Test-SqlAvailabilityGroup`  has to offer? (I don't know the answer, I don't use it, but there may be opportunity here).

ijeb: I used this blog post [https://blogs.msdn.microsoft.com/sqlalwayson/2012/02/08/the-alwayson-health-model-part-1-health-model-architecture/](url).

The function should always return the same state information as the dashboard in SSMS (Health/Error/Warning). Otherwise it's still necessary to check the dashboard.

The health state is determined with a set of PBM policies. Dbatools does not have a function to evaluate a PBM policy (I think this would also need the sqlserver module).
I think smo or T-SQL cannot be used.

I do not see another option.

potatoqualitee:
k, will close for now and we'll work out something later, even if it's creating the underlying commands

PR #4634, opened 2018-11-22
Remove WindowsBase dependency, RIP CredentialPrompt

IISResetMe:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [X] Build system
 

### Purpose
 
In the interest of compiling dbatools.dll against .NET Core, I've removed `CredentialPrompt`, since that depended on `WindowsBase` (which isn't available in Core)

### Approach

The only internal consumer of the CredentialPrompt class was the constructor `DbaCredentialParameter()` which is never used anywhere else. Removed both completely and then removed the reference to WindowsBase in the project file


potatoqualitee:
thank you ! 💯💯💯

Issue #4610, opened 2018-11-21
Add database to availability group (on all nodes including backup/restore/seeding)


ijeb:
---
name: Feature request/idea :bomb:
about: Suggest a new feature or enhancement

---

### Summary of new feature

It's easy to create an availability group, for example
New-DbaAvailabilityGroup -Primary sqlserver-0 -Secondary sqlserver-1 -Name test-ag -ClusterType Wsfc -SeedingMode Automatic -Database db1

But after that, if I want to add db2, I have to take many actions:
Backup full/log / restore /  Add-DbaAgDatabase on all nodes.

It would be usefull to put this part of the code in a separate function. DBA's often want to remove a database from an ag and later add it again (including the seeding).

### Proposed technical details (if applicable)

New function with basically this code. Not the same as Add-DbaAgDatabase.

        # Add databases
        Write-ProgressHelper -StepNumber ($stepCounter++) -Message "Adding databases"

        $allbackups = @{
        }

        foreach ($db in $Database) {
            if ($SeedingMode -eq "Automatic") {
                if ($Pscmdlet.ShouldProcess($Primary, "Backing up $db to NUL")) {
                    $null = Backup-DbaDatabase -BackupFileName NUL -SqlInstance $Primary -SqlCredential $PrimarySqlCredential -Database $db
                }
            }

            if ($Pscmdlet.ShouldProcess($Primary, "Adding $db to $Name")) {
                $null = Add-DbaAgDatabase -SqlInstance $Primary -SqlCredential $PrimarySqlCredential -AvailabilityGroup $Name -Database $db
            }

            foreach ($second in $secondaries) {
                if ($Pscmdlet.ShouldProcess($second.Name, "Adding $db to $Name")) {
                    $primarydb = Get-DbaDatabase -SqlInstance $Primary -SqlCredential $PrimarySqlCredential -Database $db
                    $seconddb = Get-DbaDatabase -SqlInstance $second -Database $db

                    if ((-not $seconddb -or $Force) -and $SeedingMode -ne 'Automatic') {
                        try {
                            if (-not $allbackups[$db]) {
                                if ($UseLastBackup) {
                                    $allbackups[$db] = Get-DbaBackupHistory -SqlInstance $primarydb.Parent -Database $primarydb.Name -IncludeCopyOnly -Last -EnableException
                                } else {
                                    $fullbackup = $primarydb | Backup-DbaDatabase -BackupDirectory $SharedPath -Type Full -EnableException
[...]

PR #4648, opened 2018-11-23
Move Add databases to Add-DbaAgDatabase

ijeb:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality) #4610
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Use the same process/code for adding a database to an existing availability group.

### Commands to test
New-DbaAvailabilityGroup -Primary sqlserver-0 -Secondary sqlserver-1 -Name ag-1 -ClusterType Wsfc -SeedingMode Automatic -Database db1 -Confirm:$false
New-DbaAvailabilityGroup -Primary sqlserver-0 -Secondary sqlserver-1 -Name ag-1 -ClusterType Wsfc -SeedingMode Manual -SharedPath '\\sqlserver-0\Share' -Database db1


potatoqualitee:
hey thanks :D i think add-dbaagedatabase is missing from this pr

ijeb:
> hey thanks :D i think add-dbaagedatabase is missing from this pr

I tried to put two commits in one PR, but it didn't work ...

ijeb:
Added second commit

potatoqualitee:
looks like it worked this time 👍  looking forward to seeing the test outcome

potatoqualitee:
I added the 3 params to the unit tests to ensure we get a pass, tests are running now 👍 I do have two questions about the design

1. I saw a commented out `.SeedingMode =`, was that a question for me? Looks like it should be set or were you questioning if it should have been set when the replica joined and not at the db level? That could make sense.
2. There is a condition where there would be no output. Was that intentional?

thx

ijeb:
1 Parameter SeedingMode
My intention is:
If not specified, the setting from the availability group replica will be used (this is the case when called from New-DbaAvailabilityGroup).
If the SeedingMode is specified the function will update the ag replica setting (and use it). I can’t update the property and don’t know why this won’t work). Can’t find any example how to do it either.
2 No. I see it now and have to look at it.

potatoqualitee:
ah ha.. mmm

So the way i'd usually check is this `Get-DbaAgReplica -SqlInstance sql2016c | Get-Member` then look to see if SeedingMode is a `get;set;`

image

In this case it is, but I saw there are version restrictions. Are you using 2016 or above? It's also possible it can only be set at create? (I'm not sure, I generally just try stuff until it works)

ijeb:
It's 2017. The erros is The property 'SeedingMode' cannot be found on this object. Verify that the property exists and can be set.

potatoqualitee:
It is completely by accident that i may have figured this out? it has to be one replica at a time.

image

I didn't even really know that about PS

ijeb:
Wow thanks, that makes sense. Now the function that at least update the property SeedingMode.

potatoqualitee:
awesome! tests pass, thank you, i will merge and delete Test-DbaAvailabilityGroup

PR #4661, opened 2018-11-24
Build also against .Net Core 2.1

bergmeister:
First, I upgraded to the new csproj2017 project format using the [CsprojToVs2017 ](https://github.com/hvanbakel/CsprojToVs2017) tool by calling `./dotnet-migrate.exe migrate path/to/slnFile`
After this change, building requires either Visual Studio 2017 or the .Net Core SDK `2.1.403` as minimum version (the .Net SDK will already be installed as part of Visual Studio if users are on the latest minor version)
The next step was to add multi-targeting to build it against .Net Core 2.1 as well (i.e. in addition to `net452`), which compiles fine. The reference to `Microsoft.Management.Infrastructure` was changed to a NuGet reference as an additional improvement.

## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [x] Build system
 
### Purpose
Being able to build for .Net Core, requires the so called newer csproj2017 project format. This is not a breaking change for the existing full .Net Framework but means that Visual Studio 2017 has to be used (i.e. VS 2015 does not understand that format)


potatoqualitee:
Great news!! Thank you both! 🏆 🌮 🍸

PR #4663, opened 2018-11-24
changed internal name to Test-Windows

potatoqualitee:
from test-notwindows to test-windows. im still not sure cuz now you gotta have two () but this feels right


potatoqualitee:
unrelated failure

PR #4669, opened 2018-11-24
Set default VS-Code settings and recommend Powershell and csharp extension

bergmeister:
This is so that people do not have to configure the settings to be used for this repo manually, VS-Code will pick them up automatically and recommend the PowerShell and CSharp extensions when opening this project in the root folder, which is a common best practice for OSS projects.

## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,8 @@
+{
+    // See http://go.microsoft.com/fwlink/?LinkId=827846
+    // for the documentation about the extensions.json format
+    "recommendations": [
+        "ms-vscode.PowerShell",
+        "ms-vscode.csharp"
+    ]
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+// Place your settings in this file to overwrite default and user settings.
+{
+    "files.trimTrailingWhitespace": true,
+    "powershell.codeFormatting.preset": "OTBS"
+}
--- a/contributing.md
+++ b/contributing.md
@@ -45,12 +45,8 @@ When you are working with "objects" in SQL Server, say with databases, what vari
 [This page](https://github.com/sqlcollaborative/dbatools/wiki/Standard-Documentation) sums up what we currently use. We aim at standardizing and reducing to a set of self-documenting and reusable parameters. If you have any questions around the above do not hesitate to ask in Slack.
 
 ## Standardize formatting and indentation
-If you use VSCode, you can enable 
-```
-"files.trimTrailingWhitespace": true,
-"powershell.codeFormatting.preset": "OTBS"
-```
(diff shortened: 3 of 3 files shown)


potatoqualitee:
love it! thank you!

potatoqualitee:
whoa this is so cool!! check it out @wsmelton  and @niphlod  
 
image

wsmelton:
Yeah, this has been available for a while but we had agreed not to have tool specific files in the repository.

potatoqualitee:
ahh i dont think i understood the ramifications earlier. as end-user i find it super useful and would like to recommend gitlens and docker and perhaps remove the c# recommendation.

are there any more that are standard? /cc @SQLDBAWithABeard

wsmelton:
Docker won't do any good if it is not installed....it causes errors to be spit out in VS Code when it is not found by default. Anything that depends on a 3rd party app should not be included as recommended.

wsmelton:
Standard is also subjective, meaning folks opinion. You only need PS extension to contribute. I doubt most would even care about what GitLens extension offers. That is more around those that maintain and head up the project....but that is just my opinion :grin:

potatoqualitee:
okie what about markdownlint?

wsmelton:
If we wrote our help files in markdown like PS docs and all then yes; as it stands though I see no reason to have it on this project. (Microsoft has their own [authoring pack](docs-authoring-pack) for contributing to the Docs site format.)

SQLDBAWithABeard:
Bracket pair colouriser is the other one I use

https://marketplace.visualstudio.com/items?itemName=CoenraadS.bracket-pair-colorizer


But I agree with @wsmelton in general

I would keep these out of the repo and have a post saying what people use and show them how to use settings sync to have the same experience on any machine

Although on the other hand, it might be useful to help to keep coding guidelines such as 

number of tabs, where braces go etc

bergmeister:
The point of checking the VS code settings in is because if someone uses VS-Code, they *override* the user settings so that everything is setup how you (the project maintainers) want the contributors to format their submissions to be so that there won't by a discussion about formatting at the PR stage. As a contributor I don't want (or have the time) to change my *personal* settings every time I change working between different OSS projects. Yes, this is editor specific but is more meant to be a help in the automation of setup of expected settings, it does not mean that someone has to use VS-Code.
Recommending extension that are needed or very helpful for a project, is part of that, maybe moving the recommendation to the C# extension to the folder where the VS solution is would be slightly better if most contributions do not change C# code.

wsmelton:
We have no real reason for the C# extension. The project is for VS not VSCode.

potatoqualitee:
it wont be a universal prompt so nbd if it's in the dbatools solutions folder where only c# is at.

bergmeister:
I opened a PR to move it to the C# project folder, one can use either VS-Code or Visual Studio for working with C# projects, this just makes it easier if someone opens VS-Code in the C# project folder (which means they are interested in looking at C# code only)

PR #4687, opened 2018-11-27
New function - Update-DbaInstance

### LINKED ISSUE #4500: Update-DbaInstance
---
name: Update-DbaInstance
about: SQL Server patcher: SPs and CUs

---

### Summary of new feature

* Create a function that will complement Install-DbaInstance and re-use Adam Bertram's PSSqlUpdater https://github.com/adbertram/PSSqlUpdater/blob/master/SqlUpdater.ps1

* use Get-DbaBuildReference to keep all the details about the patches, update with filenames if needed
* Decide if we want to incorporate an automatic computer restart after installing a patch, wait for it to come back, and resume pending operations after restart

0.9.509

nvarscar:
Wish it could be smaller, but issues kept piling up until the very end, so bear with me 😇 
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #4500 )
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
SQL Server patching functionality, including remote patching, multiple versions at once, CredSSP support and so on.

### Approach
In short - I just kept throwing code at it until it gave up.

As discussed, I took [PsSqlUpdater](https://github.com/adbertram/PSSqlUpdater) from Adam Bertram as a reference and the core of it is in the PR. However, it was transformed to a point where I can no longer call it an adoption - there's almost no old code left, except for a few functions that work with remote registry (they are still huge though). All the functions were rewritten to support dbatools flow and were fully integrated with our build system (to a point where I had to add functionality to our build reference stuff). 

So, there is one main command - **Update-DbaInstance** that controls the overall flow and handles multiple installations.
There's **Install-SqlServerUpdate** - which actually handles the installation logic and calls all the checks and whistles.
There's also **Invoke-Program**, which takes care of remote execution through a protocol that allows for passing credentials, which are required for properly running SQLServer setup/patching. Those protocols are:
* CredSSP (there's Initialize-CredSSP that tries to configure both client and server to work together if possible)
* Default with PSSessionConfiguration lifehack (suggested by @niphlod), which creates a remote PSSession configuration with RunAsCredential option. This is a little less secure than CredSSP, but it's the only alternative around, it seems. I leave it up to reviewers to decide if it's secure enough for us. It serves as a fallback after CredSSP has failed, so users might end up never using it if CredSSP works as intended in their environment.

**Get-DbaBuildReference** was augmented with a couple more ParameterSets, as well as it now returns a few more properties, all of which are used by this new set of functions to ascertain current and target SQL Server builds and return proper KB names.

**Get-SqlInstanceComponent** was copied almost from PsSqlUpdater as is (with a few additions on how to figure current SQL Version), and is a slight concern since it uses remote registry features that completely ignore -Credential parameters and act on behalf of current user. I would love to address it later, but it's already too big of a change, so I would rather do it during next iteration once all this stuff was reviewed and all the suggestions implemented.

**Invoke-Command2** now supports -Authentication, -InputObject and remote PSSession configurations. Had to put Authentication type in the session name to avoid re-using a wrong session when protocols are switching, but I don't see any other ways of doing it except for changing the C# code, which I'm not ready to do just yet.

New PSSession configurations require remote WinRM service to be restarted (through... WinRM, hehe), which led to creating **Restart-WinRMService** function, which restarts the service and waits for the protocol to be brought back up.

Other stuff deserves less attention, but it's still sitting there - wrappers, code reuse stuff; some of them could actually be internal functions inside internal functions, but most of them might prove useful in other scenarios, which is why I isolated them into their own files.

### Commands to test
Ugh, there's only one, really: Update-DbaInstance. Pester tests are 100% UnitTests, without any remoting tests in place (for obvious reasons). I made this stuff work in my environment, but I'm completely unaware how it will work in your environment specifically, which is why .. please do test this stuff!

### Screenshots
Upgrading to latest version
image

No upgrade found - already at latest
image

Installing specific patch version and not finding it
image

### Learning
CredSSP had me crying at times
WinRM is a **** and de-serialization sucks
Parallel execution will come later - definitely after initial release


nvarscar:
There's a conflict due to `Test-Windows -NoWarn`, will resolve tomorrow, took me too much time to fix all the bugs before creating the PR

nvarscar:
.. aaand it's resolved. Please let me know what you guys think of this monster.

niphlod:
huuuuuuuuuuuuuge!

potatoqualitee:
The detailed description was extremely useful, thank you Kirill 👍

Zhunya:
It will be good to mention how to configure CredSSP on workstation and server

functions/Update-DbaInstance.ps1
```diff
@@ -0,0 +1,244 @@
...
+        Windows Credential with permission to log on to the remote server. Must be specified for any remote connection.
+
+    .PARAMETER Type
+        Type of the update: All | ServicePack | CumulativeUpdate. Only usable with -Latest.
+        Default: All
+
+    .PARAMETER KB
+        Install a specific update or list of updates. Can be a number of a string KBXXXXXXX.
+
+    .PARAMETER SqlServerVersion
```
potatoqualitee: Can this be Version or TargetVersion?

functions/Update-DbaInstance.ps1
```diff
@@ -0,0 +1,244 @@
...
+
+    .PARAMETER MajorVersion
+        Designed to work in conjunction with -Latest to only update specific version(s) of SQL Server. Syntax: SQL20XX or simply 20XX.
+
+    .PARAMETER Latest
+        Install latest SP and CU known to the module. More details can be found by running Test-DbaBuild -SqlInstance YourSqlServer
+
+        Note: `-Latest -Type CumulativeUpdate` will try to install CU from the latest SP available. Make sure to upgrate servers to proper SP beforehand.
+
+    .PARAMETER RepositoryPath
```
potatoqualitee: Please change to Path. Repository will be implied by docs and examples 👍

functions/Update-DbaInstance.ps1
```diff
@@ -0,0 +1,244 @@
...
+        Path to the folder(s) with SQL Server patches downloaded. It will be scanned recursively for available patches.
+        Path should be available from both server with SQL Server installation and client that runs the command.
+        All file names should match the pattern used by Microsoft: SQLServer####*-KB###-*x##*.exe
+        If a file is missing in the repository, the installation will fail.
+        Consider setting the following configuration if you want to omit this parameter: `Set-DbatoolsConfig -Name Path.SQLServerUpdates -Value '\\path\to\updates'`
+
+    .PARAMETER Restart
+        Restart automatically if required.
```
potatoqualitee: Restart Windows? or SQL or both?

functions/Update-DbaInstance.ps1
```diff
@@ -0,0 +1,244 @@
...
+        SP3 binary files will be
+
+    .EXAMPLE
+        PS C:\> Update-DbaInstance -ComputerName SQL1, SQL2 -Latest -Restart
+
+        Updates all applicable SQL Server installations on SQL1 and SQL2 with the most recent patch.
+        It will install latest ServicePack, restart the computers, install latest Cumulative Update, and finally restart the computer once again.
+
+    .EXAMPLE
+        PS C:\> Update-DbaInstance -ComputerName SQL1 -Latest -MajorVersion 2012 -Type ServicePack
```
potatoqualitee: Can/Should Latest be implied? and thus not required?

functions/Get-DbaBuildReference.ps1
```diff
@@ -58,16 +73,38 @@ function Get-DbaBuildReference {
 
     #>
     [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSUseShouldProcessForStateChangingFunctions", "")]
-    [CmdletBinding()]
+    [CmdletBinding(DefaultParameterSetName = 'Build')]
```
potatoqualitee: we've tried to move away from parametersets because they error really weirdly and confuse people. i know we have some commands that still use them but they're generally the older ones in need of a refactor.

can you change these into stop functions? if it's too much work/ablocker, we can do it later.

functions/Update-DbaInstance.ps1
```diff
@@ -0,0 +1,244 @@
...
+        Copyright: (c) 2018 by dbatools, licensed under MIT
+        License: MIT https://opensource.org/licenses/MIT
+
+        Requires Local Admin rights on destination computer(s).
+
+    .EXAMPLE
+        PS C:\> Update-DbaInstance -ComputerName SQL1 -SqlServerVersion SP3 -RepositoryPath \\network\share
+
+        Updates all applicable SQL Server installations on SQL1 to SP3.
+        SP3 binary files will be
```
potatoqualitee: will be .....? ;)

potatoqualitee:
can you add an `-InstanceName` similar to your service commands? Sometimes I just want to update one instance and not all at the same time (i lie, not me, but it seems like a likely scenario 😅)

functions/Get-DbaBuildReference.ps1
nvarscar: This one I'm slightly confused about. I love ParameterSets and if some of them confuse people, then the developer did something very wrong. Get-Help with defined parameterSets is much more nice, providing parameters grouped by usage. And they error weirdly when people are not providing proper parameters, not because parameterSets are evil. 
In this particular scenario, I want to be able to call function in four different ways. I don't see a benefit in mashing all the parameters together and leave users to figure it out, when we can define independent sets and have a nice and friendly way of confining users in appropriate boundaries.
But this is definitely not the hill where I'm ready to make a stand. If you say stop-fucntions, that's the way to go.

functions/Update-DbaInstance.ps1
nvarscar: or not will be! 💀 Will fix in the next set of commits :)

nvarscar:
> can you add an `-InstanceName` similar to your service commands? Sometimes I just want to update one instance and not all at the same time (i lie, not me, but it seems like a likely scenario 😅)

I thought about it, but then shared features got in a way - not sure how to properly ascertain shared features version and whether to include them into the update with limited subset of instances. Maybe in future releases? Or this is a must for initial release?

functions/Update-DbaInstance.ps1
potatoqualitee: Yeah I wouldn't want to use `-Lastest$false` ever, perhaps the param needs to be renamed? or made into a string with a validateset?

functions/Get-DbaBuildReference.ps1
potatoqualitee: > then the developer did something very wrong. 

Perhaps, but it's the way I naturally work. I didn't realize what the different blocks were in Get-Help for years, and always had to google the weird generic "parameter could not be resolved" error, and I spent about an hour before realizing I used two commands from conflicting parametersets. I would have appreciated if I was just told "You cannot specify Y when using X". But PS hasn't given us that option so we all decided a while back to give clarity to our users by telling them in human words that they can't use X when using Y. I wish parametersets would have been the solution, but they're just not user-friendly.

functions/Update-DbaInstance.ps1
potatoqualitee: lol thank you

potatoqualitee:
We can wait till a future release. I believe that Shared Features must be updated at the time the first one is updated, and that's fine, they're just shared features.

potatoqualitee:
This is dope af so far! I pointed it to a directory filled with a bunch of garbage plus the update file and it found it!!!! 😍 

image

But I did get one warning?

````
VERBOSE: [10:37:34][Update-DbaInstance] Launching installation on sql2017.base.local with following params: {"MajorVersion":[],"Latest":true,"Type":"CumulativeUpdate"}
VERBOSE: [10:37:34][Resolve-DbaBuild] Looking for 14.0.3038
VERBOSE: [10:37:34][Resolve-DbaBuild] We have 25 builds in store for this Release
WARNING: [10:37:34][Get-SqlServerVersion] Failed to process SQL versions | Version string portion was too short or too long.
VERBOSE: [10:37:34][Install-SqlServerUpdate] Found 1 applicable SQL Server version(s): 2017 (14.0.3038)
````

potatoqualitee:
Then it found it, you genius!

image

potatoqualitee:
This can be caught earlier. You can do $Computer.IsLocalhost to detect if it's local, then require the credential as one of the first checks.

image

potatoqualitee:
Can this red be caught? 

image

potatoqualitee:
image

so cool 👍

potatoqualitee:
It is not finding any instances on my localhost, even tho I ran as admin

image


Perhaps Get-DbaRegistryRoot can be used as a failsafe?

image


This machine, as you can see, has two instances. Btw, we have `Test-ElevationRequirement` that you can use in the event that admin is required when running locally.

nvarscar:
> This can be caught earlier. You can do $Computer.IsLocalhost to detect if it's local, then require the credential as one of the first checks.

My logic here is that if there is no update needed, there is no need to ask for credentials. It will only fail when it found an update to be installed and triggers Invoke-Program, which indeed checks for .IsLocalHost

nvarscar:
> Perhaps Get-DbaRegistryRoot can be used as a failsafe?
> 
> image
> 

It says network path unavailable, something with name resolution? Anyways, I'll change how this stuff is gathered in a next iteration, please let me know if that still happens. 

> This machine, as you can see, has two instances. Btw, we have `Test-ElevationRequirement` that you can use in the event that admin is required when running locally.

Test-ElevationRequirement is implemented, thanks for the suggestion!

potatoqualitee:
Unsure what to make of this, ill try on another server

image

nvarscar:
Hmm, maybe it tries to re-use the existing session after a restart of WinRM and it is invalid.. Gotta dig into that direction for a little bit. Thanks for the report!

potatoqualitee:
So, now I see why you wanted latest to be explicit. I was like "why isn't it finding my stuff?" 🤔  Then I realized it was not the latest, and specified the KB. Once we get this going, I'll be adding the Write-ProgressHelper that will expclitly say "KB and CU not specified, looking for latest" or something.

So far, none of my servers have worked. Is it the session reuse (i do restart ps each time btw) or am I missing something on my side? I did assume that if I can Invoke-Command, this would work.

image

nvarscar:
@potatoqualitee  I've pushed a few commits, old sessions will be removed after WinRM restart is triggered.
I wonder, why CredSSP is not working in your environment. Could you test `Invoke-Command -Authentication CredSSP` and check if there is something else that needs to be configured for CredSSP to make it work?

potatoqualitee:
man, no dice so far. this is insane. issue is on my side, it seems

potatoqualitee:
bravo! gonna merge then start a new PR for my progress bar changes

potatoqualitee:
💯💯💯🎉🎉🎉🎉🎊🎊🎊🎊💃💃💃

PR #4722, opened 2018-11-30
Config v4

FriedrichWeinmann:
> Not ready to merge! Will cause a test bloodbath

Submitting for initial run-up, is guaranteed to have 100% broken configuration.

> New commands to publish:

 - Export-DbatoolsConfig
 - Import-DbatoolsConfig
 - Reset-DbatoolsConfig
 - Unregister-DbatoolsConfig

> New internal commands of note:

 - Resolve-DbaPath
 - Join-DbaPath


potatoqualitee:
BE STILL MY HEART

potatoqualitee:
Once this merges, shall I replcae all `Join-Path` with `Join-DbaPath` ? Same for resolve? THANKS FOREVER FOR THIS

FriedrichWeinmann:
Replacing `Join-Path` with `Join-DbaPath` is totally cool - it does all the same things and more.
The performance impact is negligible.

`Resolve-Path` with `Resolve-DbaPath` depends:
- When accessing resources we know have to exist - specifically files within dbatools itself - stick to `Resolve-path` (better performance - especially for stuff during import!)
- When targetting paths outside of the module and wanting the extra utility or conformity: By all means.

Its specific design goals:
- Be able to ensure the provider
- Be able to resolve the path to a single file that needs not actually exist (e.g. for creating exports to file)
- Be able to ensure only one path can be resolved from it.

So either going at it with its specific use-cases in mind or just blanket for most internal use will work - so long as you keep it out of the import sequence itself.

potatoqualitee:
READY TO MERGEEEEEEEEEEEEEEEEEEEEEEEEEE XPLAT FA SHO

PR #4731, opened 2018-11-30
Restore-DbaDatabase - added dbaools default outputs

potatoqualitee:
Stuart, I moved the routine to an internal routine in `begin` to easily enable multi instance restore as requested by 👞  and I thought we supported it anyway.

i also added `ComputerName, InstanceName` to the output and ordered it properly. While I was at it, i changed `restoreinstance` to `restoreserver` to be closer to our `server` standard. Tho i recall `server` caused scoping issues in the past.

image


wsuhoey:
👞

potatoqualitee:
Tests failed for multi, so i reverted multi and left in our default outputs/new column aliases. those passed so i will mergre then work on restore-dbabackup multi changes

PR #4767, opened 2018-12-05
Fixing Remove-DbaOrphanUser

### LINKED ISSUE #4766: Remove-DbaOrphanUser doesn't properly check for users

- [X ] Ensure you are able to reproduce it on the latest released version (_we release often_)
- [X ] Verified this bug is not already reported in an issue
- [X ] Verified errors are not related to permissions
- [X ] Can reproduce in a clean PowerShell session (_clean = `powershell -NoProfile`_)

## Steps to Reproduce

```powershell
Remove-DbaOrphanUser -SqlInstance $devInstance -Database Sandbox -Verbose # Always returns 0 orphans
'domain\AdamWenger' | Remove-DbaOrphanUser -SqlInstance $devInstance -Database Sandbox -Verbose # Always removes all available orphans, not just the ones piped in
```

## Expected Behavior

Orphans are removed as specified (all or based on piping)

## Actual Behavior

No Piping results in an inability to find orphaned users, and any users in the pipe results in all users being removed.

## Environmental data

Powershell 5.1

Adam-Wenger:
## Type of Change

 - [X ] Bug fix (non-breaking change, fixes #4766 )
 

### Purpose
 

Fixing bug in Remove-DbaOrphanUser to appropriately determine which orphans users should be deleted
### Approach

The fix here syncs the logic with Repair-DbaOrphanUser, and will no longer try to match $User variable if it is empty, but utilize it when present

### Commands to test

I created 2 orphaned users in my Sandbox database to test/verify.  First, check that both are going to be dropped when explicitly named, then that only one is dropped when piped in.  Finally confirm all will be found/dropped if no piped in user information is provided
```
'domain\AdamWenger', 'domain\SecondOrphan' | Remove-DbaOrphanUser -SqlInstance $devInstance -Database Sandbox -Verbose -WhatIf # should drop both
'domain\AdamWenger' | Remove-DbaOrphanUser -SqlInstance $devInstance -Database Sandbox -Verbose -WhatIf # Should only drop the listed user
Remove-DbaOrphanUser -SqlInstance $devInstance -Database Sandbox -Verbose -WhatIf # Should drop all orphans
```


potatoqualitee:
excellent, thanks much Adam!

PR #4770, opened 2018-12-05
VSCode settings.json updates

potatoqualitee:
hello I would like to add these as the defaults.  inviting gentle feedback ;)


potatoqualitee:
YAY two approvals! 😻 thank you both

PR #4776, opened 2018-12-05
Fixed Add-DbaAgReplica

potatoqualitee:
interesting -- adding it after the fact is a lot different from adding it during the creation of the AG. Anyway, this fixes that. It just looked like it joined but it actually didnt.


potatoqualitee:
unrelated test failure

PR #4778, opened 2018-12-06
Update Test-DbaDbOwner so it will check all databases

### LINKED ISSUE #4753: Test-DBADBOwner fails to return database ownership for non-accessible databases
Test-DBADBOwner fails to return database ownership where IsAccessible property -eq $false. example would be a mirror database that is in the restoring state. 

Previously, it took a long time to respond when getting ownership for non-accessible database, however this may be resolved now.

Get-DBADatabase returns appropriate database ownership on these non-accessible databases just fine.

joeywas:
Update Test-DbaDbOwner so it will check all dbs, not just the ones where IsAccessible = $True

Set default parm set to NoPipe so if nothing is piped in, it will prompt for SqlInstance parameter.

Added contextual example of pipage ennoblement

## Type of Change

 - [x] Bug fix (non-breaking change, fixes #4753)

### Purpose
Test-DbaDbOwner should not filter out "non-accessible" databases. Previously, SMO was very slow to enumerate properties on non-accessible databases, but this issue has been resolved. If user wants to filter to only Accessible databases, Get-DbaDatabase -OnlyAccessible can be piped to Test-DbaDbOwner

### Approach
Removed IsAccessible checks from Test-DbaDbOwner.

### Commands to test
The examples! I added one!

### Learning
I learned how to handle objects from the pipe (Thanks @potatoqualitee!) as well as use conditional parameters with parameter sets

### DIFF
--- a/functions/Test-DbaDbOwner.ps1
+++ b/functions/Test-DbaDbOwner.ps1
@@ -57,10 +57,14 @@ function Test-DbaDbOwner {
 
         Returns all databases where the owner does not match 'DOMAIN\account'.
 
+    .EXAMPLE
+        PS C:\> Get-DbaDatabase -SqlInstance localhost -OnlyAccessible | Test-DbaDbOwner
+
+        Gets only accessible databases and checks where the owner does not match 'sa'.
     #>
-    [CmdletBinding()]
+    [CmdletBinding(DefaultParameterSetName = "NoPipe")]
     param (
-        [parameter(Mandatory)]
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Thanks so much for the PR! The example is fab and the isaccessible is a good change but we avoid parametersets in dbatools because they cause weird errors if the person uses them wrong. In this case, what you'd do is `if (-not $InputObject -and -not $Sqlinstance) { Stop-Function -Message "you must specify blablah ....return}` 👍

Issue #4786, opened 2018-12-07
Add SchemaName to Copy-DbaDbTableData output


Adam-Wenger:
---
Title: Add SchemaName to Copy-DbaDbTableData output :bomb:
About: It would be nice to explicitly see the schema name on the ouput for Copy-DbaDbTableData, especially for folks who have the same table name in multiple schemas.

---

### Summary of new feature

* Add SchemName to the ouput pscustomobject output

### Proposed technical details (if applicable)

* n/a

### Latest version of dbatools as of writing

* 0.9.710

Adam-Wenger:
Closing - Fixed via #4787 and released in 0.9.712

PR #4787, opened 2018-12-07
Update Copy-DbaDbTableData output & bugfix

Adam-Wenger:
Adding (Source and Destination) Schema to the output; helpful in cases where you have >1 schema with the same named table.
Also fixed -AutoCreateTable -> It was not properly replacing the schema name for the new table if it differed from the source schema.  It was retaining the original source schema due to bug in regex

## Type of Change

 - [X ] Bug fix (non-breaking change, fixes <didn't log bug, found during feature 4786>)
 - [X ] New feature (non-breaking change, adds functionality #4786 )
 

### Purpose
 

Fixes AutoCreateTable when the Source schema differs from the desired destination schema.  the current code just used the source due to typo in the regex.

Second change is to add SourceSchema and DestinationSchema to the pscustomobject output of the function so that you can clearly see what schema the tables are in

### Approach

Added the values to the output.

### Commands to test

```
$copyArgs = @{

   SqlInstance = $devInstance;
   Database = "Sandbox";
   Table = 'ref.YourTable';
   Destination = $devInstance;
   DestinationDatabase = "Sandbox";
   DestinationTable = 'new.NewName';
   AutoCreateTable = $true;
   KeepIdentity = $true;
   KeepNulls = $true;
   Truncate = $true;
   Verbose = $true;
}

Copy-DbaDbTableData @copyArgs;
```
### Screenshots

image

### DIFF
--- a/functions/Copy-DbaDbTableData.ps1
+++ b/functions/Copy-DbaDbTableData.ps1
@@ -200,7 +200,7 @@ function Copy-DbaDbTableData {
                 }
             }
         }'
-        
+
         Add-Type -ReferencedAssemblies System.Data.dll -TypeDefinition $sourcecode -ErrorAction Stop
         if (-not $script:core) {
             try {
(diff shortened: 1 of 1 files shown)


potatoqualitee:
excellent, thank you adam!

Issue #4783, opened 2018-12-06
Get-DbaDbFile should accept pipeline from Get-DbaDatabase


joeywas:
---
name: Get-DbaDbFile should accept pipeline from Get-DbaDatabase :bomb:
about: Get-DbaDbFile should accept pipeline from Get-DbaDatabase

---

### Summary of new feature

* We want pipeline goodness for all! Part of that is making sure Get-DbaDbFile can accept pipeline input from Get-DbaDatabase

### Proposed technical details (if applicable)

* Properties that Get-DbaDbFile would snag from the pipeline would be sqlinstance and database

### Latest version of dbatools as of writing

* 0.9.710

Adam-Wenger:
Closing - #4788 merged this into 0.9.715

PR #4788, opened 2018-12-08
Get-DbaDbFile extended pipeline support #4783

Adam-Wenger:
Adding $SmoDatabase pipeline parameter to support pipelining in an SMO Database object array.  If SmoDatabase is provided, it takes precidence over SqlInstance and Database; they're overwritten with the contents of SmoDatabase.  Added Region block around the queries so I could fold them in VSCode.

## Type of Change

### Purpose
 

Adding pipeline support to Get-DbaDbFile so that it accepts SMO Database objects (request from #4783 was to support piping from Get-DbaDatabase

### Approach

I Added a ValueFromPipeline parameter $SmoDatabase.  If this new parameter is passes, the function essentially ignores $SqlInstance and $Database parameters, and will overwrite their values with what is in the $SmoDatabase object.

**I'm not sure this is the correct approach for supporting multiple pipeline object types**

Originally I tried to maintain support for additional items coming from SqlInstance and Database, but it resulted in a lot of duplicated results being returned as the objects in $Database just kept growing, and running the process{} per pipe item item against multiple $Database objects instead of just the one on the pipe, so that is why I ultimately opted to go with the either/or approach for parameters instead of aggregating them.

I've also updated the ParameterSets for the parameters to try to be more explicit that the new SmoDatabase parameter is incompatible with SqlInstance & Database

### Commands to test

```
$db = Get-DbaDatabase -SqlInstance $devInstance, $qaInstance -Database "Audit", "Staging", "Warehouse"
$db | Get-DbaDbFile | Select-Object SqlInstance, LogicalName, TypeDescription, Database | Format-Table -AutoSize
Get-DbaDbFile -SmoDatabase $db | Select-Object SqlInstance, LogicalName, TypeDescription, Database | Format-Table -AutoSize
$devInstance | Get-DbaDbFile -Database "Warehouse" | Select-Object SqlInstance, LogicalName, TypeDescription, Database | Format-Table -AutoSize
```
### Screenshots

image

### DIFF
--- a/functions/Get-DbaDbFile.ps1
+++ b/functions/Get-DbaDbFile.ps1
@@ -66,23 +66,44 @@ function Get-DbaDbFile {
         [DbaInstanceParameter[]]$SqlInstance,
         [PSCredential]$SqlCredential,
         [Alias("Databases")]
-        [parameter(ParameterSetName = "Default")]
-        [parameter(ParameterSetName = "Pipe")]
         [object[]]$Database,
         [object[]]$ExcludeDatabase,
-        [parameter(ParameterSetName = "SmoDatabase", Mandatory, ValueFromPipeline)]
+        [parameter(ValueFromPipeline)]
         [Microsoft.SqlServer.Management.Smo.Database[]]$SmoDatabase,
         [object[]]$InputObject,
         [Alias('Silent')]
         [switch]$EnableException
     )
 
+    begin {
+
+        # Used to ensure parameter validation checks only run once.
+        $initialParameterCheck = $true;
+    }
+
     process {
+
+        # The check only needs to be run once, but cannot be done in begin{} blcok since it doens't know if we'll receive pipeline input
+        if($initialParameterCheck) {
+
+            if ($SqlInstance -and $SmoDatabase) {
+
+                Stop-Function -Message "You cannot specify both SqlInstance and SmoDatabase." -Continue -EnableException $EnableException;
+            }
+
+            if ($SmoDatabase -and $Database) {
+
+                Write-Message -Level Verbose -Message "Parameter Database is not being used since SmoDatabase was provided as well, using that instead.";
+            }
+
+            # initial check completd, set variable to $false so we don't check each iteration
+            $initialParameterCheck = $false;
+        }
 
         if ($SmoDatabase) {
 
             # Select-Object Unique is here so that when $SmoDatabae is used as a parameter (not piped) it doesn't loop thorugh the same instance
-            # more than once if you have more than one database for a given instance
+            # more than once if you have more than one database for a given instance, or if you have more than one instance with the same database
             $Database = $SmoDatabase.Name | Select-Object -Unique;
             $SqlInstance = $SmoDatabase.SqlInstance | Select-Object -Unique;
         }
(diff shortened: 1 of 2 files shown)


Adam-Wenger:
I saw Potatoqualitee‘s note in another PR that we shouldn’t use ParameterSets. I’ll swap it out for the if() logic in the morning.

Adam-Wenger:
Pipelines replaced with If logic now, ready for review/comments.

functions/Get-DbaDbFile.ps1
```diff
@@ -50,6 +53,12 @@ function Get-DbaDbFile {
 
         Will return an object containing all file groups and their contained files for the Impromptu and Trading databases on the sql2016 SQL Server instance
 
+    .EXAMPLE
+        PS C:\> $db = Get-DbaDatabase -SqlInstance sql2016 -Database Impromptu, Trading;
```
potatoqualitee: Semicolons are not encouraged in PowerShell as they aren't required and aesthetically pleasing.

functions/Get-DbaDbFile.ps1
```diff
@@ -19,6 +19,9 @@ function Get-DbaDbFile {
     .PARAMETER ExcludeDatabase
         The database(s) to exclude - this list is auto-populated from the server
 
+    .PARAMETER SmoDatabase
+        A SMO Database object that represents the database to reference
```
potatoqualitee: This is the intent of InputObject: InputObject should be the SMODatabase.

`[Microsoft.SqlServer.Management.Smo.Database[]]$InputObject,` You can do that, and then InputObject +=.

functions/Get-DbaDbFile.ps1
```diff
@@ -59,20 +68,54 @@ function Get-DbaDbFile {
...
+        [parameter(ValueFromPipeline)]
+        [Microsoft.SqlServer.Management.Smo.Database[]]$SmoDatabase,
         [object[]]$InputObject,
         [Alias('Silent')]
         [switch]$EnableException
     )
 
+    begin {
+
+        # Used to ensure parameter validation checks only run once.
```
potatoqualitee: The issue with introducing `$initialParameterCheck = $true` is that it would be the first in the toolset and thus the behavior would be inconsistent. Please let it fail repeatedly.

functions/Get-DbaDbFile.ps1
```diff
@@ -59,20 +68,54 @@ function Get-DbaDbFile {
...
+        if ($initialParameterCheck) {
+
+            if ($SqlInstance -and $SmoDatabase) {
+
+                Stop-Function -Message "You cannot specify both SqlInstance and SmoDatabase." -Continue -EnableException $EnableException;
+            }
+
+            if ($SmoDatabase -and $Database) {
+
+                Write-Message -Level Verbose -Message "Parameter Database is not being used since SmoDatabase was provided as well, using that instead.";
```
potatoqualitee: Both can be used with no issue. If you do `InputObject += Get-DbaDatabase...` then it wouldn't matter which one they pick.

potatoqualitee:
I've gone ahead and made the changes. take a look if you're interested 👍

Adam-Wenger:
Thanks Chrissy; I’ll take a look at these.

PR #4805, opened 2018-12-12
Get-DbaDbSpace Piping Support

potatoqualitee:


potatoqualitee:
i didnt want to support it so the functionality has been removed hence not using it cuz test-dep doesnt stop it

wsmelton:
If you want to stop supporting `IncludeSystemDbs`  then that pattern needs to be updated across all the other commands that use it. We have `IncludeSystemDatabases`, `IncludeSystemObjects`, `IncludeFilter` used in multiple commands. I'm fine with using the pattern that is being updated in this PR with all of those.

potatoqualitee:
mm this one just didnt seem to fit, i htink user/system belongs in get-dbadatabase which ones were you thinking?

PR #4814, opened 2018-12-13
Update Start-DbaMigration.ps1

### LINKED ISSUE #4809: UseLastBackup should imply BackupRestore
Experienced behavior: invoking Start-DbaMigration with -UseLastBackup but no -BackupRestore threw an error that I needed to specify -BackupRestore or -DetachAttach.

Expected behavior: invoking Start-DbaMigration with -UseLastBackup implies -BackupRestore (since it doesn't make sense with -DetachAttach).

1DontEx1st:
* -UseLastBackup also sets -BackupRestore
* raises error when both -DetachAttach  and (-BackupRestore or -UseLastBackup) are set
* copied "missing" logic checks from Begin into Process

## Type of Change
 - [X] Bug fix (non-breaking change, fixes #4809 )

### Purpose
Add some additional logic checks, and make -UseLastBackup imply -BackupRestore

Also - I may have misunderstood the need for doubling-up on logic checks (as per the "what-if" note), but I added all of the pertinent checks from Begin into Process.

### DIFF
--- a/functions/Start-DbaMigration.ps1
+++ b/functions/Start-DbaMigration.ps1
@@ -217,16 +217,16 @@ function Start-DbaMigration {
         Test-DbaDeprecation -DeprecatedOn 1.0.0 -Parameter NetworkShare -CustomMessage "Using the parameter NetworkShare is deprecated. This parameter will be removed in version 1.0.0 or before. Use SharedPath instead."
         
         if ($Exclude -notcontains "Databases") {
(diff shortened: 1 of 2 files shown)


potatoqualitee:
This looks so good! Thank you 🙇 Can you add any pics of it working? That'll speed things up. If not, I'll likely have some time to test tomorrow.

PR #4827, opened 2018-12-15
Save-DbaDiagnosticQueryScript - add output enhancements, fix bad link for 2019

potatoqualitee:
image

/cc @andrekamman


potatoqualitee:
one pass which fixes something in dev, so merging

PR #4842, opened 2018-12-18
fixing most recent fork as default and making sure it fires when gbh …

### LINKED ISSUE #4841: Get-DbaBackupHistory fork check not firing automatically, and sub optimal default choice

## Expected Behavior

Whenever `Get-DbaBackupHistory` is run against an instance with multiple recovery fork's history it should warn that this is the case

Should pick up the most recent recovery fork as that's what most people are expecting it to do

## Actual Behavior

Doesn't always fire

returns the last by lsn

## Notes

People should really clean up their backup history every now and then ;)

Stuart-Moore:
…is called on its own

## Type of Change

 - [x] Bug fix (non-breaking change, fixes #4841 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Ensures that the recovery fork check is called every time GBH is called on it's own (ie not by another command)

Sets default recoveryfork to the latest as that seems to be what people are expecting

### Approach
Added a check for being called in a scriptblock which is the the other outcome from Get-PsCallstack 

Set $RecoveryForkID to the latest recoveryforkid

### DIFF
--- a/functions/Get-DbaBackupHistory.ps1
+++ b/functions/Get-DbaBackupHistory.ps1
@@ -260,7 +260,7 @@ function Get-DbaBackupHistory {
                     if ($RecoveryFork) {
                         $recoveryForkSqlFilter = "AND backupset.last_recovery_fork_guid ='$RecoveryFork'"
                     }
-                    if ($null -eq (Get-PsCallStack)[1].Command) {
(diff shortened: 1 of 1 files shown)


potatoqualitee:
This is great, thank you for all of your work on your fabulous commands 🙇

PR #4848, opened 2018-12-19
Fixed random creation of values based on faker

sanderstad:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
When using the specific faker assigned types like Name it would create a random string instead of a name. 

### Approach
Added another section that creates the values

### Commands to test
Invoke-DbaDbDataMasking

### Screenshots
With the firstname changed as random string
image

With the firstname changed as type firstname from bogus
image


potatoqualitee:
Thanks! passes testes locally

PR #4849, opened 2018-12-19
updated buildref

niphlod:


potatoqualitee:
sorry niph, got a conflict :(

niphlod:
should be the same from https://github.com/sqlcollaborative/sqlcollaborative.github.io/blob/master/assets/dbatools-buildref-index.json

PR #4857, opened 2018-12-20
More enhancements for Update-DbaInstance

nvarscar:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Adding new features to the existing set of commands:
* Users can now specify which authentication to use; CredSSP is still a default for remote connections with credentials, but is no longer a requirement per se. Will be useful when patch repository is accessible locally or does not require authentication.
* If a target machine (not a localhost) has a restart pending, the function will now restart the machine prior to beginning the installation 
* Passing credentials through the insecure connection now asks for confirmation (but only once)
* Invoke-Program now has a -Fallback option for remote execution with Credentials that enables execution through...
* New internal function Invoke-CommandWithFallback - a wrapper for Invoke-Command2 that does a retry in case an initial connection (CredSSP in particular) has failed due to transport issues.
* Minor fixes and improvements across the board

### Screenshots
Running in a non-elavated session without means to configure CredSSP:

image

Only needs one confirmation of using insecure transit:

image

Restarting computer prior to starting the installation due to pending restart:

image

### DIFF
--- a/functions/Update-DbaInstance.ps1
+++ b/functions/Update-DbaInstance.ps1
@@ -167,6 +167,8 @@ function Update-DbaInstance {
         [switch]$EnableException
     )
     begin {
+        $notifiedCredentials = $false
+        $notifiedUnsecure = $false
         #Validating parameters
         if ($PSCmdlet.ParameterSetName -eq 'Version') {
             foreach ($v in $Version) {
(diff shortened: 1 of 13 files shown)


potatoqualitee:
daaaaaaaaaaaamn! this is awesome! 🏅 🏅 🏅

Issue #3451, opened 2018-03-30
Read only intent for Invoke-DbaSqlQuery


michalporeba:
## Feature Request

Ability to execute read only queries against secondary replicas

### Problem to solve

Neither `Invoke-SqlCmd` , `Invoke-SqlCmd2` nor `Invoke-DbaSqlQuery` seems to support the read only application intent at the moment. I would like to be able to choose which replica to run the query against without the need to explicitly check. 

### Additional information

I was thinking of adding `-ReadOnly` parameter to `Invoke-DbaSqlQuery`  rather than piping in filtered results from `Get-DbaAgReplica` ( or something similar which could get rather complex very soon).  `Connect-SqlInstance` and `Invoke-SqlCmd2` would have to be modified to use read only application intent. 

`Connect-DbaInstance` already implements `-ApplicationIntent` parameter.

Would it be worth adding MultiSubnetFailover to the connection string too?

niphlod:
guess is a fix for connect-dbainstance more than any other function. 
We can't really support every param combination in each function, so if you need "detailed" parameters, your scenario will be

$conn = Connect-DbaInstance .....
Invoke-DbaSqlQuery -SqlInstance $conn

wsmelton:
@niphlod is correct on this one. If you need a custom connection property the implemented path is to create your connection first and then pass that connection to the command. The `Connect-DbaInstance` supports both `ApplicationIntent` and `MultiSubnetFailover`. 

`Invoke-SqlCmd2` utilizes `System.Data.SqlClient` to make connections. So if you need a custom connection property for that command you would have to pass in that full connection string via `SQLConnection` parameter.

In reference to `ApplicationIntent` for `Invoke-SqlCmd`...[we tried to get it added](https://trello.com/c/deWheByZ/123-applicationintent-support) but did not get enough votes or attention. I'm not sure they still utilize Trello so if you want to submit user feedback through the new site you can prompt getting this parameter added there.

michalporeba:
I will give it a go but I'm not sure if this will be that simple. I was hopging to be able to run a query against all of the databases in my cluster which hosts multiple AGs. The connection string has to be decided on a database level if we cannot use the correct listener with ReadOnly intent. I'll do some testing when I've got a moment and will post an update here. 

As always, thanks for your input, @niphlod, @wsmelton. Much appreciated.

wsmelton:
There is a command in the module that will output a connection string, can't recall name. That may help in the process.

[...]

Issue #4695, opened 2018-11-27
Removing unit test: params.Count - $defaultParamCount | Should Be $paramCount


potatoqualitee:
This second test does not seem to bring benefit. It's duplicate information that we just gotta recalc after any additions. Seems like comparing our known params with actual params is perfect.
```
$paramCount = X
$defaultParamCount = 11
        [object[]]$params = (Get-ChildItem function:\Get-DbaLogin).Parameters.Keys
        $knownParameters = 'SqlInstance', 'SqlCredential', 'x','y','z'
        It "Should contain our specific parameters" {
            ( (Compare-Object -ReferenceObject $knownParameters -DifferenceObject $params -IncludeEqual | Where-Object SideIndicator -eq "==").Count ) | Should Be $paramCount
        }
        It "Should only contain $paramCount parameters" {
            $params.Count - $defaultParamCount | Should Be $paramCount
        }
````

 I'd like to remove 
```
$paramCount = X
$defaultParamCount = 11

 It "Should only contain $paramCount parameters" {
            $params.Count - $defaultParamCount | Should Be $paramCount
        }
```

ijeb:
I updated Copy-DbaServerAudit.Tests.ps1 accordingly

potatoqualitee:
fixed! thank you @jpomfret !

PR #4893, opened 2019-01-02
ReadOnly option for Invoke-DbaQuery

michalporeba:
## Type of Change

New Feature #3451
 

### Purpose
 
To be able to easily run queries read only queries on secondary nodes in AG

### Approach

When `Invoke-DbaQuery` is invoked with `-ReadOnly` parameter the ReadOnly application intent is set which allows the read only queries to succeed even if the specific database on the current instance is a read only replica and not primary. 

### Commands to test

`Invoke-DbaQuery ... -ReadOnly`


wsmelton:
Use of custom properties for the connection string should be done via `Connect-DbaInstance … -ApplicationIntent ReadOnly` and then passing in that server object to the command.

The internal function should not be touched to support this feature.

wsmelton:
Hit enter a bit to early....I see no reason to modify an internal function to support this feature/ability as this function (1) is used with everything and (2) we don't need to update an internal function in order to support one feature that only one command will ever use.

wsmelton:
Which this was all discussed in that issue as well.

wsmelton:
Yeah, also testing (at least on PS Core) this change has broken any command I use with a SQL Login. Not all of our test utilize a SQL Login to test the credential so assuming the checks are covering every scenario would be false.

image

Extensive testing would have to be done if I'm overruled and this feature is approved.

michalporeba:
Thanks Shawn for your comments. I see one test is failing and I will look more into it this afternoon. I see the internal function is light on the tests, but if it is used so much it will be tested with every other test of the function that is using it. Still, it would be good to improve testing I think. 

As for the discussion about the feature itself it would be good to go to the original issue #3451.

functions/Invoke-DbaQuery.ps1
```diff
@@ -118,7 +121,8 @@ function Invoke-DbaQuery {
         [switch]$MessagesToOutput,
         [parameter(ValueFromPipeline)]
         [Microsoft.SqlServer.Management.Smo.Database[]]$InputObject,
-        [switch]$EnableException
+        [switch]$EnableException,
+        [switch]$ReadOnly
```
wsmelton: `EnableException` should always be the last parameter, just for tab purposes.

tests/Invoke-DbaQuery.Tests.ps1
```diff
@@ -4,10 +4,10 @@ Write-Host -Object "Running $PSCommandPath" -ForegroundColor Cyan
 
 Describe "$CommandName Unit Tests" -Tag 'UnitTests' {
     Context "Validate parameters" {
-        $paramCount = 13
+        $paramCount = 14
```
wsmelton: Check #4695 and you might go ahead and adjust this test to match the new pattern for unit testing.

functions/Invoke-DbaQuery.ps1
michalporeba: done, I'll be pushing the change with the few more updates shortly

tests/Invoke-DbaQuery.Tests.ps1
michalporeba: I'll look into it.

michalporeba: Done.

michalporeba:
I'm really out of shape. So many commits for so little. Sorry :)

potatoqualitee:
i do the same @michalporeba 🤗

potatoqualitee:
squash and merge, now all 1 💯

wsmelton:
Little commits are best and user more commonly than you may think. It makes it easier to revert or remove those commits around breaking changes if they are discovered during testing.

Issue #4889, opened 2019-01-02
v0.9.732 Invoke-SQLCMD2 Credentials parameter missing


skattie22:
Hi
It appears that Invoke-SQLCmd2 has been replaced by Invoke-DBAQuery
But seeing at Invoke-SQLCmd2 still exists, why oh why did the -Credential parameter get removed?
and on calling Invoke-SQLCmd2 it does not state that it is deprecated...?

wsmelton:
`Invoke-DbaQuery` does not replace `Invoke-Sqlcmd2`. `Invoke-Sqlcmd2` is a function added to dbatools module out of convenience, and is maintained by a different individual. (We simply host it under [sqlcollaborative\invoke-sqlcmd2](https://github.com/sqlcollaborative/Invoke-SqlCmd2))

The `-Credential` parameter was not removed.

It is not being deprecated.

skattie22:
Thanks Shawn, I don't understand why the 2 commands below are so different?
0.9.501
`Invoke-Sqlcmd2 [-ServerInstance] <String[]> [[-Database] <String>] [-Query] <String> 
    [[-Credential] <PSCredential>] [[-Encrypt]] [[-QueryTimeout] <Int32>] 
    [[-ConnectionTimeout] <Int32>] [[-As] <String>] [[-SqlParameters] <IDictionary>] 
    [[-AppendServerInstance]] [[-ParseGO]] `[<CommonParameters>]`

0.9.732
`Invoke-Sqlcmd2 [-ServerInstance] <String> [-Database] <String> [[-Query] <String>] 
    [[-CommandType] {Text | StoredProcedure | TableDirect}] [[-CommandParameters] <Array>] 
    [[-Username] <String>] [[-Password] <String>] [[-QueryTimeout] <Int32>] 
    [[-ConnectionTimeout] <Int32>] [[-InputFile] <String>] [[-As] <String>] 
    [[-CallingFunction] <String>] [[-ApplicationName] <String>] [[-ReturnStream]] 
    [<CommonParameters>]`

potatoqualitee:
thx shawn you are right, Credential is aliased so would still work. skattie the commands are different because invoke-dbaquery adheres to our toolset conventions and has more goodies baked in 👍

skattie22:
Thanks Chrissy, but I am comparing 2 versions of Invoke-Sqlcmd2, forget invoke-dbaquery, the 2 commands above are very different...or am I missing something?

[...]

PR #4904, opened 2019-01-04
delete invoke-sqlcmd2.ps1

potatoqualitee:
fixes confusion in #4889

what i dont get is that..there's no export of this command so im unsure how users were accesssing it unless it was directly via psm1


wsmelton:
That particular issue I think was due to older version of the module.

PR #4915, opened 2019-01-08
Update Test-DbaLastBackup parameter help

shaneis:
Changed -IncludeCopyOnly help from "will not" to "will"

## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [x] Documentation
 - [ ] Build system
 

### Purpose
 
Current documentation says that if you do not want to include copy-only backups then you should use the switch `-IncludeCopyOnly`.

But the parameter has Include in its name, and a quick check of the code confirms that if you include this switch then it does include copy-only backups.

This change should hopefully make the documentation technically correct which is the best kind of correct ;)

### Approach

Makes the documentation clearer to remove the confusion.

### Commands to test

`Get-Help -Name Test-DbaLastBackup -Parameter IncludeCopyOnly`

### DIFF
--- a/functions/Test-DbaLastBackup.ps1
+++ b/functions/Test-DbaLastBackup.ps1
@@ -61,7 +61,7 @@ function Test-DbaLastBackup {
         The name of the SQL Server credential on the destination instance that holds the key to the azure storage account.
 
     .PARAMETER IncludeCopyOnly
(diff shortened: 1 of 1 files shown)


potatoqualitee:
fabu thank you

PR #4958, opened 2019-01-15
Figure out the new name before doing the checks.

cfradenburg:
Otherwise we may check to see if "sourceInstance default profile" exists on the destination then trying to create "destinationInstance default profile" which already exists.

## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
When trying to update a profile on a server named "serverb default profile" from a server with a profile named "servera default profile" it tried to create a duplicate mail profile and failed.

### Approach
Figure out the new mail profile name before doing the existence checks

### Commands to test
Copy-DbaDbMail -Source servera -Destination serverb -Verbose -force

### DIFF
--- a/functions/Copy-DbaDbMail.ps1
+++ b/functions/Copy-DbaDbMail.ps1
@@ -188,6 +188,8 @@ function Copy-DbaDbMail {
             foreach ($profile in $sourceProfiles) {
 
                 $profileName = $profile.name
+                $newProfileName = $profileName -replace [Regex]::Escape($source), $destinstance
+                Write-Message -Message "Updating profile name from $profileName to $newProfileName." -Level Verbose
                 $copyMailProfileStatus = [pscustomobject]@{
                     SourceServer      = $sourceServer.Name
                     DestinationServer = $destServer.Name
@@ -198,30 +200,30 @@ function Copy-DbaDbMail {
                     DateTime          = [Sqlcollaborative.Dbatools.Utility.DbaDateTime](Get-Date)
                 }
 
-                if ($profiles.count -gt 0 -and $profiles -notcontains $profileName) {
+                if ($profiles.count -gt 0 -and $profiles -notcontains $newProfileName) {
                     continue
                 }
 
-                if ($destProfiles.name -contains $profileName) {
+                if ($destProfiles.name -contains $newProfileName) {
                     if ($force -eq $false) {
-                        If ($pscmdlet.ShouldProcess($destinstance, "Profile $profileName exists at destination. Use -Force to drop and migrate.")) {
+                        If ($pscmdlet.ShouldProcess($destinstance, "Profile $newProfileName exists at destination. Use -Force to drop and migrate.")) {
                             $copyMailProfileStatus.Status = "Skipped"
                             $copyMailProfileStatus.Notes = "Already exists"
                             $copyMailProfileStatus | Select-DefaultView -Property DateTime, SourceServer, DestinationServer, Name, Type, Status, Notes -TypeName MigrationObject
-                            Write-Message -Message "Profile $profileName exists at destination. Use -Force to drop and migrate." -Level Verbose
+                            Write-Message -Message "Profile $newProfileName exists at destination. Use -Force to drop and migrate." -Level Verbose
                         }
                         continue
                     }
 
-                    If ($pscmdlet.ShouldProcess($destinstance, "Dropping profile $profileName and recreating.")) {
+                    If ($pscmdlet.ShouldProcess($destinstance, "Dropping profile $newProfileName and recreating.")) {
                         try {
-                            Write-Message -Message "Dropping profile $profileName." -Level Verbose
-                            $destServer.Mail.Profiles[$profileName].Drop()
+                            Write-Message -Message "Dropping profile $newProfileName." -Level Verbose
+                            $destServer.Mail.Profiles[$newProfileName].Drop()
                             $destServer.Mail.Profiles.Refresh()
                         } catch {
                             $copyMailProfileStatus.Status = "Failed"
                             $copyMailProfileStatus | Select-DefaultView -Property DateTime, SourceServer, DestinationServer, Name, Type, Status, Notes -TypeName MigrationObject
-                            Stop-Function -Message "Issue dropping profile." -Target $profileName -Category InvalidOperation -InnerErrorRecord $_ -Continue
+                            Stop-Function -Message "Issue dropping profile." -Target $newProfileName -Category InvalidOperation -InnerErrorRecord $_ -Continue
                         }
                     }
                 }
--- a/functions/Copy-DbaDbMail.ps1
+++ b/functions/Copy-DbaDbMail.ps1
@@ -124,6 +124,8 @@ function Copy-DbaDbMail {
             Write-Message -Message "Migrating accounts." -Level Verbose
             foreach ($account in $sourceAccounts) {
                 $accountName = $account.name
+                $newAccountName = $accountName -replace [Regex]::Escape($source), $destinstance
+                Write-Message -Message "Updating account name from $accountName to $newAccountName." -Level Verbose
                 $copyMailAccountStatus = [pscustomobject]@{
                     SourceServer      = $sourceServer.Name
                     DestinationServer = $destServer.Name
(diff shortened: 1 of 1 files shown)


wsmelton:
Can you provide code that produces the issue this PR is fixing since we do not have an assigned issue?

potatoqualitee:
Thank you both!  The code appears to be included `Copy-DbaDbMail -Source sql2016 -Destination sql2017 -Verbose -force`. 

I have tested the changes locally and I think it's just one step away from finished. 

image

cfradenburg:
Chrissy,

From what you're sending it looks like sql2016 has two mail profiles, "sql2016 dba" and "The DBA Team" and sql2017 has a profile named "sql2016 dba" because that's the value that's throwing the duplicate error.  The output shows that the profile name got replaced so it wouldn't have dropped the "sql2016 dba" profile but it shouldn't have tried to create it either.

While trying to get the $newProfileName variable to have the right value I had to muck with the regex that's used to replace.  Here's what I used:
$newProfileName = $profileName -replace [Regex]::Escape($source), $destinstance

And here's what the SQL replace uses:
$sql = $sql -replace [Regex]::Escape("'$source'"), "'$destinstance'"

It hadn't occurred to me at the time but I'm betting the replace only happens in the SQL if the instance name is surrounded by quotes.  Which is backed up by the fact that the instance name isn't replaced in our description:

DEBUG: 236 | [08:19:26][Copy-DbaDbMailProfile] EXEC msdb.dbo.sysmail_add_profile_sp @profile_name=N'serverb',
                @description=N'Default mail profile  for servera instance.'

I removed the quotes and double quotes and it works on my machine but before I commit that change, do you know why those were there in the first place?  I want to make sure that doesn't break anything else.  They've been in there since the file was created and I'm not well enough versed in git to quickly find the Copy-DbDatabaseMail file to see if there was anything in there that would say why.

potatoqualitee:
hey @cfradenburg ! thanks for taking a deeper look. I should have clarified that the duplicate happened after I ran the script for a second time. So the first run went well, but the second one failed. 

The reason the quotes are around the name in other commands is because we want to be as accurate as possible. For instance, let's say someone names their server `sql` - if no quotes are there, it's possible that the destination script will replace `sqlmail` so it'd be `sql2017mail`. So if it's possible to include the quotes _in the final script_ somehow, that'd be ideal.

wsmelton:
By "Code to reproduce" I mean T-SQL that will create a profile and such on each source and destination so we can duplicate the setup that this fix is supposed to be addressing.

As well, if we have that we can update the test to also validate it continues to support that scenario going forward.

cfradenburg:
So all I need to do to be completely thorough is come up with a Regex that finds cases where it's between an opening single quote and a closing single quote but not when it's between a closing single quote and an opening single quote.  Should be easy.  Thankfully I have several chickens to sacrifice.

On a more serious note, what do you think about just replacing "@profile_name=N'<expr>'" and "@account_name=N'<expr>'"?  It'll leave the description with the old name if it's there which may not be desired but would leave a bread crumb trail so it also may be.  I could also add that one in if you think that's preferred.  I'll make sure I test cases where the profile and account have just the server names and where it has text either side.

I'll also put another commit in to do the same check I already committed for account name.  We don't put our server names in there so I hadn't caught it.  The same regex also gets applied to the mail configuration but I don't think I need to do anything with that.  I don't think I need to do anything for mail server either because the account name should already be updated when it pulls the mail server script but I'll double check that.

- [ ] Create code to create profiles and accounts with: Just servername, servername with leading text, servername with trailing text (There aren't DBATools commands for these yet so it'll be quicker to create one in SQL)
- [ ] Add account name dupe check
- [ ] Check to see if the regex to update mail server script needs to be revised
- [ ] Update SQL script regex once we decide on a method.

It'll probably take a couple days for me to be able to get through all this but I'll work on it as I have time.

potatoqualitee:
`what do you think about just replacing "@profile_name=N''" and "@account_name=N''"?` totally! that is a great idea, actually. thanks so much 🙇

cfradenburg:
Code to set up test profiles:

DECLARE @beforeText NVARCHAR(MAX),
	@textBefore NVARCHAR(MAX),
	@serverOnly NVARCHAR(MAX)

SET @beforeText = CAST(SERVERPROPERTY('ServerName') AS NVARCHAR(MAX)) + ' before text'
SET @textBefore = 'text before ' +  CAST(SERVERPROPERTY('ServerName') AS NVARCHAR(MAX))
SET @serverOnly = CAST(SERVERPROPERTY('ServerName') AS NVARCHAR(MAX))

EXEC msdb.dbo.sysmail_add_account_sp @account_name = @beforeText,
                                     @email_address = N'test@test.org',
                                     @description = @beforeText,
									 @mailserver_name = 'email.test.org',
									 @mailserver_type = 'SMTP';

EXEC msdb.dbo.sysmail_add_account_sp @account_name = @textBefore,
                                     @email_address = N'test@test.org',
                                     @description = @textBefore,
									 @mailserver_name = 'email.test.org',
									 @mailserver_type = 'SMTP';

EXEC msdb.dbo.sysmail_add_account_sp @account_name = @serverOnly,
                                     @email_address = N'test@test.org',
                                     @description = @serverOnly,
									 @mailserver_name = 'email.test.org',
									 @mailserver_type = 'SMTP';


EXEC msdb.dbo.sysmail_add_profile_sp @profile_name = @beforeText,
                                     @description = @beforeText;
EXEC msdb.dbo.sysmail_add_profileaccount_sp @profile_name = @beforeText,
                                            @account_name = @beforeText,
                                            @sequence_number = 1;
EXEC msdb.dbo.sysmail_add_principalprofile_sp @principal_name = N'guest',
                                              @profile_name = @beforeText,
                                              @is_default = 1;

EXEC msdb.dbo.sysmail_add_profile_sp @profile_name = @textBefore,
                                     @description = @textBefore;
EXEC msdb.dbo.sysmail_add_profileaccount_sp @profile_name = @textBefore,
                                            @account_name = @textBefore,
                                            @sequence_number = 1;
EXEC msdb.dbo.sysmail_add_principalprofile_sp @principal_name = N'guest',
                                              @profile_name = @textBefore,
                                              @is_default = 1;

EXEC msdb.dbo.sysmail_add_profile_sp @profile_name = @serverOnly,
                                     @description = @serverOnly;
EXEC msdb.dbo.sysmail_add_profileaccount_sp @profile_name = @serverOnly,
                                            @account_name = @serverOnly,
                                            @sequence_number = 1;
EXEC msdb.dbo.sysmail_add_principalprofile_sp @principal_name = N'guest',
                                              @profile_name = @serverOnly,
                                              @is_default = 1;

potatoqualitee:
fabulous! 💞 thank you! if my tests pass, is it ready to merge?

cfradenburg:
Not yet, that's just pulling the check I made for profile into account.  I still need to update the regex to update the SQL that gets run for accounts, profiles, and mail servers so it'll use the correct name if the server name in the profile has text either before or after it.

potatoqualitee:
pawesome thank you 🐾

cfradenburg:
It should be all set now.  If you notice any issues let me know.

potatoqualitee:
thanks so much. im unsure whats going on with my setup, first i did this 

image


and was like, okay maybe my leftover stuff caused issues, so i went delete `sql2016 dba` on the destination. then i tried again

image

then I asked myself if maybe the source server has issues but it doesnt seem like it

image

does that provide enough information? i'd like to figure this part out before trying with other tests. This was the specific failure

image

cfradenburg:
That's unexpected.  I ran the script above and it worked fine on a named instance with text after.  I even threw a single quote into a name to make sure that would work.  What also has me even more puzzled is if I take the code that should be generated from the source instance and plug it into a [RegEx tester](http://regexstorm.net/tester) it still comes out the way that it should.

`EXEC msdb.dbo.sysmail_add_profile_sp @profile_name=N'sql2016 dba',
                @description=N''
EXEC msdb.dbo.sysmail_add_principalprofile_sp @principal_name=N'guest', @profile_name=N'sql2016 dba', @is_default=0`

Can you confirm those commands are what's generated for your system?

potatoqualitee:
Sorry for the delay, I'm a bit behind. Will get to this when I get a moment.

wsmelton:
Tested on local docker containers and all good...
image

Checking on full domain environment shortly.

wsmelton:
From Azure lab domain, everything looks good:

image

wsmelton:
However, since this is changing the server name should we have that also reflected in the output object?

image

potatoqualitee:
oh thank you 🙇 

congrats on your azure lab!! if i were running the commands, i'd like to see the original name. perhaps the updated name can be thrown into the notes in a new PR 👍 

Thank you both so very much 💯

potatoqualitee:
your code is now in the PowerShell Gallery and in master 🎉

cfradenburg:
> However, since this is changing the server name should we have that also reflected in the output object?
> 
> image

My thought was to display the profile that the copy was requested from but I'm up for updating the code to do this.  I updated the delete profile/account messages to reference the new name so it's clear what's getting dropped but my thought was that the source is the name that should be reported so if for some reason fails it's clear what profile it's coming from.  Another option would be to report it as "text before wsql2016-01, renamed to text before wsql2016-02" but that gets rather wordy so I don't think that's a prime choice.

Updating the name in the output object would allow it to be more easily used in further commands which is a strong argument for doing that.  If you want me to submit a new PR for that let me know.

PR #4961, opened 2019-01-15
In progress: Improve and consolidate Unit Tests

jpomfret:
## Type of Change

 - [X] Bug fix (non-breaking change, address #4695 )
 - [X] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Current (single) unit test does not fail if you add a parameter to the function. Rewrote this test to fail if parameters are removed or added to functions.

### Screenshots

Missing Param
image
Extra Param
image
All happy
image

### DIFF
--- a/tests/Get-DbaTraceFlag.Tests.ps1
+++ b/tests/Get-DbaTraceFlag.Tests.ps1
@@ -4,15 +4,11 @@ Write-Host -Object "Running $PSCommandPath" -ForegroundColor Cyan
 
 Describe "$CommandName Unit Tests" -Tag 'UnitTests' {
     Context "Validate parameters" {
-        $paramCount = 4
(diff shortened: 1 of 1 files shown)


wsmelton:
Just noting the images provided seem to all show failures...is that intended?

jpomfret:
My bad, got the wrong image on the last one. I did find this doesn't work for all functions. If we're using should process it fails currently. Will have. fix for you today.

potatoqualitee:
Thanks so much for your help with this 🙇 does it make any sense to just remove whatif and confirm from all collections?

jpomfret:
Yeah I think so, that is my new improved unit test idea. I'll get it to you when I get to the office (in a couple of hours).

PR #4983, opened 2019-01-21
adjusting example to use splatting

wsmelton:
Just updating example to use splatting instead of the line breaks, and removed comment lines as those are covered by the example description.


potatoqualitee:
beautiful, thank you!

PR #4987, opened 2019-01-21
Updating editor settings for VS Code

wsmelton:
Added additional settings that I found turned off by default in VS Code now. Also added comments so folks can see a bit more info on why we set what.


potatoqualitee:
excellent, thank you!

PR #4993, opened 2019-01-23
Fix lower case / upper case

### LINKED ISSUE #4985: Get-DbaUserPermission not working on SQL Servers with case-sensitive collation
Hi,

We were trying to use the Get-DbaUserPermission but we never received any result.
After investigation, we noticed that the views were not created correctly because they contain wrong column names because we're using an instance with case sensitive collation.
We corrected the STIG.sql file so that all the views and functions were created correctly and also corrected the SQL statements in the ps1 script as well but we never got it working…

Could it be possible to help us out and verify what should be changed to make this script working on an instance with case sensitive collation?

Thanks
Best REgards
Pieter

ijeb:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #4985 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Code should run on a case sensitive server/database.

Also fix this line
$sql = $sql.Replace($db.Name, "<TARGETDB>")
This is wrong because of tempdb on the first line will be replaced.

When tested, results of a database after tempdb (alphabetically) were wrong.

### Commands to test
Get-DbaUserPermission -SqlInstance sql2 | Out-GridView

### DIFF
--- a/bin/stig.sql
+++ b/bin/stig.sql
@@ -483,7 +483,7 @@ USE tempdb;
                         P.[State],
                         P.[Grantor],
                         P.[Grantor Type],
-                        P.[source view]
+                        P.[Source View]
                     FROM
                         STIG.database_permissions P
                         INNER JOIN Targets T ON T.[Principal] = P.[Grantee]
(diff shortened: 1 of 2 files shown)


potatoqualitee:
🙇 🙇 🙇  thank you so much, you da MVP

PR #5011, opened 2019-01-25
bug report - proposed changes

potatoqualitee:
Proposed changes to let the community know that we will focus our resources and time on backup|restore. This should ensure that these two very essential commands work in as many environments as possible, at the expense of ensuring the Copy-DbaDatabase wrapper works in specific instances.

This should help reduce the number of times @pollusb and others encounter issues with SQL 2000.


potatoqualitee:
My wording in the bug report is awkward. If anyone has any suggestions, let me know. Otherwise, I'll merge after 24 hours.

potatoqualitee:
both it is 👍

PR #5053, opened 2019-02-05
Fixes for Update-DbaInstance: parallelism, exceptions

### LINKED ISSUE #4988: Get-SqlInstanceComponent - Internal function needs throw used instead of Stop-Function
Following lines need to be updated to either utilize `throw` used instead of calling `Stop-Function`, or simply don't wrap the call to invoke in a try/catch so it throws the desired error properly to the parent caller.

https://github.com/sqlcollaborative/dbatools/blob/5a3e806868725cc7729f006ab915614a8d278550/internal/functions/Get-SqlInstanceComponent.ps1#L300-L304

In conjunction with this the reference call made in the parent command should be wrapped in try/catch block with `Stop-Function` so it shows proper messaging system text instead of red errors:

https://github.com/sqlcollaborative/dbatools/blob/5a3e806868725cc7729f006ab915614a8d278550/functions/Update-DbaInstance.ps1#L289-L293

nvarscar:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #4988 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Fixing issue raised by @wsmelton as well as some annoying parallelism fix that can't be tested in Appveyor (adding `-ImportFunctions` to the `Invoke-Parallel` fixes inability to use internal commands inside the script block).

### Approach
no EnableException in the internals
try-catches everywhere

### Commands to test
Update-DbaInstance -ComputerName srv1, srv2

### DIFF
--- a/functions/Update-DbaInstance.ps1
+++ b/functions/Update-DbaInstance.ps1
@@ -157,7 +157,7 @@ function Update-DbaInstance {
         [string[]]$KB,
         [Alias("Instance")]
         [string]$InstanceName,
-        [string[]]$Path,
+        [string[]]$Path = (Get-DbatoolsConfigValue -Name 'Path.SQLServerUpdates'),
         [switch]$Restart,
         [switch]$Continue,
         [ValidateNotNull()]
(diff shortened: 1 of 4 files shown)


potatoqualitee:
@wsmelton - can you test?

potatoqualitee:
thanks so much @nvarscar  🙇

PR #5103, opened 2019-02-19
New-DbaDbMailAccount

potatoqualitee:
new command with tests - another command is still needed tho, one that makes mail servers

`Microsoft.SqlServer.Management.Smo.Mail.MailServer`


wsmelton:
Reviewing this now to fix some of the issues on the test failures.

One note is that you don't create mail servers with DB Mail. You just create profiles and accounts. The mail server ican be setup within a given profile or each account can have a mail server. Your ideal setup would be your profile to have the SMTP configured so it is reusable and does not require you to adjust all the accounts.

wsmelton:
Debugging this one a bit more because I'm getting the same failure on my container that Appveyor shows.

wsmelton:
Ok....took me to long to figure that one out. Missing the parameter name for `Stop-Function`.

Fixed and the command fails properly when the mail server does not exists.

potatoqualitee:
heyy thanks @wsmelton ! fixed a merge issue and tests have been running better, let's see if this works

potatoqualitee:
finalllllly i can merge this!

potatoqualitee:
the problem was indeed my code which was suitable to work only when the server name was typed w/o aliases. so localhost\sql2016 failed cuz it was appveyor-win\slq2016. yay pester tests

PR #5129, opened 2019-02-27
fix path issues in psm1

potatoqualitee:
fixes this

image


potatoqualitee:
Will merge on tests passing. I've done some manual tests and it works on linux as well.

wsmelton:
What environment setup is required for this error to occur? It does not exist on any machine I've tried it on for my environments.

potatoqualitee:
i have no idea what has to be in place, just that I kept running into it 👎

PR #5132, opened 2019-02-28
Get-DbaServerRoleMember - Correcting call to child call for Get-DbaLogin

wsmelton:
Correct passing connection into child call to `Get-DbaLogin`


potatoqualitee:
LGTM 👍

PR #5151, opened 2019-03-05
Fix bug in Get-DbaDbFile NextGrowthEventSize

### LINKED ISSUE #5147: Get-DbaDbFile - incorrectly accounting for pages when calculating NextGrowthEventSize
The NextGrowthEventSize value returned by Get-DbaDbFile appears to be calculated incorrectly due to a double accounting for pages on non-percent growth types and by not accounting for pages on percent growth types.

## Steps to Reproduce

```sql
/**** create a temporary database for example ****/
CREATE DATABASE [SizeTestDB]
 ON  PRIMARY 
( NAME = N'SizeTestDB', FILENAME = N'C:\TempDatabase\SizeTestDB.mdf' , SIZE = 256MB , MAXSIZE = UNLIMITED, FILEGROWTH = 65536KB )
 LOG ON 
( NAME = N'SizeTestDB_log', FILENAME = N'C:\TempDatabase\SizeTestDB_log.ldf' , SIZE = 256MB , MAXSIZE = 2048GB , FILEGROWTH = 10%)


```

```powershell
Get-DbaDbFile -sqlinstance localhost -database SizeTestDB
```

## Expected Behavior

For the .mdf, the NextGrowthEventSize should be 64 MB and for the .ldf it should be 25.6 MB

## Actual Behavior

For non-Percent growth types it looks like it may be trying to calculate pages twice. The underlying T-SQL gets the [growth] value and multiplies it by 8 initially
```sql
case mf.is_percent_growth when 1 then df.growth else df.Growth*8 end as Growth
```
...but then that result is multiplied by 8 again in the powershell
```powershell
 $nextgrowtheventadd = [dbasize]($result.Growth * 8 * 1024)
```
So the .mdf file's 64 MB (8192 pages) growth value ends up calculated to a NextGrowthEventSize of 512MB instead of the expected 64 MB. I think one of the page conversions should be removed.

For Percent growth types the underlying T-SQL gets the [growth] value and it is multiplied in Powershell against the size value which doesn't appear to be converted from pages. So the 256 MB (32768 pages) .ldf file with a 10% growth type is showing a NextGrowthEventSize of 3.20 MB instead of 25.6 MB. 

```powershell
$nextgrowtheventadd = [dbasize]($result.size * ($result.Growth * 0.01) * 1024)
```
Should be something like
```powershell
$nextgrowtheventadd = [dbasize]($result.size * 8 * ($result.Growth * 0.01) * 1024)
```

image

## Environmental data

 - PowerShell: 5.1.17134.590
 - - SQL Server: Microsoft SQL Server 2016 (SP1-CU7-GDR) (KB4057119) - 13.0.4466.4 (X64)

IsItGreyOrGray:
## Type of Change
 - [x] Bug fix (non-breaking change, fixes #5147)
 
### Purpose
The $nextgrowtheventadd variable used to return NextGrowthEventSize in Get-DbaDbFile is incorrectly accounting for pages. 

For Percent growth types, pages are not currently accounted for and for non-Percent growth types pages are accounted for twice - once in the initial T-SQL call and again in the calculation for $nextgrowtheventadd.

### Approach
Line 215 - Update $nextgrowtheventadd calculation for Percent growth types to account for file size being returned in pages.

Line 217 - Update $nextgrowtheventadd calculation for non-Percent growth types to not account for file size being returned in pages as the "* 8" calculation is already performed in the T-SQL calls when growth is not percent-based. 

### Screenshots
Current version 

- mdf file is set to 64 MB growth but is showing 512 MB for the NextGrowthEventSize
- ldf file is set to 10% growth but is showing 3.20 MB for the NextGrowthEventSize

bug

Fixed version 

- 256 MB .mdf file is set to 64 MB growth and is showing 64 MB for the NextGrowthEventSize
- 256 MB .ldf file is set to 10% growth and is showing 25.60 MB for the NextGrowthEventSize

fixed

### DIFF
--- a/functions/Get-DbaDbFile.ps1
+++ b/functions/Get-DbaDbFile.ps1
@@ -212,9 +212,9 @@ ON fd.Drive = LEFT(df.physical_name, 1);
                     }
                 }
                 if ($result.GrowthType -eq "Percent") {
-                    $nextgrowtheventadd = [dbasize]($result.size * ($result.Growth * 0.01) * 1024)
+                    $nextgrowtheventadd = [dbasize]($result.size * 8 * ($result.Growth * 0.01) * 1024)
                 } else {
-                    $nextgrowtheventadd = [dbasize]($result.Growth * 8 * 1024)
+                    $nextgrowtheventadd = [dbasize]($result.Growth * 1024)
                 }
                 if (($nextgrowtheventadd.Byte -gt ($MaxSize.Byte - $size.Byte)) -and $maxsize -gt 0) {
                     [dbasize]$nextgrowtheventadd = 0


potatoqualitee:
awesome, thank you! will release to gallery shortly 🎉

PR #5156, opened 2019-03-05
CommandExecute.sql only runs if $Solution = "All"

dgmyrs:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Fixes issue with Install-DbaMaintenanceSolution where it does not run CommandExecute if the Solution is not "All", even though it is required for all of the maintenance solutions.

### Approach
parens around the match statement so -not is applied correctly.


potatoqualitee:
thanks so much! 💯 💯

PR #5179, opened 2019-03-10
Consolidate connects 🤞

potatoqualitee:
i was tired of changing code in two places so I added the necessary params to -Dba and rerouted -Sql. And then added our specific application name.


potatoqualitee:
i'm still considering it but so far, the ClientName and distinction that makes it manageable in the event we need to.

potatoqualitee:
both of them do, to disable exceptions in connect-dbainstance, you have to `-DisableException`. by default, it acts like Connect-SqlInstance

potatoqualitee:
im gonna close this branch, it's failing for no good reason. works locally. i wonder if appveyor is up to no good.

PR #5182, opened 2019-03-10
Firewall commands

### LINKED ISSUE #3519: New firewall rules command
I've completed the primary logic for the firewall rules command I've been working on since last December 2017. I've created this issue to start a tracked discussion on it.

I need a little help getting this prepared for my "first" PR (I use "first" in quotes because this code isn't spoon fed to me this time ;) ref: https://github.com/sqlcollaborative/dbatools/pull/2695 and I'm not sure I count that).

The code imports the `NetSecurity` module and adds new firewall rules to the `Windows Firewall with Advanced Security` application, as well as setting the Display Group for all of these newly created rules to "SQL" Group (by default) for easier management later.

Given the above, what should this new firewall command be named (not sure if it should be Add, Set, New)? I'm thinking options could be like:
Add-DbaFirewallRules
New-DbaFirewallRules
Set-DbaFirewallRules
something else?

potatoqualitee:
putting this back on the map. i considered making this backwards compat to older versions of windows, but it's far too much work. Using the built-in fw ps commands is 💯 

Win8/Win2012+

Fixes #3519


potatoqualitee:
closing for now but will keep the `fw` branch

PR #5183, opened 2019-03-11
Fix debug and Path is null issues

### LINKED ISSUE #4965: dbatools module import errors
[importModuleOutputs.txt](https://github.com/sqlcollaborative/dbatools/files/2764584/importModuleOutputs.txt)

## Steps to Reproduce

$dbatools_serialimport=$true
import-module  dbatools -Verbose

## Expected Behavior

dbatools module should load

## Actual Behavior

Using the latest build of dbatools and importing the module i get

WARNING: The module import has hit a timeout while waiting for some background tasks to finish.
This may result in some commands not working as intended.
This should not happen under reasonable circumstances, please file an issue at:
https://github.com/sqlcollaborative/dbatools/issues
Or contact us directly in the #dbatools channel of the SQL Server Community Slack Channel:
https://dbatools.io/slack/
Timeout waiting for temporary runspaces reached! The Module import will complete, but some things may not work as intended

If i then unload the module and add $dbatools_serialimport=$true i get 
```
Exception calling "Invoke" with "3" argument(s): "Cannot bind argument to parameter 'Path' because it is null."
At line:203 char:5
+     $scriptBlock.Invoke($script:PSModuleRoot, "$script:DllRoot\smo",  ...
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + FullyQualifiedErrorId : CmdletInvocationException
```

The module did load once during repo but i cant explain why but it now consistently fails
```
## Environmental data

PSVersion                      5.1.14409.1018                                                                                                                                                                                                                    
BuildVersion                   10.0.14409.1018                                                                                                                                                                                                                   
```

potatoqualitee:
note that debug now works by default, so imports for developers will be a bit slower.

i also fixed the path not found issue - fixes #4965


bin/smo/smo-deps.txt
```diff
@@ -1,3 +1,4 @@
+cd C:\github\dbatools\bin\smo
```
wsmelton: Why is this a hard coded path?

PR #5188, opened 2019-03-11
disable DLL test that keeps failing

potatoqualitee:
i dont know how to fix but it produces a big ol bunch of red as seen here: https://ci.appveyor.com/project/sqlcollaborative/dbatools/builds/22991623/job/78r25q6da5m78nc9

so i disabled it until it's fixed by someone with knowledge of the dll tests


potatoqualitee:
it fails because it's missing not because of legit fails

PR #5203, opened 2019-03-14
Improve Handling for TLS

sheldonhull:
## Type of Change
 - [x] Bug fix

### Purpose
AWS instances with restricted internet settings fail to process webrequest without basic parsing. A better practice for rest based requests is to use `Invoke-RestMethod`. 

### Approach
- Timeout parameter added for allowing override of AWS metadata
- New internal function cloned from `Invoke-TlsWebRequest` that is using `Invoke-RestMethod`
- Proper handling of failure to connect with `try{} catch{}` should ensure timeout exception handled properly and warned, bypassing attempt to connect to AWS metadata further.
- *new* added 3 properties related to CPU metrics to improve the results for Get-DbaComputerSystem in reviewing basics of OS performance specs

### Commands to test
1. Get-DbaComputerSystem
2. Invoke-TlsRestMethod *new internal helper*

# Note
1. Without AWS infrastructure can't verify this with pester test, and didn't see any `Invoke-TLS*` tests built. The new error handling worked, and confirmed on my own AWS instance that results pulled correctly, but you'll want to verify as well. 
2. Not sure of any other location to add detail to.

### DIFF
--- a/functions/Get-DbaComputerSystem.ps1
+++ b/functions/Get-DbaComputerSystem.ps1
@@ -16,9 +16,6 @@ function Get-DbaComputerSystem {
     .PARAMETER IncludeAws
         If computer is hosted in AWS Infrastructure as a Service (IaaS), additional information will be included.
 
-    .PARAMETER TimeoutSec
-        Default: 5 sec
-        Timeout in seconds for AWS related metaquery.
     .PARAMETER EnableException
         By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
         This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
@@ -58,7 +55,6 @@ function Get-DbaComputerSystem {
         [DbaInstanceParameter[]]$ComputerName = $env:COMPUTERNAME,
         [PSCredential]$Credential,
         [switch]$IncludeAws,
-        [int]$TimeoutSec = 5,
         [switch][Alias('Silent')]
         $EnableException
     )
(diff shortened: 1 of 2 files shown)


functions/Get-DbaComputerSystem.ps1
```diff
@@ -55,6 +58,7 @@ function Get-DbaComputerSystem {
         [DbaInstanceParameter[]]$ComputerName = $env:COMPUTERNAME,
         [PSCredential]$Credential,
         [switch]$IncludeAws,
+        [int]$TimeoutSec = 5,
```
wsmelton: Why does this need to be a parameter? As well, we had it set to 15 previously, if the parameter is to stay in place it needs to be set to 15 seconds.

wsmelton: Talking with Chrissy this needs to be removed and implemented as a module configuration. 

However from my point of view I see no reason for it. The site you are hitting is a local site in AWS servers only. I believe we can simply set it to a value and leave it.

wsmelton:
When the timeout parameter is removed I'm good for this to be merged.

functions/Get-DbaComputerSystem.ps1
sheldonhull: Sounds good to me. thanks!

wsmelton:
I have not had time but Azure finally added an REST API on Azure VMs for similar information if you happen to want to tackle that one 😁

wsmelton:
Failure was unrelated on Appveyor

niphlod:
BTW: this sparked a bug in https://github.com/sqlcollaborative/dbatools/issues/5336 . Remote instances do not have our own local commands, so invoke-command2 on anything other than localhost fails miserably.

I'll see what can be done in terms of "remoting" the two needed functions

PR #5204, opened 2019-03-14
Library Update to better support Smo Server Reuse and Azure

potatoqualitee:
Certain Properties in the SMO recycler section of `[dbainstanceparameter]` caused the connection to initiate and reset. So I modified this to the bare minimum.

I can't figure out how to keep the Port, so we'll see where that takes us. So far, the tests are passing.

Initially, I added a NetPort in Connect-DbaInstance but removed it because I couldn't get it to work.


potatoqualitee:
test failure due to unrelated, one pass, also got fred's approval

PR #5217, opened 2019-03-17
better error message

### LINKED ISSUE #5091: Include the SqlInstance in the failure to connect message for Get-DbaDatabase
---
name: Include sqlinstance name in Get-DbaDatabase failure to connect message
about: 

---

### Summary of new feature

When using a CMS group with Get-DbaDatabase, it would be helpful to include the SqlInstance name in the failure message.

Pretty sure it wouldn't be any more than changing the message from:

to

Stop-Function -Message "Failure to connect to $instance" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue

0.9.757

potatoqualitee:
fixes #5091

image

### DIFF
--- a/functions/Add-DbaAgReplica.ps1
+++ b/functions/Add-DbaAgReplica.ps1
@@ -119,7 +119,7 @@ function Add-DbaAgReplica {
             try {
                 $server = Connect-SqlInstance -SqlInstance $instance -SqlCredential $SqlCredential -MinimumVersion 11
             } catch {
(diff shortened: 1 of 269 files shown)


wsmelton:
We need to fix the error message system and not have to keep doing find and replace for this stuff. Commented in the issue.

wsmelton:
I recall the target used to be included in the write message output... That was why we changed the connection message to what it is now.

potatoqualitee:
not everything has a target instance, i'm not so good at adding that one. but if someone would like to do a mass replace for that and then add to write-message, i'd accept a PR that overwrites these changes 👍

PR #5223, opened 2019-03-17
Copy-DbaDbTableData - should fix #5080

### LINKED ISSUE #5080: copy-dbadbtabledata causses faillure on max pool size
I am filling a table in a few servers with fake data. The data comes from WideWorldImporters and is send to one other db on the same server and two other SQL2014 and SQL2012 servers. It is doing this in a loop.
When it has ran for some time, the output changes to:
Failure |timeout expired. The timeout period elapsed prior to obtaining a connection from the pool.

For me this looks like the connections are not closed at the end of this script. 
Could you investigate?
we are using 0.9.714

potatoqualitee:


potatoqualitee:
integration tests pass 👍

PR #5237, opened 2019-03-18
Agent Commands - remove blue output

potatoqualitee:
totally over these test failures, just gonna merge. it's Output -> Verbose


potatoqualitee:
(appveyor is being hit by a DDoS which is why we're getting failures)

PR #5238, opened 2019-03-18
Backupfix for core

### LINKED ISSUE #5070: [System.IO.Path]::GetExtension() does not accept objects as input in PS6

## Steps to Reproduce

```powershell
PS> Restore-DbaDatabase -SqlInstance "localhost" -Path $Path

Cannot find an overload for "GetExtension" and the argument count: "1".
At C:\Program Files\PowerShell\Modules\dbatools\0.9.755\allcommands.ps1:73252 char:17
+             if ([System.IO.Path]::GetExtension($p).Length -eq 0) {
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : NotSpecified: (:) [], MethodException
+ FullyQualifiedErrorId : MethodCountCouldNotFindBest
```

## Explanation

Forgive me if I get any of the terminology wrong.

The root error is in how Get-DbaBackupInformation passes file names to Read-DbaBackupHeader, so I would expect this error to pop up in other functions as well.

Restore-DbaDatabase calls Get-DbaBackupInformation, which collects metadata of all backup files in a folder by putting all the file names in an array. These file names should be passed as strings. However, sometimes they are passed as objects. Then the array is passed to Read-DbaBackupHeader which tries to determine whether each element in the array has an extension or not: [System.IO.Path]::GetExtension($file). Apparently in PS5.1 when the array would consist of objects these were implicitly converted to strings by this function. This was basically always wrong to begin with, because the result would be "@{file.extension}" instead of "file.extension". In PS6 GetExtension apparently does not accept objects at all anymore.

This behavior can be observed by running the following code in PS5.1 and PS6 and comparing the output:
```powershell
$file = New-Object PSObject -Property @{FullName="C:\Windows\system.ini"}
[System.IO.Path]::GetExtension($file)
[System.IO.Path]::GetExtension($file -as [String])
[System.IO.Path]::GetExtension($file.FullName)
```

## Proposed Solution

In lines 227, 231, 258, 259 and 260 of Get-DbaBackupInformation change the code from
```powershell
$Files += [[innercode]]
```
to
```powershell
$Files += ( [[innercode]] ).FileName
```
This might not be complete, but it would solve it for my scenario.

## Environmental data

```powershell
PSVersion                      6.1.2
GitCommitId                    6.1.2
OS                             Microsoft Windows 6.3.9600
```

potatoqualitee:
will not merge this one until we can get some passes in appveyor but i wanted to put it on the agenda. fixes #5070

### DIFF
--- a/functions/Get-DbaBackupInformation.ps1
+++ b/functions/Get-DbaBackupInformation.ps1
@@ -224,11 +224,11 @@ function Get-DbaBackupInformation {
                             Write-Message -Level Verbose -Message "Skipping Log Backups as requested"
                         } else {
                             Write-Message -Level Verbose -Message "OLA - Getting folder contents"
-                            $Files += (Get-XpDirTreeRestoreFile -Path $f -SqlInstance $server).FileName
+                            $Files += (Get-XpDirTreeRestoreFile -Path $f -SqlInstance $server).FullName
                         }
                     } else {
                         Write-Message -Message "Testing a folder $f" -Level Verbose
-                        $Files += $Check = (Get-XpDirTreeRestoreFile -Path $f -SqlInstance $server).FileName
+                        $Files += $Check = (Get-XpDirTreeRestoreFile -Path $f -SqlInstance $server).FullName
                         if ($null -eq $check) {
                             Write-Message -Message "Nothing returned from $f" -Level Verbose
                         }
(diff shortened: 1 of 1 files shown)


potatoqualitee:
didnt work

wsteen:
Your first commit is what did the trick for me, but it was a while back. Maybe I'm overlooking something. I'm curious as to why it didn't work for you.

PR #5258, opened 2019-03-20
better connection string support

potatoqualitee:
fixes connectionstring bug too


potatoqualitee:
thank you @nvarscar  🙇

PR #5296, opened 2019-03-23
Manage Instance doesn't script WithReplace

potatoqualitee:
it may be a bug we need to report

image

this is a workaround

### DIFF
--- a/functions/Invoke-DbaAdvancedRestore.ps1
+++ b/functions/Invoke-DbaAdvancedRestore.ps1
@@ -258,6 +258,10 @@ function Invoke-DbaAdvancedRestore {
                                 $script = $script.TrimEnd() + ' WITH KEEP_CDC'
                             }
                             if ($true -ne $OutputScriptOnly) {
+                                if (($Restore.Script($server) -match '\ URL') -and $WithReplace) {
+                                    # restore does not generate WITH REPLACE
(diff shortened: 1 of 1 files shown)


potatoqualitee:
This seems a bit slower than WithReplace btw.

potatoqualitee:
Confirmed unsupported by Jovan.

WITH REPLACE option is not supported in RESTORE statement in Managed Instance - see https://docs.microsoft.com/en-us/azure/sql-database/sql-database-managed-instance-transact-sql-information#restore-statement . Here is a list of differences: https://docs.microsoft.com/en-us/azure/sql-database/sql-database-managed-instance-transact-sql-information#restore-statement

Stuart-Moore:
There's a load of other bits it does't support. Would suggest we do a check at the top of the restore and then ignore the bits that don't exist (like file moves, etc).

potatoqualitee:
should i close this PR? If they use WithReplace should we recommend they use Remove then Restore, instead? ty 💯

Stuart-Moore:
Yes, I've a working version that hides the drop under withreplace so user doesn't notice.  

I'll push up a draft PR in the morning to track the changes as I make them.  

Trying to make it all as transparent as poss for the users  
--  
Stuart Moore

potatoqualitee:
thanks so much 🙇

PR #5298, opened 2019-03-24
Set-DbaStartupParameter - getting 1.0 ready

### LINKED ISSUE #5292: Set-DbaStartupParameter - Examples and bug with trace flag

This command overall just needs to be reworked/reviewed for 1.0 release.

## Steps to Reproduce

```powershell
Set-DbaStartupParameter -SqlInstance myserver -TraceFlags 1117,1118 -WhatIf
```

## Bug 1

`-TraceFlags` should be singular

## Bug 2

The examples show passing in TraceFlags parameter with `-SingleUser` which is not realistic to be done. Unless I'm missing something

## Bug 3

Issuing the above command spits out this warning:

> WARNING: [10:55:51][Set-DbaStartupParameter] MasterData value must be provided

So that parameter is not validated to be provided at all, nor do I think it is required to just change trace flags.

~~## Bug 4~~

~~I don't know where this comes from but I'm running this on my localhost (dev machine) that is just a workgroup, so this may be from that fact alone. The what if output:~~

~~> What if: Performing the operation "Set variable" on target "Name: __dbatools_interrupt_function_78Q9VPrM6999g6zo24Qn83m09XF56InEn4hFrA8Fwhu5xJrs6r Value: True"~~

wsmelton:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #5292 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Working through #5292 and overall getting ready for 1.0 (e.g. apply standard on formatting, try/catch around child calls to other dbatools commands, etc.)


wsmelton:
Ok, as long as test pass everything is good to go on this one...

functions/Set-DbaStartupParameter.ps1
```diff
@@ -207,43 +206,48 @@ function Set-DbaStartupParameter {
...
-        #Get Current parameters:
-        $currentstartup = Get-DbaStartupParameter -SqlInstance $SqlInstance -Credential $Credential
-        $originalparamstring = $currentstartup.ParameterString
+        # Get Current parameters (uses WMI) -- requires elevated session
+        try {
+            $currentStartup = Get-DbaStartupParameter -SqlInstance $SqlInstance -Credential $Credential -EnableException
+        } catch {
+            Stop-Function -Message "Unable to gather current startup parameters" -Target $SqlInstance -ErrorRecord $_
+            return
+        }
```
wsmelton: This is particularly important to do anytime we are calling other public functions in our commands. Because this was not including `EnableException` and wrapped in a try/catch it was falsely assuming SingleUser was true at current line 237 because the object didn't have any content to begin with.... so `f (!($currentStartup.SingleUser))` will return true when that property is true *AND* when `$currentStartup -eq $null` ...so the object itself had no content.

wsmelton:
Had to add few quick fixes for the copy startup proc command that I found missed.

potatoqualitee:
looks good, ty!

PR #5314, opened 2019-03-27
Read-DbaBackupHeader - Support for PS Core

### LINKED ISSUE #5070: [System.IO.Path]::GetExtension() does not accept objects as input in PS6

wsmelton:
At least it seems to work on my testing

## Type of Change

 - [x] Bug fix (non-breaking change, fixes #5070 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Details are in the related issue #5070 

image

### DIFF
--- a/functions/Read-DbaBackupHeader.ps1
+++ b/functions/Read-DbaBackupHeader.ps1
@@ -101,8 +101,9 @@ function Read-DbaBackupHeader {
     )
 
     begin {
-        foreach ($p in $path) {
-            if ([System.IO.Path]::GetExtension($p).Length -eq 0) {
+        foreach ($p in $Path) {
+            Write-Message -Level Verbose -Message "Checking: $p"
+            if ([System.IO.Path]::GetExtension($p.ToString()).Length -eq 0) {
                 Stop-Function -Message "Path ($p) should be a file, not a folder" -Category InvalidArgument
                 return
             }
--- a/functions/Read-DbaBackupHeader.ps1
+++ b/functions/Read-DbaBackupHeader.ps1
@@ -103,8 +103,8 @@ function Read-DbaBackupHeader {
     begin {
         foreach ($p in $Path) {
             Write-Message -Level Verbose -Message "Checking: $p"
-            if ([System.IO.Path]::GetExtension($p.ToString()).Length -eq 0) {
-                Stop-Function -Message "Path ($p) should be a file, not a folder" -Category InvalidArgument
+            if ([System.IO.Path]::GetExtension("$p").Length -eq 0) {
+                Stop-Function -Message "Path ("$p") should be a file, not a folder" -Category InvalidArgument
                 return
             }
         }


wsmelton:
@StartAutomating just `$p` is the problem it has running on PS Core. Therefore it is not sufficient.

wsmelton:
It will also be hard for `$p` to be null since we require the Path parameter to be provided.

StartAutomating:
"$p", not just $p.  

If I read your fix correctly, the problem was that the path wasn't coerced to a string automagically by PowerShell (which is weird, since pathinfos should cast cleanly to strings (this may be PSH's bug)). 

$p.ToString() will fix this.  "$p" will do the same, but will tend to run faster (because Powershell will have to interpret less).  It's a really minor drag either way (1.05x-1.5x faster over 10000 runs).

wsmelton:
Use of `"$p"` takes it longer (981ms vs 1.4 sec)...but does seem to correct the warning.

StartAutomating:
I'm just getting used to the codebase, but if just .ToString() didn't do it, then you're almost certainly experiencing something I liked to call a "null burp".

Basically, this is when you pass in a list that unexpectedly contains a null, which can throw off lots of little things within PowerShell.

Taking a look at the function in question, a few things jump out at me:
* [object[]] pipelining is generally not a great thing (anything can be an object, including a null)
* It seems like the function is designed to take a File path (I'm not sure if you're yet accounting for if that was a PowerShell path or not)
* If there aren't any other scenarios than piping in a file or a string that could be a file name, then there are cleaner ways to do this that should sidestep the bug
* If this bug originated from something that calls this, then I'd like to find the source of the null burp.

To be brief, if this is primarily taking a bunch of files as input, it's not doing it the "right" way, and this will make the script more error prone.

This type of fix is a band-aid overall.  Please let me know if you'd like something more surgical to be done here.

wsmelton:
Yeah, there is a good bit more to it than that...

StartAutomating:
I'm happy to help more directly if you can provide context and contact information (sorry, I'm new and while I believe we've met I don't believe I have your info).

wsmelton:
> [object[]] pipelining is generally not a great thing (anything can be an object, including a null)
This is required for TEPP so we can populate values where we have it configured. We simply standardized this for piping purposes. We are aware that things can be a null and in situations where we know 100% they are we account for it in the code.

> It seems like the function is designed to take a File path (I'm not sure if you're yet accounting for if that was a PowerShell path or not)

I don't follow what you mean by PowerShell path...it is whatever path the backups are expected to be read from. The path directory is passed from a parent command call, this command wants the file path itself but we add a check to make sure it is a file path and not just a directory.

> If there aren't any other scenarios than piping in a file or a string that could be a file name, then there are cleaner ways to do this that should sidestep the bug
There are many scenarios that use this command. The backup and restore commands are many pieces and are not modified lightly anymore these days because they are the most used commands we have.

> If this bug originated from something that calls this, then I'd like to find the source of the null burp.
The bug only shows up when it is being called from the parent command `Get-DbaBackupHistory` because we iterate over every file we find in the directory/path passed into that parent command, and each file is passed into this one to read the header. The reading is done asynchronously to speed up that process, in the event someone passed in a directory that has 4 days worth of 15-minute log backup files.

wsmelton:
@potatoqualitee test finally pass, the ones that fail are from some Google address being used... 🤷‍♂️ 

```
Describe : Resolve-DbaNetworkName Unit Tests
Context  : Testing basic name resolution
Name     : It should test 8.8.8.8 with Turbo = False
Result   : Failed
Message  : Expected strings to be the same, but they were different.
           Expected length: 19
           Actual length:   3
           Strings differ at index 0.
           Expected: 'google-public-dns-a'
           But was:  'DNS'
           -----------^
```

StartAutomating:
> This is required for TEPP so we can populate values where we have it configured. We simply standardized this for piping purposes. We are aware that things can be a null and in situations where we know 100% they are we account for it in the code.

TEPP?
 
> It seems like the function is designed to take a File path (I'm not sure if you're yet accounting for if that was a PowerShell path or not)

> I don't follow what you mean by PowerShell path...it is whatever path the backups are expected to be read from. The path directory is passed from a parent command call, this command wants the file path itself but we add a check to make sure it is a file path and not just a directory.

I meant to check two things:
* that only files are meant to be piped in (seems like yes )
* that said files could come from a relative path in PowerShell (..\foo\bar) vs an absolute path ( c:\foo\bar) 
 
> If there aren't any other scenarios than piping in a file or a string that could be a file name, then there are cleaner ways to do this that should sidestep the bug
> There are many scenarios that use this command. The backup and restore commands are many pieces and are not modified lightly anymore these days because they are the most used commands we have.

Do those scenarios pipe anything other than files or paths to files?  I understand the imperative to tread lightly, and am mainly trying to grow in understanding now before potenially readdressing some of the performance problems I see here at a later date.

> If this bug originated from something that calls this, then I'd like to find the source of the null burp.

> The bug only shows up when it is being called from the parent command Get-DbaBackupHistory because we iterate over every file we find in the directory/path passed into that parent command, and each file is passed into this one to read the header. The reading is done asynchronously to speed up that process, in the event someone passed in a directory that has 4 days worth of 15-minute log backup files.


Ok, it took a while, but I believe I see how the problem is occurring.  Apologies, I don't know how to quote other files.

The HistoryObject.Fullname property is populated like so:  $group.Group.Path

This will only work in PowerShell v3+ (fine enough), but it can really, really bite you and isn't ideal.  What this actually says is "find me all of the items in this group's Path property".  If this property is null, most later versions of PowerShell are smart enough to skip it.  However, as I'm sure you've encountered,[Dbnull]::Value -ne $null.  Give me some more context and I can recommend how this changes.

wsmelton:
We can discuss further in an issue if you would like to create one or in [dbatools-dev](https://sqlcommunity.slack.com/messages/C3EJ852JD/) channel of SQL Community Slack...there is no reason to go into detail on all this in this PR.

potatoqualitee:
Thank you @wsmelton - i updated that test and merged. and thank you @StartAutomating for the feedback! We have the channel `#dbatools-dev` in the SQL community Slack at dbatools.io/slack if you'd like to join 👍

PR #5321, opened 2019-03-29
Import-DbaCsv - auto-map Column Names

petervandivier:
## Type of Change

 - [X] Ran manual Pester test and has passed (`.\tests\manual.pester.ps1`)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 

```powershell
(Get-Help Import-DbaCsv).Parameters.Parameter | Where Name -EQ "ColumnMap"
````

The parameter help for `ColumnMap` states...

>By default, the bulk copy tries to automap columns. When it doesn't work as desired, this parameter will help. Check out the examples for more information.

...but does it though? Consider the following scenario:

```powershell
# (Get-Module dbatools).Version -eq '0.9.798'
$myDb = "YUNoBcpGud"
New-DbaDatabase -SqlInstance "." -Name $myDb -Verbose

$tblName = "foo"

$mkTbl = @"
create table dbo.$tblName (
     id int not null primary key identity
    ,foo varchar(100) 
    ,baz varchar(100) 
    ,bin varchar(100) 
    ,bar varchar(100) 
);
"@

Invoke-DbaQuery -ServerInstance "." -Database $myDb -Query $mkTbl -Verbose

$myTxt = @"
foo,bar,baz
aaa,111,1900-01-01
bbb,222,1900-02-02
ccc,333,1900-03-03
"@

$fileName = "foo.csv"
$myTxt | Set-Content $fileName

$importSpec = @{
    Path        = $fileName
    SqlInstance = "." 
    Database    = $myDb
    Schema      = "dbo"
    Table       = $tblName
}

Import-DbaCsv @importSpec -Verbose

Import-Csv $fileName
Invoke-DbaQuery -ServerInstance "." -Database $myDb -Query "select * from $tblName" | ft

<#
foo bar baz       
--- --- ---       
aaa 111 1900-01-01
bbb 222 1900-02-02
ccc 333 1900-03-03



id foo baz        bin bar
-- --- ---        --- ---
 1 111 1900-01-01        
 2 222 1900-02-02        
 3 333 1900-03-03        
#>
```

...some pro-level "mapping" there, huh? "`foo`" data has been lost. "`bar`" data have been transposed. [We can do better](https://youtu.be/SLILjDx0SO0).

### Approach

Using the cached `$firstline` object from earlier in the function, we build the `$ColumnMap` object for the user. 

At this time it is exact-match only. Fuzzy-matching, exclude-cols, and other validations can be conceived of, but I'm not certain at this time if those are more likely to be helpful additions or confusing noise. 

The current implementation is to require the user to supply the new `-AutoColumnMap` switch to access this, but it might be worth considering this as the default behavior instead? In the event that the current default is desirable, the switch name is discoverable to intellisense for sensibly similar params. 

### Commands to test

```powershell
$autoSpec = @{
    Path        = $fileName
    SqlInstance = "." 
    Database    = $myDb
    Schema      = "dbo"
    Table       = $tblName
    AutoColumnMap = $true
    Truncate    = $true
}
Import-DbaCsv @betterSpec -Verbose

Invoke-DbaQuery -ServerInstance "." -Database $myDb -Query "select * from $tblName" | ft
<#

id foo baz        bin bar
-- --- ---        --- ---
 1 aaa 1900-01-01     111
 2 bbb 1900-02-02     222
 3 ccc 1900-03-03     333
#>
```

Currently, `bcp` bails out if the supplied mapping is invalid (mapped col does not exist in source for example)

> VERBOSE: Performing the operation "Performing import from C:\Users\peter\Desktop\foo.csv" on target "NP:.".
> WARNING: [07:19:49][Import-DbaCsv] Failure | 'baz' field header not found.
> Parameter name: name

### DIFF
--- a/functions/Import-DbaCsv.ps1
+++ b/functions/Import-DbaCsv.ps1
@@ -519,7 +519,7 @@ function Import-DbaCsv {
                         $bulkCopy.NotifyAfter = $NotifyAfter
                         $bulkCopy.EnableStreaming = $true
 
-                        if ($AutoMapColumns) {
+                        if ($AutoColumnMap) {
                             if ($ColumnMap) {
                                 Write-Message -Level Verbose -Message "ColumnMap was supplied. Additional auto-mapping will not be attempted."
                             } else {
--- a/functions/Import-DbaCsv.ps1
+++ b/functions/Import-DbaCsv.ps1
@@ -242,6 +242,7 @@ function Import-DbaCsv {
         [switch]$KeepNulls,
         [string[]]$Column,
         [hashtable[]]$ColumnMap,
+        [switch]$AutoColumnMap,
         [switch]$AutoCreateTable,
         [switch]$NoProgress,
         [switch]$NoHeaderRow,
@@ -514,6 +515,18 @@ function Import-DbaCsv {
                         $bulkCopy.BatchSize = $BatchSize
                         $bulkCopy.NotifyAfter = $NotifyAfter
                         $bulkCopy.EnableStreaming = $true
+
+                        if ($AutoMapColumns) {
+                            if ($ColumnMap) {
+                                Write-Message -Level Verbose -Message "ColumnMap was supplied. Additional auto-mapping will not be attempted."
+                            } else {
+                                $ColumnMap = New-Object -TypeName "System.Collections.Hashtable"
+
+                                $firstline -split $Delimiter | ForEach-Object {
+                                    $ColumnMap.Add($PSItem,$PSItem)
+                                }
+                            }
+                        }
 
                         if ($ColumnMap) {
                             foreach ($columnname in $ColumnMap) {
--- a/functions/Import-DbaCsv.ps1
+++ b/functions/Import-DbaCsv.ps1
@@ -61,6 +61,9 @@ function Import-DbaCsv {
     .PARAMETER ColumnMap
         By default, the bulk copy tries to automap columns. When it doesn't work as desired, this parameter will help. Check out the examples for more information.
 
+    .PARAMETER AutoColumnMap
+        If this switch is enabled, bcp will attempt to map exact-match columns names from the source document to the target table.
+
     .PARAMETER AutoCreateTable
         Creates a table if it does not already exist. The table will be created with sub-optimal data types such as nvarchar(max)
 
--- a/functions/Import-DbaCsv.ps1
+++ b/functions/Import-DbaCsv.ps1
@@ -526,7 +526,7 @@ function Import-DbaCsv {
                                 $ColumnMap = New-Object -TypeName "System.Collections.Hashtable"
 
                                 $firstline -split $Delimiter | ForEach-Object {
-                                    $ColumnMap.Add($PSItem,$PSItem)
+                                    $ColumnMap.Add($PSItem, $PSItem)
                                 }
                             }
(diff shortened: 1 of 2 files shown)


petervandivier:
I'm planning to play around with more extended validations locally this morning, but by and large most of them require an invocation of `Get-DbaDbTable` to be accurate, so I my first instinct is to discuss them separately and merge them into my fork piecemeal to keep this thread cogent. I'm open to input though :smile: 

 * Attempt fuzzy-matching to target? i.e.: column_name -> ColumnName
 * Remove column names from the mapping that cannot be matched?
 * Warn & bail for non-insertable cols (computed, identity, etc) - but I'm not sure how this should interact with `-KeepNulls`, `-KeepIdentity`, etc

Require `-Confirm` switch for all of the above?

Also notable: I haven't yet tested what happens in the presence of an interaction w/ `-NoHeaderRow`

functions/Import-DbaCsv.ps1
```diff
@@ -515,6 +519,18 @@ function Import-DbaCsv {
                         $bulkCopy.NotifyAfter = $NotifyAfter
                         $bulkCopy.EnableStreaming = $true
 
+                        if ($AutoMapColumns) {
```
potatoqualitee: Thank you this is great! I considered something in the past, but didnt include it because my default was to do it by default and it doesn't work as often as not mapping somehow?

This looks nice, except `if ($AutoMapColumns) {` should be `if ($AutoColumnMap) {` yes?

wsmelton:
I would say right now we keep it tied to `-AutoColumnMap ` but for 1.0 release we can break that and just make it the default behavior ... @potatoqualitee ?

functions/Import-DbaCsv.ps1
petervandivier: yup, should prob have done a bit more testing & noticed that. I imagine it snuck by while I was importing/re-importing/editting just before I pushed. 

tbh, I'm not really familiar with how the appveyor pester for this function works (where is `SuperSmall.csv` saved for example, and how to define a target schema within the appveyor db against which to test this) but I'd be happy to do a bit more build out if you wanna slap a <kbd>DO NOT MERGE</kbd>🏷  on this and give me a couple days to figure it out 😄

tests/Import-DbaCsv.Tests.ps1
```diff
@@ -29,7 +29,7 @@ Describe "$CommandName Integration Tests" -Tag "IntegrationTests" {
     Context "Works as expected" {
         $results = $path | Import-DbaCsv -SqlInstance $script:instance1 -Database tempdb -Delimiter `t -NotifyAfter 50000 -WarningVariable warn
         It "accepts piped input and doesn't add rows if the table does not exist" {
-            $resulst | Should -Be $null
+            $results | Should -Be $null
```
petervandivier: ☝️  just want to double check that this _was_ in fact a typo that I fixed, and not an intentional bit of funny business?

potatoqualitee: totally a typo, thank you, @petervandivier !

potatoqualitee:
now im wondering if this should be default behavior. lemme merge and check then i'll decide before publishing 🌮

PR #5265, opened 2019-03-21
New command - Copy-DbaStartupProcedure

wsmelton:
## Type of Change

 - [ ] Bug fix (non-breaking change, fixes #<enter issue number>)
 - [x] New feature (non-breaking change, adds functionality, fixes #555)
 - [ ] Breaking change (effects multiple commands or functionality)
 - [ ] Ran manual Pester test and has passed (`.\tests\manual.pester.ps1)
 - [ ] Adding code coverage to existing functionality
 - [ ] Pester test is included
 - [ ] If new file reference added for test, has is been added to github.com/sqlcollaborative/appveyor-lab ?
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Allow migration of startup procedures for a given instance.

Current issue with this is the ability to just script out the code is broken. The `GO` terminator is missing in `Script()` method and the startup procedures have the definition to both create the procedure and set it as a startup all in the same script output. However because of this the error: `'CREATE/ALTER PROCEDURE' must be the first statement in a query batch` is spit out trying to use the script output.

I'll come back and put a test in after a while.

PR #5326, opened 2019-03-31
Startup procs

sqllensman:
## Type of Change

 - [X] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Change #5265 added new command Copy-DbaStartupProcedure
This adds the matching Get-DbaStartupProcedure and Set-DbaStartupProcedure commands to provide complete coverage.

### Commands to test
Get-DbaStartupProcedure 
Set-DbaStartupProcedure

### DIFF
--- a/functions/Get-DbaStartupProcedure.ps1
+++ b/functions/Get-DbaStartupProcedure.ps1
@@ -1,0 +1,86 @@
+function Get-DbaStartupProcedure {
+    <#
+    .SYNOPSIS
+        Get-DbaStartupProcedure gets startup procedures (user defined procedures within master database) from a SQL Server.
+
+    .DESCRIPTION
+        By default, this command returns for each SQL Server instance passed in, the name and schema for all procedures in the master database that are marked as a startup procedure.
+
+    .PARAMETER SqlInstance
+        The target SQL Server instance or instances.
+
+    .PARAMETER SqlCredential
+        Allows you to login to servers using SQL Logins instead of Windows Authentication (AKA Integrated or Trusted). To use:
+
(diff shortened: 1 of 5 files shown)


wsmelton:
I'm not sure I see a reason to have the get command, `Get-DbaDbStoredProcedure` outputs the SMO object which allows you to script out the given procedure. This one will not being that it is just an enumerator.

wsmelton:
I would rather see a Enable/Disable function instead of a Set...makes it more clear to users and more discover-able.

sqllensman:
Happy to modify to use Enable / Disable as this would match the TraceFlag commands.

The Get procedure can be modified to return same output as Get-DbaStoredProcedure.

However once we go down this route should I also modify to allow the Get procedure to be piped to Disable procedure?

sqllensman:
Also if Set function is to be split into multiple functions would it make sense to move code that parses an object name into Schema and Name into an internal function rather than maintaining multiple copies?

potatoqualitee:
> I would rather see a Enable/Disable function instead of a Set
Agreed

> However once we go down this route should I also modify to allow the Get procedure to be piped to Disable procedure?
Awesome, yes please!

> The Get procedure can be modified to return same output as Get-DbaStoredProcedure
Agreed, that'd help simplify your code, too 👍 

Thanks so much!

sqllensman:
Please close this pull request and I will submit a new request with expanded scope.

potatoqualitee:
looking forward to it, @sqllensman

PR #5393, opened 2019-04-20
fix Test-DbaConnection command

potatoqualitee:
was this horror

image

now is this

image


potatoqualitee:
actually i take it back, it'll look like this to match the other error

image

potatoqualitee:
also instancename is no longer blank if it's default

potatoqualitee:
dang, what do we do about this

image

that machine is online but sql isnt running, however, what if psremoting is online?

wsmelton:
Since it is a test connection I think it needs to return an object regardless of issue and the inject contain the results of all the test performed.

Issue #5397, opened 2019-04-22
New function: Backup-DbaServiceMasterKey


maddhatter:
---
name: Add `Backup-DbaServiceMasterKey` 🔐 

---

### Summary of new feature
Add a new `Backup-DbaServiceMasterKey` command, similar to the existing `Backup-DbaDbMasterKey`.

### Proposed technical details (if applicable)
SMO exposes the same Export command used in `Backup-DbaDbMasterKey` on the instance's SMK:
```powershell
$server = Connect-DbaInstance ServerA

$server.ServiceMasterKey.Export

OverloadDefinitions
-------------------
void Export(string path, string password)
```
T-SQL Command reference: https://docs.microsoft.com/en-us/sql/t-sql/statements/backup-service-master-key-transact-sql?view=sql-server-2017

### Latest version of dbatools as of writing

0.9.809

potatoqualitee:
🎉

PR #5398, opened 2019-04-22
Add Backup-DbaServiceMasterKey command

maddhatter:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Add `Backup-DbaServiceMasterKey` command for backing up the service master key (as proposed in #5397 

### Approach

Copied, then modified the `Backup-DbaDbMasterKey`.

### Commands to test

(These are the examples)
```powershell
Backup-DbaServiceMasterKey -SqlInstance server1\sql2016
Backup-DbaServiceMasterKey -SqlInstance Server1 -Path \\nas\sqlbackups\keys
```

### DIFF
--- a/functions/Backup-DbaServiceMasterKey.ps1
+++ b/functions/Backup-DbaServiceMasterKey.ps1
@@ -127,7 +127,7 @@ function Backup-DbaServiceMasterKey {
                 Add-Member -Force -InputObject $masterkey -MemberType NoteProperty -Name Filename -value $filename
                 Add-Member -Force -InputObject $masterkey -MemberType NoteProperty -Name Status -value $status
 
-                Select-DefaultView -InputObject $masterkey -Property ComputerName, InstanceName, SqlInstance, Database, 'Filename as Path', Status
+                Select-DefaultView -InputObject $masterkey -Property ComputerName, InstanceName, SqlInstance, 'Filename as Path', Status
             }
         }
     }
--- a/functions/Backup-DbaServiceMasterKey.ps1
+++ b/functions/Backup-DbaServiceMasterKey.ps1
@@ -1,0 +1,134 @@
+function Backup-DbaServiceMasterKey {
+    <#
+    .SYNOPSIS
+        Backs up specified database master key.
+
+    .DESCRIPTION
+        Backs up specified database master key.
+
+    .PARAMETER SqlInstance
+        The target SQL Server instance or instances.
+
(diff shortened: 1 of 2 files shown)


potatoqualitee:
yessssssssss! I was hoping to see this one! thank you so very much :bowtie:

maddhatter:
This is my first time trying to contribute a new command, so I'm a little iffy on the tests/ if the command name needs added to `dbatools.psm1`.

Another big ? is the auto-generated filename for the export. I went with
`<instance>-SMK-<timestamp>.key`.

functions/Backup-DbaServiceMasterKey.ps1
```diff
@@ -0,0 +1,131 @@
...
+
+        Logs into sql2016 with Windows credentials then backs up the service master key to the \\nas\sqlbackups\keys directory.
+
+    #>
+    [CmdletBinding(SupportsShouldProcess, ConfirmImpact = 'Low')]
+    param (
+        [parameter(Mandatory, ValueFromPipeline)]
+        [DbaInstanceParameter[]]$SqlInstance,
+        [PSCredential]$SqlCredential,
+        [PSCredential]$Credential,
```
wsmelton: I don't think we need to use just Credential here because of the standard we use for use of that parameter name.

We use this across commands to mean the cred used for remote access to the OS side of a server.

If this is just for encryption of the key I would say something like KeyCredential may be more appropriate...open to other name suggestions but I do think it needs to be something different than just Credential.

maddhatter: I was trying to stay consistent with the signature of `Backup-DbaDbMasterKey` since they do such similar things. I do think it's confusing though - I like KeyCredential, makes it much more obvious.

wsmelton: Yeah...if @potatoqualitee agrees we might want to update the parameter naming on that backup command just so we don't mix it up. I don't think there is a real reason to take a credential object just for the password, an example can easily be provided that covers the use of that secure string type that is needed.

But we will let @potatoqualitee make final call....

potatoqualitee: I prefer using credentials for passwords because i dont have to look up the syntax, i just have to use it. that said, i like the idea of updating the name! this is a fab PR, and will merge.

PR #5411, opened 2019-04-24
Add multi-IP and subnet support to Add-DbaAgListener

### LINKED ISSUE #5410: Modify Add-DbaAgListener to support multiple IP addresses and subnets
---
name: Feature request/idea :bomb:
about: As is, `Add-DbaAgListener` cannot work with listeners that need more than one IP or subnet.

---

### Summary of new feature

* Multi-IP Listener
* Cross-subnet litener
* Choose subnet for DHCP listener

* 0.9.809

briantist:
## Type of Change

 - [X] Pester test is included (sort of, see note)
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Closes #5410 

Adds support for multiple IP addresses and subnets to `Add-DbaAgListener`, and adds support for choosing subnet on a DHCP-based listener.

#### Parameter Changes
`$IPAddress` & `$SubnetMask` have had their types changed from `[ipaddress]` to `[ipaddress[]]`.
`$SubnetIP` has been added (also as `[ipaddress[]]`).

##### Parameter Considerations
To keep backward compatibility, invoking the command with the same parameters as the old version, will produce the same results. `$SubnetMask`'s default was preserved, `$SubnetIP` is optional in all cases (it can be calculated). 

Many parameter validation checks were added to ensure various combinations are valid and make sense. 

In addition, to support easier common cases, `$SubnetIP` and `$SubnetMask`, which should generally match in count to `$IPAddress`, can be single values, and if so they are applied to every IP.

For example, this means you can do `-IPAddress 1.1.1.1,2.2.2.2,3.3.3.3 -SubnetMask 255.0.0.0`, without having to do `-SubnetMask 255.0.0.0,255.0.0.0,255.0.0.0`.

If one `$SubnetIP` is specified, it is also used across all supplied IPs. 
If no `$SubnetIP` is specified, but `$IPAddress` is, then subnet is calculated from each IP and its associated netmask, so this parameter is often not needed, but is useful when you want a DHCP listener in one of multiple available subnets.

#### Execution Changes
The code has been modified so that iterates over the supplied IPs to add them to the listener object. Validation of whether the IPs/subnet are appropriate happens at the SQL level, so the object throws an error if, for example, not all of your replicas are represented in the subnets you chose.

### Commands to test
Any combinations of `$IPAddress`, `$SubnetIP`, and `$SubnetMask`, with one or more values in each field.

### Tests
I modified the test that checks for parameters. 
I made changes to the integration tests that I think could set it up for future tests, but don't really know how to run them or verify that, could use a hand there.


wsmelton:
Haven't checked Appveyor but the formatting looks like it was modified from mobile view so you may need to address that for test to pass.

wsmelton:
The test should be reverted as we don't add the function to the test file.

wsmelton:
Yeah, checked Appveyor and failure is because the test file is no longer valid so it definetly needs to be reverted back to the original.

```
Describe : dbatools Tests missing
Context  : Every function should have tests
Name     : It Add-DbaAgListener has validate parameters unit test
Result   : Failed
Message  : Expected 'Context "Validate parameters"' to be found in file 'tests\Add-DbaAgListener.tests.ps1', but it was not found.
```

potatoqualitee:
Excellent, thank you for the PR, @briantist  (lol, great username). I will revert the test as highlighted by @wsmelton and see if we get some passes.

functions/Add-DbaAgListener.ps1
```diff
@@ -92,27 +102,73 @@ function Add-DbaAgListener {
...
+                }
+            }
+        }
+
         foreach ($ag in $InputObject) {
             if ((Test-Bound -Not -ParameterName Name)) {
                 $Name = $ag.Name
             }
-            if ($Pscmdlet.ShouldProcess($ag.Parent.Name, "Adding $($IPAddress.IPAddressToString) to $($ag.Name)")) {
+            if ($Pscmdlet.ShouldProcess($ag.Parent.Name, "Adding $($IPAddress) to $($ag.Name)")) {
```
potatoqualitee: hey brian, why did the IPAddressToString get removed? I recall it being required for the proper output. Do you have a screenshot to share?

functions/Add-DbaAgListener.ps1
```diff
@@ -92,27 +102,73 @@ function Add-DbaAgListener {
...
 
+        if ($Dhcp) {
+            if (Test-Bound -ParameterName IPAddress) {
+                Stop-Function -Message "You cannot specify both an IP address and the Dhcp switch."
+                return
+            }
+
+            if ($SubnetMask.Count -gt 1) {
+                Stop-Function -Message "You can only specify a single subnet when using Dhcp."
+            }
```
potatoqualitee: please add a return after the Stop-Function

potatoqualitee:
Thanks much! after those two issues have been addressed, we can merge 💯

briantist:
Thanks all for the quick and warm response, fastest I've gotten trying to contribute to an open source project! 🎉 

Sorry for the test mess; I really screwed that up. It was a bad paste into the wrong file, so not only were my changes to the tests gone, my last fixes to the function were too. 

I think I resolved it in this latest push; thanks for your patience!

functions/Add-DbaAgListener.ps1
briantist: Strange, it shouldn't have been required previously, because the default stringification is the same, a la:
```
"$([ipaddress]'127.0.0.1')"
```

The reason I took it out was because `$IPAddress` became an array and I was expecting to use the default array-to-string functionality of joining with `$OFS` (space). Which does work, but as I respond to this I realize it should have worked the same without changing it too, since `.IPAddressToString` would just return an array via the PowerShell array member magic, which I just confirmed as well:

```powershell
"$(([ipaddress[]]@('127.0.0.1','127.0.0.2')))"
"$(([ipaddress[]]@('127.0.0.1','127.0.0.2')).IPAddressToString)"
```

So yeah.. the clarity / intent of the original is indeed better!
(added back)

potatoqualitee:
Excellent, thank you for the fix, your results have been merged and will be published to master/the PS Gallery momentarily 💃 🚀

PR #5417, opened 2019-04-26
Fix Set-DbaNetworkCertificate not granting permissions to private key

maddhatter:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
On a Windows Server 2016 (version 1607, Build 1439.2848) server, `Set-DbaNetworkCertificate` does not grant permissions to the private key.

### Approach

This appears to be because the `PrivateKey` property of the certificate is `$null`, and the `PrivateKey` is used to populate the path to the underlying certificate file.

Added a check to see if `PrivateKey` is null, and if it is use a different method to determine the underlying file. Also added a check to be sure that `$KeyFullPath` is a valid path to a file and abort if it isn't.

### Commands to test

```powershell
# On a Windows 2016 box (I'm using a newly provisioned Azure IaaS server built from the SQL Server 2017 marketplace image)
Set-DbaNetworkCertificate -Certificate $cert
```

### Learning

- [Stack Overflow answer to this problem](https://stackoverflow.com/a/52364549)

### DIFF
--- a/functions/Set-DbaNetworkCertificate.ps1
+++ b/functions/Set-DbaNetworkCertificate.ps1
@@ -71,10 +71,10 @@ function Set-DbaNetworkCertificate {
         [string]$Thumbprint,
         [switch]$EnableException
     )
-    
+
     process {
         # Registry access
-        
+
         if (Test-FunctionInterrupt) { return }
 
         if (-not $Certificate -and -not $Thumbprint) {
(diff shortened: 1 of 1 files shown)


potatoqualitee:
awesome solution, thank you! will merge when tests pass :bowtie:

potatoqualitee:
🎉 🎉 🎉

PR #5428, opened 2019-04-29
New-DbaAccessToken - initial commit

potatoqualitee:
maybe it'll be New-DbaAzAccessToken ? Unsure how much we'll be supporting AWS.

Anyway, here's a potential new command. We can explore, like one or two other commands, having an internal dependency on an Az module. Like, I kinda want to make it way easier to do a new storage token. Right now, I use a GUI for them. Or maybe not? 🤔


potatoqualitee:
gonna close then draft

PR #5442, opened 2019-04-30
Get-DbaToolsLog - Adding LastError parameter

wsmelton:
Added `-LastError` to log command in order to grab just the last error written to the log.

Example use case:

```powershell
try {
    Get-DbaDiskSpace -ComputerName DoesNotExists -EnableException
} catch {
    Get-DbatoolsLog -LastError
}
```


potatoqualitee:
i like it 👍

PR #5469, opened 2019-05-02
Add-DbaAgReplica - Remove default value of seeding mode

potatoqualitee:
if there's a mixed version AG, it'll pass the seeding mode  to lower versions and fail


potatoqualitee:
appveyor is having dns issues 👎

PR #5481, opened 2019-05-06
Measure-DbaBackupThroughput - adjusting output and formatting

wsmelton:
General formatting for 1.0 standard.
Removing default view filter so it matches output standards.


potatoqualitee:
thank you! failure due to appveyoru - has been reported

PR #5487, opened 2019-05-06
Update Copy-DbaAgentSchedule.Tests.ps1 with Integration Test

gbargsley:
## Type of Change

 - [X] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system

### DIFF
--- a/tests/Copy-DbaAgentSchedule.Tests.ps1
+++ b/tests/Copy-DbaAgentSchedule.Tests.ps1
@@ -12,8 +12,35 @@ Describe "$CommandName Unit Tests" -Tag 'UnitTests' {
         }
     }
 }
-<#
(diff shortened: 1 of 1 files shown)


gbargsley:
So Copy-DbaAgentSchedule copies all schedules and does not allow for a specific one to be copied.  How can I determine if Instance 2 has any Agent Schedules so I know how many total their will be after I add my test.

joshcorr:
Why not do a BeforeAll inside the Context block to gather the current number of Agent Schedules before running the copy in the test. That would ensure you know how many before moving. Or you could also write it to be Something like `$results | Should Be Greaterthan 1`

potatoqualitee:
Thanks much! It says fail but it's another test. Thanks so much, please keep em coming!

PR #5501, opened 2019-05-08
XML updates to better support azure

potatoqualitee:
not done yet, but want it on the agenda


wsmelton:
I'm not sure I follow this...you are removing support for what SQL Server allows in order to support using XE against Azure SQL?

If that is the cause I would recommend simply creating an additional copy of the templates, one for Azure SQL and one for usual SQL Server.

The event file target is supported for Azure SQL as well, you write the events to Azure Storage account.

potatoqualitee:
the default xmls failed _really_ hard when they had any filename associated so I basically removed the defaults but now i gotta add functionality back. it's hard.

potatoqualitee:
gonna do after 1.0

PR #5529, opened 2019-05-15
Adding ExtractPath parameter to multiple commands

### LINKED ISSUE #5512: [Bug] Update-DbaInstance trying to extract to path name too long
[Invoke-DbaAdvancedUpdate] Extraction failed with exit code 16389

The command is detecting our sqlclone VHDs as a suitable large disk to copy and extract the file to. The issue here being the path to our sql clone VHDs is quite long, too long to extract the update to.

Can we tell it to ignore sql clone disks AND/OR specify the path e.g. "C:\updates\"

I believe default naming is "SqlClone_VhdMount" e.g.

\SqlClone_VhdMount\dbatools_KB4488536_Extract_f5ee5e8fe25e4a949f74e5bab470f18a

Thanks,

garethnewman:
## Type of Change

 - [ X] Bug fix (non-breaking change, fixes #5512

 

### Purpose
Add an extract path variable $ExtractPath, allows to specify a manual location to extract too.
If no path specified, command behaves as before.
Added a hint to use -ExtractPath in the extract error.


functions/Invoke-DbaAdvancedUpdate.ps1
```diff
@@ -53,7 +56,8 @@ Function Invoke-DbaAdvancedUpdate {
         [ValidateSet('Default', 'Basic', 'Negotiate', 'NegotiateWithImplicitCredential', 'Credssp', 'Digest', 'Kerberos')]
         [string]$Authentication = 'Credssp',
         [pscredential]$Credential,
-        [switch]$EnableException
+        [switch]$EnableException,
+        [string]$ExtractPath
```
wsmelton: Ensure that `EnableException` is always the last parameter for usability purposes (tabbing). Same for CBH, it should be before Confirm in this case.

functions/Invoke-DbaAdvancedUpdate.ps1
```diff
@@ -40,6 +40,9 @@ Function Invoke-DbaAdvancedUpdate {
         This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
         Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
 
+    .PARAMETER ExtractPath
+        Lets you specify a location to extract the update file to on the system requiring the update. e.g. C:\temp
+
     .EXAMPLE
     PS C:\> Invoke-DbaAdvancedUpdate -ComputerName SQL1 -Action $actions
```
wsmelton: An example should be added for docs to draw attention to the usage of the parameter added.

functions/Invoke-DbaAdvancedUpdate.ps1
garethnewman: Ok, got it. CBH?

functions/Invoke-DbaAdvancedUpdate.ps1
garethnewman: got it

garethnewman: because Invoke-DbaAdvancedUpdate is called by Update-Dbainstace I wasn't sure how valid an example would be?

functions/Invoke-DbaAdvancedUpdate.ps1
garethnewman: I moved it, not sure about the CBH bit.

garethnewman:
@wsmelton anything else need doing to this?

functions/Invoke-DbaAdvancedUpdate.ps1
wsmelton: This is not an internal only function, it is public and usable by any user. So an example is needed.

functions/Update-DbaInstance.ps1
```diff
@@ -189,7 +202,7 @@ function Update-DbaInstance {
             }
         }
         $actions = @()
-        $actionTemplate = @{}
+        $actionTemplate = @{ }
```
wsmelton: please remove this space

functions/Invoke-DbaAdvancedUpdate.ps1
wsmelton: An example is still needed for this command. It is not an internal only command.

functions/Update-DbaInstance.ps1
potatoqualitee: this space was added by vs code if i had to guess

garethnewman: yeah my vs code likes to "format" powershell, I'll remove it.

functions/Invoke-DbaAdvancedUpdate.ps1
garethnewman: So it seems to get called from Update-DbaInstance using an action list, I am really not sure how to write the example for that. Perhaps someone could help me out.

There is currently only 1 example which assumes a list of actions has been generated somewhere else with no explanation how.

Also synopsis says
"Designed for internal use"

wsmelton: This function is not located in the internal directory and is exported as a public function. It can say designed or whatever else it wants to in the CBH, it still remains as a public function and therefore requires an example for the parameter. I don't know how to spell it any different.

potatoqualitee: I have added the example and also switched up the order of the help (not required necessarily, but works with my ocd ;)) Check out more info here: https://github.com/sqlcollaborative/dbatools/pull/5529/commits/d247567b09d756b8b7441605bc0eacaa9d58af73

garethnewman: Thanks @cl I'll have a look.

PR #5530, opened 2019-05-15
Remove tags for 1.0

potatoqualitee:
breaks boh, but im ready to start


niphlod:
why ?

potatoqualitee:
they werent maintained and the plan was to remove them in 1.0 anyway. but basically like waffle, was a good idea but just didnt work with my workflow and i hindered the results

PR #5539, opened 2019-05-16
Bug 5537 rename command

### LINKED ISSUE #5537: Rename watch-dbaupdate
to watch-dbatoolsupdate

garethnewman:
- [X] Bug fix (non-breaking change, fixes #5537)

### Purpose
renaming a command


potatoqualitee:
thanks again! so sorry, i missed on the last one, that it should not be DbaTools but rather Dbatools. If it's not too burdensome, can you also rename `Install-DbaWatchUpdate` and `Uninstall-DbaWatchUpdate`

Renaming the file itself to change the casing may require a delete, add, recommit

🙇

PR #5560, opened 2019-05-18
Remove-DbaDbUser - Address bug on piping support

### LINKED ISSUE #4971: Unable to pipe Get-DbaUser into Remove-DbaUser when multiple users are selected on the same database

- [X] Ensure you are able to reproduce it on the latest released version (_we release often_)
- [X] Verified this bug is not already reported in an issue
- [X] Verified errors are not related to permissions
- [X] Can reproduce in a clean PowerShell session (_clean = `powershell -NoProfile`_)

## Steps to Reproduce

```sql
CREATE USER USER01 WITHOUT LOGIN;
CREATE USER USER02 WITHOUT LOGIN;
```

```powershell
#Returns 2 users
get-dbadbuser -SqlInstance SQLSRV01 -Database DB01| Where-Object{ $_.Name -match "^USER(.*)" } | select Name

# Removes 1 user then errors
get-dbadbuser -SqlInstance SQLSRV01 -Database DB01| Where-Object{ $_.Name -match "^USER(.*)" } | Remove-DbaUser

#Workaround
get-dbadbuser -SqlInstance SQLSRV01 -Database DB01| Where-Object{ $_.Name -match "^USER(.*)" } | %{Remove-DbaUser -SqlInstance $_.SqlInstance -Database $_.Database -User $_.Name }

```

## Expected Behavior

Both users should be removed

## Actual Behavior

Single user is removed. 
```
An error occurred while enumerating through a collection: Collection was modified; enumeration operation may not execute..
At line:36361 char:17
+                 $users | ForEach-Object {
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (Microsoft.SqlSe...dListEnumerator:SmoSortedListEnumerator) [], RuntimeException
    + FullyQualifiedErrorId : BadEnumeration
```

### PowerShell: 
```
PSVersion                      5.1.14393.2636
BuildVersion                   10.0.14393.2636
```

### SQL Server: 
Microsoft SQL Server 2014 (SP2-CU13) (KB4456287) - 12.0.5590.1 (X64)   Aug  1 2018 01:23:36   Copyright (c) Microsoft Corporation  Enterprise Edition: Core-based Licensing (64-bit) on Windows NT 6.3 <X64> (Build 9600: ) (Hypervisor)

dbaduck:
## Type of Change

 - [X] Bug fix (non-breaking change, fixes #4971 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Fixing the Bug to remove the collection so that you don't get the error message about modifying the collection.

### DIFF
--- a/functions/Remove-DbaDbUser.ps1
+++ b/functions/Remove-DbaDbUser.ps1
@@ -203,12 +203,16 @@ function Remove-DbaDbUser {
     }
 
     process {
-        if ($InputObject) {
-            Remove-DbUser $InputObject
+            if ($InputObject) {
+            $server = Connect-DbaInstance -SqlInstance $InputObject.Parent.Parent.Name 
+            $user = $server.Databases[$InputObject.Parent.Name].Users[$InputObject.Name]
+
+            Remove-DbUser $user
+
         } else {
             foreach ($instance in $SqlInstance) {
                 try {
-                    $server = Connect-SqlInstance -SqlInstance $instance -SqlCredential $sqlcredential
+                    $server = Connect-DbaInstance -SqlInstance $instance -SqlCredential $sqlcredential
                 } catch {
                     Stop-Function -Message "Error occurred while establishing connection to $instance" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
                 }
--- a/functions/Remove-DbaDbUser.ps1
+++ b/functions/Remove-DbaDbUser.ps1
@@ -194,7 +194,7 @@ function Remove-DbaDbUser {
 
     process {
         if ($InputObject) {
-            $server = Connect-DbaInstance -SqlInstance $InputObject.Parent.Parent.Name
+            $server = Connect-SqlInstance -SqlInstance $InputObject.Parent.Parent.Name
             $user = $server.Databases[$InputObject.Parent.Name].Users[$InputObject.Name]
 
             Remove-DbUser $user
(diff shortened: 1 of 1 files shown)


wsmelton:
@dbaduck Please note that we utlize an internal function for making connections in the functions. I have reverted this change back as `Connect-DbaInstance` is a public functions that we have for user access.

wsmelton:
If we can get the test to pass on formatting this one is ready to merge.

dbaduck:
Thanks. That makes sense. I will make sure that I use the right one hereafter.

Ben Miller

wsmelton:
My test shows this solution is not working....it is prompting me to enter a user:
image

We should also update the test to actually validate piping support from both the parameter use and the filter use. I tested with just `Get-DbaDbUser ... | Remove-DbaDbUser` and it actually works properly. The problem is when that input object comes from `Where-Object` being used to filter the results down. This fails: `Get-DbaDbUser .. | Where-Object Name -match "mypattern" | Remove-DbaDbUser`

wsmelton:
Which reading the issue the input object being modified by the where object was noted.

functions/Remove-DbaDbUser.ps1
```diff
@@ -204,11 +194,15 @@ function Remove-DbaDbUser {
 
     process {
         if ($InputObject) {
-            Remove-DbUser $InputObject
+            $server = Connect-SqlInstance -SqlInstance $InputObject.Parent.Parent.Name
```
wsmelton: This method won't suffice as it does not have the full context of the connection to the instance. It will be missing connection information such as the login used to connect.

image

wsmelton:
Ok, so found the prompt for the user is because of the way parameter sets are being used in this command. It does not support piping the connection because the SqlInstance, SqlCredential are only allowed in the user parameter set and not with the input object.

potatoqualitee:
Thank you for bugfix and approval!

wsmelton:
This is actually broken and doesn't work as-is, we will need to do another PR.

potatoqualitee:
nooooo what

wsmelton:
It works in some scenarios and doesn't in others.

functions/Remove-DbaDbUser.ps1
potatoqualitee: oh wait, this is an easy fix

 $server = $InputObject.Parent.Parent

PR #5578, opened 2019-05-22
Add Azure Data Studio support to Get-DbaRegisteredServer

potatoqualitee:
A couple more enhancements can be made regarding the connection string but so far, pumped about this support


potatoqualitee:
image

PR #5580, opened 2019-05-23
Add support for alternative credentials in Azure Data Studio in Get-DbaRegisteredServer

potatoqualitee:
well, sql login, but not MFA yet


potatoqualitee:
garrr one pass

PR #5592, opened 2019-05-24
Rename Test-DbaVirtualLogFile to Measure-DbaDbVirtualLogFile

### LINKED ISSUE #2407: Test-DbaVirtualLogFile - this commands needs to actually test something
This command as-is does not test anything, it simply gets the VLF count and returns the information.

It should actually test the information. I would propose the current version of Test-DbaVirtualLogFile be renamed to `Get-DbaDbVirtualLogFile` or just `Get-DbaDbVlf`.

Then a proper Test-DbaVirtualLogFile command is written that validates the VLF count and offer some recommendation based on that count. I would expect based on the [gold standard](https://www.sqlskills.com/blogs/kimberly/transaction-log-vlfs-too-many-or-too-few/).

I don't know if we want to work out calculating based on the size of the current log file, how many VLFs it should have if you grew the file out properly (based on the gold standard). If it does not match then we should pop out a warning on it. I would say that is what a proper test command should provide, not just output of DBCC LOGINFO.

potatoqualitee:
fixes #2407

### DIFF
--- a/dbatools.psd1
+++ b/dbatools.psd1
@@ -113,7 +113,7 @@
         'Set-DbaDbOwner',
         'Test-DbaAgentJobOwner',
         'Set-DbaAgentJobOwner',
-        'Test-DbaDbVirtualLogFile',
+        'Measure-DbaDbVirtualLogFile',
         'Get-DbaDbRestoreHistory',
         'Get-DbaTcpPort',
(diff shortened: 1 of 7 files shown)


potatoqualitee:
one pass 🙄

PR #5608, opened 2019-05-26
Update Appveyor.prep for Path.DbatoolsExport

joshcorr:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Ensure that the DbatoolsConfig `Path.DbatoolsExport` folder is present in appveyor so `Export-dba` tests don't fail. Also includes a file cleanup for `Export-DbaInstance`
### Approach

Add the Static path to the Appveyor.prep.ps1 script. Since this runs before the Module is imported there is no good way to read the configured value, so if the default changes in the config then this will need to be changed manually.

### DIFF
--- a/tests/appveyor.prep.ps1
+++ b/tests/appveyor.prep.ps1
@@ -19,6 +19,12 @@ if (-not(Test-Path 'C:\Program Files\WindowsPowerShell\Modules\Pester\4.4.2')) {
     Install-Module -Name Pester -Force -SkipPublisherCheck -MaximumVersion 4.4.2 | Out-Null
 }
 
+#Setup DbatoolsConfig Path.DbatoolsExport path
+Write-Host -Object "appveyor.prep: Create Path.DbatoolsExport" -ForegroundColor DarkGreen
+if (-not(Test-Path 'C:\Users\appveyor\Documents\DbatoolsExport')) {
+    New-Item -Path C:\Users\appveyor\Documents\DbatoolsExport -ItemType Directory | Out-Null
+}
+
 
 #Get opencover.portable (to run DLL tests)
 Write-Host -Object "appveyor.prep: Install opencover.portable" -ForegroundColor DarkGreen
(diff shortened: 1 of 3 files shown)


joshcorr:
So... it looks like after fixing the appveyor environment this test revealed that there is some issue with `Export-DbaRepServerSettings` being called from `Export-DbaInstance` (Yeah Tests!) It looks like `Export-DbaRepServerSettings` doesn't have a validation for the `-Path` parameter.

potatoqualitee:
awesome! :D i do appreciate when tests discover bugs. would you have time/ be able to fix it and just commit to this branch?

joshcorr:
Sure, I can give it a shot. I can't reproduce yet in my own environment only happens to me in Appveyor.. Think I need to add something similar like what is in `Export-DbaSpConfigure`

joshcorr:
OMG 😮  that worked the first time..

potatoqualitee:
🥚 cellent!! does this mean that there will be issues in other envs like agent or people who have never run dbatools before?

tests/appveyor.prep.ps1
```diff
@@ -19,6 +19,12 @@ if (-not(Test-Path 'C:\Program Files\WindowsPowerShell\Modules\Pester\4.4.2')) {
     Install-Module -Name Pester -Force -SkipPublisherCheck -MaximumVersion 4.4.2 | Out-Null
 }
 
+#Setup DbatoolsConfig Path.DbatoolsExport path
+Write-Host -Object "appveyor.prep: Create Path.DbatoolsExport" -ForegroundColor DarkGreen
+if (-not(Test-Path 'C:\Users\appveyor\Documents\DbatoolsExport')) {
```
wsmelton: A cleaner method might be to do this in each individual environment setup as the module is already loaded we won't have to hardcode the path. Meaning you simply can run `Get-DbatoolsConfig -Name Path.DbatoolsExport` to retrieve the path and validate in Appveyor it is there first.... Or you simply duplicate this process in the test themselves, which gives more context as to why it is there.

joshcorr: so that would get moved to `tests\appveryor.SQL<ver>.ps1` instead of where it currently is?

joshcorr:
> 🥚 cellent!! does this mean that there will be issues in other envs like agent or people who have never run dbatools before?

I don't think so as I don't use the Dbatools config and didn't have any issues until now. Problem is that it is defined in the Config, but the path wasn't present on the system

PR #5613, opened 2019-05-27
Errors when database is not present/incorrect.

shaneis:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
If a database that is not present/incorrect is passed into `New-DbaDbUser`, the function should error out.
This is different from the silently returning that it was originally doing.

Attempted to add a Pester Unit test for this but it would not throw. Added a `-Skip` to the test until I can figure out what is causing it issues.

### Approach

Code now has a `if ($null -eq $databases -or $databases.Count -eq 0) { Stop-Function... }` block to error out.

### Commands to test

New-DbaDbUser -SqlInstance < instance name > -database 'FakeyMcFakeFake' -User 'ItDoesn'tMatter'


potatoqualitee:
awww yeahhhh! thank you 🙇

PR #5626, opened 2019-05-28
Update build ref index

gbargsley:
Build reference updated with 2016 CU7 and 2017 CU15.


potatoqualitee:
fantastic! thank you 💯 🥇

PR #5692, opened 2019-06-08
Export-DbaScript with ScriptBatchTerminator

potatoqualitee:
@caderoux - can you test if this works? 

First, I fixed another bug - we shouldn't be using `$script` as a variable. Second, we were overwriting some settings `if ($ScriptingOptionsObject.ScriptBatchTerminator) {` so I made it a bit more flexible.

Let me know if this fixes it for you.


potatoqualitee:
LOL, my attempt sucked. i'll try again later.

PR #5695, opened 2019-06-08
Copy-DbaLogin - Force / Group issue

potatoqualitee:
If your account is part of a Windows group, using Copy-DbaLogin with Force can result in killing your own access. This can be fixed using Reset-DbaAdmin but it's not preferable. Add warning and stop functoin  to address.

image


potatoqualitee:
Note that this warning/stoppage will not appear if you use `-Login` or `-ExcludeLogin` because it's assumed you will not drop your group.

PR #5705, opened 2019-06-10
Fixed Add-DbaAgDatabase

GeorgePalacios:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
To fix an issue where the Add-DbaAgDatabase command was targeting the Primary server for a restore operation rather than the secondary

### Approach
Changed the function to pass Instance SMO objects around rather than AG objects, and convert between the two only when needed. This means we don't need to reconnect to the secondary AG later on in the function and we can be sure we've got the correct server

### Commands to test
Commands in the help


potatoqualitee:
Solid changes! Thank you. Works on docker with weird ports and SQL auth so we should be set.

PR #5755, opened 2019-06-17
Add-DbaRegServer - remove overlap from azure data studio

potatoqualitee:
The Get was too broad

image


potatoqualitee:
one test fail, POS. everything else passes 👍

PR #5759, opened 2019-06-17
Read-DbaXeFile simplified DLL detection

potatoqualitee:
wtf


potatoqualitee:
im suspicious but will check why it fails later

PR #5764, opened 2019-06-17
Fix Secondary Log Shipping Monitor

sanderstad:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
When you assign a secondary monitor server, the command would not use that monitor server although it was set.

### Approach
Somewhere along the way variables for the primary server were passed along to the internal log shipping command that also makes the changes for the secondary monitor.

### Commands to test
Invoke-DbaDbLogShipping

### Screenshots
I used two different values because I did not have enough resources to test this on two different instances.
I used "localhost" for the primary and my local hostname (IT-SStad-Mobile) as the secondary. Technically they're the same but the values should be represented differently

**The right secondary server is used** 
image

**The right secondary monitor server is passed**
image


potatoqualitee:
this is gorgeous!  thank you 🙇

PR #5769, opened 2019-06-18
Add support for SQL Server 2000 and piping to Restore-DbaDatabase

potatoqualitee:
Previously, it kept saying "No backups passed through". This ignores the check which doesn't work only on sql2000 and I also saw sql2000 had some dupes so I added a Sort-Object -Unique to ensure there were no dupes. 

Code used to test: `Get-DbaDatabase -SqlInstance sql2000 -Database pubs32 |Backup-DbaDatabase -Path C:\temp |Restore-DbaDatabase -SqlInstance sql2000 -WithReplace`


potatoqualitee:
didnt work, maybe i changed too much

PR #5770, opened 2019-06-18
Backup-DbaDatabase - enable pipe w/o path on SQL2000

potatoqualitee:
the `.\whatever.bak` format did not work for sql2000


potatoqualitee:
unsure what's up with this branch. i feel like it's broken. Will merge and undo if required.

potatoqualitee:
azure should not be impacted by the change. it's even explicitly excluded (???)

PR #5773, opened 2019-06-18
.NET Core on Mac OS loads no errors

potatoqualitee:


potatoqualitee:
one pass - other failure was caused by maxed out azure, cleaned up

PR #5783, opened 2019-06-18
Add commands to psd1 and psm1 and some renames of new files

potatoqualitee:


potatoqualitee:
passed except for a thing i fixed 💯

PR #5785, opened 2019-06-18
X86 Fix

potatoqualitee:
Does not import non-compat DLLs in x86. really a workaround for a couple impacted commands but i dont expect many people to use x86.


potatoqualitee:
lookin bon 👌 
image

PR #5809, opened 2019-06-24
Add-Dba-Ag-Database - Added fix for named instances

GeorgePalacios:
$SecondaryInstance.netname wasn't the correct AG replica name

## Type of Change

 - [x] Bug fix (non-breaking change, fixes #5808  )
 - [ ] New feature (non-breaking change, adds functionality)
 - [ ] Breaking change (effects multiple commands or functionality)
 - [ ] Ran manual Pester test and has passed (`.\tests\manual.pester.ps1)
 - [ ] Adding code coverage to existing functionality
 - [ ] Pester test is included
 - [ ] If new file reference added for test, has is been added to github.com/sqlcollaborative/appveyor-lab ?
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
To fix #5808

### Approach
Construct the secondary replica name using the ComputerName + InstanceName properties

### Commands to test
Add-DbaAgDatabase with a named instance as the secondary

PR #5811, opened 2019-06-24
Endpoint / Replica fixes

potatoqualitee:
wrapping things I noticed in previous PR #5809

### DIFF
--- a/functions/Add-DbaAgDatabase.ps1
+++ b/functions/Add-DbaAgDatabase.ps1
@@ -185,10 +185,10 @@ function Add-DbaAgDatabase {
                     $secondaryInstanceReplicaName = $secondaryInstanceReplicaName, $secondaryInstance.InstanceName -join "\"
                 }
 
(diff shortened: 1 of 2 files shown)


potatoqualitee:
Confirmed working on docker.

functions/Add-DbaAgDatabase.ps1
```diff
@@ -185,10 +185,10 @@ function Add-DbaAgDatabase {
                     $secondaryInstanceReplicaName = $secondaryInstanceReplicaName, $secondaryInstance.InstanceName -join "\"
                 }
 
-                $agreplica = Get-DbaAgReplica -SqlInstance $Primary -SqlCredential $SecondarySqlCredential -AvailabilityGroup $ag.name -Replica $secondaryInstanceReplicaName
+                $agreplica = Get-DbaAgReplica -SqlInstance $Primary -SqlCredential $SqlCredential -AvailabilityGroup $ag.name -Replica $secondaryInstanceReplicaName
```
potatoqualitee: The reason sqlcredential is used is because it connects to the primary to get the info about the secondaries. the reason it was not returning anything before was that  `-Replica $secondaryInstanceReplicaName` was using an incomplete name.

Issue #5840, opened 2019-06-28
[Bug] Export-DbaUser is enumerating and displaying the current directory contents


haroldnorton:
When running the cmdlet, whatever folder is the current directory, the contents are being dumped to the host screen at the end.

sqllensman:
The code has following statement at end of function

Get-ChildItem -Path $script:pathcollection.Path

This does not seem correct.
Should be using
Get-ChildItem -Path $FilePath and only calling this if $Passthru is not specified

Issue #5732, opened 2019-06-12
Return more properties about Login (Get-DbaLogin)


ClaudioESSilva:
### Summary of new feature
Add more properties about the login using `LoginProperty` T-SQL function. 
Namely:
``` sql
BadPasswordCount
BadPasswordTime
DaysUntilExpiration
HistoryLength
IsMustChange
LockoutTime
PasswordHash
PasswordLastSetTime
```

### Proposed technical details (if applicable)
Add to `Get-DbaLogin` all properties that [LoginProperty](https://docs.microsoft.com/en-us/sql/t-sql/functions/loginproperty-transact-sql?view=sql-server-2017) T-SQL function returns and which are not by default on the SMO object.

Example get `PasswordLastSetTime`:
``` sql
SELECT LOGINPROPERTY ('TestLogin', 'PasswordLastSetTime');
GO
```

### Latest version of dbatools as of writing
Latest released version of dbatools module: v0.9.834

PR #5841, opened 2019-06-29
Return more properties about Login

sqllensman:
## Type of Change

 - [ ] Pester test is included
 - [X] If new file reference added for test, has is been added to github.com/sqlcollaborative/appveyor-lab ?
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Implement the changes for Feature request #5732 

### Approach
Added option to use SQL Query to get LoginProperty information for
BadPasswordCount
BadPasswordTime
DaysUntilExpiration
HistoryLength
IsMustChange
LockoutTime
PasswordHash
PasswordLastSetTime

### Commands to test
Get-DbaLogin

### DIFF
--- a/functions/Export-DbaUser.ps1
+++ b/functions/Export-DbaUser.ps1
@@ -415,11 +415,12 @@ function Export-DbaUser {
                 $sql += "`r`nGO"
             }
             if (-not $Passthru) {
-                $sql | Out-File -Encoding UTF8 -FilePath $FilePath -Append:$Append -NoClobber:$NoClobber
+                $sql | Out-File -Encoding UTF8 -FilePath  -Append:$Append -NoClobber:$NoClobber
+                Get-ChildItem -Path $FilePath
             } else {
                 $sql
             }
         }
-        Get-ChildItem -Path $script:pathcollection.Path
+
     }
 }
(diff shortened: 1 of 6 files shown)


potatoqualitee:
great idea! how about `-Detailed` instead of `GetLoginProperty` ?

sqllensman:
Ok.
I will update now

sqllensman:
The ealier Pull request to fix #5840 should have failed as it was missing the $FilePath parameter at Line 418

wsmelton:
I disagree with the detailed parameter here. We spent previous time removing that from every command moving toward 1.0.0.

Why are we know reverting back to it?

sqllensman:
I have no opinion either way.
Only reason I added a switch is avoid overhead of running query for every login if information is not wanted.

potatoqualitee:
Detailed was used without thought or patterns in previous commands. In this case, like Patrick mentioned, it's being used to avoid the overhead of running a query for every login if information is not wanted. 

In this case, Detailed is justified and intentional.

wsmelton:
So switch it to use SMO instead of the query. I haven't verified all but some of those can be returned via SMO: https://docs.microsoft.com/en-us/dotnet/api/microsoft.sqlserver.management.smo.login.passwordhashalgorithm?view=sql-smo-140.17283.0

PR #5852, opened 2019-07-02
Fix Export-DbaInstance logic error for Resource Governor script

potatoqualitee:
thanks for the solution @sqllensman !


potatoqualitee:
is this missing a `\` ?

PR #5854, opened 2019-07-03
Alert category

### LINKED ISSUE #5376: Export-DbaInstance - add user defined types
Found that this function doesn't script user defined types - and without them (if they exist) many procedures don't work.
Here they are -  `select * from sys.types where is_user_defined = 1`
Could you please add these types import to this function?

sqllensman:
## Type of Change

 - [X] Bug fix (non-breaking change, fixes #5376)
 - [X] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
In order to fix one of issues of #5376, namely scripting out alert categories it is necessary to add the functions to Add, Remove or Retrieve Alert Categories in SQL Agent. (similar to Job Categories)

Also fixed minor formatting issues in the AgentJobCategory functions 

### Commands to test
Get-DbaAgentAlertCategory
New-DbaAgentAlertCategory
Remove-DbaAgentAlertCategory
New-DbaAgentJobCategory
Remove-DbaAgentAlertCategory

### DIFF
--- a/functions/Get-DbaAgentAlertCategory.ps1
+++ b/functions/Get-DbaAgentAlertCategory.ps1
@@ -60,11 +60,11 @@ function Get-DbaAgentAlertCategory {
             }
 
             # get all the alert categories
-            $alertCategories = $server.JobServer.AlertCategories 
+            $alertCategories = $server.JobServer.AlertCategories
             if (Test-Bound -ParameterName Category) {
-                $alertCategories = $exportObjects | Where-Object { $Category -contains $_ }
+                $alertCategories = $alertCategories | Where-Object { $_.Name -in $Category }
             }
-              
+
             # Set the default output
             $defaults = 'ComputerName', 'InstanceName', 'SqlInstance', 'Name', 'ID', 'AlertCount'
 
@@ -79,7 +79,7 @@ function Get-DbaAgentAlertCategory {
                     Add-Member -Force -InputObject $cat -MemberType NoteProperty -Name ComputerName -value $server.ComputerName
                     Add-Member -Force -InputObject $cat -MemberType NoteProperty -Name InstanceName -value $server.ServiceName
                     Add-Member -Force -InputObject $cat -MemberType NoteProperty -Name SqlInstance -value $server.DomainInstanceName
-                    Add-Member -Force -InputObject $cat -MemberType NoteProperty -Name JobCount -Value $alertCount
+                    Add-Member -Force -InputObject $cat -MemberType NoteProperty -Name AlertCount -Value $alertCount
 
                     # Show the result
                     Select-DefaultView -InputObject $cat -Property $defaults
@@ -94,7 +94,7 @@ function Get-DbaAgentAlertCategory {
 
     end {
         if (Test-FunctionInterrupt) { return }
-        Write-Message -Message "Finished retrieving job category." -Level Verbose
+        Write-Message -Message "Finished retrieving alert category." -Level Verbose
     }
 
 }
--- a/functions/Get-DbaAgentAlertCategory.ps1
+++ b/functions/Get-DbaAgentAlertCategory.ps1
@@ -45,13 +45,14 @@ function Get-DbaAgentAlertCategory {
     [CmdletBinding()]
     param (
         [parameter(Mandatory, ValueFromPipeline)]
-        [DbaInstance[]]$SqlInstance,
+        [DbaInstanceParameter[]]$SqlInstance,
         [PSCredential]$SqlCredential,
         [string[]]$Category,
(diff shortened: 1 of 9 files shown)


potatoqualitee:
ohhh shoooooo! thank you 🙇‍♀

functions/Remove-DbaAgentAlertCategory.ps1
```diff
@@ -0,0 +1,117 @@
...
+                        $currentCategory = $server.JobServer.AlertCategories[$cat]
+
+                        Write-Message -Message "Removing alert category $cat" -Level Verbose
+
+                        $currentCategory.Drop()
+                    } catch {
+                        Stop-Function -Message "Something went wrong removing the alert category $cat on $instance" -Target $cat -Continue -ErrorRecord $_
+                    }
+
+                } #if should process
```
potatoqualitee: Thanks so much for the new commands! 

Some dbatools commands (and the ones that you based this off) still comment brackets in this manner, but it's only because I haven't cleaned them up yet ;) With common editors like VS Code, it's easy to collapse so comments like these are unnecessary.

In general, we want to reserve comments to explain why something unusual or complex is being done, instead of what is being done.

functions/Remove-DbaAgentAlertCategory.ps1
```diff
@@ -0,0 +1,117 @@
...
+
+            } # for each category
+
+        } # for each instance
+
+    } # end process
+
+    end {
+        if (Test-FunctionInterrupt) { return }
+        Write-Message -Message "Finished removing alert category." -Level Verbose
```
potatoqualitee: same for this, it seems unnecessary and it counters the way most other commands are. i know it's from sort of a template, but i haven't gotten to clean those up yet. sorrrrrrryyyyy.

functions/Get-DbaAgentAlertCategory.ps1
```diff
@@ -0,0 +1,100 @@
...
+    .EXAMPLE
+        PS C:\> Get-DbaAgentAlertCategory -SqlInstance sql1 -Category 'Severity Alert'
+
+        Return all the agent alert categories that have the name 'Severity Alert'.
+
+    #>
+    [CmdletBinding()]
+    param (
+        [parameter(Mandatory, ValueFromPipeline)]
+        [DbaInstance[]]$SqlInstance,
```
wsmelton:
```suggestion
        [DbaInstanceParameter[]]$SqlInstance,
```

The short name works but full name is preferred.

functions/Get-DbaAgentAlertCategory.ps1
```diff
@@ -0,0 +1,100 @@
...
+        [PSCredential]$SqlCredential,
+        [string[]]$Category,
```
wsmelton:
```suggestion
        [object[]]$Category,
```

@potatoqualitee we used to do these as object types for TEPP, are we still shooting for that?

functions/Get-DbaAgentAlertCategory.ps1
```diff
@@ -0,0 +1,100 @@
...
+                    Add-Member -Force -InputObject $cat -MemberType NoteProperty -Name ComputerName -value $server.ComputerName
+                    Add-Member -Force -InputObject $cat -MemberType NoteProperty -Name InstanceName -value $server.ServiceName
+                    Add-Member -Force -InputObject $cat -MemberType NoteProperty -Name SqlInstance -value $server.DomainInstanceName
+                    Add-Member -Force -InputObject $cat -MemberType NoteProperty -Name AlertCount -Value $alertCount
+
+                    # Show the result
+                    Select-DefaultView -InputObject $cat -Property $defaults
+                }
+            } catch {
+                Stop-Function -ErrorRecord $_ -Target $instance -Message "Failure. Collection may have been modified" -Continue
```
wsmelton: This seems like a template message...or I just have not seen it used before. Why would we say "may have been modified".

functions/Get-DbaAgentAlertCategory.ps1
potatoqualitee: it should be `string`, we generally moved away from `object` after we started doing `InputObject`, tho there are exceptions. we can shoot for tepp but it's not required. it's sort of a pain in the butt so i stopped 😅

sqllensman:
As all of these issues are inherited from the AgentJobSchedule functions I will update both these new functions and the AgentJobSchedule functions. 

Also should I be adding piping support so that results of the Get functions can be piped to Remove functions?

potatoqualitee:
that would be great and appreciated. thank you, patrick!

Issue #5853, opened 2019-07-03
[Bug] Import-DbaCsv Imports header row by default


Nullh:
#### PowerShell version:

```
Name                           Value                                                                                     
----                           -----                                                                                     
PSVersion                      5.1.16299.967                                                                             
PSEdition                      Desktop                                                                                   
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                                                                   
BuildVersion                   10.0.16299.967                                                                            
CLRVersion                     4.0.30319.42000                                                                           
WSManStackVersion              3.0                                                                                       
PSRemotingProtocolVersion      2.3                                                                                       
SerializationVersion           1.1.0.1        
                                                                           
```

#### dbatools Module version:

```

Name    : dbatools
Path    : C:\Program Files\WindowsPowerShell\Modules\dbatools\1.0.12\dbatools.psd1
Version : 1.0.12

Name    : dbatools
Path    : C:\Program Files\WindowsPowerShell\Modules\dbatools\1.0.0\dbatools.psd1
Version : 1.0.0

Name    : dbatools
Path    : C:\Program Files\WindowsPowerShell\Modules\dbatools\0.9.834\dbatools.psd1
Version : 0.9.834

```

#### SQL Server: 
```sql
Microsoft SQL Server 2019 (CTP3.1) - 15.0.1700.37 (X64)   Jun 19 2019 12:44:55   Copyright (C) 2019 Microsoft Corporation  Developer Edition (64-bit) on Linux (Ubuntu 16.04.6 LTS) <X64>
-- Also
Microsoft SQL Server 2017 (RTM-CU11) (KB4462262) - 14.0.3038.14 (X64)   Sep 14 2018 13:53:44   Copyright (C) 2017 Microsoft Corporation  Developer Edition (64-bit) on Linux (Ubuntu 16.04.5 LTS)
```

### Report

Import-DbaCsv imports the first row of data, even if it's header data. 

The help for the -NoHeaderRow flag specifies it should use the first row of the import file as a header row by default, but behaviour is the same if this flag is specified or not.

The same behaviour happens if you use the -AutoCreateTable flag or of importing into an existing table with or without data.

#### Host used

- [x] powershell.exe
- [x] ISE
- [x] VS Code
- [ ] Other (please specify)

If anything other than powershell.exe was used, please confirm that you can duplicate the issue with powershell.exe

- [x] Still buggy in powershell.exe

#### Errors Received

No error message thrown.

[...]

PR #5860, opened 2019-07-04
5853 Fix Import-DbaCsv not reading args

Nullh:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Fix for [[Bug] Import-DbaCsv Imports header row by default](https://github.com/sqlcollaborative/dbatools/issues/5853)

### Approach
The current system of forcing members of the CsvReader class seem to no longer work. 
Changed to use one of the constructors that takes all our args, but moved encoding to the StreamReader instantiation as for some reason the CsvReader constructors using that don't work!

### Commands to test
Import-DbaCsv using examples in help. Specific issues were test files with a header row always importing the header, and not being able to set a delimiter or string delimiter.

### DIFF
--- a/functions/Import-DbaCsv.ps1
+++ b/functions/Import-DbaCsv.ps1
@@ -128,6 +128,8 @@ function Import-DbaCsv {
         You can also choose AdvanceToNextLine which basically ignores parse errors.
 
     .PARAMETER Encoding
+        By default, set to UTF-8.
+
         The encoding of the file.
 
     .PARAMETER NullValue
--- a/functions/Import-DbaCsv.ps1
+++ b/functions/Import-DbaCsv.ps1
@@ -254,7 +254,7 @@ function Import-DbaCsv {
         [ValidateSet('AdvanceToNextLine', 'ThrowException')]
         [string]$ParseErrorAction = 'ThrowException',
         [ValidateSet('ASCII', 'BigEndianUnicode', 'Byte', 'String', 'Unicode', 'UTF7', 'UTF8', 'Unknown')]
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Yessssssssssssss! Just the fix I was hoping for! Let me take a closer look 👍

functions/Import-DbaCsv.ps1
```diff
@@ -300,7 +302,7 @@ function Import-DbaCsv {
                 [System.Data.SqlClient.SqlTransaction]$transaction
             )
             $reader = New-Object LumenWorks.Framework.IO.Csv.CsvReader(
-                (New-Object System.IO.StreamReader($Path)),
+                (New-Object System.IO.StreamReader($Path, [System.Text.Encoding]::$Encoding)),
```
potatoqualitee: I remember this problem now! It was hard as hell. I did not and would not have thought to move the encoding here, thank you so much 🙇‍♀️

potatoqualitee:
beautiful 💯  thanks for solving this problem. will release to Gallery and master shortly

functions/Import-DbaCsv.ps1
Nullh: Still at a loss to explain why those constructors wouldn't work! 
My C# ain't good enough to figure it out :)

potatoqualitee: mine either! i thought perhaps the docs were wrong but i used powershell to tell me the constructors and they said the same thing 🤔

PR #5863, opened 2019-07-04
Save-DbaUpdate - fixes #5337

### LINKED ISSUE #5337: Save-DbaUpdate
it would be awesome to have a Save-DbaUpdate that saves the updates required for a server. So specify the server, it gets the version, builds the chain until the most recent patch, searches (maybe this? https://www.catalog.update.microsoft.com/Search.aspx?q=sql+server) and saves them all to a path. 

It would also be helpful to be able to just specify the build. 

This could help with both Update-DbaInstance and Install-DbaInstance

/cc @nvarscar what do you think? i believe you have a lot of this code in your stuff.

potatoqualitee:
image

image

image

PAARRTTTTY 

Unsure if we should clean and surface Get-KBLink. It's pretty useful.


potatoqualitee:
Going to surface kb link so that we can make it pipable and avoid dupes / confirm what we're seeing

potatoqualitee:
Here is output of Get-DbaUpdateDetail. More details will be added as I have time. Gotta parse so it's time consuming.

image

PR #5867, opened 2019-07-07
Get-DbaKbUpdate - add more fields and a simple view

potatoqualitee:
what do yall think of this command's name? should it be Get-DbaUpdate or UpdateDetail? 

Anyway, this is what it looks like now, gonna turn it into a module for PS Gallery

image


potatoqualitee:
changed the command names to `-DbaKbUpdate`

potatoqualitee:
one pass 🙄

PR #5918, opened 2019-07-19
Export-DbaUser - Address User parameter properly filtering objects

VitalKoshalew:
Pull request https://github.com/sqlcollaborative/dbatools/pull/5627 has broken -User parameter. This patch restores fiter by user capability.

## Type of Change

 - [x] Bug fix (non-breaking change, fixes regession introduced in pull request 5627)
 

### Purpose
Fix regression (bug) introduced in pull request 5627: -User parameter is effectively ignored.

### Approach
Restore missing filtering as it was implemented before pull request 5627.

### Commands to test
Export-DbaUser -SqlInstance SQLServerName -User User1, User2 -Path resultsfoldername


wsmelton:
Please clarify how the referenced PR broke the command?

wsmelton:
The history of this file in our repository shows the code you added never existed prior to October 2018 (as far back as I bothered going). So this is a true bug, that for some reason no one has brought up; which is somewhat odd.

potatoqualitee:
The fix looks solid, thank you so much for fixing this big ol oversight! :100

PR #5941, opened 2019-08-01
Make Replication Parameter Type Generic

potatoqualitee:
Fixes this:

image

Hoping a side effect will be that it fixes azure too


potatoqualitee:
one pass

PR #5947, opened 2019-08-06
New Command - Get-DbaDbLogSpace

jpomfret:
## Type of Change

 - [X] Pester test is included
 - [ ] Nunit test is included
 - [X] Documentation
 - [ ] Build system
 

### Purpose
New command that will retrieve transaction log space usage. Uses `DBCC SQLPERF` so works on versions back to 2008.
Opening to name changes - not sure if I should have went with Get-DbaDbTransactionLogSpace or Get-DbaDbLogSpace ... or neither 😄 

### Approach
Returns custom object that meets dbatools standard output (I think) that details size of transaction log per database, percentage used and based on those calculates the space used.

### Commands to test
Included examples in CBH,
```
Get-DbaDbTLogSpace -SqlInstance Server1
Get-DbaDbTLogSpace -SqlInstance Server1 -Database Database1, Database2
Get-DbaDbTLogSpace -SqlInstance Server1 -ExcludeDatabase Database3
Get-DbaDbTLogSpace -SqlInstance Server1 -ExcludeSystemDatabase
```
### Screenshots
image

### Learning
[DBCC SQLPERF (Transact-SQL)](https://docs.microsoft.com/en-us/sql/t-sql/database-console-commands/dbcc-sqlperf-transact-sql?view=sql-server-2017)

### DIFF
--- a/functions/Get-DbaDbLogSpace.ps1
+++ b/functions/Get-DbaDbLogSpace.ps1
@@ -1,4 +1,4 @@
-Function Get-DbaDbTLogSpace {
+Function Get-DbaDbLogSpace {
     <#
     .SYNOPSIS
         Gets information on the database transaction log usage for each instance(s) of SQL Server.
@@ -37,30 +37,30 @@ Function Get-DbaDbTLogSpace {
         License: MIT https://opensource.org/licenses/MIT
 
     .LINK
-        https://dbatools.io/Get-DbaDbTLogSpace
+        https://dbatools.io/Get-DbaDbLogSpace
 
     .EXAMPLE
-        PS C:\> Get-DbaDbTLogSpace -SqlInstance Server1
+        PS C:\> Get-DbaDbLogSpace -SqlInstance Server1
 
         Returns the transaction log usage information for all databases on Server1
 
     .EXAMPLE
-        PS C:\> Get-DbaDbTLogSpace -SqlInstance Server1 -Database Database1, Database2
+        PS C:\> Get-DbaDbLogSpace -SqlInstance Server1 -Database Database1, Database2
 
         Returns the transaction log usage information for both Database1 and Database 2 on Server1
 
     .EXAMPLE
-        PS C:\> Get-DbaDbTLogSpace -SqlInstance Server1 -ExcludeDatabase Database3
+        PS C:\> Get-DbaDbLogSpace -SqlInstance Server1 -ExcludeDatabase Database3
 
         Returns the transaction log usage information for all databases on Server1, except Database3
 
     .EXAMPLE
-        PS C:\> Get-DbaDbTLogSpace -SqlInstance Server1 -ExcludeSystemDatabase
+        PS C:\> Get-DbaDbLogSpace -SqlInstance Server1 -ExcludeSystemDatabase
 
         Returns the transaction log usage information for all databases on Server1, except the system databases
 
     .EXAMPLE
-        PS C:\> Get-DbaRegisteredServer -SqlInstance cmsServer | Get-DbaDbTLogSpace -Database Database1
+        PS C:\> Get-DbaRegisteredServer -SqlInstance cmsServer | Get-DbaDbLogSpace -Database Database1
 
         Returns the transaction log usage information for Database1 for a group of servers from SQL Server Central Management Server (CMS).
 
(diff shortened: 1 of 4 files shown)


wsmelton:
Oh we got to remove that T from the name :grimacing:

jpomfret:
I can do that - happy with `Get-DbaDbLogSpace` ?

wsmelton:
🎉 yes, much better looking name. Since it has the `Db` prefix it should be known that it will be a t-log.

functions/Get-DbaDbTLogSpace.ps1
```diff
@@ -0,0 +1,118 @@
...
+        [string[]]$Database,
+        [string[]]$ExcludeDatabase,
+        [switch]$ExcludeSystemDatabase,
+        [switch]$EnableException
+    )
+
+    process {
+        foreach ($instance in $SqlInstance) {
+            try {
+                $server = Connect-SqlInstance -SqlInstance $instance -SqlCredential $SqlCredential
```
wsmelton: Do we need to validate versions that support the output or is it all the same from 2005+?

I wonder if we need to spit warning if the server passed in is for Azure SQL? The command will run when you are on Azure SQL but it won't return information about user databases. It only works against the system databases.

functions/Get-DbaDbTLogSpace.ps1
```diff
@@ -0,0 +1,118 @@
...
+            if ($ExcludeDatabase) {
+                $dbs = $dbs | Where-Object Name -NotIn $ExcludeDatabase
+            }
+
+            if ($ExcludeSystemDatabase) {
+                $dbs = $dbs | Where-Object IsSystemObject -eq $false
+            }
+
+            try {
+                $logspace = Invoke-DbaQuery -SqlInstance $server -Query "dbcc sqlperf(logspace)" | Where-Object { $dbs.name -contains $_.'Database Name' }
```
wsmelton: I'm on the fence on this one...will let @potatoqualitee chime in.

I'm for:
```suggestion
                $logspace = $server.Query("dbcc sqlperf(logspace)")
```

wsmelton: I would also prefer we validate version and for 2012+ use the new DMV. [sys.dm_db_log_space_usage](https://docs.microsoft.com/en-us/sql/relational-databases/system-dynamic-management-views/sys-dm-db-log-space-usage-transact-sql).

This requires executing within the context of each database but also includes an additional property for determining if you need to take a log backup (`log_space_in_bytes_since_last_backup`):
image

wsmelton: The image is against Azure SQL database as well.

functions/Get-DbaDbTLogSpace.ps1
jpomfret: It is the same from 2008+, I don't have a 2005 box to test with. @potatoqualitee do you have 2005 in your lab?

functions/Get-DbaDbTLogSpace.ps1
jpomfret: I wondered about checking for versions and using the newer dmv, chose to use dbcc instead to make it easy. I didn't think about Azure though - I'll add in this logic.

What's the benefit of using $server.Query over Invoke-DbaQuery?

wsmelton: You have additional overhead using Invoke-DbaQuery because it does more than just `$server.Query()` will....it calls the async command which is overkill for this purpose.

wsmelton: I can see responses coming about wanting the log space info since last log backup with the name of this command. If we do not want to add it now that is fine, it can just be added later on but I think it is needed. As well, the new DMV will actually return information for Azure SQL Database, where the DBCC command does not.

jpomfret: Oh I'll add the logic for using the DMV today, and that makes sense to change to Server.Query. Was just curious for the reasoning, thanks!

potatoqualitee:
this looks fab, thank you for the review, Shawn. agreed with @wsmelton 's name change - please do not forget to rename the file 💯  will test on older versions

potatoqualitee:
works all the way back to sql2000!

image

jpomfret:
Good shout on the file name! 
 - changed function & file name (& test file)
 - logic so if we are 2012 or higher use the new dmv
 - use $server.Query instead of invoke-dbaquery
 - change test for calculation to only run on versions less than 2012

Thanks both for the review 😸

functions/Get-DbaDbLogSpace.ps1
```diff
@@ -0,0 +1,139 @@
...
+                    } catch {
+                        Stop-Function -Message "Unable to select from sys.dm_db_log_space_usage on $instance." -ErrorRecord $_ -Target $db -Continue
+                    }
+                    [pscustomobject]@{
+                        ComputerName        = $server.ComputerName
+                        InstanceName        = $server.ServiceName
+                        SqlInstance         = $server.DomainInstanceName
+                        Database            = $db.name
+                        LogSize             = [dbasize]($logspace.total_log_size_in_bytes)
+                        LogSpaceUsedPercent = $logspace.used_log_space_in_percent
```
wsmelton: Random tidbit, that we don't have to put here...but did y'all know there is an overload for `.ToString()` where you can actually make it include the `%` sign on the output?

```
[PS 6.2.2] [08:53:00] > C:\Users\Shawn Melton
[1] > (45/254).ToString("P")
17.72%
```

functions/Get-DbaDbLogSpace.ps1
```diff
@@ -0,0 +1,139 @@
...
+                        LogSpaceUsed        = [dbasize]($logspace.used_log_space_in_bytes)
+                    }
+                }
+            } else {
+                write-host 'under 2012'
```
wsmelton: No, no no no no no no....please remove 😁

functions/Get-DbaDbLogSpace.ps1
```diff
@@ -0,0 +1,139 @@
...
+                try {
+                    $logspace = $server.Query("dbcc sqlperf(logspace)") | Where-Object { $dbs.name -contains $_.'Database Name' }
+                } catch {
+                    Stop-Function -Message "Unable to run dbcc sqlperf on $instance." -ErrorRecord $_ -Target $db -Continue
+                }
```
wsmelton: I don't think we need a try/catch for both with a custom message. Think we can just stick with a generic message as we do on other commands like `Unable to collect log space data on $instance`

functions/Get-DbaDbLogSpace.ps1
jpomfret: I did not know that, should I add it?

functions/Get-DbaDbLogSpace.ps1
jpomfret: my bad - this was for testing, thought i'd removed them all haha

functions/Get-DbaDbLogSpace.ps1
potatoqualitee: i dont think so, unless % is ignored in calculations

wsmelton: No, not to be added to our code...this is used for when you are ready to just display the data.

jpomfret: I got excited.... removed it :)

potatoqualitee:
LGTM 💃 !

PR #5954, opened 2019-08-10
Changes to add export functions for Roles

sqllensman:
## Type of Change

 - [X] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Add function to export Database and Server Roles
### Approach

Adds code that wraps T-SQL scripts that fully generate the Role scripts

This command is based off of John Eisbrener's post "Fully Script out a MSSQL Database Role"
Reference:  https://dbaeyes.wordpress.com/2013/04/19/fully-script-out-a-mssql-database-role/

Includes changes to
Get-DbaDbRole to fix issue with offline databases
Get-DbaServerRole to exclude AzureDB's

Used with Permission

### Commands to test
Get-DbaDbRole
Get-DbaServerRole
Export-DbaDbRole
Export-DbaServerRole


wsmelton:
Export of server roles I can understand, but what purpose would there be in exporting database roles?

sqllensman:
Getting the full permissions of a role is note easy via any current process (at least that I am aware of)

This is why the script from John Eisbrener was created

potatoqualitee:
thank you so much! your tests are glorious 🌞 💯 

Just a few changes needed:

Was testing how far this can go back and found that it needs a try catch around the enumeration of the dbs or unprocessable dbs need to be skipped.

image

2005 was only slightly problematic

image

and 2000 needs to be skipped

image

sqllensman:
I will look into the enumeration errors and also issue with SQL 2005 or earlier.
May take a day as this was done at SQLSaturday Auckland and I won't be home until tomorrow

wsmelton:
You will also need to exclude Azure SQL from being passed in since these queries are not supported. You can just add `-AzureUnsupported` to your `Connect-SqlInstance` call to handle this one.

sqllensman:
Changes added to exclude unsupported versions and fix problems with offline databases
Also replaced calls to Invoke-SqlQuery with .Query calls

potatoqualitee:
So close! You can do `-MinimumVersion` on Connect-SqlInstance if you can't get it to work.

image

sqllensman:
Do you have more details on where this error is being generated?
I have tried this against multiple SQL Server 2005 instances and cannot reproduce the above error.

The error looks like some issue with Collation differences . What is the collation of SQL2005 instance you are testing against?

wsmelton:
I'd also verify what compat-level the database is in (on 80, those queries would not run).

sqllensman:
Thanks @wsmelton 
I tested on a 2005 instance and putting a Database into compat-level 70 is able to reproduce the error.
Using compat-level 80 did not cause error but I will restrict to compat-level 90 or higher

potatoqualitee:
ohhh good call, i wouldnt have known

potatoqualitee:
looking great thank you!

PR #5955, opened 2019-08-12
A couple fixes

potatoqualitee:
- Remove exclamation points
- Fix a bug with Test-DbaConnection


potatoqualitee:
one pass - previous failures were due to azure being junky

PR #5975, opened 2019-08-15
Support Excluding Agent Server Properties during Migration

potatoqualitee:
Chose plural because it's all and not just some, ever


potatoqualitee:
Works as expected 👍  
image

Issue #1383, opened 2017-06-01
If $source exists as text within a trigger, it will materially change the content of the trigger


ConstantineK:
## Is this:
- [ ] *feature* 
- [x] *bug*:

In https://github.com/sqlcollaborative/dbatools/blob/master/functions/Copy-SqlServerTrigger.ps1#L143
we replace the instance name with the destination name, but in the case that the SQL Server name is also a keyword or something that is used in code, we might be making too greedy of a replacement. 

We need to replace only the location of the server within the portion that contains it instead of in the entire trigger.

#### Example:
1. Use Copy-SqlServerTrigger to copy a trigger
2. Name your source server any keyword which also appears in the trigger
3. The destination server (if the trigger is even created) will have a broken trigger.

ConstantineK:
This approach appears in a few other commands, which I also want to document here.

ConstantineK:
https://github.com/sqlcollaborative/dbatools/search?utf8=%E2%9C%93&q=%5BRegex%5D%3A%3AEscape%28%22%27%24source%27%22%29&type= There are >20 examples of this in the code, this looks like a burn down list after we find the general approach we want to take.

wsmelton:
This should reference all the commands that use that pattern in code, along with the line number:
https://github.com/sqlcollaborative/dbatools/pull/1322

wsmelton:
Chatting with @ConstantineK about this one in Slack as it is intriguing. @potatoqualitee I would say this revolves around the need to have the actual regex statement in the copy commands to begin with. I mean when you script a trigger (server or database) the is no reference to the server it came from in that script.

A SQL Job maybe the closes you get to needing it but even that is going to be special occasions. Unless someone has set the job server to a different server or something would be only reason it shows something other than `(local)`.

potatoqualitee:
so far this seems to be an effective way to handle most migrations. I will suggest `Export-DbaScript` for more power over the process.

PR #5979, opened 2019-08-16
Random fixes

potatoqualitee:
found these while using the commands extensively

### DIFF
--- a/functions/Copy-DbaAgentJob.ps1
+++ b/functions/Copy-DbaAgentJob.ps1
@@ -248,6 +248,8 @@ function Copy-DbaAgentJob {
                             $sql = $sql -replace [Regex]::Escape("@owner_login_name=N'$missingLogin'"), [Regex]::Escape("@owner_login_name=N'$saLogin'")
                         }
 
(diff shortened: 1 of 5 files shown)


functions/Copy-DbaAgentJob.ps1
```diff
@@ -248,6 +248,8 @@ function Copy-DbaAgentJob {
                             $sql = $sql -replace [Regex]::Escape("@owner_login_name=N'$missingLogin'"), [Regex]::Escape("@owner_login_name=N'$saLogin'")
                         }
 
+                        $sql = $sql -replace [Regex]::Escape("@server=N'$($sourceserver.DomainInstanceName)'"), [Regex]::Escape("@server=N'$($destServer.DomainInstanceName)'")
```
wsmelton: FYI #1383 was why we removed this in most every copy command.

potatoqualitee: Thank you. I recalled that convo but vaguely so I was happy to revisit it. 

The replacement ck talked about was too broad. This one is specifically for the impacted server ala @server=. Without it, I couldn't even migrate a number of my jobs. I had to export the script then modify manually.

PR #5983, opened 2019-08-16
New Command - Add-DbaRoleMember

### LINKED ISSUE #5982: [New Command] Add-DbaRoleMember for managing membership of server roles
Will add current logins to the provided roles. Will also include support for the two server roles included in Azure SQL/DW (`dbmanager` and `loginmanager`)

wsmelton:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality fixes #5982)
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Address missing functionality for adding logins and custom roles to server-level roles.

Required adjusting other commands around server-level roles to match naming and output. Piping support requires that we have a property output as `ServerRole` being that this is the parameter used across the commands. I left the "Role" property as default output and simply added the `ServerRole` to support piping. 

Unsure how we want to standardize these commands. On one hand having the `Role` property gives a command property across the server and database level commands, but on calling the command we differ. There is a need on the Add command in this PR to have the Role property because you of the membership support with custom roles.

- Updated some spots in `New-DbaServerRole` as it contained references to being related to database roles. (Probably just from copy/paste.)
- Updated test for the `Add-DbaDbRoleMember` as it was using static values in the test and not the test users actually being created in the `BeforeAll` block.
- Adjusted `New-DbaServerRole` to call `Get-DbaServerRole` instead of outputting from the role object so we properly show what was created. This also allows the same output we do in Get command to be in the New command.
- Adjusted output of `Get-DbaServerRole` to account for calling it in `New-DbaServerRole` so it would not cause any breaking changes. (Just had to add Owner property.)

Adjusting 
### Approach
Same approach as other server-level role commands on parameters.

### Commands to test
See examples


dbatools.psd1
```diff
@@ -491,6 +491,7 @@
         'Get-DbaDbLogSpace',
         'Export-DbaDbRole',
         'Export-DbaServerRole',
+        'Add-DbaServerRoleMember',
```
potatoqualitee: because of core and linux, this also needs to go in the psm1 👎

wsmelton:
Not sure I follow our format tester. I ran that invoke command against the file and this is all it changed:
image

Does that even have anything to do with OTSB formatting?

wsmelton:
I've included an example of using Azure SQL but until we fix the blockage on how the server name is being passed it is broken. This has to be fixed (it is not passing the proper server name when it comes to Azure SQL):

> WARNING: [23:27:00][Add-DbaServerRoleMember] Failure access dbatoolspbi.database.windows.net | The SqlAzureDatabase - [TCP:dbatoolspbi.database.windows.net,1433] is  not supported.

potatoqualitee:
yesssssssssssssssssssssss! and the OTSB, it just adds that and im like ok 😅

PR #5992, opened 2019-08-24
Connect-DbaInstance - catch issue

potatoqualitee:


potatoqualitee:
gonan disable that test

PR #5996, opened 2019-08-27
Copy-DbaDbTableData - allows copy using query with same dest/source server

jpomfret:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [X] Documentation
 - [ ] Build system
 

### Purpose
Currently if you choose the same server and database for the source and destination you can't use a query into a table. It checks and notices you are using same server/database and then stops the function.

### Approach
- Check if a query was used before stopping the function for matching server names
- Change the label on the should process so it gets the correct source name

### Commands to test
Added to CBH also.
```
$params = @{
    SqlInstance = 'server1'
    Destination = 'server1'
    Database = 'AdventureWorks2017'
    DestinationDatabase = 'AdventureWorks2017'
    Table = '[Person].[EmailPromotion]'
    BatchSize = 10000
    Query = "SELECT * FROM [Person].[Person] where EmailPromotion = 1"
}

Copy-DbaDbTableData @params
```


potatoqualitee:
yall are the best <3 thank you!

Issue #5945, opened 2019-08-05
Sync-DbaAvailabilityGroup - doesn't connect to Replicas using custom port


andrew-bickerton:
#### PowerShell version:
```
Name                           Value                                                                                                                                                                                     
----                           -----                                                                                                                                                                                     
PSVersion                      5.1.16299.251                                                                                                                                                                             
PSEdition                      Desktop                                                                                                                                                                                   
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                                                                                                                                                                   
BuildVersion                   10.0.16299.251                                                                                                                                                                            
CLRVersion                     4.0.30319.42000                                                                                                                                                                           
WSManStackVersion              3.0                                                                                                                                                                                       
PSRemotingProtocolVersion      2.3                                                                                                                                                                                       
SerializationVersion           1.1.0.1                                                                                                                                                                                   
```

#### dbatools Module version:
```
Name    : dbatools
Path    : C:\Program Files\WindowsPowerShell\Modules\dbatools\1.0.23\dbatools.psd1
Version : 1.0.23
```

#### SQL Server: 
```
Microsoft SQL Server 2017 (RTM-CU15) (KB4498951) - 14.0.3162.1 (X64)   May 15 2019 19:14:30   Copyright (C) 2017 Microsoft Corporation  Developer Edition (64-bit) on Windows Server 2016 Standard 10.0 <X64> (Build 14393: ) (Hypervisor) 
```

```sql
/* REPLACE WITH output of @@LANGUAGE */
```

### Report
When using Sync-DbaAvailabilityGroup as:
```
Sync-DbaAvailabilityGroup -Primary "PrimaryInstance,CustomPort" -AvailabilityGroup "AGName"
```

Response received is: 
WARNING: [12:12:44][Sync-DbaAvailabilityGroup] Error occurred while establishing connection to  | Access is denied

Through monitoring I've been able to determine that Sync-DbaAvailabilityGroup is retrieving the list of Replica names, but not their ports.  And is only trying to connect to the Replicas against port 1433.

There is a workaround:
```
Sync-DbaAvailabilityGroup -Primary "PrimaryInstance,CustomPort" -Secondary @("ReplicaOne,CustomPort","ReplicaTwo,CustomPort")
```

An ideal fix would be to have the powershell function determine the port number for the Replicas.  
If cannot determine port number perhaps a better sequence would be: 
1. Try connecting using the same port number as used on the primary node?
[...]

Issue #5919, opened 2019-07-22
Option to pass compatibility to Get-DbaDbCompatibility as an integer. 


ttoennies:
Get-DbaDbCompatibility returns compatibility as a string while Set-DbaDbCompatibility needs an integer.  Please consider adding a switch to Get-DbaDbCompatibility so that an integer could be returned which allow us to write a command such as this:

$Comp = Get-DbaDbCompatibility -SqlInstance psqlb7sm102 -Database master 

Set-DbaDbCompatibility -SqlInstance psqlb7sm102 -Database MyUserDB -TargetCompatibility $Comp 

Essentially setting the compatibility of a user database to the version of the server.  

Thank you.

ttoennies:
Get-DbaDbCompatibility returns compatibility as a string while Set-DbaDbCompatibility needs an integer.  Please consider adding a switch to Get-DbaDbCompatibility so that an integer could be returned which allow us to write a command such as this:

$Comp = Get-DbaDbCompatibility -SqlInstance psqlb7sm102 -Database master 

Set-DbaDbCompatibility -SqlInstance psqlb7sm102 -Database MyUserDB -TargetCompatibility $Comp 

Essentially setting the compatibility of a user database to the version of the server.  

Thank you.

wsmelton:
I see it as stating that Set command needs to support the string value returned by the get command to support proper piping.

The set command has the conversion code already in it we just don't allow the property and type correctly.

ttoennies:
The documentation suggests that the Set command needs the value as an integer; so how would the piping work?

This command:

Get-DbaDbCompatibility -SqlInstance server01 -Database master | Set-DbaDbCompatibility -SqlInstance server01 -Database userdb

produces this error:

Set-DbaDbCompatibility : The input object cannot be bound to any parameters for the command either because the command does not take pipeline input or the input and its properties do not 
match any of the parameters that take pipeline input.

While this command: 

$comp = Get-DbaDbCompatibility -SqlInstance server01 -Database master 

Set-DbaDbCompatibility -SqlInstance server01 -Database UserDB -TargetCompatibility $comp

Produces this error:

Set-DbaDbCompatibility : Cannot process argument transformation on parameter 'TargetCompatibility'. Cannot convert the "@{ComputerName=server01; 
SqlInstance=server01; Database=master; Compatibility=Version130}" value of type "System.Management.Automation.PSCustomObject" to type "System.Int32".

[...]

Issue #5905, opened 2019-07-16
Test-DbaLastBackup Output for if else statement possibly using wrong variable


JuleSQL:
### PowerShell Version:

```powershell
Name                           Value
----                           -----
PSVersion                      5.1.14393.3053
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.14393.3053
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1


```

### DBATools Version:

```powershell
Name    : dbatools
Path    : C:\Program Files\WindowsPowerShell\Modules\dbatools\1.0.21\dbatools.psd1
Version : 1.0.21

```
### Environmental information

```powershell
LCID             Name             DisplayName
----             ----             -----------
1033             en-US            English (United States)
```

#### SQL Server: 

```sql
Microsoft SQL Server 2017 (RTM-CU15) (KB4498951) - 14.0.3162.1 (X64)   May 15 2019 19:14:30   Copyright (C) 2017 Microsoft Corporation  Enterprise Edition (64-bit) on Windows Server 2016 Datacenter 10.0 <X64> (Build 14393: ) (Hypervisor) 
```

```sql
us_english
```

#### Host used

- [x] VS Code

If anything other than powershell.exe was used, please confirm that you can duplicate the issue with powershell.exe

- [x] Still buggy in powershell.exe

#### Errors Received
No error, variable for the backup size was not properly represented.

#### Steps to Reproduce
If you make -MaxSize smaller than the Database backup size, it will produce the message showing the backup size missing (Shown in 'Actual Behavior' section :
```
[...]

PR #5999, opened 2019-08-27
Addressing various submitted issues

wsmelton:
## Type of Change

 - [x] Bug fix
 - [ ] Breaking change
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Fixing various items to try and clean up some of our issues.

### Get-DbaDbCompatibility (breaking change)

#5919

`Get-DbaDbCompatibility` previously output the `Compatibility` property as a string (e.g. `Version140`). This PR adds a property, `Level`, to have the integer value be available.

### Get-DbaTcpPort

#5929

This command included a call to `Invoke-ManagedComputerCommand` but was passing a `$Credential` variable that the command never accepted as a parameter. So I updated the command to now accept that object.

In turn this fixes an issue `Test-DbaConnection` has because we were not passing in the actual Credential to `Get-DbaTcpPort`. I updated that call to now pass the Credential properly.

### Output messages

#5945
#5905

The other two items are just simply output messages.


potatoqualitee:
Thanks so much!! 

Tests for Get-DbaDbCompatibility need to be updated - they are currently failing, which is great becuase it means they work ;)

Should we add an extra field for CompatabilityVersion or some other name instead? The reason it used VersionXX is because thats how SMO outputs it.

wsmelton:
The error I believe is on the context calling for master and it is from deadlock. I had this happen running it locally one time, but any other time the test runs fine.

wsmelton:
This is failing due to an issue with Appveyor, the test runs finally locally now. I cannot make it issue any error in the last context block with the last adjustment.

I also just caught we need to add an artifact to Appveyor to output the `$error` object contents. The log file we capture won't include errors that Pester hits, so those exceptions are never visible.

wsmelton:
Oh well.

potatoqualitee:
aw why? i was looking forward to this one

wsmelton:
I'll be reworking it but should have kept my own standard of not including so much in one PR 😬

potatoqualitee:
haha K 🏆 !!

PR #6047, opened 2019-09-15
Update VS Code settings and PSSA rules file

wsmelton:
Update the psd1 file we include in the repository with proper rules for VS Code to enforce on our workspace. 

Applied the rules file to `powershell.scriptAnalysis.settingsPath` so the PS Extension knows what rules should be checked. This allows you to use the Problems panel in VS Code:
image


wsmelton:
No clue why that test failed, very random.

wsmelton:
Is there anything pending to merge this @potatoqualitee ?

PR #6058, opened 2019-09-23
Export-DbaInstance - issue nesting folders & BREAKING CHANGE removed param

jpomfret:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
**BREAKING CHANGE** - while I was working on the issue below I noticed the `$FilePath` parameter is never used... I removed it, but should that be aliased still?

When you pass in two SqlInstances the second servers files would be nested within the first servers folder:
image

### Approach
Don't reuse `$Path` as a variable within the loop, it overwrites the `$Path` parameter value.

### Commands to test
`Export-DbaInstance -SqlInstance "mssql1","mssql2" -SqlCredential $credential -Path C:\temp\ -Exclude $exclude`


potatoqualitee:
Thank you. I shall accept the breaking change - it wasn't working before anyway. An alias seems like a good idea. Please update the failing test ⭐️

potatoqualitee:
🙇

Issue #6059, opened 2019-09-23
Add 2019 versions to the build ref


jpomfret:
### Summary of new feature
Not sure if this is a new feature/enhancement or a bug. There is no 2019 versions in the build ref yet. Can this be added? Got the below off of http://sqlserverbuilds.blogspot.com/#sql2019x.
(This can be assigned to me if they can be added... 😄 

https://sqlcollaborative.github.io/builds

- 15.0.1900.47 - Microsoft SQL Server 2019 Release Candidate Refresh for Big Data Clusters only (RC1.1)
- 15.0.1900.25 - Microsoft SQL Server 2019 Release Candidate 1 (RC1)
- 15.0.1800.32 - Microsoft SQL Server 2019 Community Technology Preview 3.2 (CTP 3.2)
- 15.0.1700.37 - Microsoft SQL Server 2019 Community Technology Preview 3.1 (CTP 3.1)
- 15.0.1600.8 - Microsoft SQL Server 2019 Community Technology Preview 3.0 (CTP 3.0)
- 15.0.1500.28 - Microsoft SQL Server 2019 Community Technology Preview 2.5 (CTP 2.5)
- 15.0.1400.75 - Microsoft SQL Server 2019 Community Technology Preview 2.4 (CTP 2.4)
- 15.0.1300.359 - Microsoft SQL Server 2019 Community Technology Preview 2.3 (CTP 2.3)
- 15.0.1200.24 - Microsoft SQL Server 2019 Community Technology Preview 2.2 (CTP 2.2)
- 15.0.1100.94 - Microsoft SQL Server 2019 Community Technology Preview 2.1 (CTP 2.1)
- 15.0.1000.34 - Microsoft SQL Server 2019 Community Technology Preview 2.0 (CTP 2.0)

wsmelton:
Because it is not production we have not added it.

potatoqualitee:
does it need to be in production to be added? I think if it's useful for anyone's current environment, we can add. tho i can also see how that could introduce mess to the build website

wsmelton:
No, that is just the decision we made while back during CTP. It would have required constantly updating that based on when they released CTPs. Since it is RC, I'm find with just adding those, but not the CTP releases.

wsmelton:
Although, @niphlod would have to verify if RC can be added. Not sure if it would pick it up properly.

jpomfret:
[...]

PR #6067, opened 2019-09-25
Add 2019 versions to build ref

jpomfret:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Adding the 2019 builds - discussed in #6059


wsmelton:
Need to update the date at the top of the file.

wsmelton:
The only ones to be added will be the release candidates. You can just match how the ones for 2017 are structured.

The CTPs can be removed as we don't support or need to worry about those.

jpomfret:
@wsmelton is it formatted correctly now? I think the test failures would be expected since we don't have an RTM yet? Should I try and change those?

wsmelton:
We must not have added the RCs for SQL Server 2017 until the RTM was released. I would rather not change the test because they are there to ensure we have the proper structure and we just have not built in dealing with RC releases. 

We can likely just hold off adding them for now.

potatoqualitee:
I'd like to get the RC's in and want to help Jess get her demo to look good 💅 . Can @niphlod help modify the tests?

wsmelton:
RCs are added and included in the build index. But we do not add them until there is an RTM, unless we want to make up what the RTM version may or may not be and have it published as that.

jpomfret:
Could we exclude 2019 versions from the tests until RTM comes out? Then when that's added we could remove that exclusion? If we opened an issue to track it?

https://github.com/sqlcollaborative/dbatools/blob/8c28ffd8d98f3bb0b59393a7b6423083a659a336/tests/Get-DbaBuildReference.Tests.ps1#L97

wsmelton:
I'm working on the test now. The way RCs are released changed a bit with 2019 and to be accurate with the information we put out publicly requires the test validate the information a bit.

jpomfret:
Sounds good! Thank you, I appreciate it!

wsmelton:
Done. PR #6071 submitted.

PR #6087, opened 2019-10-03
Correcting connection messages and removing unused code

wsmelton:
- Connection error messages that didn't conform to standard.
- Removed `return` from `Read-DbaTraceFile` since it was to process multiple servers.


wsmelton:
So I didn't change any code on the read trace file command except correct the connection error message. However now the mocks that are in this test are broke for some reason in Appveyor.

It runs fine on my local test. Someone else may want to try it on their system to see if it works or breaks for them.

PR #6099, opened 2019-10-09
Added SQL Patches 2017 CU17 and 2016 SP2 CU10 to Build Ref json

gbargsley:


wsmelton:
Need to update the last modified date at the top.

wsmelton:
Some of the test started failing but build ref tests passed when viewing the logs.

PR #6105, opened 2019-10-13
Fix example resolve dbanetworkname

Splaxi:
## Type of Change

 - [X] Documentation
 

### Purpose


wsmelton:
This change is invalid because the examples are there to show you can run the command against one instance or against two.

Splaxi:
I just downloaded the latest from PowershellGallery and ran the following command, based on the examples:

```
Resolve-DbaNetworkName -SqlInstance sql2014\SQL2008R2
```
image

The `-SqlInstance` parameter is not a valid parameter for this command. I might be missing something obvious because my testing is so narrow, but I would argue that the examples that I removed are not valid as they are being presented at the moment.

wsmelton:
Ok, that is fine but your PR is not addressing that issue at all.

wsmelton:
If you want to update the PR to address the parameter reference that is fine but both examples are to stay in place...otherwise the PR can just be closed and we will address it with other doc fixes that are to be done.

Splaxi:
I'm sorry - but I don't think I quite understand what I'm missing.

There are a total of 4 examples as of today in the function. 2 of these are using the `-SqlInstance` parameter, which is invalid.

As an effort in showing the users that they can run against multiple instances, across multiple servers, I updated the last example with the `Get-DbaRegServer` function to show just that.

Should I provide the following examples instead:

```
Resolve-DbaNetworkName -ComputerName ServerA

Resolve-DbaNetworkName -ComputerName ServerA, ServerB

Get-DbaRegServer -SqlInstance sql2016\sqlexpress | Resolve-DbaNetworkName

Get-DbaRegServer -SqlInstance sql2016\sqlexpress, sql2014 | Resolve-DbaNetworkName
```

If you feel the effort is to high to guide me in the right direction, I'm fine with you closing the PR. I'll just skip testing this function going forward.

Please note that english is not my native language, so I might misunderstand something based on the language barrier.

wsmelton:
The change you have in this PR just moved an instance name to the example. There are no changes made to correct the parameter reference:

image

I would expect to see the following:

```
   .EXAMPLE
        PS C:\> Resolve-DbaNetworkName -Computer sql2016

        Returns a custom object displaying InputName, ComputerName, IPAddress, DNSHostName, DNSDomain, Domain, DNSHostEntry, FQDN, DNSHostEntry  for the SQL instance sql2016
    .EXAMPLE
        PS C:\> Resolve-DbaNetworkName -Computer sql2016, sql2014

        Returns a custom object displaying InputName, ComputerName, IPAddress, DNSHostName, DNSDomain, Domain, DNSHostEntry, FQDN, DNSHostEntry  for the SQL instance sql2016 and sql2014
```

wsmelton:
If you want to add a 3rd example to show the piping that would be acceptable as well...just need the other two examples put back and corrected.

PR #6109, opened 2019-10-13
Fix example get dbacmobject

Splaxi:
## Type of Change

 - [X] Documentation
 

### Purpose


wsmelton:
I'm not sure how your test runs since it is not provided but the example is valid. PowerShell binding defaults to positions, and the class name parameter is first parameter of the command. So the example is valid as-is.

Splaxi:
Hi Shawn,

Sorry for nagging you, but my manual testing also shows that the current example is not valid:

```
PS C:\> Get-DbaCmObject win32_OperatingSystem
```

I'm running Windows 10, with the following details: 

image

I would argue that we either need to change:
https://github.com/sqlcollaborative/dbatools/blob/f571af240222a94ac2902ec8a838744814e4a217/functions/Get-DbaCmObject.ps1#L74-L77

Into:
```
[Parameter(Mandatory, ParameterSetName = "Class", Position=0)]
        [Alias('Class')]
        [string]
        $ClassName,
```

or accept this PR that updates the example to show people that they need to include the -ClassName parameter.

The way I'm reading and understanding the Position binding, it actually requires you to specify the position attribute for the parameter, before PowerShell will bind it for you. 

Based on this docs: https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_functions_advanced_parameters?view=powershell-6#position-argument

Especially this section:
_If the Position argument isn't specified, the parameter name, or a parameter name alias or abbreviation, must precede the parameter value whenever the parameter is used in a command._

PR #6111, opened 2019-10-13
Fix example export dbalogin

Splaxi:
## Type of Change

 - [X] Documentation
 

### Purpose


functions/Export-DbaLogin.ps1
```diff
@@ -128,7 +128,7 @@ function Export-DbaLogin {
         Exports ONLY logins FROM sqlserver2014a with permissions on databases HR and Accounting
 
     .EXAMPLE
-        PS C:\> Export-DbaLogin -SqlInstance sqlserver2008 -Login realcajun, netnerds -Path C:\temp\login.sql -ExcludeGoBatchSeparator
+        PS C:\> Export-DbaLogin -SqlInstance sqlserver2008 -Login realcajun, netnerds -Path C:\temp\login.sql
```
wsmelton: This is actually invalid example based on the description being `without the GO batch separator`. Your test is not likely going to pick this up because we moved how this works to our configurations. In order for this example to be valid an additional command has to be added.

```suggestion
        PS C:\> Set-DbatoolsConfig -FullName formatting.batchseparator -Value $null
        PS C:\> Export-DbaLogin -SqlInstance sqlserver2008 -Login realcajun, netnerds -Path C:\temp\login.sql
```

Splaxi: Thanks for pointing that out. You are right, my testing is fairly simply, and only catches obvious issues, so I wouldn't catch this at all.

PR #6131, opened 2019-10-17
Fix Pipeline Input for Invoke-DbaAgFailover

maddhatter:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 

Piping AGs to the `Invoke-DbaAgFailover` command complains about a missing `SqlInstance` parameter:

```powershell
Get-DbaAvailabilityGroup SERVER | Invoke-DbaAgFailover

cmdlet Invoke-DbaAgFailover at command pipeline position 2
Supply values for the following parameters:
SqlInstance[0]:
```

### Approach

Add `ParameterSetName` attributes to the `SqlInstance` and `InputObject` parameters to allow both to work as expected.

### Commands to test

The examples in the help work:
```powershell
# Pipeline
Get-DbaAvailabilityGroup SERVER | Invoke-DbaAgFailover

# No pipeline
Invoke-DbaAgFailover -SqlInstance SERVER -AvailabilityGroup TestAG
```

### DIFF
--- a/functions/Invoke-DbaAgFailover.ps1
+++ b/functions/Invoke-DbaAgFailover.ps1
@@ -63,11 +63,11 @@ function Invoke-DbaAgFailover {
     #>
     [CmdletBinding(SupportsShouldProcess, ConfirmImpact = 'High')]
     param (
-        [parameter(Mandatory)]
+        [parameter(ParameterSetName = "NoPipe", Mandatory)]
         [DbaInstanceParameter[]]$SqlInstance,
         [PSCredential]$SqlCredential,
         [string[]]$AvailabilityGroup,
-        [parameter(ValueFromPipeline)]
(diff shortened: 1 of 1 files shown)


wsmelton:
Appreciate the contribution but we prefer to not use parameter sets because we cannot control the error that comes from their general use.

If you'd like to create an bug report (not sure one exists for this already) for what you found is the issue we can discuss in the issue how it can be handled with our code base.

potatoqualitee:
The `Mandatory` can be removed and we can just allow the user to process nothing - some of our commands do that.

Alternatively, you can remove the `Mandatory` then do a check in `process` for either `Sqlinstance` or `InputObject` and catch it with a `Stop-Function` if nothing was passed 👍

Issue #6133, opened 2019-10-17
Test-DbaLastBackup - change default, or add parameter to use the disk device type


MediocreDBA:
### Summary of new feature
When using the Test-DbaLastBackup function if there was a backup done to a virtual device (typically through a third party tool calling VSS, Veeam, VMWare snapshot, etc) after the last full backup, this virtual backup is attempted to be restored, which then fails.

Either defaulting to a device type of disk, or adding a paramater to specify should resolve the issue. 

### Proposed technical details (if applicable)
This is already supported in Get-DbaDbBackupHistory which is how the backup files are being gathered, so should be an isolated change to this call.

### Latest version of dbatools as of writing
1.0.54

PR #6134, opened 2019-10-17
Update Test-DbaLastBackup to specify device type of disk

MediocreDBA:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
To limit backups used in Test-DbaLastBackup to only those of the disk device type. Issue #6133

### Approach

I included the -DeviceType Disk in the calls to Get-DbaBackupHistory. I don't believe any other device types would be usable through this function, so expanding to other types isn't necessary. If it turns out there are other use cases, it should be a simple change to include them.

### DIFF
--- a/functions/Test-DbaLastBackup.ps1
+++ b/functions/Test-DbaLastBackup.ps1
@@ -64,7 +64,10 @@ function Test-DbaLastBackup {
 
     .PARAMETER MaxSize
         Max size in MB. Databases larger than this value will not be restored.
-
+        
+    .PARAMETER IncludeDiskDeviceOnly
+        If this switch is enabled, only backups to a disk device will be included
+        
     .PARAMETER AzureCredential
         The name of the SQL Server credential on the destination instance that holds the key to the azure storage account.
 
@@ -166,6 +169,7 @@ function Test-DbaLastBackup {
         [switch]$CopyFile,
         [string]$CopyPath,
         [int]$MaxSize,
+        [switch]$IncludeDiskDeviceOnly,        
         [switch]$IncludeCopyOnly,
         [switch]$IgnoreLogBackup,
         [string]$AzureCredential,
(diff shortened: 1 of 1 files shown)


functions/Test-DbaLastBackup.ps1
```diff
@@ -274,18 +274,19 @@ function Test-DbaLastBackup {
                 $CopyFile = $false
             }
 
-            Write-Message -Level Verbose -Message "Getting recent backup history for $($db.Name) on $instance."
+            Write-Message -Level Verbose -Message "Getting recent backup 
+            for $($db.Name) on $instance."
```
potatoqualitee: Thanks so much for the PR! It appears this carriage return was unintentional.

functions/Test-DbaLastBackup.ps1
```diff
@@ -274,18 +274,19 @@ function Test-DbaLastBackup {
...
 
             if (Test-Bound "IgnoreLogBackup") {
                 Write-Message -Level Verbose -Message "Skipping Log backups as requested."
                 $lastbackup = @()
-                $lastbackup += $full = Get-DbaDbBackupHistory -SqlInstance $sourceserver -Database $dbname -IncludeCopyOnly:$IncludeCopyOnly -LastFull -WarningAction SilentlyContinue
-                $diff = Get-DbaDbBackupHistory -SqlInstance $sourceserver -Database $dbname -IncludeCopyOnly:$IncludeCopyOnly -LastDiff -WarningAction SilentlyContinue
+                $lastbackup += $full = Get-DbaDbBackupHistory -SqlInstance $sourceserver -Database $dbname -IncludeCopyOnly:$IncludeCopyOnly -LastFull -DeviceType Disk -WarningAction SilentlyContinue
+                $diff = Get-DbaDbBackupHistory -SqlInstance $sourceserver -Database $dbname -IncludeCopyOnly:$IncludeCopyOnly -LastDiff -DeviceType Disk -WarningAction SilentlyContinue
```
potatoqualitee: I am unsure of other people's use cases or if we can even support them properly. This suits my needs but I don't know if it would break other people's workflows.

What do you think about adding a new parameter and defaulting it to Disk instead? What do you think @wsmelton ?

MediocreDBA: Doesn't bother me, i'm okay with it either way

wsmelton: A new parameter can be added but we should not default it to disk. URL backups is a separate device type and would break workflows for those users.

MediocreDBA: Ahh good point, not something I use in my environment today. I'll go ahead and add it as a parameter matching how Get-DbaDbBackupHistory uses it. So by default the functionality wouldn't change, or you could specify an array of types to use.

potatoqualitee:
Why'd you close your PR @MediocreDBA ? 😢  You could have updatedyouru branch and it'd update on our side.

PR #6143, opened 2019-10-19
Install-DbaWhoIsActive - Cross Platform Support Added

### LINKED ISSUE #6126: Install-DbaWhoIsActive with -LocalFile parameter deletes the local file
### Summary of new feature

I was testing out the Install-DbaWhoIsActive cmdlet and using the -LocalFile parameter.  The stored procedure already existed on the target server, so it was updated, but the local file was then deleted.  I understand why you would want to delete it if it was the downloaded zip file.  Is there a way to have it not delete the local file if you use the -LocalFile parameter?
Thank you

I'm using version 1.0.50 of dbatools module.

joshcorr:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #6126)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Add cross platform support for installing sp_WhoisActive from Linux or Mac.
Also remove logic for deleting the file after installing as requested in #6126 
After looking at other `Install-` cmdlets it looks like we delete the ZIP (which this is doing), but not the other files as a cache.

### Approach

Replace explicit pathing with `Join-Path` and add a check for `Unblock-File` when ran on Windows.
Also moved the Function name to the xplat section of *.psm1

### Commands to test

Run any example from any system running PowerShell Core.
`Install-DbaWhoIsActive -SqlInstance sqlserver2014a -SqlCredential $cred`

### Screenshots

While the screen shot shows it working from a cached file, it also worked from downloading directly from GitHub.
image

### DIFF
--- a/functions/Install-DbaWhoIsActive.ps1
+++ b/functions/Install-DbaWhoIsActive.ps1
@@ -136,7 +136,7 @@ function Install-DbaWhoIsActive {
             # Unpack
             # Unblock if there's a block
             if ($PSCmdlet.ShouldProcess($env:computername, "Unpacking zipfile")) {
-                if (Test-Windows) {
+                if (Test-Windows -NoWarn) {
                     Unblock-File $zipfile -ErrorAction SilentlyContinue
                 }
                 try {
--- a/functions/Install-DbaWhoIsActive.ps1
+++ b/functions/Install-DbaWhoIsActive.ps1
@@ -89,8 +89,8 @@ function Install-DbaWhoIsActive {
         if ($Force) { $ConfirmPreference = 'none' }
 
         $DbatoolsData = Get-DbatoolsConfigValue -FullName "Path.DbatoolsData"
-        $temp = ([System.IO.Path]::GetTempPath()).TrimEnd("\")
-        $zipfile = "$temp\spwhoisactive.zip"
+        $temp = ([System.IO.Path]::GetTempPath())
+        $zipfile = Join-Path -Path $temp -ChildPath "spwhoisactive.zip"
 
         if ($LocalFile -eq $null -or $LocalFile.Length -eq 0) {
             $baseUrl = "https://github.com/amachanic/sp_whoisactive/archive"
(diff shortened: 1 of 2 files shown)


functions/Install-DbaWhoIsActive.ps1
```diff
@@ -136,8 +136,9 @@ function Install-DbaWhoIsActive {
             # Unpack
             # Unblock if there's a block
             if ($PSCmdlet.ShouldProcess($env:computername, "Unpacking zipfile")) {
-
-                Unblock-File $zipfile -ErrorAction SilentlyContinue
+                if ([Environment]::OSVersion.Platfrom -eq 'Win32NT') {
```
wsmelton: I would prefer to use `$PSVersionTable.Platform` as that is what we are doing in other places in the module.

joshcorr: I am basing this change/type off of the following article.. (https://powershell.org/2019/02/tips-for-writing-cross-platform-powershell-code/) $PSVersionTable.Platfrom returns null on 5.1

joshcorr: [Environment]::OSVersion.Platfrom returns a value regardless of version.

joshcorr: Although I just realized it is misspelled.. 😨

wsmelton: https://github.com/sqlcollaborative/dbatools/blob/e8d440c0e09a4937b7301696cb3c7dba7128d705/dbatools.psm1#L11-L15

joshcorr: I understand that is how it is currently being done, but maybe we could change the way we are doing it. [Environment]::OSVersion is supported back to .NET 1.1 and give a ubiquitous answer.. WinNT32 or Unix. It is just a little more concise. I guess that is why I like it. I have the spell change staged, but if you want the long form I can rewrite.

wsmelton: My only issue with it is it takes longer to run...even longer in PS6+ than just using `$PSVersionTable`.

image

image

wsmelton: While that is just showing ms time, we start doing it everywhere in commands that we get to cross-platform and it starts adding up.

wsmelton: So I have an alternative. We just use the internal function `Test-Windows`
```suggestion
                if (Test-Windows) {
```

joshcorr: Cool.. that works too.. was about to suggest something like that. I am going to look at the other `Install-` cmdlets as well, so I knew this was going to come up again..

joshcorr:
oooh @wsmelton  before approving that I may add `-nowarn` so that it doesn't write-message

PR #6144, opened 2019-10-19
Update-Dbatools - Cross Platform Support Added

joshcorr:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Update `install.ps1` so that it supports installation of Dbatools on Linux and Mac.
While I assume most people use PSGallery to install the module and update. The cmdlet `Update-DbaTools` relies upon this script. If the module is installed via gallery then this script is already cross plat, but if for some reason `-Dev` switch then it would fail. 

Don't think this is a breaking change, but do think @potatoqualitee should be aware since it is an entry point to the module for some users. 

### Approach

Add checks for code not supported on Unix
Add dynamic check for user home path in PSModulePath

### Commands to test

`Update-Dbatools -Verbose -Dev`
or on a system with Dbatools not installed from the gallery (but running in memory)
`Update-Dbatools -Verbose` 

### Screenshots

Install on Windows
image

Install on Linux Container
image

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,9 +1,9 @@
 <!-- Below information IS REQUIRED with every PR -->
 ## Type of Change
 <!-- What type of change does your code introduce -->
- - [ ] Bug fix (non-breaking change, fixes #<enter issue number>)
- - [ ] New feature (non-breaking change, adds functionality)
- - [ ] Breaking change (effects multiple commands or functionality)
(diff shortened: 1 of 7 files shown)


wsmelton:
@joshcorr you will have to re-sync your branch since you are changing the psm1 file in each PR to resolve the conflict.

joshcorr:
ugh, sorry for the git sprawl.

wsmelton:
We can't merge this one with a rebase. It can cause issues for us because we don't rebase master.

If you can just create a new branch with the changes and then PR it again.

PR #6146, opened 2019-10-20
Resubmit: Update-Dbatools - Cross Platform Support Added

joshcorr:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Update `install.ps1` so that it supports installation of Dbatools on Linux and Mac.
While I assume most people use PSGallery to install the module and update. The cmdlet `Update-DbaTools` relies upon this script. If the module is installed via gallery then this script is already cross plat, but if for some reason `-Dev` switch then it would fail. 
Resubmitted per #6144 

Don't think this is a breaking change, but do think @potatoqualitee should be aware since it is an entry point to the module for some users. 

### Approach

Add checks for code not supported on Unix
Add dynamic check for user home path in PSModulePath

### Commands to test

`Update-Dbatools -Verbose -Dev`
or on a system with Dbatools not installed from the gallery (but running in memory)
`Update-Dbatools -Verbose` 

### Screenshots

Install on Windows
image

Install on Linux Container
image

### DIFF
--- a/dbatools.psm1
+++ b/dbatools.psm1
@@ -740,6 +740,7 @@ $script:xplat = @(
     'Export-DbaServerRole',
     'Get-DbaBuildReference',
     'Install-DbaWhoIsActive',
+    'Update-Dbatools',
     'Add-DbaServerRoleMember'
 )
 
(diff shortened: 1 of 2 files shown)


potatoqualitee:
acceptable change, thank you! anad thanks for all of your work with getting more commands on linux! 🎉

PR #6169, opened 2019-10-30
*-DbaDiagnosticQuery - Cross Platform Support Added

joshcorr:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose

Provide the ability to `Save-` and `Invoke-` the scripts created by Glenn Barry on Linux.
Since `Invoke-TLSWebRequest` is really just a wrapper for `Invoke-WebRequest` (Which works differently on core regardless of OS), this PR addresses some of those differences.
This is another chip away at #4647 

### Approach

Replace use of `InnerText` with `OuterHTML` and some regex since `InnerText` isn't supported on core. (See https://github.com/PowerShell/PowerShell/issues/2867)

### Commands to test

`Save-DbaDiagnosticQueryScript -Path /my/path/here`

### Screenshots

On Linux
image

Same Code on Windows
image
image

### DIFF
--- a/functions/Save-DbaDiagnosticQueryScript.ps1
+++ b/functions/Save-DbaDiagnosticQueryScript.ps1
@@ -72,16 +72,17 @@ function Save-DbaDiagnosticQueryScript {
     $glenberrysql = @()
     $RssPostFilter = "SQL Server Diagnostic Information Queries for*"
     $DropboxLinkFilter = "*dropbox.com*"
-    $LinkTitleFilter = "*Diagnostic*"
+    $LinkTitleFilter = "*Diagnostic Information Queries*"
+    $ExcludeSpreadsheet = "*BlankResutsSpreadsheet*"
 
     foreach ($post in $Feed.item) {
         if ($post.title -like $RssPostFilter) {
             # We found the first post that matches it, lets go visit and scrape.
             $page = Get-WebData -uri $post.link
-            $glenberrysql += ($page.Links | Where-Object { $_.href -like $DropboxLinkFilter -and $_.innerText -like $LinkTitleFilter } | ForEach-Object {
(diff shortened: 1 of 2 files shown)


functions/Save-DbaDiagnosticQueryScript.ps1
```diff
@@ -72,16 +72,17 @@ function Save-DbaDiagnosticQueryScript {
...
-    $LinkTitleFilter = "*Diagnostic*"
+    $LinkTitleFilter = "*Diagnostic Information Queries*"
+    $ExcludeSpreadsheet = "*BlankResutsSpreadsheet*"
 
     foreach ($post in $Feed.item) {
         if ($post.title -like $RssPostFilter) {
             # We found the first post that matches it, lets go visit and scrape.
             $page = Get-WebData -uri $post.link
-            $glenberrysql += ($page.Links | Where-Object { $_.href -like $DropboxLinkFilter -and $_.innerText -like $LinkTitleFilter } | ForEach-Object {
+            $glenberrysql += ($page.Links | Where-Object { $_.href -like $DropboxLinkFilter -and $_.outerHTML -like $LinkTitleFilter -and $_.outerHTML -notlike $ExcludeSpreadsheet } | ForEach-Object {
```
wsmelton: Why did this have to change got both Windows and Linux? It worked previously on Windows didn't it?

joshcorr: It did, I was making to the same code worked on both platforms without having to get into writing `if/else`. The previous code also overlooked the blank Excel files that I believe he is now including. This is something that didn't work in Core on windows as well. So it is more of a Core vs WindowsPS issue.

wsmelton: Yeah I think the iwr changed in core...

PR #6181, opened 2019-11-05
Implementing paths

FriedrichWeinmann:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Purpose
These new commands introduce better path resolution for common paths usually covered under environment variables ($env:temp, $env:AppData, $env:LocalAppdata, $env:ProgramData). They should find viable paths cross-platforms.

They also allow _redirecting_ those paths at the user's convenience. For example when the default paths are unsuitable due to permissions reasons.

At the same time, they are a foundation to be able to easily create more, named paths utilizing the same tools.

### Notes
- All internal use was corrected / switched over.
- Take care not to use this in scriptblocks that are invoked remotely, as the remote system might not have dbatools


wsmelton:
This seems to overcomplicate things to me. Why would we not use environment variables that are in PowerShell? These all work the same on any OS?

FriedrichWeinmann:
The paths should work on all relevant OSes.
The environment variables _will_ be picked up if available, but no longer solely dependent on that.
However this implementation also gives the user the ability to explicitly redirect it using the configuration system (while offering a user-friendly front-end).
Chrissy requested this due to frequent issues when running dbatools from the SQL Server Agent.

wsmelton:
This needs to be added to the list of things to document for contributors if we implement it. Otherwise we will see folks just use env on the code changes.

FriedrichWeinmann:
Probably even better to include tests to catch "unauthorized" $env: usage.
Which I'll throw in this afternoon to make the tests happy (see currently failed tests). 0:30am, so not updating this right now anymore :)

potatoqualitee:
agreed, that'd be great. one thing tho is that we have to ensure that we can still use $env within a Scriptblock in Invoke-Command/2 since the dbatools command will not be available remotely.

wsmelton:
Need to add `Save-DbaDiagnosticQueryScript` to this PR if you can Fred. It uses the MyDocuments folder path.

potatoqualitee:
heck yah, these are failures i can fix 👍

potatoqualitee:
fake failure 👍  merging, thanks, fred!

PR #6198, opened 2019-11-14
fixing import error on SMO already available

FriedrichWeinmann:
# Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Changes

Changed the decision, when dbatools will import the SQL libraries during import.
Previously checked for SMO being loaded, colliding with SqlSever and other script samples out there.
The new check may under rare circumstances try loading the libraries when there should be no need, merely costing a little import times.

It should be the lesser risk and have the lesser impact


Jaykul:
Hehe. This looks good ...

wsmelton:
We still need this to error because you can end up with versions of SMO that we have not tested against.

PR #6251, opened 2019-12-17
AvailabilityGroup backup history added

Stuart-Moore:
## Type of Change

 - [x] Pester test is included
 - [x] If new file reference added for test, has is been added to github.com/sqlcollaborative/appveyor-lab ?
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Have Get-DbaDbBackupHistory play nice with AG backup history spread across nodes

### Approach
Little bit of recursion

Added another property to the BackupHistory object: AvailabilityGroupName
Select-DbaBackupInformation will now default to using AvailabilityGroupName if it's populated

If the SqlInstance is an AG listener then fire off a gddbh query to each node. For last,
 pipe the results through Select-DbaBackupInformation to filter down to the full chain

If an instance is specified and the requested dbs are in an AG, then it will query all nodes in the AG for that database, but only use the instance for the other dbs.

Removed ParameterSet as it doesn't appear to do anything anymore.


potatoqualitee:
ohhh @wsmelton - i think you've been wanting this for a while!

Stuart-Moore:
Be good to get @wsmelton to give it a thorough test. I'm sure he's got some more 'interesting' AG configurations than I've got to hand. Logic says it'll be fine, but reality's a harsh master

functions/Get-DbaDbBackupHistory.ps1
```diff
@@ -259,6 +308,26 @@ function Get-DbaDbBackupHistory {
...
+                    Write-Message -Level Verbose -Message "Fetching history from replicas for db $($adb.name)"
+                    if ($adb.GetType().name -ne 'Database') {
+                        $adb = Get-DbaDatabase -SqlInstance $server -Database $adb.name
+                    }
+                    $AvailabilityGroupBase = $adb.parent.AvailabilityGroups[$adb.AvailabilityGroupName]
+                    $null = $PSBoundParameters.Remove('SqlInstance')
+                    $null = $PSBoundParameters.Remove('Database')
+                    $AgLoopResults = Get-DbaDbBackupHistory -SqlInstance $AvailabilityGroupBase.AvailabilityGroupListeners.name -database $adb.Name @PSBoundParameters
+                    $AvailabilityGroupName = $AvailabilityGroupBase.name
+                    $AgLoopResults.Foreach{ $_.AvailabilityGroupName = $AvailabilityGroupName }
```
wsmelton: This is not compatible with PS3, the foreach method was only added in PS4.

Stuart-Moore: thanks for spotting, I'll get that changed

Stuart-Moore: pushed up fixed version

Stuart-Moore:
Failure appears to be nthing to do with this. Everyone happy if I merge?

wsmelton:
Ok. So you might be able to answer this scenario without me having to dig through code...or see what was changed.

I have 3 node AG setup, the 3rd async replica is located at a DR site and is a named instance, but also because of security requires a custom port (not using 1433). I know in SMO that SQL Server class for AvailabilityGroup knows nothing of this port and only has the server name. When I pass in the listener in order to pull that backup history how is your code going to connect to that 3rd instance?

Now mind you I completely understand if it will break because we have the same issue in the DSC module for SQL Server and AG builds. If you can lay out how it connects to the other replicas to pull backup history I'm good with merging....

potatoqualitee:
can we merge? 🐶 👀

wsmelton:
It is fine with me to merge, just as long as folks understand it will have bugs that we cannot fix/resolve due to limitations in SMO....which I think should be rare cases.

potatoqualitee:
Thank you 💯

PR #6257, opened 2019-12-19
Get-DbaUserPermission - Fix path for cross-platform support

ggsince88:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Add cross-platform support 

### Approach
Using standard `Join-Path` instead of `\` which fails on macOS and Linux.

### DIFF
--- a/functions/Get-DbaUserPermission.ps1
+++ b/functions/Get-DbaUserPermission.ps1
@@ -215,7 +215,8 @@ function Get-DbaUserPermission {
                     Stop-Function -Message "The database $db is not accessible" -Continue
                 }
 
-                $sql = [System.IO.File]::ReadAllText("$script:PSModuleRoot\bin\stig.sql")
+                $sqlFile = Join-Path -Path $script:PSModuleRoot -ChildPath "bin" -AdditionalChildPath "stig.sql"
+                $sql = [System.IO.File]::ReadAllText("$sqlFile")
                 $sql = $sql.Replace("<TARGETDB>", $db.Name)
 
                 #Create objects in active database


potatoqualitee:
heyy thank you for the fix @ggsince88! let me take a closer look at the code

potatoqualitee:
Unfortunately, it looks like `AdditionalChildPath` is not supported in all operating systems. Perhaps explore using Resolve-Path instead?

functions/Get-DbaUserPermission.ps1
```diff
@@ -215,7 +215,8 @@ function Get-DbaUserPermission {
                     Stop-Function -Message "The database $db is not accessible" -Continue
                 }
 
-                $sql = [System.IO.File]::ReadAllText("$script:PSModuleRoot\bin\stig.sql")
+                $sqlFile = Join-Path -Path $script:PSModuleRoot -ChildPath "bin" -AdditionalChildPath "stig.sql"
```
wsmelton: Ah, yeah approved to quickly.. should be changed to:
```suggestion
                $sqlFile = Join-Path -Path $script:PSModuleRoot -ChildPath "bin\stig.sql"
```

wsmelton:
Yeah, the AdditionalChildPath is only on PS 6+

functions/Get-DbaUserPermission.ps1
potatoqualitee: the PR was created beecause \ will not work with Linux but it can be potentially `Resolve-Path (Join-Path -Path $script:PSModuleRoot -ChildPath "bin\stig.sql")`

``

wsmelton: Yes, `Join-Path` handles converting the slash based on the OS for us.

potatoqualitee: Yesss! Just tried it and did not know that. I thought it was just Resolve-path.

potatoqualitee:
Yay tests! Wouldn't have known otherwise 🎈

potatoqualitee:
gonna merge then add your changes, @wsmelton

PR #6277, opened 2020-01-03
create Get-DbaHideInstance to see if Hide Instance setting

### LINKED ISSUE #6276: Get-DbaHideInstance
### Summary of new feature

- There are commands to enable and disable this feature but no command to check the status before doing so.  Plus we writing a test in dbachecks for CIS and this would give dbachecks a way to check the setting without having to put a lot of code into dbachecks.

* Clear/concise description of what the problem is that the new feature can solve.

- Use the same code in either Disable-DbaHideInstance or Enable-DbaHideInstance except don't change the values, just display them.

* 1.0.79

tboggiano:
## Type of Change

 - [X] New feature (non-breaking change, adds functionality, fixes #6276 
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
To able to view the Hide Instance property for an instance.

### Approach
Same method as ForcedEncryption some magic around reading registry keys using wmi.

### Commands to test
Examples in the help

### Screenshots
image

### DIFF
--- a/functions/Get-DbaHideInstance.ps1
+++ b/functions/Get-DbaHideInstance.ps1
@@ -57,8 +57,7 @@ function Get-DbaHideInstance {
                 Write-Message -Level Verbose -Message "Resolving hostname."
                 $resolved = $null
                 $resolved = Resolve-DbaNetworkName -ComputerName $instance -Credential $Credential -EnableException
-            }
-            catch {
+            } catch {
                 $resolved = Resolve-DbaNetworkName -ComputerName $instance -Credential $Credential -Turbo
             }
 
@@ -70,8 +69,7 @@ function Get-DbaHideInstance {
                 $sqlwmi = Invoke-ManagedComputerCommand -ComputerName $resolved.FullComputerName -ScriptBlock {
                     $wmi.Services
                 } -Credential $Credential -ErrorAction Stop | Where-Object DisplayName -eq "SQL Server ($($instance.InstanceName))"
-            }
-            catch {
+            } catch {
                 Stop-Function -Message "Failed to access $instance" -Target $instance -Continue -ErrorRecord $_
             }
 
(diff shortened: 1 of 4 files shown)


functions/Get-DbaHideInstance.ps1
```diff
@@ -0,0 +1,140 @@
...
+    .PARAMETER Credential
+        Allows you to login to the computer (not sql instance) using alternative Windows credentials
+
+    .PARAMETER EnableException
+        By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
+        This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
+        Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
+
+    .NOTES
+        Tags:   HideInstance, Security
```
wsmelton: Just tag as `Instance, Security`

Will make it more discoverable.

functions/Get-DbaHideInstance.ps1
```diff
@@ -0,0 +1,140 @@
...
+
+            if ([System.String]::IsNullOrEmpty($regroot)) {
+                $regroot = $sqlwmi.AdvancedProperties | Where-Object {
+                    $_ -match 'REGROOT'
+                }
+                $vsname = $sqlwmi.AdvancedProperties | Where-Object {
+                    $_ -match 'VSNAME'
+                }
+
+                if (![System.String]::IsNullOrEmpty($regroot)) {
```
wsmelton: Please `-not` here instead of `!` for easier readability.

Not sure if the other commands may have done this or not...not something I was able to catch prior if so.

potatoqualitee: I also prefer `-not` and remove the ! when possible 💯

functions/Get-DbaHideInstance.ps1
tboggiano: The tag in the other commends Enable and Disable is HideInstance, so I thought for consistency it would be best to have the same, I could be wrong.

wsmelton: Ah, yeah sorry. I am going to be cleaning all those up, just have not gotten to it.

If you want to make them all consistent in this PR that would be awesome!

functions/Get-DbaHideInstance.ps1
tboggiano: done.

functions/Get-DbaHideInstance.ps1
tboggiano: Ok, all three are the same, I like consistency.

PR #6279, opened 2020-01-06
Module - Fix Unable to find type ProtectedData

### LINKED ISSUE #6278: [Bug] - Unable to find type [System.Security.Cryptography.ProtectedData].
i have a second system reporting this issue. will fix with 

`Add-Type -AssemblyName System.Security`

per https://stackoverflow.com/questions/48797700/powershell-date-type-unable-to-find

potatoqualitee:
fixes #6278

### DIFF
--- a/dbatools.psm1
+++ b/dbatools.psm1
@@ -5,7 +5,6 @@ param(
     $Option = @{ }
 )
 
-
 $start = [DateTime]::Now
 
 if (($PSVersionTable.PSVersion.Major -lt 6) -or ($PSVersionTable.Platform -and $PSVersionTable.Platform -eq 'Win32NT')) {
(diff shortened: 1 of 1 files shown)


wsmelton:
Would it be better to do this just in the command that needs it instead of module wide?

potatoqualitee:
i believe it's a module-wide problem. i can't even load the module on the servers that have this issue.

wsmelton:
Should be coming from this internal function:

https://github.com/sqlcollaborative/dbatools/blob/cd7a897cf042666e3bdbd18b6fdefb77dd4991fd/internal/functions/Get-DecryptedObject.ps1

potatoqualitee:
I ran into the issue when running a block that only included Get-DbaRegisteredServer, Invoke-DbaQuery and Write-DbaDbTableData

potatoqualitee:
that command also executes `Add-Type -AssemblyName System.Security` explicitly so it appears to be something else

PR #6280, opened 2020-01-06
New-DbaClientAlias remove array, make servename singular

potatoqualitee:


wsmelton:
This parameter is not even used in this function the way I am reading it, line 81 is commented out.

potatoqualitee:
@wsmelton its used in $serverstring which is used as arg3

wsmelton:
If that is the case then there is some cleanup to be done because we are still passing `$ServerName` as `$arg[0]` and should not be.

https://github.com/sqlcollaborative/dbatools/blob/b8659a8aadfb04d68231df1c9fba8b1cfa7749c9/functions/New-DbaClientAlias.ps1#L79-L83

https://github.com/sqlcollaborative/dbatools/blob/b8659a8aadfb04d68231df1c9fba8b1cfa7749c9/functions/New-DbaClientAlias.ps1#L133

PR #6289, opened 2020-01-09
Return the domain of the user instead of the domain of the group

### LINKED ISSUE #6286: Find-DbaLoginInGroup putting wrong domain name with login
We have a scenario where we have users in 2 different domains and the older users have 2 different SamAccountName values. For example in one domain they are FirstName.LastName and in the other domain they are FirstIntialLastName. (domain1\john.smith & domain2\jsmith)

It looks like it is because _$memberDomain = $ads.Name_ doesn't rotate so it using the wrong domain and ends up looking like 
Domain1\john.smith
Domain1\jsmith

awickham10:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #6286 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Fix issue #6286.

### Approach

Return the domain of the group member instead of the domain of the group.

### Commands to test

Hard to test because a multi-domain Active Directory is required. The change was tested by the issue reporter.

### Screenshots

N/A

### Learning

N/A

### DIFF
--- a/functions/Find-DbaLoginInGroup.ps1
+++ b/functions/Find-DbaLoginInGroup.ps1
@@ -81,7 +81,7 @@ function Find-DbaLoginInGroup {
                     $group = [System.DirectoryServices.AccountManagement.GroupPrincipal]::FindByIdentity($ads, $groupName);
                     $subgroups = @()
                     foreach ($member in $group.Members) {
-                        $memberDomain = $ads.Name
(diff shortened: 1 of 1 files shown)


potatoqualitee:
awesome, thank you Andrew!

PR #6346, opened 2020-01-31
Restore-DbaDatabase. Set initial catalog to master

bajmccormack:
$RestoreInstance is now set using Connect-DbaInstance instead of
Connect-Sqlinstance. This change means that instead of connecting
to the inital catalogue based on your SQL login, you will always
connect to master before restoring the database. This is needed
in case your default connection puts you in the context of the
database you are trying to restore.

## Type of Change

 - [ Yes ] Bug fix (non-breaking change, fixes # )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 

So that if restoring a database which is the same as the one you are automatically connected to, the restore will no longer fail. It's not a common scenario but it happened to me today where my initial catalog was the same as the DB I wanted to restore.

### Approach

It sets the database context to master for all connections.

### Commands to test

Get-ChildItem '\\BACKUPS\DBName1\' | Restore-DbaDatabase -SqlInstance 'DEVSQL1' -DatabaseName 'DBName1' -UseDestinationDefaultDirectories -WithReplace

### DIFF
--- a/functions/Restore-DbaDatabase.ps1
+++ b/functions/Restore-DbaDatabase.ps1
@@ -381,7 +381,7 @@ function Restore-DbaDatabase {
 
         #region Validation
         try {
-            $RestoreInstance = Connect-DbaInstance -SqlInstance $SqlInstance -SqlCredential $SqlCredential -Database Master
+            $RestoreInstance = Connect-SqlInstance -SqlInstance $SqlInstance -SqlCredential $SqlCredential -Database master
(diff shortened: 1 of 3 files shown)


potatoqualitee:
Thanks so much BJ! Appreciate the PR. Great points - I will merge after making a slight change.

PR #6348, opened 2020-02-03
Find-DbaInstance - not finding on custom port

FriedrichWeinmann:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Changes

Due to an error in string comparisons on the data type, the SqlInstance property would not display any values on results where no instance name was found (e.g. setting ScanType to TCPPort only).

This would prevent Find-DbaInstance from finding the instance when no other kind of instance was found on the targeted server.

Providing a default value of empty string to the field on the object resolved the issue (also improved some comparison performance, but on a very minor level)


potatoqualitee:
Thank you for the prompt PR 🏆 So far, we have one fail but it is unrelated.

potatoqualitee:
💯 much appreciated

PR #6374, opened 2020-02-22
Added optional User parameter to Set-DbaPrivilege

ben-thul:
* In order to prepare for changing a service account, permissions should
be able to be set for the new account ahead of the change in order to
minimize service restarts. Adding an optional parameter for the user to
set allows for this possibility. If provided, it skips the remote
service enumeration and attempts to set the requested permissions for
the passed in user.

* Fixed what appeared to be a bug where... none of the permissions
were to be set.

## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Added optional User parameter to Set-DbaPrivilege

### Approach
Check if new parameter is provided. If so, populate the list of accounts to add permissions to with that user (rather than through service enumeration).

### Commands to test
Set-DbaPrivilege

### DIFF
--- a/functions/Set-DbaPrivilege.ps1
+++ b/functions/Set-DbaPrivilege.ps1
@@ -91,8 +91,7 @@ function Convert-UserNameToSID ([string] `$Acc ) {
                         $SQLServiceAccounts = @();
                         if (Test-Bound 'User') {
                             $SQLServiceAccounts += $User;
-                        }
-                        else {
+                        } else {
                             Write-Message -Level Verbose -Message "Getting SQL Service Accounts on $computer"
                             $SQLServiceAccounts += (Get-DbaService -ComputerName $computer -Type Engine).StartName
                         }
@@ -103,7 +102,7 @@ function Convert-UserNameToSID ([string] `$Acc ) {
                                 param ($ResolveAccountToSID,
                                     $SQLServiceAccounts,
                                     $Type
-                                    )
+                                )
                                 . ([ScriptBlock]::Create($ResolveAccountToSID))
                                 $temp = ([System.IO.Path]::GetTempPath()).TrimEnd("");
                                 $tempfile = "$temp\secpolByDbatools.cfg"
--- a/functions/Set-DbaPrivilege.ps1
+++ b/functions/Set-DbaPrivilege.ps1
@@ -60,7 +60,8 @@ function Set-DbaPrivilege {
         [Parameter(Mandatory)]
         [ValidateSet('IFI', 'LPIM', 'BatchLogon', 'SecAudit')]
         [string[]]$Type,
-        [switch]$EnableException
+        [switch]$EnableException,
+        [string]$User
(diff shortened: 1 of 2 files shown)


functions/Set-DbaPrivilege.ps1
```diff
@@ -83,17 +87,23 @@ function Convert-UserNameToSID ([string] `$Acc ) {
                         Invoke-Command2 -Raw -ComputerName $computer -Credential $Credential -ScriptBlock {
                             $temp = ([System.IO.Path]::GetTempPath()).TrimEnd(""); secedit /export /cfg $temp\secpolByDbatools.cfg > $NULL;
                         }
-                        Write-Message -Level Verbose -Message "Getting SQL Service Accounts on $computer"
-                        $SQLServiceAccounts = (Get-DbaService -ComputerName $computer -Type Engine).StartName
+
+                        $SQLServiceAccounts = @();
+                        if ([string]::IsNullOrEmpty($User) -eq $false) {
```
wsmelton: Please use Test-Bound for validation of parameter being passed.

wsmelton:
Re-reading this a bit now...

Why would you need to assign any user outside of the SQL Server service account to any of these permissions: `'IFI', 'LPIM', 'BatchLogon', 'SecAudit'` ?

wsmelton:
The test failures are due to the parameter being added, the test has to be updated as well. Then the formatting is off somehow, so you will need to run the format command noted in the test output:

> It Set-DbaPrivilege.ps1 is adopting OTSB formatting style. Please run Invoke-DbatoolsFormatter against the failing file and commit the changes.

ben-thul:
In re: your comment as to "why?", I want to set those permissions on an account that will become (but is not currently) the service account. My specific use case right now is I'm changing the service account on the services across the environment. By allowing a user to be passed into this cmdlet, it allows me to set the required permissions for that account without first having to change the account and *then* give it the required privileges (thus preventing at least one service restart).

potatoqualitee:
Thank you very much 🙇 !

potatoqualitee:
merged! will publish momentarily

wsmelton:
The permissions are automatically assigned to the account you set if you use SqlWmi or SSCM... It has been that way since 2005... There is no reason to even run the command in your use case.

ben-thul:
> 
> 
> The permissions are automatically assigned to the account you set if you use SqlWmi or SSCM... It has been that way since 2005... There is no reason to even run the command in your use case.
> 
> Only exception would be IFI, but that shouldn't require a restart to take effect.

I'm not sure if you're saying you don't see a need for this _change_ or this _cmdlet_.

Either way, I'm going through my environment and changing the service account using `Update-DbaServiceAccount` and, anecdotally, none of the required permissions were assigned to the new service account. The change to this cmdlet was informed by that observation. Happy to learn something though.

PR #6383, opened 2020-02-26
Improve datamasking performance

sanderstad:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
When dealing with larger data sets, the masking command would either use a lot of memory or crash.
Also the speed of the updates drastically decreases when dealing with more than a million rows.

### Approach
The command now works in batches. The update statement will be separated in batches and executed on the destination database.

Prior to the masking the command check if the table contains an identity column.
If it has an identity column, it will create a nonclustered index on that column. 
If it doesn't have an identity column, the table will be altered by adding a column called "MaskingID" containing an identity.
Afterwards the command will remove both the index and if needed change the table back by removing the identity column.

### Commands to test
Invoke-DbaDbDataMasking

### Screenshots
image

image


potatoqualitee:
hell yeah 😍

PR #6395, opened 2020-03-02
Added filter of system files added in SQL 2019 install

gbargsley:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
With the SQL 2019 installation the following four files are now being dropped in the DATA folder by default.

model_msdbdata.mdf
model_msdblog.ldf
model_replicatedmaster.ldf
model_replicatedmaster.mdf

The current version of the function identifies these as orphaned because they are not hosted by SQL Server, but I believe they are there for AG replication of system files or for some other SQL related feature.  I would not recommend displaying these as orphaned as deleting them might break the SQL 2019 instance.

### Approach

Added them as system files which then removes them from the output of the function.

### Commands to test

run Find-DbaOrphanedFiles against a SQL 2019 instance 

### Screenshots

Before: (ver 0.1.100)
image

After fix from PR:
image

### DIFF
--- a/functions/Find-DbaOrphanedFile.ps1
+++ b/functions/Find-DbaOrphanedFile.ps1
@@ -168,7 +168,7 @@ function Find-DbaOrphanedFile {
         }
 
         $FileType += "mdf", "ldf", "ndf"
-        $systemfiles = "distmdl.ldf", "distmdl.mdf", "mssqlsystemresource.ldf", "mssqlsystemresource.mdf"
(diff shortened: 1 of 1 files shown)


wsmelton:
This needs an issue to know the background of those files and how you determined SQL Server 2019 installation is putting them there on the server.

gbargsley:
#6396

gbargsley:
Yes on multiple boxes.  I cannot find it in any MS docs, but it seems related to a feature.

potatoqualitee:
Interesting! I was able to find some docs, though it wasn't explicitly about the system dbs, just talks about them in passing

https://docs.microsoft.com/en-us/sql/big-data-cluster/deployment-high-availability?view=sql-server-ver15

Thanks much, great work, merging 💯

wsmelton:
So an additional change is required then to also add `containedag_master` and `containedag_msdb` as documented in that link.

PR #6401, opened 2020-03-05
Add ExcludeCategory param to Get-DbaAgentJob

ben-thul:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 

Add an ExcludeCategory param to Get-DbaAgentJob

### Approach

Mimic'd existing implementation of ExcludeJob param in the same cmdlet

### Commands to test

Get-DbaAgentJob

### DIFF
--- a/functions/Get-DbaAgentJob.ps1
+++ b/functions/Get-DbaAgentJob.ps1
@@ -31,6 +31,9 @@ function Get-DbaAgentJob {
     .PARAMETER Category
         Return jobs associated with specific category
 
+    .PARAMETER ExcludeCategory
+        Categories to exclude - jobs associated with these categories will not be returned.
+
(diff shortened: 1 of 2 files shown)


wsmelton:
@ben-thul you will need to update the test associated with this command as well so the test will pass.

Should have a unit test that includes list of the parameters, we validate each command's parameter list. If you can also add an integration test would be nice, if you are up to it but not required.

ben-thul:
Happy to do it.

In the interest of teaching one to fish, can you provide some guidance on how to run these locally? In the future, I'd like to run these before submitting the PR so there's at least a chance of not having the CI tests fail. Thanks in advance.

PR #6442, opened 2020-03-25
Add optional Name parameter to Get-DbaAgentAlert

ben-thul:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Added an optional Name parameter to the Get-DbaAgentAlert cmdlet to allow for fetching just some of the alerts (in addition to the current behavior of "get all of them").

### Commands to test
Added an example to the documentation for Get-DbaAgentAlert


functions/Get-DbaAgentAlert.ps1
```diff
@@ -49,6 +57,7 @@ function Get-DbaAgentAlert {
         [DbaInstanceParameter[]]$SqlInstance,
         [PSCredential]
         $SqlCredential,
+        [string[]]$Name,
```
wsmelton: This should be called `Alert` as it is the name of the SMO object we are returning. Just like with `Get-DbaAgentSchedule` and `Get-DbaAgentOperator` they are named `Schedule` and `Operator` respectfully.

As well, an additional parameter for `ExcludeAlert` should be added to return all but the provided list as you have done in other command(s).

ben-thul: Sounds good to me. I've done all of that and while I was at it also added support for regexes on both include and exclude.

wsmelton: The pattern support can be included within each original Alert and ExcludeAlert. We do not need additional parameters for that, causes more confusion on user side. 

I would just update the help to state that each parameter also accepts wildcards or regex patterns.

functions/Get-DbaAgentAlert.ps1
```diff
@@ -71,17 +84,26 @@ function Get-DbaAgentAlert {
...
 
-            foreach ($alert in $alerts) {
-                $lastraised = [dbadatetime]$alert.LastOccurrenceDate
+            if (Test-Bound 'Alert') {
+                $alerts = $alerts | where Name -like $Alert;
+            }
+
+
+            if (Test-Bound 'ExcludeAlert') {
+                $alerts = $alerts | where Name -notlike $ExcludeAlert;
```
wsmelton, line 93: It should handle both scenarios where they provided a wildcard or they only want one to return. Current state if I provided "Backup" it will return any job that has the term "backup" in the job name and not just the one I called "Backup".

wsmelton: We have another command that handles both scenarios in the same parameter, but I'm forgetting which one it is.

wsmelton: I don't want to break functionality of the current parameter. The other commands I think we just check for the `*`... `if ($Alert -match '`*') {`, or something of that nature.

ben-thul: Hmm... I tested it just now and it doesn't seem to work that way. On my local instance, I have several alerts that replication installed. If I run `Get-DbaAgentAlert -SqlInstance . -ExcludeAlert 'Replication'` I get all of the alerts (i.e. no filtering happened as none of the alerts are called `Replication`). Contrast that with if I run `Get-DbaAgentAlert -SqlInstance . -ExcludeAlert 'Replication*'` then those alerts are excluded as expected.

I'd expect it to work as you describe if the `-notmatch` operator were used instead of `-notlike`.

potatoqualitee:
How is this looking? Ready to merge? Thank you 🙇

ben-thul:
> 
> 
> How is this looking? Ready to merge? Thank you 🙇

Good question. It is unclear to me what is still outstanding with respect to this PR.

potatoqualitee:
Ah, k, @wsmelton - ready to merge? Thanks to you both 💯

wsmelton:
Can't merge yet. It is not working to be as it should... Previous comments and screenshots provided.

ben-thul:
I'm not seeing any screenshots in the PR. Where can I find them?

As to previous comments, perhaps the aforementioned screenshots shed some light on what you're trying to say. However, I just tested the change on my local instance and I'm not seeing the behavior that you describe in https://github.com/sqlcollaborative/dbatools/pull/6442#pullrequestreview-390050553.

wsmelton:
Ha, forgot to actually submit the review 😬

ben-thul:
Solid. That gives me something to address. I'll circle back on that and see what's up. Thanks!

functions/Get-DbaAgentAlert.ps1
```diff
@@ -71,17 +84,33 @@ function Get-DbaAgentAlert {
 
             $alerts = $server.Jobserver.Alerts
 
-            foreach ($alert in $alerts) {
-                $lastraised = [dbadatetime]$alert.LastOccurrenceDate
+            if (Test-Bound 'Alert') {
+                $tempAlerts = @();
+
+                foreach ($a in $Alert) {
+                    $tempAlerts += $alerts | where Name -like $a;
```
wsmelton, line 91: Instead of iterating over them twice...would it not be simpler to just but this filter in the output block?

Use of the `+=` adds memory usage if user is hitting a server that has a ton of alerting setup.

potatoqualitee: Hope you dont mind, I will merge as the fix works and passes tests (tests that failed are unrelaated)

wsmelton: Our test pass because we are not testing anything specific for this command. The integration test for this command does not do anything other than making sure the unfiltered output includes an alert we created in Appveyor instance.

potatoqualitee: ohh nooo

ben-thul: I'm happy to both add to the tests and write some better tests. But under separate cover (since this PR is about to have a two month birthday... they grow up so fast!).

Can either of you suggest a cmdlet that currently has decent tests that I can emulate?

wsmelton: It's already been merged so you have to do a new PR 😉

wsmelton: Likely `Get-DbaDatabase` is pretty solid test, may have more than what is needed for this but shows how to mock some internal commands (though should not be needed for this one). Main thing is just checking the code coverage to make sure at least some portion of code for the parameters is being executed. End result of a test won't necessarily get 100% coverage but the base functionality the command should have, we want to test for and make sure it returns good....especially areas of a command we have "special" logic.

ben-thul: Roger that on "separate PR". And thanks for pointing me in a direction. I'll see what I can do!

PR #6449, opened 2020-03-30
Tests failing - testing reversions

potatoqualitee:
github and appveyor were bombing, leading us to believe that some things were passing when they were not. I will revert to working copies until tests work again then notify the impacted parties. So far, i do think it's one file impacting multiple commands.


potatoqualitee:
sweet that worked - something else failed

PR #6475, opened 2020-04-14
Updates to Get-DbaService and Restart-DbaService to fix #6458

### LINKED ISSUE #6458: [Bug] Restart-DbaService fails when Polybase feature or Launchpad is installed
### Environmental information

```
#### PowerShell version:

PSVersion                      5.1.18362.628                                                                                                                                                  
BuildVersion                   10.0.18362.628                                                                                                                                                 




#### dbatools Module version:


Name    : dbatools
Path    : C:\Program Files\WindowsPowerShell\Modules\dbatools\1.0.104\dbatools.psd1
Version : 1.0.104

Name    : dbatools
Path    : C:\Program Files\WindowsPowerShell\Modules\dbatools\1.0.103\dbatools.psd1
Version : 1.0.103

Name    : dbatools
Path    : C:\Program Files\WindowsPowerShell\Modules\dbatools\1.0.91\dbatools.psd1
Version : 1.0.91




```

PS C:\WINDOWS\system32> Get-Culture

1031             de-DE            German (Germany) 

#### SQL Server:
Microsoft SQL Server 2017 (RTM-CU19) (KB4535007) - 14.0.3281.6 (X64)   Jan 23 2020 21:00:04   Copyright (C) 2017 Microsoft Corporation  Enterprise Edition (64-bit) on Windows Server 2016 Standard 10.0 <X64> (Build 14393: ) (Hypervisor) 

and

Microsoft SQL Server 2019 (RTM-CU3) (KB4538853) - 15.0.4023.6 (X64)   Mar  4 2020 00:59:26   Copyright (C) 2019 Microsoft Corporation  Enterprise Edition: Core-based Licensing (64-bit) on Windows Server 2019 Standard 10.0 <X64> (Build 17763: ) (Hypervisor) 

```sql
```

#### Host used

- [x ] powershell.exe
- [x ] ISE
- [x ] VS Code
- [ ] Other (please specify)

- [x ] Still buggy in powershell.exe

#### Errors Received

WARNING: [10:08:59][Update-ServiceStatus] (MSSQLSERVER on MYSERVER) The attempt to stop the service returned the
 following error: The service cannot be stopped because other services that are running are dependent on it.
WARNING: [10:08:59][Update-ServiceStatus] (MSSQLSERVER on MYSERVER) Run the command with '-Force' switch to
force the restart of a dependent SQL Agent

#### Steps to Reproduce

SQL Server 2017 with R intergration installed, e.g. Launchpad service is available
SQL Server 2019 with Polybase integration, e.g. SQLPBENGINE and SQLPBDMS

Import-Module dbatools
$target = "MyTarget"
Restart-DbaService -ComputerName $target -Type Engine   -force 

#### Expected Behavior

That the DB Egnine would restart with all dependant services

#### Actual Behavior

Engine does not restart, only the Agent. 
Warning that there are dependencies:
The attempt to stop the service returned the following error: The service cannot be stopped because other services that are running are dependent on it.

tkwj:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #6458  )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Fix the restart bug in #6458 and a couple other issues discovered in the process

### Approach
**Changes in Get-DbaService**

- Added Launchpad to list of values for parameter Type

- Got Launchpad's service type enum ID from
$mc = New-Object Microsoft.SqlServer.Management.Smo.Wmi.ManagedComputer
$mc.Services | Where-Object DisplayName -like "*Launchpad*"
image

- Added Launchpad to the section that handles instance name for per-instance type services
In doing this, also added PolyBase and FullText which are instance scoped, this will fix calls to Get-DbaService with -Instance provided (bugs uncovered as part of this effort)

**Changes in Restart-DbaService**
- Added Launchpad and PolyBase to list of values for parameter Type. This allows them to be 
Found that calls to Get-DbaService using the "Silent" parameter changed this to EnableException, as the Silent alias was removed awhile back
-In the section that adds SQL Agent to the restart collection when Engine is detected, I added the PolyBase and Launchpad types as well. Essentially the same code for Agent was kept but now that logic is done for each of "Agent", "PolyBase", "Launchpad" types

- The call into Get-DbaService was also changed to use -WarningAction SilentlyContinue. This is because otherwise Get-DbaService will produce a warning for each Launchpad / PolyBase that it searches for and doesn't find. This wasn't needed before because Agent is always there with Engine.
This is only in the section where dependent services are added. If someone does Restart-DbaService - Type Launchpad explicitly on a machine without it, they'll get the service not found warning as expected.

**Testing**
I manually tested this on:
SQL 2016 with R & PolyBase
SQL 2017 with R, Python, PolyBase
SQL 2019 with R, Python, PolyBase (both options included)
SQL 2014, just for sanity
used a mix of named instances and a default instance
used a variety of commands, including piping such as Get-DbaService -Type Launchpad | Restart-DbaService
ran manual.pester on *DbaService.Tests.ps1

### Commands to test
Prerequisites:
Have PolyBase or in-database R/Python services installed in SQL 2016 and above
Then Run:
```powershell
Restart-DbaService -Type Engine -Force
```

### Screenshots
image

### DIFF
--- a/functions/Restart-DbaService.ps1
+++ b/functions/Restart-DbaService.ps1
@@ -116,20 +116,25 @@ function Restart-DbaService {
         $processArray = [array]($processArray | Where-Object { (!$InstanceName -or $_.InstanceName -in $InstanceName) -and (!$Type -or $_.ServiceType -in $Type) })
         foreach ($service in $processArray) {
             if ($Force -and $service.ServiceType -eq 'Engine') {
+                $dependentServices = @()
                 foreach ($dependentService in @("Agent", "PolyBase", "Launchpad")) {
                     if (!($processArray | Where-Object { $_.ServiceType -eq $dependentService -and $_.InstanceName -eq $service.InstanceName -and $_.ComputerName -eq $service.ComputerName })) {
                         Write-Message -Level Verbose -Message "Adding $dependentService service to the list for service $($service.ServiceName) on $($service.ComputerName), since -Force has been specified"
-                        #Construct parameters to call Get-DbaService
-                        $serviceParams = @{
-                            ComputerName  = $service.ComputerName
-                            InstanceName  = $service.InstanceName
-                            Type          = $dependentService
-                            WarningAction = 'SilentlyContinue'
-                        }
-                        if ($Credential) { $serviceParams.Credential = $Credential }
-                        if ($EnableException) { $serviceParams.EnableException = $EnableException }
-                        $processArray += @(Get-DbaService @serviceParams)
+                        $dependentServices += $dependentService
                     }
+                }
+                if ($dependentServices.Count -gt 0) {
+                    #Construct parameters to call Get-DbaService
+                    $serviceParams = @{
+                        ComputerName  = $service.ComputerName
+                        InstanceName  = $service.InstanceName
+                        Type          = $dependentServices
+                        WarningAction = 'SilentlyContinue'
+
+                    }
+                    if ($Credential) { $serviceParams.Credential = $Credential }
+                    if ($EnableException) { $serviceParams.EnableException = $EnableException }
+                    $processArray += @(Get-DbaService @serviceParams)
                 }
             }
         }
--- a/functions/Restart-DbaService.ps1
+++ b/functions/Restart-DbaService.ps1
@@ -19,7 +19,7 @@ function Restart-DbaService {
 
     .PARAMETER Type
         Use -Type to collect only services of the desired SqlServiceType.
-        Can be one of the following: "Agent","Browser","Engine","FullText","SSAS","SSIS","SSRS"
+        Can be one of the following: "Agent", "Browser", "Engine", "FullText", "SSAS", "SSIS", "SSRS", "PolyBase", "Launchpad"
 
     .PARAMETER Timeout
         How long to wait for the start/stop request completion before moving on. Specify 0 to wait indefinitely.
@@ -87,7 +87,7 @@ function Restart-DbaService {
         [DbaInstanceParameter[]]$ComputerName = $env:COMPUTERNAME,
         [Alias("Instance")]
         [string[]]$InstanceName,
-        [ValidateSet("Agent", "Browser", "Engine", "FullText", "SSAS", "SSIS", "SSRS")]
+        [ValidateSet("Agent", "Browser", "Engine", "FullText", "SSAS", "SSIS", "SSRS", "PolyBase", "Launchpad")]
         [string[]]$Type,
         [parameter(ValueFromPipeline, Mandatory, ParameterSetName = "Service")]
         [Alias("ServiceCollection")]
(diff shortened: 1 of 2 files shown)


functions/Restart-DbaService.ps1
```diff
@@ -115,17 +115,22 @@ function Restart-DbaService {
...
         foreach ($service in $processArray) {
-            if ($Force -and $service.ServiceType -eq 'Engine' -and !($processArray | Where-Object { $_.ServiceType -eq 'Agent' -and $_.InstanceName -eq $service.InstanceName -and $_.ComputerName -eq $service.ComputerName })) {
-                Write-Message -Level Verbose -Message "Adding Agent service to the list for service $($service.ServiceName) on $($service.ComputerName), since -Force has been specified"
-                #Construct parameters to call Get-DbaService
-                $serviceParams = @{
-                    ComputerName = $service.ComputerName
-                    InstanceName = $service.InstanceName
-                    Type         = 'Agent'
+            if ($Force -and $service.ServiceType -eq 'Engine') {
+                foreach ($dependentService in @("Agent", "PolyBase", "Launchpad")) {
```
nvarscar, line 119: Get-DbaService supports returning multiple types at once, I'm pretty sure it would be faster to get all three types initially and then add missing services, as opposed to doing three requests - one for each service.

tkwj: @nvarscar Thanks for the review! I didn't consider the multiple types going in.
I have kept the loop, to prevent duplicates from being added to the $processArray, but it is now just building a collection of $dependentService. Now $dependentService is being used for the single call to Get-DbaService

functions/Restart-DbaService.ps1
```diff
@@ -115,17 +115,27 @@ function Restart-DbaService {
...
+                    $serviceParams = @{
+                        ComputerName  = $service.ComputerName
+                        InstanceName  = $service.InstanceName
+                        Type          = $dependentServices
+                        WarningAction = 'SilentlyContinue'
+
+                    }
+                    if ($Credential) { $serviceParams.Credential = $Credential }
+                    if ($EnableException) { $serviceParams.EnableException = $EnableException }
+                    $processArray += @(Get-DbaService @serviceParams)
```
wsmelton, line 137: Considering how this command can be used I would consider doing this in a different manner for performance. Using `+=` can build on memory usage for a session in PowerShell.

I'm aware this pattern is used throughout our module, and appears especially in this command; but arrays in PowerShell are immutable so this process basically duplicates that array each time making a larger array object to hold the additional elements. Not something we have to deal with in this PR right now, but something to consider going forward that we need to stop using `+=` in our code for this module.

wsmelton: Reference docs that speak on this as well: https://docs.microsoft.com/en-us/windows-server/administration/performance-tuning/powershell/script-authoring-considerations#array-addition

nvarscar: I always wondered how is it worth a trouble when it comes to using arrays with less than 10 members. I would understand the necessity if it was an array with 100+ items.
I agree that we should conform to best practices more though. And just like you said, our code is plagued with this particular syntax.

wsmelton: Well it is worth the trouble in that to build the objects for those 10 members only takes 1ms compared to 8ms because it was duplicating that array multiple times. Then you have that command run against 100 servers...it can make huge differences; particularly being that this command is likely used in build/deployment cycles frequently.

tkwj: Thanks for the info about array addition. I can definitely see the point, specially when things are at a large scale. I'll try not to use any additional `+=` when submitting future PRs!

potatoqualitee:
sweet, approved, thanks all! 🏆

PR #6490, opened 2020-04-26
Fix get dba locale setting

Splaxi:
## Type of Change

 - [X] Documentation
 
### Purpose
Fix / update example to match what parameters are allowed for the function / cmdlet.


wsmelton:
@Splaxi please just combine PRs when it comes to doc updates. We have stated this previously I recall but we do not require a single PR for every doc updates.

PR #6516, opened 2020-05-05
Update Export-DbaScript.ps1 - Invalid $scriptBatchTerminator condition

i7slegend:
Invalid $scriptBatchTerminator condition (always return false)

## Type of Change

 - [ ] New feature (non-breaking change)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system

### DIFF
--- a/functions/Export-DbaScript.ps1
+++ b/functions/Export-DbaScript.ps1
@@ -285,8 +285,10 @@ function Export-DbaScript {
                             } else {
                                 $ScriptingOptionsObject.FileName = $null
                                 foreach ($scriptpart in $scripter.EnumScript($object)) {
-                                    if ($scriptBatchTerminator) {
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Thanks so much, @i7slegend  - this seems reasonable. Can you tell how you found it and what output changed after you made the change? Also, does line 256 have to be changed as well?

i7slegend:
For line 288 only.
I wanted to `scriptData` for a table, but with `scriptBatchTerminator` it's not impossible cuz uses `Script` method (`scriptData` and `Script` method is not applicable together). But if i remove `scriptBatchTerminator` then doesn't generate batch seperators. I looked the source code and found this mistake.

potatoqualitee:
Works for me! Thanks again 🏆

PR #6546, opened 2020-05-20
Restart-DbaDatabase flow control fixes

FriedrichWeinmann:
# Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Change

Fixed flow control to not silently kill callstacks

### DIFF
--- a/functions/Restore-DbaDatabase.ps1
+++ b/functions/Restore-DbaDatabase.ps1
@@ -406,7 +406,7 @@ function Restore-DbaDatabase {
         try {
             $RestoreInstance = Connect-SqlInstance -SqlInstance $SqlInstance -SqlCredential $SqlCredential -Database master
         } catch {
-            Stop-Function -Message "Error occurred while establishing connection to $SqlInstance" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
+            Stop-Function -Message "Error occurred while establishing connection to $SqlInstance" -Category ConnectionError -ErrorRecord $_ -Target $instance
             return
         }
 
@@ -526,7 +526,9 @@ function Restore-DbaDatabase {
                 $bound['Path'] = $bh
                 Restore-DbaDatabase @bound
             }
-            break
+            # Flag function interrupt to silently not execute end
+            ${__dbatools_interrupt_function_78Q9VPrM6999g6zo24Qn83m09XF56InEn4hFrA8Fwhu5xJrs6r} = $true
+            return
         }
         if ($PSCmdlet.ParameterSetName -like "Restore*") {
             if ($PipeDatabaseName -eq $true) {


FriedrichWeinmann:
A bit additional context for those who didn't track the Twitter chat at the time (as well as some more thorough explanation than what you can do on Twitter anyway):

The command was silently failing, ending the calling script, no error, no nothing.

This happens when you call `break` or `continue` outside of a loop, as those two statements are designed to kill the entire callstack until they find a loop to break or continue with.

The complications arise, when you now look at dbatools' opt-in exception model. The safe way to end a function* without throwing an exception is to use `return` instead of one of the previously mentioned statements.
The Problem with that is, that it only ends the current block (begin/process/end) and then continues on with the next in that chain.
We deal with this by having Stop-Function write a variable into the caller scope, that later on is picked up by Test-FunctionInterrupt.
You can see that in the function Code at the beginning of process and end:

```powershell
```

Looking at the Change where this PR removes the break:
In this instance we want to silently end the function, but it is not an error, so we don't actually want to use Stop-Function, which would at least write a warning. Instead we explicitly write that (somewhat wonkyly named) variable and use return.

This works well enough, but now we have another scenario we need to deal with:
Let's say we have a function that accepts Input from the pipeline. It could receive any number of Inputs, and maybe only one of them is bad Input, the rest works well enough. Also: That input might be provided from pipeline or explicitly.

To deal with the from-pipeline vs. explicitly bound situation, we use a foreach loop in process.
Now if we want to use the same helper function `Stop-Function` for dealing with bad input - which with `-EnableException` _will_ lead to a terminating exception the caller can catch/try - we also need a mode, where it will not actually kill the entire command in situations, where an input item is bad and we want to just write a warning and continue with the next item.
So in that scenario, it must not write the flag that `Test-FunctionInterrupt` uses to kill the entire function.

Which is where the `-Continue` parameter comes in, telling it to not set that flag and to call `continue` instead, having it skip to the next item in that foreach loop.
This means, this should _never_ be used to kill the entire function (though it reliably will), because of this above-mentioned habit of continue killing the entire callstack until it finds a loop.

The change this PR applies to the begin block is where it was misused:
If we cannot connect to the SQL instance, no other success can be possible and we want to kill the function, so we need to use the ...

```powershell
Stop-Function -Message "Error occurred while establishing connection to $SqlInstance" -Category ConnectionError -ErrorRecord $_ -Target $instance
return
```

... notation.

> Note: We were unable to implement the return statement from within `Stop-Function` (as it only ends `Stop-Function`) so the return in the next line is necessary, while we can call `continue` for the use-case from within.

Using `-Continue` here meant it would silently kill everything if the caller was using it in interactive non-exception mode and the connection failed.
It would still write a warning though, so interactive users were not quite as confused.

PR #6576, opened 2020-06-08
Updated bad logic in Find-DbaAgentJob

gbargsley:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
There is a logic flow error in this command.  If you do not provide a job or step, it loads all jobs from the server.  Then in the parameter checks that follow it attempts to add Enabled, Failed, Not Scheduled jobs to that list.  Those jobs are already in the list since all jobs were loaded.  The += on those parameters was changed to just = so it only sets job list to those filtered.

### Approach

Adds only jobs that meet the parameter specified

### Commands to test

`Find-DbaAgentJob` -SqlInstance $servers -NoSchedule 
Find-DbaAgentJob -SqlInstance $servers -IsFailed
Find-DbaAgentJob -SqlInstance $servers -IsDisabled
`

### DIFF
--- a/functions/Find-DbaAgentJob.ps1
+++ b/functions/Find-DbaAgentJob.ps1
@@ -198,16 +198,16 @@ function Find-DbaAgentJob {
 
             if ($IsDisabled) {
                 Write-Message -Level Verbose -Message "Finding job/s that are disabled"
-                $output += $jobs | Where-Object IsEnabled -eq $false
+                $output = $jobs | Where-Object IsEnabled -eq $false
             }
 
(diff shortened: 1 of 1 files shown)


wsmelton:
Can you update the test to ensure this logic is validated?

gbargsley:
How do you tell what jobs are part of the tests?  I see there are already some checks in the test for the function. But I do not know how to know the status of the jobs deployed.

PR #6577, opened 2020-06-10
New feature - Clone login with a new name and DB permissions

### LINKED ISSUE #3864: New Feature - Clone login with all DB-level permissions included
---
name: New feature - clone a certain login including all the DB-level permissions :bomb:
about: Either a new function, or a new switch in New-DbaLogin to get permissions from each database of the source login and apply them to a new login.

---

### Summary of new feature

* Right now there is no certain way of cloning user permissions fully. Sync-DbaLoginPermissions only does so on the login level, using underlying function Update-SqlPermissions, that only covers database level permissions and roles; all the object-level permissions are ignored. There is Export-DbaDbUser, which, if used together with Export-DbaDbLogin and some regex magic, might do the trick. However, I would rather have that functionality built-in into module, as this is one of the most popular questions I get: create new login A with permissions like user B.
* Options to consider:
  * `Get-DbaLogin LoginA | New-DbaLogin LoginB -IncludeDBPermissions`
  * `Copy-DbaDbPrivileges -From LoginA -To LoginB`
  * `(Export-DbaUser -Login LoginA -IncludeLoginScript) -replace 'LoginA','LoginB'`

This will all come down to the following actions inside a function:
1. Get databases that source login has users in.
2. Get DB permissions script for each of the users in that list.
3. Add login creation part if necessary 
4. Replace names from LoginA to LoginB using regex.
5. Run the statements against the target server

* 0.9.382

nvarscar:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #3864 )
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Make login cloning a reality!

### Approach
When `-LoginRenameHashtable` parameter is specified, `Copy-DbaLogin` no longer tries to rename the newly created login. It creates a new login using `New-DbaLogin` function based on the old login object.
The new switch `-ObjectLevel` also clones permissions by calling `Export-DbaUser`, that now supports a new switch `-Template`, which replaces username and login name with templated strings, which can be replaced afterwards by new login/user names. When the switch is absent, the old behavior is preserved, taking care of database-level permissions and roles only.
I had to rename some variables because otherwise I was constantly confused what am I working with, but that, unfortunately, made the PR quite unreadable due to GitHub being annoying and not recognizing the changes. VSCode with GitLens highlights the changes just fine.

### Commands to test
```
Copy-DbaLogin -Source $script:instance1 -Login tester -Destination $script:instance2 -LoginRenameHashtable @{ tester = 'tester_new' } -ObjectLevel
```

### Learning
Reading own old code is confusing.


potatoqualitee:
Finalllllllllllllllllly! 🤩

functions/Copy-DbaLogin.ps1
```diff
@@ -51,9 +51,15 @@ function Copy-DbaLogin {
...
 
+    .PARAMETER NewSid
+        Ignore sids from the source login objects to generate new sids on the destination server. Useful when copying login onto the same server
+
     .PARAMETER LoginRenameHashtable
         Pass a hash table into this parameter to be passed into Rename-DbaLogin to update the Login and mappings after the Login is completed.
 
+    .PARAMETER ObjectLevel
+        Include object-level permissions for each user associated with copied login.
+
```
wsmelton, line 62: Please provide examples of use on the new parameters...

functions/Export-DbaUser.ps1
```diff
@@ -46,6 +46,9 @@ function Export-DbaUser {
     .PARAMETER Passthru
         Output script to console, useful with | clip
 
+    .PARAMETER Template
+        Script user as a templated string that contains tokens {templateUser} and {templateLogin} instead of username and login
+
```
wsmelton, line 51: Please provide an example of the parameter...

functions/New-DbaLogin.ps1
```diff
@@ -58,6 +58,9 @@ function New-DbaLogin {
     .PARAMETER Disabled
         Create the login in a disabled state
 
+    .PARAMETER DenyWindowsLogin
+        Create the login and deny Windows login ability
+
```
wsmelton, line 63: Please provide an example...

nvarscar:
Thanks for the review! Added examples and fixed param help a bit.

potatoqualitee:
thanks so much to you both 🙇 we'll have to scream about this one from the rooftops!

PR #6622, opened 2020-06-25
Grant-DbaAgPermission - Fix usage of $server to enable creation of login

### LINKED ISSUE #6613: [Bug] Grant-DbaAgPermission - creation of login failing (Cannot bind argument to parameter 'SqlInstance' because it is null)
This code throws the error:
```
$null = Grant-DbaAgPermission -SqlInstance $server -Login $serviceAccount -Type Endpoint -Permission Connect
WARNING: [14:52:40][Grant-DbaAgPermission] Failure | Cannot bind argument to parameter 'SqlInstance' because it is null.
```

Here is the responsible code (starting line 124):
```
            if ($perm -contains "CreateAnyDatabase") {
                try {
                    $server = Connect-SqlInstance -SqlInstance $instance -SqlCredential $sqlcredential
                } catch {
[...]
                }
[...]
            } elseif ($Login) {
                $InputObject += Get-DbaLogin -SqlInstance $instance -SqlCredential $SqlCredential -Login $Login
                foreach ($account in $Login) {
                    if ($account -notin $InputObject.Name) {
                        try {
                            $InputObject += New-DbaLogin -SqlInstance $server -Login $account -EnableException
                        } catch {
                            Stop-Function -Message "Failure" -ErrorRecord $_ -Target $instance
                            return
                        }
                    }
                }
            }
```

The failing commandlet is "New-DbaLogin -SqlInstance $server -Login $account -EnableException" because the variable $server is only set in the other branch of the outer if. It should use $instance instead, like the Get-DbaLogin some line above.

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #6613 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
I brought the call to Connect-SqlInstance one level up
to serve all following parts with the variable $server.

I changed "elseif" to "if" to create missing logins even
if permissions contain CreateAnyDatabase.

I changed the Stop-Function to continue after error
to harmonize these calls within the commandlet.

### DIFF
--- a/functions/Grant-DbaAgPermission.ps1
+++ b/functions/Grant-DbaAgPermission.ps1
@@ -121,30 +121,28 @@ function Grant-DbaAgPermission {
         }
 
         foreach ($instance in $SqlInstance) {
+            try {
+                $server = Connect-SqlInstance -SqlInstance $instance -SqlCredential $sqlcredential
+            } catch {
+                Stop-Function -Message "Error occurred while establishing connection to $instance" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
+            }
             if ($perm -contains "CreateAnyDatabase") {
-                try {
-                    $server = Connect-SqlInstance -SqlInstance $instance -SqlCredential $sqlcredential
-                } catch {
-                    Stop-Function -Message "Error occurred while establishing connection to $instance" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
-                }
-
                 foreach ($ag in $AvailabilityGroup) {
                     try {
                         $server.Query("ALTER AVAILABILITY GROUP $ag GRANT CREATE ANY DATABASE")
                     } catch {
-                        Stop-Function -Message "Failure" -ErrorRecord $_ -Target $instance
-                        return
+                        Stop-Function -Message "Failure" -ErrorRecord $_ -Target $instance -Continue
                     }
                 }
-            } elseif ($Login) {
-                $InputObject += Get-DbaLogin -SqlInstance $instance -SqlCredential $SqlCredential -Login $Login
+            }
+            if ($Login) {
+                $InputObject += Get-DbaLogin -SqlInstance $server -SqlCredential $SqlCredential -Login $Login
                 foreach ($account in $Login) {
                     if ($account -notin $InputObject.Name) {
                         try {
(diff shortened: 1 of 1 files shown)


functions/Grant-DbaAgPermission.ps1
```diff
@@ -121,30 +121,28 @@ function Grant-DbaAgPermission {
...
+            if ($Login) {
+                $InputObject += Get-DbaLogin -SqlInstance $server -SqlCredential $SqlCredential -Login $Login
                 foreach ($account in $Login) {
                     if ($account -notin $InputObject.Name) {
                         try {
                             $InputObject += New-DbaLogin -SqlInstance $server -Login $account -EnableException
                         } catch {
-                            Stop-Function -Message "Failure" -ErrorRecord $_ -Target $instance
-                            return
+                            Stop-Function -Message "Failure" -ErrorRecord $_ -Target $instance -Continue
```
wsmelton, line 145: So if it fails to grant one login you want it to continue failing?

andreasjordan: Yes. I've seen this behavior at other parts of this commandlet and in other commandlets.
But you are the expert and have the overview of the general structure in the commandlets.
So what do you suggest?

wsmelton: Well since this one is actually trying to pull in the account, if it was not able to do that then continuing would cause additional errors (e.g. that `$InputObject` would be null/empty)...at least I would think it would.

I'm not sure if the test for this command is that extensive to cover it.

andreasjordan: No it wouldn't. If $InputObject would be $null then the foreach ($account in $InputObject) would just do nothing. 
If passing in two logins via $Login, and the first fails to be created because of a typing error, do we want to create and grant the second or do we want the commandlet to fail?

andreasjordan:
@wsmelton 
Sorry, I think I completly messed up this thing. I wanted to follw you and changed the Stop-Function and commited. But now there is another change ($server / $account.Parent) I don't know where it came from. Can you please help me here? What can I do to get this working?

potatoqualitee:
This looks right to me. Thanks to you both! 🏆

PR #6631, opened 2020-07-01
Return TraceFlags as a collection rather than CSV

ben-thul:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Change the type of the TraceFlags member in the object returned from `Get-DbaStartupParameter` from a CSV (i.e. `string`) to `int[]`. This will allow for enhanced functionality for things like `TraceFlags.Contains(1118)`.

### Approach
Once all of the parsing is done, apply a type cast.

### Commands to test
Just the examples from the docs should be fine.


potatoqualitee:
This is a breaking change, but it makes sense and aligns better with PowerShell

PR #6649, opened 2020-07-08
added lookup for EncryptionAlgorithm

matt2005:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality )
 

### Purpose
Adds EncryptionAlgorithm to Get-DbaDbEncryption for databases encrypted by Certificates

### Approach
Lookup on existing object

### Commands to test
Get-DbaDbEncryption -SqlInstance 'server'

Output will contain encryption algorithm for Encrypted Databases

### Screenshots
NA

### DIFF
--- a/functions/Get-DbaDbEncryption.ps1
+++ b/functions/Get-DbaDbEncryption.ps1
@@ -136,7 +136,7 @@ function Get-DbaDbEncryption {
                         $returnCertificate.Owner = $serverCertificate.Owner
                         $returnCertificate.Object = $serverCertificate
                         $returnCertificate.ExpirationDate = $serverCertificate.ExpirationDate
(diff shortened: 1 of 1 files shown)


functions/Get-DbaDbEncryption.ps1
```diff
@@ -136,6 +136,7 @@ function Get-DbaDbEncryption {
                         $returnCertificate.Owner = $serverCertificate.Owner
                         $returnCertificate.Object = $serverCertificate
                         $returnCertificate.ExpirationDate = $serverCertificate.ExpirationDate
+                        $returnCertificate.EncryptionAlgorithm = $db.DatabaseEncryptionKey.Properties.where( { $psitem.name -eq 'EncryptionAlgorithm' }).value
```
wsmelton, line 139: At this time we are still supporting PowerShell v3, so this will need to be adjusted to use `Where-Object` instead of the where method.

matt2005: Edited to use where-object

PR #6654, opened 2020-07-10
Test get agent schedule

MikeyBronowski:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
The test was failing on en-GB locale as it was expected 12:00:00 AM for midnight instead getting 00:00:00.

### Approach

Used the same approach as for date + the default startTime is 000000 so placed that in the test. Tested it with 12:00:00 AM in the test too.


wsmelton:
I don't completely follow what this PR is solving...

The command is returning the message properly but our test was not validating it with the right format?

PR #6657, opened 2020-07-11
Install-DbaSqlWatch - Fix overflow in ProgressHelper

andreasjordan:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
Calculate and use TotalSteps to prevent overflow in ProgressHelper when installing on more than one instance.

To get TotalSteps inside ProgressHelper, the number of calls in the calling commandlet is used. This is only correct, if every call is only run once. But here two of the calls are inside the foreach-SqlInstance-loop, so with more than one SqlInstance, the percentage exceeds 100 percent and the error "The 133 argument is greater than the maximum allowed range of 100. Supply an argument that is less than or equal to 100 and then try the command again." is raised. 

Is there a general way to use ProgressHelper in those situations?

### DIFF
--- a/functions/Install-DbaSqlWatch.ps1
+++ b/functions/Install-DbaSqlWatch.ps1
@@ -182,6 +182,7 @@ function Install-DbaSqlWatch {
             return
         }
 
+        $totalSteps = $stepCounter + $SqlInstance.Count * 2
         foreach ($instance in $SqlInstance) {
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Thank you so much! I'm not sure what the ultimate solution would be. I think in other commands I reset the counter in the loop.

PR #6691, opened 2020-07-30
Add-DbaAgDatabase - fix for SQL Server 2014

### LINKED ISSUE #6608: [Bug] SeedingMode property forced to set in New-DbaAvailabilityGroup/SQL Server 2014 version
If we run `New-DbaAvailabilityGroup` against a 2014 instance it will fail with the following error while trying to add the database to availability groups.

```
PS❤️> $error | where {$_.exception -like '*seedingmode*'} | select exception

Exception
---------
System.Management.Automation.SetValueInvocationException: Exception setting "SeedingMode": "Cannot write property SeedingMode. This property is not available on SQL Server 2014."
```
The problem is obvous here, `SeedingMode` is an invalid property in 2014 version, so we'll need to add an additional condition, perhaps something like 

```
if (Get-DbaAgDatabase -SqlInstance $Primary -SqlCredential $PrimarySqlCredential -AvailabilityGroup $Name -Database $Database) {
    $dbdone = $true
} else {
    if ($server.VersionMajor -lt 13) {
    $null = Add-DbaAgDatabase -SqlInstance $Primary -SqlCredential $PrimarySqlCredential -AvailabilityGroup $Name -Database $Database -SharedPath $SharedPath -Secondary $Secondary -SecondarySqlCredential $SecondarySqlCredential 
    } 
    else {
    $null = Add-DbaAgDatabase -SqlInstance $Primary -SqlCredential $PrimarySqlCredential -AvailabilityGroup $Name -Database $Database -SeedingMode $SeedingMode -SharedPath $SharedPath -Secondary $Secondary -SecondarySqlCredential $SecondarySqlCredential 
    }
    $dbwait++
    Start-Sleep -Seconds 1
	}
```
Environment:
```

PS❤️> $PSVersionTable 

PSVersion                      5.1.14393.3471
BuildVersion                   10.0.14393.3471


PS❤️> Get-Culture

1033             en-US            English (United States)


PS❤️> Get-Module -Name dbatools -ListAvailable


    Directory: C:\Program Files\WindowsPowerShell\Modules


ModuleType Version    Name                                ExportedCommands
---------- -------    ----                                ----------------
Script     1.0.113    dbatools                            {Select-DbaObject, Set-DbatoolsConfig, Start-DbaMigration, Copy-DbaDatabase...}


PS❤️> $conn = Connect-DbaInstance -SqlInstance $server
PS❤️> $conn.Version

Major  Minor  Build  Revision
-----  -----  -----  --------
12     0      5556   -1


PS❤️> $conn.Language
English (United States)

```

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #6608 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
I split up the changes in three commits to be able to see three diffs.
First change is the bugfix: Only set the property in versions that support this property.
Second change is just to make things nice: Having variables start with lowercase and using using variable that was set.
Third change is to align tests with similar tests in New-DbaAvailabilityGroup, to correct order of tests (tests had wrong nesting) and to test before creating the smo object.

### DIFF
--- a/functions/Add-DbaAgDatabase.ps1
+++ b/functions/Add-DbaAgDatabase.ps1
@@ -149,6 +149,14 @@ function Add-DbaAgDatabase {
                 Stop-Function -Message "$($db.Name) is already joined to $($ag.Name)" -Continue
             }
 
+            if ($SeedingMode -eq "Automatic" -and $primary.VersionMajor -lt 13) {
+                Stop-Function -Message "Automatic seeding mode only supported in SQL Server 2016 and above" -Continue
+            }
+
+            if (-not $UseLastBackup -and -not $SharedPath -and $SeedingMode -ne 'Automatic') {
+                Stop-Function -Message "You must specify a SharedPath when adding databases to a manually seeded availability group" -Continue
+            }
+
             if (-not $Secondary) {
                 try {
                     $secondarynames = ($ag.AvailabilityReplicas | Where-Object Role -eq Secondary).Name
@@ -204,16 +212,7 @@ function Add-DbaAgDatabase {
                     Stop-Function -Continue -Message "Secondary replica $($secondaryInstanceReplicaName) for availability group $($ag.name) not found on $($primary.Name)"
                 }
 
-                if ($SeedingMode -eq "Automatic" -and $secondaryInstance.VersionMajor -le 12) {
-                    Stop-Function -Continue -Message "Automatic seeding mode not supported on SQL Server versions prior to 2016 - Instance $($secondaryInstance.Name)"
-
-                    if (-not $SharedPath -and -not $UseLastBackup) {
-                        Stop-Function -Continue -Message "Automatic seeding not supported and no $SharedPath provided. Database not added to instance $($secondaryInstance.Name)"
-                    }
-                }
-
-
-                if ($SeedingMode -and $secondaryInstance.VersionMajor -gt 12) {
+                if ($SeedingMode -and $secondaryInstance.VersionMajor -ge 13) {
                     $agreplica.SeedingMode = $SeedingMode
                     $agreplica.Alter()
                 }
--- a/functions/Add-DbaAgDatabase.ps1
+++ b/functions/Add-DbaAgDatabase.ps1
@@ -153,10 +153,6 @@ function Add-DbaAgDatabase {
                 Stop-Function -Message "Automatic seeding mode only supported in SQL Server 2016 and above" -Continue
             }
 
-            if ($SeedingMode -eq 'Manual' -and -not $UseLastBackup -and -not $SharedPath) {
-                Stop-Function -Message "You must specify a SharedPath or UseLastBackup when adding databases to a manually seeded availability group" -Continue
-            }
-
             if (-not $Secondary) {
                 try {
                     $secondarynames = ($ag.AvailabilityReplicas | Where-Object Role -eq Secondary).Name
--- a/functions/Add-DbaAgDatabase.ps1
+++ b/functions/Add-DbaAgDatabase.ps1
@@ -153,11 +153,9 @@ function Add-DbaAgDatabase {
                 Stop-Function -Message "Automatic seeding mode only supported in SQL Server 2016 and above" -Continue
             }
 
-            # disabled, we first have to discuss the need for this test and the consequences of this test
-            # was not active in previous version due to wrong placement
-            #            if (-not $UseLastBackup -and -not $SharedPath -and $SeedingMode -ne 'Automatic') {
-            #                Stop-Function -Message "You must specify a SharedPath when adding databases to a manually seeded availability group" -Continue
-            #            }
+            if ($SeedingMode -eq 'Manual' -and -not $UseLastBackup -and -not $SharedPath) {
+                Stop-Function -Message "You must specify a SharedPath or UseLastBackup when adding databases to a manually seeded availability group" -Continue
+            }
 
             if (-not $Secondary) {
                 try {
--- a/functions/Add-DbaAgDatabase.ps1
+++ b/functions/Add-DbaAgDatabase.ps1
@@ -153,9 +153,11 @@ function Add-DbaAgDatabase {
                 Stop-Function -Message "Automatic seeding mode only supported in SQL Server 2016 and above" -Continue
             }
 
-            if (-not $UseLastBackup -and -not $SharedPath -and $SeedingMode -ne 'Automatic') {
-                Stop-Function -Message "You must specify a SharedPath when adding databases to a manually seeded availability group" -Continue
-            }
+            # disabled, we first have to discuss the need for this test and the consequences of this test
+            # was not active in previous version due to wrong placement
+            #            if (-not $UseLastBackup -and -not $SharedPath -and $SeedingMode -ne 'Automatic') {
+            #                Stop-Function -Message "You must specify a SharedPath when adding databases to a manually seeded availability group" -Continue
+            #            }
 
             if (-not $Secondary) {
                 try {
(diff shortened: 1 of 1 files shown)


andreasjordan:
I think there are other issues with this command, I will work on them as soon as this pull request is merged into development.

andreasjordan:
I try to have a closer look at the failing AppVeyor test - but it will take some time.

andreasjordan:
Ok, no AG related command show up in AppVeyor test - but others fail. What's the problem?

andreasjordan:
@wsmelton I think this pull request is ready for a merge. If you need anything else, give me a sign.

functions/Add-DbaAgDatabase.ps1
```diff
@@ -137,18 +137,28 @@ function Add-DbaAgDatabase {
...
+            if ($SeedingMode -eq "Automatic" -and $primary.VersionMajor -lt 13) {
+                Stop-Function -Message "Automatic seeding mode only supported in SQL Server 2016 and above" -Continue
+            }
+
+            # disabled, we first have to discuss the need for this test and the consequences of this test
+            # was not active in previous version due to wrong placement
+            #            if (-not $UseLastBackup -and -not $SharedPath -and $SeedingMode -ne 'Automatic') {
+            #                Stop-Function -Message "You must specify a SharedPath when adding databases to a manually seeded availability group" -Continue
+            #            }
+
```
wsmelton, line 161: Why would we not want to verify that either `UseLastBackup` or `SharedPath` is provided when seeding mode defaults to manual?

andreasjordan: I'll start a discussion about that as soon as this pull request is merged. 
Short answer is: Because then the tests would fail.
Long answer is: In a real world scenario I always have an availability group with two servers, so I need a shared path for the backups or have to use already available backups.
But in a demo scenario (and in the scenario of the tests) I have an availability group with two instances on the same server. If they run under the same account, the restore can access the local backups. Have a look at the tests: They don't use UseLastBackup nor SharedPath, but the default manual seeding. That works because both instances are on the same server.
So I put these lines there to not forget to start a discussion about that, but commented them out to be able to leave the tests untouched.

wsmelton: What tests are you referring to? As well, our commands are not meant to support demo environments and should be written for production situations/configurations; especially for AGs. 

The following code at line 229 is where things will start throwing oddball errors if the user does not provide at least one of those two parameters...that is the whole point of validation and why we do that in our commands (as a standard), to control error messages and ensure they are readable to the user.

```
if ($UseLastBackup) {
                                $allbackups[$db] = Get-DbaDbBackupHistory -SqlInstance $primarydb.Parent -Database $primarydb.Name -IncludeCopyOnly -Last -EnableException
                            } else {
                                $fullbackup = $primarydb | Backup-DbaDatabase -BackupDirectory $SharedPath -Type Full -EnableException
                                $logbackup = $primarydb | Backup-DbaDatabase -BackupDirectory $SharedPath -Type Log -EnableException
                                $allbackups[$db] = $fullbackup, $logbackup
```

If you do not first validate that either `UseLastBackup` parameter or `SharedPath` parameter are provided what output is shown to show the user that a missing parameter was provided to the parent command and that it was just not an issue with the Backup command?

wsmelton: As well, if you want to start a discussion on the comment block of code please open an issue and add that block of code...then remove from this PR. If in the end, we decided that it is not required then we do not have to go back and do a PR just to clean up dead code.

andreasjordan: These tests are failing now:

```
Describe : Add-DbaAgDatabase Integration Tests
Context  : adds ag db
Name     : It returns proper results
Result   : Failed
Message  : Expected 'dbatoolsci_addagdb_agroup', but got $null.

Describe : Get-DbaAgDatabase Integration Tests
Context  : gets ag db
Name     : It returns results
Result   : Failed
Message  : Expected 'dbatoolsci_getagdb_agroup', but got $null.

Describe : New-DbaAvailabilityGroup Integration Tests
Context  : adds an ag
Name     : It returns an ag with a db named
Result   : Failed
Message  : Expected 'dbatoolsci_addag_agroupdb', but got $null.

Describe : Suspend-DbaAgDbDataMovement Integration Tests
Context  : resumes  data movement
Name     : It returns resumed results
Result   : Failed
Message  : Expected 'dbatoolsci_resumeagdb_agroup', but got $null.

```

andreasjordan: Here is how the test ag is set up:

`$ag = New-DbaAvailabilityGroup -Primary $script:instance3 -Name $agname -ClusterType None -FailoverMode Manual -Database $dbname -Confirm:$false -Certificate dbatoolsci_AGCert
`

So no `SharedPath` and no `UseLastBackup`. And I even had a look at `New-DbaAvailabilityGroup` and the parameter `UseLastBackup`: It is not passed to the `Add-DbaAgDatabase`, so there is another bug.

So I think we have two options: Keep the test and fix `New-DbaAvailabilityGroup` and all the tests or delete the test. As this is a branch and pull request for fixing the 2014 issue, I would like to delete the test here to be able to merge and solve the 2014 issue. The functionality would not be touched, it would just be a fix for 2014. For the parameter `UseLastBackup` we need a general discussion - because it just does not what it says.

@wsmelton What do you think?

wsmelton: If the test was passing before and fails from your change, what exactly are we fixing?

andreasjordan: Foreword: Cooperation across time zones, languages ​​and cultures is not easy and I think we keep talking about each other. 

Please have a look at these lines of code:

```
if ($SeedingMode -eq "Automatic" -and $secondaryInstance.VersionMajor -le 12) {
                    Stop-Function -Continue -Message "Automatic seeding mode not supported on SQL Server versions prior to 2016 - Instance $($secondaryInstance.Name)"

                    if (-not $SharedPath -and -not $UseLastBackup) {
                        Stop-Function -Continue -Message "Automatic seeding not supported and no $SharedPath provided. Database not added to instance $($secondaryInstance.Name)"
                    }
                }
```

There is the test for `$SharedPath` and `$UseLastBackup`. Because of the `Stop-Function`, this code was never executed. So there was never a working test for these parameters. And the pester-tests don't use these parameters and work well. So there are working cases where none of these parameters are needed.
Can we agree on that?

To keep this a non breaking change and only a bugfix for version 2014, I am not allowed to introduce a new test that will let the pester-tests fail. And so I'm also not allowed to change the location of the "inactive" test to a location where it is "active" - because that would be a breaking change.

So I will commit the deletion of the test to fix the bug and not break anything else. 
But I will later on work on the ag commands and especially the parameters `$SharedPath` and `$UseLastBackup` - but consistent in all related commands.

wsmelton: `-Continue` this ensures the `Stop-Function` is to continue executing and not throw the full exception.

andreasjordan: > `-Continue` this ensures the `Stop-Function` is to continue executing and not throw the full exception.

Yes and no. It continues with the command, but does quit the loop. So the `if (-not $SharedPath -and -not $UseLastBackup) {` is never ever executed.

So there never was a test for these parameters. So I can not put "the validation for those two parameters [...] back in the command". I just can't. Sorry to say. I just can't. I can add them to the command - but that would be a breaking change. And I will not do that in a pull request for a bugfix,
And sure I could position them behind a `Stop-Function` like it was - but I will not do that either.

potatoqualitee:
not sure if this was closed by accident, but the code looks good and valid 👍  we often get reports of seedingmode/version issues. thank you for the catch, Andres.

PR #6697, opened 2020-08-01
Test-DbaLastBackup - Add support for MaxTransferSize and BufferCount

### LINKED ISSUE #6694: extend test-dbaLastBackup with -MaxTransferSize -BufferCount
I am using the function test-dbaLastBackup to test my backups 24/7 on a test server. And without these changes the server is slow.
These changes allow me to use -maxtransfersize 4128768 -buffercount = 100 settings for the restore action, and by this gain 50% in restore speed.

the parameters should only be for passtrough, and for testing of the parameters sanity the underlaying Restore-DbaDatabase is responsible

Fiander:
… can give 50% speed improvement on restores

## Type of Change

 - [ X] New feature (non-breaking change, adds functionality, fixes #6694)
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Give a possible speedup for people using Test-DbaLastBackup.

### Approach

Test-DbaLastBackup uses Restore-DbaDatabase to do a restore of a database.
Restore-DbaDatabase has two parameters that when used use more memory to restore a database.
Test-DbaLastBackup did not use those parameters. 
I have added those two parameters in a passthrough way in Test-DbaLastBackup.

by adding -MaxTransferSize 4128768 -BufferCount = 100 as parameters to Test-DbaLastBackup i get a 50% higher throughput in my restores ( differs from system to system )

because the parameters are passthrough, i did not add a sanitize section. sanitizing is the job of Restore-DbaDatabase. 

Also did a small section of code cleanup by adding a splat variable.

### Learning
[https://www.mssqltips.com/sqlservertip/4935/optimize-sql-server-database-restore-performance/](url)

### DIFF
--- a/functions/Test-DbaLastBackup.ps1
+++ b/functions/Test-DbaLastBackup.ps1
@@ -89,11 +89,6 @@ function Test-DbaLastBackup {
     .PARAMETER Confirm
         If this switch is enabled, you will be prompted for confirmation before executing any operations that change state.
 
-    .PARAMETER EnableException
-        By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
-        This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
-        Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
-
     .PARAMETER MaxTransferSize
         Parameter to set the unit of transfer. Values must be a multiple of 64kb and a max of 4GB
         Parameter is used as passtrough for Restore-DbaDatabase.
@@ -102,6 +97,11 @@ function Test-DbaLastBackup {
         Number of I/O buffers to use to perform the operation.
         Refer to https://msdn.microsoft.com/en-us/library/ms178615.aspx for more detail
         Parameter is used as passtrough for Restore-DbaDatabase.
+
+    .PARAMETER EnableException
+        By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
+        This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
+        Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
 
     .NOTES
         Tags: DisasterRecovery, Backup, Restore
@@ -184,9 +184,9 @@ function Test-DbaLastBackup {
         [string]$AzureCredential,
         [parameter(ValueFromPipeline)]
         [Microsoft.SqlServer.Management.Smo.Database[]]$InputObject,
-        [switch]$EnableException,
         [int]$MaxTransferSize,
-        [int]$BufferCount
+        [int]$BufferCount,
+        [switch]$EnableException
     )
     process {
         if ($SqlInstance) {
--- a/functions/Test-DbaLastBackup.ps1
+++ b/functions/Test-DbaLastBackup.ps1
@@ -95,7 +95,7 @@ function Test-DbaLastBackup {
 
     .PARAMETER BufferCount
         Number of I/O buffers to use to perform the operation.
-        Refer to https://msdn.microsoft.com/en-us/library/ms178615.aspx for more detail
+        Refererence: https://msdn.microsoft.com/en-us/library/ms178615.aspx#data-transfer-options
         Parameter is used as passtrough for Restore-DbaDatabase.
 
     .PARAMETER EnableException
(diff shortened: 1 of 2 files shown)


functions/Test-DbaLastBackup.ps1
```diff
@@ -175,7 +184,9 @@ function Test-DbaLastBackup {
         [string]$AzureCredential,
         [parameter(ValueFromPipeline)]
         [Microsoft.SqlServer.Management.Smo.Database[]]$InputObject,
-        [switch]$EnableException
+        [switch]$EnableException,
```
wsmelton, line 187: Please move the additional parameters before EnableException. This one should always be the last parameter. In both help and the param block.

Fiander: i have moved them up as requested, and added the parameters to the test script.

functions/Test-DbaLastBackup.ps1
```diff
@@ -89,6 +89,15 @@ function Test-DbaLastBackup {
     .PARAMETER Confirm
         If this switch is enabled, you will be prompted for confirmation before executing any operations that change state.
 
+    .PARAMETER MaxTransferSize
+        Parameter to set the unit of transfer. Values must be a multiple of 64kb and a max of 4GB
+        Parameter is used as passtrough for Restore-DbaDatabase.
+
+    .PARAMETER BufferCount
+        Number of I/O buffers to use to perform the operation.
+        Refer to https://msdn.microsoft.com/en-us/library/ms178615.aspx for more detail
```
wsmelton, line 98:
```suggestion
        Refererence: https://msdn.microsoft.com/en-us/library/ms178615.aspx#data-transfer-options
```

functions/Test-DbaLastBackup.ps1
```diff
@@ -89,6 +89,15 @@ function Test-DbaLastBackup {
     .PARAMETER Confirm
         If this switch is enabled, you will be prompted for confirmation before executing any operations that change state.
 
+    .PARAMETER MaxTransferSize
+        Parameter to set the unit of transfer. Values must be a multiple of 64kb and a max of 4GB
```
wsmelton, line 93: If we have specific requirements around the value will the error returned by provided an invalid value be clear to the user?

wsmelton: Unsure if maybe the Restore-DbaDatabase command handles that for us, have not tested/checked.

wsmelton:
Forgot to also ask to add an example for the parameters. If you want to just use the one noted in the issue is fine.

functions/Test-DbaLastBackup.ps1
Fiander: in Restore-DbaDatabase this is done:
`            if (($MaxTransferSize % 64kb) -ne 0 -or $MaxTransferSize -gt 4mb) {
                Stop-Function -Category InvalidArgument -Message "MaxTransferSize value must be a multiple of 64kb and no greater than 4MB"
                return
            }
            if ($BlockSize) {
                if ($BlockSize -notin (0.5kb, 1kb, 2kb, 4kb, 8kb, 16kb, 32kb, 64kb)) {
                    Stop-Function -Category InvalidArgument -Message "Block size must be one of 0.5kb,1kb,2kb,4kb,8kb,16kb,32kb,64kb"
                    return
                }
            }`

because of this, and that i only wanted a minimum impact so passthrough style, i did not touch the parameters.
i could copy code above, but that would double that part of code, and not really add anything.

wsmelton: Since these are parameters from Restore command we should match the help that command provides as well. So the .PARAMETER for each from the restore command can just be copied/pasted into this one to make them consistent.

wsmelton: Tabbing through trying to use the command with the new commands I think it would be more fitting to have these two parameters go after the DeviceType parameter (both in help and param block). That seems to flow better when folks are tabbing through the list.

wsmelton:
The error handling is not picking up the exception thrown by the restore command on the parameter validation, but that is something I can address later.

image

wsmelton:
The test for formatting is also failing on the changes, the splatting is not formatted properly.

If you will import the branch into your PowerShell session and then run this command, it will fix the formatting issues and then you can simply commit that change:

```Invoke-DbatoolsFormatter .\functions\Test-DbaLastBackup.ps1```

Fiander:
euhg? 
you are trying to put a bigint in a int? That is an error that triggers before the test-dbaLastBackup is actualy being called. That means that restore-dbaBackup is also not called. That is also what the error says: cannot convert to int32. value to big or to small.

if you want to test that exeption i would sugest an integer(32bit ) that is not a multiple of 64KB, eg: 123456

i will look into the failing check concerning splatting.

Fiander:
p.s. where can i find the the tests that are failing? all i see is a test for a notebook that misses a file.

wsmelton:
You would go into the Details link, in Appveyor click on the job that is failing and at the bottom, we output the information of the failed Pester tests.

PR #6706, opened 2020-08-03
Get-DbaAgentJob unexcepted behavior if JobSteps.DatabaseName is array

### LINKED ISSUE #6699: [Bug] Function Get-DbaAgentJob - unexcepted behavior with few job steps
#### SQL Server:

```sql
Microsoft SQL Server 2016 (SP1-CU5) (KB4040714) - 13.0.4451.0 (X64)   Sep  5 2017 16:12:34   Copyright (c) Microsoft Corporation  Standard Edition (64-bit) on Windows Server 2016 Standard 6.3 <X64> (Build 14393: ) (Hypervisor) 
```

```sql
us_english
```

### Report

https://github.com/sqlcollaborative/dbatools/blob/a1c053f8a1d1cf9a16a14909f015430cc75415c4/functions/Get-DbaAgentJob.ps1#L123

#### Host used

- [X] powershell.exe
- [ ] ISE
- [ ] VS Code
- [ ] Other (please specify)

- [ ] Still buggy in powershell.exe

#### Steps to Reproduce

Create job with few (2 or more) steps for database. Next need to call function Get-DbaAgentJob with param $Database which contains name in created early job. Function doesn't return excepted behavior - not include this job.

#### Expected Behavior

Return jobs which contains DatabaseName = $Database

#### Actual Behavior

Doesn't return job if it has 2 or more steps because "-in" operator excepted next params order:
1) value what need to find
2) array

But JobSteps.DatabaseName can be array too.

i7slegend:
## Type of Change

 - [X] Bug fix (non-breaking change, fixes #6699 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
https://github.com/sqlcollaborative/dbatools/issues/6699
### Approach

Fix unexcepted behavior with JobSteps.DatabaseName which can be array of object if a job has 2 or more steps.

### DIFF
--- a/functions/Get-DbaAgentJob.ps1
+++ b/functions/Get-DbaAgentJob.ps1
@@ -79,10 +79,16 @@ function Get-DbaAgentJob {
         PS C:\> $servers | Get-DbaAgentJob | Out-GridView -PassThru | Start-DbaAgentJob -WhatIf
 
         Find all of your Jobs from SQL Server instances in the $servers collection, select the jobs you want to start then see jobs would start if you ran Start-DbaAgentJob
+
     .EXAMPLE
-       PS C:\> Get-DbaAgentJob -SqlInstance sqlserver2014a | Where-Object Category -eq "Report Server" | Export-DbaScript -Path "C:\temp\sqlserver2014a_SSRSJobs.sql"
+        PS C:\> Get-DbaAgentJob -SqlInstance sqlserver2014a | Where-Object Category -eq "Report Server" | Export-DbaScript -Path "C:\temp\sqlserver2014a_SSRSJobs.sql"
 
         Exports all SSRS jobs from SQL instance sqlserver2014a to a file.
+
+    .EXAMPLE
+        PS C:\> Get-DbaAgentJob -SqlInstance sqlserver2014a -Database msdb
+
+        Finds all jobs on sqlserver2014a that T-SQL job steps associated with msdb database
     #>
     [CmdletBinding()]
     param (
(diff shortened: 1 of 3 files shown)


functions/Get-DbaAgentJob.ps1
```diff
@@ -120,7 +120,7 @@ function Get-DbaAgentJob {
             }
             if ($Database) {
                 $jobs = $jobs | Where-Object {
-                    $_.JobSteps.DatabaseName -in $Database
+                    $_.JobSteps.DatabaseName | Where-Object { [string]$_ -in $Database }
```
wsmelton, line 123: This is already a `Where-Object` why does it need a second call to `Where-Object`?

i7slegend: JobSteps.DatabaaeName is something like an aggregate property. I don't find any documentation about this but If you create job with two steps you found what condition in current version dbatools doesn't return this job because excepts first operand in the "-in" operator is value (not an array / collection).

wsmelton: Understand that portion but the only change required for this is to change the `-in` to `-Contains`

wsmelton: Actually you can reformat that line of code as well from this:
```
            if ($Database) {
                $jobs = $jobs | Where-Object {
                    $_.JobSteps.DatabaseName -in $Database
                }
            }
```

To this:
```
            if ($Database) {
                $jobs = $jobs | Where-Object { $_.JobSteps.DatabaseName -contains $Database }
            }
```

wsmelton:
Thanks for the catch on this one @i7slegend 

I made the adjustment to the logic as you noted, also updated the test to cover this scenario properly, and then added an example (saw we hadn't included one for use of that parameter).

functions/Get-DbaAgentJob.ps1
i7slegend: I don't think what "contains" can be apply here because $Database is array too.
I can't test it now, but if it is works - okay.

wsmelton: Oh, ok...now I follow. Expanded the test to pass in multiple databases to the parameter and issue is it is trying to compare an Object to a String array.

wsmelton:
_The stupid `Invoke-DbaDbClone` test failing again.._ 😠 

Does not pertain to this PR.

PR #6707, opened 2020-08-03
Reset-DbaAdmin - Corrected some wrong variables

andreasjordan:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
Corrected some obviously wrong variables and changed the case of some varible names to the standard.

### DIFF
--- a/functions/Reset-DbaAdmin.ps1
+++ b/functions/Reset-DbaAdmin.ps1
@@ -184,7 +184,7 @@ function Reset-DbaAdmin {
                     Write-Message -Level Verbose -Message "First attempt using ICMP failed. Trying to connect using sockets. This may take up to 20 seconds."
                     $tcp = New-Object System.Net.Sockets.TcpClient
                     try {
-                        $tcp.Connect($hostName, 135)
+                        $tcp.Connect($baseaddress, 135)
                         $tcp.Close()
                         $tcp.Dispose()
                     } catch {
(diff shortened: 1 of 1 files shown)


andreasjordan:
In addition to the changes made: Can someone have a look at the variable $password. Is it necessary or can it be replaced by $SecurePassword?

andreasjordan:
And another question: Why is it necessary to get $ipaddr and $hostName? 
$hostName is used to open the session and $ipaddr is used to get the services.
This can not be done with $baseaddress (= $instance.ComputerName)?

wsmelton:
The functionality in the command to prompt for password stays.

wsmelton:
Trying with the host and IP is required due to various customer configurations in both networks and servers. This command has been around for many years and some changes were made around it because of user feedback.

wsmelton:
Memory serves the IP and hostname stuff had to deal with users that have multiple-tier DNS or multiple site domains, where the IP had to be used vs the hostname. We had same situations around the resolve command for that too....but can't recall if we ever put all that "story" in an issue/PR.

andreasjordan:
> Memory serves the IP and hostname stuff had to deal with users that have multiple-tier DNS or multiple site domains, where the IP had to be used vs the hostname. We had same situations around the resolve command for that too....but can't recall if we ever put all that "story" in an issue/PR.

Thanks for the info. The reason for the question was to learn something about the modul and the command. I think there is a bug inside that part of the code, but I'll do more testing and open an issue then.

andreasjordan:
> The functionality in the command to prompt for password stays.

@wsmelton I don't want to delete this. Just wanted to understand the reason for the new variable, because later on the use is not consistent in my eyes. $password is used to construct the sql, but later the login test is done with $SecurePassword. But in the case the user is prompted for the password, $SecurePassword is never updated with that so the test will fail. Is this intended?

wsmelton:
The password prompt does work as I've used it many times in resetting instances.

PR #6725, opened 2020-08-08
Adding ArgumentList to Update-DbaInstance

### LINKED ISSUE #6271: [Bug] Export-DbaUser outputs USE <DB> regardless of ScriptingOptions
Apologies for not filling out the form below.  Line 213 of functions/Export-DbaUser.ps1 is:

`$outsql += "USE [" + $db.Name + "]"`

This negates any ScriptingOptions.IncludeDatabaseContext passed to the function via the -ScriptingOptionsObject param. It even negates the default option set on line 194:

`$ScriptingOptionsObject.IncludeDatabaseContext = $false`

I believe this line 213 should be replaced with something like

`if ($ScriptingOptionsObject.IncludeDatabaseContext) {
    $outsql += "USE [" + $db.Name + "]"
}`

### Environmental information

```powershell
# Replace with output of Get-Culture
```

#### SQL Server: 

```sql
/* REPLACE WITH output of @@VERSION */
```

```sql
```

#### Host used

- [ ] powershell.exe
- [ ] ISE
- [x] VS Code
- [ ] Other (please specify)

- [ ] Still buggy in powershell.exe

### LINKED ISSUE #6643: Update-DbaInstance add option for /SkipRules=RebootRequiredCheck
### Summary of new feature
Microsoft has an undocumented option when performing patching/installs ---- "/SkipRules=RebootRequiredCheck" this skip rules allows you to install a patch even if a pending reboot is found on the machine.  We typically only reboot machines during their maintenance window, but if we miss that window and a new CU is released in between, we can still apply the patch by passing in the command line /SkipRules=RebootRequiredCheck

By providing this functionality, as for instance an optional flag such as -SkipRebootCheck then we have the option to use it or not.

The whole intent of this is that even if a reboot is pending on the machine, it may be for a variety of reasons and none of which may apply to your SQL Server.  If that is the case, then you can use the /SkipRules=RebootRequiredCheck to ignore any pending reboot and still apply a patch.

here are some examples of how we use it in our shop:

```
-- installing a patch
"\\server\SQL Server Installs-Patches\2016\SP2\CU13\SQLServer2016-KB4549825-x64.exe" /SkipRules=RebootRequiredCheck

-- uninstall a patch
"D:\SQL Patches\2016\SP2\CU11\SQLServer2016-KB4527378-x64.exe" /ACTION=RemovePatch /InstanceName=INS1_D /SkipRules=RebootRequiredCheck

-- installing the base install with a config file, UpdateEnabled & SkipRules
"D:\SQL Patches\2012\setup.exe" /ACTION=INSTALL /IACCEPTSQLSERVERLICENSETERMS /CONFIGURATIONFILE="D:\SQLSERVER\Configuration Templates\SQL2012_ConfigurationFile_IIS1Dev.ini" /UpdateEnabled=True /UpdateSource="D:\SQL Patches\2012\all-patches" /SkipRules=RebootRequiredCheck

```

1.0.112

nvarscar:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality, fixes #6643 )
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Add an ability to provide custom command line parameters to executable calls

### Approach
ArgumentList passed on to Invoke-Program

### DIFF
--- a/functions/Invoke-DbaAdvancedUpdate.ps1
+++ b/functions/Invoke-DbaAdvancedUpdate.ps1
@@ -32,6 +32,9 @@ Function Invoke-DbaAdvancedUpdate {
     .PARAMETER ExtractPath
         Lets you specify a location to extract the update file to on the system requiring the update. e.g. C:\temp
 
+    .PARAMETER ArgumentList
+        A list of extra arguments to pass to the execution file.
+
     .PARAMETER WhatIf
         Shows what would happen if the command were to run. No actions are actually performed.
 
@@ -72,6 +75,7 @@ Function Invoke-DbaAdvancedUpdate {
         [string]$Authentication = 'Credssp',
         [pscredential]$Credential,
         [string]$ExtractPath,
+        [string[]]$ArgumentList,
         [switch]$EnableException
 
     )
@@ -145,7 +149,7 @@ Function Invoke-DbaAdvancedUpdate {
             # Extract file
             Write-ProgressHelper -ExcludePercent -Activity $activity -Message "Extracting $($currentAction.Installer) to $spExtractPath"
             Write-Message -Level Verbose -Message "Extracting $($currentAction.Installer) to $spExtractPath"
-            $extractResult = Invoke-Program @execParams -Path $currentAction.Installer -ArgumentList "/x`:`"$spExtractPath`" /quiet" -Fallback
+            $extractResult = Invoke-Program @execParams -Path $currentAction.Installer -ArgumentList @("/x`:`"$spExtractPath`"", "/quiet") -Fallback
             if (-not $extractResult.Successful) {
                 $msg = "Extraction failed with exit code $($extractResult.ExitCode), try specifying a location instead using -ExportPath"
                 $output.Notes += $msg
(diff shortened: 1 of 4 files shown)


functions/Invoke-DbaAdvancedUpdate.ps1
```diff
@@ -32,6 +32,9 @@ Function Invoke-DbaAdvancedUpdate {
     .PARAMETER ExtractPath
         Lets you specify a location to extract the update file to on the system requiring the update. e.g. C:\temp
 
+    .PARAMETER ArgumentList
+        A list of extra arguments to pass to the execution file.
+
```
wsmelton, line 37: Your example includes the forwardslash, should we add that tidbit or template example to do parameter description?

It may be assumed but just want to check.

nvarscar: That's a good idea. Even though there is now a new example, it would be beneficial to have that info in both places.

nvarscar: I've updated the parameter in the user-facing command. Invoke-DbaAdvancedUpdate is never meant to be run by a user and is only public due to runspace-related requirements.

nvarscar: Since it somehow failed on an unrelated test I re-pushed the commit that updates help for both commands.

potatoqualitee:
ill try a repush as well

potatoqualitee:
hayle yeah, tests pass, merge city. thanks for including the test!

PR #6773, opened 2020-08-28
Fix Log Shipping command to exclude monitor procedure when server is higher than 2012

sanderstad:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
The check for the major version should return a value but it was based on a variable that does not have that property.

### Approach
Changed the check to use the correct variable and now it returns the right value

### Commands to test
New-DbaLogShippingSecondaryDatabase

### Screenshots
afbeelding


potatoqualitee:
hell yes thanks a ton, will publish momentarily

PR #6777, opened 2020-08-29
Query Store Set and Get functions shouldn't try to get data from model

tboggiano:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 

- Model database properties cannot be set for query store so added to the exclude databases, since it can be set also add to the exclude dbs in the Get command as well

### Approach

- Just changed the $Exclude parameter

### Commands to test

Set-DbaDbQueryStoreOption -SqlInstance $SQLInstance -Database "model" -State "ReadWrite" -MaxSize 2048
Get-DbaDbQueryStoreOption -SqlInstance $SQLInstance -Database "model"

### Screenshots

image
image


wsmelton:
Tests are failing on format, possible your editor modified it in some manner. If you will run the `Invoke-DbatoolsFormatter` on the files and that should fix them.

SMO does not seem to support changing model settings for query store but it can be done last I checked... Also noted here: https://www.sqlskills.com/blogs/erin/sql-server-query-store-default-settings/

tboggiano:
The article by Erin states in bold "That you can only truly enable and use Query Store against one system database: msdb.", it can't be change in model.  That's why SMO errored out.  SQL will not error out but it actually doesn't do anything either.

wsmelton:
I'm not concerned with **using** or **enabling** Query Store in the model database, It is that you are allowed to control the settings of QueryStore for databases created on the instance (which is the intent of the model database). If SMO does not support it and T-SQL does then should we not adjust to support it as well in the functions?

If the version of SMO we are using in the module does not support newer features of Query Store and we need to update it please file a bug around this.

tboggiano:
I file a bug this evening when I get back to my test machine.

tboggiano:
Yes and Get command will not return any results rather you use SMO or T-SQL, Microsoft wants it to by a mystery.  So I can change it all to T-SQL like the 2017 and 2019 features as I did before in Set so we can set model, but then you can't view it or Copy it with the other functions, silly Microsoft.

wsmelton:
Just need to update the test to use something besides the model DB and we can just leave it as-is until users come asking.

potatoqualitee:
k, updated! @tboggiano  - your branch was super out of date, by like 300+ commits. You'll need to sync from origin when you get a chance 👍

potatoqualitee:
sweet - tests pass now, will merge 💯 thank you both!

Issue #6604, opened 2020-06-22
Support for non default ports when querying cmdlets related to Availability Groups.


amon09:
### Summary of new feature

Currently if I run `Get-DbaDbBackupHistory` cmdlet against an instance that contains availability groups, it would fail to get results from all replicas. For example, for a default Sql Server instance, the replica name would be `SERVERNAME` and for a named instance it would always be `SERVERNAME\INSTANCENAME`. If the default instance does not listen on `1433` or if the browser service is disabled/UDP 1434 is blocked(for named instances), the cmdlet will fail to return results for all replicas. 

```
$AvailabilityGroupBase = ($server.AvailabilityGroups | Where-Object { $_.AvailabilityGroupListeners.name -eq $agShortInstance })
$AgLoopResults = Get-DbaDbBackupHistory -SqlInstance $AvailabilityGroupBase.AvailabilityReplicas.name @PSBoundParameters -AgCheck -IncludeCopyOnly
$AvailabilityGroupName = $AvailabilityGroupBase.name
Foreach ($agr in $AgLoopResults) {
    $agr.AvailabilityGroupName = $AvailabilityGroupName
}
```

Similarly there are other cmdlets, for e.g `Sync-DbaAvailabilityGroup` that could potentially run into the same issue (unless we use the `-Secondary` parameter explicitly in this case).

```
if ($InputObject) {
      $Secondary += (($InputObject.AvailabilityReplicas | Where-Object Name -ne $server.DomainInstanceName).Name | Select-Object -Unique)
 }

 if ($Secondary) {
     $Secondary = $Secondary | Sort-Object
     $secondaries = @()
     foreach ($computer in $Secondary) {
         try {
             $secondaries += Connect-SqlInstance -SqlInstance $computer -SqlCredential $SecondarySqlCredential
                } 
```

### Proposed technical details (if applicable)

I'm sure there is a better elegant way to achieve this, although the following changes for `Get-DbaDbBackupHistory` works well for me because we use a common non-default port per cluster. I would assume there are people who may have unique ports per instance within the cluster itself though.

```
$port = (Get-DbaTcpPort -SqlInstance $instance).port
if ((($server.AvailabilityGroups.count -gt 0) -and ($server.AvailabilityGroups.AvailabilityGroupListeners)) -and ($agCheck -ne $True)) {
...
[...]

PR #6778, opened 2020-08-30
Get-DbaDbBackupHistory - change AgCheck to SuppressAgCheck and use it in Backup-DbaDatabase

andreasjordan:
This change is not a classic bug fix, but it will help in situations described in #6604 and generally speed up Backup-DbaDatabase because in the call to get the record for the just taken backup it does not ask other instances.

I changed the name of a parameter to better reflect its purpose. And it makes this parameter a public parameter and not only an internal one. The name "SuppressAgCheck" is only an idea, it could also be "NoAgCheck" or some other name. But I think "AgCheck" is wrong because this is positiv, so I assume some kind of active check related to AGs - but the parameter itself is negative and suppresses some action. But I added an alias for backword compatibility, just in case someone uses this (internal) parameter.

I already mentioned this branch here: https://github.com/sqlcollaborative/dbatools/issues/6604#issuecomment-672638390 and hoped for a discussion on it. So with this pull request I want to get feedback on my ideas so that my ideas are not quietly lost.

I also fixed some case issues that VSCode reported.

### DIFF
--- a/functions/Get-DbaDbBackupHistory.ps1
+++ b/functions/Get-DbaDbBackupHistory.ps1
@@ -229,7 +229,7 @@ function Get-DbaDbBackupHistory {
             }
             $AgResults = @()
             $ProcessedAgDatabases = @()
-            if (($server.AvailabilityGroups.count -gt 0) -and (-not $SuppressAgCheck)) {
+            if (($server.AvailabilityGroups.count -gt 0) -and ($SuppressAgCheck -ne $True)) {
                 $agShortInstance = $instance.FullName.split('.')[0]
                 if ($agShortInstance -in ($server.AvailabilityGroups.AvailabilityGroupListeners).Name) {
                     # We have a listener passed in, just query the dbs specified or all in the AG
@@ -309,7 +309,7 @@ function Get-DbaDbBackupHistory {
             if ($ExcludeDatabase) {
                 $databases = $databases | Where-Object Name -NotIn $ExcludeDatabase
             }
-            if (($server.AvailabilityGroups.count -gt 0) -and (-not $SuppressAgCheck)) {
+            if (($server.AvailabilityGroups.count -gt 0) -and ($SuppressAgCheck -ne $True)) {
                 $adbs = $databases | Where-Object Name -In $server.AvailabilityGroups.AvailabilityDatabases.Name
                 $adbs = $adbs | Where-Object Name -NotIn $ProcessedAgDatabases
                 ForEach ($adb in $adbs) {
--- a/functions/Get-DbaDbBackupHistory.ps1
+++ b/functions/Get-DbaDbBackupHistory.ps1
@@ -152,6 +152,12 @@ function Get-DbaDbBackupHistory {
 
         Will query all replicas in the Availability Group with AgListener and return the backup chain (Full, Diff and Log) to restore to the most rececnt point in time
 
+    .EXAMPLE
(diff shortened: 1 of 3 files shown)


andreasjordan:
I added an example and would be happy if a native english speeker could have a look and rephrase it.

wsmelton:
The discussion we've had was to actually completely remove it. Don't believe that we every fully decided but there has been confusion around it and just adding a new parameter name I don't think will help.

If folks want to get backup history around an AG it deserves a dedicated command. I think that will make it more clear.

We use exclude parameters to filter objects in our module. If this was a test command then I'd see use for a parameter to skip some check or functionality but not with a Get function.

It will offer easier adoption and allow I think for separation of duties (so to speak) to just make a new command. Something like `Get-DbaAgBackupHistory`. We can also have better control of how it accepts instances with non-default ports potentially as well.

Just my money's worth...

andreasjordan:
I would vote for removing the AG-check-code from `Get-DbaDbBackupHistory`. That would speed up `Backup-DbaDatabase` as well. And although it is documented, checking other instances than the one you submit in the command is a bit "surprising".

I would also vote for a new command `Get-DbaAgBackupHistory`. There we could have a similar parameter set like in `Add-DbaAgDatabase` with an optional parameter `-Secondary` for those with custom ports. And we could have `-AvailabilityGroup` to get the history for all databases in that group.

The code would become more readable because we have no recursion anymore.

Should I try to create the new command - as part of my training in developing dbatools?

Just mention @potatoqualitee and @Stuart-Moore as authors of the command here.

andreasjordan:
@wsmelton Do you remember where the discussion took place and who was involved? I would like to push this issue a little bit...

wsmelton:
Slack with Stuart.

functions/Get-DbaDbBackupHistory.ps1
```diff
@@ -173,11 +179,12 @@ function Get-DbaDbBackupHistory {
         [switch]$LastLog,
         [string[]]$DeviceType,
         [switch]$Raw,
-        [bigint]$LastLsn,
```
Stuart-Moore, line 176: This stays, don't remove it

functions/Get-DbaDbBackupHistory.ps1
```diff
@@ -173,11 +179,12 @@ function Get-DbaDbBackupHistory {
...
-        [switch]$IncludeMirror,
```
Stuart-Moore, line 177: Dont remove this please

Stuart-Moore:
Rip it out

You're going to end up with some recursion in the new cmd  as you're  going to have to call Get-DbaDbBackuphistory across replicas anyway to make sure it's using the same filtering logic and output types.

Please don't try to rewrite that as we've just about got agreement on these methods after $DEITY knows how long :) and the wrong output type will bork the restore commands when using `TrustBackupHistory` options

ignore the review bits, was confused as you moved some stuff around and I'm suffering on bad hotel wifi

andreasjordan:
Started a brand new pull request: #6799 so closing this one.

wsmelton:
Are we not going to rip out the parameter and associated code?

@Stuart-Moore unless I'm misreading your post...

andreasjordan:
I will certainly continue to work on it and rip out the parameter and assiciated code. I just would like to have a confirmation that the new code does exactly what the old code did.
Or is it enough to be able to return to this commit point later? Then I will probably work on that today.

PR #6815, opened 2020-09-08
Change Message of Stop-Function to include $instance after failure of Connect-DbaInstance or Connect-SqlInstance

andreasjordan:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
### Purpose
Having more info for the user, especially when passing more instances to SqlInstance.

### Approach
I just used the message that was already used in most commands.

Sometimes also changed the Target to the correct value.
Sometimes added Category.
Sometimes changed dbmailRecord to ErrorRecord because Stop-Function does not have that parameter.


wsmelton:
This PR should have been discussed prior to altering our standard message that is used across every command in the module.

PR #6818, opened 2020-09-08
proposal for pester tests compatible with pester4-5

niphlod:
As discussed on Slack, weird proposal to make syntax more akin to what Pester4 expects (e.g. -Be instead of -Be) which on Pester5 is required.

Also, this apparently (hence the "proposal" name of the branch) makes it work in both Pester4 and Pester5.

Picked up a few things in the process:
- initialization needs to be in a BeforeAll
- no code should run in Context
- `$MyInvocation.MyCommand.Name` is not relied upon anymore
  - hence `$CommandName = (Get-Item $PSCommandPath).Name.Replace(".Tests.ps1", "")`
- `-Throw` matched with a "contains" approach on Pester4. Pester5 needs to include a wildcard such as in `-Like`
  - hence ```$PesterV5Throw = ''
    if ((Get-Command Invoke-Pester -ErrorAction SilentlyContinue).Version -ge '5.0.0') {
        $PesterV5Throw = '*'
    }``` , agreed, it can be more nice, but this is just for show
- validate params stealed from @wsmelton 's "new" method, seems perfectly legit
- setting costants.ps1 moved to the BeforeAll of integrationtests (it seems that setting it in the "global" BeforeAll doesn't work)

Think this as a starting point and not a finalized proposal, I'm sure more things will come up (e.g. for tests with mocks) but at least it enables who does tests to be more in line with what Pester4 started to introduce (leaving compat with Pester3 style, which a lot of our tests still rely upon) and what Pester5 expects (no backwards compat).
For the vast majority of our tests this kind of rewrite is not a big deal and moves us a lot forward.

Again, this is to start make moves to make sure we CAN test with Pester5 when such a need will arise. My only con right now with Pester5 is that it doesn't support coverage report.

### DIFF
--- a/tests/Invoke-DbaQuery.Tests.ps1
+++ b/tests/Invoke-DbaQuery.Tests.ps1
@@ -1,14 +1,18 @@
-$CommandName = $MyInvocation.MyCommand.Name.Replace(".Tests.ps1", "")
-Write-Host -Object "Running $PSCommandPath" -ForegroundColor Cyan
-. "$PSScriptRoot\constants.ps1"
+BeforeAll {
+    $CommandName = (Get-Item $PSCommandPath).Name.Replace(".Tests.ps1", "")
+    Write-Host -Object "Running $PSCommandPath" -ForegroundColor Cyan
+    $PesterV5Throw = ''
+    if ((Get-Command Invoke-Pester -ErrorAction SilentlyContinue).Version -ge '5.0.0') {
+        $PesterV5Throw = '*'
+    }
+}
 
 Describe "$CommandName Unit Tests" -Tag 'UnitTests' {
     Context "Validate parameters" {
-        [object[]]$params = (Get-Command $CommandName).Parameters.Keys | Where-Object { $_ -notin ('whatif', 'confirm') }
-        [object[]]$knownParameters = 'SqlInstance', 'SqlCredential', 'Database', 'Query', 'QueryTimeout', 'File', 'SqlObject', 'As', 'SqlParameters', 'AppendServerInstance', 'MessagesToOutput', 'InputObject', 'ReadOnly', 'EnableException', 'CommandType'
-        $knownParameters += [System.Management.Automation.PSCmdlet]::CommonParameters
         It "Should only contain our specific parameters" {
-            (@(Compare-Object -ReferenceObject ($knownParameters | Where-Object { $_ }) -DifferenceObject $params).Count ) | Should Be 0
+            [array]$params = ([Management.Automation.CommandMetaData]$ExecutionContext.SessionState.InvokeCommand.GetCommand($CommandName, 'Function')).Parameters.Keys
+            [object[]]$knownParameters = 'SqlInstance', 'SqlCredential', 'Database', 'Query', 'QueryTimeout', 'File', 'SqlObject', 'As', 'SqlParameters', 'AppendServerInstance', 'MessagesToOutput', 'InputObject', 'ReadOnly', 'EnableException', 'CommandType'
+            Compare-Object -ReferenceObject $knownParameters -DifferenceObject $params | Should -BeNullOrEmpty
         }
     }
     Context "Validate alias" {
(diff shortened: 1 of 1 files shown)


tests/Invoke-DbaQuery.Tests.ps1
```diff
@@ -1,14 +1,18 @@
...
+}
 
 Describe "$CommandName Unit Tests" -Tag 'UnitTests' {
     Context "Validate parameters" {
-        [object[]]$params = (Get-Command $CommandName).Parameters.Keys | Where-Object { $_ -notin ('whatif', 'confirm') }
-        [object[]]$knownParameters = 'SqlInstance', 'SqlCredential', 'Database', 'Query', 'QueryTimeout', 'File', 'SqlObject', 'As', 'SqlParameters', 'AppendServerInstance', 'MessagesToOutput', 'InputObject', 'ReadOnly', 'EnableException', 'CommandType'
-        $knownParameters += [System.Management.Automation.PSCmdlet]::CommonParameters
         It "Should only contain our specific parameters" {
-            (@(Compare-Object -ReferenceObject ($knownParameters | Where-Object { $_ }) -DifferenceObject $params).Count ) | Should Be 0
+            [array]$params = ([Management.Automation.CommandMetaData]$ExecutionContext.SessionState.InvokeCommand.GetCommand($CommandName, 'Function')).Parameters.Keys
```
wsmelton, line 13: You could probably move this to the BeforeAll could you not? (I see the test failed on it not seeing anything in the variable

niphlod: "global" scope. Lemme reconstruct the before-cleanup logic :D

niphlod:
yeah, so https://github.com/sqlcollaborative/dbatools/pull/6818#discussion_r485194623 problem is that BeforeAll outside a Context is only available for Pester5 (and it's the "recommended" section to do setups such as filename), Pester4 never runs it. 
Adding $CommandName directly in restores functionality, albeit when running with Pester4 you won't get 
`Describing Invoke-DbaDbQuery Unit Tests` but rather `Describing Unit Tests` (for the same reason, such as `$CommandName` is not available when printing the Context)

tests/Invoke-DbaQuery.Tests.ps1
```diff
@@ -60,11 +66,11 @@ Describe "$CommandName Integration Tests" -Tag "IntegrationTests" {
         foreach ($result in $results) {
             $result.TestColumn | Should -Be 'hello'
         }
-        'tempdb' | Should -Bein $results.dbname
+        'tempdb' | Should -BeIn $results.dbname
     }
     It "stops when piped databases and -Database" {
         $dbs = Get-DbaDatabase -SqlInstance $script:instance2, $script:instance3
-        { $dbs | Invoke-DbaQuery -Query "Select 'hello' as TestColumn, DB_NAME() as dbname" -Database tempdb -EnableException } | Should Throw "You can't"
+        { $dbs | Invoke-DbaQuery -Query "Select 'hello' as TestColumn, DB_NAME() as dbname" -Database tempdb -EnableException } | Should -Throw "You can't$PesterV5Throw"
```
wsmelton, line 73: I'm not fond of the way this looks and would just soon either switch everything to Pester 5 or hold off and leave it all at Pester 4.

niphlod: or we can just `Should -Throw`. In any case leaving this as "pester4-style" isn't a big deal, the move to pester5 will just have to append/prepend a *

potatoqualitee:
i have no good feedback :(

niphlod:
> i have no good feedback :(

meaning you don't like it or "I'm unable to rate this good vs bad" ?

Pester5 right now is a little slower and missing coverage. But it helps a lot in matter of scoping (Pester4 has some apparently unfixable drawbacks in that area, but we kinda "know" them all) and apparently mocking (which I have not experienced yet).
The fact that you can only run tests on PS5 IMHO is not that big of a deal.

So, IMHO, the only sane move now for dbatools is wait for Pester5 to be more "sound", but in the meantime we should still make sure at least we don't use Pester3 syntax (use `-Be` instead of `Be`) ... and we try to move most of the code running inside a `Context` (and not within an `It`) to `Before*/After*`).

wsmelton:
I will go ahead and as I touch a test update that unit test section with this format/syntax.

I'm up for waiting so will close this PR now.

PR #6819, opened 2020-09-08
Update-DbaServiceAccount, use FullComputerName

### LINKED ISSUE #6473: [Bug] Update-DbaServiceAccount not using FQDN
When I try to run Update-DbaServiceAccount with a FQDN (since I'm running this on a client in a different domain than the server), it loses the domain name in the FQDN and so it fails to connect since it can't find the server. 

Update-DbaServiceAccount - it uses ComputerName as opposed to FullComputerName in Get-DbaService. Per @nvarscar this is on lines 160 and 176.

niphlod:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #6473 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Make it work when passed an FQDN

### Approach
This command is a bit of a mashup because it can take multiple formats (also, Get-DbaService already resolves correctly fullcomputername, but it allows being passed also a generic object, not checking if the input is indeed already "processed" by Get-DbaService).

Problem described on the issue is that if you pass `-ComputerName foo.bar.baz` down the line Invoke-ManagedComputerCommand is left (possibly) with `foo` only. 
Whenever ever something must target the server rather than the instance, FullComputerName in the commands should always be used.
This is more evident with commands resolving the hostname from an instance  ... here it works taking a ComputerName instead of a SqlInstance, but you can still pass a SqlInstance to it. 
Rule of thumb (to simplify things) should be if a command like this doesn't stricly need a live instance to connect to, accept a ComputerName as a string and use it verbatim. No Resolve-DbaNetworkName whatsoever. But if more flexibility is added and Resolve-DbaNetworkName is needed, use FullComputerName if you need to operate on the host running the SQL Service.

The config in bypass (`Get-DbatoolsConfigValue -FullName commands.resolve-dbanetworkname.bypass`) was born because the operations to resolve properly all properties are rather convoluted to make it work in all cases (sql not being very collaborative and disjoint domains being the most "shitty situation").
Here we just have a client not running in the same domain, so FQDN is the "natural" way to pass a reference to a Computer (and an instance, if need be).
For people living in "netbios-resolves-all/never-gonna-give-anything-other-than-hostname" the bypass in config speeds things up because it just avoids doing the logic (returns every property as the string passed in, in this case it would remain `foo.bar.baz` for both ComputerName and FullComputerName).

So yes, it would fix it but again, I see no reason why Update-DbaServiceAccount should work in disjoint domains (or when client is on a different domain than server) only if a default setting is altered. 
This update should make it work in all cases (hopefully it'll pass tests, which I have reorganized)


potatoqualitee:
looks good!

potatoqualitee:
happy to have you back 🤗

PR #6824, opened 2020-09-09
New-/Get-DbaEndpoint - Added IPAddress and fixed SslPort

andreasjordan:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
I have a client who wants to use a dedicated network for hadr communication, so we bind the hadr_endpoint only on one ipadress in a special network. This parameter was missing, so I had to use SQL. I want to use dbatools the next time...

Also removed SslPort as this is not a property of the smo (try to use it and you'll see - or have a look at https://docs.microsoft.com/en-us/dotnet/api/microsoft.sqlserver.management.smo.tcpprotocol?view=sql-smo-160).

Changed the Get- to return a Fqdn with ipaddress if not 0.0.0.0, because this is most likely what is needed for Add-DbaAgReplica. But we can discuss that point.

I have not updated the tests because other pull requests are pending and they would conflict on the tests for parameters.

Other issues I found while coding:
```
                $thisport = (Get-DbaEndpoint -SqlInstance $server).Protocol.Tcp
                $measure = $thisport | Measure-Object ListenerPort -Maximum

                if ($thisport.ListenerPort -eq 0) {
                    $tcpPort = 5022
                } elseif ($measure.Maximum) {
                    $maxPort = $measure.Maximum
                    #choose a random port that is greater than the current max port
                    $tcpPort = $maxPort + (New-Object Random).Next(1, 500)
                } else {
                    $maxPort = 5000
                    #choose a random port that is greater than the current max port
                    $tcpPort = $maxPort + (New-Object Random).Next(1, 500)
                }
```
I think that this does not works as expected - I would just use 5022 as a default if mirroring and otherwise fail if port is not provided. With more than one instance on a system, the instance can not know if 5022 is not used by another instance on that machine. So just checking an local instance is just "weigh in safety" and not garantied to work.

```
                    if ($Certificate) {
                        $outscript = $endpoint.Script()
                        $outscript = $outscript.Replace("ROLE = ALL,", "ROLE = ALL, AUTHENTICATION = CERTIFICATE $cert,")
                        $server.Query($outscript)
```
This only works if $Role = 'All', otherwise the replace would not work.

For all that like Screenshots, here we go:

image

### DIFF
--- a/functions/New-DbaEndpoint.ps1
+++ b/functions/New-DbaEndpoint.ps1
@@ -28,11 +28,13 @@ function New-DbaEndpoint {
     .PARAMETER Role
         The type of role. Defaults to All. Options: All, None, Partner, Witness
 
+    .PARAMETER IPAddress
+        Specifies the IP address that the endpoint will listen on. The default is ALL. This means that the listener will accept a connection on any valid IP address.
+
+        Currently only IPv4 is supported by this command.
+
     .PARAMETER Port
         Port for TCP. If one is not provided, it will be auto-generated.
-
-    .PARAMETER SslPort
-        Port for SSL
 
     .PARAMETER Certificate
         Database certificate used for authentication.
@@ -87,7 +89,12 @@ function New-DbaEndpoint {
         PS C:\> New-DbaEndpoint -SqlInstance localhost\sql2017 -Type DatabaseMirroring -Port 5055
 
         Creates a database mirroring endpoint on localhost\sql2017 which uses alternative port 5055
-    #>
+
+    .EXAMPLE
+        PS C:\> New-DbaEndpoint -SqlInstance localhost\sql2017 -Type DatabaseMirroring -IPAddress 192.168.0.15 -Port 5055
+
+        Creates a database mirroring endpoint on localhost\sql2017 which binds only on ipaddress 192.168.0.15 and port 5055
+        #>
     [CmdletBinding(SupportsShouldProcess, ConfirmImpact = 'Low')]
     param (
         [parameter(Mandatory, ValueFromPipeline)]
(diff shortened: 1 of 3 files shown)


functions/New-DbaEndpoint.ps1
```diff
@@ -153,22 +160,18 @@ function New-DbaEndpoint {
                 }
             }
 
-            if ($Pscmdlet.ShouldProcess($server.Name, "Creating endpoint $Name of type $Type using protocol $Protocol and if TCP then using Port $tcpPort")) {
+            if ($Pscmdlet.ShouldProcess($server.Name, "Creating endpoint $Name of type $Type using protocol $Protocol and if TCP then using IPAddress $IPAddress and Port $tcpPort")) {
```
wsmelton, line 163: I would think this a bit confusing to some users seeing the IP in the WhatIf output if it was not provided:

image

Some of the WhatIf messages are not ideal in our commands, so you are welcome to update/rearrange this to be more informative if you want. I would not see adding "Creating" until the actual call to the if block for certificate and `.Create()` call. We could just have this split to say "Setting..." for setting protocol, owner, IP (if provided), etc. Then when it goes to actual create have the "Creating...". I'd also wrap the call to `Get-DbaEndpoint` if we chose to rearrange so it does not execute that command.

andreasjordan: But '0.0.0.0' is the equivalent for the default "ALL" and that's what is there inside the smo if you don't set it - try with the old version of the module. That's why I don't have a `if ($IPAddress) { ... }` in there but just set the default of '0.0.0.0'.
If you would like to rephrase the -WhatIf messages: No problem, just please open your own pull request for that.

potatoqualitee:
Just tested it and you're right - SSL port is not a part of TCP, it's part of HTTP. Would you be able to add it back and assign it to the appropriate protocol?

image

There seems to be some ongoing discord that I'd like to help resolve. Please ping me on Slack, @andreasjordan, when you have a PR to review and you and I can work together on it.

andreasjordan:
> Would you be able to add it back and assign it to the appropriate protocol?

Sure, will do this in the next hours.

potatoqualitee:
This looks good to me - that test is failing but it's the test that keeps failing :/

PR #6832, opened 2020-09-11
Build Ref update - SQL Server 2017 CU22

wsmelton:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [x] Build Reference Update
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Adding SQL Server 2017 CU22 to build ref


niphlod (approved):

niphlod:
aargh, that damn regr about get-dbabackuphistory fails (unrelated to this pr), but with good reason..... will investigate this evening (either the query is still borked or the setup part doesn't recreate the same condition between runs).

andreasjordan:
> aargh, that damn regr about get-dbabackuphistory fails (unrelated to this pr), but with good reason..... will investigate this evening (either the query is still borked or the setup part doesn't recreate the same condition between runs).

After each backup you should check for the LSN and do more work and backups until the LSN is high enough so that the restore to the old backup results in a lower LSN at the next backup (that's what I did manually). Keep in mind that while taking the backup the LSN increases again, so the LSN in the fist chain has to be much bigger than the one from the backup that you want to restore for test.

niphlod:
> > aargh, that damn regr about get-dbabackuphistory fails (unrelated to this pr), but with good reason..... will investigate this evening (either the query is still borked or the setup part doesn't recreate the same condition between runs).
> 
> After each backup you should check for the LSN and do more work and backups until the LSN is high enough so that the restore to the old backup results in a lower LSN at the next backup (that's what I did manually). Keep in mind that while taking the backup the LSN increases again, so the LSN in the fist chain has to be much bigger than the one from the backup that you want to restore for test.

yeah, figured as much, will inspect later ^_^

PR #6835, opened 2020-09-12
New-DbaEndPoint: removed sslport and http options

potatoqualitee:
It didn't work anyway. Usually I try to support old stuff but it would be too much work for no payoff. I had to go back to SQL 2008 to create an HTTP endpoint and even then it yelled at me. 

image

Now I'm unsure if Via, SharedMemory and NamedPipes belongs there either.

### DIFF
--- a/functions/New-DbaEndpoint.ps1
+++ b/functions/New-DbaEndpoint.ps1
@@ -20,7 +20,7 @@ function New-DbaEndpoint {
         The name of the endpoint. If a name is not specified, one will be auto-generated.
 
     .PARAMETER Type
-        The type of endpoint. Defaults to DatabaseMirroring. Options: DatabaseMirroring, ServiceBroker, Soap, TSql
+        The type of endpoint. Defaults to DatabaseMirroring. Options: DatabaseMirroring, ServiceBroker, TSql
 
     .PARAMETER Protocol
         The type of protocol. Defaults to tcp. Options: Tcp, NamedPipes, Via, SharedMemory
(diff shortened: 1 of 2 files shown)


potatoqualitee:
Accctually! SOAP seems to be the issue, not HTTP? Continuing to research

image

potatoqualitee:
Yep, so far, it's SOAP

image

potatoqualitee:
K, so far, the protocols are looking good 

image

But to properly support things fully, we need to add the following for HTTP

````
AuthenticationRealm     Property   string AuthenticationRealm {get;set;}
ClearPort               Property   int ClearPort {get;set;}
DefaultLogonDomain      Property   string DefaultLogonDomain {get;set;}
HttpAuthenticationModes Property   Microsoft.SqlServer.Management.Smo.HttpAuthenticationModes HttpAuthenticationModes {get;set;}
HttpPortTypes           Property   Microsoft.SqlServer.Management.Smo.HttpPortTypes HttpPortTypes {get;set;}
IsCompressionEnabled    Property   bool IsCompressionEnabled {get;set;}
SslPort                 Property   int SslPort {get;set;}
UserData                Property   System.Object UserData {get;set;}
WebSite                 Property   string WebSite {get;set;}
WebSiteUrlPath          Property   string WebSiteUrlPath {get;set;}
````

TCP is already taken care of 👍

potatoqualitee:
gotta just rework new-dbaendpoint, closing

wsmelton:
Found an example one for http if you need something to play against to make sure what is created works...

https://mobile.developer.com/net/asp/article.php/3767311/Creating-Native-Web-Services-in-SQL-Server.htm

Gives example of what they create and set on the endpoint.

potatoqualitee:
Thank you, @wsmelton ! I'm looking for an HTTP endpoint that doesn't use SOAP but rather, TCP. I cannot find one example on the internet. This would make great livestream content, actually. I think I'll do some CASP studying today then livestream how I'm gonna figure this out.

wsmelton:
Why would you use one over to?

potatoqualitee:
whatcha mean? no idea of the use-case. It's just possible, so I'd like to enable it.

wsmelton:
Use it http over tcp?

It may let you but those two protocols can't mix?

The only use cases for http is hitting API or I've seen some with reporting services. None showed using TCP with http.

Might ask on MVP DL... If it used I'm sure someone on that thread can share some info.

PR #6840, opened 2020-09-14
Get-DbaEndpoint - Some suggestions for improvements

andreasjordan:
Not a real pull request, only some ideas I want to show to @potatoqualitee. Is there a better way than a pull request?

I tested this code with https://dbatools.io/docker/ and it works.

Additional comment:
If the target is a clustered instance, then we have to make sure not to use the name of the physical computer but the virtual name. 
Based on this lonely source (don't know if we can trust) the serverproperty('MachineName') is .NetName:
https://social.technet.microsoft.com/Forums/en-US/6238806a-8f09-4594-95c2-5fcb923a72fd/obtain-machine-name-from-sql-server-instance-name?forum=sqlsmoanddmo
I still have no failover cluster instance to test...

### DIFF
--- a/functions/Get-DbaEndpoint.ps1
+++ b/functions/Get-DbaEndpoint.ps1
@@ -66,24 +66,18 @@ function Get-DbaEndpoint {
                 Stop-Function -Message "Error occurred while establishing connection to $instance" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
             }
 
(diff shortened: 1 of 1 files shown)


potatoqualitee:
lemme boot up my cluster and test machinename 💯

potatoqualitee:
this is way better! works for docker and machinename does work for clusters, though not for azure. azure shows the proper computername, however.

PR #6842, opened 2020-09-14
Connect-DbaInstance - Added output to -Debug

andreasjordan:
Ok, @potatoqualitee here is the pull request. Please play around with it, it is very interesting, I've learned a lot.

Sorry, now time for more info or screenshots. Just connect to some instances:
```
$server = Connect-DbaInstance -SqlInstance $instance -SqlCredential $cred -Debug
```

### DIFF
--- a/functions/Connect-DbaInstance.ps1
+++ b/functions/Connect-DbaInstance.ps1
@@ -300,8 +300,7 @@ function Connect-DbaInstance {
         [string]$Thumbprint = (Get-DbatoolsConfigValue -FullName 'azure.certificate.thumbprint'),
         [ValidateSet('CurrentUser', 'LocalMachine')]
         [string]$Store = (Get-DbatoolsConfigValue -FullName 'azure.certificate.store'),
-        [switch]$DisableException,
-        [switch]$VeryVerbose
+        [switch]$DisableException
     )
     begin {
         $azurevm = Get-DbatoolsConfigValue -FullName azure.vm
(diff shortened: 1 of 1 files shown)


functions/Connect-DbaInstance.ps1
```diff
@@ -300,7 +300,8 @@ function Connect-DbaInstance {
         [string]$Thumbprint = (Get-DbatoolsConfigValue -FullName 'azure.certificate.thumbprint'),
         [ValidateSet('CurrentUser', 'LocalMachine')]
         [string]$Store = (Get-DbatoolsConfigValue -FullName 'azure.certificate.store'),
-        [switch]$DisableException
+        [switch]$DisableException,
+        [switch]$VeryVerbose
```
wsmelton, line 304: We have support for this within our message system.

You can add this value as a message type to the Write-Message call instead of using a parameter on the command itself.

andreasjordan: Ok, good to know. Can you give me a hint where to find more info or an example?
And someone should rephrase some of the messages if this should ever go into production.
This is only a first step to give more insight about how and where the values are set.

andreasjordan:
Ok, I have a screenshot now:

image

If I connect using a full name "sqlix.ordix.de", then the returned .ComputerName is just the .NetName which is just the hostname without the domain. 
But I think I've seen screenshots with a full name returned as .ComputerName. And I would expect to have the $instance.ComputerName to be transfered to $server.ComputerName, because if I need the .NetName I can just use it as it is part of the official smo.

If someone could try this on other environments and send screenshots or outputs that would be great.

wsmelton:
I'd have to go back through our git log (GitHub doesn't show all file history before of the file renamed we did) but I recall we intially started out using netname but in domain environments that have multiple forest and domains it will be the wrong value and not the machine the user actually connected with or something.

Fairly sure we have some conversation recorded on here but I'll have to go looking through history.

In reference to the messaging system that is all written up in the wiki: https://github.com/sqlcollaborative/dbatools/wiki/Implementing-the-messaging-system (this is Fred's brain child and equivalent to what you get in PSFrameworks module).

wsmelton:
I've probably got something turned around on the netname. We went through many iterations trying to fix massive bugs in that connect command so not sure why we ended up doing some things.

andreasjordan:
Ok, I used:
```
        Write-Message -Level VeryVerbose -Message "Starting process block veryverbose"
        Write-Message -Level SomewhatVerbose -Message "Starting process block SomewhatVerbose"
```
But then they show up in "normal" -Verbose, which I don't want to.

So I will change everything to Debug...

wsmelton:
Oh yeah forgot about that being treated as all verbose.

That data gets written to our log file but not sure we have mechanism to keep it off host.

andreasjordan:
Added more info to show that $instance.FullSmoName is used in the smo constructor and so goes into $server.Name - so .Name has the full name when the $instance was build with the full name.

potatoqualitee:
Awesome, PR, thank you @andreasjordan ! This was very helpful.

PR #6844, opened 2020-09-14
Set-DbaExtendedProtection, shush and speed it

niphlod:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Tests for this lasted too damn long. 

### Approach
The only thing we tested was for the same result given different parameters for "Value", which the command accept both "text" and "numeric".
For 7 times (3 parameters x 2 "types" plus the default) the test ran, basically to do the same exact thing EXCEPT setting the proper value in the registry.
But we paid a hefty price.... we basically just needed to test that the reg value is set accordingly (multiple tests aggravated by querying WMI bening too damn slow).
So, change of plans :D 
We just test if the command works ONCE, with the default param.
And given we can't use classes as params (like enums, because of PS5), we just mock the s**t around it.

I was lucky already because the passed Value is outputted in the Verbose Stream (L134), so we check that.

I reintroduced also Test-ShouldProcess (https://github.com/sqlcollaborative/dbatools/pull/3325 , hopefully just "lost in translation" and not "removed for a good reason"), which we can definitely not enforce in any command but it's the only way to run pester tests (in this case definitely shushing the tests, but for things more complicated it makes perfect sense). 
Prolly I'll give the same treatment to other "big talkers" (Invoke-DbaDiagnosticQuery ^_^)

### Screenshots
WAS
immagine

IS
immagine

Also, took
immagine

and now takes
immagine

### DIFF
--- a/tests/Set-DbaExtendedProtection.Tests.ps1
+++ b/tests/Set-DbaExtendedProtection.Tests.ps1
@@ -19,28 +19,28 @@ Describe "$commandname Integration Tests" -Tags "IntegrationTests" {
             $results = Set-DbaExtendedProtection -SqlInstance $script:instance1 -EnableException *>$null
             $results.ExtendedProtection -eq "0 - Off"
         }
-	}
-	Context "Command works when passed different values" {
-		BeforeAll {
-			Mock Test-ShouldProcess { $false } -ModuleName dbatools
+    }
+    Context "Command works when passed different values" {
+        BeforeAll {
+            Mock Test-ShouldProcess { $false } -ModuleName dbatools
             Mock Invoke-ManagedComputerCommand -MockWith {
-				param (
-				$ComputerName,
-				$Credential,
-				$ScriptBlock,
-				$EnableException
-			)
-			   $server = [DbaInstanceParameter[]]$script:instance1
-			   @{
-					DisplayName = "SQL Server ($($instance.InstanceName))"
-					AdvancedProperties = @(
-							@{
-								Name = 'REGROOT'
-								Value = 'Software\Microsoft\Microsoft SQL Server\MSSQL10_50.SQL2008R2SP2'
-							}
-						)
-				}
-			} -ModuleName dbatools
+                param (
+                $ComputerName,
+                $Credential,
+                $ScriptBlock,
+                $EnableException
+            )
+               $server = [DbaInstanceParameter[]]$script:instance1
+               @{
+                    DisplayName = "SQL Server ($($instance.InstanceName))"
+                    AdvancedProperties = @(
+                            @{
+                                Name = 'REGROOT'
+                                Value = 'Software\Microsoft\Microsoft SQL Server\MSSQL10_50.SQL2008R2SP2'
+                            }
+                        )
+                }
+            } -ModuleName dbatools
         }
         It "Set explicitly to '0 - Off' using text" {
             $results = Set-DbaExtendedProtection -SqlInstance $script:instance1 -Value Off -EnableException -Verbose 4>&1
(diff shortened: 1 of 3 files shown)


wsmelton:
We can test for WhatIf in our tests (and actually should be in some of our major commands). 

I'm not sure I understand the purpose of the internal command for `Test-ShouldProcess`. I don't recall that ever existing before to have been removed.

tests/Set-DbaExtendedProtection.Tests.ps1
```diff
@@ -16,34 +16,57 @@ Describe "$CommandName Unit Tests" -Tag 'UnitTests' {
...
         It "Default set and returns '0 - Off'" {
-            $results = Set-DbaExtendedProtection -SqlInstance $script:instance1 -EnableException
+            $results = Set-DbaExtendedProtection -SqlInstance $script:instance1 -EnableException *>$null
             $results.ExtendedProtection -eq "0 - Off"
         }
+    }
+    Context "Command works when passed different values" {
+        BeforeAll {
+            Mock Test-ShouldProcess { $false } -ModuleName dbatools
+            Mock Invoke-ManagedComputerCommand -MockWith {
```
wsmelton, line 26: Is this no longer an integration test now that we are mocking the main call by the command and manipulating what the output will look like for users that want to run our test framework locally?

wsmelton:
I can see why the time changed so drastically because you took out the call to Invoke-ManagedCommand, but mocking that call it is no longer an integration test.

Any function we have that calls `Invoke-Command2` can speed up by changing our default configuration values for certain things. I took your test but removed all the mocks so the command is actually tested against a live instance. I replaced the `BeforeAll` with this:

```
        BeforeAll {
            $WhatIfPreference = $false
            Set-DbatoolsConfig -FullName 'commands.resolve-dbanetworkname.bypass' -Value $true
            Set-DbatoolsConfig -FullName 'computermanagement.badconnectiontimeout' -Value '00:00:05'
            Set-DbatoolsConfig -FullName 'computermanagement.cache.disable.all' -Value $true
        }
```

And had these results:

_No BeforeAll block, no mocks_

image

Added the `BeforeAll` block above:

image

niphlod:
> I'm not sure I understand the purpose of the internal command for `Test-ShouldProcess`. I don't recall that ever existing before to have been removed.

yeah, if you look at https://github.com/sqlcollaborative/dbatools/pull/3325 we agreed upon it and then the PR remained unmerged. Long story short, Pester can't mock ShouldProcess. Also, output from shouldprocess cannot be intercepted. Granted, right now the function works exactly as it was, with the only plus side of being completely silent when tests are run. But it could be used to test if the message of ShouldProcess actually is consistent....which I cannot find a way to do it if we don't use Test-ShouldProcess


> We can test for WhatIf in our tests (and actually should be in some of our major commands).

hum, couldn't find that much but I confess I just did a quick search



> I can see why the time changed so drastically because you took out the call to `Invoke-Command2`, but mocking that call it is no longer an integration test.

it still needs a live instance, or we need to mock other things. I just mocked the WMI call, and yes, unfortunately it still is not a unittest. I'm just mocking the most expensive call for the function, not mocking all.
The point is that the function does a lot of things (read, spend a lot of time) just to figure out where to put the value in the registry.
There's a call untouched and unmocked for testing that the whole logic works.
IMHO there's no added value on testing that logic 6 more times just to make sure that when `-Value 0` is passed the same value ends in the same place of when passed `-Value Off`, which is what the 6 mocked calls are for (i.e. just assuring parameters are handled correctly `0 { "Off" } 1 { "Allowed" } 2 { "Required" }`).

potatoqualitee:
@wsmelton please let me know when you are happy with this, and we can merge

potatoqualitee:
gonna go ahead and merge this bad boy

PR #6846, opened 2020-09-16
Find dba orphaned file recurse

nmbell:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Add a new parameter, -Recurse, to Find-DbaOrphanedFile that will enable recursive search of 
directories specified in the -Path parameter.

### Approach

Update the logic to control the value of the 'depth' parameter passed to xp_dirtree when recursion is required (i.e. 0). 
Also update the SQL query to construct the full paths from the results returned.

### Commands to test

Find-DbaOrphanedFile

### DIFF
--- a/functions/Find-DbaOrphanedFile.ps1
+++ b/functions/Find-DbaOrphanedFile.ps1
@@ -150,8 +150,8 @@ function Find-DbaOrphanedFile {
                 )
                 SELECT DISTINCT f.fullpath
                 FROM fullpaths AS f
-                WHERE f.fs_filename NOT IN( 'xtp', '5', '`$FSLOG', '`$HKv2', 'filestream.hdr', '"+ $($SystemFiles -join "','") + "' )
-                AND f.fs_fileextension IN('"+ $($FileTypes -join "','") + "')
+                WHERE f.fs_filename NOT IN( 'xtp', '5', '`$FSLOG', '`$HKv2', 'filestream.hdr', '" + $($SystemFiles -join "','") + "' )
+                AND f.fs_fileextension IN('" + $($FileTypes -join "','") + "')
                 AND f.is_file = 1;
                 "
 
@@ -232,7 +232,7 @@ function Find-DbaOrphanedFile {
             }
 
             # Reset all the arrays
-            $sqlpaths = $userpaths = @(); $dirtreefiles = @{}
+            $sqlpaths = $userpaths = @(); $dirtreefiles = @{ }
 
             # Gather a list of files known to SQL Server
             $sqlfiles = Get-SqlFileStructure $server
--- a/functions/Find-DbaOrphanedFile.ps1
+++ b/functions/Find-DbaOrphanedFile.ps1
@@ -107,7 +107,7 @@ function Find-DbaOrphanedFile {
 
     begin {
         function Get-SQLDirTreeQuery {
-            param($SqlPathList,$UserPathList,$FileTypes,$SystemFiles,[Switch]$Recurse)
+            param($SqlPathList, $UserPathList, $FileTypes, $SystemFiles, [Switch]$Recurse)
 
             $q1 = "
                 CREATE TABLE #enum (
@@ -156,10 +156,10 @@ function Find-DbaOrphanedFile {
 
             # build the query string based on how many directories they want to enumerate
             $sql = $q1
-            $sql += $($SqlPathList | Where-Object { $_ -ne '' } | ForEach-Object { "$([System.Environment]::Newline)$($q2.Replace('dirname',$_).Replace('recurse','1'))" })
+            $sql += $($SqlPathList | Where-Object { $_ -ne '' } | ForEach-Object { "$([System.Environment]::Newline)$($q2.Replace('dirname',$_).Replace('recurse','1'))" } )
             If ($UserPathList) {
                 $recurseVal = If ($Recurse) { '0' } Else { '1' }
-                $sql += $($UserPathList | Where-Object { $_ -ne '' } | ForEach-Object { "$([System.Environment]::Newline)$($q2.Replace('dirname',$_).Replace('recurse',$recurseVal))" })
+                $sql += $($UserPathList | Where-Object { $_ -ne '' } | ForEach-Object { "$([System.Environment]::Newline)$($q2.Replace('dirname',$_).Replace('recurse',$recurseVal))" } )
             }
             $sql += $query_files_sql
             Write-Message -Level Debug -Message $sql
@@ -199,7 +199,7 @@ function Find-DbaOrphanedFile {
 
             return $dbfiletable.Tables.Filename | ForEach-Object {
                 [PSCustomObject]@{
-                    Filename = $_
+                    Filename       = $_
                     ComparisonPath = [IO.Path]::GetFullPath($(Format-Path $_))
                 }
             }
@@ -255,7 +255,7 @@ function Find-DbaOrphanedFile {
             $sql = Get-SQLDirTreeQuery -SqlPathList $sqlpaths -UserPathList $userpaths -FileTypes $fileTypeComparison -SystemFiles $systemfiles -Recurse:$Recurse
             $dirtreefiles = $server.Databases['master'].ExecuteWithResults($sql).Tables[0] | ForEach-Object {
                 [PSCustomObject]@{
-                    Fullpath = $_.fullpath
+                    Fullpath       = $_.fullpath
                     ComparisonPath = [IO.Path]::GetFullPath($(Format-Path $_.fullpath))
                 }
             }
@@ -263,28 +263,28 @@ function Find-DbaOrphanedFile {
             # Output files in the dirtree not known to SQL Server
             foreach ($file in $dirtreefiles | Where-Object { $_.ComparisonPath -notin $sqlfiles.ComparisonPath }) {
 
-                    $result = [PSCustomObject]@{
-                        Server         = $server.Name
-                        ComputerName   = $server.ComputerName
-                        InstanceName   = $server.ServiceName
-                        SqlInstance    = $server.DomainInstanceName
-                        Filename       = $file.Fullpath
-                        RemoteFilename = Join-AdminUnc -Servername $server.ComputerName -Filepath $file.Fullpath
-                    }
-
-                    if ($LocalOnly -eq $true) {
-                        $result | Select-Object -ExpandProperty Filename
-                        continue
-                    }
-
-                    if ($RemoteOnly -eq $true) {
-                        $result | Select-Object -ExpandProperty RemoteFilename
-                        continue
-                    }
-
-                    $result | Select-DefaultView -Property ComputerName, InstanceName, SqlInstance, Filename, RemoteFilename
-
+                $result = [PSCustomObject]@{
+                    Server         = $server.Name
+                    ComputerName   = $server.ComputerName
+                    InstanceName   = $server.ServiceName
+                    SqlInstance    = $server.DomainInstanceName
+                    Filename       = $file.Fullpath
+                    RemoteFilename = Join-AdminUnc -Servername $server.ComputerName -Filepath $file.Fullpath
                 }
+
+                if ($LocalOnly -eq $true) {
+                    $result | Select-Object -ExpandProperty Filename
+                    continue
+                }
+
+                if ($RemoteOnly -eq $true) {
+                    $result | Select-Object -ExpandProperty RemoteFilename
+                    continue
+                }
+
+                $result | Select-DefaultView -Property ComputerName, InstanceName, SqlInstance, Filename, RemoteFilename
+
+            }
 
         }
     }
--- a/functions/Find-DbaOrphanedFile.ps1
+++ b/functions/Find-DbaOrphanedFile.ps1
@@ -232,7 +232,7 @@ function Find-DbaOrphanedFile {
             }
 
             # Reset all the arrays
-            $sqlpaths = @(); $dirtreefiles = @{}
+            $sqlpaths = $userpaths = @(); $dirtreefiles = @{}
 
             # Gather a list of files known to SQL Server
             $sqlfiles = Get-SqlFileStructure $server
--- a/functions/Find-DbaOrphanedFile.ps1
+++ b/functions/Find-DbaOrphanedFile.ps1
@@ -150,8 +150,8 @@ function Find-DbaOrphanedFile {
                 )
                 SELECT DISTINCT f.fullpath
                 FROM fullpaths AS f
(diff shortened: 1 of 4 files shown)


potatoqualitee:
Thanks so much for the PR, @nmbell ! Unfortunately, the test failed and I couldn't immediately identify the cause :/

nmbell:
Hi Chrissy

Thanks for reaching out. No problem, I’m still working on it. I think I have a fix; I’ll commit this evening after work and we’ll see what happens 😊.

Nathan

nmbell:
OK, I'm stumped. If anyone could provide any insight as to why the build is failing, it'd be much appreciated:

https://ci.appveyor.com/project/sqlcollaborative/dbatools/builds/35288490/job/nf9eu0kyhl3wi3la
https://ci.appveyor.com/project/sqlcollaborative/dbatools/builds/35288714/job/in5lk63ss48fsiea
Build is failing because either the database isn't getting created, or the function isn't finding the files. I've run the function on my laptop, my workstation, and the QA server at work, both as it would normally be used and by running the test line-by-line in the console, and it gives me exactly the expected results, no errors.

https://ci.appveyor.com/project/sqlcollaborative/dbatools/builds/35288714/job/mqux8564ukuiyrm2
"Find-DbaOrphanedFile.ps1 is not compliant with the OTBS formatting style. Please run Invoke-DbatoolsFormatter against the failing file and commit the changes."
I've run Invoke-DbatoolsFormatter against the file but I'm still getting this error and I'm not sure why. I've run Invoke-ScriptAnalyzer -Settings CodeFormattingOTBS but it doesn't return anything.

wsmelton:
Can the test be run against your test environment?

You can create a constants.local.ps1 in the root directory of the repo and set those instances to your environment and then our tests use those instances.

Generally if the test is successful locally it will run successful in appveyor...generally.

wsmelton:
According to the output in the link the tests database is trying to create the log file over the admin share...

```
Running C:\github\dbatools\tests\Find-DbaOrphanedFile.Tests.ps1

64CREATE DATABASE [dbatoolsci_orphanedfile] ON PRIMARY (NAME = [dbatoolsci_orphanedfile_DATA], FILENAME = 'C:\Program Files\Microsoft SQL Server\MSSQL13.SQL2016\MSSQL\DATA\dbatoolsci_orphanedfile_DATA.mdf') LOG ON (NAME = [dbatoolsci_orphanedfile_TLOG], FILENAME = '\\APPVYR-WIN\C$\Program Files\Microsoft SQL Server\MSSQL13.SQL2016\MSSQL\DATA\dbatoolsci_orphanedfile_TLOG.ldf')

65Running C:\github\dbatools\tests\Find-DbaStoredProcedure.Tests.ps1

```

niphlod:
tests are a bit funky but let's go over the changes first .... why should the instance enum the admin shares ? there are shops with servers that have no admin shares available, and this will surely break.
So, IMHO:
- no need for the whole comparisonpath logic
- no need for split-adminunc at all
- mssql should enum the dirs as the instance sees them, if they're local, they are local and they don't have to be translated to an unc path (also because all your translations are done "hoping" that admin shares are actually available)

PS: Join-AdminUNC has been created to make life to users easier if they DO have admin shares available, but it's not something our functions use outside of the realm of the PS session: in this case   MSSQL can't be assured to be able to enumerate it and, sincerely, I don't find a reason why you changed the defaults from the fullpath to the admin-share-representation-of-it pretty much everywhere.

niphlod:
@nmbell : give me a shout if you can't figure it out, I streamlined your code and it all seems to work as it should, tests too.

nmbell:
In SSMS, if I run:

`USE [master]`
`GO`
`CREATE DATABASE [dbatoolsci_orphanedfile] ON`
`PRIMARY (NAME = [dbatoolsci_orphanedfile_DATA], FILENAME = 'C:\Program Files\Microsoft SQL Server\MSSQL14.SQL2017DEVELOPER\MSSQL\DATA\dbatoolsci_orphanedfile_DATA.mdf')`
`LOG ON  (NAME = [dbatoolsci_orphanedfile_TLOG], FILENAME = '\\MYCOMPUTER\C$\Program Files\Microsoft SQL Server\MSSQL14.SQL2017DEVELOPER\MSSQL\DATA\dbatoolsci_orphanedfile_TLOG.ldf')`

then:

`USE [dbatoolsci_orphanedfile]`
`GO`
`SELECT [name], state_desc FROM sys.databases WHERE database_id = DB_ID()`
`SELECT physical_name FROM [dbatoolsci_orphanedfile].sys.database_files`

I get:

image

Then running the original version of Find-DbaOrphanedFiles:

`Find-DbaOrphanedFile -SqlInstance 'MYCOMPUTER\SQL2017DEVELOPER'`

Returns the following:

`ComputerName   : THINKPADP40YOGA`
`InstanceName   : SQL2017DEVELOPER`
`SqlInstance    : THINKPADP40YOGA\SQL2017DEVELOPER`
`Filename       : C:\Program Files\Microsoft SQL Server\MSSQL14.SQL2017DEVELOPER\MSSQL\DATA\dbatoolsci_orphanedfile_TLOG.ldf`
`RemoteFilename : \\THINKPADP40YOGA\C$\Program Files\Microsoft SQL Server\MSSQL14.SQL2017DEVELOPER\MSSQL\DATA\dbatoolsci_orphanedfile_TLOG.ldf`

which is incorrect. Why is this? In the function, the calls to Get-SqlDefaultPaths returns only a local path. The comparison is only matching the paths as text strings rather than e.g. using Get-Item to compare the file _objects_ (which I imagine would create a huge performance penalty). The file paths need to be translated to some common format (which I believe was also the point of the Comparison (now ComparisonPath) property). I suppose I could have gone the other way and converted everything to local paths, but UNC's are, well, universal. 

When you say 'admin shares', are you thinking specifically of folders that have been setup as a 'network share'? I don't know about older servers, but on every machine I've worked on (Windows Server 2008 R2 I think is the oldest version for me), you can refer to any file or folder with its UNC path, regardless of whether it's available via an explicit 'shared folder' or not. Once the function has gathered the file paths (in whatever form they are returned), they are thereafter only treated as strings. There is no need to hope the shares are available, because it's irrelevant. The user has the choice of working with either Filename or RemoteFilename in the output as suits their needs.

It's a fair point if the function is no longer compatible with SQL 2000, however my goal is only to present working code for consideration. If the project administrators decide that backward compatibility is more valuable than the new functionality, then fair enough. I could only hope that at some point in the future the scales would tip the other way.

nmbell:
I fixed the issue with the build failing due to formatting. I needed to downgrade my version of PSScriptAnalyzer from 1.19.1 to 1.18.2. I tried running the tests locally with Pester 5.0.4 and PSScriptAnalyzer 1.19.1 (latest versions at time of writing) and got the following formatting errors (edited for clarity):

`PS>& 'D:\Users\...\Documents\WindowsPowerShell\Modules\dbatools\tests\dbatools.Tests.ps1'`
`Running D:\Users\...\Documents\WindowsPowerShell\Modules\dbatools\tests\dbatools.Tests.ps1`
`Starting discovery in 1 files.`
`Running D:\Users\...\Documents\WindowsPowerShell\Modules\dbatools\tests\dbatools.Tests.ps1`
`Discovery finished in 217.94s.`
`[-] dbatools style.formatting.Get-DbaDbFile.ps1 is not compliant with the OTBS formatting style. Please run I...`
`[-] dbatools style.formatting.Get-DbaFilestream.ps1 is not compliant with the OTBS formatting style. Please r...`
`[-] dbatools style.formatting.Get-DbaPfAvailableCounter.ps1 is not compliant with the OTBS formatting style. ...`
`[-] dbatools style.formatting.Test-DbaDeprecatedFeature.ps1 is not compliant with the OTBS formatting style. ...`
`[-] dbatools style.formatting.Test-DbaDiskAlignment.ps1 is not compliant with the OTBS formatting style. Plea...`
`[-] dbatools style.formatting.Set-FileSystemSetting.ps1 is not compliant with the OTBS formatting style. Plea...`
`[-] dbatools style.formatting.Update-ServiceStatus.ps1 is not compliant with the OTBS formatting style. Pleas...`

It seems like the newer version of PSScriptAnalyzer would break the build if it was upgraded on the build server. @potatoqualitee: just thinking, rather than failing the build because of incorrect formatting, what about applying formatting after a successful build? Could you trigger it off commits to the main development branch? That way contributors don't have to worry about whether a difference in their local version will break the build off their pull request commits, and the clonable/distributable code base always complies with whichever version is running on the build server.

niphlod:
> When you say 'admin shares', are you thinking specifically of folders that have been setup as a 'network share'? I don't know about older servers, but on every machine I've worked on (Windows Server 2008 R2 I think is the oldest version for me), you can refer to any file or folder with its UNC path, regardless of whether it's available via an explicit 'shared folder' or not. Once the function has gathered the file paths (in whatever form they are returned), they are thereafter only treated as strings. There is no need to hope the shares are available, because it's irrelevant. The user has the choice of working with either Filename or RemoteFilename in the output as suits their needs.

Meddling with `HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Services\LanmanServer\Parameters\AutoShareWks` or simply not being an administrator on the server renders `\\computername\diskname$` unusable (hello, hardening).

> There is no need to hope the shares are available, because it's irrelevant

Huh oh... They absolutely need to be available (and that was the point of my previous comment) !

I get you wanted an universal representation of a path, but you are making MSSQL use `\\computername\diskname$` to enumerate dirs, which it simply cannot do on all servers: let's suppose master is on `D:\SQLData` dir, you're invoking `xp_dirtree \\computername\d$\SQLData`. Not only it doesn't work always (for hardening but also in other cases, like disjoint domains.... a better implementation would enumerate on localhost)....it's really not needed at all (or doesn't seems so from the code you added). 
Again, I really not see the added benefit of forcing all paths in the code to assume the `\\computername\disk$` format over just local plain fullpaths .... if you can name one, this is the time :D 

> It's a fair point if the function is no longer compatible with SQL 2000, however my goal is only to present working code for consideration. If the project administrators decide that backward compatibility is more valuable than the new functionality, then fair enough. I could only hope that at some point in the future the scales would tip the other way.

If you look at the function, yes, we do support SQL2000. The new feature is cool, but doesn't strictly *need* a cte to work: it's perfectly achievable without.

potatoqualitee:
Thanks everyone for your participation 💯 ! I appreciate the attention to detail and agree with @niphlod's suggestions. I also like the idea of auto-formatting on behalf of the user, but I'm not sure leaving GitHub Actions/additional commits to fix it is ideal. @niphlod your thoughts?

niphlod:
about formatting: the point is always the same....assure a consistent formatting throughout the module. Everyone has their IDEs and not every IDE is settable to format-on-save on a single style.

We just leverage the one from Invoke-ScriptAnalyzer because THAT one is consistent :D 
I'm not sure if a pre-commit-hook can be set on git to avoid pushing unformatted files.
 
Anywayyyyy....I'm going to investigate why scriptanalyzer changed the defaults (maybe it's just a `@{}` vs `@{ }` type of thing), but the point is that we DO have Invoke-Dbatoolsformatter .... and that we're one call of Invoke-Dbatoolsformatter away to consistently change one file style to a single one (and even the whole module if the new OTSB replaces all occurrences of `@{}` to `@{ }`).

As for "failing fast & providing a button to fix it" or "not failing and making sure who merges does it instead of the user", is just a matter of additional burden on the merger. We chose to fail fast to avoid the burden on mergers.
BTW, a few weeks ago on slack I was advocating to add a call to Invoke-Dbatoolsformatter to manual.pester.ps1, so at least users running tests will have the file formatted automatically before submitting.

nmbell:
The function was gathering a list of directory paths in $sqlpaths which is passed in to Get-SQLDirTreeQuery, but if any of those directories, which are passed to xp_dirtree, are in UNC format, the server won't return results if it doesn't support UNC paths. OK, now I see why it wouldn't work. I've updated the code to guarantee only local paths are passed in.

Without the Split-AdminUNC function, or some equivalent, I don't think there's any way for this:
image
to get to this:
`Filename : C:\Program Files\Microsoft SQL Server\MSSQL14.SQL2017DEVELOPER\MSSQL\DATA\dbatoolsci_orphanedfile_TLOG.ldf`
I think if you accept that SQL allows for specifying UNC paths, then converting back to a local path format is a necessity to maintain the function output.

Regarding recursion not strictly needing a CTE, I'll leave that for someone else to tackle. For now I've updated to keep the original behavior on pre-2005 versions.

And finally, a successful build :-). Creating a database with a UNC file path obviously doesn't work on the build server, so this will just have to remain a testing gap. Test of the recursive logic is working good, though :-).

nmbell:
My point about the formatting wasn't about whether it should be the committer or the merger that applies formatting (the issue is just as likely to occur if the merger's version of PSScriptAnalyzer is not the same as the build server, isn't it?). Having the build machine apply the formatting after a successful build side-steps the issue entirely: it will always be correct and never a burden for anyone. It was a non-trival effort to figure out why the build was failing even after I'd applied the formatting locally; OK, not a _huge_ deal, but still a frustration that could have been avoided. The build setup isn't detailed anywhere (or is it?), so any new contributor is likely to experience the same. I included an output in one of the builds to expose the module versions: https://ci.appveyor.com/project/sqlcollaborative/dbatools/builds/35296201/job/r612h2n02ueqy36e. If formatting can't be automated, maybe at least the module versions can be exposed somewhere appropriate as a permanent part of the build output...?

niphlod:
> The function was gathering a list of directory paths in $sqlpaths which is passed in to Get-SQLDirTreeQuery, but if any of those directories, which are passed to xp_dirtree, are in UNC format, the server won't return results if it doesn't support UNC paths. OK, now I see why it wouldn't work. I've updated the code to guarantee only local paths are passed in.
> 
> Without the Split-AdminUNC function, or some equivalent, I don't think there's any way for this:
> image
> to get to this:
> `Filename : C:\Program Files\Microsoft SQL Server\MSSQL14.SQL2017DEVELOPER\MSSQL\DATA\dbatoolsci_orphanedfile_TLOG.ldf`
> I think if you accept that SQL allows for specifying UNC paths, then converting back to a local path format is a necessity to maintain the function output.

not really. if SQL is using local paths, you must enum local paths. 
if SQL is using UNC paths, you must enum UNC paths... because if the database is mounted of course the server has access to the share (at least the base folder of it, subdirectories may break ACL inheritance and refuse to be enumerated, but that's nothing SQL or you can solve).
point being you enum what SQL has mounted, without translations (also because of the following comma ^_^)

Your split-adminunc just works for paths that are \\servername\diskname$\path (admin shares) and not \\servername\sharename\path (normal share).... which from MSSQL 2012 is fully supported, db files on any SMB share.
So, you're not fixing a problem, you're introducing one.
 
> Regarding recursion not strictly needing a CTE, I'll leave that for someone else to tackle. For now I've updated to keep the original behavior on pre-2005 versions.

NP, I'll push a fixed version

PR #6867, opened 2020-09-19
Set-DbaTcpPort - Add parameter Restart to restart the service

### LINKED ISSUE #6865: [Feature] Install-DbaInstance / Invoke-DbaAdvancedInstall - Restart service after changing port with Set-DbaTcpPort
### Summary of new feature

Install-DbaInstance can change the port after the installation and does this in Invoke-DbaAdvancedInstall via Set-DbaTcpPort here:

https://github.com/sqlcollaborative/dbatools/blob/f19e8248310602d15eaf5f2fd35aa22e478be956/functions/Invoke-DbaAdvancedInstall.ps1#L292-L295

But it does not invoke the needed restart of the service.

We could just issue a Restart-DbaService.
But on the other hand we may want to change Set-DbaTcpPort to do the restart. Or just tell the user that a restart is needed.
But if Set-DbaTcpPort has a warning output that would also show up at Install-DbaInstance.

Maybe the best way:
Add a parameter like -RestartService to Set-DbaTcpPort. If set, restart the service, if not, display a warning.
Invoke-DbaAdvancedInstall can use this switch to silently restart the service.

@nvarscar : What do you think?

andreasjordan:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality, fixes #6865 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Be able to silently restart the service so that the new value is used. Will fix issues with Install-DbaInstance when using -Port

### Approach
Added new parameter Restart

### Commands to test
see help

I have also slightly reworked the code, I hope thats ok.

### DIFF
--- a/functions/Invoke-DbaAdvancedInstall.ps1
+++ b/functions/Invoke-DbaAdvancedInstall.ps1
@@ -291,7 +291,7 @@ function Invoke-DbaAdvancedInstall {
     }
     # change port after the installation
     if ($Port) {
-        $null = Set-DbaTcpPort -SqlInstance "$($ComputerName)\$($InstanceName)" -Credential $Credential -Port $Port -EnableException:$EnableException -Confirm:$false
+        $null = Set-DbaTcpPort -SqlInstance "$($ComputerName)\$($InstanceName)" -Credential $Credential -Port $Port -Restart -EnableException:$EnableException -Confirm:$false
     }
     # restart if necessary
     try {
--- a/functions/Set-DbaTcpPort.ps1
+++ b/functions/Set-DbaTcpPort.ps1
@@ -102,10 +102,10 @@ function Set-DbaTcpPort {
             $tcp = $wmiinstance.ServerProtocols | Where-Object {
                 $_.DisplayName -eq 'TCP/IP'
             }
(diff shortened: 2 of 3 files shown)


andreasjordan:
I still need a bit of help with wording:

```
.PARAMETER IpAddress
        Which IpAddress should the portchange , if omitted AllIP (0.0.0.0) will be changed with the new port number. 
```

wsmelton:
This command should not support restarting, only setting.

If you want to combine commands to do multiple things then an invoke command did be created.

nvarscar:
Why is that @wsmelton ? It would only enhance user experience of the command would support restarting without having to restart the server as a separate command.

wsmelton:
It is practice with PowerShell.

If you want to start combining commands to preform a complete process that is what the Invoke verb is for. The set verb is for changing the state of something, the fact that restarting the service in SQL Server reference for that to take effect is irrelevant


The user experience should be focused on what called the set command, the install command. You make user experience easier by restarting on that command.

andreasjordan:
I will close this one. I have submitted some smaller changes here: #6880 and #6881

PR #6874, opened 2020-09-24
Fix boolean logic and inline docs

mwsutherland:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [x] Documentation
 - [ ] Build system
 

### Purpose
Fix some boolean logic and the cmdlet name in the inline documentation

### Approach
Line 119, as originally written,

`if (-not $InputObject -and (-not $Schedule -or $ScheduleUid)) {`

would only allow you to get past this point if you have either $InputObject or have $Schedule and not have $ScheduleUid, which makes no sense and contradicts the error message.

I believe the `-not` should have been outside the parentheses, as in

`if (-not $InputObject -and -not ($Schedule -or $ScheduleUid)) {`

but it seems more clear to use the logical equivalent

`if (-not ($InputObject -or $Schedule -or $ScheduleUid)) {`

to say "If you don't have at least one of these, you may not continue"

### Commands to test
Neither of these should return the error "Please enter the schedule or schedule uid"

`Remove-DbaAgentSchedule -SqlInstance sql1 -ScheduleUid 'bf57fa7e-7720-4936-85a0-87d279db7eb7'`
`Remove-DbaAgentSchedule -SqlInstance sql1 -Schedule 'schedule1'`

### DIFF
--- a/functions/Remove-DbaAgentSchedule.ps1
+++ b/functions/Remove-DbaAgentSchedule.ps1
@@ -1,10 +1,10 @@
 function Remove-DbaAgentSchedule {
     <#
     .SYNOPSIS
-        Remove-DbaAgentJobSchedule removes a job schedule.
+        Remove-DbaAgentSchedule removes a job schedule.
 
     .DESCRIPTION
-        Remove-DbaAgentJobSchedule removes a job in the SQL Server Agent.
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Thank you so much, @mwsutherland! This looks good and changes make sense 💯

PR #6882, opened 2020-09-27
Get-DbaAgBackupHistory - Redesign

andreasjordan:
## Type of Change

 - [x] Breaking change (effects multiple commands or functionality, fixes # )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
I have reworked Get-DbaAgBackupHistory to show what I would expect from this command.
@Stuart-Moore - can you please have a look and give feedback if that's the right direction?
The diff is not very helpful as I have changed the fundamental structure.


potatoqualitee:
btw, github has draft PRs. when you create the PR, click the down arrow then Draft instead of regular 💯

andreasjordan:
@Stuart-Moore - I would be happy to get your feedback here

Stuart-Moore:
Sorry, Bits and works 2nd busiest week of the year have gotten in the way.

All looks good with the function to me

The only thing I'd like to see is a test where the output get piped into Restore-DbaDatabase just to confirm the output is good.

I'd happily approve with that in place

For bonus credit, with the lack of AG could we build a mocked test so filtering can be tested? If you look at Test-DbaBackupInformation.Tests.ps1 you should get an idea of how we can abstract away from needing a real instance for testing. This would also allow contributors without an AG to test locally

andreasjordan:
Hi @Stuart-Moore , things are not that easy as expected.

My setup at the moment: 
image

So I have a local full backup of AdventureWorks on the primary and a full and log backup on a share that where taken on the primary and restored on the secondary.

image
I can use Get-DbaAgBackupHistory with either all the replicas or the listener and I get all the backups.

image
I can get the last full or log backup. 

image
But using -Last results in this output, which is different and the output of Select-DbaBackupInformation.

In Get-DbaDbBackupHistory there is a complete seperate code path for -Last, but here the command Select-DbaBackupInformation is used.

So I have two questions:
1. What output should be returned when using -Last?
2. What parameter should be used when piping to Restore-DbaDatabase?

Stuart-Moore:
afternoon,
on the original the last parameter causes the gdbh to build up a backup chain by gabbing the last full, then the last diff and then all the logs needed to complete the chain. My guess is that your backups are so close together that this might be causing an issue with the 

Would it be possible to get the raw  output from the objects? To do that, instead of piping into ft, could you do this:
```
| ConvertTo-Json -Depth 10 | Out-File  -FilePath <path of your choosing>
```

and then attach them to this ticket. Should give us all the raw LSN information that gets used under the hood.

andreasjordan:
Here is the requested file.
[DbaAgBackupHistory.txt](https://github.com/sqlcollaborative/dbatools/files/5392639/DbaAgBackupHistory.txt)

There is only one log backup, it was taken at SRV1 and restored on SRV2 (as part of adding the database to the availability group). Because of that, every instance knows about this backup and returns it when asked by Get-DbaDbBackupHistory.

This issue will probably disappear when I start taking new backups on just one of the instances.

andreasjordan:
I now took a full backup on SRV1 (which is primary at the moment) and then a log backup on SRV1 and a log backup on SRV2.

```
Get-DbaAgBackupHistory -SqlInstance Adventure2017 -AvailabilityGroup Adventure2017 -Database AdventureWorks -Type Log | 
    Sort-Object -Property End | 
    Format-Table -Property ComputerName, End, Path, FirstLsn, LastLsn, LastRecoveryForkGUID

Get-DbaAgBackupHistory -SqlInstance Adventure2017 -AvailabilityGroup Adventure2017 -Database AdventureWorks -Last | 
    Format-Table -Property ComputerName, End, Path, FirstLsn, LastLsn, LastRecoveryForkGUID
```

This is the output:
image

So the last log backup taken on SRV2 is missing, being filtered out by Select-DbaBackupInformation.

andreasjordan:
Ups, have to do more research in my code as...

```
Get-DbaAgBackupHistory -SqlInstance Adventure2017 -AvailabilityGroup Adventure2017 -Database AdventureWorks -Last | 
    Format-Table -Property ComputerName, End, Path, FirstLsn, LastLsn, LastRecoveryForkGUID

Get-DbaAgBackupHistory -SqlInstance Adventure2017 -AvailabilityGroup Adventure2017 -Database AdventureWorks | Select-DbaBackupInformation -ServerName Adventure2017 |
    Format-Table -Property ComputerName, End, Path, FirstLsn, LastLsn, LastRecoveryForkGUID
```

image


So Select-DbaBackupInformation is doing it correct...

andreasjordan:
Found the problem in my code: Had forgotten to remove the -Last when running Get-DbaDbBackupHistory...

andreasjordan:
I had a deeper look at Select-DbaBackupInformation to find out why the first log backup for setting up the availability group is returned twice.
I found out that the log backups are grouped by BackupSetID and because this id is different on the two servers there a two groups and so two results.

The question is: As this is an edge case that we query the backup history just after creating the availability group and before the next full backup - can we just leave the code of Select-DbaBackupInformation as it is? Or is there an easy way to filter out the duplicate?

andreasjordan:
To prove that the information is good to do a restore:

image

andreasjordan:
> For bonus credit, with the lack of AG could we build a mocked test so filtering can be tested? If you look at Test-DbaBackupInformation.Tests.ps1 you should get an idea of how we can abstract away from needing a real instance for testing. This would also allow contributors without an AG to test locally

Sorry, but I don't understand this. I don't see any mocked test there. In the BeforeAll part there are real backups taken from the $script:instance1 to the folder C:\Temp\backups - or am I missing something here?

Stuart-Moore:
Sorry, was thinking more about putting that in the tests. But as it's using the BackupHIstory object all the way through and Select-DbaBackupInformation has enough testing on that's not so important

In https://github.com/sqlcollaborative/dbatools/blob/development/tests/Test-DbaBackupInformation.Tests.ps1, Line 65 onwards show how you can mock the function calls so you don't need to have a db or instance to run the tests. All the 'backups' being tested in that file a rehydrated JSON objects created as I mentioned above.

By mocking the connection call and the calls to dbadbhistory you could pretend that the AG exists, so all you're testing it the logic in your function

It's not essential now you're relying on functions that already have a lot of testing around them.

I'm happy to merge without that if you're happy with it?

Stuart-Moore (commented):
If @andreasjordan is happy with the code then I'm happy to approve it. Just holding off in case he's got any last words, so if anyone else wants to merge it if I'm not around, go ahead

Stuart-Moore (approved):
If @andreasjordan is happy with the code then I'm happy to approve it. Just holding off in case he's got any last words, so if anyone else wants to merge it if I'm not around, go ahead

andreasjordan:
I'm happy with the code.

Stuart-Moore:
Merged. Sorry for the delays. Hoping work quietens down soon and I can do the fun  stuff again. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

PR #6889, opened 2020-09-29
Remove 2019 CU7 from build reference

jpomfret:
Since CU7 has been pulled should be remove it for now so it doesn't show as latest cu?

https://support.microsoft.com/en-us/help/4570012/cumulative-update-7-for-sql-server-2019

## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Currently we're showing that 2019 CU7 should be the latest CU, but it's been pulled with issues

### Approach
removed the object from json

### Learning
https://support.microsoft.com/en-us/help/4570012/cumulative-update-7-for-sql-server-2019


potatoqualitee:
good call, thank you!

niphlod:
nooooope

andreasjordan:
There was a discussion about that on slack #dbatools-dev

niphlod:
nooooope
edit: nope, CU7 is still a version and is still the latest released. the fact that today you can't install it doesn't vouch for removal....this isn't get-dbabuildreferencethatICanInstallToday....

potatoqualitee:
good call 😅 but won't people's test show that they're behind? we dont have a Retracted flag. Can we add this version after MS releases a more recent working update?

jpomfret:
Oh my bad - I had a look at issues here but not slack 😄 

@potatoqualitee  - That was my thought, it shows I'm not on the latest, but is it the latest safe build... we almost need a flag for pulled (I think there is a 2012 version that also is no longer in use).

niphlod:
test-dbabuild tells you you're a release behind. .... and you are if you're on cu6. 
we do not have a "recommended" flag anywhere (and frankly it's hugely out of scope)... the decision of what to install is left to the user, always. we just tell them where they are.

jpomfret:
Totally agree @niphlod - just don't want it to feel like we're suggesting they should be on CU7..

niphlod:
> Totally agree @niphlod - just don't want it to feel like we're suggesting they should be on CU7..

we're not, anywhere, recommending to install things.

jpomfret:
Would a better Could we add something to the build ref that states this is not a supported build?

niphlod:
it's still supported, just not recommended. people are running it to fix SOME bugs  (in slack someone said it fixed an AG error they were having) if they can live with the OTHER bugs (no snap, no checkdb).

potatoqualitee:
While we aren't necessarily recommending things, I think it's safer to remove it until there's a newer release. People's checks for running the latest version can potentially fail over an update that's not even available anymore and we can't help them with the current implementation.

jpomfret:
Ah interesting. Yeah all good points, I just was patching SQL... noticed I wasn't on the latest, since I was on CU6 and threw this PR in 😆 

Sorry for missing the slack discussion

niphlod:
@potatoqualitee : they can't update if the exe is not available...they'll just know that there is a new build. and when they'll try to install it it won't be available.

potatoqualitee:
but their tests will still fail 😢

niphlod:
they need to ^_^ they're not on the latest.

potatoqualitee:
🥚but they can no longer upgrade to the latest 🐔

niphlod:
mwhahaha.. today, they can't. some days ago, they could.

tl;dr: this is something for another cmdlet, aka "get-thingsIcanInstall". nothing to do with test-dbabuild or get-dbabildreference. maybe tangent to install-dbainstance, but very limited point of contact.

jpomfret:
😆 - I have caused quite the issue this morning! 

For consistency I still feel like we should remove it

We don't list 2016 CU9 which was removed by Microsoft but still a legit build (I have some and wondered why they weren't in the build ref)
image

https://sqlperformance.com/latest-builds/sql-server-2016
image

potatoqualitee:
I hate to disagree with simone and andreas 💔, but I agree with Jess 💖. It makes the most sense that we remove it.

niphlod:
2016 cu9 should definitely be there. heck, you can't install most of the 2000 version, but is still there.

potatoqualitee:
I can install SQL Server 2000, no problem 😇 I'd like to institute a policy that pulled releases are either removed entirely or not added until a replacement is available.

Please forgive me, niph 🙏 🤗 !

niphlod:
well, when you will be able to install 8.0.2199 give me a call :P

PR #6894, opened 2020-09-29
formatter, require 1.18.2 till problems with 1.19.1 get sorted out

niphlod:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Make sure users use the same module version we use in appveyor to check formatting.
Issues around PSSA 1.19.1 "breaking" changes have been listed on slack.
Long story short: VSCode and PSSA no longer match exactly, and it doesn't seem something we can "tune".
New PSSA capitalizes correctly commands and their parameters (wow) but also changes all operators to the correct casing (i.e. `-In` over `-in`, `-EQ` over `-eq`, etc).

### DIFF
--- a/functions/Invoke-DbatoolsFormatter.ps1
+++ b/functions/Invoke-DbatoolsFormatter.ps1
@@ -39,17 +39,18 @@ function Invoke-DbatoolsFormatter {
     begin {
         $invokeFormatterVersion = (Get-Command Invoke-Formatter -ErrorAction SilentlyContinue).Version
         $HasInvokeFormatter = $null -ne $invokeFormatterVersion
+        $ScriptAnalyzerCorrectVersion = '1.18.2'
         if (!($HasInvokeFormatter)) {
-            Stop-Function -Message "You need PSScriptAnalyzer version 1.18.2 installed"
-            Write-Message -Level Warning "     Install-Module -Name PSScriptAnalyzer -RequiredVersion '1.18.2'"
+            Stop-Function -Message "You need PSScriptAnalyzer version $ScriptAnalyzerCorrectVersion installed"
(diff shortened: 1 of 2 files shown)


andreasjordan:
Why not having `$ScriptAnalyzerCorrectVersion = '1.18.2'` in Invoke-DbatoolsFormatter as well?

niphlod:
yep, you're totally right

wsmelton:
> Issues around PSSA 1.19.1 "breaking" changes have been listed on slack.

Please include these here as Slack is not accessible by all, nor will it be there in two months 😉 If you have the issue numbers for the repo you can just reference the major ones if needed....will help show them the effect they are having as well.

niphlod:
from slack

"""
1: our bin\PSScriptAnalyzerRules.psd1 has slightly different config from PSSA's (both 1.18.1 and 1.19.1) profile. it is PSUseConsistentWhitespace.CheckOperator that WE set to false, but PSSA doesn't on OTBS. In VSCode setting it to true makes it alert about Where-Object not using the correct case comparison (-eq vs -EQ). So we shush it probably for good reason.

2. running invoke-formatter with bin\PSScriptAnalyzerRules.psd1 turns -eq to -EQ, no matter what PSUseConsistentWhitespace.CheckOperator value is set. -eq turns to -EQ when PSUseCorrectCasing.Enable is $true, which has always been in all places (our vscode profile, PSSA's profiles in 1.18.1 and 1.19.1).
so, the behaviour changed from 1.18.1 and 1.19.1. we might come closer to that behaviour if we set PSUseConsistentWhitespace.CheckOperator to $true in vscode profile, but still leaves something like -in not turned to -In .

that being said I'm unsure on how VSCode can format the code differently using the same codebase. theoretically VSCode calls the underlying method provided by PSSA's module, but it seems it is not always the case.
"""

PR #6902, opened 2020-10-02
Fix #6869 - cope with change in dbname case due to log shipping or other user intereference

### LINKED ISSUE #6869: Error when Restoring database, due to Test-DbaBackupInformation] Trying to restore  from multiple sources databases
when restoring a database, I am receiving the error, Test-DbaBackupInformation] Trying to restore DATABASE from multiple sources databases. The .bak file restores ok, but when adding .trn files, I am gettng the warning message.  Looking at the header for the 2 files, I see the database name in the .bak file is mixed and lower case, DatabaseName, for example. Looking at the .trn file, it is all uppercase, DATABASENAME.  Looking at the Test-DbaBackupInformation, I see that there is a comparison. My input database name is mixed case. The below code does an -eq.  I believe this is causing the issue.  Suggestion would be to uppercase or lowercase both variables so comparison would be equal.

          #Test we're only restoring backups from one database, or hilarity will ensure
            $DbHistory = $InternalHistory | Where-Object { $_.Database -eq $Database }
            if (( $DbHistory | Select-Object -Property OriginalDatabase -Unique ).Count -gt 1) {
                Write-Message -Message "Trying to restore $Database from multiple sources databases" -Level Warning
                $VerificationErrors++

Stuart-Moore:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Issue #6869 raised a case where a restore fails due to a change in case in a db's name due to log shipping

We've not encountered it before in the 3 years the code's been live, but as I have a fix, might as well put it in. 

### Approach
Uppercase the db name when doing the -unique check

This should be safe as SQL appears to do the same when it checks for db existence when you create a new one


niphlod:
Ahem, in CS instances there can be the same db name with different casing...

image

Stuart-Moore:
d'oh, guess what I do my best to not have configured :)

As it's a user induced error we'll ignore it then as we're not going to please everyone and it's made it this far with only being mentioned once.

niphlod:
yeah, no worries ....... things with CS/CSAS instances are weird for people coming from a Windows background.
TBF if you try to create it via SSMS it fails, but not because it checks for the same db with case-insensitivity, but because "file a.mdf already exists", due to NTFS being case-insensitive... you just have to specify different paths before pressing "OK".
With linux instead all goes switfly without a hiccup.

so someone having the same db name with different casing is definitely something remote, but can totally happen.
weird custom logshipping script *should* be even more remote, and MS could introduce a fix for it happening in the first place.

PR #6905, opened 2020-10-04
Improve unique indexes

sanderstad:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
1. Memory usage
The command uses a lot of memory when it has to deal with a large table with a unique index.
It could lead to 10's of GBs for a single process.

2. Static values
The command also did not (or no longer) supported static values.

### Approach
1. Memory usage
The generation of the unique values is now far more sophisticated.
It will create a table in "tempdb" which contains all the columns that are included in all the unique indexes.
A unique index will be created on the temporary table for all the columns.
It will keep generating new values for failed inserts in the temporary table

2. Static values
There is a new property in the configuration file called "StaticValue".
If a static value is set it will generate a column update and not go for RBAR
Added static data type for allowed data types

### Commands to test
Invoke-DbaDbDataMasking
Get-DbaRandomizedValue
New-DbaDbMaskingConfig

### Screenshots
Before the masking
1 TableValues_Before

Unique values generated
2 UniqueValuesTempDB

After the masking
3 TableValues_After

Using static values
4 UsingStaticValues

The run fo the masking
5 RunOfMasking


wsmelton:
@sanderstad can you go ahead and update the gitingore file to exclude `debug.log`. I noticed Code now generates this file for some reason and puts it at the root of the project. We need to exclude this from being added to the user's PRs now.

wsmelton:
You will need to delete it once and commit that change so it gets removed from this PR as well.

wsmelton:
The PR test is failing on a parameter change.

tests/Convert-DbaMaskingValue.Tests.ps1
```diff
@@ -0,0 +1,110 @@
+$CommandName = $MyInvocation.MyCommand.Name.Replace(".Tests.ps1", "")
+Write-Host -Object "Running $PSCommandPath" -ForegroundColor Cyan
+. "$PSScriptRoot\constants.ps1"
+
+Describe "$CommandName Unit Tests" -Tag 'UnitTests' {
+    Context "Validate parameters" {
+        [object[]]$params = (Get-Command $CommandName).Parameters.Keys | Where-Object { $_ -notin ('whatif', 'confirm') }
+        [object[]]$knownParameters = 'Value', 'DataType', 'Nullable'
```
wsmelton, line 8: Needs `EnableException` added

wsmelton: If you want to check the unit test on `Install-DbaFirstResponderKit` we changed the pattern for this test...just have not updated all the other test on it yet.

sanderstad:
> debug.log

Will do!

wsmelton (approved):
Test pass so I'm good.

potatoqualitee:
Fabulousity of 10! Thanks Sander and great to see you back, @wsmelton 🥳 !

PR #6906, opened 2020-10-05
[Feature] Adding Dev Container functionality to the project

wsmelton:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Adds [devcontainer](https://code.visualstudio.com/docs/remote/containers) functionality to the project.

This also provides support for use with GitHub Codespaces for those users that have been added to the beta program for it.

Write-ups will be added to the Wiki and I'll try to get a video created as well to show it in action.

### General Info

The image built (see the `dockerfile`) pulls the latest SQL Server 2019 image, modifies the following for the image:
- Install PowerShell (pwsh)
- Enable SQL Agent

Both of the above only increase the image by around 100MB, the biggest part of that is I'm updating Ubuntu on it first (`apt-get update`.

The reference in the `dockerfile` to `USER root` is because the SQL Server 2019 image was changed to run under a non-root user, and to install PowerShell it has to be as root.

The `docker-compose` file then proceeds to set up each container: 
- Each one is named `dbatoolsx` (1..3)
- User is `root` as well, because for bind mounts and such with Docker Desktop it works better, and if folks want to modify things on the image to play they need root to do it.
- Opens port 1433 on each one to 140x (x being container number). (this allows if you run this from VS Code to be able to access those containers from the host machine.
- The volume is only configured on the first container. So you can adjust that if you want to say have an easy way to add files to the container for reference. Otherwise, it's just something used by devcontainer and GitHub Codespaces.

Three containers are created in reference to how Appveyor environment is configured. If I can figure out how to detect we are in devcontainers or GitHub workspaces I'll update our constants file to adjust for that. Then folks can more easily run our tests individually or as a whole. (Our test may need updating at this time as well to detect the OS version they are running on since the Windows-only are going to fail.)

There is one issue with GitHub Codespaces that I'm trying to get an answer to that it does seem to delete three files specifically after it loads up. It is deleting our psd1, psm1 and our `dbatools.dll` file. I don't know if there may be an exclusion list I need to configure so it ignores those or what...TBD. It is not a show stopper as users will just need to remove those changes to put them back for right now, and is only effecting Codespaces use.


wsmelton:
This is a branch within the repo. So the easiest way to test this is to pull down the repo, then checkout that branch.

Doing this in VS Code once the branch is loaded (checked out) it will detect the devcontainer and you are going to be prompted to load it up.

wsmelton:
Still working out how the repo is loaded into the devcontainer once you pull our repository down.

- `git clone https://github.com/sqlcollaborative/dbatools.git` (or your forked repo)
- VS Code loads the project
- Detects devcontainer
- Prompt to install all necessary extensions
- Prompts to reopen in container
- ? (project is not showing just the contents of the devcontainer folder)

wsmelton:
Seems the best chance I have to detect where this is running is with `$env:TERM_PROGRAM` in both VS Code container and GitHub Codespaces this value is set to `vscode`.

wsmelton:
More to figure out on this one so I'll close this PR and reopen once I get VS Code side figured out more.

PR #6912, opened 2020-10-08
buildreference, introduce retired kbs

niphlod:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Handle retired KBs. Took a while (sorry) but we do need to manage them somehow.

### Approach
Add the "Retired" attribute to the build, make sure that Get-DbaBuildReference returns a warning if the build is "officially retired".
Test-DbaBuild recognizes the same warning but additionally skips that version if -MaxBehind is passed.

### Commands to test
Look at the added tests

### Screenshots
image
image

As soon as the buildref is committed to the other repo, this is the result

image

and, of course, if it is indeed the last cu to be retired, it will keep the "latest cu" column ticked in the previous one, so if CU8 didn't exist yet, it'll show this
image

### DIFF
--- a/bin/dbatools-buildref-index.json
+++ b/bin/dbatools-buildref-index.json
@@ -1,5 +1,5 @@
 {
-    "LastUpdated": "2020-10-02T00:00:00",
+    "LastUpdated": "2020-10-08T00:00:00",
     "Data": [
         {
             "Version": "8.0.47",
@@ -3988,6 +3988,12 @@
             "KBList": "4505830"
         },
         {
+            "CU": "CU9",
+            "Version": "13.0.5479",
+            "KBList": "4515435",
+            "Retired": true
+        },
+        {
             "CU": "CU10",
             "Version": "13.0.5492",
             "KBList": "4524334"
@@ -4246,6 +4252,12 @@
             "CU": "CU6",
             "Version": "15.0.4053",
             "KBList": "4563110"
+        },
+        {
+            "CU": "CU7",
+            "Version": "15.0.4063",
+            "KBList": "4570012",
+            "Retired": true
         },
         {
             "CU": "CU8",
--- a/functions/Get-DbaBuildReference.ps1
+++ b/functions/Get-DbaBuildReference.ps1
@@ -331,9 +331,15 @@ function Get-DbaBuildReference {
                 $Detected.KB = $el.KBList
                 if (($Build -and $el.Version -eq $Build) -or ($Kb -and $el.KBList -eq $currentKb)) {
                     $Detected.MatchType = 'Exact'
+                    if ($el.Retired) {
+                        $Detected.Warning = "This version has been officially retired by Microsoft"
+                    }
                     break
                 } elseif ($MajorVersion -and $Detected.SP -contains $ServicePack -and (!$CumulativeUpdate -or ($el.CU -and $el.CU -eq $CumulativeUpdate))) {
                     $Detected.MatchType = 'Exact'
+                    if ($el.Retired) {
+                        $Detected.Warning = "This version has been officially retired by Microsoft"
+                    }
                     break
                 }
             }
--- a/functions/Test-DbaBuild.ps1
+++ b/functions/Test-DbaBuild.ps1
@@ -226,6 +226,7 @@ function Test-DbaBuild {
                             VersionObject = $el.VersionObject
                             SP            = $lastsp
                             CU            = $el.CU
+                            Retired       = $el.Retired
                         }
                     }
                 }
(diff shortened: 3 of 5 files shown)


niphlod:
I'm unsure about the wording. The competition is between "Retired" and "Withdrawn"... Vote away :D

wsmelton:
Define a retired KB?

wsmelton:
If this will only apply to SQL Server 2016 and back, then I'd say I'm ok with this. However SQL Server 2017+ the KBs/CUs are not retired because of the servicing model those products are under now. The support life cycle for a CU under the service model will be for the EOL of the production version itself. That is why the "End of Support" date is the same on every CU for SQL Server 2017 and 2019, it's until the support lifecycle ends for the major version.

In reference to the terms. Retired means you can't use it in software terms or it is physically going to be removed and not available anymore. Microsoft won't do that because of Enterprise Support agreements. Withdrawn would have the same meaning, and can be confused with CUs that Microsoft actually did pull (none of which we actually added to the build ref file). Which because of this I don't see this as being needed at all unless there was some wanting discussion for it outside of an issue?

niphlod:
I feel it's getting more and more difficult to roll this thing up the hill. We NEED to identify a build: the fact that MS finds out there's a bug and based on the severity of that bug recommends users to not install it are completely untied.
Summing up, IMHO these are separate concepts:
- users CAN have a version if in some point in time MS released a patch
- MS can tell users to not install it
- MS can deny the download of said patch at some point in history (and this is true for both versions out of support (e.g. SQL2000 patches) and for "bugged" patches, like CU7 for 2019)

So, on one end, if an instance is 15.0.4063, IT IS CU7. And removing it from the index made impossible for dbatools to tell the users BOTH if they were on something MS actually did not recommend AND where they were.

On the other end, @potatoqualitee and others felt like LISTING a CU7 as "the latest CU released from MS" (which is a plain fact if CU8 hadn't been released yet) might suggest that people on CU6 are "hinted" to install CU7. I was "left alone" saying that WHAT to install is a choice for the user to make, and not dbatools, and certainly not a job for Get-DbaBuildReference or Test-DbaBuild.
BUT, I heard SOME more complaints, which were:
-  "users on CU6 testing for being on the latest CU are "failing" but we should not  "hint" that a CU7 is installable
- we should flag in *some* way at least releases that are pulled officially by Microsoft

This PR IMHO solves all ends: does exactly what we did NOT listing VALID version numbers in regards of "testing" and "hinting" and it gives the possibility for dbatools to:
 - know what releases your instances are on, irregardless of the fact that in some point in history MS "officially recommend to not stay at that version"
 - signalling without doubt the difference between the existence of a version 'cause a patch has been released VS the existence of a warning about the stability of it

Given all of the above, let's come up with the correct english term for "MS put a big yellow warning on it and if you are there you should either uninstall it or go to the next level": I'm not native English so I see and already recognized that maybe "retired" is not the best wording. 
BTW, I'm not convinced that what MS is reporting has good wording too (they signal it as "Removed", even if the patch is clearly still "downloadable").

wsmelton:
> BTW, I'm not convinced that what MS is reporting has good wording too (they signal it as "Removed", even if the patch is clearly still "downloadable").

SQL Server 2019 CU7 is not downloadable at all right now. Whether users were able to download it at the time it was first released is out of our control. Unless we are going to provide written proof from Microsoft that the KB is "removed" (that is not retired or withdrawn) we should not include it on our list. For CU7 we could include it but all other data points would have to be addressed as well (e.g. "End of Support" would be null and not include a date). Just putting a value of "removed" would not be sufficient for me to believe our buildref data as a whole.

> On the other end, @potatoqualitee and others felt like LISTING a CU7 as "the latest CU released from MS" (which is a plain fact if CU8 hadn't been released yet) might suggest that people on CU6 are "hinted" to install CU7.

If the others have ever called support or gone through a security standards audit (PCI, etc.) they would find the first thing they are told by Microsoft support is to install the latest CU released and retest the "bug" or "issue" the cases is referencing. Security standards only tell you to be on the latest version of the product support. Microsoft tells you in every KB it is recommended to be on the latest CU of their product. Why shouldn't we?

image

wsmelton:
Referencing SQL Server 2019 CU7: https://support.microsoft.com/en-us/help/4570012/cumulative-update-7-for-sql-server-2019

Every download link takes you to CU8. The only possible place it "might" be available in the WSUS catalog but I recall a blog post coming out that stated it was physically removed from every corner of the Internet to be downloaded because of how bad it was.

wsmelton:
If @potatoqualitee and others wanted it, I'll let them review as I see no purpose or reason for this to be added.

niphlod:
> Microsoft tells you in every KB it is recommended to be on the latest CU of their product. Why shouldn't we?

best practices are out of scope too. 
If everyone follows what is recommended, there would be no point in Test-DbaBuild.
I'm not pro, nor against, being on the latest. But it still is a CHOICE. The webpage, Get-DbaBuildReference and Test-DbaBuild are all TOOLS to make that choice.

> SQL Server 2019 CU7 is not downloadable at all right now

again, SQL2000 patches are not downloadable too. CU7, in addition to not being "downloadable", has been marked at a certain point in history by Microsoft as "there is a known problem, please uninstall it".

Once again, there are three "main" POVs.
- Mine is that the index must list every version you can possibly have, CU7 not being available from a certain point on doesn't mean it didn't exist at all or users did not install it. With this PR we can inform users about an "official warning by Microsoft" too.
- @potatoqualitee doesn't want it to appear because, till this PR, we didn't signal clearly that that version is a "no-go"
- Yours is that we should list only what MS supports and remains "downloadable". But you'd concur with me here that stretching that, everything out of support shouldn't be there, the most obvious example is all the versions of SQL2000, or CTP, or RC.

potatoqualitee:
i actually like this compromise a lot. fantastic solution, niph 😍 ! whatcha think, @jpomfret ?

potatoqualitee:
If all goes well, ill merge this one during a presentation tomorrow, too 😅

jpomfret:
I like this too, completes the build reference with all builds... but also indicates builds which are retired\removed\missing. Thanks @niphlod for the hard work, I love this buildref!

potatoqualitee (commented):
this looks great!

Issue #5380, opened 2019-04-17
Export-DBAInstance - add "-Append" option


SvenLowry:
---
name: Feature request/idea :bomb:
about: Suggest a new feature or enhancement

---

### Summary of new feature

Add the "-Append" option to Export-DBAInstance to allow the user to choose whether the scripts overwrite existing files or not

* Please provide the latest released version of dbatools module.
0.9.804

potatoqualitee:
Good call, thanks for the request 👍

potatoqualitee:
resolved - will be available in dbatools 1.0 in 20 says

sqllensman:
This actually does not do anything in current code as all scripts are written to a sub directory that includes a timestamp.
Line 173: $path = Join-DbaPath -Path $Path -Child "$($server.name.replace('\', '$'))-$timenow"

The file can never exist prior to being written.

Also for  code paths where same file is being updated with additional content eg Resource Governor, Policy Management etc then Append flag is required to avoid overwriting the content

potatoqualitee:
That is the Default behavior. But my intention was to allow them to be as specific as they'd like with Path and FilePath, which would benefit from -Append

sqllensman:
As the code is currently written
FilePath parameter is unused
Path parameter is used to create a new Sub Directory that has SQL instance name and a timestamp.

As the SqlInstance parameter is an array, then it needs something like this to handle more than 1 instance being passed into the function.

The concept of having an Append switch would only make sense if there is also an option to create the Sub-directories without the timestamp. 

There is also a difficulty in that each file has a number prefix and it may never be certain that filename will be reused as the number may alter.



The

potatoqualitee:
damn, i lost track of what i was doing there. thanks for reopening, ill take a closer look and fix.

sqllensman:
I was looking at this function in terms of fixing issues with #5402 but I have difficulty in seeing how having an Append switch makes much sense or can be reliably implemented

The only reliable way that I can think of to achieve this is if all data was written to a single file. In this case you could implement Append logic but would also want to implement a NoClobber switch to ensure you are not inadvertently overwriting a file.

While this can technically be done I can't really imagine what use this would be. 
Having one file with multiple versions of configuration may be useful for some audit purpose but the same would be true of the timestamped directories

If this requirement really exists then it can be achieved by some a script that could be run afterward that concatenates and appends all of the individual files

potatoqualitee:
[...]

PR #6919, opened 2020-10-10
Export-DbaInstance - Append doesn't do what it appears

### LINKED ISSUE #6918: [Bug] Export-DbaInstance doesn't export all objects advertised
### Report

When using `Export-DbaInstance` without `-Append` set to `$true`, object "groups" that contain multiple steps like `agent` and `dbmail` only export files for the last item in the group.

For example, the file for SQL Agent only produces the Agent jobs. The other items (Categories, Operators, Alerts, Proxies & Schedules) are not exported because `Get-DbaAgentJob | Export-DbaScript` doesn't append, it overwrites (`$Append` is passed through from the `Export-DbaInstance` parameters but it's not set to `$true` by default).

A `-Append` switch was proposed in #5380 but  I think that's meant for a different use case than was implemented with the switch as we see it today.

#### Host used

- [X] powershell.exe
- [ ] ISE
- [ ] VS Code
- [ ] Other (please specify)

#### Errors Received
No errors produced - unexpected behavior

#### Steps to Reproduce
`Execute-DbaInstance -sqlinstance MYINSTANCE -path c:\tmp`

#### Expected Behavior
All objects for AgentServer, ResourceGovernor, PolicyManagement, CentralManagementServer, DatabaseMail should be included in their respective script files.

#### Actual Behavior
Only the last item in each of the listed "groups" is exported to the script file.

### Environmental information

🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨

```powershell
PowerShell Version        : 5.1.14393.3866
dbatools latest installed : 1.0.124
```

#### SQL Server:

```sql
Microsoft SQL Server 2019 (RTM-CU6) (KB4563110) - 15.0.4053.23 (X64) 
	Jul 25 2020 11:26:55 
	Copyright (C) 2019 Microsoft Corporation
	Developer Edition (64-bit) on Windows Server 2016 Standard 10.0 <X64> (Build 14393: ) (Hypervisor)
```

```sql
us_english
```

alevyinroc:
## Type of Change

 - [X] Bug fix (non-breaking change, fixes #6918  )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
This is probably going to be a temporary measure until #5380 is completed but my hope is that we'll help people get what they're expecting in the meantime. Addresses #6918

In #5380 a `-Append` switch was proposed and I _think_ the intention there was to have a single file exported which contains _all_ scripts for the instance. But this being defaulted to `$false` (well, `$null`) prevents the complete scripts from being produced in the current implementation where we have one script per "group" and there are multiple object types per "group" (such as Agent and DBMail).

If I'm exporting AgentServer, I would expect _everything_ for AgentServer to be exported. For DBMail, if we don't export everything, DBMail isn't fully configured by the script that's produced - configuration, accounts, profiles & server settings aren't there.

### Approach
Default `-Append` to `$true`. I'm defaulting to this because it is what people expect from the function - 
### Commands to test
`Export-DbaInstance -sqlinstance MYINSTANCE`

With `-Append` set to `$true` (which will now be the default), Agent Operators, Schedules, Proxies, etc. should be included in the` X-sqlagent.sql` script.
With `-Append` set to `$false`, Agent Operators, Schedules, Proxies, etc. will be included in the` X-sqlagent.sql` script.

### DIFF
--- a/functions/Export-DbaInstance.ps1
+++ b/functions/Export-DbaInstance.ps1
@@ -144,7 +144,7 @@ function Export-DbaInstance {
         [ValidateSet('Databases', 'Logins', 'AgentServer', 'Credentials', 'LinkedServers', 'SpConfigure', 'CentralManagementServer', 'DatabaseMail', 'SysDbUserObjects', 'SystemTriggers', 'BackupDevices', 'Audits', 'Endpoints', 'ExtendedEvents', 'PolicyManagement', 'ResourceGovernor', 'ServerAuditSpecifications', 'CustomErrors', 'ServerRoles', 'AvailabilityGroups', 'ReplicationSettings')]
         [string[]]$Exclude,
         [string]$BatchSeparator = 'GO',
-        [switch]$Append,
+        [switch]$Append = $true,
         [Microsoft.SqlServer.Management.Smo.ScriptingOptions]$ScriptingOption,
         [switch]$NoPrefix = $false,
         [switch]$ExcludePassword,


potatoqualitee:
mmm, im not so sure about this one but need some time to articulate why 😅  ill bbiab

potatoqualitee:
okay, so the reason is because it's semantically incorrect. We'd have to change the name of the variable like "NoAppend". A switch that defaults to true would require a `:$false` to work. 

Is the issue just with Agent? If so, I'd have no problem throwing all of that into one file.

potatoqualitee:
Thank you @alevyinroc - going to resolve any outstanding issues in https://github.com/sqlcollaborative/dbatools/pull/6957

PR #6950, opened 2020-11-01
Improve Export-DbaUser

### LINKED ISSUE #6710: Export-DbaUser not overwritten the file
### Environmental information

PowerShell Version        : 5.1.18362.113
dbatools latest installed : 1.0.111
Culture of OS             : pt-PT

#### SQL Server:

```sql
/* REPLACE WITH output of @@VERSION */
```
Microsoft SQL Server 2017 Enterprise Edition 
```sql
```
us_english
#### Host used

- [x] powershell.exe
- [x] ISE
- [ ] VS Code
- [x] Other (please specify) 
ADS

- [x] Still buggy in powershell.exe

#### Errors Received
No errors received.

#### Steps to Reproduce

Export-DbaUser -SqlInstance DevInst -Database someDB -FilePath "C:\Temp\test.txt"

I also tried mixing the -Append and -NoClobber parameters to false without success. The behaviour is the same.

I also tried on the actual latest version (v.1.0.115) and is the same.

#### Expected Behavior

Was expecting to have the same behaviour like the Export-DbaLogin that the file is always overwritten.

#### Actual Behavior

Execute sequentially the command above makes the code to be appending to the file  and not overwritten it as was expected..

ClaudioESSilva:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #6710 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 

- Fix default behaviour - Overwrite existing file (instead of append)
- Add missing `-Encoding` parameter.

### Approach

I have fixed the wrong default behaviour (reported on #6710). With this version, it will overwrite by default as other `Export-Dba*` commands.

I took the opportunity and I have also added the `-Encoding` parameter because was missing. (we have it on `Export-Dba*` commands)

### DIFF
--- a/functions/Export-DbaUser.ps1
+++ b/functions/Export-DbaUser.ps1
@@ -158,7 +158,7 @@ function Export-DbaUser {
     begin {
         $null = Test-ExportDirectory -Path $Path
 
-        $outsql = $script:pathcollection = @()
+        $outsql = $script:pathcollection = $instanceArray = @()
         $GenerateFilePerUser = $false
 
         $versions = @{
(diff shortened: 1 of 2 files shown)


potatoqualitee:
Thank you, friend! one pass

PR #6957, opened 2020-11-05
Fix 6837 and 6918 for Export-DbaInstance

### LINKED ISSUE #6837: [Bug] Export-DbaInstance - Honor date format configuration
As mentioned in #6833 the command does not honor our date configuration formatting.uformat.

### Report

The command should honor the `uformat` date configuration for the folder naming.

#### Steps to Reproduce

Run `Export-DbaInstance -SqlInstance SomeServer`

#### Expected Behavior

Output of folder name to be: `C:\temp\dbatools\lab-sql-01-20200914074942`

#### Actual Behavior

Got: `C:\temp\dbatools\lab-sql-01-09142020074942`

#### References:

image

### LINKED ISSUE #6918: [Bug] Export-DbaInstance doesn't export all objects advertised

lancasteradam:
- Updated the format of the date used when creating the export dir for #6837.
- Replace the ‘GO’ with (Get-DbatoolsConfigValue -FullName 'formatting.batchseparator').
- Provided a suggested fix for the Append issue discussed in #6919 and #6918 based on the comment from @wsmelton here: https://github.com/sqlcollaborative/dbatools/issues/6918#issuecomment-706737694
- Fixed the policy export because it was not able to be exported by Export-DbaScript due to incompatible API.
- Removed the Get-DbaDbMail call since it was resulting in duplicate SQL.
- Moved the Get-DbaResourceGovernor call to be after the other components so that the resulting SQL is run in the correct order.
- Ensure test dirs are deleted after the tests run.
- Added an integration test for dir name for #6837.
- Added integration tests for all of the exported files and their contents. These tests can be expanded in the future as needed.
- Updated the documentation regarding the database restoration scripts.

## Type of Change

 - [x] Bug fix (non-breaking change, fixes #6837 and fixes #6918  )
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [x] Documentation
 - [ ] Build system
 

### Purpose and approach
 
The original goal of this effort was to fix 6837 but various other issues were encountered along the way, so there are additional changes included. There is overlap here with the effort in https://github.com/sqlcollaborative/dbatools/pull/6919 and in this proposed fix for the "Append" it is essentially the same, i.e. the Append is always going to be true. Since Export-DbaInstance is currently creating new dynamic files during each execution it needs Append=true because the multi-step exports such as DBMail require it.

The integration tests that have been added provide placeholders for future expansion as needed.


potatoqualitee:
Thanks so much, @lancasteradam ! @wsmelton - can I get your opinion on this? Seems like if we go this route, we'll have to either Remove $Append or add a `-NoAppend`. I can't see how to avoid a breaking change. @alevyinroc also submitted a similar PR so it's something that does need to be addressed.

functions/Export-DbaInstance.ps1
```diff
@@ -163,6 +163,10 @@ function Export-DbaInstance {
         $ScriptingOptions = New-Object Microsoft.SqlServer.Management.Smo.ScriptingOptions
         $ScriptingOptions.ScriptBatchTerminator = $true
 
+        # The way this command is currently written it makes sense for it to always append, since the files are dynamically created and some of the exported objects are gathered in multiple steps and then written to the same file.
+        # This issue is discussed at: https://github.com/sqlcollaborative/dbatools/pull/6919 and https://github.com/sqlcollaborative/dbatools/issues/6918
+        $Append = $true
+        $ScriptingOptions.AppendToFile = $true
```
wsmelton, line 169: We can't provide a parameter of `-Append` and always set it to true, that defeats the purpose of a switch parameter. As-is if this is going to be done the parameter has to be removed because it gives a false sense to users that it will not append by default.

wsmelton: Going to move this discussion to the issue because this can't be done using this method.

functions/Export-DbaInstance.ps1
```diff
@@ -173,7 +177,7 @@ function Export-DbaInstance {
             } catch {
                 Stop-Function -Message "Error occurred while establishing connection to $instance" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
             }
-            $timeNow = (Get-Date -uformat "%m%d%Y%H%M%S")
+            $timeNow = (Get-Date -UFormat (Get-DbatoolsConfigValue -FullName 'formatting.uformat'))
```
wsmelton, line 180: Do we want to add a short note to this in the comments? Just so users are now aware the format of the date will change on the files.

functions/Export-DbaInstance.ps1
```diff
@@ -351,9 +355,23 @@ function Export-DbaInstance {
...
-                $null = Get-DbaPbmPolicy -SqlInstance $server | Export-DbaScript -FilePath "$exportPath\$fileCounter-policymanagement.sql" -Append:$Append -BatchSeparator $BatchSeparator -NoPrefix:$NoPrefix
+
+                $outputFilePath = "$exportPath\$fileCounter-policymanagement.sql"
+                $scriptText = ""
+                $policyObjects = @()
+
+                # the policy objects are a different set of classes and are not compatible with the SMO object usage in Export-DbaScript
+                $policyObjects += Get-DbaPbmCondition   -SqlInstance $server
+                $policyObjects += Get-DbaPbmObjectSet   -SqlInstance $server
+                $policyObjects += Get-DbaPbmPolicy      -SqlInstance $server
```
wsmelton, line 366: The tabbing out of the parameter spacing is not needed here. It just looks unnormal to me, that is used when you deal with syntax like PSCustomObjects but not parameters.

functions/Export-DbaScript.ps1
```diff
@@ -154,19 +154,19 @@ function Export-DbaScript {
         $commandName = $MyInvocation.MyCommand.Name
         $prefixArray = @()
 
-        $appendToScript = $Append
+        # if the Append switch is used then ensure the scripting options are updated
+        $appendToScript = $false
+        if ($Append.IsPresent) {
+            $ScriptingOptionsObject.AppendToFile = $true
+            $appendToScript = $true
+        }
```
wsmelton, line 162: We use `Test-Bound` to check for parameter switch being passed in.

functions/Export-DbaScript.ps1
```diff
@@ -154,19 +154,19 @@ function Export-DbaScript {
...
         if ($ScriptingOptionsObject) {
             # Check if BatchTerminator is consistent
             if (($($ScriptingOptionsObject.ScriptBatchTerminator)) -and ([string]::IsNullOrWhitespace($BatchSeparator))) {
                 Write-Message -Level Warning -Message "Setting ScriptBatchTerminator to true and also having BatchSeperarator as an empty or null string may produce unintended results."
             }
-
-            if ($ScriptingOptionsObject.AppendToFile) {
-                Write-Message -Level Verbose -Message "The ScriptingOptionsObject AppendToFile setting of true will override Append parameter."
-                $appendToScript = $true
-            }
```
wsmelton, line 167: Why are we removing the verbose message here?

tests/Export-DbaInstance.Tests.ps1
```diff
@@ -4,23 +4,334 @@ Write-Host -Object "Running $PSCommandPath" -ForegroundColor Cyan
...
 Describe "$CommandName Unit Tests" -Tag 'UnitTests' {
     Context "Validate parameters" {
-        [object[]]$params = (Get-Command $CommandName).Parameters.Keys | Where-Object {$_ -notin ('whatif', 'confirm')}
-        [object[]]$knownParameters = 'SqlInstance', 'SqlCredential', 'Credential', 'Path', 'NoRecovery', 'IncludeDbMasterKey', 'Exclude', 'BatchSeparator', 'ScriptingOption', 'NoPrefix', 'ExcludePassword', 'Append', 'EnableException'
-        $knownParameters += [System.Management.Automation.PSCmdlet]::CommonParameters
         It "Should only contain our specific parameters" {
-            (@(Compare-Object -ReferenceObject ($knownParameters | Where-Object {$_}) -DifferenceObject $params).Count ) | Should Be 0
+            [object[]]$params = (Get-Command $CommandName).Parameters.Keys | Where-Object { $_ -notin ('whatif', 'confirm') }
+            [object[]]$knownParameters = 'SqlInstance', 'SqlCredential', 'Credential', 'Path', 'NoRecovery', 'IncludeDbMasterKey', 'Exclude', 'BatchSeparator', 'ScriptingOption', 'NoPrefix', 'ExcludePassword', 'Append', 'EnableException'
+            $knownParameters += [System.Management.Automation.PSCmdlet]::CommonParameters
```
wsmelton, line 10: Is there a reason you are moving these into the It block on your PRs? I noticed you did this on the last PR you had submitted.

functions/Export-DbaInstance.ps1
potatoqualitee: I agree here

functions/Export-DbaInstance.ps1
potatoqualitee: But the default stays the same, no?

lancasteradam: Yes, the $timeNow is used for naming the new parent folder that gets created each time the export is done. It will use the default provided by formatting.uformat (yyyyMMddHHmmss as requested in #6837).

I will add a note regarding this in the command description.

functions/Export-DbaScript.ps1
lancasteradam: My thought here was that the Append param should be the overriding factor instead of the ScriptingOptionsObject. From the caller's perspective I think this is more intuitive. Using the -Append param wouldn't be overwritten by another object's state.

tests/Export-DbaInstance.Tests.ps1
lancasteradam: Yes, with Pester 5.0.2 the following error occurs due to scoping of the variables. So, the fix I have been implementing is to just move the test code inside the It blocks so the test logic has the necessary scope visibility.

[-] Export-DbaInstance Unit Tests.Validate parameters.Should only contain our specific parameters 405ms (245ms|159ms)
 ParameterBindingValidationException: Cannot bind argument to parameter 'ReferenceObject' because it is null.
 at <ScriptBlock>, C:\Users\lancasteradam\Documents\GitHub\forks\dbatools\tests\Export-DbaInstance.Tests.ps1:11

wsmelton: We are not updating to Pester 5 any time soon. When we do it will be to 5.1 because many things and structures around the discovery process that Jakub implemented changed and improved.

This can stay but please don't make these changes going forward on future PRs.

Our test will be written to support Pester 5+ as a whole so we handle applying the same pattern across the board.

functions/Export-DbaInstance.ps1
```diff
@@ -97,7 +105,10 @@ function Export-DbaInstance {
         Add scripting options to scripting output for all objects except Registered Servers and Extended Events.
 
     .PARAMETER Append
-        Append to the target file instead of overwriting.
+        Note: this parameter is deprecated and will be removed in a future release.
```
wsmelton, line 108: If it's deprecated, remove it.

functions/Export-DbaInstance.ps1
```diff
@@ -119,17 +130,24 @@ function Export-DbaInstance {
...
 
-        Exports everything but logins and database restore scripts to C:\dr\sqlcluster
+        Exports everything but logins and database restore scripts to C:\dr\sqlcluster\<machinename>$sqlcluster-yyyyMMddHHmmss
 
-.EXAMPLE
+    .EXAMPLE
         PS C:\> Export-DbaInstance -SqlInstance sqlcluster -Path C:\servers\ -NoPrefix
 
-        Exports everything to C:\servers but scripts do not include prefix information.
+        Exports everything to C:\servers\<machinename>$sqlcluster-yyyyMMddHHmmss but scripts will not include prefix information.
```
wsmelton, line 143: Please provide a true example.

Just make up a machine name and date value.

functions/Export-DbaInstance.ps1
```diff
@@ -119,17 +130,24 @@ function Export-DbaInstance {
         PS C:\> Export-DbaInstance -SqlInstance sqlserver\instance
 
         All databases, logins, job objects and sp_configure options will be exported from
-        sqlserver\instance to an automatically generated folder name in Documents.
+        sqlserver\instance to an automatically generated folder name in Documents. For example, C:\Users\<username>\Documents\DbatoolsExport\<machinename>$sqlcluster-yyyyMMddHHmmss
```
wsmelton, line 133: Please move this to one line. Word wrapping is handled by `Get-Help`.

For the example path, I would just put `%userprofile%\Documents...`. Please also make up a machine name and date value.

functions/Export-DbaInstance.ps1
```diff
@@ -119,17 +130,24 @@ function Export-DbaInstance {
...
 
     .EXAMPLE
         PS C:\> Export-DbaInstance -SqlInstance sqlcluster -Exclude Databases, Logins -Path C:\dr\sqlcluster
 
-        Exports everything but logins and database restore scripts to C:\dr\sqlcluster
+        Exports everything but logins and database restore scripts to C:\dr\sqlcluster\<machinename>$sqlcluster-yyyyMMddHHmmss
```
wsmelton, line 138: Please provide a fake machine name and date value for the example.

functions/Export-DbaInstance.ps1
```diff
@@ -119,17 +130,24 @@ function Export-DbaInstance {
...
+    .EXAMPLE
         PS C:\> Export-DbaInstance -SqlInstance sqlcluster -Path C:\servers\ -NoPrefix
 
-        Exports everything to C:\servers but scripts do not include prefix information.
+        Exports everything to C:\servers\<machinename>$sqlcluster-yyyyMMddHHmmss but scripts will not include prefix information.
+
+    .EXAMPLE
+        PS C:\> Export-DbaInstance -SqlInstance sqlcluster -Path C:\servers\ -Overwrite
+
+        Exports everything to C:\servers\<machinename>$sqlcluster and will overwrite/refresh existing files in that folder.
```
wsmelton, line 148: Please provide a fake machine name.

functions/Export-DbaInstance.ps1
```diff
@@ -160,22 +179,29 @@ function Export-DbaInstance {
         $elapsed = [System.Diagnostics.Stopwatch]::StartNew()
         $started = Get-Date
 
-        $ScriptingOptions = New-Object Microsoft.SqlServer.Management.Smo.ScriptingOptions
-        $ScriptingOptions.ScriptBatchTerminator = $true
-
+        if ($Append) {
+            Write-Message -Level Warning -Message "The -Append parameter is deprecated for Export-DbaInstance. It will be removed in a future release."
+        }
```
wsmelton, line 184: Please remove this.

functions/Export-DbaInstance.ps1
```diff
@@ -186,254 +212,198 @@ function Export-DbaInstance {
...
-                $null = Get-DbaRegServerGroup -SqlInstance $server | Export-DbaScript -FilePath "$exportPath\$fileCounter-regserver.sql" -Append:$Append -BatchSeparator 'GO' -NoPrefix:$NoPrefix
-                $null = Get-DbaRegServer -SqlInstance $server | Export-DbaScript -FilePath "$exportPath\$fileCounter-regserver.sql" -Append:$Append -BatchSeparator 'GO' -NoPrefix:$NoPrefix
-                Get-ChildItem -ErrorAction Ignore -Path "$exportPath\$fileCounter-regserver.sql"
-                if (-not (Test-Path "$exportPath\$fileCounter-regserver.sql")) {
-                    $fileCounter--
+                $outputFilePath = "$exportPath\regserver.xml"
+                if ($Overwrite) {
+                    $null = Export-DbaRegServer -SqlInstance $server -FilePath $outputFilePath -Overwrite
+                } else {
+                    $null = Export-DbaRegServer -SqlInstance $server -FilePath $outputFilePath
```
wsmelton, line 269: This can be changed to one-line call:

`$null = Export-DbaRegServer -SqlInstance $server -FilePath $outputFilePath -Overwrite:$Overwrite`

functions/Export-DbaScript.ps1
```diff
@@ -154,19 +154,19 @@ function Export-DbaScript {
         $commandName = $MyInvocation.MyCommand.Name
         $prefixArray = @()
 
-        $appendToScript = $Append
+        # if the Append switch is used then ensure the scripting options are updated
+        $appendToScript = $false
+        if ($Append) {
+            $ScriptingOptionsObject.AppendToFile = $true
+            $appendToScript = $true
```
wsmelton, line 161: The check for `$Append` should be removed and just set the options to true.

tests/Export-DbaInstance.Tests.ps1
```diff
@@ -4,23 +4,351 @@ Write-Host -Object "Running $PSCommandPath" -ForegroundColor Cyan
...
+
+    It "Export replication settings" {
+        $results = Export-DbaInstance -SqlInstance $testServer -Path $exportDir -Exclude 'AgentServer', 'Audits', 'AvailabilityGroups', 'BackupDevices', 'CentralManagementServer', 'Credentials', 'CustomErrors', 'DatabaseMail', 'Databases', 'Endpoints', 'ExtendedEvents', 'LinkedServers', 'Logins', 'PolicyManagement', 'ResourceGovernor', 'ServerAuditSpecifications', 'ServerRoles', 'SpConfigure', 'SysDbUserObjects', 'SystemTriggers'
+
+        $results.FullName | Should -Exist
+        $results.Length | Should -BeGreaterThan 0
+    }
+
+    It "Export system db user objects" {
+        $results = Export-DbaInstance -SqlInstance $testServer -Path $exportDir -Exclude 'AgentServer', 'Audits', 'AvailabilityGroups', 'BackupDevices', 'CentralManagementServer', 'Credentials', 'CustomErrors', 'DatabaseMail', 'Databases', 'Endpoints', 'ExtendedEvents', 'LinkedServers', 'Logins', 'PolicyManagement', 'ReplicationSettings', 'ResourceGovernor', 'ServerAuditSpecifications', 'ServerRoles', 'SpConfigure', 'SystemTriggers'
```
wsmelton, line 345: The way this has to be done begs for us to have an `-Include` parameter instead of `-Exclude`. // @potatoqualitee

wsmelton (changes requested):

functions/Export-DbaScript.ps1
wsmelton: This block of code needs to be removed, and line 158 can simply be changed to set it to `$true` since the parameter has been removed. Comment on line 157 can be deleted.

wsmelton: This block of code needs to be removed, and line 158 can simply be changed to set it to `$true` since the parameter has been removed. Comment on line 157 can be deleted.

wsmelton (changes requested):

functions/Export-DbaScript.ps1
lancasteradam: This code is for Export-DbaScript, which needs to keep the Append param to support its own set of use cases. We are only removing the Append param from the Export-DbaInstance command.

wsmelton (approved):

potatoqualitee:
🥳  thank you both!

potatoqualitee:
Shoot, I didn't realize the 1-, 2-, 3- was removed. Why was that?

potatoqualitee:
Shoot, I didn't realize the 1-, 2-, 3- was removed. Why was that, @lancasteradam ?

lancasteradam:
@potatoqualitee, the fileCounter was removed to ensure consistent filenames, whether they be written to an existing dir (the overwrite scenario) or to a new timestamped dir (the default scenario). By having predictable filenames it simplifies additional processing that the caller may want to do. The issue with keeping the fileCounter as part of the filename is that it could change depending on the -Exclude param. For example, on one call the customerrors.sql could have the '1-' prefix and on another call it could have '2-'.

#5380 can also be marked as closed. The topic of consistent filenames was mentioned here: https://github.com/sqlcollaborative/dbatools/issues/5380#issuecomment-509148821

Let me know if you agree on these items or if additional changes should be done.

PR #6959, opened 2020-11-06
Masking improve memoryusage

sanderstad:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
I already made improvements with the unique values being in a database instead of memory.
I wanted to do the same thing for the deterministic values to make the process more robust and better to use for larger data sets.

During the process of implementing data masking for several edge-case databases I found a couple of things that needed to improve as well.
One example of this was the way the data masking process handled the composite actions where it would mask certain column multiple times which was not efficient.

### Approach
The main approach was to use the same method of saving the deterministic values in the temp database and clean them up at the end of the process in the same way I did it for the unique values.

**Changes that were made:**
1. Moved deterministic values to table instead of memory
2. Changed the process of importing dictionaries to use the temporary deterministic table
3. Fixed tracking of time
4. Removed redundant code that was no longer used
5. Improved error handling
6. Fixed progress bars to show correct batch numbers
7. Fixed verbose message for progress
8. Improved composite actions 
9. Fixed value conversion for nullable columns with spaces as values
10. Fixed status message in return object when masking fails

### Commands to test
- Invoke-DbaDbDataMasking
- Convert-DbaMaskingValue


potatoqualitee:
One failure, so one pass! thanks @sanderstad  💯

PR #6969, opened 2020-11-12
New command - Install-DbaMultiTool

### LINKED ISSUE #6954: [Feature] Add installer for dba-multitool
### Summary of new feature

Hi 👋 all! I have a suite of utilities under the name [dba-multitool](https://github.com/LowlyDBA/dba-multitool) (formerly Express SQL). To be brief, they are:

> ... a suite of scripts for the long haul: optimizing storage, on-the-fly documentation, general administrative needs, and more. Each script relies solely on T-SQL to ensure it is secure, requires no third-party software, and can be installed in seconds.

As far as I know, no other projects cover the functional areas that my scripts do, so I think they would be a good value add to DbaTools. 

I've been working on these for over 2 years and don't have any plans to stop maintaining them.

Mimic `Install-DbaFirstResponderKit` and the like to install them from the source repository. I have a single installer script that handles them all. 

They don't (generally) support versions < 2012, so a version check would be the only deviation from the existing installer patterns. 

1.0.5 - Conroy

---

I am happy to make the appropriate PR if this request is approved 👍

LowlyDBA:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality, fixes #6954  )
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 

Add an installer for [dba-multitool](https://dba-multitool.org) T-SQL scripts as per #6954. There are currently four utility scripts: `sp_doc`, `sp_estindex`, `sp_helpme`, and `sp_sizeoptimiser`. 

As discussed in the issue, I am happy to work on `Invoke-` functions for these later if the interest arises. 

### Approach

More or less a copy of `Install-DbaFirstResponderKit` with a few minor tweaks. <s>Added an additional test scenario for 2008R2 to have better coverage for a "skipped" status for instances below the minimum version (2012).</s>

While I do have an all-in-one installer script, installing them individually seemed like a better option for more granular results.

### Commands to test

`Install-DbaMultiTool`


dbatools.psd1
```diff
@@ -540,6 +540,7 @@
         'Update-Dbatools',
         'Install-DbaWhoIsActive',
         'Install-DbaFirstResponderKit',
+        'Install-DbaMultiTool',
```
wsmelton, line 543: Please reset this file. We will handle updating the manifest, more than our psd1 has to be udpated.

tests/Install-DbaMultiTool.second.Tests.ps1
```diff
@@ -0,0 +1,36 @@
+$CommandName = $MyInvocation.MyCommand.Name.Replace(".second.Tests.ps1", "")
```
wsmelton, line 1: Please put all tests into the same file.

wsmelton (changes requested):

tests/Install-DbaMultiTool.second.Tests.ps1
LowlyDBA: That was my original approach, but doing so results in [this error](https://ci.appveyor.com/project/sqlcollaborative/dbatools/builds/36324572/job/yt1g6extt9mjtha1), that I am not sure the cause behind. 

Using the [second file](https://github.com/sqlcollaborative/dbatools/blob/development/contributing.md#appveyor-setup) approach (from contributing.md) results in that test succeeding. 

It is a minor addition to the existing tests for similar commands, so can also just be removed altogether potentially as well.

wsmelton: Your test only needs to run against `$script:instance3`. I do not see any reason to have it run on other versions.

tests/Install-DbaMultiTool.Tests.ps1
```diff
@@ -0,0 +1,131 @@
...
+            Add-Content $sqlScript (New-Guid).ToString()
+            $result = Install-DbaMultiTool -SqlInstance $script:instance2 -Database $database -Verbose:$false
+            $result[0].Status -eq "Error" | Should -Be $true
+        }
+    }
+    Context "Testing DBA MultiTool installer with LocalFile" {
+        BeforeAll {
+            $branch = "master"
+            $database = "dbatoolsci_multitool_$(Get-Random)"
+            $server = Connect-DbaInstance -SqlInstance $script:instance3
```
wsmelton, line 59: Please use the same instance across your test.

The AfterAll cleans up the objects created from within the last context so the use of the same instance in each context should not cause any contention.

wsmelton: Using all three instances we have built for Appveyor is not necessary for this command.

tests/Install-DbaMultiTool.Tests.ps1
```diff
@@ -0,0 +1,131 @@
...
+            $database = "dbatoolsci_multitool_$(Get-Random)"
+            $server = Connect-DbaInstance -SqlInstance $script:instance1
+            $server.Query("CREATE DATABASE $database")
+
+            $outfile = "DBA-MultiTool-$branch.zip"
+            Invoke-WebRequest -Uri "https://github.com/LowlyDBA/dba-multitool/archive/$branch.zip" -OutFile $outfile
+            if (Test-Path $outfile) {
+                $fullOutfile = (Get-ChildItem $outfile).FullName
+            }
+            $resultsLocalFile = Install-DbaMultiTool -SqlInstance $script:instance1 -Database $database -Branch $branch -LocalFile $fullOutfile -Force
```
wsmelton, line 113: This should also be possible by lowering the compatibility level of the database unless your stored procedures validate against the version of the instance instead.

LowlyDBA: The stored procedures do use the Major Version of the instance, not the compatibility level. That is why I was trying to use the 2008 instance for the skip tests.

wsmelton: Easiest way to handle that if you want to keep the test is mocking the connect command of our module where the property for the version is returned as 2008 R2.

LowlyDBA: Going to remove that test for the time being as it'd probably require re-writing portions of the actual function and I would like to hold off on that for the time being

LowlyDBA:
@wsmelton Last issue:

I switched all tests to use `instance3` in commit "[use one instances type for all tests](https://github.com/sqlcollaborative/dbatools/pull/6969/commits/cad6f68320627eb75f4a01176810ce5e0f71bccf)", but the resulting [appveyor build](https://ci.appveyor.com/project/sqlcollaborative/dbatools/builds/36335138) still tried to use 2008R2/2016:

image

The subsequent commit "[switch to 2016 instance](https://github.com/sqlcollaborative/dbatools/pull/6969/commits/06220341d6d56f7053dff103a3b99b9ddf304297)" changed them all to use `instance2` and that [built successfully](https://ci.appveyor.com/project/sqlcollaborative/dbatools/builds/36335400). Is there something I am missing as to why it failed?

wsmelton:
Instance failure seems like it is coming from Appveyor.

LowlyDBA:
Defaulted back to using the same test instance as `Install-DbaFirstResponderKit`, which makes the tests file more or less the same now.

wsmelton (approved):

niphlod (approved):

potatoqualitee:
thanks so much, all! 🥳  will merge, then add to psd1 and psm1 and publish shortly

PR #6979, opened 2020-11-17
[New Command] Move-DbaDbFile

### LINKED ISSUE #1806: [New-Command] Request: Move-DbaDatabaseFile
## Is this:
- [x] *feature* 
- [ ] *bug*:

## System Details

- Operating system name and version: Windows Server 2012 R2
```
PSVersion                      5.1.14409.1005
BuildVersion                   10.0.14409.1005
```

- SQL Server version for source/target: 2012, 2014, 2016

Can a cmdlet be created that will take standard arguments that must include:
SqlServer/Instance
DestinationDataFilePath
DestinationLogFilePath

Request based on the in-place upgrade process. If SQL installed using standard paths, upon upgrade (ex. 2014 to 2016) user databases will remain in `Drive:\Program Files\Microsoft SQL Server\MSSQL12.INSTANCENAME\MSSQL\Data`. 

IF upgrading with standard paths, the desired action would be to detach and reattach data files after moving to version specific directory, in this example `Drive:\Program Files\Microsoft SQL Server\MSSQL13.INSTANCENAME\MSSQL\Data` but MSFT leaves the files in the original version directory.

Example call to this new cmdlet would be:
```powershell
Move-DbaDatabaseFiles -Instance 'localhost\InstanceName' `
-Database 'myDbName' `
-DestinationDataFilePath 'MDFDrive:\Program Files\Path\To\My DB Root\Data' `
-Destination 'LDFDrive:\Program Files\Path\To\My DB Root\Data'
```

1. Retrieve current path using `Get-DbaDatabaseFile` and retrieve physical path of mdf and ldf files
2. Move or Copy mdf files from `($DBFileObj | ?{$_.TypeDescription -eq 'ROWS'}).PhysicalPath` to `$DestinationDataFilePath\$($DBFileObj.LogicalName).mdf`
3. Move or Copy ldf files from `($DBFileObj | ?{$_.TypeDescription -eq 'LOG'}).PhysicalPath` to `$DestinationDataFilePath\$($DBFileObj.LogicalName).ldf`
4. Offline/Single User mode the Database
5. Detach database/db files from SQL Instance
6. Attach database/db files to SQL instance using new path
7. Upon success dbcc, delete original files `$DBFileObj | %{ Remove-Item -Force -Path $_.PhysicalPath}`

ClaudioESSilva:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality, fixes #1806 )
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
A new command that allows move database files between paths.
Useful when we need to reorganize drives/mount points space or move a database to different paths.

### IMPORTANT
All my tests until now were done locally. 
Currently, I don't have access to an environment with remote instances to test the command.
If you are reading this and you can test it outside of the server, please let me know the results.
As soon as I can, I will also do those tests.

### Approach

We need to indicate which database and the logical name we want to move and to where.
1 - Database will be set offline.
2 - File(s) copied
3 - For each file copied the metadata will be updated to point to the new path
4 - Optional - The deletion of the source files just happens if the `-DeleteAfterMove` switch is specified.
5 - Database set online

The destination needs to be accessible by SQL Server. Using `Test-DbaPath` to validate that.

#### 3 Major call types
1 - Move files by type (Data / Log or both)
``` powershell
Move-DbaDbFile -SqlInstance sql2017 -Database dbatools -FileType Data -FileDestination "D:\DATA2"
``` 
image

2 - Specify which file you want to move using its logical name and the destination.
``` powershell
$fileToMove=@{
    'dbatools'='D:\DATA3'
    'dbatools_log'='D:\LOG2'
}
Move-DbaDbFile -SqlInstance sql2019 -Database dbatools -FileToMove $fileToMove
```
While copying
image

When finished with success
image

3 - If we need some help creating the hashtable with the mapping of logical name and current folders we can use
``` powershell
Move-DbaDbFile -SqlInstance sql2017 -Database dbatools -FileStructureOnly
```
image

### Commands to test

`Move-DbaDbFile`

### Learning

~~To do the copy, `Start-BitsTransfer` is used. During my tests, I found that sometimes ACLs were not replicated.
This leads to the database can't be set online as the moved data file can't be read by SQL Server.
To solve this I have used `-ACLFlags DACL` .~~

> ~~Copy the DACL ("Access Control List") information with the file.~~

As the parameter `-ACLFlags` of `Start-BitsTransfer` is not available on every PS version, I have a fallback to `Get-Acl ... | Set-Acl ... ` approach


niphlod:
will test asap

functions/Move-DbaDbFile.ps1
```diff
@@ -0,0 +1,366 @@
+function Move-DbaDbFile {
+    <#
+    .SYNOPSIS
+        Moves database files from one location to another.
+
+    .DESCRIPTION
+        Moves database files from one location to another.
```
wsmelton, line 7: I would recommend defining location. 

SQL Server 2016 brought support for storing data files on Azure Blob storage service...are we going to support that in this command? If not I would recommend layout what location means.

Example:

> Move database files from one local drive or folder to another.

wsmelton: Which to even move it to Azure Blob can only be done via backup/restore, but to prevent confusion or questions I'd say local drives/folders.

functions/Move-DbaDbFile.ps1
```diff
@@ -0,0 +1,366 @@
...
+        PS C:\> Move-DbaDbFile -SqlInstance sql2019 -Database dbatools -FileToMove $fileToMove
+
+        Declares a hashtable that says for each logical file the new path.
+        Copy each dbatools database file referenced on the hashtable on the sql2019 instance from the current location to the new mentioned location (D:\DATA3 and D:\LOG2 paths).
+        Before it puts database offline and after copy each file will update database metadata and it ends by set the database back online
+
+    .EXAMPLE
+        PS C:\> Move-DbaDbFile -SqlInstance sql2017 -Database dbatools -FileStructureOnly
+
+        Shows the current database file structure (without filenames).
```
wsmelton, line 88: I would show the output of this command to be a complete example...

functions/Move-DbaDbFile.ps1
```diff
@@ -0,0 +1,366 @@
...
+    begin {
+        if ((Test-Bound -ParameterName FileType) -and (-not(Test-Bound -ParameterName FileDestination))) {
+            Stop-Function -Category InvalidArgument -Message "You need to specify the -FileDestination parmeter. Quitting."
+        }
+    }
+
+    process {
+        if (Test-FunctionInterrupt) { return }
+
+        if (!$FileType -and !$FileToMove -and !$FileStructureOnly ) {
```
wsmelton, line 123: Please expand `!` to use `-not` for readability.

functions/Move-DbaDbFile.ps1
```diff
@@ -0,0 +1,366 @@
...
+                } else {
+                    # let's start checking if we can access .ComputerName
+                    $testPS = $false
+                    if ($SqlCredential) {
+                        # why does Test-PSRemoting require a Credential param ? this is ugly...
+                        $testPS = Test-PSRemoting -ComputerName $server.ComputerName -Credential $SqlCredential -ErrorAction Stop
+                    } else {
+                        $testPS = Test-PSRemoting -ComputerName $server.ComputerName -ErrorAction Stop
+                    }
+                    if (!($testPS)) {
```
wsmelton, line 206: Please use `-not`

functions/Move-DbaDbFile.ps1
ClaudioESSilva: I agree. I updated it.
Thank you, Shawn.

functions/Move-DbaDbFile.ps1
ClaudioESSilva: Done

functions/Move-DbaDbFile.ps1
ClaudioESSilva: Done

functions/Move-DbaDbFile.ps1
ClaudioESSilva: Done

ClaudioESSilva:
Found that `-ACLFlags` on `Start-BitsTransfer` is not available on every PS version. Changed to `Get-Acl -Path $physicalName | Set-Acl $destination`.

Found that `Test-PSRemoting` is not using `-UseSSL` by relying on our dbatools config. Added this.

ClaudioESSilva:
I was doing more testing and found two issues.

1 - When the database is already offline and we use the `-FileType Data` parameter, the selected filter is not being honoured.
2 - Old servers (win Server 2012) have `BitsTransfer` v1.0 which can't be used remotely. Error yield is: "The remote use of BITS is not supported". In this scenario, we need to fall back to `Copy-Item`. 
I will add the logic to the script block and this way one try should be enough.

I will also add an improvement. When running locally do not use `Invoke-Command2` (which will use PSRemoting)

wsmelton:
Oh please don't use `Copy-Item` 🥺 🙏 

Robocopy may be an option but I can't recall what version of Windows Server they started shipping it with. I know 2016+ includes but not sure about 2012/2012 R2.

wsmelton:
@niphlod are we good with this one?

If so I'll update the manifest so it gets added in next release.

functions/Move-DbaDbFile.ps1
```diff
@@ -0,0 +1,438 @@
...
+                'dbatools_log'='D:\Log'
+            }
+
+    #>
+
+    [CmdletBinding(SupportsShouldProcess, ConfirmImpact = "Medium")]
+    param (
+        [parameter(Mandatory, ValueFromPipeline)]
+        [DbaInstanceParameter]$SqlInstance,
+        [PSCredential]$SqlCredential,
```
wsmelton, line 106: Since remoting is involved to the server for supporting moving files over `Invoke-Command`, and use of `Start-BitsTransfer` this requires the need for `Credential` parameter as well.

Both commands accept a parameter of `-Credential` and the use of the `SqlCredential` should not be used for those calls.

functions/Move-DbaDbFile.ps1
```diff
@@ -0,0 +1,438 @@
...
+                if ([DbaValidate]::IsLocalhost($server.ComputerName)) {
+                    # locally ran so we can just use Start-BitsTransfer
+                    $ComputerName = $server.ComputerName
+                    $locally = $true
+                } else {
+                    # let's start checking if we can access .ComputerName
+                    $testPS = $false
+                    if ($SqlCredential) {
+                        # why does Test-PSRemoting require a Credential param ? this is ugly...
+                        $testPS = Test-PSRemoting -ComputerName $server.ComputerName -Credential $SqlCredential -ErrorAction Stop
```
wsmelton, line 219: This should pass `$Credential`

wsmelton: The comment can be removed as well. The parameter is required because utilizing WSMan requires elevated rights by default on a Windows Server.

functions/Move-DbaDbFile.ps1
```diff
@@ -0,0 +1,438 @@
...
+                    } else {
+                        $testPS = Test-PSRemoting -ComputerName $server.ComputerName -ErrorAction Stop
+                    }
+                    if (-not ($testPS)) {
+                        # let's try to resolve it to a more qualified name, without "cutting" knowledge about the domain (only $server.Name possibly holds the complete info)
+                        $Resolved = (Resolve-DbaNetworkName -ComputerName $server.Name).FullComputerName
+                        if ($SqlCredential) {
+                            $testPS = Test-PSRemoting -ComputerName $Resolved -Credential $SqlCredential -ErrorAction Stop
```
wsmelton, line 227: Should pass `$Credential`

functions/Move-DbaDbFile.ps1
```diff
@@ -0,0 +1,438 @@
...
+                                            Write-Verbose "Try copying using Start-BitsTransfer."
+                                            Start-BitsTransfer -Source $physicalName -Destination $destination -ErrorAction Stop
+                                        } else {
+                                            Write-Verbose "Can't use Bits. Using Copy-Item instead"
+                                            Copy-Item -Path $physicalName -Destination $destination -ErrorAction Stop
+                                        }
+
+                                        Get-Acl -Path $physicalName | Set-Acl $destination
+                                    }
+                                    Invoke-Command2 -ComputerName $ComputerName -Credential $SqlCredential -ScriptBlock $scriptBlock -ArgumentList $physicalName, $destination
```
wsmelton, line 310: This should pass `$Credential`

functions/Move-DbaDbFile.ps1
```diff
@@ -0,0 +1,438 @@
...
+                                # Try using UNC paths
+                                try {
+                                    $physicalNameUNC = Join-AdminUnc -ServerName $ComputerName -Filepath $physicalName
+                                    $destinationUNC = Join-AdminUnc -ServerName $ComputerName -Filepath $destination
+
+                                    if ($PSCmdlet.ShouldProcess($database, "Copying file $physicalNameUNC to $destinationUNC using UNC path for $ComputerName")) {
+
+                                        try {
+                                            Write-Message -Level Verbose -Message "Try copying using Start-BitsTransfer with UNC paths."
+                                            Start-BitsTransfer -Source $physicalNameUNC -Destination $destinationUNC -ErrorAction Stop
```
wsmelton, line 322: Should likely allow for `$Credential` as well but could leave it alone until we start seeing the use of the command for this use case.

functions/Move-DbaDbFile.ps1
```diff
@@ -0,0 +1,438 @@
...
+                                try {
+                                    if ($PSCmdlet.ShouldProcess($database, "Deleting source file $physicalName")) {
+                                        if ($locally) {
+                                            Remove-Item -Path $physicalName -ErrorAction Stop
+                                        } else {
+                                            $scriptBlock = {
+                                                $source = $args[0]
+                                                Remove-Item -Path $source -ErrorAction Stop
+                                            }
+                                            Invoke-Command2 -ComputerName $ComputerName -Credential $SqlCredential -ScriptBlock $scriptBlock -ArgumentList $physicalName
```
wsmelton, line 363: Should pass `$Credential`

wsmelton (changes requested):
- Suggestions on docs and messages
- Need to include the `Credential` parameter because we are triggering remote calls to the server.

niphlod:
we're good, let the users try it and let's see their usecases before overengineering it

sheldonhull:
Perfect timing. I have a gerry rigged hybrid version of using backup/restore and so on, and it's not elegant. It's pretty high risk option when the cleanup step gets involved if doing backup and restore vs move for example. 

Is this ready for usage as I'd love to give it a try to free up some space on disks if it's stable and just needing some real-world usage feedback. Much appreciate this getting some love. Wanted to contribute myself on it but no capacity at this time :-(

ClaudioESSilva:
@wsmelton thank you very much. I got underwater with other topics and only today was able to come back here.

ClaudioESSilva:
@sheldonhull I have already used once on work.
Like any other command please do your tests for your scenario. As @niphlod said we don't want to overengineer it, I would say it's ready for the most common scenarios. However, if you find anything you think can be improved let us know so we can discuss it.

PR #6982, opened 2020-11-20
Add "PasswordMustChange" to New-DbaLogin command

ClaudioESSilva:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Make possible to specify that login being created need to change password at next login

### Approach

When `-PasswordMustChange` is specified, we will enforce the `PasswordExpirationEnabled` and `PasswordPolicyEnforced` to true as they are required for the must change flag be turn on.

### Commands to test
New-DbaLogin with `-PasswordMustChange`

### Screenshots

image

### Learning
As `MustChangePassword` property of the login is just a `get`, we need to use the [ChangePassword](https://docs.microsoft.com/en-us/dotnet/api/microsoft.sqlserver.management.smo.login.changepassword?view=sql-smo-160#Microsoft_SqlServer_Management_Smo_Login_ChangePassword_System_Security_SecureString_System_Boolean_System_Boolean_) method

### DIFF
--- a/functions/New-DbaLogin.ps1
+++ b/functions/New-DbaLogin.ps1
@@ -513,8 +513,10 @@ function New-DbaLogin {
                         if ($currentPasswordMustChange -ne $newLogin.MustChangePassword) {
                             try {
                                 $newLogin.ChangePassword($SecurePassword, $true, $true)
-                                #$newLogin.Alter()
                                 Write-Message -Level Verbose -Message "Login $loginName has been marked as must change password."
+
+                                # We need to refresh login after ChangePassword. Otherwise, MustChangePassword will appear as False
+                                $server.Logins[$loginName].Refresh()
                             } catch {
                                 Write-Message -Level Verbose -Message "Failed to marked as must change password in $loginName on $instance using SMO."
                             }
@@ -525,7 +527,12 @@ function New-DbaLogin {
                         if ($usedTsql) {
                             $server.Logins.Refresh()
                         }
-                        Get-DbaLogin -SqlInstance $server -Login $loginName
+
+                        if ($loginType -eq 'SqlLogin') {
+                            Get-DbaLogin -SqlInstance $server -Login $loginName | Select-Object ComputerName, InstanceName, SqlInstance, Name, LoginType, CreateDate, LastLogin, HasAccess, IsLocked, IsDisabled, MustChangePassword
+                        } else {
+                            Get-DbaLogin -SqlInstance $server -Login $loginName
+                        }
                     } catch {
                         Stop-Function -Message "Failed to create login $loginName on $instance." -Target $credential -InnerErrorRecord $_ -Continue
                     }
--- a/functions/New-DbaLogin.ps1
+++ b/functions/New-DbaLogin.ps1
@@ -533,11 +533,8 @@ function New-DbaLogin {
                             $server.Logins.Refresh()
                         }
 
-                        if ($loginType -eq 'SqlLogin') {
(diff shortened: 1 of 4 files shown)


functions/New-DbaLogin.ps1
```diff
@@ -493,12 +508,31 @@ function New-DbaLogin {
...
+
                         #Display results
                         # If we ever used T-SQL, the smo is some times not up to date and should be refreshed
                         if ($usedTsql) {
                             $server.Logins.Refresh()
                         }
-                        Get-DbaLogin -SqlInstance $server -Login $loginName
+
+                        if ($loginType -eq 'SqlLogin') {
+                            Get-DbaLogin -SqlInstance $server -Login $loginName | Select-Object ComputerName, InstanceName, SqlInstance, Name, LoginType, CreateDate, LastLogin, HasAccess, IsLocked, IsDisabled, MustChangePassword
```
nvarscar, line 532: We should return a Login object in all cases. Select-Object would create a custom object instead. Did you want to use Select-DefaultView

functions/New-DbaLogin.ps1
```diff
@@ -493,12 +508,31 @@ function New-DbaLogin {
                                 }
                             }
                         }
+
+                        #Process the MustChangePassword property
+                        if ($currentPasswordMustChange -ne $newLogin.MustChangePassword) {
+                            try {
+                                $newLogin.ChangePassword($SecurePassword, $true, $true)
```
nvarscar, line 515: If it requires password change, then we need to validate that the password is specified when the MustChangePassword 
 parameter is used.

functions/New-DbaLogin.ps1
ClaudioESSilva: Excellent catch! 
Changed. Thanks.

functions/New-DbaLogin.ps1
ClaudioESSilva: Added validation on the top of the function.
Thanks!

functions/New-DbaLogin.ps1
nvarscar: Not a problem. I'm still wondering though why SqlLogin needs a special treatment here.

functions/New-DbaLogin.ps1
nvarscar: Looks good!

functions/New-DbaLogin.ps1
ClaudioESSilva: It's just my suggestion and how I would like it to have it. 

Windows login can't be created with must change pwd flag. Thus, don't make sense to output. 
However, if we are creating one (or more) SQL login and we specify the flag I find it useful to know if it needs to change the password. I see it like we have the isDisabled.

wsmelton: I see this handled better by `Get-DbaLogin` than with the new function. Just add this logic to the Get command.

wsmelton: The same reason could be said on the output itself of `Get-DbaLogin`..so no real reason to not just do it there.

ClaudioESSilva: - Rollback the change of the output on the New-DbaLogin
- Added MustChangePassword to the default output of Get-DbaLogin
- Took the chance to add a new switch parameter -MustChangePassword that will just output logins that need to change password.

Do you agree?

potatoqualitee:
Thanks to everyone who helped review and thank you @ClaudioESSilva for adding this functionality 🙇

PR #6988, opened 2020-11-23
Update Get-DbaLogin.ps1

williammartins40:
Correct me if I'm wrong, but currently I haven't found an exit for the 'Detailed' parameter.

So I suggest you add the detail output and create the 'else' if the detailed output is false.

## Type of Change

 - [ X ] Adding code coverage to existing functionality
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system


niphlod:
hum .... the fact that there are some detailed properties and that by default those are not shown is not the same thing .... when you pass `-Detailed` you can loop over the results and do, e.g., `$login.BadPasswordCount` or  `$results | Select-Object Name, BadPasswordCount`

williammartins40:
Thanks,

Applying `$results | Get-Member` I could see all the properties, so I will close this pr.

williammartins40:
@niphlod helped me to better understand the Get-DbaLogin process and that there would be no need to change the detailed item.

PR #6338, opened 2020-01-26
Modifications to fix issues with Azure SQL Db

sqllensman:
## Type of Change

 - [X] Bug fix (non-breaking change, fixes #6337 )
 - [ ] New feature (non-breaking change, adds functionality, fixes # )
 - [ ] Breaking change (effects multiple commands or functionality, fixes # )
 - [ ] Ran manual Pester test and has passed (`.\tests\manual.pester.ps1)
 - [ ] Adding code coverage to existing functionality
 - [ ] Pester test is included
 - [ ] If new file reference added for test, has is been added to github.com/sqlcollaborative/appveyor-lab ?
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Fix issues with  Write-DbaDataTable when using an Azure SQL Database as targe

### Approach
Add test to avoid appending Database name to name of table used for Bulk Copy
This is not supported for an Azure SQL Database

Removed code that excluded adding database name to connection string for Azure databases

### Commands to test
Write-DbaDataTable

Have run successful manual tests against Azure SQL Database and Azure Managed Instance and normal SQL Server installation

PR #7009, opened 2020-12-02
Write-DbaDbTable - Reuse SMO Connection

awickham10:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Re-use the existing SMO connection when performing the SQL bulk copy. Prior to this change a temporary table could not be used because the session the table was created in was different than session used to copy the data.

### Approach
Instead of passing a connection string to the SqlBulkCopy constructor, copy the connection object.

### Commands to test
Write-DbaDbTable with a temporary table.

### DIFF
--- a/functions/Write-DbaDbTableData.ps1
+++ b/functions/Write-DbaDbTableData.ps1
@@ -491,7 +491,7 @@ function Write-DbaDbTableData {
             }
         }
         # Create SqlBulkCopy object - Database name needs to be appended as not set in $server.ConnectionContext
-        $bulkCopy = New-Object Data.SqlClient.SqlBulkCopy("$($server.ConnectionContext.ConnectionString);Database=$databaseName", $bulkCopyOptions)
+        $bulkCopy = New-Object Data.SqlClient.SqlBulkCopy($server.ConnectionContext.SqlConnectionObject, $bulkCopyOptions, $null)
 
         $bulkCopy.DestinationTableName = $fqtn
         $bulkCopy.BatchSize = $BatchSize


awickham10:
My one concern with this is "Database" is no longer being passed in the connection. Maybe this needs to be added to the `$server = Connect-DbaInstance`.

potatoqualitee:
Thanks @awickham10! @andreasjordan what do you think of this change?

andreasjordan:
I try to test this today.

andreasjordan:
Hi @awickham10 , what is the reason to use the constructor `SqlBulkCopy(SqlConnection, SqlBulkCopyOptions, SqlTransaction)` with a SqlTransaction of $null instead of just using `SqlBulkCopy(String, SqlBulkCopyOptions)`?

awickham10:
> Hi @awickham10 , what is the reason to use the constructor `SqlBulkCopy(SqlConnection, SqlBulkCopyOptions, SqlTransaction)` with a SqlTransaction of $null instead of just using `SqlBulkCopy(String, SqlBulkCopyOptions)`

There were only two constructors that accepted the SqlConnection object - one with just the connection and the other with all three. 

My biggest concern is the database is no longer passed to the SqlBulkCopy so will this end up being a breaking change for a bunch of people? I haven't tested out the before/after.

andreasjordan:
@awickham10 : Sorry, yes, you are right with the constructors - it was to early in the morning.

wsmelton:
Just an FYI that this will break Azure SQL access. The database was being appended in that manner explicitly because of how SMO connection is created it will at times have `master` set as the database. This method was used in order to support Azure SQL DB last I recall.

A quick test on this PR shows it does break:

image

wsmelton:
For reference #6338 is where that line of code was explicitly added to correct connecting to Azure SQL DB.

wsmelton:
Is there an issue tracking why the change in this PR is required?

andreasjordan:
Looking at PR #6338: What about adding the `if ($server.isAzure) {` again?

wsmelton:
There is no issue sharing why this change is required from the start. What is broken with the command as-is?

andreasjordan:
> My biggest concern is the database is no longer passed to the SqlBulkCopy so will this end up being a breaking change for a bunch of people? I haven't tested out the before/after.

@awickham10 : You definitly need to add ` -Database $Database` in line 425 to have the database in the connection context. But that gives you another error: WriteToServer requires an open and available connection. The current status of the connection is 'Closed'.

I work on this because I like to learn how the commands work. And I think this change might (!) help to reuse connections. But Shawn is right: Do we need this change? What is broken?

andreasjordan:
Ok, the error is only thrown when using the new experimental code path of Connect-DbaInstance, because then the connections are closed very fast. This is needed just before the constructor:
```
        if (-not $server.ConnectionContext.IsOpen) {
            $server.ConnectionContext.SqlConnectionObject.Open()
        }
```
Then it "works on my maschine".

awickham10:
> There is no issue sharing why this change is required from the start. What is broken with the command as-is?

You cannot use Write-DbaDbTable with a temp table because the bulk copy command is in a separate session.

wsmelton:
@awickham10 You can if they are global temp tables (i.e., `##temptb`) instead of session only (i.e., `#temptb`).

awickham10:
I was just using a temp table as a staging table for some reference data. While troubleshooting it I noticed the issue and a possible solution existed so I pursued it since I had already been investigating what was going on. I'll close this out.

wsmelton:
I'm not apposed to addressing support for session temp tables of that is the consensus. We need to make sure functionality with Azure SQL is not broken. (Our test won't catch it being broke.)

andreasjordan:
@potatoqualitee : You asked me what I think of this change. I did some tests yesterday and thought about it.

I like the change. Only one connection is used, which leads to the fact that temporary tables can be used which are only accessible in the current session. And with frequent use the connection limit is not reached as quickly. Also the general procedure to establish a connection with Connect-DbaInstance and then use it is better implemented.

If the risk is considered high, we can only activate the change if the experimental codepath of Connect-DbaInstance is used.

PR #7015, opened 2020-12-04
Copy-DbaDatabase - Add EA to internal function

### LINKED ISSUE #6915: Copy-DbaDatabase fails when destination is a mountpoint
On version 1.0.123

The destination server have the same folder structure as the source, but one of the destination folders is a drive mapped to a path (which we use on all our SQL Servers instead of drive letters)

Copy-DbaDatabase -Source sqlmigratesql2.agr.bah -Destination sqlmigratesql4.agr.bah -DestinationSqlCredential $sacred -AllDatabases -DetachAttach -ReuseSourceFolderStructure -Reattach

First a warning is displayed:
WARNING: [08:58:26][Mount-SqlDatabase] Issue mounting database. | Unable to open the physical file
"c:\mssql\crmdata\agrtmp_Primary.mdf". Operating system error 2: "2(The system cannot find the file specified.)".
CREATE DATABASE failed. Some file names listed could not be created. Check related errors.
Type                    Name                 Status  Notes
----                    ----                 ------  -----
Database                xxxxxxxbooking       Skipped Already exists on destination
Database (DetachAttach) agrtmp               Failed  Failed to attach database to destination
And when trying to do that database the following error occurs:
Start-BitsTransfer : The destination volume has changed. If the disk is removable, it might have been replaced with a d
ifferent disk. Reinsert the original disk and resume the job.
At C:\Program Files\WindowsPowerShell\Modules\dbatools\1.0.123\allcommands.ps1:5327 char:29
+ ...             Start-BitsTransfer -Source $from -Destination $remotefile ...
    + CategoryInfo          : InvalidOperation: (:) [Start-BitsTransfer], Exception
    + FullyQualifiedErrorId : StartBitsTransferCOMException,Microsoft.BackgroundIntelligentTransfer.Management.NewBits
   TransferCommand

It seems that the Start-BitsTransfer is the culprit and the code for the transfer (Start-SqlFileTransfer) never falls back to the copy-item command in the third try-catch block (which it seems the code intended to)

Trying the command below result in the same error
Start-BitsTransfer -Source .\myfile.txt -Destination \\sqlmigratesql4.agr.bah\c$\mssql\crmdata\myfile.txt

System, Administrators and MSSQLSERVER have full permissions on the folder (and mapped drive) - relaxing the permissions doesn't solve the BITS problem, so the bug is that the Copy-Item never gets executed

wsmelton:
Adding EA to the `Start-BitsTransfer` so the catch block and final `Copy-Item` is properly hit.

## Type of Change

 - [x] Bug fix (non-breaking change, fixes #6915 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system

### DIFF
--- a/functions/Copy-DbaDatabase.ps1
+++ b/functions/Copy-DbaDatabase.ps1
@@ -279,7 +279,9 @@ function Copy-DbaDatabase {
             return
         }
 
-        if ($Force) { $ConfirmPreference = 'none' }
+        if ($Force) {
+            $ConfirmPreference = 'none' 
+        }
 
         function Join-Path {
             <#
@@ -422,7 +424,7 @@ function Copy-DbaDatabase {
                             if ($destServer.VersionMajor -lt 10) {
                                 $directory = "$directory\FTDATA"
                             }
-                            $fileName = Split-Path($physical) -leaf
+                            $fileName = Split-Path($physical) -Leaf
                             $d.physical = "$directory\$fileName"
                         }
                         $d.logical = $logical
@@ -490,7 +492,7 @@ function Copy-DbaDatabase {
             $fileStructure = [pscustomobject]@{
                 "databases" = $dbcollection
             }
-            Write-Progress -id 1 -Activity "Processing database file structure" -Status "Completed" -Completed
+            Write-Progress -Id 1 -Activity "Processing database file structure" -Status "Completed" -Completed
             return $fileStructure
         }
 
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Thank you! 🥳

wsmelton:
If we don't want to deal with the new format style that Code enforces with PSSA we should add exclusion to turn that off for our workspace.

Can't remember when it started doing that.

potatoqualitee:
That'd be great! It supposedly shouldn't be impacting us. Well, our tests aren't failing about it because we're using an older version, but new versions are doing it. Niph suggested we submit a bug report to PSScriptAnalzyser, but I haven't had a chance yet. Is that something we can ignore using our .vscode settings?

wsmelton:
It's not a bug in formatting really because it is the proper casing for the parameters. They have always been in all caps. Code or the extension just never enforced it.

If we don't want it to be I'll see if I can find a way to block it.

PR #7022, opened 2020-12-08
update constants to support devcontainer

wsmelton:
Add support for our test framework to be used within devcontainers.

The environment variables within each option (GitHub CodeSpaces and VS Code Remote Container) offer the ability to detect whether we are running within that environment:

GitHub CodeSpaces: `$env:CODESPACES` (only one I think should be needed)
VS Code: `$env:REMOTE_CONTAINER` and `TERM_PROGRAM = vscode`


potatoqualitee:
awesome! merciii

PR #7023, opened 2020-12-08
New command: Remove-DbaDbTableData

### LINKED ISSUE #1009: Remove-DbaDataByBatch

feature, post 1.0. Unsure of the name, the current one kinda sucks but "Chunk" is gross.

### Details
This post by aaron bertrand is useful and can be implemented using PowerShell and some calculations. 

https://sqlperformance.com/2013/03/io-subsystem/chunk-deletes

lancasteradam:
This new command addresses #1009 by providing a command that can table deletes in batches. The following features are included:

- Param for log backup file path
- Param for batch size
- Param for table name (optional if the DeleteSql param is passed in)
- Param for DeleteSql statement for advanced control by the caller
- Timings are recorded and returned to the caller
- Checkpoints are performed for a simple model db and log backups for full/bulk logged
- Integration tests are included
- Documentation and examples

## Type of Change

 - [x] New feature (non-breaking change, adds functionality, fixes #1009 )
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [x] Documentation
 - [ ] Build system
 

### Purpose
 
The purpose of this PR is to add a new command that can do batch deletes as described in #1009.

### Approach

This command is structured similarly to Remove-DbaDbData. The difference is that the deletes are done in batches so that the log file size can be controlled.

### Commands to test

See the integration tests.

### Learning
As mentioned in #1009 the posting by Aaron Bertrand is the example that this command is patterned from.


potatoqualitee:
WOW! How fun!! @Stuart-Moore check it out 🤩  Thank you @lancasteradam

lancasteradam:
@potatoqualitee ,  thanks! Dbatools is a great project to contribute to.

functions/Remove-DbaDbTableData.ps1
```diff
@@ -0,0 +1,306 @@
+function Remove-DbaDbTableData {
+    <#
+    .SYNOPSIS
+        Removes table data using a batch technique from a database(s) for each instance(s) of SQL Server.
+
+    .DESCRIPTION
+        This command does a batch delete of table data using the technique described by Aaron Bertrand here: https://sqlperformance.com/2013/03/io-subsystem/chunk-deletes. The main goal of this command is to ensure that the log file size is controlled while deleting data. This command can be used for doing both very large deletes or small deletes. Foreign keys are not temporarily removed, so the caller needs to perform deletes in the correct order with dependent tables or enable cascading deletes. When a database is using the full or bulk_logged recovery model this command will take log backups at the end of each batch. If the database is using the simple recovery model then CHECKPOINTs will be performed. The object returned will contain metadata about the batch deletion process including the log backup details.
+
+    .PARAMETER SqlInstance
```
wsmelton, line 9: We likely need some validation control for this parameter to ensure it is not an Azure SQL instance because this command won't work against those. Since it is not calling `Connect-SqlInstance` directly we can't use the `-AzureUnsupported`

Stuart-Moore:
Hi. Great command, and thanks for submitting it. But......

The unannounced log backup could catch people out, not everyone reads the cbh, and copy and paste from t’internet is common for accidental DBAs. 

I agree you’re doing best practice.  Just that shoving a log backup somewhere someone isn’t expecting it to be could cause a problem if someone has to recovery afterwards.

Might be better to have the log backup as an optional extra rather than a default?

wsmelton:
I'd probably expand that to add checks to know if the database is involved in special configurations:

- Log Shipping (will cause their LS to go out of sync)
- Availability Group (would only error on secondaries)
- Mirroring (same as AG)
- Replication (not sure adverse effects here)

niphlod:
Just my 2 cents: I think an "order by" is missing, at least as an option (e.g. I can sustain failure within the loop, but in the zillions of rows to be deleted I wanna be sure every loop is committed with a proper order).
Also, I usually run these in low deadlock priority, via "SET DEADLOCK_PRIORITY LOW;". Also, prolly it's better to use "SET NOCOUNT ON;SET XACT_ABORT ON;" for performance & precaution.

lancasteradam:
Thanks everyone for the feedback. I'll look into these items.

lancasteradam:
I've incorporated the items mentioned above. Let me know if these are on the right track. 

@wsmelton , let me know if there is something different you'd like to see for the special configurations. 

@niphlod , for the 'order by' scenario I agree it is important. We could use the 'WhereSql' param to accomplish that, so I added an example to the command description. Let me know if you agree with that approach. I think this will help to keep the command more simple and have fewer parameters.

niphlod:
hi, that's not an orderby... it's cheating with a performance hit and a really bad example, as you'd need to repeat the where statement both in the delete and in the "IN" subselect ;-) . At this point accepting a "-Stmt" will do the exact same thing, at the cost of being reeeeeally verbose (not that much more, with the orderby added)...
As we're pushing for "best practices" I think the code will look cleaner and be more snappy with something like this
```
WITH ToDelete as (
	SELECT TOP (<howmanyrows>) onecol(preferably the pk, or *)
	FROM <tbname>
	WHERE <wherestmt>
	ORDER BY <whatever>
) 
DELETE FROM ToDelete;
```
if the instance is at >= 2005 . For 2000, either you "complicate the code" or you complicate examples or .... don't support it.

lancasteradam:
@niphlod , thanks for that feedback. I see what you mean. I like the idea of having a -Stmt param instead of the -FromSql and -WhereSql params. It will further reduce complexity of the command and provide the flexibility we are looking for. The command will then support two categories of use cases:

1. Simplistic batch deletion from a single table via the -Table param
2. Advanced batch deletion based on a specified SQL delete statement via the -DeleteStmt param

potatoqualitee:
@niphlod give me the approval and i'll merge 👍

functions/Remove-DbaDbTableData.ps1
niphlod: @lancasteradam : is this still valid ? I feel that removing data in batches when done on Azure is just a matter of skipping the log backup no matter what, given backups are "managed"

niphlod:
whoops, missed the notification. Added a relevant comment for @lancasteradam

functions/Remove-DbaDbTableData.ps1
```diff
@@ -0,0 +1,326 @@
...
+            }
+
+            if ($DeleteSql -inotmatch "delete") {
+                Stop-Function -Message "The -DeleteSql param must be a DELETE statement with a TOP (N) clause. See the command description for more details."
+                return
+            }
+        }
+
+        if (Test-Bound LogBackupPath) {
+            $null = Test-ExportDirectory -Path $LogBackupPath
```
niphlod, line 145: Here Test-DbaPath should be used instead. Test-ExportDirectory tests if the path is there from the perspective of the account running PS, Test-DbaPath instead does it via the SQL service account. If the result is False, a warning should be raised and the function stopped.

lancasteradam: ok, thanks, I'll make that change.

functions/Remove-DbaDbTableData.ps1
lancasteradam: @niphlod, I think it is still valid. I like the suggestion that @wsmelton made about ensuring it is not an Azure SQL instance. I can add validation code for that. This will keep the command simpler and focused for on-prem scenarios. Let me know if you agree with that approach.

niphlod: hum, I'm not convinced...why can't we allow removing table data in batches for Azure ? It's Remove-DbaDbTableData, not Remove-DbaDbTableDataAndBackup. @wsmelton , @Stuart-Moore  your thoughts on this ? I feel we can make so that if you don't pass a LogBackupPath backups are not taken (and issue a warning?). This could "solve" the issue of "shoving a log backup".
IMHO it's perfectly fine to "defaulting" to on-premise "logic" of "remove-and-backup" but on Azure you can just leverage the "remove" logic without the backup part. BTW @lancasteradam sorry this is taking so long but it's going to be an outstanding function ^_^

wsmelton: The cautionary tale that exist with Azure SQL is that because you don't control the backups large batch deletion can cause 40552 errors. This means Azure SQL killed the transaction because of log space.

If error handling is in place to catch that properly, then I'm good with including it. Users will just have to be aware of the limits with deleting data on Azure SQL DBs.

niphlod: that try/catch should intercept it and later on the function checks for an empty errormessage before continuing so I think it's perfectly managed there. At this point the missing bit to accomodate all reviews is to make the backup optional (if a LogBackupPath is not passed).

lancasteradam:
Thanks everyone for the feedback. I've incorporated the latest suggested changes. Let me know if anything else needs updating.

functions/Remove-DbaDbTableData.ps1
```diff
@@ -0,0 +1,350 @@
...
+            }
+
+            foreach ($db in $dbDatabases) {
+
+                $instance = $db.Parent
+
+                if (Test-Bound LogBackupPath) {
+                    $pathCheck = Test-DbaPath -SqlInstance $instance -Path $LogBackupPath
+                    if (-not $pathCheck) {
+                        Stop-Function -Message "The service account for $instance.Name is not able to create log backups in $LogBackupPath."
```
niphlod, line 236: This should be just $instance

lancasteradam: thanks, I'll make that change

lancasteradam:
@wsmelton and @niphlod, let me know if any additional changes are needed for this new command.

niphlod (approved):
nothing on my end, looks good for a first round of eager users to test it

PR #7040, opened 2020-12-19
Save-DbaKbUpdate - Additional tags for docs

wsmelton:
Add tags to have `Save-DbaKbUpdate` included with `Update-DbaInstance`


wsmelton:
Be nice if our test would consistently pass in PRs 😞

wsmelton:
Dang it...missed the branch being set wrong.

PR #7046, opened 2020-12-20
Copy-DbaDbTableData – workaround for integer overflow/wraparound in the legacy bulk copy library

### LINKED ISSUE #6927: Copy-DbaDbTableData displays negative row count in progress area for billions of rows assuming 32 UINT casting
powershell.exe
Windows Server 2019
EN-US
SQL server

lancasteradam:
Copy-DbaDbTableData – workaround for integer overflow/wraparound in the legacy bulk copy library

- Added an inline function to adjust the count of total rows copied when the legacy bulk copy library experiences integer wraparound/overflow

## Type of Change

 - [x] Bug fix (non-breaking change, fixes #6927 )
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [x] Documentation
 - [ ] Build system
 

### Purpose
It seems that the legacy bulk copy library is still using a 4 byte integer to keep track of the number of rows copied. These links below all indicate a 4 byte int is used:

https://docs.microsoft.com/en-us/sql/relational-databases/native-client-odbc-how-to/bulk-copy/bulk-copy-a-select-result-set-odbc?view=sql-server-ver15 (sample code uses an sdword)

https://stackoverflow.com/questions/1188384/sqlbulkcopy-row-count-when-complete (open discussion about what would happen for higher row count values exceeding the range of a signed int)

https://docs.microsoft.com/en-us/sql/relational-databases/native-client-odbc-extensions-bulk-copy-functions/bcp-exec (sample  code uses an int)

### Approach
Use a running total and do arithmetic at the times when the signed integer wraps around from positive to negative values and negative to positive values (see the screenshots below to see the row count progression).

Alternate approach discarded:  Run a select count(1) query to determine how many rows were added each time Add_SqlRowsCopied is invoked. This approach won't work since the bulk copy has a tablock and running select count(1) on a very large table would add performance overhead to the command.

I'm open to other approaches if there is a better way to solve the issue. 

If anyone has some inside contacts or knowledge on the legacy bulk copy library it would be interesting to confirm the issue and find out if it is something that can be fixed by Microsoft. 

This issue affects other dbatools commands that use Add_SqlRowsCopied:
Import-DbaCsv.ps1
Write-DbaDbTableData.ps1

Once we've reached a consensus on the approach I can update those two commands.

### Commands to test
The examples and integration tests have example commands.

### Screenshots
dbatools 6927 rowsCopied field

dbatools 6927

Handling of integer wraparound
6927 fix positive to negative

6927 fix negative to positive

### DIFF
--- a/functions/Copy-DbaDbTableData.ps1
+++ b/functions/Copy-DbaDbTableData.ps1
@@ -37,13 +37,13 @@ function Copy-DbaDbTableData {
     .PARAMETER Table
         Specify a table to use as a source. You can specify a 2 or 3 part name.
         If the object has special characters please wrap them in square brackets.
-        
+
         Note: Cannot specify a view if a table value is provided
 
     .PARAMETER View
         Specify a view to use as a source. You can specify a 2 or 3 part name (see examples).
         If the object has special characters please wrap them in square brackets.
-        
+
         Note: Cannot specify a table if a view value is provided
 
     .PARAMETER DestinationTable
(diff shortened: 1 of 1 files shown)


potatoqualitee:
this is soooooo good! thank you @lancasteradam 🥳  hope you're having a wonderful holiday season. thank you for all of your contributions this year 🙇

lancasteradam:
Thanks @potatoqualitee. Likewise Happy Holidays to you and the entire dbatools team!

PR #7049, opened 2020-12-21
Get-DbaHelp. refactor logic to internal

niphlod:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Refactor the entry point to get a more parseable object suitable for further processing.
Adds the same logic of https://github.com/sqlcollaborative/docs/blob/multiple_pages/assets/templates/to_md.ps1 to be able to output the MD "representation" of the help.
People wanting to refactor help now just need to do
```
Import-Module .\dbatools.psm1 -Force
Get-DbaHelp Get-DbaHelp -OutputAs "MDString" | Out-File Get-DbaHelp.md
& code Get-DbaHelp.md
```
which is still cumbersome but in order to get the help you need to reload the module anyway so..... one step closer.
Previously you needed to know a whole lot of things in order to understand what to touch and how to preview changes
https://github.com/sqlcollaborative/docs/issues/46#issuecomment-736089862
I'm unsure if it's time to have a public function that gets the help in MD just to avoid loading the psm1.

### Approach
Refactor what was inside Find-DbaCommand and add the bits to transform into MarkDown the inline help


wsmelton:
Are we trying to rewrite the help system that PowerShell offers?

When we publish the module the help is dumped into an XML file. (Can't recall if we strip it from the function itself right off).

niphlod:
lol, I swear this happens every time I push something for the docs... we're not trying to do anything if not help contributors doing their thing faster. 
In this case help amend the inline help so it's readable. 
And the "amendments" will benefit docs.dbatools.io (which is seemingly not reeeeeally your favourite) or whatever tool will be used in the future to produce documentation.
Every time I'm doing a recap, sorry if you are already informed on some of those.
dbatools does not maintain external sources for the help. All comes from the inline docs, which are a requirement for every function.
Find-DbaCommand was written to circumvent a really sluggish performance when searching a command (and/or keywords, and/or tags, and/or author, and/or lots of other things). It's been there for a while. 
docs.dbatools.io just parses the same file, and builds the SPA.
We're approaching the "single page for each command" agenda, in order to be more SEO friendly over at sqlcollaborative/docs (multiple_pages branch, long awaited).

Doing the logic here instead of on docs.dbatools.io to have one page per command has no cons, and the added benefits that:
- there will (hopefully) be a single "logic" to "adjust" the inline help to be more in line with a markdown help
- you can preview what the help will look like without rebuilding the help of the entire module
- a foreach loop is all that's needed to have an offline portable help in MD, which wasn't there before

potatoqualitee:
internal looks great to me! thanks for all of your help with the docs 🙇  !! hope you have an awesome new year 🥳  sorry ive been slightly absent, im obsessed with a couple other projects.

PR #7071, opened 2021-01-10
Invoke-DbaDbPiiScan - Update to code base and improvements

### LINKED ISSUE #6801: Return all scanned Pattern from a single field
### Summary of new feature
Function : Invoke-DbaDbPiiScan

* Current function Invoke-DbaDbPiiScan returns only one row per field(column), for the very first hit it finds in Jason file. For example, If we have a column with data belonging to multiple PII categories (like SSN, Email, Credit Cards), it would be good to return the details for each category as a separate output (3 output for above example). It could return duplicate records so making it an optional parameter could be an idea for this.

* Probably an inner loop with output should do and adding this feature as an optional parameter

1.0.115

lancasteradam:
- Refactored the scan logic so that all known names and patterns are checked. The results are a unique set based on the criteria of the known names/patterns.
- Changed the matching logic so that it can obtain all matches and not just the first as requested in #6801.
- Added integration tests for the other command parameters
- Fixed a bug with multiple column names passed in for -Column
- Fixed a bug related to loading the default known names and patterns (the command param variables KnownNameFilePath and PatternFilePath were being overwritten)
- Added a note to the documentation on how to create the custom scan definitions files.

## Type of Change

 - [x] Bug fix (non-breaking change, fixes #6801 )
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [x] Documentation
 - [ ] Build system
 

### Purpose
 
Addressing the issue described in #6801 so that scan results are complete. Also addressed a few minor issues encountered along the way.

### Commands to test

See the integration tests.


functions/Invoke-DbaDbPiiScan.ps1
```diff
@@ -37,10 +38,10 @@ function Invoke-DbaDbPiiScan {
         Amount of rows to sample to make an assessment. The default is 100
 
     .PARAMETER KnownNameFilePath
-        Points to a file containing the custom known names
+        Points to a file containing the custom known names. Custom scan definitions can be specified using the format seen in <dbatools module root>\bin\datamasking\pii-knownnames.json.
 
     .PARAMETER PatternFilePath
-        Points to a file containing the custom patterns
+        Points to a file containing the custom patterns. Custom scan definitions can be specified using the format seen in <dbatools module root>\bin\datamasking\pii-patterns.json.
```
wsmelton, line 44: Users should not be modifying these files.

sanderstad (approved):
Nice addititions. Thank you for that!

PR #7100, opened 2021-01-19
Fix Get-DbaDatabase join for Azure databases

paddycull:
Databases where the owner is not in the sys.database_principals table do not get returned due to an INNER JOIN. Updated this to LEFT JOIN so all Azure databases are returned.

## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
I found that some of my Azure databases were not getting returned when I ran Get-DbaDatabase. I discovered that the reason for this is because the Database Owner's for the missing databases did not have an entry in sys.database_principals, and due to the query in line 261 in Get-DbaDatabase.ps1 being an "INNER JOIN" between "sys.databases" and "sys.database_principals", if the Database Owner SID is not in the sys.database_principals then that database does not get returned. 

### Approach

I've updated that to a "LEFT JOIN" so all databases get returned, even if the database owner can not be found within sys.database_principals. The function still returns the Azure Active Directory generated owner string if it is not found within sys.database_principals.

### Commands to test

````
Get-DbaDatabase  -SqlInstance "azureinstance.database.windows.net" -SqlCredential $Cred
````

### Screenshots

Here is the functionality before changing the query to a "LEFT JOIN";

Databases where the owner is not in the sys.database_principals table do not get returned. Here is the database list in Azure - I have removed the names, but there are 8 user databases listed.
image

And here is the output of Get-DbaDatabase for the same instance. Again, names are removed, but there are only 5 user databases returned;
image

After updating the command to a "LEFT JOIN", all databases are returned as expected, all 8 user databases are returned. I've included part of the "Owner" of the databases to demonstrate the newly returned databases and owner string. 
image


wsmelton:
How is a database created in Azure SQL with no owner?

paddycull:
The owner of the database is an Azure Active Directory user - in Database Properties that comes up as an Azure generated string;
image

But that user does not exist in sys.database_principals, so it can't be resolved and the database does not get returned due to the INNER JOIN clause.

paddycull:
Just to add to the above, I done some digging and found the owner of the database is part of an AAD group that has been given admin access to the instance via the Azure portal - so the user has admin access, but his individual account has not been explicitly added to the instance itself, which is why the user not in sys.database_principals,

potatoqualitee:
hey @paddycull - so sorry for the delay. This part of the lockdown has been brutal. We'll work to get your PR merged this week ❤️  Thank you for your patience!

paddycull:
No problem at all! 

That would be brilliant , thanks ! 😄

potatoqualitee:
Thanks so much, sorry again for the delay. The lockdown is still brutal 😓  This should be published to the gallery today 👍

paddycull:
Awesome, thanks very much ! 🙂

PR #7105, opened 2021-01-21
 Write-DbaDbTableData: Moving Connect to Server step to the beginning of the code.

miguelmolgar:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
I saw that in command Write-DbaDbTableData $server variable was called (line 394) before being initialised in line 430. So I moved the block of code where the connection is set to the beginning, as the first step of the command. 
### Approach

Moving part of the code will make that the connection of the server is made before checking if it is an Azure SQL server or not.
### Commands to test

Write-DbaDbTableData with an Azure SQL Server using a private enpoint.

### DIFF
--- a/functions/Write-DbaDbTableData.ps1
+++ b/functions/Write-DbaDbTableData.ps1
@@ -399,7 +399,7 @@ function Write-DbaDbTableData {
         $tableName = $fqtnObj.Name
 
         $quotedFQTN = New-Object System.Text.StringBuilder
-       
+
         if ($server.ServerType -ne 'SqlAzureDatabase') {
             <#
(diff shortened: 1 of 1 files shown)


miguelmolgar:
Bad branch to merge (master instead of development)

wsmelton:
That is by design @miguelmolgar . There are internal functions within the command that reference an object created further down in the workflow. The command does not have any issue that I'm aware of functioning against Azure SQL this way. If it does please create an issue to show more details.

PR #7150, opened 2021-02-19
Add config item to bypass CredSSP initialization

### LINKED ISSUE #7148: Configuration setting to bypass Initialize-CredSSP
### Summary of new feature

Create a configuration setting to bypass initialization of CredSSP authentication by `Initialize-CredSSP`.
Allow CredSSP authentication to be used, but only using the existing configuration already in place.

The reason I am proposing this is because CredSSP configuration is managed by Group Policy in my organization, and does not need to be managed by dbatools. If I run `Enable-WSManCredSSP` on my corporate workstation, it results in an access denied error.

I think this would be a reasonable way to allow users who aren't able to manage their own CredSSP settings a way to avoid unnecessary errors, or advanced users who *really* want to manage their own CredSSP settings to avoid configuration munging.

Create new configuration setting something like `commands.initialize-credssp.bypass` : 

```
Set-DbatoolsConfig -FullName 'commands.initialize-credssp.bypass' -Value $false -Initialize -Validation bool -Description "Do not attempt to configure CredSSP authentication, use existing configuration as-is"
```

Then add a conditional in `Initialize-CredSSP`:

```
if ( ( Get-DbatoolsConfigValue -FullName 'commands.initialize-credssp.bypass' ) -eq $true ) {
    Write-Message -Level Verbose -Message "CredSSP initialization bypass (commands.initialize-credssp.bypass) is set to $true";
    return;
}
```

1.0.139

kozloski:
* add configuration setting "commands.initialize-credssp.bypass"
* add check for configuration item value in Initialize-CredSSP function,
  log message and return if true

## Type of Change

 - [X] New feature (non-breaking change, adds functionality, fixes #7148 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Add a configuration item to bypass initialization of CredSSP configuration. This can be useful in cases where the user is not allowed to configure CredSSP on their computer or servers, or where a user already has CredSSP configured and does not want the existing configuration to be altered.

### Approach
Add a configuration item "commands.initialize-credssp.bypass" with a default value of $false to internal/configurations/settings/commands.ps1 

In internal/functions/Initialize-CredSSP.ps1, add a check for this configuration value.
If true, log a message indicating CredSSP initialization is bypassed, and return from the function with no further action.

### DIFF
--- a/functions/Install-DbaInstance.ps1
+++ b/functions/Install-DbaInstance.ps1
@@ -24,6 +24,10 @@ function Install-DbaInstance {
         Note that the dowloaded installation media must be extracted and available to the server where the installation runs.
         NOTE: If no ProductID (PID) is found in the configuration files/parameters, Evaluation version is going to be installed.
 
+        When using CredSSP authentication, this function will configure CredSSP authentication for PowerShell Remoting sessions.
+        If this is not desired (e.g.: CredSSP authentication is managed externally, or is already configured appropriately,)
+        it can be disabled by setting the dbatools configuration option 'commands.initialize-credssp.bypass' value to $true.
+
     .PARAMETER SqlInstance
         The target computer and, optionally, a new instance name and a port number.
         Use one of the following generic formats:
--- a/functions/Update-DbaInstance.ps1
+++ b/functions/Update-DbaInstance.ps1
@@ -25,6 +25,10 @@ function Update-DbaInstance {
         CVE-2018-0886 security update is required for both local and remote hosts. If CredSSP connections are failing, make sure to
         apply recent security updates prior to doing anything else.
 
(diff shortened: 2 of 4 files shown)


nvarscar:
See https://github.com/sqlcollaborative/dbatools/issues/7148#issuecomment-782357926

nvarscar (changes requested):
Please add information about this setting to the Description of the public functions (install/update-dbainstance), so that users are aware how to use it.

nvarscar (approved):
Looks good

potatoqualitee:
woo! thank you, @nvarscar! And thank you so much for the contribution, @kozloski

PR #7155, opened 2021-02-20
Update New-DbaDirectory.ps1

GLoeffler14:
Fix 2 bugs.

## Type of Change

 - [ XX] Bug fix (non-breaking change, fixes # )
 

### Purpose
1.)  the $exists variable did not properly reference the enumerated items inside the collection, so only single servers could be used.
2.) when checking for the folder already existing, the method FileExists needed to be added to make sure it checked to see if the folder existed, not just that an object was returned, which was always the case.> 

### Approach
<allows an array to be used to create folders on more than 1 server>

### DIFF
--- a/functions/New-DbaDirectory.ps1
+++ b/functions/New-DbaDirectory.ps1
@@ -76,9 +76,9 @@ function New-DbaDirectory {
 
         $Path = $Path.Replace("'", "''")
 
-        $exists = Test-DbaPath -SqlInstance $SqlInstance -SqlCredential $SqlCredential -Path $Path
+        $exists = Test-DbaPath -SqlInstance $Instance -SqlCredential $SqlCredential -Path $Path
 
-        if ($exists) {
+        if ($exists.FileExists) {
             Stop-Function -Message "$Path already exists" -Target $server -Continue
         }
 


wsmelton:
@GLoeffler14 this PR does not properly utilize our template. Please open a new PR and provide the requested details (do not delete content).

functions/New-DbaDirectory.ps1
```diff
@@ -76,9 +76,9 @@ function New-DbaDirectory {
 
         $Path = $Path.Replace("'", "''")
 
-        $exists = Test-DbaPath -SqlInstance $SqlInstance -SqlCredential $SqlCredential -Path $Path
+        $exists = Test-DbaPath -SqlInstance $Instance -SqlCredential $SqlCredential -Path $Path
```
wsmelton, line 79: This should be passing the `$server` object, internal calls to other public functions have the main object passed so it does not cause additional connections to be opened in SQL Server.

```
$exists = Test-DbaPath -SqlInstance $server -Path $Path
```

functions/New-DbaDirectory.ps1
```diff
@@ -76,9 +76,9 @@ function New-DbaDirectory {
...
 
-        if ($exists) {
+        if ($exists.FileExists) {
```
wsmelton, line 81: The purpose of the command is to create a directory and not a file. `$FileExists` will always be false if they provide a path with a file and not an explicit container.

It depends on what bug you are trying to fix as to what this needs to be changed to but overall the original check is fine because we do not output an object if the path provided does not exists.

GLoeffler14:
> @GLoeffler14 this PR does not properly utilize our template. Please open a new PR and provide the requested details (do not delete content).

sorry, first PR...  just created a bug request and will redo.  hopefully will get it right this time.

PR #7164, opened 2021-02-22
Export-DbaLogin -  ensure the ScriptingOptions object with IncludeIfNotExists set to true is used when scripting the db users

### LINKED ISSUE #7158: Existance Check Create User
Export-DBALogin currently does an "IF USER NOT EXISTS" check when creating logins to the instance, which is great.  However, in generating the TSQL file it does not put a similar check around the create user statement, which when coupled with a restore/refresh we run into errors at times.

My request is simple, can that check be added to the SQL file generation?

Thanks!

lancasteradam:
Export-DbaLogin - Ensure the ScriptingOptions object with IncludeIfNotExists set to true is used when scripting the db users

## Type of Change

 - [x] Bug fix (non-breaking change, fixes #7158 )
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Ensure the ScriptingOptions object with IncludeIfNotExists set to true is used when scripting the db users.

### Approach

Minor refactor of the code.

### Commands to test

The "Should Export a specific user" integration test has an example.

### DIFF
--- a/functions/Export-DbaLogin.ps1
+++ b/functions/Export-DbaLogin.ps1
@@ -461,16 +461,17 @@ function Export-DbaLogin {
 
                         $outsql += "`r`nUSE [$dbName]`r`n"
 
+                        $scriptOptions = New-DbaScriptingOption
+                        $scriptVersion = $sourceDb.CompatibilityLevel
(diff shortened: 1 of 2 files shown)


nvarscar (approved):
Makes sense, thanks for the submission!

potatoqualitee:
Thanks again to you both 🙇

PR #7181, opened 2021-03-06
New commands for interacting with sequence objects

### LINKED ISSUE #7180: Add commands to manage sequence objects
### Summary of new feature
Add commands to manage and interact with sequence objects .

Add the typical commands for CRUD operations on sequence objects as well as a command to obtain the next sequence value for usage in scripting.

1.0.139

lancasteradam:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality, fixes #7180 )
 - [x] Pester test is included
 - [ ] Nunit test is included
 - [x] Documentation
 - [ ] Build system
 

### Purpose
 
Add new commands for interacting with sequence objects. Due to a limitation in SMO I've added the restriction that the New-DbaDbSequence command will only support SQL 2012 and higher.

### Approach

Implemented the commands to manage sequences and included a command to get a sequence value in case that is needed in scripts/pipelines.

### Commands to test

See the new integration tests.

### DIFF
--- a/functions/Get-DbaDbSequence.ps1
+++ b/functions/Get-DbaDbSequence.ps1
@@ -41,7 +41,7 @@ function Get-DbaDbSequence {
         Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
 
     .NOTES
-        Tags: Data, Database, Migration, Sequence, Table
+        Tags: Data, Sequence, Table
         Author: Adam Lancaster https://github.com/lancasteradam
 
         dbatools PowerShell module (https://dbatools.io)
@@ -80,10 +80,14 @@ function Get-DbaDbSequence {
             return
         }
 
+        # caller specified the instance info
         foreach ($instance in $SqlInstance) {
-            $InputObject += Get-DbaDatabase -SqlInstance $instance -SqlCredential $SqlCredential -Database $Database
+            foreach ($db in (Get-DbaDatabase -SqlInstance $instance -SqlCredential $SqlCredential -Database $Database)) {
+                $db.Sequences | Where-Object { $_.Schema -eq $Schema -and $_.Name -eq $Name }
+            }
         }
 
+        # caller has piped in one or more databases
         foreach ($db in $InputObject) {
             $db.Sequences | Where-Object { $_.Schema -eq $Schema -and $_.Name -eq $Name }
         }
--- a/functions/Get-DbaDbSequence.ps1
+++ b/functions/Get-DbaDbSequence.ps1
@@ -1,0 +1,91 @@
+function Get-DbaDbSequence {
+    <#
+    .SYNOPSIS
+        Finds a sequence.
+
+    .DESCRIPTION
+        Finds a sequence in the database(s) specified.
+
+    .PARAMETER SqlInstance
+        The target SQL Server instance or instances. This can be a collection and receive pipeline input to allow the function
+        to be executed against multiple SQL Server instances.
+
+    .PARAMETER SqlCredential
+        Login to the target instance using alternative credentials. Accepts PowerShell credentials (Get-Credential).
+
+        Windows Authentication, SQL Server Authentication, Active Directory - Password, and Active Directory - Integrated are all supported.
+
+        For MFA support, please use Connect-DbaInstance.
+
+    .PARAMETER Database
+        The target database(s).
+
+    .PARAMETER Name
+        The name of the sequence.
+
+    .PARAMETER Schema
+        The name of the schema for the sequence. The default is dbo.
+
+    .PARAMETER InputObject
+        Allows piping from Get-DbaDatabase.
+
+    .PARAMETER WhatIf
+        Shows what would happen if the command were to run. No actions are actually performed.
+
+    .PARAMETER Confirm
+        Prompts you for confirmation before executing any changing operations within the command.
+
+    .PARAMETER EnableException
+        By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
+        This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
+        Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
+
+    .NOTES
+        Tags: Data, Database, Migration, Sequence, Table
+        Author: Adam Lancaster https://github.com/lancasteradam
+
+        dbatools PowerShell module (https://dbatools.io)
+        Copyright: (c) 2021 by dbatools, licensed under MIT
+        License: MIT https://opensource.org/licenses/MIT
+
+    .LINK
+        https://dbatools.io/Get-DbaDbSequence
+
+    .EXAMPLE
+        PS C:\> Get-DbaDbSequence -SqlInstance sqldev01 -Database TestDB -Name TestSequence
+
+        Finds the sequence TestSequence in the TestDB database on the sqldev01 instance.
+
+    .EXAMPLE
+        PS C:\> Get-DbaDatabase -SqlInstance sqldev01 -Database TestDB | Get-DbaDbSequence -Name TestSequence -Schema TestSchema
+
+        Using a pipeline this command finds the sequence named TestSchema.TestSequence in the TestDB database on the sqldev01 instance.
+    #>
+    [CmdletBinding(SupportsShouldProcess, ConfirmImpact = 'Low')]
+    param (
+        [DbaInstanceParameter[]]$SqlInstance,
+        [PSCredential]$SqlCredential,
+        [string[]]$Database,
+        [Parameter(Mandatory)]
+        [string]$Name,
+        [string]$Schema = 'dbo',
+        [parameter(ValueFromPipeline)]
+        [Microsoft.SqlServer.Management.Smo.Database[]]$InputObject,
+        [switch]$EnableException
+    )
+    process {
+
+        if ((Test-Bound -ParameterName SqlInstance) -and (Test-Bound -Not -ParameterName Database)) {
+            Stop-Function -Message "Database is required when SqlInstance is specified"
+            return
+        }
+
+        foreach ($instance in $SqlInstance) {
+            $InputObject += Get-DbaDatabase -SqlInstance $instance -SqlCredential $SqlCredential -Database $Database
+        }
+
+        foreach ($db in $InputObject) {
+            $db.Sequences | Where-Object { $_.Schema -eq $Schema -and $_.Name -eq $Name }
+        }
+    }
+}
--- a/functions/New-DbaDbSequence.ps1
+++ b/functions/New-DbaDbSequence.ps1
@@ -4,7 +4,7 @@ function New-DbaDbSequence {
         Creates a sequence.
 
     .DESCRIPTION
-        Creates a sequence in the database(s) specified.
+        Creates a sequence in the database(s) specified. SQL Server 2012 and higher are supported.
 
     .PARAMETER SqlInstance
         The target SQL Server instance or instances. This can be a collection and receive pipeline input to allow the function
@@ -118,6 +118,10 @@ function New-DbaDbSequence {
         }
 
         foreach ($db in $InputObject) {
+
+            if ($db.Parent.VersionMajor -lt 11) {
+                Stop-Function -Message "This command only supports SQL Server 2012 and higher." -Continue
+            }
 
             if (($db.Sequences | Where-Object { $_.Schema -eq $Schema -and $_.Name -eq $Name })) {
                 Stop-Function -Message "Sequence $Name already exists in the $Schema schema in the database $($db.Name) on $($db.Parent.Name)" -Continue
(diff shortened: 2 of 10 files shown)


functions/Get-DbaDbSequence.ps1
```diff
@@ -0,0 +1,91 @@
...
+    )
+    process {
+
+        if ((Test-Bound -ParameterName SqlInstance) -and (Test-Bound -Not -ParameterName Database)) {
+            Stop-Function -Message "Database is required when SqlInstance is specified"
+            return
+        }
+
+        foreach ($instance in $SqlInstance) {
+            $InputObject += Get-DbaDatabase -SqlInstance $instance -SqlCredential $SqlCredential -Database $Database
```
wsmelton, line 84: Please change this to output as the objects are found instead of building a collection.

functions/Get-DbaDbSequence.ps1
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        foreach ($db in $InputObject) {
+            $db.Sequences | Where-Object { $_.Schema -eq $Schema -and $_.Name -eq $Name }
```
wsmelton, line 88: If creating Sequences can only be done in SQL Server 2012 is getting them not restricted?

functions/New-DbaDbSequence.ps1
```diff
@@ -0,0 +1,190 @@
...
+        if ((Test-Bound -ParameterName SqlInstance) -and (Test-Bound -Not -ParameterName Database)) {
+            Stop-Function -Message "Database is required when SqlInstance is specified"
+            return
+        }
+
+        foreach ($instance in $SqlInstance) {
+            $InputObject += Get-DbaDatabase -SqlInstance $instance -SqlCredential $SqlCredential -Database $Database
+        }
+
+        foreach ($db in $InputObject) {
```
wsmelton, line 120: This command would be best to work per instance instead of per database. Unless you found this to be just as performant?

Consider if user provides 12 instances and each of those instances has 15 databases, how does this command handle that processing? That is a lot of object to cache and then iterate over versus caching 15 objects on each instead instead of 180 databases in one go.

functions/New-DbaDbSequence.ps1
```diff
@@ -0,0 +1,190 @@
...
+                        } else {
+                            $newSequence.SequenceCacheType = [Microsoft.SqlServer.Management.Smo.SequenceCacheType]::CacheWithSize
+                            $newSequence.CacheSize = $CacheSize
+                        }
+                    } else {
+                        $newSequence.SequenceCacheType = [Microsoft.SqlServer.Management.Smo.SequenceCacheType]::DefaultCache
+                    }
+
+                    $newSequence.Create()
+                    $newSequence
```
wsmelton, line 183: This should reach back to the instance (using the same connection as used to create it) and grab the sequence just created instead of depending on the cached object. 

Execution `Create()` will not always give good enough indicator that the object was indeed created.

functions/Select-DbaDbSequenceNextValue.ps1
```diff
@@ -0,0 +1,89 @@
...
+    .PARAMETER Confirm
+        Prompts you for confirmation before executing any changing operations within the command.
+
+    .PARAMETER EnableException
+        By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
+        This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
+        Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
+
+    .NOTES
+        Tags: Data, Database, Migration, Sequence, Table
```
wsmelton, line 44:
```suggestion
        Tags: Data, Sequence, Table
```

functions/Remove-DbaDbSequence.ps1
```diff
@@ -0,0 +1,99 @@
...
+    .PARAMETER Confirm
+        Prompts you for confirmation before executing any changing operations within the command.
+
+    .PARAMETER EnableException
+        By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
+        This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
+        Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
+
+    .NOTES
+        Tags: Data, Database, Migration, Sequence, Table
```
wsmelton, line 44:
```suggestion
        Tags: Data, Sequence, Table
```

functions/New-DbaDbSequence.ps1
```diff
@@ -0,0 +1,190 @@
...
+    .PARAMETER Confirm
+        Prompts you for confirmation before executing any changing operations within the command.
+
+    .PARAMETER EnableException
+        By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
+        This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
+        Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
+
+    .NOTES
+        Tags: Data, Database, Migration, Sequence, Table
```
wsmelton, line 65:
```suggestion
        Tags: Data, Sequence, Table
```

functions/Get-DbaDbSequence.ps1
```diff
@@ -0,0 +1,91 @@
...
+    .PARAMETER Confirm
+        Prompts you for confirmation before executing any changing operations within the command.
+
+    .PARAMETER EnableException
+        By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
+        This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
+        Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
+
+    .NOTES
+        Tags: Data, Database, Migration, Sequence, Table
```
wsmelton, line 44:
```suggestion
        Tags: Data, Sequence, Table
```

functions/Set-DbaDbSequence.ps1
```diff
@@ -0,0 +1,159 @@
...
+    .PARAMETER Confirm
+        Prompts you for confirmation before executing any changing operations within the command.
+
+    .PARAMETER EnableException
+        By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
+        This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
+        Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
+
+    .NOTES
+        Tags: Data, Database, Migration, Sequence, Table
```
wsmelton, line 62:
```suggestion
        Tags: Data, Sequence, Table
```

functions/Set-DbaDbSequence.ps1
```diff
@@ -0,0 +1,159 @@
...
+                            $sequence.SequenceCacheType = [Microsoft.SqlServer.Management.Smo.SequenceCacheType]::CacheWithSize
+                            $sequence.CacheSize = $CacheSize
+                        }
+                    } else {
+                        $sequence.SequenceCacheType = [Microsoft.SqlServer.Management.Smo.SequenceCacheType]::DefaultCache
+                    }
+
+                    $sequence.Alter()
+                    $sequence.Refresh()
+                    $sequence
```
wsmelton, line 152: This should get the sequence from the instance again and not depend on the cached object/connection.

functions/Set-DbaDbSequence.ps1
```diff
@@ -0,0 +1,159 @@
...
+                } catch {
+                    Stop-Function -Message "Failure on $($db.Parent.Name) to modify the sequence $Name in the $Schema schema in the database $($db.Name)" -ErrorRecord $_ -Continue
```
wsmelton, line 154: Should this also include the instance that the database resides on?

functions/Remove-DbaDbSequence.ps1
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        foreach ($db in $InputObject) {
+
+            if ($Pscmdlet.ShouldProcess($db.Parent.Name, "Removing the sequence $Name in the $Schema schema in the database $($db.Name) on $($db.Parent.Name)")) {
+                try {
+                    $sequenceToDrop = $db.Sequences | Where-Object { $_.Schema -eq $Schema -and $_.Name -eq $Name }
+                    $sequenceToDrop.Drop()
+                } catch {
+                    Stop-Function -Message "Failure on $($db.Parent.Name) to drop the sequence $Name in the $Schema schema in the database $($db.Name)" -ErrorRecord $_ -Continue
```
wsmelton, line 94: Should this also include the instance that the database resides on?

functions/New-DbaDbSequence.ps1
```diff
@@ -0,0 +1,190 @@
...
+                            $newSequence.CacheSize = $CacheSize
+                        }
+                    } else {
+                        $newSequence.SequenceCacheType = [Microsoft.SqlServer.Management.Smo.SequenceCacheType]::DefaultCache
+                    }
+
+                    $newSequence.Create()
+                    $newSequence
+                } catch {
+                    Stop-Function -Message "Failure on $($db.Parent.Name) to create the sequence $Name in the $Schema schema in the database $($db.Name)" -ErrorRecord $_ -Continue
```
wsmelton, line 185: Should this also include the instance that the database resides on?

functions/Select-DbaDbSequenceNextValue.ps1
```diff
@@ -0,0 +1,89 @@
+function Select-DbaDbSequenceNextValue {
```
wsmelton, line 1: I see no reason for this to be a command by itself. That content can be added to the object for `Get-DbaSequence` and simply output the additional property as a default output.

functions/Set-DbaDbSequence.ps1
```diff
@@ -0,0 +1,159 @@
...
+
+        foreach ($instance in $SqlInstance) {
+            $InputObject += Get-DbaDatabase -SqlInstance $instance -SqlCredential $SqlCredential -Database $Database
+        }
+
+        foreach ($db in $InputObject) {
+
+            if ($Pscmdlet.ShouldProcess($db.Parent.Name, "Modifying the sequence $Name in the $Schema schema in the database $($db.Name) on $($db.Parent.Name)")) {
+                try {
+                    $sequence = $db | Get-DbaDbSequence -Schema $Schema -Name $Name
```
wsmelton, line 115: Best to use the `-Database` parameter of the command for performance.

wsmelton (changes requested):

functions/Get-DbaDbSequence.ps1
lancasteradam: yes, will do

functions/Get-DbaDbSequence.ps1
lancasteradam: Yes, the New-DbaDbSequence checks for SQL 2012 or higher. The other commands are not restricted and could be used for the older unsupported versions of SQL.

functions/New-DbaDbSequence.ps1
lancasteradam: I'm glad you mentioned this. Yes, in this case it does perform well. Creating a sequence in 300 databases took about 25 seconds on my dev machine and there was no abnormal memory usage.

I see what you mean about processing the batch per instance and I agree that is better. However, this pattern of doing the "$InputObject +=" seems to be the common approach for supporting both the pipeline input and the normal command invocation with parameters.

functions/New-DbaDbSequence.ps1
lancasteradam: ok, will do

functions/Set-DbaDbSequence.ps1
lancasteradam: ok, will do

functions/Set-DbaDbSequence.ps1
lancasteradam: Yes, it has the $($db.Parent.Name) at the beginning of the message string.

functions/Select-DbaDbSequenceNextValue.ps1
lancasteradam: Would there be a new param for Get-DbaDbSequence such as -SelectNextValue so that the caller could advance the sequence and obtain the next value? I would be concerned about the Get-DbaDbSequence function changing state on the server (i.e. advancing the sequence). The goal for this Select command is to provide a scripting command to call "select next value for <sequence>".

functions/Set-DbaDbSequence.ps1
lancasteradam: Unfortunately that doesn't seem possible. If the caller had originally piped in the db object we won't have the SQLInstance and SQLCredential info that would be required to use the -Database parameter.

lancasteradam:
Thanks @wsmelton  for the feedback, much appreciated.

potatoqualitee:
Looks fab! Thank you 🙇🏼  I'll add this to the psm1 and the psd1 and publish today.

PR #7207, opened 2021-03-22
Fix misnamed variable in Backup-DbaDatabase?

LordFlashmeow:
If you try to perform a backup on multiple databases with the `-ReplaceInName` and the `-FilePath` flag, you get hit with the error "1 BackupFile specified, but more than 1 database." 

According to the examples, this should be possible: `Backup-DbaDatabase -SqlInstance Server1\Prod -Path \\filestore\backups\servername\instancename\dbname\backuptype -FilePath dbname-backuptype-timestamp.trn -Type Log -ReplaceInName`

I think this is because there's a misnamed variable that doesn't exist anywhere else. When I set `$ReplaceInFile = $true` it works, and when it's false, the command fails and raises the error message. 

I'm pretty sure that `$ReplaceInFile` is meant to be `$ReplaceInName`, but clarification otherwise would also be appreciated.

## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Fix a bug that prevents using custom file names for backups with multiple databases

### Approach

Change a misnamed variable that isn't set or documented anywhere

### Commands to test

Before using the fix, verify that it's broken:

```
PS C:\> Backup-DCbaDatabase -SqlInstance Server1\Prod -Path \\filestore\backups\servername\instancename\dbname\backuptype -FilePath dbname-backuptype-timestamp.trn -Type Log -ReplaceInName
Warning: 1 BackupFile specified, but more than 1 database.

PS C:\>$ReplaceInFile = $true
PS C:\>Backup-DbaDatabase -SqlInstance Server1\Prod -Path \\filestore\backups\servername\instancename\dbname\backuptype -FilePath dbname-backuptype-timestamp.trn -Type Log -ReplaceInName

Success...
```


LordFlashmeow:
Realized that this doesn't really fix the underlying issue. The names of the subsequent database files are still the defaults, not the ones set in the command.

PR #7236, opened 2021-04-07
Update Install-DbaMaintenanceSolution.ps1

Ant-Green:
Added Queue and QueueDatabase to the list of required scripts for anyone who wants to use the DatabasesInParallel options of the jobs so they don't have to go and find them separately.

QueueDatabase creates an FK to Queue so had to add a Sort-Order on the for-each to ensure the right order of deployment.

Added cleanup task for the two tables for $ReplaceExisting

Changed existence check for the Queue & QueueDatabase to Stop should they exist and not have $ReplaceExisting enabled

## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
Complete installation of the Ola tool set with the two tables to allow parallel runs

### Approach

Added the two scripts to the right places in the code and tested they deploy as needed

### Commands to test

Install-DbaMaintenanceSolution

select name, create_date from sys.objects 
where name in ('QueueDatabase','Queue')
order by create_date desc

### Screenshots

image


niphlod:
sorry to barge in, see https://github.com/sqlcollaborative/dbatools/pull/6704. The TLDR is "in order to not break backward compat, do not assume everyone wants Queue and QueueDatabase created"

Ant-Green:
So that was essentially to stop any issues when anyone was doing a "LocalFile" installation then who has a very old copy of the maintenance solution.

@DatabasesInParallel has been an option now for many months, without the Queue & QueueDatabase tables being present it breaks all 3 of the procs for anyone who uses that option, resulting in a manual fix to get the two scripts and create them, all in for automation etc we did our first automated build and the jobs all failed as Queue and QueueDatabase wasn't created.

niphlod:
and, still, people can choose to not use @DatabaseInParallel, so why force them ? 
I'm using Queue and QueueDatabase too, but the rationale behind Install-DbaMaintenanceSolution hasn't changed.

Ant-Green:
So the rational is as @DatabasesInParallel is not an option like @LogToTable on the main scripts, that anyone who is manually modifying the jobs to use parallel should manually go and create the Queue and QueueDatabase objects

niphlod:
not really. It's just "make Queue and QueueDatabase creation optional, as is CommandLog". 
re-quote from the latest "agreed-upon" post:

> Soooooo, IMHO, we should have done:
> 
> InstallJobs is passed ? If yes
> Solution All --> executes MaintenanceScript, Queue and QueueDatabase. If LogToTable is passed, triggers the creation of CommandLog (because at this point is a strict requirement)
> every other choice of solution ends up with installing only the specified procedure, but jobs are created nonetheless for everything so you may end up with jobs configured calling procedures you don't have
> InstallJobs is not passed
> Solution All --> executes Queue, QueueDatabase, CommandExecute, CommandLog and the three procedures
> every other choice ends up with just CommandExecute and the selected procedure 'cause you seem to want just the basics, with the exception of CommandLog created when -LogToTable is passed
> In this way, LogToTable triggers creation only if "configuration" is agreed upon, and becomes a "please install another bit on top of the three procedures", keeping the "object creator" role in all cases.
> You get CommandLog created with InstallJobs if you pass also LogToTable, or when you choose to install just one procedure and still pass LogToTable.
> But, different from now, you don't get it if you don't install jobs and choose All, without passing -LogToTable.
> 
> At that point the pain point of LogToTable being both "config" and "object creator" is removed.
> 
> Enter "but since some time Ola's support processing databases in parallel".
> If we want to keep "Solution" tied to "all 3 or one of the three procedures" vs "all objects that may be used by the solution" then we can add a InstallParallel that triggers Queue and QueueDatabase, and make them behave as LogToTable, strictly with an "object creator" role.
> To make sure the entire matrix of parameters of Ola's is supported, users will need to pass -Solution All -LogToTable -InstallParallel (-Solution being a choice of all three procedures, -LogToTable triggering CommandLog and InstallParallel triggering Queue and QueueDatabase).

Ant-Green:
Believe all the changes niphlod mentioned above have now been captured

functions/Install-DbaMaintenanceSolution.ps1
```diff
@@ -139,7 +142,8 @@ function Install-DbaMaintenanceSolution {
         [switch]$InstallJobs,
         [string]$LocalFile,
         [switch]$Force,
-        [switch]$EnableException
+        [switch]$EnableException,
+        [switch]$InstallParallel
```
niphlod, line 146: keep EnableException as the last one, sorry, forgot to mention it in the chat. Everything else looks good.

niphlod (changes requested):

niphlod:
failure on appveyor unrelated, I'll restart it but everything for this PR is good. @wsmelton , @potatoqualitee  ? what do you think ?

functions/Install-DbaMaintenanceSolution.ps1
```diff
@@ -53,6 +53,9 @@ function Install-DbaMaintenanceSolution {
     .PARAMETER Confirm
         If this switch is enabled, you will be prompted for confirmation before executing any operations that change state.
 
+    .PARAMETER InstallParallel
+        If this switch is enabled, the Queue and QueueDatabase tables are created.
```
jpomfret, line 57: Could you add a note that these tables are used by the DatabasesInParallel option?

functions/Install-DbaMaintenanceSolution.ps1
```diff
@@ -139,7 +142,9 @@ function Install-DbaMaintenanceSolution {
         [switch]$InstallJobs,
         [string]$LocalFile,
         [switch]$Force,
+        [switch]$InstallParallel,
```
jpomfret, line 145: Could you also add an example to the help for this new param please

functions/Install-DbaMaintenanceSolution.ps1
```diff
@@ -315,7 +320,7 @@ function Install-DbaMaintenanceSolution {
                 $required = @('CommandExecute.sql')
             }
 
-            if ($LogToTable) {
+            if ($LogToTable -and $InstallJobs -eq $false) {
```
jpomfret, line 323: Also - not following this add, wouldn't we want the CommandLog table to be created whether jobs are installed or not? (or did I miss that it's added in later if InstallJobs is true?

jpomfret (commented):
Thanks for the work @Ant-Green and thanks for the review @niphlod.

Agree it's looking good - couple of ideas and a question.
Thanks!

functions/Install-DbaMaintenanceSolution.ps1
Ant-Green: That has been done, waiting to push/commit for any further tweaks.

functions/Install-DbaMaintenanceSolution.ps1
Ant-Green: Example added, waiting to push/commit for further tweaks

`
    .EXAMPLE
        PS C:\> Install-DbaMaintenanceSolution -SqlInstance RES14224 -Database DBA -InstallParallel

        This will create the Queue and QueueDatabase tables for uses when manually changing jobs to use the @DatabasesInParallel = 'Y' flag`

functions/Install-DbaMaintenanceSolution.ps1
Ant-Green: If -InstallJobs:$true then the only script which can be run is the "MaintenanceSolution.sql" script which will force install the commandlog table anyway.

The add was done to make it the decision between "Install" and "Config" and to while the maintenancesolution will skip over the object already being present with the existence check, to stop unnecessary scripts from being run.

I walked through the logic with @niphlod over Slack and advised this is the route from #6704 that was envisaged

jpomfret: Cool gotcha - thanks for clarrifying

jpomfret (approved):
Thanks!

PR #7246, opened 2021-04-18
Remove allcommands.ps1 (will not impact Gallery)

potatoqualitee:
This will help speed up development within VS Code. But I'll still include allcommands.ps1 when published to Gallery.

People installing right from GitHub - their import will slow down by about 5 seconds but overall, each of our devs will benefit. This has long been requested by Shawn and I agree.


jpomfret (approved):
makes sense to me

potatoqualitee:
sweet, thanks! I tried it from the gallery and then local and both still work, as expecting. should make for a better dev experience for us 👍🏼

PR #7293, opened 2021-05-07
Get-DbaServerRoleMember - Use PSCustomObject to prevent modification of SMO

### LINKED ISSUE #7214: [Bug] Get-DbaServerRoleMember - Update of $login.Role not correct
This code:

```
            foreach ($role in $roles) {
                $members = $role.EnumMemberNames()
                foreach ($member in $members) {
                    $l = $server.Logins | Where-Object { $_.Name -eq $member }
                    if ($l) {
                        Add-Member -Force -InputObject $l -MemberType NoteProperty -Name Role -Value $role.Name
                    }
                }
            }
```

will set the property only to the last role in the loop. So the information is not correct if the login is member of more than one role.

Is it really necessary to change the login object here? I discovered the function by chance and actually find it quite nice, but it should then better be integrated directly into Get-DbaLogin and contain an array of the roles.

Get-DbaDbRoleMember has the same problem.

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #7214 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
prevent modification of login SMO

### Approach
Output a PSCustomObject

### DIFF
--- a/functions/Get-DbaServerRoleMember.ps1
+++ b/functions/Get-DbaServerRoleMember.ps1
@@ -150,13 +150,13 @@ function Get-DbaServerRoleMember {
                     $l = $server.Logins | Where-Object { $_.Name -eq $member }
 
                     if ($l) {
-                        Add-Member -Force -InputObject $l -MemberType NoteProperty -Name ComputerName -Value $server.ComputerName
-                        Add-Member -Force -InputObject $l -MemberType NoteProperty -Name InstanceName -Value $server.ServiceName
-                        Add-Member -Force -InputObject $l -MemberType NoteProperty -Name SqlInstance -Value $server.DomainInstanceName
-                        Add-Member -Force -InputObject $l -MemberType NoteProperty -Name Role -Value $role.Name
-
-                        # Select object because Select-DefaultView causes strange behaviors when assigned to a variable (??)
-                        Select-Object -InputObject $l -Property 'ComputerName', 'InstanceName', 'SqlInstance', 'Role', 'Name'
+                        [PSCustomObject]@{
+                            ComputerName = $server.ComputerName
+                            InstanceName = $server.ServiceName
+                            SqlInstance  = $server.DomainInstanceName
+                            Role         = $role.Name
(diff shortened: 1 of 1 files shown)


potatoqualitee:
😬 not a fan of this solution, i'm so sorry. I used to default to pscustomobject but now I like to rely on SMO. serverrolemembers are indeed logins, just like in SSMS. is there any other way to better address your concerns?

andreasjordan:
I have commented in the issue.

andreasjordan:
Appveyor errors not related to change, can we run them again?

jpomfret:
Rerunning tests...

andreasjordan:
Thanks Jess, all passed now.

functions/Get-DbaServerRoleMember.ps1
```diff
@@ -99,7 +99,7 @@ function Get-DbaServerRoleMember {
             Write-Message -Level Verbose -Message "Attempting to connect to $instance"
 
             try {
-                $server = Connect-SqlInstance -SqlInstance $instance -SqlCredential $SqlCredential
+                $server = Connect-DbaInstance -SqlInstance $instance -SqlCredential $SqlCredential
```
wsmelton, line 102: Unless we are going to do this across all the commands, this should stay with internal function.

andreasjordan:
My very personal opinion: We should change Connect-SqlInstance to Connect-DbaInstance in all commands because Connect-SqlInstance has no additional use. And I think we discussed this already, but can find it anymore...

wsmelton:
Raise an issue and do it one PR, piecemeal is not the way to implement a standard.

functions/Get-DbaServerRoleMember.ps1
potatoqualitee: I will merge then set it back to be included in the mass change if we make it. Thanks for the review @jpomfret and @wsmelton !

PR #7294, opened 2021-05-07
Mount-DbaDatabase - Removed Code that will never be used

andreasjordan:
Not a bug, but I don't like unused code...

All existing databases will be skipped (see a few lines above), so no need to test for existing system databases.


potatoqualitee:
Sweet! Thank you both 🙇🏼

PR #7306, opened 2021-05-10
New commands: New-DbaDbFileGroup, Set-DbaDbFileGroup, and Remove-DbaDbFileGroup

lancasteradam:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [x] Documentation
 - [ ] Build system
 
### Purpose
- New commands for creating, updating, and removing filegroups.
- Made a minor fix for Disable-DbaFilestream to change RunningValue to [RunValue ](https://docs.microsoft.com/en-us/dotnet/api/microsoft.sqlserver.management.smo.configuration.filestreamaccesslevel)

### Approach
Basic commands for filegroup create, update, remove. All feedback is welcome.

### Commands to test
See the command examples and the integration tests.


potatoqualitee:
Thank you, Adam! Will hopefully take a look this weekend.

functions/Remove-DbaDbFileGroup.ps1
```diff
@@ -0,0 +1,106 @@
...
+        PS C:\>Get-DbaDatabase -SqlInstance sqldev1 -Database TestDb | Remove-DbaDbFileGroup -FileGroupName HRFG1
+
+        Passes in the TestDB database via pipeline and removes the HRFG1 filegroup on the TestDb database on the sqldev1 instance.
+    #>
+    [CmdletBinding(SupportsShouldProcess, ConfirmImpact = 'High')]
+    param (
+        [DbaInstanceParameter[]]$SqlInstance,
+        [PSCredential]$SqlCredential,
+        [string[]]$Database,
+        [string]$FileGroupName,
```
jpomfret, line 66: Wondering if FileGroup could accept multiple filegroups to be deleted at once?

functions/New-DbaDbFileGroup.ps1
```diff
@@ -0,0 +1,124 @@
...
+        Login to the target instance using alternative credentials. Accepts PowerShell credentials (Get-Credential).
+
+        Windows Authentication, SQL Server Authentication, Active Directory - Password, and Active Directory - Integrated are all supported.
+
+        For MFA support, please use Connect-DbaInstance.
+
+    .PARAMETER Database
+        The target database(s).
+
+    .PARAMETER FileGroupName
```
jpomfret, line 23: Could you rename this to just FileGroup for all three functions please? We tend to drop the `name`.

jpomfret (commented):
This is great work - few ideas\comments for discussion. Thanks for your contributions!

lancasteradam:
Thanks @jpomfret, much appreciated. I'll look into those items.

lancasteradam:
@jpomfret, thanks again for the feedback. I've incorporated all of the suggestions:

1. Renamed the param for Filegroup.
2. Added support for multiple filegroups names to be passed in for the Set and Remove commands.
3. Added pipeline support for filegroup objects.
4. Added support to set a filegroup as read/write.
5. Added the option to turn off the auto grow of all files.

functions/Set-DbaDbFileGroup.ps1
```diff
@@ -0,0 +1,181 @@
...
+    #>
+    [CmdletBinding(SupportsShouldProcess, ConfirmImpact = 'High')]
+    param (
+        [DbaInstanceParameter[]]$SqlInstance,
+        [PSCredential]$SqlCredential,
+        [string[]]$Database,
+        [string[]]$FileGroup,
+        [switch]$Default,
+        [switch]$ReadOnly,
+        [switch]$ReadWrite,
```
jpomfret, line 104: I think it might be better to handle these as a single param instead of two switches... this currently doesn't fail and seems to set it to ReadWrite.
`Set-DbaDbFileGroup -SqlInstance mssql1 -Database DatabaseAdmin -FileGroup test2New -ReadWrite -ReadOnly`

You could leave it as a switch `$ReadOnly` and people can use `$ReadOnly:$False` to it back to ReadWrite.. or perhaps make it a string that accepts Enabled or Disabled as values?

Tagging @potatoqualitee and @wsmelton - do you have preferences or do we have a standard we should follow here?

functions/Set-DbaDbFileGroup.ps1
```diff
@@ -0,0 +1,181 @@
...
+        [switch]$AutoGrowAllFiles,
+        [switch]$AutoGrowSingleFile,
```
jpomfret, line 106: Same for these two, perhaps `$AutoGrowAllFiles` with Enabled\Disabled options?

wsmelton:
This needs to be closed and created as PR for each command.

wsmelton:
It is easier to review single command than trying to review multiple into one. Believe this has been mentioned before but may be wrong.

jpomfret:
Apologies on the close\reopen - miss press on the app

potatoqualitee:
These three related commands seem reasonable to me as one PR. I'd prefer making it more convenient for contributors when possible, and allow multi files PRs unless it doesn't make sense for an unforeseen reason 🙇

jpomfret:
Makes sense to me @potatoqualitee - thanks.

@lancasteradam - as for the param decisions... I think I'm leaning towards the best option being to make them both switches that when used set the property, and to unset it `$ReadOnly:$false` ... what do you think? Feels like that fits with SSMS having checkboxes?

image

lancasteradam:
@jpomfret, I agree, that sounds good. I'll make those changes.

jpomfret (approved):
looks good - thanks!

jpomfret:
unrelated test failures

PR #7313, opened 2021-05-12
New functions - synonyms

MikeyBronowski:
## Type of Change

 - [ ] Bug fix (non-breaking change, fixes # )
 - [x] New feature (non-breaking change, adds functionality, fixes # )
 - [ ] Breaking change (effects multiple commands or functionality, fixes # )
 - [x] Ran manual Pester test and has passed (`.\tests\manual.pester.ps1)
 - [ ] Adding code coverage to existing functionality
 - [x] Pester test is included
 - [ ] If new file reference added for test, has is been added to github.com/sqlcollaborative/appveyor-lab ?
 - [ ] Nunit test is included
 - [x] Documentation
 - [ ] Build system
 

### Purpose
 
Adding functions to manage database synonyms

### Approach

As a base, I used functions for DbRoles.

### Commands to test

Get-DbaDbSynonym
New-DbaDbSynonym
Remove-DbaDbSynonym

Issue #6885, opened 2020-09-27
[Bug] - Incorrect usage of InputObject


andreasjordan:
As this is the case in a lot of commands, I will use a general example:

```
function Test-DbaInputObject {
    [CmdletBinding(SupportsShouldProcess, ConfirmImpact = 'Low')]
    param (
        [string[]]$SqlInstance,
        [PSCredential]$SqlCredential,
        [parameter(ValueFromPipeline)]
        [string[]]$InputObject,
        [switch]$EnableException
    )
    process {
        Write-Verbose -Message "starting process"

        if (($null -eq $SqlInstance) -and ($null -eq $InputObject)) {
            Write-Warning -Message "You must supply either -SqlInstance or an Input Object"
            return
        }
        if ($null -ne $SqlInstance) {
            $InputObject += $SqlInstance
        }
        Write-Verbose -Message "InputObject.Count = '$($InputObject.Count)'"

        foreach ($instance in $InputObject) {
            Write-Verbose -Message "instance = '$instance'"
        }
    }
}
```

The basic idea is to add the values from the named parameter to the objects that come down the pipeline.

And for some scenarios, that works:

```
Test-DbaInputObject -SqlInstance a -Verbose
'a' | Test-DbaInputObject -Verbose
```
![image](https://user-images.githubusercontent.com/66946165/94370537-eccde680-00f0-11eb-882d-f86a98d3d954.png)

```
Test-DbaInputObject -SqlInstance a, b -Verbose
'a', 'b' | Test-DbaInputObject -Verbose
```
![image](https://user-images.githubusercontent.com/66946165/94370548-040cd400-00f1-11eb-9969-e66cd809787e.png)

```
'a' | Test-DbaInputObject -SqlInstance b, c, d -Verbose
```
![image](https://user-images.githubusercontent.com/66946165/94370589-3e767100-00f1-11eb-93c1-ae485f80adc8.png)

But when more than one object is piped and the named parameter is used as well, the named parameter is processed more than once:

```
'a', 'b' | Test-DbaInputObject -SqlInstance c, d -Verbose
```
![image](https://user-images.githubusercontent.com/66946165/94370665-8d240b00-00f1-11eb-9948-55aaf3067451.png)

andreasjordan:
I thinks that would be a correct way to handle this:

```
function Test-DbaInputObject {
    [CmdletBinding(SupportsShouldProcess, ConfirmImpact = 'Low')]
    param (
[...]

PR #7320, opened 2021-05-16
New function : Remove-DbaDbView

MikeyBronowski:
## Type of Change

 - [x] Pester test is included
 - [ ] Nunit test is included
 - [x] Documentation
 - [ ] Build system
 

### Purpose
 
Expanding functionality to manage database views.

### Commands to test

```powershell
Get-DbaDbView -SqlInstance localhost -Database Mikey -View Ocean | Remove-DbaDbView

or


Remove-DbaDbView -SqlInstance localhost -Database Mikey -View Ocean -Confirm:$false
```


functions/Remove-DbaDbView.ps1
```diff
@@ -0,0 +1,150 @@
...
+            return
+        }
+
+        if ($SqlInstance) {
+            $InputObject = $SqlInstance
+        }
+
+        foreach ($input in $InputObject) {
+            $inputType = $input.GetType().FullName
+            switch ($inputType) {
```
wsmelton, line 109: Use of this pattern is not required in every command. I know we had a few that started using it from contributors but is not part of usual template of commands.

Use of this block also convolutes it more than it needs to be. PowerShell inherits the ability to bind parameter values based on both property name or value. It goes through this process regardless of what you have set on the parameter attributes.

In the end this command only needs to accept input from `Get-DbaDbView` and not a database, server or the DbaInstance parameter itself.

functions/Remove-DbaDbView.ps1
```diff
@@ -0,0 +1,150 @@
...
+    param (
+        [parameter(ValueFromPipeline)]
+        [DbaInstance[]]$SqlInstance,
+        [PSCredential]$SqlCredential,
+        [string[]]$Database,
+        [string[]]$ExcludeDatabase,
+        [string[]]$View,
+        [switch]$IncludeSystemDbs,
+        [parameter(ValueFromPipeline)]
+        [object[]]$InputObject,
```
wsmelton, line 93:
```suggestion
        [Microsoft.SqlServer.Management.Smo.View[]]$InputObject,
```

This resolves any need to having to check what type is the input object. The object we output from `Get-DbaDbView` includes SqlInstance name and PowerShell will bind that for us without any hassle or extra code.

https://github.com/sqlcollaborative/dbatools/blob/de024aaa616fe37de7c8eedb796a2352ab08529a/functions/Get-DbaDbView.ps1#L167-L175

MikeyBronowski: Would that work just remove- standalone, without piping from get-

I would probably use get | remove myself as "safer" option anyway.

wsmelton: If you are calling remove function by itself you don't use the input object parameter.

functions/Remove-DbaDbView.ps1
MikeyBronowski: Makes sense. Do you know any "model" function(s) that follow the best practices and desired coding approach I could have a look at?

functions/Remove-DbaDbView.ps1
MikeyBronowski: Do my changes make more sense now?

functions/Remove-DbaDbView.ps1
```diff
@@ -0,0 +1,118 @@
...
+        if ((Test-Bound -ParameterName SqlInstance) -and (Test-Bound -Not -ParameterName Database)) {
+            Stop-Function -Message "Database is required when SqlInstance is specified"
+            return
+        }
+
+        foreach ($instance in $SqlInstance) {
+            $InputObject += Get-DbaDbView -SqlInstance $instance -SqlCredential $SqlCredential -Database $Database -View $View -ExcludeSystemView -ExcludeDatabase $ExcludeDatabase
+        }
+
+        foreach ($db in $InputObject) {
```
andreasjordan, line 106: $InputObject is not a list of databases, but a list of views. So I would call the variable $view.

MikeyBronowski: Good catch

functions/Remove-DbaDbView.ps1
```diff
@@ -0,0 +1,118 @@
...
+    )
+
+    process {
+
+        if ((Test-Bound -ParameterName SqlInstance) -and (Test-Bound -Not -ParameterName Database)) {
+            Stop-Function -Message "Database is required when SqlInstance is specified"
+            return
+        }
+
+        foreach ($instance in $SqlInstance) {
```
andreasjordan, line 102: I don't see any need in this foreach loop, as Get-DbaDbView can get the array in $SqlInstance as -SqlInstance. You could just test with `if ($SqlInstance) { ... }`. I know that this is widely used, but maybe we can streamline the code here and in other functions.

functions/Remove-DbaDbView.ps1
```diff
@@ -0,0 +1,118 @@
...
+            $InputObject += Get-DbaDbView -SqlInstance $instance -SqlCredential $SqlCredential -Database $Database -View $View -ExcludeSystemView -ExcludeDatabase $ExcludeDatabase
+        }
+
+        foreach ($db in $InputObject) {
+
+            if ($Pscmdlet.ShouldProcess($db.Parent.Name, "Removing the view $db in the database $($db.Parent.Name) on $instance")) {
+                try {
+                    $viewToDrop = $db.Parent.Views | Where-Object { $_.Name -eq $db.Name }
```
andreasjordan, line 110: Why don't you just use `$db.Drop()`? Testing for just the Name without Schema would maybe return and drop more views (if there are views with this name in other schemas).

andreasjordan:
And have a look at #6885.

functions/Remove-DbaDbView.ps1
```diff
@@ -0,0 +1,118 @@
...
+            return
+        }
+
+        if ($SqlInstance) {
+            $InputObject = Get-DbaDbView -SqlInstance $SqlInstance -SqlCredential $SqlCredential -Database $Database -View $View -ExcludeSystemView -ExcludeDatabase $ExcludeDatabase
+        }
+
+        foreach ($vw in $InputObject) {
+
+            if ($Pscmdlet.ShouldProcess($vw.SqlInstance, "Removing the view $vw in the database $($vw.Parent.Name)")) {
```
andreasjordan, line 108: As more than one SqlInstance is allowed, maybe also add the instance (.Parent.Parent.Name) in the message.

functions/Remove-DbaDbView.ps1
MikeyBronowski: > And have a look at #6885.

Thank you - I will have a look. Sounds like a great discussion and examples. Just need a fresh head and coffee :)

MikeyBronowski:
> And have a look at #6885.

Thank you - I will have a look. Sounds like a great discussion and examples. Just need a fresh head and coffee :)

functions/Remove-DbaDbView.ps1
```diff
@@ -0,0 +1,118 @@
...
+        [string[]]$View,
+        [parameter(ValueFromPipeline)]
+        [Microsoft.SqlServer.Management.Smo.View[]]$InputObject,
+        [switch]$EnableException
+    )
+
+    process {
+
+        if ((Test-Bound -ParameterName SqlInstance) -and (Test-Bound -Not -ParameterName Database)) {
+            Stop-Function -Message "Database is required when SqlInstance is specified"
```
andreasjordan, line 98: But the parameter documentation says: "The database(s) to process. This list is auto-populated from the server. If unspecified, all databases will be processed." What do you want?

functions/Remove-DbaDbView.ps1
```diff
@@ -0,0 +1,118 @@
+function Remove-DbaDbView {
+    <#
+    .SYNOPSIS
+        Removes a database view from database(s) for each instance(s) of SQL Server.
+
+    .DESCRIPTION
+        The Remove-DbaDbView removes view(s) from database(s) for each instance(s) of SQL Server.
+
+    .PARAMETER SqlInstance
+        The target SQL Server instance or instances. This can be a collection and receive pipeline input to allow the function to be executed against multiple SQL Server instances.
```
andreasjordan, line 10: No, no pipeline input for SqlInstance.

functions/Remove-DbaDbView.ps1
```diff
@@ -0,0 +1,118 @@
...
+    .PARAMETER Database
+        The database(s) to process. This list is auto-populated from the server. If unspecified, all databases will be processed.
+
+    .PARAMETER ExcludeDatabase
+        The database(s) to exclude. This list is auto-populated from the server.
+
+    .PARAMETER View
+        The view(s) to process. If unspecified, all views will be processed.
+
+    .PARAMETER IncludeSystemDbs
```
andreasjordan, line 28: You don't use this parameter at the moment.

functions/Remove-DbaDbView.ps1
MikeyBronowski: > As more than one SqlInstance is allowed, maybe also add the instance (.Parent.Parent.Name) in the message.

That's currently my output when I run :
```powershell
Get-dbadbview -SqlInstance $s1,$s2 -ExcludeSystemView -Database Mikey | Remove-DbaDbView -Verbose -Confirm:$false
```

```
VERBOSE: Performing the operation "Removing the view [dbo].[a] in the database Mikey" on target "sql1".
VERBOSE: Performing the operation "Removing the view [dbo].[a] in the database Mikey" on target "sql2".
```

functions/Remove-DbaDbView.ps1
```diff
@@ -0,0 +1,118 @@
...
+        If this switch is enabled, views can be removed from system databases.
+
+    .PARAMETER InputObject
+        Enables piped input from Get-DbaDbView or Get-DbaDatabase
```
andreasjordan, line 32: Only Get-DbaDbView, no more Get-DbaDatabase

functions/Remove-DbaDbView.ps1
```diff
@@ -0,0 +1,118 @@
...
+
+        Website: https://dbatools.io
+        Copyright: (c) 2021 by dbatools, licensed under MIT
+        License: MIT https://opensource.org/licenses/MIT
+
+    .LINK
+        https://dbatools.io/Remove-DbaDbView
+
+    .EXAMPLE
+        PS C:\> Remove-DbaDbView -SqlInstance localhost -Database dbname -View "customview1", "customview2"
```
andreasjordan, line 57: I would remove the " here.

functions/Remove-DbaDbView.ps1
```diff
@@ -0,0 +1,118 @@
...
+        PS C:\> Remove-DbaDbView -SqlInstance localhost, sql2016 -Database db1, db2 -View view1, view2, view3
+
+        Removes view1, view2, view3 from db1 and db2 on the local and sql2016 SQL Server instances.
+
+    .EXAMPLE
+        PS C:\> Remove-DbaDbView -SqlInstance localhost -Database msdb -IncludeSystemDbs -View view1, view2, view3
+
+        Removes view1, view2, view3 from db1 and db2 from system database (msdb) on the local SQL Server instances.
+
+    .EXAMPLE
```
andreasjordan, line 71: You only allow views SMOs to be piped in...

MikeyBronowski: I still can remove views like that
`Remove-DbaDbView -SqlInstance $s1 -Database Mikey -View b,c -Confirm:$false`

functions/Remove-DbaDbView.ps1
MikeyBronowski: Yeah, the same about "Schema" was there originally, but the Get-DbaDbView does not have it either. Wanted to finish the Remove- first so it is satisfactory and then add more options to the Get-DbaDbView as well.

wsmelton:
Overall I guess we need to decide, here since there is no issue for this:

`Remove-DbaDbSchema` got approved with iterating and taking input object of a Database (not the schema) https://github.com/sqlcollaborative/dbatools/blob/development/functions/Remove-DbaDbSchema.ps1

Whereas for the master key removal it only processes inputs from the Get master key command: https://github.com/sqlcollaborative/dbatools/blob/development/functions/Remove-DbaDbSchema.ps1

If we consider how would we do this for say tables or even stored procedures (if we chose to create one for that)
We do not have a whole lot of Remove functions yet for the various objects in a databases so now is the time to decide. // @potatoqualitee  @jpomfret

functions/Remove-DbaDbView.ps1
wsmelton: There is no reason to use this as the assumption is the user has decided either from providing the database name or piping what databases will be included or excluded.

dbatools.psd1
```diff
@@ -544,6 +544,7 @@
         # solvable filesystem issues or other workarounds
         'Install-DbaSqlWatch',
         'Uninstall-DbaSqlWatch',
+        'Remove-DbaDbView',
```
wsmelton, line 547: Please remove this from the PR, maintainer control this changes to the manifest and psm1 files.

wsmelton (changes requested):

functions/Remove-DbaDbView.ps1
andreasjordan: Oh, yes - I forgot the target.

dbatools.psd1
MikeyBronowski: Sorry, I must have clicked the wrong button :)

functions/Remove-DbaDbView.ps1
```diff
@@ -0,0 +1,118 @@
...
+        Login to the target instance using alternative credentials. Accepts PowerShell credentials (Get-Credential).
+
+        Windows Authentication, SQL Server Authentication, Active Directory - Password, and Active Directory - Integrated are all supported.
+
+        For MFA support, please use Connect-DbaInstance.
+
+    .PARAMETER Database
+        The database(s) to process. This list is auto-populated from the server. If unspecified, all databases will be processed.
+
+    .PARAMETER ExcludeDatabase
```
wsmelton, line 22: We can remove this function as it is not used in our other Remove functions but also they should control via input or Database the exact ones to use or not.

tests/Remove-DbaDbView.Tests.ps1
```diff
@@ -0,0 +1,74 @@
...
+    }
+}
+
+Describe "$CommandName Integration Tests" -Tags "IntegrationTests" {
+    BeforeAll {
+        $server = Connect-DbaInstance -SqlInstance $script:instance2
+        $view1 = "dbatoolssci_view1_$(Get-Random)"
+        $view2 = "dbatoolssci_view2_$(Get-Random)"
+        $dbname1 = "dbatoolsci_$(Get-Random)"
+        $null = New-DbaDatabase -SqlInstance $script:instance2 -Name $dbname1
```
wsmelton, line 22:
```suggestion
        $null = New-DbaDatabase -SqlInstance $server -Name $dbname1
```

functions/Remove-DbaDbView.ps1
wsmelton: Yes, all of our commands should accept input from a string or `Get-Content serverlist.txt | Verb-Dba...`

This should not be removed and is the template description used across all of our public functions.

wsmelton:
If you look at the process you'd use in an environment for removing objects from n+1 databases I would think it best to have the InputObject accept a Database class, but I can see cases where users want to select them based on the parent Get function of the object.

andreasjordan:
I would only allow piped input of type of object to remove. This way we only have two parameter sets: Using the command with the same parameters as the Get- command (which will internally just call get Get- command with these parameters and then drop the returning objects) or looping through the piped objects and just drop them.

I personally think that a lot of the commands that allow different piped input objects and a mix of piped and parameter input are not deeply tested - especially with arrays as piped inputs and the additional use of parameters with arrays. On the long run I would suggest some breaking changes to reduce functionality but improve flawlessness. But this is my personal opinion and no one has to follow me.

MikeyBronowski:
@wsmelton @andreasjordan Thank you for all the comments and reviews. I will park it for now until the "standard" is decided for Remove-* functions.
Do you want me to remove (oh irony) the Remove- function from #7313  until this decision is made?

andreasjordan:
I would just wait with further work on #7313, maybe change it to a draft.

MikeyBronowski:
> I would just wait with further work on #7313, maybe change it to a draft.

I think the only concern for that one was for the .Drop() method in the Remove-* function, the other two were ok?

potatoqualitee:
I will need a better brain day to think about the Removes that we'll do, but I'm leaning pscustomobject. I do that with  my newer modules 👍🏼

jpomfret:
Hey folks,
Had a think about this and I think I'd personally go for piping from `Get-` into the `Remove-` so you pass an object of the thing you want to get rid of into the remove. However, @potatoqualitee I'm interested in seeing an example of using PSCustomObject - I'm happy to be convinced otherwise if there is a better way.

I think my preference on that is because I'm thinking about DscResources and how the Get\Set\Test kind of fit together.

Side note - are these templates up-to-date for our current preferences and standards?  If not once we decide on how to handle this I'm happy to work on updating those so we can easily point people to them

https://github.com/sqlcollaborative/dbatools-templates/

potatoqualitee:
Agreed, the piped input should be type of object to remove.

functions/Remove-DbaDbView.ps1
```diff
@@ -0,0 +1,105 @@
...
+        if ($SqlInstance) {
+            $InputObject = Get-DbaDbView -SqlInstance $SqlInstance -SqlCredential $SqlCredential -Database $Database -View $View -ExcludeSystemView -ExcludeDatabase $ExcludeDatabase
+        }
+
+        foreach ($vw in $InputObject) {
+
+            if ($Pscmdlet.ShouldProcess($vw.SqlInstance, "Removing the view $vw in the database $($vw.Parent.Name)")) {
+                try {
+                    $viewToDrop = $vw.Parent.Views | Where-Object { $_.Name -eq $vw.Name }
+                    $viewToDrop.Drop()
```
andreasjordan, line 98: This is not needed and dangerous. Just use the view object you already have: $vw.Drop()

functions/Remove-DbaDbView.ps1
```diff
@@ -0,0 +1,105 @@
...
+        [string[]]$View,
+        [parameter(ValueFromPipeline)]
+        [Microsoft.SqlServer.Management.Smo.View[]]$InputObject,
+        [switch]$EnableException
+    )
+
+    process {
+
+        if ($SqlInstance) {
+            $InputObject = Get-DbaDbView -SqlInstance $SqlInstance -SqlCredential $SqlCredential -Database $Database -View $View -ExcludeSystemView -ExcludeDatabase $ExcludeDatabase
```
andreasjordan, line 90: What about `$InputObject = Get-DbaDbView @PSBoundParameters`? This way only the used parameters are send, what prevents problems when the called function uses Test-Bound.

functions/Remove-DbaDbView.ps1
```diff
@@ -0,0 +1,105 @@
...
+        [DbaInstanceParameter[]]$SqlInstance,
+        [PSCredential]$SqlCredential,
+        [object[]]$Database,
+        [object[]]$ExcludeDatabase,
+        [string[]]$View,
+        [parameter(ValueFromPipeline)]
+        [Microsoft.SqlServer.Management.Smo.View[]]$InputObject,
+        [switch]$EnableException
+    )
+
```
andreasjordan, line 86: I would like to have a begin block with a test that only $InputObject XOR $SqlInstance can be used.

andreasjordan:
I think we are very close to finishing this.

functions/Remove-DbaDbView.ps1
andreasjordan: Ups. Not in the begin block, but at the beginning of the process block.

functions/Remove-DbaDbView.ps1
MikeyBronowski: Was not aware of that. I will give it a go next week :) Thank you

MikeyBronowski:
> I think we are very close to finishing this.

Thanks for all the reviews @andreasjordan . It was purely based on some existing function, but I am learning here a lot

andreasjordan:
Hi @MikeyBronowski , can I help you with this?

potatoqualitee:
I'll ping @MikeyBronowski on Twitter 😊

MikeyBronowski:
> Hi @MikeyBronowski , can I help you with this?

I will make some changes today and run tests. Appreciate your help!

andreasjordan:
The last Remove-DbaDb* command we redesigned was Remove-DbaDbSequence: https://github.com/sqlcollaborative/dbatools/blob/development/functions/Remove-DbaDbSequence.ps1

We had some good discussions (#7408) about the functionallity and the style. And I would like to use this as a role model for similar commands. So can you please have a look and try to align this command with Remove-DbaDbSequence?

So basic principles are:
* Two ways of input: Pipelining from Get- command or using same parameters as Get- command. In the later case just issue Get- command to get the objects to remove.
* PSCustomObject as output for every object with properties Status and IsRemoved to indicate success or failure.

MikeyBronowski:
> The last Remove-DbaDb* command we redesigned was Remove-DbaDbSequence: https://github.com/sqlcollaborative/dbatools/blob/development/functions/Remove-DbaDbSequence.ps1
> 
> We had some good discussions (#7408) about the functionallity and the style. And I would like to use this as a role model for similar commands. So can you please have a look and try to align this command with Remove-DbaDbSequence?
> 
> So basic principles are:
> 
> * Two ways of input: Pipelining from Get- command or using same parameters as Get- command. In the later case just issue Get- command to get the objects to remove.
> * PSCustomObject as output for every object with properties Status and IsRemoved to indicate success or failure.

oh, ok. I will have a look then.

functions/Remove-DbaDbView.ps1
```diff
@@ -0,0 +1,77 @@
+function Remove-DbaDbView {
+    <#
+    .SYNOPSIS
+        Removes a database view from database(s) for each instance(s) of SQL Server.
+
+    .DESCRIPTION
+        The Remove-DbaDbView removes view(s) from database(s) for each instance(s) of SQL Server.
```
wsmelton, line 7:
```suggestion
        Drops a database view(s), with supported piping from Get-DbaDbView.
```

functions/Remove-DbaDbView.ps1
```diff
@@ -0,0 +1,77 @@
+function Remove-DbaDbView {
+    <#
+    .SYNOPSIS
+        Removes a database view from database(s) for each instance(s) of SQL Server.
```
wsmelton, line 4:
```suggestion
        Removes a database view(s) from each database and SQL Server instance.
```

functions/Remove-DbaDbView.ps1
```diff
@@ -0,0 +1,116 @@
+function Remove-DbaDbView {
+    <#
+    .SYNOPSIS
+        Removes a database view(s) from each database and SQL Server instance.
+
+    .DESCRIPTION
+        Removes a database view(s), with supported piping from Get-DbaDbView.
+
+    .PARAMETER SqlInstance
+        The target SQL Server instance or instances. This can be a collection and receive pipeline input to allow the function to be executed against multiple SQL Server instances.
```
andreasjordan, line 10: This text has to be changed, as the parameter does not accept pipeline input.

functions/Remove-DbaDbView.ps1
```diff
@@ -0,0 +1,116 @@
...
+        Windows Authentication, SQL Server Authentication, Active Directory - Password, and Active Directory - Integrated are all supported.
+
+        For MFA support, please use Connect-DbaInstance.
+
+    .PARAMETER Database
+        The target database(s).
+
+    .PARAMETER View
+        The name(s) of the view(s).
+
```
andreasjordan, line 24: The Get- command has a parameter ExcludeSystemView - what do you think: Would it be worth adding this here as well?

functions/Remove-DbaDbView.ps1
```diff
@@ -0,0 +1,116 @@
...
+        License: MIT https://opensource.org/licenses/MIT
+
+    .LINK
+        https://dbatools.io/Remove-DbaDbView
+
+    .EXAMPLE
+        PS C:\> $views = Get-DbaDbView -SqlInstance localhost, sql2016 -Database db1, db2 -View view1, view2, view3
+        PS C:\> $views | Remove-DbaDbView
+
+        Removes view1, view2, view3 from db1 and db2 on the local and sql2016 SQL Server instances.
```
andreasjordan, line 55: Can you add a second example to show the non-piped usage?

tests/Remove-DbaDbView.Tests.ps1
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+}
+
+Describe "$CommandName Integration Tests" -Tag "IntegrationTests" {
+
+    BeforeAll {
+
+        $instance2 = Connect-DbaInstance -SqlInstance $script:instance2
+        $null = Get-DbaProcess -SqlInstance $instance2 | Where-Object Program -match dbatools | Stop-DbaProcess -Confirm:$false
```
andreasjordan, line 20: I think this line should be removed - as tests migth run in parallel.

tests/Remove-DbaDbView.Tests.ps1
```diff
@@ -0,0 +1,48 @@
...
+            Compare-Object -ReferenceObject $knownParameters -DifferenceObject $params | Should -BeNullOrEmpty
+        }
+    }
+}
+
+Describe "$CommandName Integration Tests" -Tag "IntegrationTests" {
+
+    BeforeAll {
+
+        $instance2 = Connect-DbaInstance -SqlInstance $script:instance2
```
andreasjordan, line 19: When we put the returnd Server SMO from Connect-DbaInstance into a variable - we call it $server.

tests/Remove-DbaDbView.Tests.ps1
```diff
@@ -0,0 +1,48 @@
...
+        $null = New-DbaDatabase -SqlInstance $instance2 -Name $dbname1
+
+        $view1 = "dbatoolssci_view1_$(Get-Random)"
+        $view2 = "dbatoolssci_view2_$(Get-Random)"
+        $null = $instance2.Query("CREATE VIEW $view1 (a) AS (SELECT @@VERSION );" , $dbname1)
+        $null = $instance2.Query("CREATE VIEW $view2 (b) AS (SELECT 1);", $dbname1)
+    }
+
+    AfterAll {
+        $null = $dbname1 | Remove-DbaDatabase -Confirm:$false
```
andreasjordan, line 31: That works? Without the info in which instance the database should be removed?

MikeyBronowski: Good call.

tests/Remove-DbaDbView.Tests.ps1
potatoqualitee: This was added by me because it helps with appveyor's limited resources (tho the new experimental path may help with exhaustion). Tests do not run in parallel as they are isolated to synchronous commands in a VM. Please leave it in. 

Also, I'd like to prioritize the merging of this PR as it may cause burnout in the author and he used my commands as a baseline 🤗

functions/Remove-DbaDbView.ps1
potatoqualitee: Good idea! However, we can add it in a later PR, let's get this one merged.

andreasjordan: I agree.

tests/Remove-DbaDbView.Tests.ps1
potatoqualitee: using $instance2 is a common pattern in our tests. We can leave as is.

tests/Remove-DbaDbView.Tests.ps1
andreasjordan: Thanks for the info. So no change needed here.

tests/Remove-DbaDbView.Tests.ps1
potatoqualitee: Once this test is fix and the example is added, this will be mergeable. I'll take care of it!

tests/Remove-DbaDbView.Tests.ps1
andreasjordan: Ok.

functions/Remove-DbaDbView.ps1
potatoqualitee: I'll fix this as well.

potatoqualitee: We may have to fix this in other remove commands too in another PR

functions/Remove-DbaDbView.ps1
potatoqualitee: done! once tests pass, i should be able to merge unless you find any mistakes

andreasjordan (approved):
Yes!

potatoqualitee:
Woo@ Merging and will publish shortly.

PR #7356, opened 2021-05-22
Test-DbaTempDbConfig checks SQL 2017 and 2019 for TF1118 and it is not needed

gbargsley:
## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
 
When deploying a new server recently I noticed that the TF1118 best practice check is still being executed against SQL 2016, 17 and 19 where the functionality is enabled by default.  This in my opinion causes confusing on best practice check when best practice is returned as false, but the functionality is enabled by default.  So it is like a false positive.
### Approach

I added a version check and changed the output text for SQL 2016 and above for this check.
### Screenshots

SQL 2019 output
image

SQL 2014 output
image


potatoqualitee:
Looks great! Thanks for the catch 🙇🏼

PR #7357, opened 2021-05-23
Remove extra bogus.dll

### LINKED ISSUE #7303: [Bug] Two copies of Bogus.dll in the module
### Report

PR #5331 added `Bogus.dll` but we also have this file living under our 3rd-party folder folder [bin/libraries/third-party/Bogus](https://github.com/sqlcollaborative/dbatools/tree/development/bin/libraries/third-party/Bogus).

I do not find anywhere explicitly loading the DLL from either location so I would assume it is inherited in some fashion.

#### Errors Received
N/A

#### Expected Behavior

We only have one copy of a given DLL file.

#### Actual Behavior

We have two copies.

potatoqualitee:
fixes #7303

bogus is loaded from the bogus dir

https://github.com/sqlcollaborative/dbatools/blob/development/internal/scripts/libraryimport.ps1#L95


jpomfret (approved):
agreed

PR #7390, opened 2021-05-31
Get-/Set-DbaTcpPort - Rewriting the commands and using Get-/Set-DbaNetworkConfiguration

andreasjordan:
Just a draft with first ideas.
There have been issues with the commands, so this is also a bugfix.

## Type of Change

 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system


potatoqualitee:
Tests pass but Set has no integration tests :/ Would you be so kind as to write at least one that ensure the command works as expected?

andreasjordan:
How are the instances in AppVayor configured? Can I change the port without affecting other tests?
If they are on static port, I could just change that and then change it back.
If they are on dynamic port, I could change to static and then back to dynamic with Set-DbaNetworkConfiguration.

andreasjordan:
I can just change the configuration without restarting the service or also restarting the service. What should the test do?

niphlod:
the service_restarts env should be the best option here, as it runs pretty destructive tests already.
instances are configured using these scripts \tests\appveyor.SQL2017.ps1,\tests\appveyor.SQL2016.ps1

potatoqualitee:
Fantastic! Thank you.

PR #7411, opened 2021-06-05
New-DbaDbUser - Remove wrong example

### LINKED ISSUE #6270: [bug] New-DbaDbUser support for input via pipeline
Example #4 for New-DbaDbUser is incorrect.  It states:
`Get-DbaDbUser -SqlInstance sqlserver1 -Database DB1 | New-DbaDbUser -SqlInstance sqlserver2 -Database DB1`

When I attempt to execute similar code, I get "The input object cannot be bound to any parameters for the command either because the command does not take pipeline input or the input and its properties do not match any of the parameters that take pipeline input." 

Which makes sense because none of the params for `New-DbaDbUser` accept pipeline input.
It would be nice if it did in fact accept pipeline input so we could copy one or more users from one DB to another.  In the short term, removing the example from the docs would  be sufficient.

https://github.com/sqlcollaborative/dbatools/blob/master/functions/New-DbaDbUser.ps1

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #6270 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
Just removed the wrong example.

### DIFF
--- a/functions/New-DbaDbUser.ps1
+++ b/functions/New-DbaDbUser.ps1
@@ -71,11 +71,6 @@ function New-DbaDbUser {
 
         Creates a new sql user named user1 mapped to Login1 in the specified database.
 
-    .EXAMPLE
-        PS C:\> Get-DbaDbUser -SqlInstance sqlserver1 -Database DB1 | New-DbaDbUser -SqlInstance sqlserver2 -Database DB1
-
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Agreed with the solution. Unsure why that example was ever added 🤔

PR #7414, opened 2021-06-05
SQL Watch commands cannot be used in PowerShell 7

### LINKED ISSUE #6567: Install-DbaSqlWatch broken in PowerShell 7
### Summary of new feature

* I've tried to install Install-DbaSqlWatch with PowerShell 7.1 which didn't work. This is IMHO not a bug, but rather a new feature as it does work with PowerShell 6 just fine.

wsmelton:
Tempoary fix until we address import process.

## Type of Change

 - [x] Bug fix (non-breaking change, fixes #6567 )
 - [ ] Pester test is included
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
The DLL used is not supported in dotnet Core, only .NET 4.6 making this command only compatible with Windows PowerShell.

SQL Watch commands in addition to any command trying to use the `Microsoft.SqlServer.Dac.dll` are now going to throw a message that PowerShell Core is not supported if they are on Windows.

The commands are only imported at this time if they are Windows but we do not address the Edition of PowerShell in our import process. Our import process needs to have an additional check on PowerShell Edition so we can properly exclude these commands on import and lessen the confusion.

### DIFF
--- a/functions/Export-DbaDacPackage.ps1
+++ b/functions/Export-DbaDacPackage.ps1
@@ -120,6 +120,11 @@ function Export-DbaDacPackage {
     process {
         if (Test-FunctionInterrupt) { return }
 
+        if ($PSEdition -eq 'Core') {
+            Stop-Function -Message "PowerShell Core is not supported, please use Windows PowerShell."
+            return
+        }
+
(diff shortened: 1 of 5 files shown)


wsmelton:
Odd failure of `Get-DbaDbSharePoint`, not related to PR.

PR #7416, opened 2021-06-06
Connect-DbaInstance - Set new codepath as default

andreasjordan:
Not sure if we are ready to do this now, but I want to see if all tests pass.

### DIFF
--- a/functions/Connect-DbaInstance.ps1
+++ b/functions/Connect-DbaInstance.ps1
@@ -18,6 +18,14 @@ function Connect-DbaInstance {
 
         To execute SQL commands, you can use $server.ConnectionContext.ExecuteReader($sql) or $server.Databases['master'].ExecuteNonQuery($sql)
 
+        On Nov 6, 2020 we added a complete new way of processing inside of this command, available via the config switch sql.connection.experimental.
+        Now it is time to use this as the default, but still allow users to use the old code path via a config switch. Please use:
+        Set-DbatoolsConfig -FullName sql.connection.legacy -Value $true
+
+        Some parameters are not supportet by the new code path. If your code relies on them, please use the legacy code path and please open an issue:
+        https://github.com/sqlcollaborative/dbatools/issues
+        This will help us to know which parameters are used and we will try to implement them in the new code path as quick as possible.
+
     .PARAMETER SqlInstance
         The target SQL Server instance or instances. This can be a collection and receive pipeline input to allow the function to be executed against multiple SQL Server instances.
 
@@ -899,7 +907,7 @@ function Connect-DbaInstance {
             #>
 
             if ($AccessToken) {
-                Stop-Function -Message "AccessToken is only supported in the new experimental code path. See documentation for details."
+                Stop-Function -Message "AccessToken is not supported in the legacy code path."
                 return
             }
 
--- a/functions/Connect-DbaInstance.ps1
+++ b/functions/Connect-DbaInstance.ps1
@@ -851,7 +851,7 @@ function Connect-DbaInstance {
                         Write-Message -Level Debug -Message "Setting ConnectionContext.SqlExecutionModes to '$SqlExecutionModes'"
                         $server.ConnectionContext.SqlExecutionModes = $SqlExecutionModes
                     }
-                    if (Test-Bound -ParameterName 'StatementTimeout') {
+                    if ($null -ne $StatementTimeout) {
                         Write-Message -Level Debug -Message "Setting ConnectionContext.StatementTimeout to '$StatementTimeout'"
                         $server.ConnectionContext.StatementTimeout = $StatementTimeout
                     }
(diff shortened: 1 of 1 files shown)


andreasjordan:
These commands change the StatementTimeout of the ConnectionContext. If we pass in a server SMO, this is changed even after the command finishes. Connect-DbaInstance now copies the context if a server SMO and this parameter is passed in. So all these commands should be changed to include this parameter when Connect-DbaInstance is unsed instead of Connect-SqlInstance:

* Backup-DbaDatabase
* Restore-DbaDatabase
* Get-DbaTopResourceUsage
* Import-DbaCsv
* Invoke-DbaDbLogShipping
* Invoke-DbaDbShrink
* Invoke-DbaDbUpgrade
* Set-DbaDbCompression
* Test-DbaDbCompression
* Start-DbccCheck

This is primarily a reminder for myself.

andreasjordan:
These commands use AzureUnsupported and we should check regularly if this is realy not supported on Azure: 
* Backup-DbaDatabase  
* Copy-DbaLogin
* Get-DbaLogin
* Set-DbaLogin
* Get-DbaDefaultPath
* Get-DbaUserPermissions
* Get-DbaXESession
* New-DbaCustomError
* Remove-DbaCustomError

andreasjordan:
This command uses ApplicationIntent
* Invoke-DbaQuery

Will add this to Connect-DbaInstance as well - we should later rewrite Invoke-DbaQuery.

andreasjordan:
This command uses NonPooledConnection:
* Install-DbaFirstResponderKit

andreasjordan:
Difference to the old code path: $StatementTimeout
Parameter is definded as:
[int]$StatementTimeout = (Get-DbatoolsConfigValue -FullName 'sql.execution.timeout'),
Default config is: 0
In the old code path the test is:
if (Test-Bound -ParameterName 'StatementTimeout') { $server.ConnectionContext.StatementTimeout = $StatementTimeout }
So it is only set when parameter is used, not if default is used.
But SMO default is 600 - so this is active in the old code path.

To not change the behavior we should set the config value to 600 by default.

Or any other ideas?
@potatoqualitee @wsmelton

wsmelton:
> These commands change the StatementTimeout of the ConnectionContext. If we pass in a server SMO, this is changed even after the command finishes. Connect-DbaInstance now copies the context if a server SMO and this parameter is passed in. So all these commands should be changed to include this parameter when Connect-DbaInstance is unsed instead of Connect-SqlInstance:

A configuration item or user can set it by first creating a connection before calling the given command. If you start adding a timeout parameter to commands then we are going to end up getting a request to add it to all 500+, there is no reason to do this from the start.

wsmelton:
> These commands use AzureUnsupported and we should check regularly if this is realy not supported on Azure:

They are unsupported because we don't manage the connection in a way that Azure can handle and/or the object management is not supported by SMO in Azure (newer version might be to-date that I'm aware there is no "official" support with SMO against Azure SQL, only MI and Synapsis in limited mode).

wsmelton:
Please put PRs as draft if they are not ready to be reviewed fully as well.

functions/Connect-DbaInstance.ps1
```diff
@@ -18,6 +18,14 @@ function Connect-DbaInstance {
 
         To execute SQL commands, you can use $server.ConnectionContext.ExecuteReader($sql) or $server.Databases['master'].ExecuteNonQuery($sql)
 
+        On Nov 6, 2020 we added a complete new way of processing inside of this command, available via the config switch sql.connection.experimental.
```
wsmelton, line 21: Reads easier if you reference the build version instead of the date with modules.

andreasjordan:
> > These commands change the StatementTimeout of the ConnectionContext. If we pass in a server SMO, this is changed even after the command finishes. Connect-DbaInstance now copies the context if a server SMO and this parameter is passed in. So all these commands should be changed to include this parameter when Connect-DbaInstance is unsed instead of Connect-SqlInstance:
> 
> A configuration item or user can set it by first creating a connection before calling the given command. If you start adding a timeout parameter to commands then we are going to end up getting a request to add it to all 500+, there is no reason to do this from the start.

Sorry - we have a missunderstanding. The parameter should not be added to the command, it should only be added to the internal call to Connect-DbaInstance. I'll show it later in details.

andreasjordan:
About the StatementTimeout:

Import-DbaCsv does it the right way:
https://github.com/sqlcollaborative/dbatools/blob/8bda66fa60772068289d815a8a92906c4ee174bf/functions/Import-DbaCsv.ps1#L442
It requests a connection with no StatementTimeout. Only problem: That does not work if $instance is already a server SMO, as it is just returned by Connect-SqlInstance. 

Invoke-DbaDbUpgrade does change $instance permanently if it is already a server SMO:
https://github.com/sqlcollaborative/dbatools/blob/8bda66fa60772068289d815a8a92906c4ee174bf/functions/Invoke-DbaDbUpgrade.ps1#L132-L133
Instead it should use:
```
$server = Connect-DbaInstance -SqlInstance $instance -SqlCredential $SqlCredential -StatementTimeout 0
```
(Because `[Int32]::MaxValue` should be 0 / Yes, we have to use Connect-DbaInstance instead of Connect-SqlInstance to let this work / and the new code path has also to be used)

andreasjordan:
Sorry for this pull request, will close this shortly. There are so many optimizations now that I will only include these changes in the new path for now and not set it as default yet. 
But I am now very happy with the code and would like to provide it as default soon and then also adapt the examples.

Issue #6479, opened 2020-04-17
[Bug] Install-DbaInstance fails when action = 'AddNode' on SQL2016+


OmegaMadLab:
<!--
🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨
The core team may close bug reports that do not follow our provided template.

The information we ask for helps to better troubleshoot the report. We release frequently and often, this information helps to resolve the issue more efficiently.
-->

<!--
dbatools 1.0 has been released as of June 20, 2019
Please ensure you are on this version before submitting an issue
-->

### Environmental information

<!--
Run below command, paste results below:
& {"``````";"#### PowerShell version:`n$($PSVersionTable | Out-String)"; "`n#### dbatools Module version:`n$(gmo dbatools -List | select name, path, version | fl -force | Out-String)";"``````"} | clip
-->
```
#### PowerShell version:

Name                           Value                                                                                                                                                                                                                                                                                                                      
----                           -----                                                                                                                                                                                                                                                                                                                      
PSVersion                      5.1.17763.1007                                                                                                                                                                                                                                                                                                             
PSEdition                      Desktop                                                                                                                                                                                                                                                                                                                    
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                                                                                                                                                                                                                                                                                                    
BuildVersion                   10.0.17763.1007                                                                                                                                                                                                                                                                                                            
CLRVersion                     4.0.30319.42000                                                                                                                                                                                                                                                                                                            
WSManStackVersion              3.0                                                                                                                                                                                                                                                                                                                        
PSRemotingProtocolVersion      2.3                                                                                                                                                                                                                                                                                                                        
SerializationVersion           1.1.0.1                                                                                                                                                                                                                                                                                                                    




#### dbatools Module version:


Name    : dbatools
Path    : C:\Program Files\WindowsPowerShell\Modules\dbatools\1.0.105\dbatools.psd1
Version : 1.0.105




```

#### SQL Server:
<!-- Paste output of `SELECT @@VERSION` -->
```sql
Microsoft SQL Server 2019 (RTM-CU3) (KB4538853) - 15.0.4023.6 (X64)   Mar  4 2020 00:59:26   Copyright (C) 2019 Microsoft Corporation  Developer Edition (64-bit) on Windows Server 2019 Datacenter 10.0 <X64> (Build 17763: ) (Hypervisor) 
```
### Report

<!--
Things to consider:
- Errors received are not related to permissions?
[...]

PR #7437, opened 2021-06-14
Install-DbaInstance - Test for allowed ACTION before adding SQLTEMPDBFILECOUNT

andreasjordan:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose

I want to be able to run ACTION=Uninstall and use `-ConfigurationFile` for that. 
It's a similar case like #6479 that was addressed with #6480.
To be totally flexible I asked the documentation what ACTIONs allow SQLTEMPDBFILECOUNT:
https://docs.microsoft.com/en-us/sql/database-engine/install-windows/install-sql-server-from-the-command-prompt
So I added these 5 ACTIONs.
I moved the test so that there is no unsused call to the computer for no reason.

@nvarscar : What do you think? If you would like to solve it some other way, I'm open for feedback.

### DIFF
--- a/functions/Install-DbaInstance.ps1
+++ b/functions/Install-DbaInstance.ps1
@@ -613,7 +613,7 @@ function Install-DbaInstance {
             if ($canonicVersion -gt '10.0') {
                 $execParams += '/IACCEPTSQLSERVERLICENSETERMS'
             }
-            if ($canonicVersion -ge '13.0' -and (-Not $configNode.SQLTEMPDBFILECOUNT)) {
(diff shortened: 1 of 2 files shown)


nvarscar (approved):
That makes sense, no reason to add tempdb file count upon uninstall

potatoqualitee:
Thank you both! 🥳

PR #7465, opened 2021-06-23
Copy-DbaLinkedServer - New parameter to exclude copying sensitive info

### LINKED ISSUE #6422: Copy-DbaLinkedServer: Add option to Exclude Password
### Summary of new feature

Would like to use the Copy-DbaLinkedServer feature but because of the use of the Get-DecryptedObject this is not possible due to inability to connect via DAC.  In enterprises where there are regulations and audits based on access to sensitive data, there is lock down for the Browser service and it will not be running.  The option is available in the Export-DbaLinkedServer option and default of the results of a Get-DBALinkedServer "Script()" call so it would be helpful to include option for this scenario.  

Code is based on this action
https://blog.netspi.com/decrypting-mssql-database-link-server-passwords/

potatoqualitee:
## Type of Change
 - [x] New feature (non-breaking change, adds functionality, fixes #6422 )

The current approach may fix it, but I'm on a smaller laptop and haven't read the rest of the code. I'll take another look when I'm using a bigger monitor.

### DIFF
--- a/functions/Copy-DbaLinkedServer.ps1
+++ b/functions/Copy-DbaLinkedServer.ps1
@@ -37,6 +37,9 @@ function Copy-DbaLinkedServer {
     .PARAMETER UpgradeSqlClient
         Upgrade any SqlClient Linked Server to the current Version
 
+    .PARAMETER ExcludePassword
+        Copies the linked server without any sensitive information.
+
     .PARAMETER WhatIf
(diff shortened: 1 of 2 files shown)


potatoqualitee:
k, tested this a number of times locally

andreasjordan (approved):
Looks good.

potatoqualitee:
awesome, thank you!

PR #7466, opened 2021-06-23
Get-DbaDbMirror - Add new parameter for Database filtering

### LINKED ISSUE #5881: Suggestion for Get-DbaDbMirror
Would be useful to natively filter on database, so it can be used to pass stuff through without extensive filtering.

I am using it to automate the decommissioning of the mirroring partners as part of a migration and the mirror is broken.

(Alternative add an option to remove-dbadbmirror that cleans up the orphaned pairs)

potatoqualitee:
## Type of Change
 - [x] New feature (non-breaking change, adds functionality, fixes #5881 )

We don't have integration tests for this and I don't have any mirrored databases right now (tho I can explore in appveyor). @BilliousBT can you give this a shot?

Ty!

### DIFF
--- a/functions/Invoke-DbaDbMirroring.ps1
+++ b/functions/Invoke-DbaDbMirroring.ps1
@@ -365,7 +365,7 @@ function Invoke-DbaDbMirroring {
                     Stop-Function -Continue -Message "Failure with the new last part" -ErrorRecord $_
                 }
             }
-write-warning HELLO
+
             if ($Pscmdlet.ShouldProcess("console", "Showing results")) {
                 $results = [pscustomobject]@{
                     Primary  = $Primary
--- a/functions/Invoke-DbaDbMirroring.ps1
+++ b/functions/Invoke-DbaDbMirroring.ps1
@@ -191,7 +191,7 @@ function Invoke-DbaDbMirroring {
             foreach ($currentmirror in $Mirror) {
                 $stepCounter = 0
                 try {
-                    $dest = Connect-SqlInstance -SqlInstance $currentmirror -SqlCredential $currentmirrorSqlCredential
+                    $dest = Connect-SqlInstance -SqlInstance $currentmirror -SqlCredential $MirrorSqlCredential
 
                     if ($Witness) {
                         $witserver = Connect-SqlInstance -SqlInstance $Witness -SqlCredential $WitnessSqlCredential
@@ -250,7 +250,7 @@ function Invoke-DbaDbMirroring {
                     if ($Pscmdlet.ShouldProcess("$currentmirror", "Restoring full and log backups of $primarydb from $Primary")) {
                         foreach ($currentmirrorinstance in $currentmirror) {
                             try {
-                                $null = $allbackups | Restore-DbaDatabase -SqlInstance $currentmirrorinstance -SqlCredential $currentmirrorSqlCredential -WithReplace -NoRecovery -TrustDbBackupHistory -EnableException
+                                $null = $allbackups | Restore-DbaDatabase -SqlInstance $currentmirrorinstance -SqlCredential $MirrorSqlCredential -WithReplace -NoRecovery -TrustDbBackupHistory -EnableException
                             } catch {
                                 Stop-Function -Message "Failure" -ErrorRecord $_ -Target $dest -Continue
                             }
--- a/functions/Invoke-DbaDbMirroring.ps1
+++ b/functions/Invoke-DbaDbMirroring.ps1
@@ -205,6 +205,7 @@ function Invoke-DbaDbMirroring {
                 Write-ProgressHelper -StepNumber ($stepCounter++) -Message "Validating mirror setup"
                 # Thanks to https://github.com/mmessano/PowerShell/blob/master/SQL-ConfigureDatabaseMirroring.ps1 for the tips
 
+                $params.Database = $dbName
                 $validation = Invoke-DbMirrorValidation @params
 
                 if ((Test-Bound -ParameterName SharedPath) -and -not $validation.AccessibleShare) {
@@ -215,8 +216,10 @@ function Invoke-DbaDbMirroring {
                     Stop-Function -Continue -Message "This mirroring configuration is not supported. Because the principal server instance, $source, is $($source.EngineEdition) Edition, the mirror server instance must also be $($source.EngineEdition) Edition."
                 }
 
-                if ($validation.MirroringStatus -ne "None") {
-                    Stop-Function -Continue -Message "Cannot setup mirroring on database ($dbName) due to its current mirroring state: $($primarydb.MirroringStatus)"
+                foreach ($status in $validation.MirroringStatus) {
+                    if ($status -ne "None") {
+                        Stop-Function -Continue -Message "Cannot setup mirroring on database ($dbName) due to its current mirroring state on primary: $status"
+                    }
                 }
 
                 if ($primarydb.Status -ne "Normal") {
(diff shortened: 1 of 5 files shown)


potatoqualitee:
K - I tested the heck out of this locally but can't run them on appveyor because the underlying hostname has to be different. In the future when I move most commands to GH Actions, we'll have two containers that'll work

potatoqualitee:
lol sql on linux does not support mirroring, guess we'll have to rely on my lab

potatoqualitee:
It turns out that as I was adding this, I found some other bugs, so I added in the fixes for them.

potatoqualitee:
Please ignore that -Match casing, I need to figure out how to get VS Code to behave again.

Anyway! WOOHOO! Even though it says this

image

I still got it to work on Linux

image

functions/Invoke-DbaDbMirroring.ps1
```diff
@@ -215,8 +216,10 @@ function Invoke-DbaDbMirroring {
                     Stop-Function -Continue -Message "This mirroring configuration is not supported. Because the principal server instance, $source, is $($source.EngineEdition) Edition, the mirror server instance must also be $($source.EngineEdition) Edition."
                 }
 
-                if ($validation.MirroringStatus -ne "None") {
-                    Stop-Function -Continue -Message "Cannot setup mirroring on database ($dbName) due to its current mirroring state: $($primarydb.MirroringStatus)"
+                foreach ($status in $validation.MirroringStatus) {
```
andreasjordan, line 219: Why have you changed from `if` to `foreach`? You have one database, so you get one result from Invoke-DbMirrorValidation.
This has a bad effect on the -Continue in Stop-Function as this only continues the new foreach and not the old (outer) foreach.

functions/Invoke-DbaDbMirroring.ps1
```diff
@@ -317,18 +320,20 @@ function Invoke-DbaDbMirroring {
...
-                    if ($Pscmdlet.ShouldProcess("primary, mirror and witness (if specified)", "Creating login $account and granting CONNECT ON ENDPOINT")) {
-                        $null = New-DbaLogin -SqlInstance $source -Login $account -WarningAction SilentlyContinue
-                        $null = New-DbaLogin -SqlInstance $dest -Login $account -WarningAction SilentlyContinue
-                        try {
-                            $null = $source.Query("GRANT CONNECT ON ENDPOINT::$primaryendpoint TO [$account]")
-                            $null = $dest.Query("GRANT CONNECT ON ENDPOINT::$currentmirrorendpoint TO [$account]")
-                            if ($witserver) {
-                                $null = New-DbaLogin -SqlInstance $witserver -Login $account -WarningAction SilentlyContinue
-                                $witserver.Query("GRANT CONNECT ON ENDPOINT::$witnessendpoint TO [$account]")
+                    if ($account) {
```
andreasjordan, line 323: How can $account ever be $false?

tests/Get-DbaDbMirror.Tests.ps1
```diff
@@ -4,16 +4,45 @@ Write-Host -Object "Running $PSCommandPath" -ForegroundColor Cyan
...
+Describe "$commandname Integration Tests" -Tag "IntegrationTests" {
+    BeforeAll {
+        $null = Get-DbaProcess -SqlInstance $script:instance2, $script:instance3 | Where-Object Program -Match dbatools | Stop-DbaProcess -Confirm:$false -WarningAction SilentlyContinue
+        $server = Connect-DbaInstance -SqlInstance $script:instance2
+        $db1 = "dbatoolsci_mirroring"
+        $db2 = "dbatoolsci_mirroring_db2"
+
+        Remove-DbaDbMirror -SqlInstance $script:instance2, $script:instance3 -Database $db1, $db2 -Confirm:$false
+        Remove-DbaDatabase -SqlInstance $script:instance2, $script:instance3 -Database $db1, $db2 -Confirm:$false
+        $null = Get-DbaDatabase -SqlInstance $script:instance2, $script:instance3 -Database $db1, $db2 | Remove-DbaDatabase -Confirm:$false
```
andreasjordan, line 26: These two line do the same, don't they?

tests/Invoke-DbaDbMirroring.Tests.ps1
```diff
@@ -4,18 +4,18 @@ Write-Host -Object "Running $PSCommandpath" -ForegroundColor Cyan
...
-            (@(Compare-Object -ReferenceObject ($knownParameters | Where-Object {$_}) -DifferenceObject $params).Count ) | Should Be 0
+            (@(Compare-Object -ReferenceObject ($knownParameters | Where-Object { $_ }) -DifferenceObject $params).Count ) | Should Be 0
         }
     }
 }
 
 Describe "$commandname Integration Tests" -Tag "IntegrationTests" {
     BeforeAll {
-        $null = Get-DbaProcess -SqlInstance $script:instance2 | Where-Object Program -match dbatools | Stop-DbaProcess -Confirm:$false -WarningAction SilentlyContinue
+        $null = Get-DbaProcess -SqlInstance $script:instance2 | Where-Object Program -Match dbatools | Stop-DbaProcess -Confirm:$false -WarningAction SilentlyContinue
```
andreasjordan, line 18: Maybe we should always use this syntax: `Where-Object { $_.Program -match 'dbatools' }`

andreasjordan:
Ok, see my comments, that is all I found.

functions/Invoke-DbaDbMirroring.ps1
potatoqualitee: No, I just moved it down. I was not accounting for multiple results in an earlier test. So it would work for db1 but not db1, db2

functions/Invoke-DbaDbMirroring.ps1
potatoqualitee: if it's blank, like with docker containers somehow

image

tests/Get-DbaDbMirror.Tests.ps1
potatoqualitee: lol yeah, i was having a hard time removing them. I'll remove the extra

functions/Invoke-DbaDbMirroring.ps1
andreasjordan: Ok, I see.

tests/Get-DbaDbMirror.Tests.ps1
```diff
@@ -23,7 +23,6 @@ Describe "$commandname Integration Tests" -Tag "IntegrationTests" {
 
         Remove-DbaDbMirror -SqlInstance $script:instance2, $script:instance3 -Database $db1, $db2 -Confirm:$false
         Remove-DbaDatabase -SqlInstance $script:instance2, $script:instance3 -Database $db1, $db2 -Confirm:$false
-        $null = Get-DbaDatabase -SqlInstance $script:instance2, $script:instance3 -Database $db1, $db2 | Remove-DbaDatabase -Confirm:$false
         $null = $server.Query("CREATE DATABASE $db1")
         $null = $server.Query("CREATE DATABASE $db2")
```
andreasjordan: While the `$null =` is not needed with `.Query()` here, I think it should be used in the first two line to prevent the output of the commands.

tests/Invoke-DbaDbMirroring.Tests.ps1
potatoqualitee: it feels more formal and right but bruce payette blessed this syntax as it works on ps3 and adding a script block makes it way slower

tests/Get-DbaDbMirror.Tests.ps1
potatoqualitee: agreed! i tend to add extra $nulls around in case something changes in the future

potatoqualitee:
Thanks for the review! Updating and merging 🥳

PR #7471, opened 2021-06-29
Get- / Test- / Set-DbaPowerPlan - Complete redesign to reduce duplicate code

### LINKED ISSUE #6397: [Bug] Set-DbaPowerPlan fails on non-english language Windows installations
### Environmental information

Run below command, paste results below:
```
#### PowerShell version:

PSVersion                      5.1.17763.1007                                                                          
BuildVersion                   10.0.17763.1007                                                                         

#### dbatools Module version:
Name    : dbatools
Path    : C:\Program Files\WindowsPowerShell\Modules\dbatools\1.0.100\dbatools.psd1
Version : 1.0.100
```

```powershell
1031             de-DE            Deutsch (Deutschland)
```

### Report

Running Set-DbaPowerPlan on a fresh Windows 2019 Datacenter VM installed with the German language fails to set, because the 'High Performance' plan could not be found by the command.

This is similar to issues #2769 and #245

#### Host used

- [x] powershell.exe
- [ ] ISE
- [ ] VS Code
- [ ] Other (please specify)

#### Errors Received

WARNUNG: [10:03:19][Set-DbaPowerPlanInternal] Couldn't find the requested Power Plan 'High Performance' on SQL2019.

#### Steps to Reproduce

Windows 2019 Datacenter VM in German (presumably any non-english install)
Set-DbaPowerPlan -ComputerName {CompterName}

#### Expected Behavior

Use the GUID values for the PowerPlan info stored in the registry

#### Actual Behavior

Fails to load the power plan information from the registry because it is searching for the English language names.

andreasjordan:
## Type of Change

 - [x] Breaking change (effects multiple commands or functionality, fixes #6397 )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose

Reducing duplicate  code and duplicate attempts to get Power Plan information from the target computer.

### Approach

Added -List to Get-DbaPowerPlan and return InstanceIds in Test-DbaPowerPlan so that Test-DbaPowerPlan can be used in Set-DbaPowerPlan.
I have removed the ability to pipe from Get-DbaPowerPlan and Set-DbaPowerPlan as I can't see a use case for it.

### DIFF
--- a/functions/Set-DbaPowerPlan.ps1
+++ b/functions/Set-DbaPowerPlan.ps1
@@ -82,7 +82,7 @@ function Set-DbaPowerPlan {
         foreach ($computer in $ComputerName) {
             try {
                 Write-Message -Level Verbose -Message "Getting and testing Power Plans on $computer."
-                $change = Test-DbaPowerPlan -ComputerName $computer -Credential $Credential -CustomPowerPlan $CustomPowerPlan -EnableException
+                $change = Test-DbaPowerPlan -ComputerName $computer -Credential $Credential -PowerPlan $PowerPlan -EnableException
             } catch {
                 Stop-Function -Message "Can't get Power Plan Info for $computer. Check logs for more details." -Continue -ErrorRecord $_ -Target $computer
             }
--- a/functions/Set-DbaPowerPlan.ps1
+++ b/functions/Set-DbaPowerPlan.ps1
@@ -18,16 +18,8 @@ function Set-DbaPowerPlan {
     .PARAMETER Credential
         Specifies a PSCredential object to use in authenticating to the server(s), instead of the current user account.
 
-    .PARAMETER PowerPlan
-        Specifies the Power Plan that you wish to use.
-
-        We use the English phrase "High Performance" by default. To specify Power Plans in another language, use this parameter (-PowerPlan Höchstleistung).
-
     .PARAMETER CustomPowerPlan
-        Specifies the name of a custom Power Plan to use.
-
-    .PARAMETER InputObject
-        Enables piping from Get-DbaPowerPlan
+        Specifies the name of a custom Power Plan to use. Use Get-DbaPowerPlan -List to get all available Power Plans on a computer.
 
     .PARAMETER EnableException
         By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
@@ -59,14 +51,7 @@ function Set-DbaPowerPlan {
         Sets the Power Plan to High Performance. Skips it if its already set.
 
     .EXAMPLE
-        PS C:\> Set-DbaPowerPlan -ComputerName sql2017 -PowerPlan Höchstleistung
-
-        Sets the PowerPlan on a German system to Höchstleistung.
-
-        We use the English phrase "High Performance" by default. To specify Power Plans in another language, use the -PowerPlan parameter.
-
-    .EXAMPLE
-        PS C:\> 'Server1', 'Server2' | Set-DbaPowerPlan -PowerPlan Balanced
+        PS C:\> 'Server1', 'Server2' | Set-DbaPowerPlan -CustomPowerPlan Balanced
 
         Sets the Power Plan to Balanced for Server1 and Server2. Skips it if its already set.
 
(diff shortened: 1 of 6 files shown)


functions/Set-DbaPowerPlan.ps1
```diff
@@ -83,160 +68,68 @@ function Set-DbaPowerPlan {
...
     #>
     [CmdletBinding(SupportsShouldProcess)]
-    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSUseOutputTypeCorrectly", "", Justification = "PSSA Rule Ignored by BOH")]
     param (
         [parameter(ValueFromPipeline)]
         [DbaInstance[]]$ComputerName,
         [PSCredential]$Credential,
-        [string]$PowerPlan = 'High Performance',
+        [Alias("PowerPlan")]
         [string]$CustomPowerPlan,
```
potatoqualitee, line 76: "Custom" has a specific meaning in Power Plan world. High Performance is not custom, it is standard. Not sure I see the value in changing this parameter. In addition, it is not mandatory and I can't find where it is set. Does it ever get set?

functions/Set-DbaPowerPlan.ps1
```diff
@@ -83,160 +68,68 @@ function Set-DbaPowerPlan {
...
-                            }
-                        } else {
-                            Stop-Function -Message "Couldn't set Power Plan on $computer." -Category ConnectionError -ErrorRecord $_ -Target $computer
-                            return
+            } else {
+                if ($Pscmdlet.ShouldProcess($computer, "Changing Power Plan from $powerPlan to $powerPlanRequested")) {
+                    [System.Guid]$powerPlanGuid = $change.RecommendedInstanceId
+                    $scriptBlock = {
+                        Param ($Guid)
+                        $powerSetActiveSchemeDefinition = '[DllImport("powrprof.dll", CharSet = CharSet.Auto)] public static extern uint PowerSetActiveScheme(IntPtr RootPowerKey, Guid SchemeGuid);'
```
potatoqualitee, line 107: awesome!

functions/Test-DbaPowerPlan.ps1
```diff
@@ -51,7 +52,7 @@ function Test-DbaPowerPlan {
     #>
     param (
         [parameter(ValueFromPipeline)]
-        [DbaInstance[]]$ComputerName = $env:COMPUTERNAME,
```
potatoqualitee, line 54: If this is not set, should it be mandatory as well?

tests/Set-DbaPowerPlan.Tests.ps1
```diff
@@ -5,7 +5,7 @@ Write-Host -Object "Running $PSCommandPath" -ForegroundColor Cyan
 Describe "$CommandName Unit Tests" -Tag 'UnitTests' {
     Context "Validate parameters" {
         [object[]]$params = (Get-Command $CommandName).Parameters.Keys | Where-Object {$_ -notin ('whatif', 'confirm')}
-        [object[]]$knownParameters = 'ComputerName', 'Credential', 'PowerPlan', 'CustomPowerPlan', 'InputObject', 'EnableException'
```
potatoqualitee, line 8: Would removing the InputObject remove the authentication as well? Are you outputting Credential in Get?

potatoqualitee (commented):
Thanks so much!

tests/Set-DbaPowerPlan.Tests.ps1
andreasjordan: Yes, Get- returns the credential it gets. I removed InputObject because I think it does not add any value.
`Get-DbaPowerPlan -SqlInstance XYZ | Set-DbaPowerPlan` does not make any sense, because I don't want to set the power plan I just have. This would just set the best practice plan as `Set-DbaPowerPlan -SqlInstance XYZ` would do with less code.
And `Test-DbaPowerPlan -SqlInstance XYZ | Set-DbaPowerPlan` does not make any sense as well.
So I can only think of piping in ComputerNames - and that works.

functions/Test-DbaPowerPlan.ps1
andreasjordan: No, because I can pipe in from `Get-DbaPowerPlan` and in that case $computerName must not be set (so I had to remove the default).
If we remove InputObject here as well, then we would have to make $ComputerName mandetory. I don't see any added value in doing `Get-DbaPowerPlan -SqlInstance XYZ | Test-DabPowerPlan` because `Test-DabPowerPlan -SqlInstance XYZ` would be shorter. So if you agree, we can remove InputObject in all power plan commands and just accept ComputerName as piped input.

functions/Set-DbaPowerPlan.ps1
andreasjordan: The magic is in Test-DbaPowerPlan. There is the best practice powerplan hard coded. But if you use $CustomPowerPlan, then this would be considered best practive. Test-DbaPowerPlan uses Get-DbaPowerPlan to get all power plans and returns all the information needed (names and IDs) to set the best practice power plan.

andreasjordan:
I would like to show how I think these three commands should be used:

```
Get-DbaPowerPlan -ComputerName ABC, XYZ
Get-DbaPowerPlan -ComputerName ABC, XYZ -Credential $cred
'ABC', 'XYZ' | Get-DbaPowerPlan
'ABC', 'XYZ' | Get-DbaPowerPlan -Credential $cred
```
To get the active power plan from two computers. So just to have a look. And alos to see if I can get the information, so if I have the correct credenmials.


```
Get-DbaPowerPlan -ComputerName ABC, XYZ -List
Get-DbaPowerPlan -ComputerName ABC, XYZ -Credential $cred -List
'ABC', 'XYZ' | Get-DbaPowerPlan -List
'ABC', 'XYZ' | Get-DbaPowerPlan -Credential $cred -List
```
To get all power plans from two computers. Use this to get the info which power plans with wich names are available in my infrastructure. Helps to decide which is the in my case best power plan to use. Yes, I need the name in the next step.


```
Test-DbaPowerPlan -ComputerName ABC, XYZ
Test-DbaPowerPlan -ComputerName ABC, XYZ -Credential $cred
'ABC', 'XYZ' | Test-DbaPowerPlan
'ABC', 'XYZ' | Test-DbaPowerPlan -Credential $cred
```
Test the two computers against the default best practice power plan with the ID "8c5e7fda-e8bf-4a96-9a85-a6e23a8c635c" and the english name "High performance". Use the retuned property IsBestPractice in (pester) tests for compliance.


```
Test-DbaPowerPlan -ComputerName ABC, XYZ -CustomPowerPlan Balanced
Test-DbaPowerPlan -ComputerName ABC, XYZ -Credential $cred -CustomPowerPlan Balanced
'ABC', 'XYZ' | Test-DbaPowerPlan -CustomPowerPlan Balanced
'ABC', 'XYZ' | Test-DbaPowerPlan -Credential $cred -CustomPowerPlan Balanced
```
Test the two computers against the custom best practice power plan with the name "Balanced". Use the retuned property IsBestPractice in (pester) tests for compliance.


```
Set-DbaPowerPlan -ComputerName ABC, XYZ
Set-DbaPowerPlan -ComputerName ABC, XYZ -Credential $cred
'ABC', 'XYZ' | Set-DbaPowerPlan
'ABC', 'XYZ' | Set-DbaPowerPlan -Credential $cred
```
Set the power plan of the two computers to the default best practice power plan with the ID "8c5e7fda-e8bf-4a96-9a85-a6e23a8c635c" and the english name "High performance" if not already set. Use the returned property IsChanged if a change was performed.


```
Set-DbaPowerPlan -ComputerName ABC, XYZ -CustomPowerPlan Balanced
Set-DbaPowerPlan -ComputerName ABC, XYZ -Credential $cred -CustomPowerPlan Balanced
'ABC', 'XYZ' | Set-DbaPowerPlan -CustomPowerPlan Balanced
'ABC', 'XYZ' | Set-DbaPowerPlan -Credential $cred -CustomPowerPlan Balanced
```
Set the power plan of the two computers to the custom best practice power plan with the name "Balanced" if not already set. Use the returned property IsChanged if a change was performed.


These usages are currently also supported, but I personally see no need for them and they only add coomplexity to the code:

```
Get-DbaPowerPlan -ComputerName ABC, XYZ | Test-DbaPowerPlan
Get-DbaPowerPlan -ComputerName ABC, XYZ | Test-DbaPowerPlan -CustomPowerPlan Balanced
Get-DbaPowerPlan -ComputerName ABC, XYZ -List | Test-DbaPowerPlan
Get-DbaPowerPlan -ComputerName ABC, XYZ -List | Test-DbaPowerPlan -CustomPowerPlan Balanced
```
We were able to drastically reduce the code, if we agree to not support this type of pipeline input any more.
But it would be a breaking change...

functions/Set-DbaPowerPlan.ps1
andreasjordan: Yes, but this code is not new but only in a different place. We uses this for some months now.

tests/Set-DbaPowerPlan.Tests.ps1
potatoqualitee: Ideally, the following should be possible for reasons like grabbing the powerplan of one machine in a group and duplicating it throughout the rest of the group. `Get-DbaPowerPlan -ComputerName XYZ -Credential ad\blah | Set-DbaPowerPlan -ComputerName ABC, DEF, HIJ` (with the set using the same credential as get-dbapowerplan, probably.
 
`Test-DbaPowerPlan -ComputerName sql2012 | Set-DbaPowerPlan` is currently supported and breaking changes should be introduced only when required. Before we understood PowerShell as well as we do now, we used Test instead of Get so a lot of our presentations refer to it. I'd like to keep major breaking changes for dbatools 2.0.

Would ValueFromPipelineByPropertyName better suit your needs?

functions/Test-DbaPowerPlan.ps1
potatoqualitee: Piping in the ComputerName from Get-DbaPowerPlan should replace the default so it is not required to remove it. 

We can remove InputObject if the functionality does not change. If I could go back in time, I'd probably just use the heck out of `ValueFromPipelineByPropertyName`.

functions/Set-DbaPowerPlan.ps1
potatoqualitee: I probably said the same thing last time I saw it 😅

functions/Set-DbaPowerPlan.ps1
potatoqualitee: I do love that functionality! I'm still confused as to why `Custom` is in there. Perhaps we just swap and make CustomPowerPlan the alias?

andreasjordan: Yes, no problem. But we should have the same name in Test- and Set-.

functions/Test-DbaPowerPlan.ps1
andreasjordan: No problem with piping in ComputerName. But with piping in from Get- into InputObject sets the default to ComputerName and so not ony the computers from the pipeline are tested, but also the local computer. That's the reason for removing the default.

tests/Set-DbaPowerPlan.Tests.ps1
andreasjordan: So you want to use `Get-DbaPowerPlan -ComputerName XYZ -Credential ad\blah | Set-DbaPowerPlan -ComputerName ABC, DEF, HIJ` instead of `Set-DbaPowerPlan -ComputerName ABC, DEF, HIJ -Credential ad\blah -PowerPlan (Get-DbaPowerPlan -ComputerName XYZ -Credential ad\blah).PowerPlan`?

andreasjordan: But `Get-DbaPowerPlan` also returns ComputerName - and that should not be used from the pipeline. So we have to test what comes over the pipeline and build solid rules what to use. Not that easy...

functions/Set-DbaPowerPlan.ps1
andreasjordan: Ok, done that.

functions/Test-DbaPowerPlan.ps1
andreasjordan: Can you maybe add a comment down below with all the possible piping ways that should work? Then I have a todo list and can maybe create small mock commands for all three to show just the implementation of parameter processing.

potatoqualitee: it would be 

```
        [parameter(ValueFromPipelineByPropertyName)]
        [DbaInstance[]]$ComputerName,
        [parameter(ValueFromPipelineByPropertyName)]
        [PSCredential]$Credential,
        [parameter(ValueFromPipelineByPropertyName)]
        [Alias("CustomPowerPlan")]
        [string]$PowerPlan,
        [switch]$EnableException
```

Then you can remove the InputObject!

andreasjordan: Not that easy:
image

andreasjordan: This works: 
```
[parameter(ValueFromPipelineByPropertyName, ValueFromPipeline)]
[DbaInstance[]]$ComputerName,
```

andreasjordan:
Updated pipeline input of Test-DbaPowerPlan on suggestion of @potatoqualitee 
image

functions/Test-DbaPowerPlan.ps1
andreasjordan: Implemented that, see comment down below.

tests/Set-DbaPowerPlan.Tests.ps1
andreasjordan: If you like my implementation on Test-DbaPowerPlan, then I can try to use the same way with Set-DbaPowerPlan.

andreasjordan:
Changed Set-DbaPowerPlan as well:
image


Piping in from Test-DbaPowerPlan works as well, but I don't show an example as it should not be used. Why? Because Set-DbaPowerPlan is running Test-DbaPowerPlan itself and with a piped input from Test-DbaPowerPlan this would run twice.
But it works:
image

tests/Set-DbaPowerPlan.Tests.ps1
andreasjordan: Now everything you mentioned here works, so I will click on Resolve conversation...

potatoqualitee:
Thank you! Can you please catch this error?

image

Going through the examples now

potatoqualitee:
This may be leftovers, but can you change `Specified credentials are known to not work!` to "Login failure for $(Credential.UwserName)"

potatoqualitee:
k the rest works!

andreasjordan:
No with test on instanceId if power plan is not known on computer and added instance ids to output, but not in default view:

image

andreasjordan:
> This may be leftovers, but can you change `Specified credentials are known to not work!` to "Login failure for $(Credential.UserName)" it may be in another command

The message "Specified credentials are known to not work!" is coded here:
https://github.com/sqlcollaborative/dbatools/blob/bc84174d9c3ecef92add9c2e14835d28d4ea1e84/bin/projects/dbatools/dbatools/Connection/ManagementConnection.cs#L494

This is where it comes from insinde of the PowerPlan commands:
https://github.com/sqlcollaborative/dbatools/blob/bc84174d9c3ecef92add9c2e14835d28d4ea1e84/functions/Get-DbaPowerPlan.ps1#L72-L80

andreasjordan:
But let my try something...

andreasjordan:
Works:
image

andreasjordan:
This is the message with faild windows auth:

image

What do you suggest here? Or do we leave this as it is?

potatoqualitee:
yeah, that takes a recompile which requires changes to like 3 files :/ this will work, thank you, merging 🥳

PR #7472, opened 2021-06-29
Test-DbaConnectionAuthScheme.ps1 flag fix

slimrich:
-ntlm and -kerberos flags giving incorrect results. The if statement that returns a true/false is checking against $server.AuthScheme instead of $results.AuthScheme so is a simple fix.

## Type of Change
 - [ x ] Bug fix (non-breaking change)

### Purpose
Fix output if -ntlm or -kerberos used


andreasjordan (approved):
Good catch!

potatoqualitee:
Fantastic! Thank you so much, @slimrich. This will be released to the PowerShell Gallery within 0 to 48 hours 😅

potatoqualitee:
Released!

PR #7475, opened 2021-06-29
Repair-DbaInstanceName - Corrected test for warnings from Test-DbaInstanceName

### LINKED ISSUE #7337: [Bug] Repair-DbaInstanceName - test for Warnings not correct
Test-DbaInstanceName fills the property Warnings with "N/A" if there are no warnings:

```
            if ($null -ne $rs -or $rs.Count -gt 0) {
                if ($rs.State -eq 'Running') {
                    $rstext = "$ssrsService must be stopped and updated."
                } else {
                    $rstext = "$ssrsService exists. When it is started again, it must be updated."
                }
                $serverInfo.Warnings = $rstext
            } else {
                $serverInfo.Warnings = "N/A"
            }
```

But Repair-DbaInstanceName tests for empty string or $null:

```
            if ($nametest.Warnings.length -gt 0) {
```

So: Do we want to change Test-DbaInstanceName to send $null or do we want to change Repair-DbaInstanceName to test for "N/A"?

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #7337 )

### DIFF
--- a/functions/Repair-DbaInstanceName.ps1
+++ b/functions/Repair-DbaInstanceName.ps1
@@ -189,7 +189,7 @@ function Repair-DbaInstanceName {
                 Write-Message -Level Warning -Message "Can't contact $instance using Get-Service. This means the script will not be able to automatically restart SQL services."
             }
 
-            if ($nametest.Warnings.length -gt 0) {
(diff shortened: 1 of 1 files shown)


potatoqualitee:
great!

PR #7476, opened 2021-07-01
Get-DbaService - Rewritten command to have better error handling and messages

### LINKED ISSUE #7470: [Bug] Get-DbaService - false error no services found in relevant namespaces
#### Host used

- [x] powershell.exe
- [ ] ISE
- [ ] VS Code
- [ ] Other (please specify)

#### Errors Received

```
 No services found in relevant namespaces on <server name>. Please note that this function is available from SQL 2005 up.
```

#### Steps to Reproduce

1. Run `Get-DbaService` either in the context of an account or passing Credential
2. Ensure the account or credential used does not have valid permissions to the target endpoint

#### Expected Behavior

Provide more clear message of why services were not found.

#### Actual Behavior

We claim no services were found but in reality, the logic in the command is flawed.

We create collection variables:

https://github.com/sqlcollaborative/dbatools/blob/eb515c5ef82354b28ece0b3eaa2e72ff46e05b42/functions/Get-DbaService.ps1#L153-L154

But if the namespaces do not actually have anything we still continue processing an empty variable:

https://github.com/sqlcollaborative/dbatools/blob/eb515c5ef82354b28ece0b3eaa2e72ff46e05b42/functions/Get-DbaService.ps1#L183

Which we immediately jump down to the error message since `$services` is empty:

https://github.com/sqlcollaborative/dbatools/blob/eb515c5ef82354b28ece0b3eaa2e72ff46e05b42/functions/Get-DbaService.ps1#L261-L265

### Fix

This section of code from what I can tell is where the issue arises:

https://github.com/sqlcollaborative/dbatools/blob/eb515c5ef82354b28ece0b3eaa2e72ff46e05b42/functions/Get-DbaService.ps1#L166-L184

The foreach loop for `$services` needs to be within the `if ($namespaces)` because that is the only way we have service(s) to work on.  Everything down to line 256 should be in that `if` block, and then an `else` be used to output a cleaner message such as:

```powershell
} else {
    Stop-Function -Message "Unable to capture namespaces on $Computer. Please verify the account used has proper permissions to query for SQL Server services."
}
```

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #7470 )

See issue for details...

As with most big rewrites, diff is not very helpfull - better have a look at the target source code.

### DIFF
--- a/functions/Get-DbaService.ps1
+++ b/functions/Get-DbaService.ps1
@@ -169,7 +169,7 @@ function Get-DbaService {
                 $namespaces = Get-DbaCmObject -ComputerName $resolvedComputerName -Credential $Credential -Namespace root\Microsoft\SQLServer -Query "Select Name FROM __NAMESPACE WHERE Name Like 'ComputerManagement%'" -EnableException | Sort-Object Name -Descending
                 Write-Message -Level Verbose -Message "The following namespaces have been found: $($namespaces.Name -join ', ')."
             } catch {
-                Write-Message -Level Verbose -Message "No namespaces found in relevant namespace on $computer. Please note that this function is available from SQL 2005 up."
+                Write-Message -Level Verbose -Message "No namespaces found in relevant namespace on $computer."
             }
 
             foreach ($namespace in $namespaces) {
@@ -269,7 +269,7 @@ function Get-DbaService {
             if ($outputServices) {
                 $outputServices | Select-DefaultView -Property $defaults -TypeName DbaSqlService
             } else {
-                Write-Message -Level Verbose -Message "No services found in relevant namespaces on $computer. Please note that this function is available from SQL 2005 up."
+                Write-Message -Level Verbose -Message "No services found in relevant namespaces on $computer."
             }
         }
     }
--- a/functions/Get-DbaService.ps1
+++ b/functions/Get-DbaService.ps1
@@ -146,125 +146,130 @@ function Get-DbaService {
         }
     }
     process {
-        foreach ($Computer in $ComputerName.ComputerName) {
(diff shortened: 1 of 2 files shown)


potatoqualitee:
@nvarscar can you please take a look and approve? 🙇🏼  I'm really excited about this fix, as I encountered it as well.

potatoqualitee:
@wsmelton too if you're around

functions/Get-DbaService.ps1
```diff
@@ -146,125 +146,130 @@ function Get-DbaService {
...
+            if ($AdvancedProperties) {
+                $defaults = "ComputerName", "ServiceName", "ServiceType", "InstanceName", "DisplayName", "StartName", "State", "StartMode", "Version", "SPLevel", "SkuName", "Clustered", "VSName"
+            } else {
+                $defaults = "ComputerName", "ServiceName", "ServiceType", "InstanceName", "DisplayName", "StartName", "State", "StartMode"
+            }
+            if ($outputServices) {
+                $outputServices | Select-DefaultView -Property $defaults -TypeName DbaSqlService
             } else {
-                Stop-Function -EnableException $EnableException -Message "Failed to connect to $Computer" -Continue
+                Write-Message -Level Verbose -Message "No services found in relevant namespaces on $computer. Please note that this function is available from SQL 2005 up."
```
wsmelton, line 272: I believe we can drop the later part of this message now. It is not really relevant that I see.

andreasjordan: Ok, will do.

potatoqualitee:
One review plus I just tested it and it works! Merging 🥳

PR #7482, opened 2021-07-01
Add-DbaAgDatabase - Add missing SqlCredential when calling Test-DbaAvailabilityGroup

### LINKED ISSUE #7481: [Bug]Add-DbaAgDatabase does not honor SqlCredential
Add-DbaAgDatabase was recently re-written and the re-write does not honor the passed in SqlCredential parameter.  
https://github.com/sqlcollaborative/dbatools/commit/c864b791ad360d954a93edc065c691e117714a2e

When Add-DbaAgDatabase calls Test-DbaAvailabilityGroup, it should pass along the SqlCredential parameter if one was passed into Add-DbaAgDatabase at runtime.

This change breaks automation scenarios where dba tools is running under an account that does not have access to SQL.

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #7481 )

### DIFF
--- a/functions/Add-DbaAgDatabase.ps1
+++ b/functions/Add-DbaAgDatabase.ps1
@@ -193,6 +193,7 @@ function Add-DbaAgDatabase {
                 Write-Progress @progress
                 $testSplat = @{
                     SqlInstance            = $SqlInstance
+                    SqlCredential          = $SqlCredential
                     Secondary              = $Secondary
                     SecondarySqlCredential = $SecondarySqlCredential
                     AvailabilityGroup      = $AvailabilityGroup


potatoqualitee:
Thank you!

cmdub:
@potatoqualitee it looks like this still hasn't been merged to main/master yet. I just came across this bug on my end. Is there an ETA on when this would get merged?

andreasjordan:
Ups, looks like it was only added for one $testsplat, but not the other a feq lines below. Will open a new PR for that.

andreasjordan:
Ok, run a test on two instances where I can only work with sa:

```
$saCred = Get-Credential -Message sa -UserName sa
$db = Get-DbaDatabase -SqlInstance sql01 -SqlCredential $saCred -Database Test01
# This fails, because the secondary is connected by integrated security:
$db | Add-DbaAgDatabase -AvailabilityGroup AdventureSQL 
# This works:
$db | Add-DbaAgDatabase -AvailabilityGroup AdventureSQL -Secondary sql02 -SecondarySqlCredential $saCred
```

So I need to know your exact command to help you.

cmdub:
Hi @andreasjordan  sorry for the late response.

I thought initially this might be due to me running a slightly older version 1.1.3X but this issue still persists even on 1.1.5

Command I'm running:

```
Add-DbaAgDatabase -SqlInstance $sqlServer `
    -Database $databaseName `
    -SqlCredential $sqlCredential `
    -AvailabilityGroup $sqlAg `
    -Confirm:$false
```

- $sqlServer resolves to the string of the listener FQDN
- $databaseName resolves to the string name of the DB in question
- $sqlCredential is a PSCredential object which is used elsewhere in my script and works as expected

Give me a few and I'll add the debug output

andreasjordan:
Please use this: 
```
Add-DbaAgDatabase -SqlInstance $sqlServer `
    -Database $databaseName `
    -SqlCredential $sqlCredential `
    -SecondarySqlCredential $sqlCredential `
    -AvailabilityGroup $sqlAg `
    -Confirm:$false
```

PR #7485, opened 2021-07-02
Rename-DbaLogin - Check if user exists before attempting to rename it, require Force to rename db users too

### LINKED ISSUE #7363: [Bug] - Rename-DbaLogin won't rename logins not mapped inside a db
#### Host used

- [X] powershell.exe
- [ ] ISE
- [X] VS Code
- [ ] Other (please specify)

#### Errors Received

```
writeErrorStream      : True
PSMessageDetails      : 
Exception             : System.Exception: You cannot call a method on a null-valued expression. 
                        ---> System.Management.Automation.RuntimeException: You cannot call a 
                        method on a null-valued expression.
                           at System.Management.Automation.ExceptionHandlingOps.CheckActionPreferenc
                        e(FunctionContext funcContext, Exception exception)
                           at System.Management.Automation.Interpreter.ActionCallInstruction`2.Run(I
                        nterpretedFrame frame)
                           at System.Management.Automation.Interpreter.EnterTryCatchFinallyInstructi
                        on.Run(InterpretedFrame frame)
                           at System.Management.Automation.Interpreter.EnterTryCatchFinallyInstructi
                        on.Run(InterpretedFrame frame)
                           --- End of inner exception stack trace ---
TargetObject          : newlogin
CategoryInfo          : InvalidOperation: (hooray:String) [Write-Error], Exception
FullyQualifiedErrorId : dbatools_Rename-DbaLogin,Stop-Function
ErrorDetails          : You cannot call a method on a null-valued expression.
InvocationInfo        : System.Management.Automation.InvocationInfo
ScriptStackTrace      : at Stop-Function, C:\Program 
                        Files\WindowsPowerShell\Modules\dbatools\1.0.149\allcommands.ps1: line 88702
                        at Rename-DbaLogin<Process>, C:\Program 
                        Files\WindowsPowerShell\Modules\dbatools\1.0.149\allcommands.ps1: line 63732
                        at <ScriptBlock>, <No file>: line 1
PipelineIterationInfo : {0, 1}
```

#### Steps to Reproduce
Rename_DbaLogin -SQLInstance dbatools2 -Login sa -NewLogin newlogin

#### Expected Behavior
sa account to be renamed successfully without the need to remapped logins in the databases.

#### Actual Behavior
Failure because it tries to remap the users in the db but find no users in the db

### Environmental information

🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨

```powershell
[pscustomobject]@{
'PowerShell Version' = $PSVersionTable.PSVersion.ToString()
'dbatools latest installed' = (Get-InstalledModule -Name dbatools).Version
'Culture of OS' = (Get-Culture)
} | fl -force
```
PowerShell Version        : 5.1.17763.1852
dbatools latest installed : 1.0.149

#### SQL Server:
Microsoft SQL Server 2019 (RTM-CU10) (KB5001090) - 15.0.4123.1 (X64)   Mar 22 2021 18:10:24   Copyright (C) 2019 Microsoft Corporation  Developer Edition (64-bit) on Windows Server 2019 Datacenter 10.0 <X64> (Build 17763: ) (Hypervisor) 

us_english

potatoqualitee:
## Type of Change
 - [x] Bug fix (non-breaking change, fixes #7363 )
 
It was assumed the db user would exist since it came from an eum, so we didn't double-check.

I updated the docs to better align with the fix and I also added a force switch to also change associated db users and a check to see if the user indeed exists.

Since Force is now required to rename database users, this could be a breaking change if it worked before. I think this is safer, so if there is an issue with adding force, I'll at least add a -NoDbRename option so that people have the option to not rename db users too.

### DIFF
--- a/functions/Rename-DbaLogin.ps1
+++ b/functions/Rename-DbaLogin.ps1
@@ -102,7 +102,7 @@ function Rename-DbaLogin {
                 $output = @()
                 try {
                     $dbenums = $currentLogin.EnumDatabaseMappings()
-                    $null = $currentLogin.rename($NewLogin)
+                    $null = $currentLogin.Rename($NewLogin)
                     $output += [pscustomobject]@{
                         ComputerName  = $server.ComputerName
                         InstanceName  = $server.ServiceName
@@ -110,6 +110,8 @@ function Rename-DbaLogin {
                         Database      = $null
                         PreviousLogin = $Login
                         NewLogin      = $NewLogin
+                        PreviousUser  = $null
+                        NewUser       = $null
                         Status        = "Successful"
                     }
                 } catch {
@@ -121,6 +123,8 @@ function Rename-DbaLogin {
                         Database      = $null
                         PreviousLogin = $Login
                         NewLogin      = $NewLogin
+                        PreviousUser  = $null
+                        NewUser       = $null
                         Status        = "Failure"
                     }
                     Stop-Function -Message "Failure" -ErrorRecord $_ -Target $login -Continue
@@ -139,28 +143,33 @@ function Rename-DbaLogin {
                                 $oldname = $user.name
                                 $null = $user.Rename($NewLogin)
                                 $output += [pscustomobject]@{
-                                    ComputerName = $server.ComputerName
-                                    InstanceName = $server.ServiceName
-                                    SqlInstance  = $server.DomainInstanceName
-                                    Database     = $db.name
-                                    PreviousUser = $oldname
-                                    NewUser      = $NewLogin
-                                    Status       = "Successful"
+                                    ComputerName  = $server.ComputerName
+                                    InstanceName  = $server.ServiceName
+                                    SqlInstance   = $server.DomainInstanceName
+                                    Database      = $db.name
+                                    PreviousLogin = $null
+                                    NewLogin      = $null
+                                    PreviousUser  = $oldname
+                                    NewUser       = $NewLogin
+                                    Status        = "Successful"
                                 }
                             } catch {
                                 Write-Message -Level Warning -Message "Rolling back update to login: $Login"
-                                $currentLogin.rename($Login)
+                                $null = $currentLogin.Rename($Login)
 
                                 [pscustomobject]@{
-                                    ComputerName = $server.ComputerName
-                                    InstanceName = $server.ServiceName
-                                    SqlInstance  = $server.DomainInstanceName
-                                    Database     = $db.name
-                                    PreviousUser = $NewLogin
-                                    NewUser      = $oldname
-                                    Status       = "Failure to rename. Rolled back change."
+                                    ComputerName  = $server.ComputerName
+                                    InstanceName  = $server.ServiceName
+                                    SqlInstance   = $server.DomainInstanceName
+                                    Database      = $db.name
+                                    PreviousLogin = $null
+                                    NewLogin      = $null
+                                    PreviousUser  = $NewLogin
+                                    NewUser       = $oldname
+                                    Status        = "Failure to rename. Rolled back change."
                                 }
-                                Stop-Function -Message "Failure" -ErrorRecord $_ -Target $NewLogin -Continue
+                                Stop-Function -Message "Failure" -ErrorRecord $_ -Target $NewLogin
+                                return
                             }
                         }
                     }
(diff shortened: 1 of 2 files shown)


functions/Rename-DbaLogin.ps1
```diff
@@ -120,39 +126,43 @@ function Rename-DbaLogin {
...
+                                    SqlInstance  = $server.DomainInstanceName
+                                    Database     = $db.name
+                                    PreviousUser = $oldname
+                                    NewUser      = $NewLogin
+                                    Status       = "Successful"
+                                }
+
+                            } catch {
+                                Write-Message -Level Warning -Message "Rolling back update to login: $Login"
+                                $currentLogin.rename($Login)
```
andreasjordan, line 152: Do you really want to rollback the login rename? In line 105 there was already an output stating "Successful". So better wait with the output untill all changes are made.

potatoqualitee: agreed, thanks, will do

potatoqualitee: done! plz review

andreasjordan: don't know if `$currentLogin.rename($Login)` returns something, but in line 140 you used `$null = `. (And it's "Rename" wirh "R".)

functions/Rename-DbaLogin.ps1
```diff
@@ -116,46 +123,51 @@ function Rename-DbaLogin {
...
+                                [pscustomobject]@{
+                                    ComputerName = $server.ComputerName
+                                    InstanceName = $server.ServiceName
+                                    SqlInstance  = $server.DomainInstanceName
+                                    Database     = $db.name
+                                    PreviousUser = $NewLogin
+                                    NewUser      = $oldname
+                                    Status       = "Failure to rename. Rolled back change."
+                                }
+                                Stop-Function -Message "Failure" -ErrorRecord $_ -Target $NewLogin -Continue
```
andreasjordan, line 163: This `-Continue` will not leave the `foreach ($db in $dbenums) {` but go on with the next database user. You will have to use a `break` to leave the loop and continue with the next $instance.

andreasjordan:
If you use `-Force` you will have objects with different properties in the output. Have you testet what this looks like if you pipe the command to Format-Table?

functions/Rename-DbaLogin.ps1
potatoqualitee: good call. i was told by Fred not to use break anywhere as it breaks someone entire pipeline, so we use return instead.

functions/Rename-DbaLogin.ps1
potatoqualitee: fixed to make it more uniform!

functions/Rename-DbaLogin.ps1
andreasjordan: oh yes, return is best here.

potatoqualitee:
will merge once approved 👍🏼

andreasjordan (approved):
Looks good!

potatoqualitee:
🥳  thanks for the review!

functions/Rename-DbaLogin.ps1
wsmelton: You don't want to return in a foreach loop, that breaks the whole loop. Continue is correct method as it moves onto the next database in the loop.

wsmelton: Processing side if we fail in one database for remapping a user it is a matter if you want it to fail for all of them or keep trying.

As it stands this will tell it to keep trying, but the rest will fail if we are going to roll back the login rename. This may cause a number of random errors. Only reason the catch could get hit is if something changed after 104 when we got the `$dbenums`. The other issue would be permissions, the user had rights to rename the login but not touch any databases (e.g. giving [securityadmin ](https://docs.microsoft.com/en-us/sql/relational-databases/security/authentication-access/server-level-roles?view=sql-server-ver15#fixed-server-level-roles) but denied access to databases)

andreasjordan: Ok, I was not correct with the "return is best" as it ends the command even if the failure is in the first of more instances to process. But "continue" on the other hand would rename other database users, while the login rename is rolled back.
Suggestion: Forget about rolling back (as you don't roll back the already renamed database users anyway) and just try to rename the user in every database and report the failed ones. So the gereral rule is (I think it's the same in other commands): We do as much as we can, and report if we failed - so the DBA has to do the rest.

PR #7505, opened 2021-07-07
Invoke-Command2.ps1 internal command - fix repeated config call

slimrich:
The usessl input is ignored so changed the splat to use the input rather than re-pulling from config

## Type of Change
 - [ x ] Bug fix (non-breaking change, fixes # )

### Purpose
simple change to use input parameter instead of ignoring it

### DIFF
--- a/internal/functions/Invoke-Command2.ps1
+++ b/internal/functions/Invoke-Command2.ps1
@@ -92,7 +92,7 @@ function Invoke-Command2 {
                 Authentication = $Authentication
                 Name           = $sessionName
                 ErrorAction    = 'Stop'
-                UseSSL         = (Get-DbatoolsConfigValue -FullName 'PSRemoting.PsSession.UseSSL' -Fallback $false)
+                UseSSL         = $UseSSL
             }
             if (Test-Windows -NoWarn) {
                 $psSessionOptionsSplat = @{


potatoqualitee:
Thanks, slim! This change makes sense as, like you said, it's already pulling from config the first time around. I'll merge once tests pass 👍🏼

internal/functions/Invoke-Command2.ps1
```diff
@@ -92,7 +92,7 @@ function Invoke-Command2 {
                 Authentication = $Authentication
                 Name           = $sessionName
                 ErrorAction    = 'Stop'
-                UseSSL         = (Get-DbatoolsConfigValue -FullName 'PSRemoting.PsSession.UseSSL' -Fallback $false)
+                UseSSL         = $UseSSL
```
wsmelton, line 95: The parameter should be removed if we are not going to need `-UseSSL` on the function anymore.

slimrich: Yeah that's the other approach, only issue would be if config has $true but you want to override to $false. Honestly can't think of why you would do that though.

wsmelton: The parameter no longer does anything now with the config being set here.

PR #7511, opened 2021-07-07
Connect-DbaInstance - Add back support for Tenant

potatoqualitee:
## Type of Change
 
It seems like the new SMO libraries or .NET updates broke the way we connect using Tenants. The new experimental path is the easiest way to connect, so we will default to it.

image


potatoqualitee:
For future reference, this executes the following under the hood:

```
$token = (New-DbaAzAccessToken -Type RenewableServicePrincipal -Subtype AzureSqlDb -Tenant $tenant_id -Credential $cred).GetAccessToken()
```

wsmelton (approved):

potatoqualitee:
Thanks, Shawn!

PR #7516, opened 2021-07-08
Set-DbaPrivilege - Don't write a warning if privilege is already set

### LINKED ISSUE #7232: Enable "-WarningAction SilentlyContinue" for commands that work remote like Set-DbaPrivilege
I would like to call `Set-DbaPrivilege -ComputerName $computerName -Type IFI -WarningAction SilentlyContinue` to suppress the warning "WARNING: [...] already has Instant File Initialization Privilege on [...]".

The command uses Invoke-Command2 and inside the scriptblock there are lines with `Write-Warning` - so there might not be an easy solution.

Is there a general interest in this feature?

andreasjordan:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality, fixes #7232 )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose

When installing a second instance on a system under the same service account, and thereby setting the instant file initialization privilege again, I don't what to see a warning message, that the privilege was already set.
But on the over hand I think it's worth to have a verbose message in the else path of these ifs for debugging.

### DIFF
--- a/functions/Set-DbaPrivilege.ps1
+++ b/functions/Set-DbaPrivilege.ps1
@@ -123,7 +123,7 @@ function Convert-UserNameToSID ([string] `$Acc ) {
                                             Write-Verbose "Added $acc to Batch Logon Privileges on $env:ComputerName"
                                         } else {
                                             <# DO NOT use Write-Message as this is inside of a script block #>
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Reasonable. We waffled about this in the pat and i think this is a better approach

Issue #6896, opened 2020-10-01
[Bug] Connect-DbaInstance assumes MFA method when azure.tenantid is in dbatoolsconfig


PowerDBAKlaas:
#### Host used

- [x] powershell.exe
- [ ] ISE
- [ ] VS Code
- [ ] Other (please specify)

#### Errors Received

```
PSMessageDetails      :
Exception             : System.Exception: When using AD Universal with MFA Support, database must be specified.
TargetObject          :
CategoryInfo          : NotSpecified: (:) [], Exception
FullyQualifiedErrorId : dbatools_Connect-DbaInstance
ErrorDetails          :
InvocationInfo        : System.Management.Automation.InvocationInfo
ScriptStackTrace      : at Stop-Function, C:\Git\dbatools\internal\functions\flowcontrol\Stop-Function.ps1: line 221
                        at Connect-DbaInstance<Process>, C:\Git\dbatools\functions\Connect-DbaInstance.ps1: line 498
                        at <ScriptBlock>, <No file>: line 1
PipelineIterationInfo : {}
```

#### Steps to Reproduce
A long time ago:
`Set-DbatoolsConfig -FullName azure.tenantid -Value 'guidheremaybename' -Passthru | Register-DbatoolsConfig`  
  
Now:  
`Connect-DbaInstance -ConnectionString 'Data Source=TCP:uheresmyserverc.database.windows.net,1433;Initial Catalog=SomeDB;User Id=SomeWriterUser;Password=I**************************f; Persist Security Info=True;MultipleActiveResultSets=False;Connect Timeout=30;Encrypt=True;TrustServerCertificate=False;Application Name=dbatools PowerShell module -dbatools.io' -Debug`  
  
and:  
`Connect-DbaInstance -SqlInstance 'uheresmyserverc.database.windows.net' -DisableException -Database 'SomeDB' -SqlCredential $SomeWriterUser -Debug`
  
and:  
`Connect-DbaInstance -SqlInstance 'uheresmyserverc.database.windows.net' -DisableException -Database 'SomeDB' -SqlCredential $ITWriter -Verbose -AppendConnectionString 'User ID=SomeWriterUser;Password =I**************************f;'`  
  
#### Expected Behavior

![image](https://user-images.githubusercontent.com/22813386/94821041-6de2f180-0401-11eb-9129-bae7b36147c1.png)

#### Actual Behavior

[...]

PR #7520, opened 2021-07-10
Connect-DbaInstance - Test for GUID username when using Tenant

### LINKED ISSUE #7518: Connect-DbaInstance - Use SqlCredential to login as a service principal
Starting with 1.0.170, it is again possible to connect to an Azure SQL Database with a service principal.
At the moment, the code uses `-Tenant` to get the tenant id and `-SqlCredential` to get the service principal id and the secret.
When using `Connect-DbaInstance` directly, it is no problem to add `-Tenant`, but using other commands that use `Connect-DbaInstance` internally, they only use `-SqlInstance` and `-SqlCredential`. Then the config `azure.tenantid` has to be used because it is the default for `-Tenant`. This is maybe not the best way, because the config name looks like a general information and not a default for all connections - see #6896.

What about storing the tenant id and the service principal id together as the `UserName` of the credential? They are both guids and we can use `|` to separate them.
We already check the username for `\\` and `@` to find out if it is an active directory account. So we could also check for `|` to know that we have to get a token from Azure using `New-DbaAzAccessToken`.

So an example code would look like:
```
$tenantId = "3f03bfac-9d26-4690-8cda-783e205cb9a3" 
$clientId = "d58773f3-0acf-4760-83d3-7616a5ed9881"
$clientSecret = ConvertTo-SecureString "PutTheSecretHere" -AsPlainText -Force
$azureCredential = [PSCredential]::new("$tenantId|$clientId", $clientSecret)
Invoke-DbaQuery -SqlInstance myAzureDatabaseServer.database.windows.net -SqlCredential $azureCredential -Database myAzureDatabase -Query "SELECT foo FROM bar"
```

Feedback is welcome.

potatoqualitee:
Add in GUID check to invoke when azure.tenantid is in dbatoolsconfig or otherwise specificed

Fixes #7518 and #6896

I _could_ add a little more safety and say `(Get-DbatoolsConfigValue -FullName 'azure.tenantid' -and $SqlCredential.UserName -match '^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$')` but I'm not sure it's required

I tested this with the command line against username `beb` and with an azure client ID and it passed 👍🏼  Good work GitHub Copilot!

### DIFF
--- a/functions/Connect-DbaInstance.ps1
+++ b/functions/Connect-DbaInstance.ps1
@@ -448,6 +448,7 @@ function Connect-DbaInstance {
     }
     process {
         if (Test-FunctionInterrupt) { return }
+        # if tenant is specified with a GUID username such as 21f5633f-6776-4bab-b878-bbd5e3e5ed72 (for clientid)
         if ($Tenant -and -not $AccessToken -and $SqlCredential.UserName -match '^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$') {
             Write-Message -Level Verbose "Tenant detected, switching to experimental code path"
             try {
--- a/functions/Connect-DbaInstance.ps1
+++ b/functions/Connect-DbaInstance.ps1
@@ -448,7 +448,7 @@ function Connect-DbaInstance {
     }
     process {
         if (Test-FunctionInterrupt) { return }
-        if ($Tenant -and -not $AccessToken) {
+        if ($Tenant -and -not $AccessToken -and $SqlCredential.UserName -match '^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$') {
             Write-Message -Level Verbose "Tenant detected, switching to experimental code path"
             try {
                 Set-DbatoolsConfig -FullName sql.connection.experimental -Value $true


andreasjordan:
But Copilot failed in generating the dbatools style catch block:
```
$errormessage = Get-ErrorMessage -Record $_
Stop-Function -Message "Failed to get access token for Azure SQL DB ($errormessage)"
```
Should include `-ErrorRecord $_` and does not need `$errormessage`.

potatoqualitee:
haha that was me. Because Connect-DbaInstance throws exceptions by default, using `-ErrorRecord` produces an unhelpful error, as it does not look at `-Message` and just throws the exception.

andreasjordan:
> haha that was me. Because Connect-DbaInstance throws exceptions by default, using `-ErrorRecord` produces an unhelpful error, as it does not look at `-Message` and just throws the exception which says something like "Error 400"

Oh, forgot that. Maybe I should have a look at the catch blocks in the new code path...

potatoqualitee:
Most of the SQL Server error messages are helpful, unlike this token one :/

functions/Connect-DbaInstance.ps1
```diff
@@ -448,7 +448,7 @@ function Connect-DbaInstance {
     }
     process {
         if (Test-FunctionInterrupt) { return }
-        if ($Tenant -and -not $AccessToken) {
+        if ($Tenant -and -not $AccessToken -and $SqlCredential.UserName -match '^[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}$') {
```
wsmelton, line 451: It may help to add a comment (block maybe) of one example  or two for what that regex is supposed to catch.

6 months down I seriously doubt we will remember 😆

potatoqualitee: merci, added a comment 😊

potatoqualitee:
k merging ⛅

PR #7521, opened 2021-07-11
Convert issue templates to Issue Forms (beta)

wsmelton:
This is to convert our current issue templates from markdown to [issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms) that is currently in beta.

These exists on a public repository so you can can visually see and interact with them: [wsmelton/testing-issue-templates](https://github.com/wsmelton/testing-issue-templates)

A sampling has been created as well: [wsmelton/testing-issue-templates/issues](https://github.com/wsmelton/testing-issue-templates/issues).

**A few bugs in it as beta**

I have found that even though adding a "header" in the form will display it, it does not persist to the issue once the user submits it. I have filled feedback on that along with checkboxes even on the forms still end up showing the "tasks done" visualization under the title of the issue (😞).

**Feedback on the PR**

Recommend or suggest adjustments and I will add them on the public repo so you can see the change **prior to updating the PR**.

**Details**
The main features in forms is the ability to use drop-downs and checkboxes. Drop-down allows to provide a list of selections:

image

We can have "comments" on the template more clearly added and that content not misunderstood by users:

image

Content requested can be required, where the submitter has to be some text in the text area or select an option from a drop-down:

image

You can also pre-populate a text area with content example as well:

image


potatoqualitee:
I think this is beautiful 😍

potatoqualitee:
@jpomfret - if you're around, whatcha think? I'm ready when someone else is.

potatoqualitee:
lez do this

PR #7522, opened 2021-07-11
Fix multiple commands to restore compatibility for PSv3 - remove unsupported method `::new(`

### LINKED ISSUE #7513: [Bug] Need to remove instances of ::new()
Hopefully this can be replaced with `ArgumentList`
image

potatoqualitee:
## Type of Change
 - [X] Bug fix (non-breaking change, fixes #7513)

Please triple check for me, this is easy to mess up

I did not remove from any tests because all of our testing platforms are newer


potatoqualitee:
K I tested each of these individually

PR #7523, opened 2021-07-11
Copy-DbaLogin - Update docs to be more accurate

### LINKED ISSUE #7506: [Bug] Copy-DbaLogin - KillActiveConnection parameter help incorrect
### Report

Unless I'm reading this wrong, we state that `-Force` has to be used with `KillActiveConnection`:

https://github.com/sqlcollaborative/dbatools/blob/08b9fbafbd749be5d1ae5827cb26c490f21d4e9c/functions/Copy-DbaLogin.ps1#L63-L64

Checking where that parameter hits though we are not checking if `-Force` was provided.

https://github.com/sqlcollaborative/dbatools/blob/08b9fbafbd749be5d1ae5827cb26c490f21d4e9c/functions/Copy-DbaLogin.ps1#L321-L331

Notice in the verbose message if active connections are found we also state the use of `-Force`.

I do not really see a need to tie this parameter to Force, so believe we can update the documentation. // @potatoqualitee ?

potatoqualitee:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #7506)

### DIFF
--- a/functions/Copy-DbaLogin.ps1
+++ b/functions/Copy-DbaLogin.ps1
@@ -61,9 +61,7 @@ function Copy-DbaLogin {
         Include object-level permissions for each user associated with copied login.
 
     .PARAMETER KillActiveConnection
-        If this switch and -Force are enabled, all active connections and sessions on Destination will be killed.
-
-        A login cannot be dropped when it has active connections on the instance.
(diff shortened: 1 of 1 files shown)


potatoqualitee:
just docs, so merge

PR #7531, opened 2021-07-12
Save-DbaKbUpdate - Using Architecture and some other fixes

andreasjordan:
## Type of Change

When downloading cumulative updates for SQL Server 2014 I noticed, that also the x86 files where downloaded.
Reason is that the filtering for Architecture was not working, as `$item in $InputObject.Link` was not a list of all files for the KB, but only a single file - with no chance to see the other files for that KB. So the "Could not find architecture match, downloading all" never worked here. Maybe worked before pipeline input from Get-DbaKbUpdate was enabled.
So I reduced the two loops into one - sorry, this makes the diff not very readable again.

I also found other issues with the tests and the way the filename was build.

### DIFF
--- a/functions/Save-DbaKbUpdate.ps1
+++ b/functions/Save-DbaKbUpdate.ps1
@@ -71,8 +71,13 @@ function Save-DbaKbUpdate {
         [switch]$EnableException
     )
     process {
-        if ($Name.Count -gt 0 -and $PSBoundParameters.FilePath) {
+        if ($Name.Count -gt 1 -and $PSBoundParameters.FilePath) {
             Stop-Function -Message "You can only specify one KB when using FilePath"
+            return
+        }
+
+        if ($Architecture -eq "All" -and $PSBoundParameters.FilePath) {
+            Stop-Function -Message "You can only specify one Architecture when using FilePath"
             return
         }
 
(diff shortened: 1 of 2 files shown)


andreasjordan:
This is my use case code to download the cumulative updates in a lab environment, that I will put in "Discussions / Show and tell" after this is released:
```
Import-Module -Name dbatools

function Get-CU {
    param(
        [Parameter(Mandatory)]
        [ValidateSet('2019', '2017', '2016', '2014')]
        [string]$Version,
        [regex]$BuildBaseRegex,
        [int]$Last = 999,
        [string[]]$Exclude,
        [string]$Path = '.'
    )
    if ($null -eq $BuildBaseRegex) {
        $BuildBaseRegex = switch ($Version) {
            '2019' { '^15' }
            '2017' { '^14' }
            '2016' { '^13.0.5' }  # Based on SP2
            '2014' { '^12.0.6' }  # Based on SP3
        }
    }
    if ($Version -eq '2019') { 
        $Exclude += 'CU7'  # CU7 is not available any more
    }
    $buildrefFile = Join-Path -Path (Get-DbatoolsConfigValue -Name 'Path.DbatoolsData') -ChildPath "dbatools-buildref-index.json"
    $buildrefData = (Get-Content -Path $buildrefFile -Raw | ConvertFrom-Json).Data
    $cuData = $buildrefData | 
        Where-Object -FilterScript { $_.Version -match $BuildBaseRegex -and $_.CU -ne $null -and $_.CU -notin $Exclude } |
        Sort-Object -Property KBList |
        Select-Object -Last $Last
    foreach ($cu in $cuData) {
        $kbNr = $cu.KBList
        $cuName = $cu.CU
        $filePath = Join-Path -Path $Path -ChildPath "SQLServer$Version-KB$kbNr-$cuName-x64.exe"
        if (-not (Test-Path -Path $filePath)) {
            Save-DbaKbUpdate -Name $kbNr -FilePath $filePath
        }
    }
}

Update-DbaBuildReference
$patchPath = 'D:\Software\SQLServerPatches'
Get-CU -Version 2019 -Path $patchPath
Get-CU -Version 2017 -Path $patchPath -Last 5
Get-CU -Version 2016 -Path $patchPath -Last 5
Get-CU -Version 2014 -Path $patchPath
```

PR #7540, opened 2021-07-13
Major change: Update SMO, update dbatools library, and switch System.Data to Microsoft.Data

potatoqualitee:
This is a big one! Since there's so much potential for conflict, I'd like to get this merged as quickly as possible. If there are mild style issues, we can address in a new PR.

When/if you test this branch, remember that you will have to close VS Code and powershell/PowerShell ISE before switching branches as it's replacing a ton of DLLs, both SMO and dbatools.

I've tested this on Linux, Windows 10, Windows 2008 R2 and whatever appveyor is (2012 I think?), all with a surprising amount of luck.


potatoqualitee:
* The new SMO failed on tabs in Queries (`.Query`), so I replaced tabs with spaces in the ps1xml
* You may notice some changes to tests -- nothing impactful, I either tried to switch the instance to get appveyor to cooperate or I fixed an issue I've seen for a while
* Microsoft.Data generates mildly new connectionstrings that are incompatible with System.Data. Anything that requires a System.Data SqlConnection had to be created using System.Data or the connection string had to be converted
* I removed the usage of exclamation points in end-user messages in the dbatools library while I was upgrading the library. i also clarified some error messages. i've wanted to do this for a while and took the opportunity. 
* The new library was built with Visual Studio 2019 Community Edition though it can be recompiled using the dotnet sdk

potatoqualitee:
Got a couple more tests to fix (Pester syntax errors, mostly) but feel free to review while I clean up

potatoqualitee:
Oh, also, the .NET version is drastically newer than the old one (2 vs 5) *but* the Desktop version that we target is from 2018. This is reasonable in keeping with trying to accommodate older environments, as 2018 was 3 years ago. My `Windows Server 2008 R2` running `PS3` that gets updated when it wants to was able to run it, so I'm comfortable with the major update.

potatoqualitee:
Annnnnnnd I rewrote `New-DbaConnectionStringBuilder` to align with PowerShell standards, as it had a C# flavor. I introduced a `Legacy` parameter to allow for `System.Data` or `Microsoft.Data` and while I was at it, cleaned up. The WhatIf was not useful "making new conn string", so I removed it.

bin/build-project.ps1
```diff
@@ -55,9 +55,9 @@ Write-Verbose -Message "Building the library with command $dotnet build $Project
 
 if ($MsbuildTarget -eq 'Build') {
     if ($PSVersionTable.PSVersion.Major -ge 6) {
-        $dll = (Resolve-Path -Path "$libraryBase\netcoreapp2.1\dbatools.dll" -ErrorAction Ignore)
+        $dll = (Resolve-Path -Path "$libraryBase\netcoreapp3.1\dbatools.dll" -ErrorAction Ignore)
     } else {
-        $dll = (Resolve-Path -Path "$libraryBase\net452\dbatools.dll" -ErrorAction Ignore)
+        $dll = (Resolve-Path -Path "$libraryBase\net462\dbatools.dll" -ErrorAction Ignore)
```
wsmelton, line 60: This makes no difference because the code no longer runs.

wsmelton, line 60: Did you rebuild the library?

potatoqualitee: I did, then committed the change. But I used Visual Studio, not any of our scripts.

potatoqualitee: Yes indeed

potatoqualitee:
.NET reference for those wanting to know about the 4.6.2 requirement https://docs.microsoft.com/en-us/dotnet/framework/get-started/system-requirements

PR #7551, opened 2021-07-14
Remove extra dbatools.dll file

potatoqualitee:
We likely moved the library when we went to core and didn't clean the old library


potatoqualitee:
Tests pass on Windows and Linux, merging.

PR #7567, opened 2021-07-15
Restore-DbaDatabase - Fix SMO refresh

andreasjordan:
## Type of Change

 - [X] Bug fix (non-breaking change )

As the connection context might have been changed by Connect-DbaInstance, `.Equals` does not work here in any cases.
So be better test for the type, that works.
And we refresh the original input object.


potatoqualitee (approved):
Nice!

PR #7429, opened 2021-06-10
Replication commands - document commands as unsupported

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #6548 (and already closed #6897 )
 - [ ] New feature (non-breaking change, adds functionality, fixes # )
 - [ ] Breaking change (effects multiple commands or functionality, fixes # )
 - [ ] Ran manual Pester test and has passed (`.\tests\manual.pester.ps1)
 - [ ] Adding code coverage to existing functionality
 - [ ] Pester test is included
 - [ ] If new file reference added for test, has is been added to github.com/sqlcollaborative/appveyor-lab ?
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 
Document relication functions as currently unsupported and lead to issue #7428 for more information.

PR #7573, opened 2021-07-16
Remove SSIS Commands - Incompatible SMO version

### LINKED ISSUE #7553: Check for SSIS Compat

Yes

Now that we've got new SMOs, I still need to investigate the change's impact on the SSIS commands before 1.1.0

### Steps to Reproduce

```powershell
any
```

Yes

### Other details or mentions

_No response_

### PowerShell Host Version

any

any

potatoqualitee:
fixes #7553

Seems these have not been doable for a while -- the DLLs were not even loaded. Ultimately, these commands were always problematic because of the DLLs. Just commenting them out until... later? Unsure when they'll be ported to the new SMO.

### DIFF
--- a/dbatools.psd1
+++ b/dbatools.psd1
@@ -531,13 +531,13 @@
         # noncoresmo
         # SMO issues
         'Export-DbaUser',
-        'Get-DbaSsisExecutionHistory',
+        #'Get-DbaSsisExecutionHistory',
         'Get-DbaRepDistributor',
         'Copy-DbaPolicyManagement',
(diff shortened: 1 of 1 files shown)


andreasjordan:
Maybe add a note in the documentation of these commands like in #7429?

potatoqualitee:
I'm thinking that I'll do some DLL cleanups for both Replication and SSIS. I'll add the commands back, remove our own DLLs, add `[System.Reflection.Assembly]::LoadWithPartialName`, add warnings, etc

PR #7576, opened 2021-07-17
New Commands: New-/Get-/Remove-DbaFirewallRule

### LINKED ISSUE #3519: New firewall rules command

andreasjordan:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality, fixes #3519 )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system


wsmelton:
This will only work on Windows PowerShell 5.1, and only if the target machine (instance) is on Windows Server 2012 R2+

wsmelton:
It might work in PS4, don't have a 2012 R2 machine up.

andreasjordan:
I found a germany article about PS3 having the module NetSecurity (which includes *-NetFirewallRules) and I read this this as "We have NetSecurity on Windows Server 2012":
https://docs.microsoft.com/en-us/powershell/module/netsecurity/?view=winserver2012-ps

andreasjordan:
And the command work from a PowerShell 5.1 as well as 7.1.3.

functions/New-DbaFirewallRule.ps1
```diff
@@ -0,0 +1,281 @@
...
+    )
+
+    begin {
+        $cmdScriptBlock = {
+            # This scriptblock will be processed by Invoke-Command2.
+            $firewallRuleParameters = $args[0]
+
+            try {
+                $successful = $true
+                $cimInstance = New-NetFirewallRule @firewallRuleParameters -WarningVariable warn -ErrorVariable err -WarningAction SilentlyContinue -ErrorAction SilentlyContinue
```
potatoqualitee, line 116: Since this command may not exist, please use this before `if (-not (Get-Command New-NetFirewallRule)) { throw whatever`

functions/New-DbaFirewallRule.ps1
```diff
@@ -0,0 +1,281 @@
...
+                        $commandResult = Invoke-Command2 -ComputerName $instance.ComputerName -Credential $Credential -ScriptBlock $cmdScriptBlock -ArgumentList $config
+                    } catch {
+                        Stop-Function -Message "Failed to execute command on $($instance.ComputerName) for instance $($instance.InstanceName)." -Target $instance -ErrorRecord $_ -Continue
+                    }
+
+                    # Output information
+                    [PSCustomObject]@{
+                        ComputerName = $instance.ComputerName
+                        InstanceName = $instance.InstanceName
+                        SqlInstance  = $instance.SqlFullName.Trim('[]')
```
potatoqualitee, line 268: omg does this remove the start `[` and `]`?! That's awesome! I always do replaces.

functions/New-DbaFirewallRule.ps1
andreasjordan: This is the output, if I change `New-NetFirewallRule` to `New-NetFirewallRuleX`:
image

I think that is ok. Do you want to display a different message?

functions/New-DbaFirewallRule.ps1
andreasjordan: Yes, it does. It would be very nive if our [DbaInstanceParameter] type would have a property "SqlInstance" for those use cases...

functions/New-DbaFirewallRule.ps1
potatoqualitee: Yes, please. Let them know that the command is only supported in XYZ and up, otherwise, it's likely they will believe something is missing in dbatools.

andreasjordan:
I just don't see where my error in the test is - will continue tomorrow....

wsmelton:
I don't think the test on this command or the remove would interfere with other tests running but to be sure we might want to move this to the services group // @potatoqualitee 

@andreasjordan you can add it to this listing to accomplish that. Most of the tests that do "bad things" or directly effect the service for SQL Server we throw into this list so it is isolated:

https://github.com/sqlcollaborative/dbatools/blob/54e5c1b49a26c6b51738ac82995b3788458eb1d0/tests/pester.groups.ps1#L10-L22

potatoqualitee:
@andreasjordan - let me know if you want me to login via RDP and see what's up. Sometimes appveyor just doesn't like our tests.

andreasjordan:
The Invoke-Command2 of Remove-DbaFirewallRule does not give me a return.
Maybe it does not like "localhost"? Will try to resolve hostname.

andreasjordan:
@potatoqualitee Test work now. I'm not totally happy with the comands and would like to make some changes the next days. So will probably open the pull request for review end of this week.

potatoqualitee:
Interesting! What don't you like about it? Would you like me to wait or merge now?

andreasjordan:
Running the New- command twice give you this:
image

It's because a rule with that name exists and the "Name" is the primary key. I would like to change the message to something more readable and maybe introduce a parameter -Force to remove the rule in that case.

andreasjordan:
Maybe also create a rule for DAC?

potatoqualitee:
DAC sounds great! But let the user opt-in? I also like the idea of Force.

andreasjordan:
Just give me another 24 hours. Does not have to be in v1.1.0.

potatoqualitee:
I'm still waiting a bit for 1.1.0! Maybe Wednesday.

andreasjordan:
I'm thinking of removing -Auto and adding -Type (like in Get-DbaService) with a validate set of Engine, Browser, DAC (and then later add more?). 
If the parameter is set, then only the rules for the named types are created, failing if they are already in place with a hint to -Force for replacing them.
If the parameter is not set, then the minimal steps needed to get access to the instance are performed:
* Check if there is already a rule for the instance. If so, check if LocalPort or Program are set correctly. If so, do nothing and only report in the output object, that a rule was already in place. If settings are not correct, remove the rule and create a new one. Report in the output object, that the rule was replaced.
* Check if a rule for the Browser is needed. If so, check if a rule already is in place. If so, do nothing and report in the output object that there is an existing rule for the browser. If not, create the rule.

The new parameter -Force can only be used together with -Type to replace an existing rule. I would not test but just try to remove the rule - as Remove- does the Get- and would just not remove if there is no rule.

andreasjordan:
I would also add -Type to Get-DbaFirewallRule with the same behavior:
* If Type is used, only get those rules 
* If Type is not used, get the rule for the instance and then see if this instance needs a Browser. If so, get the rule for the Browser as well.

And I could do the same with Remove-. But I would not remove the Browser automatically, because it might be used by other instances.

functions/Get-DbaFirewallRule.ps1
```diff
@@ -0,0 +1,235 @@
...
+        Returns all firewall rules on the computer SRV1 related to SQL Server.
+        The value "AllOnComputer" only uses the computer name part of SqlInstance.
+
+    #>
+    [CmdletBinding()]
+    param (
+        [parameter(Mandatory, ValueFromPipeline)]
+        [DbaInstanceParameter[]]$SqlInstance,
+        [PSCredential]$Credential,
+        [ValidateSet('Engine', 'Browser', 'AllOnComputer')]
```
potatoqualitee, line 81: Can we get AnalysisServices, IntegrationServices, ReportingServices here too?

If so, can we get the common things listed on `Install-DbaInstance -Feature`? 

```
"All", "Engine", "Tools", "Replication", "FullText", "DataQuality", "PolyBase", "MachineLearning", "AnalysisServices",
            "ReportingServices", "ReportingForSharepoint", "SharepointAddin", "IntegrationServices", "MasterDataServices", "PythonPackages", "RPackages",
            "ReplayController", "ReplayClient", "SDK", "BIDS", 
```

functions/Get-DbaFirewallRule.ps1
```diff
@@ -0,0 +1,235 @@
...
+                            Program     = ($rule | Get-NetFirewallApplicationFilter).Program
+                            Rule        = $rule
+                        }
+                    }
+                }
+                [PSCustomObject]@{
+                    Successful = $successful
+                    Rules      = $rulesWithDetails
+                    Verbose    = $verbose
+                    Warning    = $warn
```
potatoqualitee, line 131: This is a deviation from the way we usually handle exceptions. Is this because of Remoting?

functions/Get-DbaFirewallRule.ps1
```diff
@@ -0,0 +1,235 @@
...
+            if ('AllOnComputer' -in $Type) {
+                Write-Message -Level Verbose -Message 'Returning all rules for target computer'
+                $outputRules += $rules
+            } elseif ($null -eq $Type) {
+                Write-Message -Level Verbose -Message 'Returning rule for instance and maybe for Browser'
+                # Get the rule for the instance
+                $outputRules += $rules | Where-Object { $_.Type -eq 'Engine' -and $_.InstanceName -eq $instance.InstanceName }
+                if ($outputRules.Count -eq 0) {
+                    Write-Message -Level Verbose -Message 'No rule found for instance'
+                } elseif ($outputRules.LocalPort -eq '1433') {
```
potatoqualitee, line 216: I didn't see any of this in my output, and I'm left wondering -- did we do it by TCP port or Program? Can the output be more useful?

potatoqualitee (commented):
gorgeous, just have some questions

functions/Get-DbaFirewallRule.ps1
andreasjordan: Yes, if we find the time to code that.

functions/Get-DbaFirewallRule.ps1
andreasjordan: Yes. Still not sure if this is the best way, may be changed...

andreasjordan: But the idea is to not fail on the remote machine and get as much info back to then report and fail on the local machine.

functions/Get-DbaFirewallRule.ps1
andreasjordan: It depends on the network configuration of the instance: Static port uses LocalPort, dynamic port uses Program.
What do you mean by "more useful"?

andreasjordan:
Next change will be -Force for New-DbaFirewallRule. Later today...

functions/Get-DbaFirewallRule.ps1
potatoqualitee: I'd like to see what it did. If it picked a static port, I'd like to see that port 1433 was secured. If it uses Program, I'd like to see the path.

andreasjordan: Oh, now I understand. The Get- only works by Name of the rule. So we don't need to get the configuration of the instance. Only New- needs to get the configuration to build the rule. The LocalPort and Program is then part of the output.

potatoqualitee: Oh oops, yes it was New. Can we get the details from Get?

andreasjordan: What is missing?
image

potatoqualitee: Oh weird. Nothing on yours. I may have had an old commit, my apologies. That is beautiful, can we get something as detailed for all DbaFirewallRule commands?

andreasjordan:
image

andreasjordan:
image

andreasjordan:
image

potatoqualitee:
Yesssss! 🤩  I'll take a look at it in depth then likely approve and merge tonight.

functions/Remove-DbaFirewallRule.ps1
```diff
@@ -0,0 +1,184 @@
...
+
+    .PARAMETER Credential
+        Credential object used to connect to the Computer as a different user.
+
+    .PARAMETER Type
+        Removes firewall rules for the given type(s).
+        Valid values are:
+            Engine - for the SQL Server instance
+            Browser - for the SQL Server Browser
+            AllOnComputer - for all firewall rules on the target computer related to SQL Server
```
potatoqualitee, line 30: Just realized that AllOnComputer is too scary, sounds like every firewall rule. Can you say AllSQL or AllDba or SqlRelated or AllInstance or something. Yeah I like AllInstance

potatoqualitee:
Looks awesome! Let's merge. I'll add to manifests momentarily

wsuhoey:
GASP - I can't believe my vision for this command from a few YEARS ago has finally been realized!! Thank you Andreas!!! SQUEEEEEEEEE!!!

PR #7578, opened 2021-07-18
New command - Get-DbaDbFileMapping

### LINKED ISSUE #3080: Add a OutFileMapping switch and output to Get-DbaDatabaseFile
### Problem to solve
Add an easy way to build the FileMapping Hashtable for Restore-DbaDatabase from an existing database

For example:
`$FileMapping = Get-DbaDatabaseFile -SqlInstance Instance1 -Database db1 -OutFileMapping`

Would return $FileMapping of type Hashtable like this:
```
$FileMapping = @{
"Datafile1"  = "c:\data\datafile1.mdf";
"Datafile2" = "d:\ArchiveData\File2.mdf";
"LogFile1" = "e:\LogFiles\Log1.ldf"}
```

potatoqualitee:
## Type of Change
 - [x] New feature (non-breaking change, adds functionality, fixes #3080)

### DIFF
--- a/functions/Get-DbaDbFileMap.ps1
+++ b/functions/Get-DbaDbFileMap.ps1
@@ -1,0 +1,89 @@
+function Get-DbaDbFileMap {
+    <#
+    .SYNOPSIS
+        Provides an easy way to build the FileMapping Hashtable for Restore-DbaDatabase from an existing database
+
+    .DESCRIPTION
+        Provides an easy way to build the FileMapping Hashtable for Restore-DbaDatabase from an existing database
+
+    .PARAMETER SqlInstance
+        The target SQL Server instance or instances. This can be a collection and receive pipeline input.
+
+    .PARAMETER SqlCredential
+        Login to the target instance using alternative credentials. Accepts PowerShell credentials (Get-Credential).
+
+        Windows Authentication, SQL Server Authentication, Active Directory - Password, and Active Directory - Integrated are all supported.
+
+        For MFA support, please use Connect-DbaInstance.
+
+    .PARAMETER Database
+        The database(s) to process - this list is auto-populated from the server. If unspecified, all databases will be processed.
+
+   .PARAMETER InputObject
+        Database object piped in from Get-DbaDatabase
+
+    .PARAMETER EnableException
+        By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
+        This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
+        Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
+
+    .NOTES
+        Tags: Database
+        Author: Chrissy LeMaire (@cl), netnerds.net | Andreas Jordan (@JordanOrdix), ordix.de
+
+        Website: https://dbatools.io
+        Copyright: (c) 2021 by dbatools, licensed under MIT
+        License: MIT https://opensource.org/licenses/MIT
+
+    .LINK
+        https://dbatools.io/Get-DbaDbFileMap
+
+    .EXAMPLE
+        PS C:\> $filemap = Get-DbaDbFileMap -SqlInstance sql2016 -Database test
+        PS C:\> Get-ChildItem \\nas\db\backups\test | Restore-DbaDatabase -SqlInstance sql2019 -Database test -FileMapping $filemap.FileMap
+
+        Restores test to sql2019 using the file structure built from the existing database on sql2016
+    #>
+    [CmdletBinding()]
+    param ([parameter(ValueFromPipeline)]
+        [DbaInstanceParameter[]]$SqlInstance,
+        [PSCredential]$SqlCredential,
+        [object[]]$Database,
+        [parameter(ValueFromPipeline)]
+        [Microsoft.SqlServer.Management.Smo.Database[]]$InputObject,
+        [switch]$EnableException
+    )
+
+    process {
+        if (Test-Bound -not 'SqlInstance', 'InputObject') {
+            Write-Message -Level Warning -Message "You must specify either a SQL instance or supply an InputObject"
+            return
+        }
+
+        if (Test-Bound -Not -ParameterName InputObject) {
+            $InputObject += Get-DbaDatabase -SqlInstance $SqlInstance -SqlCredential $SqlCredential -Database $Database
+        }
+
+        foreach ($db in $InputObject) {
+            if ($db.IsAccessible) {
+                Write-Message -Level Verbose -Message "Processing database: $db"
+                $fileMap = @{}
+                foreach ($file in $db.FileGroups.Files) {
+                    $fileMap[$file.Name] = $file.FileName
+                }
+                foreach ($file in $db.LogFiles) {
+                    $fileMap[$file.Name] = $file.FileName
+                }
+                [PSCustomObject]@{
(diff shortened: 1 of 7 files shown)


functions/Get-DbaDbFileMap.ps1
```diff
@@ -0,0 +1,90 @@
...
+        PS C:\> $filemap = Get-DbaDbFileMap -SqlInstance sql2016 -Database test
+        PS C:\> Get-ChildItem \\nas\db\backups\test | Restore-DbaDatabase -SqlInstance sql2019 -Database test -FileMapping $filemap.FileMap
+
+        Restores test to sql2019 using the file structure built from the existing database on sql2016
+    #>
+    [CmdletBinding()]
+    param ([parameter(ValueFromPipeline)]
+        [DbaInstanceParameter[]]$SqlInstance,
+        [PSCredential]$SqlCredential,
+        [object[]]$Database,
```
andreasjordan, line 51: I would use [string[]] here to indicate that it's the database name and not a database SMO object.

wsmelton: We use `[object[]]` @andreasjordan because TEPP populates the list of Databases for dynamic validate sets. If you use `string` type this won't be accessible.

tests/Get-DbaDbFileMap.Tests.ps1
```diff
@@ -0,0 +1,31 @@
...
+
+Describe "$CommandName Unit Tests" -Tag 'UnitTests' {
+    Context "Validate parameters" {
+        [object[]]$params = (Get-Command $CommandName).Parameters.Keys | Where-Object { $_ -notin ('whatif', 'confirm') }
+        [object[]]$knownParameters = 'SqlInstance', 'SqlCredential', 'Database', 'InputObject', 'EnableException'
+        $knownParameters += [System.Management.Automation.PSCmdlet]::CommonParameters
+        It "Should only contain our specific parameters" {
+            (@(Compare-Object -ReferenceObject ($knownParameters | Where-Object { $_ }) -DifferenceObject $params).Count ) | Should Be 0
+        }
+    }
```
wsmelton, line 13: We changed the pattern on this test to help on performance, `Get-Command` is very slow to use constantly.

```powershell
        [array]$params = ([Management.Automation.CommandMetaData]$ExecutionContext.SessionState.InvokeCommand.GetCommand($CommandName, 'Function')).Parameters.Keys
        [object[]]$knownParameters = 'SqlInstance', 'SqlCredential', 'Database', 'InputObject', 'EnableException'

            Compare-Object -ReferenceObject $knownParameters -DifferenceObject $params | Should -BeNullOrEmpty
        }
    }
```

functions/Get-DbaDbFileMap.ps1
wsmelton: Gives this ability when you use `[object[]` on a parameter where we have the backend setup for TEPP.

image

functions/Get-DbaDbFileMap.ps1
```diff
@@ -0,0 +1,90 @@
...
+                foreach ($file in $db.LogFiles) {
+                    $fileMap[$file.Name] = $file.FileName
+                }
+
+                [PSCustomObject]@{
+                    ComputerName = $db.ComputerName
+                    InstanceName = $db.InstanceName
+                    SqlInstance  = $db.SqlInstance
+                    Database     = $db.Name
+                    FileMap      = $fileMap
```
wsmelton, line 83: Can we call this `FileMapping` since it will contain multiple files and it will match the parameter name we use in the Restore command.

functions/Get-DbaDbFileMap.ps1
andreasjordan: Ok, didn't know that. Thanks. Will have a look at the commands I changed in the past.

wsmelton: We can add pretty much anything we can get from SMO or t-sql to list out. Primary one is database. Not the easiest system to deal with and not always consistent in some environments I touch.

functions/Get-DbaDbFileMap.ps1
potatoqualitee: ohh i like it, yes

functions/Get-DbaDbFileMap.ps1
potatoqualitee: wait for real? i had used object in the past to allow both SMO and text but then went to string[] instead after learning to use `InputObject` better. I'll see if string[] destroys this

wsmelton: It does 😉 because TEPP has to be an object type to return the list to tab completor.

wsmelton: We covered a few initially and at one point I had added in some as I worked on commands, but in that phase of maintainers taking a vacation for a bit 🍹 I never got back to adding support for the main parameters in commands.

functions/Get-DbaDbFileMap.ps1
wsmelton: One thing tied to this as well with the name of the command. Should we call it `Get-DbaDbFileMapping` as well? I mean when you think about how it is phrased when we are talking about it or maybe in docs as well, we term it has "you need to check the database mapping" or "what is the mapping of the database files" 🤷🏻

wsmelton: One of those is "Mapping" plural for "Map" in this sense. If we are tying the command directly to the use of the parameter on the Restore command it would make sense to call the command the same as well. Not a hard yes for me, just thinking out loud.

I would also include an update to help on Restore for that parameter and give an example in that using this command so it is found more easily.

potatoqualitee: Just renamed 👍🏼

functions/Get-DbaDbFileMap.ps1
potatoqualitee: yay, just tested, `[string[]]` does work 💯  updated

potatoqualitee:
Someone let me know when this is good to merge 👍🏼

functions/Get-DbaDbFileMapping.ps1
```diff
@@ -0,0 +1,90 @@
...
+        License: MIT https://opensource.org/licenses/MIT
+
+    .LINK
+        https://dbatools.io/Get-DbaDbFileMapping
+
+    .EXAMPLE
+        PS C:\> $filemap = Get-DbaDbFileMapping -SqlInstance sql2016 -Database test
+        PS C:\> Get-ChildItem \\nas\db\backups\test | Restore-DbaDatabase -SqlInstance sql2019 -Database test -FileMapping $filemap.FileMapping
+
+        Restores test to sql2019 using the file structure built from the existing database on sql2016
```
wsmelton, line 45: I would add this example to the restore command as well and then I'm good to merge

wsmelton (approved):

functions/Get-DbaDbFileMapping.ps1
wsmelton: Might replace the example I think the restore command has for manually building the hashtable object.

potatoqualitee: I'll add it in addition to -- I think it's important to show ppl how to build the hashtable.

potatoqualitee: Ty for the approval! Will merge once tests finish.

PR #7579, opened 2021-07-18
SSIS and Replication Commands -- Remove SSIS Commands, Update SMO/Library Import and Messages

### LINKED ISSUE #7553: Check for SSIS Compat

potatoqualitee:
fixes #7553

Creating SSIS objects is unreliable and supporting it will require a ton of DLLs that will increase the size of dbatools. We can move these commands to a new module if we ever decide to create one.

RMO, so far, is still working.


potatoqualitee:
All checks pass ✔️

PR #7582, opened 2021-07-18
Import-DbaXESessionTemplate - Add StartUpState Parameter

### LINKED ISSUE #4916: Import-DbaXESessionTemplate cannot set startup state of XE session to on
### Summary of new feature

With some extended events sessions, like database_health, the startup state should normally be set to on. Otherwise the session won't start automatically after a restart.
Import-DbaXESessionTemplate cannot set this property.

I can think of two solutions:
* Add a parameter StartUpState (ON or OFF) to Import-DbaXESessionTemplate.
* Add a property StartUpState to xetemplates-metadata.xml.

Or a combination of both.

Apply the state to the session after it's created according to the template.

Current developer branch

potatoqualitee:
Fixes #4916

Defaults to Off and Includes example. Also starts it immediately.

### DIFF
--- a/functions/Import-DbaXESessionTemplate.ps1
+++ b/functions/Import-DbaXESessionTemplate.ps1
@@ -37,6 +37,9 @@ function Import-DbaXESessionTemplate {
 
         This path is relative to the destination directory
 
+    .PARAMETER StartUpState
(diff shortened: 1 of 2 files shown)


potatoqualitee:
I need this command in another branch, all tests pass, will merge

PR #7583, opened 2021-07-19
Get-DbaDbBackupHistory - Avoid a .ToString() that pollutes $error

potatoqualitee:
Found this while cleaning up Linux paths but wanted to make it its own PR since the command is so important


potatoqualitee:
All tests pass so I'll merge. If there are any issues, we'll take a closer look.

PR #7585, opened 2021-07-19
Connect-DbaInstance - Add back Microsoft.Identity.Client support

potatoqualitee:
Helps when connecting using Connstrings that use Service Principals


potatoqualitee:
This will be merged immediately once it passes because of its low impact, the fact that I tested on Windows and Linux and ultimately, I need it in another branch.

PR #7590, opened 2021-07-19
Write-DbaDbTableData - Parse for Database so that it doesn't have to be respecified

### LINKED ISSUE #7587: Write-DbaDataTable ignores when Database is in $connstring

Yes

```
WARNING: [14:36:55][Write-DbaDbTableData] You must specify a database or fully qualified table name.
```

### Steps to Reproduce

```powershell
[682ms] C:/github/dbatools❯ $connstring = "Server=dbatoolstest.database.windows.net; Authentication=Active Directory Service Principal; Database=test; User Id=GUID; Password=xyz"
[682ms] C:/github/dbatools❯ Get-ChildItem -File | Write-DbaDataTable -SqlInstance $connstring -Table new
WARNING: [14:36:55][Write-DbaDbTableData] You must specify a database or fully qualified table name.
```

Yes

### Other details or mentions

_No response_

### PowerShell Host Version

PSVersion                      5.1.19041.1023                                                                        
BuildVersion                   10.0.19041.1023                                                                       

Azure

potatoqualitee:
Fixes #7587

Added newer Identity DLL, Forced recognition of Database, Collected the database object a different way because using the hashtable was failing, suddenly.

Left in a note that it's initializing even though `$isAzure` should be true

`WARNING: [16:10:18][Write-DbaDbTableData] Failure | unknown property LastBackupDate`

If I comment out the initialization, it does not throw that first time


andreasjordan:
There are issues with `$isAzure` - have a look at that too, please. In the new code path, it was only used in `if ($loadedSmoVersion -ge 11 -and -not $isAzure) {` (that was a bug). 
Maybe just try to test once very early with `Test-Azure` and save the result in `$isAzure` and then only use `$isAzure`. You can also completely remove `Test-Azure` and just do the test in the main code.

andreasjordan:
We will have the same problem in every command that uses `-NonPooledConnection` - so I would not change the `Connect-DbaInstance` call but change `Connect-DbaInstance` itself to better handle connection strings as input. Maybe fist create a server SMO based on the connection string itself and then pass it into `Connect-DbaInstance` again together with the additional parameters.

andreasjordan:
But maybe that is to compicated. Let's start with an easy solution: Copy and adapt some code rigth after
```
                } elseif ($inputObjectType -in 'RegisteredServer', 'ConnectionString') {
                    $server = New-Object -TypeName Microsoft.SqlServer.Management.Smo.Server -ArgumentList $serverName
                    $server.ConnectionContext.ConnectionString = $connectionString
```
Do something like the `# Test if we have to copy the connection context` code some line above.

functions/New-DbaConnectionStringBuilder.ps1
```diff
@@ -134,6 +139,9 @@ function New-DbaConnectionStringBuilder {
             if ($ColumnEncryptionSetting -eq "Enabled") {
                 $builder['Column Encryption Setting'] = "Enabled"
             }
+            if ($pooled) {
```
andreasjordan, line 142: Does this work? If -NonPooledConnection, then $NonPooledConnection is $true, $pooled is $false and $builder['Pooled'] would not be set. I would just set $builder['Pooled'] in any case.

potatoqualitee: haha i dont know, i kinda confused myself. let me test.

potatoqualitee: yeah, i had to fix a keyword but it works (I have a bunch of default params)

image

potatoqualitee:
In action -- changing Database and changing pooled state

image

Here's azure working (tho i dont have permissions to my own db)

image

potatoqualitee:
k, it's ready! we now use the `Test-Azure` extensively and immediately and I improved the check

andreasjordan:
Yes, looks good!

potatoqualitee:
WOO! Merging 🥳

PR #7595, opened 2021-07-20
Set-DbaAgentJobOwner - Only attempt to change Local job types 

### LINKED ISSUE #6891: Set-DbaAgentOwner needs to have a filter for MSX jobs
### Summary of new feature   When you run Set-DbaAgentOwner it TRIES to look at non local MSX jobs and gives the soft red error that it can't modify the job 

* Clear/concise description of what the problem is that the new feature can solve.   
We have a ton of MSX jobs and if we did not have to process through the jobs it would be a big time saver. 

### Proposed technical details (if applicable)   Add -local switch 

* e.g. Blog post reference that shows example code or functionality

### Latest version of dbatools as of writing   1.0.121

potatoqualitee:
fixes #6891

Also added a `-Type` parameter to `Get-DbaAgentJob`

### DIFF
--- a/functions/Get-DbaAgentJob.ps1
+++ b/functions/Get-DbaAgentJob.ps1
@@ -37,6 +37,9 @@ function Get-DbaAgentJob {
     .PARAMETER IncludeExecution
         Include Execution details if the job is currently running
 
+    .PARAMETER Type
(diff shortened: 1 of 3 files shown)


potatoqualitee:
not a big change and tests pass, merging 👍🏼

PR #7597, opened 2021-07-20
More Linux Compatability

potatoqualitee:
That was interesting -- decrypting secure strings is different on Linux so I created a new command to simplify the code and use PSCredential to do the decrypting. I also fixed a few other path issues


potatoqualitee:
Merging

PR #7599, opened 2021-07-21
New-Command - Get-DbaInstalledPatch

### LINKED ISSUE #5391: FR: Get-DbaSQLPatches
---
Name: Feature request/idea 💣 🥊 
About: Suggest a new feature or enhancement
---

### Summary of new feature
* Retrieves a list of SQL Patches (SP/CU/Hot-fixes) on a Computer and the instances on it.

* https://github.com/hfleitas/Monitoring
[Get-SQLUpdates.ps1](https://github.com/hfleitas/Monitoring/blob/master/Get-SQLUpdates.ps1)

* 0.9.808

potatoqualitee:
## Type of Change
 - [x] New feature (non-breaking change, adds functionality, fixes #5391 )

This is based off of @hfleitas's command. I'm up for suggestions for change of name/functionality.


andreasjordan:
image

andreasjordan:
image

So I get the same three patches for all three instances...

potatoqualitee:
InstallDate is a [dbadate] type which gives us

```
Get-DbaInstalledPatch -ComputerName sql2012 | Select -ExpandProperty InstallDate

Date        : 10/1/2016 12:00:00 AM
Day         : 1
DayOfWeek   : Saturday
DayOfYear   : 275
Hour        : 0
Kind        : Unspecified
Millisecond : 0
Minute      : 0
Month       : 10
Second      : 0
Ticks       : 636108768000000000
TimeOfDay   : 00:00:00
Year        : 2016
```

So you would have to sort by InstallDate.Date

potatoqualitee:
Interesting! I'll set up 3 instances and check it out.

andreasjordan:
`Get-DbaInstalledPatch -ComputerName SRV1 | Sort-Object InstallDate.Date | ft` works! Maybe an example for that? Might by a typical use cases to see them in order of installation.

potatoqualitee:
haha, i actually never did that before! super cool to know that it works! I'll include an example

andreasjordan:
If the command should show the patches installed on the computer, I would expect each patch to be returned only once, maybe with an array of instances where the patch is applied. 

If the command should show the patches installed on an instance, I would expect a parameter -SqlInstance to return only the patches of that instance.

potatoqualitee:
there must be a way to associate which patch patches a specific instance 🤔

andreasjordan:
Looking at the code, you only get installed patches on the computer, better call it "executed patch.exe" without any link to an instance. So you should not loop through the instances and not tell anything about the instances.
The command does not answers the questions "Is my instance up to date?" or "What have I done with my instance?" but only "What patch was executed at the computer?". If I only have one instance - then the frist two questions are alos answered. But only in this case.

potatoqualitee:
`Test-DbaBuild` answers that question. This is just to see what's installed. Are you suggesting removing the SQL instance association? I can see that.

potatoqualitee:
k will remove the association!

functions/Get-DbaInstalledPatch.ps1
```diff
@@ -0,0 +1,79 @@
...
+        [Parameter(ValueFromPipeline)]
+        [DbaInstanceParameter[]]$ComputerName = $env:COMPUTERNAME,
+        [PSCredential]$Credential,
+        [switch]$EnableException
+    )
+    process {
+        foreach ($computer in $ComputerName.ComputerName) {
+            try {
+                $patches = Invoke-Command2 -ComputerName $computer -Credential $Credential -ScriptBlock {
+                    Get-ChildItem -Path HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall | Get-ItemProperty | Sort-Object -Property @{
```
andreasjordan, line 60: Why `-Property @{ ... }`? 
Will not save a lot of seconds, but why not filter first and then sort?

potatoqualitee: because i cannot wrap my head around registry commands and copy/paste worked 😇  if you give me the code or suggest within the PR, i can copy/paste

andreasjordan:
```
Get-ChildItem -Path HKLM:\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall | Get-ItemProperty | Where-Object { $_.DisplayName -like "Hotfix*SQL*" -or $_.DisplayName -like "Service Pack*SQL*" } | Sort-Object InstallDate
```

PR #7600, opened 2021-07-21
XPlat Solution for reading XEL

potatoqualitee:
Fixes #5541

Got it working on Windows but Linux segfaults. Still, Core vs Desktop qualifies as Xplat eh 😅

PR #7603, opened 2021-07-21
Watch-DbaXESession / Read-DbaXEFile / Read-DbaAuditFile - Add Linux support

### LINKED ISSUE #5541: XPlat Solution for reading XEL
https://www.nuget.org/packages/Microsoft.SqlServer.XEvent.XELite/

potatoqualitee:
Fixes #5541 #7600

I ended up using a PowerShell module from Microsoft called `SqlServer.XEvent` that provides one command called `Read-SqlXEvent`. The DLL could be loaded but it provided some stuff that used Tasks which is not easy to implement in PowerShell so I used their PowerShell.

These commands do extend Read-SqlXEvent to make it more useful, but the output looks exactly like it used to in dbatools. They also longer use `QueryableXEventData`.

Now works on Linux without seg faulting!


potatoqualitee:
Setting to draft until I fix `Read-DbaAuditFile` too

PR #7610, opened 2021-07-23
Connect-DbaInstance - Side-step Token usage on Core

potatoqualitee:
## Type of Change

 - [x] Bug fix (non-breaking change

When attempting to generate tokens on Windows Core and Linux, we get this error:

```
Exception: Failed to get access token for Azure SQL DB (error CS1701: Assuming assembly reference 'System.Runtime, Version=4.2.2.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' used by 'Microsoft.SqlServer.ConnectionInfo' matches identity 'System.Runtime, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' of 'System.Runtime', you may need to supply runtime policy)
```

I was unable to resolve it, so I catch when Tokens are attempted on Core and provide an alternative way to login.

```
"Server=$azureserver; Authentication=Active Directory Service Principal; Database=$Database; User Id=clientid; Password=clientsecret"
```

This works across the board, as can be seen in the test that is added which passes.


potatoqualitee:
andreas approved in chat 👍🏼

PR #7612, opened 2021-07-23
New-Command - Get-DbatoolsError

### LINKED ISSUE #7611: Get-DbatoolsError
### Summarize Command's Functionality'

Instead of asking people to do `$error[0]`, we can offer `Get-DbatoolsError` and default to `[0]` or, my fav `Select-Object -First 1`

Then offer `-All`, `-First`, `-Skip`, and `-Last`

No

### Technical Details

_No response_

potatoqualitee:
## Type of Change
 - [x] New feature (non-breaking change, adds functionality, fixes #7611 )

I really like this command, it'll be helpful for the #dbatools channel in Slack and perhaps even our issue template /cc @wsmelton  @wsuhoey  

image

Also, I discovered an error on load for a missing DLL while working on this so I added it back.


potatoqualitee:
one pass and i'm tired of waiting 👍🏼

functions/Get-DbatoolsError.ps1
```diff
@@ -0,0 +1,67 @@
...
+    process {
+        if (Test-Bound -Not First, Last, Skip, All) {
+            $First = 1
+        }
+
+        if ($All) {
+            $First = $global:error.Count
+        }
+
+        $global:error | Where-Object ScriptStackTrace -match dbatools | Select-Object -First $First -Last $Last -Skip $Skip -Property CategoryInfo, ErrorDetails, Exception, FullyQualifiedErrorId, InvocationInfo, PipelineIterationInfo, PSMessageDetails, ScriptStackTrace, TargetObject
```
wsmelton, line 65: Why not call `Get-DbatoolsLog -Errors`

https://github.com/sqlcollaborative/dbatools/blob/development/functions/Get-DbatoolsLog.ps1

potatoqualitee: Our errors are never detailed enough for me. They are pretty but not for deep diving stuff. Compare the screenshot above with a similar error using dbatoolslog 

image

wsmelton: Well we can always change it. The log command retrieves errors based on `[Sqlcollaborative.Dbatools.Message.LogHost]::GetErrors()` which will be the errors for the primary session. If you query `$errors` for our stack name then you are also going to be pulling in the import errors and others that are always seen from code running async (in runspaces we generate). As long as users understand this when they use the new command.

PR #7635, opened 2021-07-28
Update-DbaServiceAccount - Allow for not restarting services

### LINKED ISSUE #6853: Update-DbaServiceAccount - Allow for not restarting services
### Summary of new feature

When service accounts are managed by PIM systems (Thycotic, BeyondTrust, etc.) offer the ability to rotate the passwords in Active Directory, and in conjunction push out those passwords to dependency resources. In most situations when SQL Server service account is being managed the need is to update the password **but not** restart the service. The password change would be done when security needs it, and then DBAs will handle restarting the service during a scheduled maintenance window.

A parameter that supports backward compatibility to how the command current works that simply offers the option to not restart the service.

Parameter name: `-NoRestart`

This parameter will be utilized to have the command not restart the agent service even if the password for the database engine is updated.

potatoqualitee:
## Type of Change
 - [x] New feature (non-breaking change, adds functionality, fixes #6853)

Adds a new `-NoRestart` switch, with examples and warnings.

### DIFF
--- a/functions/Update-DbaServiceAccount.ps1
+++ b/functions/Update-DbaServiceAccount.ps1
@@ -36,6 +36,11 @@ function Update-DbaServiceAccount {
         NETWORKSERVICE
         LOCALSYSTEM
 
+    .PARAMETER NoRestart
+        Do not immediately restart the service after changing the password.
+
+        **Note that the changes will not go into effect until you restart the SQL Services**
+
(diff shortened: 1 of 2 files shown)


wsmelton (approved):

potatoqualitee:
heck yeah, baw! 🥳

PR #7650, opened 2021-07-29
Get-DbaConnectedInstance / Disconnect-DbaInstance

potatoqualitee:
Proof of concept so far -- totally works

image

### DIFF
--- a/dbatools.psm1
+++ b/dbatools.psm1
@@ -1085,3 +1085,6 @@ $onRemoveScript = {
 }
 $ExecutionContext.SessionState.Module.OnRemove += $onRemoveScript
 Register-EngineEvent -SourceIdentifier ([System.Management.Automation.PsEngineEvent]::Exiting) -Action $onRemoveScript
+
+# Create collection for servers
+$global:connectionhash = @{ }
--- a/dbatools.psm1
+++ b/dbatools.psm1
@@ -1087,4 +1087,4 @@ $ExecutionContext.SessionState.Module.OnRemove += $onRemoveScript
 Register-EngineEvent -SourceIdentifier ([System.Management.Automation.PsEngineEvent]::Exiting) -Action $onRemoveScript
 
 # Create collection for servers
-$global:connectionhash = @{ }
+$script:connectionhash = @{ }
--- a/functions/Connect-DbaInstance.ps1
+++ b/functions/Connect-DbaInstance.ps1
@@ -964,6 +964,7 @@ function Connect-DbaInstance {
 
                 if ($SqlConnectionOnly) {
                     Write-Message -Level Debug -Message "We return only SqlConnection in server.ConnectionContext.SqlConnectionObject"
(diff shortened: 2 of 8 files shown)


andreasjordan:
What about connections to the same instance with different Credentials or with different parameters?

potatoqualitee:
we can make the "key" using psboundparameters 🤔

potatoqualitee:
but then it may become unrecognizable

andreasjordan:
Maybe an array of connection objects as the value.

potatoqualitee:
ahh! without an index!

potatoqualitee:
boo, can't detect dupes

potatoqualitee:
ahh connectionstring as an index!

potatoqualitee:
love itttttt 

image

andreasjordan:
image

potatoqualitee:
Interesting! I didn't realize that admin wasn't parsable, I'll handle it

andreasjordan:
Problem 1: Disconnect should also remove the entry from the hash.
Problem 2: If you open more that one non-pooled connection, you only have one entry in the hash for more that one connection, so only one gets closed.
Problem 3: This only works for non-pooled connection. Pooled connections can only be closed with `Clear-DbaConnectionPool`.

potatoqualitee:
1. Yep, realized that after and added a note for myself when `Get` showed too much
2. No idea how to handle this
3. Yep, it's in the description docs for Disconnect

andreasjordan:
About 2.: What about an array as the value to the key connectionstring:

Instead of `$script:connectionhash[$server.ConnectionContext.ConnectionString] = $server` something like:
```
if (-not $script:connectionhash[$server.ConnectionContext.ConnectionString]) {
    $script:connectionhash[$server.ConnectionContext.ConnectionString] = @( )
}
$script:connectionhash[$server.ConnectionContext.ConnectionString] += $server
```

You could also take the server name as key then:
```
if (-not $script:connectionhash[$server.Name]) {
    $script:connectionhash[$server.Name] = @( )
}
$script:connectionhash[$server.Name] += $server
```

potatoqualitee:
Mmm check it out? 

```
Import-Module ./dbatools.psm1 -Force; Get-DbaDatabase -SqlInstance workstation | Out-null; $a = Connect-DbaInstance -SqlInstance localhost -SqlConnectionOnly; Get-DbaDatabase -SqlInstance workstation | Out-null; Get-DbaConnectedInstance
```

andreasjordan:
As the internal key is still the connection string, it might look like this:
image

So I have one connection with connections string A and two connections with connection string B.
It's ok, but might confuse the user.

potatoqualitee:
sweet, check it now! 

image

andreasjordan:
Yes, looks good!

potatoqualitee:
Booom! Will merge once tests pass.

PR #7654, opened 2021-07-30
Internal: Better test for mutually exclusive parameters (part 1)

andreasjordan:
Don't know if it is really a bug fix, but we should better test for mutually exclusive parameters.
See #6885 for background info.
There are more commands affected, so this is only part 1.

### DIFF
--- a/functions/Add-DbaAgListener.ps1
+++ b/functions/Add-DbaAgListener.ps1
@@ -95,7 +95,7 @@ function Add-DbaAgListener {
         [switch]$EnableException
     )
     process {
-        if (Test-Bound -Not SqlInstance, InputObject) {
+        if (Test-Bound SqlInstance, InputObject -Not -Min 1 -Max 1) {
             Stop-Function -Message "You must supply either -SqlInstance or an Input Object"
             return
         }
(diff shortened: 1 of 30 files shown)


potatoqualitee:
This seems to introduce unnecessary complexities? I'm not sure why it's better.

andreasjordan:
Have a look at #6885. If we allow piped input and SqlInstance, and use `if ($SqlInstance) { $InputObject += Get-DbaXYZ }` inside the process block, the `foreach ($object in $InputObject) { }` loop runs more times than intended. 

So I would suggest to change most commands to allow EITHER piped input OR (so XOR) SqlInstance. Then it is save to set $InputObject using a Get-DbaXYZ using SqlInstance.

This pull request was intended to discuss and find the best solution. As an alternative, we could use different parameter sets in the commands. Then PowerShell itself takes care about this. But the error message is not so nice. So this solution here is user friendly. And maybe I'm wrong, but the error message "You must supply either -SqlInstance or an Input Object" means XOR, right? So the test should do that as well.

But I'm totally open do not change that at all or use a different solution. Just wanted to work again on #6885.

potatoqualitee:
> we could use different parameter sets in the commands

We decided as a team against using parametersets in most commands because the resulting error that PS gives is super unhelpful, so instead, we handle them in the code itself and give a helpful error message. I could probably do XOR but I'm not sure it's worth it to add a breaking change? The current implementation is loose, but PowerShell is loose too, yenno?

Issue #7319, opened 2021-05-15
We should really remove Connect-SqlInstance and replace it with Connect-DbaInstance


andreasjordan:
This is the only code inside of Connect-SqlInstance:

https://github.com/sqlcollaborative/dbatools/blob/de024aaa616fe37de7c8eedb796a2352ab08529a/internal/functions/Connect-SqlInstance.ps1#L58-L75

For already created SMOs (`$SqlInstance.InputObject.GetType().Name -eq 'Server'`) to Azure (`$SqlInstance.InputObject.DatabaseEngineType -eq "SqlAzureDatabase"`) it does test for `$AzureUnsupported` and would fail. Inside of Connect-DbaInstance, this is only done in the new code path, not in the old one. So we would have to address this.
Here is the reference to the new code path:
https://github.com/sqlcollaborative/dbatools/blob/de024aaa616fe37de7c8eedb796a2352ab08529a/functions/Connect-DbaInstance.ps1#L821-L823

For all other cases it just starts Connect-DbaInstance with the bound parameters and sets `-ClientName (Get-DbatoolsConfigValue -FullName 'sql.connection.clientname')` which is the default there anyway (`[string]$ClientName = (Get-DbatoolsConfigValue -FullName 'sql.connection.clientname')`).
Problem here: Connect-SqlInstance has a parameter `[switch]$NonPooled`, Connect-DbaInstance calls it `[switch]$NonPooledConnection` (without an alias - as far as I see). So this would be a bug, right?

wsmelton:
> Problem here: Connect-SqlInstance has a parameter [switch]$NonPooled, Connect-DbaInstance calls it [switch]$NonPooledConnection (without an alias - as far as I see). So this would be a bug, right?

It would have been around 4 years ago, but use of that parameter was removed from the functions we had it in. You should not find any function calling it with the internal or public function.

andreasjordan:
Let me do a little bit of research for you:
https://github.com/sqlcollaborative/dbatools/blob/048a3d8dd9d343125e6c474ce095924c5bbd5f39/functions/Get-DbaDbExtentDiff.ps1#L90
https://github.com/sqlcollaborative/dbatools/blob/048a3d8dd9d343125e6c474ce095924c5bbd5f39/functions/Install-DbaMaintenanceSolution.ps1#L303

But I have to correct myself: As "-NonPooled" is recognized as a short form of "-NonPooledConnection", everything works just fine. I would only consider it a bad design.

[...]

PR #7694, opened 2021-08-07
Internal: Replace Connect-SqlInstance by Connect-DbaInstance (part 2)

andreasjordan:
This is the second step for #7319.

@potatoqualitee Please have a look at Copy-DbaAgentServer and Copy-DbaResourceGovernor. I have not changed them as they have multiple connections to the destination. Should we remove them and instead .Refresh() the SMO?

### DIFF
--- a/functions/Add-DbaAgReplica.ps1
+++ b/functions/Add-DbaAgReplica.ps1
@@ -173,9 +173,9 @@ function Add-DbaAgReplica {
 
         foreach ($instance in $SqlInstance) {
             try {
-                $server = Connect-SqlInstance -SqlInstance $instance -SqlCredential $SqlCredential -MinimumVersion 11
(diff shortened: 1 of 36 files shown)


potatoqualitee:
Thanks for the PR! I'd prefer leaving the reconnect on AgentServer because it'd be a lot of refreshes in a specific order since it touches a number of parts. As for ResourceGov, not sure why that was done there, but I'd just leave it too.

andreasjordan:
> Thanks for the PR! I'd prefer leaving the reconnect on AgentServer because it'd be a lot of refreshes in a specific order since it touches a number of parts. As for ResourceGov, not sure why that was done there, but I'd just leave it too.

Ok, updated those two commands as well.

potatoqualitee:
merci 🙇🏼

PR #7695, opened 2021-08-08
Get-DbaInstanceInstallDate - Fix failback column check

### LINKED ISSUE #7675: Get-DbaInstanceInstallDate does not return a date for some instances

Yes

A date failed to return without error.  

### Steps to Reproduce

```powershell
```
#Requires -Version '7.0'
#Requires -Module @{ ModuleName = 'DbaTools'; ModuleVersion = '1.1.6' }
Get-DbaInstanceInstallDate -SqlInstance 'xxxx\xxxx' -IncludeWindows

ComputerName       : xxxx
InstanceName       : xxxx
SqlInstance        : xxxx\xxxx
SqlInstallDate     :
WindowsInstallDate : 2020-01-15 14:02:19.000

Yes

### Other details or mentions

I have DBATools 1.1.6.  The install for 1.1.7 fails to download at this time.  

Out of a 110 instances, only 3 fail to return an install date.  All three are from builds by the same vendor.  They might have done the image sysprep method for the install.  Anyway, the query "SELECT * FROM sys.server_principals WHERE sid = 0x010100000000000512000000" fails to return anything.  The other method "SELECT schemadate FROM sysservers" for versions prior to 9 does work.  

It appears that if the first attempt does not return a value for version 9+ cases, the second method should be used.  

VS Code (integrated terminal)

### PowerShell Host Version

PSVersion                      7.1.3
GitCommitId                    7.1.3
OS                             Microsoft Windows 10.0.19041

Microsoft SQL Server 2019 (RTM-GDR) (KB4517790) - 15.0.2070.41 (X64)   Oct 28 2019 19:56:59   Copyright (C) 2019 Microsoft Corporation  Standard Edition (64-bit) on Windows Server 2019 Standard 10.0 <X64> (Build 17763: ) (Hypervisor)

potatoqualitee:
fixes #7675 for real

### DIFF
--- a/functions/Get-DbaInstanceInstallDate.ps1
+++ b/functions/Get-DbaInstanceInstallDate.ps1
@@ -99,7 +99,7 @@ function Get-DbaInstanceInstallDate {
             if (-not $sqlInstallDate -or $sqlInstallDate -is [System.DBNull]) {
                 Write-Message -Level Verbose -Message "Trying again to get Install Date for: $instance"
                 $sql = "SELECT schemadate FROM sysservers"
-                [DbaDateTime]$sqlInstallDate = $server.Query($sql, 'master', $true).create_date
+                [DbaDateTime]$sqlInstallDate = $server.Query($sql, 'master', $true).schemadate
             }
 
             $WindowsServerName = $server.ComputerNamePhysicalNetBIOS
--- a/functions/Get-DbaInstanceInstallDate.ps1
+++ b/functions/Get-DbaInstanceInstallDate.ps1
@@ -93,7 +93,7 @@ function Get-DbaInstanceInstallDate {
             } else {
                 Write-Message -Level Verbose -Message "Getting Install Date for: $instance"
                 $sql = "SELECT schemadate FROM sysservers"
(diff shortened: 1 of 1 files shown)


andreasjordan:
Are you aware that sys.sysservers is depricated?
https://docs.microsoft.com/en-us/sql/relational-databases/system-compatibility-views/sys-sysservers-transact-sql

And line 96 still has to be changed (create_date -> schemadate).

And I don't see a difference in lines 101/102 compaired to 95/96...

potatoqualitee:
> Are you aware that sys.sysservers is deprecated?

Yes, we try to support really weird setups and in this case, very few people will ever even hit that script block because `sqlInstallDate` won't be null (unless sql sysprep is used in a weird way?)

> And line 96 still has to be changed (create_date -> schemadate).

Oops, thanks! fixed

> And I don't see a difference in lines 101/102 compaired to 95/96...

Basically, it helps if line 91 is null.

andreasjordan (approved):

potatoqualitee:
thank you for the approval 👍🏼

PR #7697, opened 2021-08-08
Use [System.Environment]::NewLine to normalize line endings

### LINKED ISSUE #7689: Export-DbaDbRole - Normalize Export File Line Endings
### Summarize Functionality

Convert `$prefix` to use `` `r`n `` line endings like the SQL Server uses and like the `$BatchSeperator` uses within the lines below.
https://github.com/sqlcollaborative/dbatools/blob/3359ca3afd5792bd957a12d625a43c35eaca96f3/functions/Export-DbaDbRole.ps1#L370-L379

No

### Technical Details

See above. Use `` `r`n `` instead of `` `n `` for the `$prefix` definition. 

This functionality works for `Export-DbaDbUser`. This feature would prevent the below message when opening the `.sql` file within SQL Server Management Studio.

image

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #7689 )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### DIFF
--- a/functions/Export-DbaDbRole.ps1
+++ b/functions/Export-DbaDbRole.ps1
@@ -357,6 +357,8 @@ function Export-DbaDbRole {
     end {
         if (Test-FunctionInterrupt) { return }
 
+        $eol = [System.Environment]::NewLine
+
         $timeNow = $(Get-Date -Format (Get-DbatoolsConfigValue -FullName 'Formatting.DateTime'))
         foreach ($dbRole in $roleCollection) {
             $instanceName = $dbRole.Instance
@@ -367,20 +369,20 @@ function Export-DbaDbRole {
             if ($NoPrefix) {
                 $prefix = $null
             } else {
-                $prefix = "/*`n`tCreated by $executingUser using dbatools $commandName for objects on $instanceName.$databaseName at $timeNow`n`tSee https://dbatools.io/$commandName for more information`n*/"
+                $prefix = "/*$eol`tCreated by $executingUser using dbatools $commandName for objects on $instanceName.$databaseName at $timeNow$eol`tSee https://dbatools.io/$commandName for more information$eol*/"
             }
 
             if ($BatchSeparator) {
-                $sql = $dbRole.SQL -join "`r`n$BatchSeparator`r`n"
+                $sql = $dbRole.SQL -join "$eol$BatchSeparator$eol"
                 #add the final GO
-                $sql += "`r`n$BatchSeparator"
+                $sql += "$eol$BatchSeparator"
             } else {
                 $sql = $dbRole.SQL
             }
 
             if ($Passthru) {
                 if ($null -ne $prefix) {
-                    $sql = "$prefix`r`n$sql"
+                    $sql = "$prefix$eol$sql"
                 }
                 $sql
             } elseif ($Path -Or $FilePath) {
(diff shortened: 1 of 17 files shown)


andreasjordan:
Please let some people review this as I have not tested it.

functions/Export-DbaUser.ps1
```diff
@@ -489,12 +492,12 @@ function Export-DbaUser {
                         $sql = "$useDatabase $outsql"
                     } else {
                         if ($useDatabase) {
-                            $sql = "$useDatabase`r`nGO`r`n" + ($outsql -join "`r`nGO`r`n")
+                            $sql = "$useDatabase${eol}GO$eol" + ($outsql -join "${eol}GO$eol")
                         } else {
-                            $sql = $outsql -join "`r`nGO`r`n"
+                            $sql = $outsql -join "${eol}GO$eol"
```
potatoqualitee, line 497: Thank you so much for this PR! What is `${eol}`? I'm not familiar with this syntax

andreasjordan: In `$eolGO` the name of the variable would include the GO. The `${}` is a bit shorter than the alternative `$($eol)`. Should I change that to `$($eol)`?

potatoqualitee: ahh! Interesting. yes, $($eol)) would be the standard way we handle these, thank you.

andreasjordan: Will do the change later today...

andreasjordan: And changed...

functions/Export-DbaDbRole.ps1
```diff
@@ -367,28 +369,28 @@ function Export-DbaDbRole {
             if ($NoPrefix) {
                 $prefix = $null
             } else {
-                $prefix = "/*`n`tCreated by $executingUser using dbatools $commandName for objects on $instanceName.$databaseName at $timeNow`n`tSee https://dbatools.io/$commandName for more information`n*/"
```
potatoqualitee, line 370: This works

functions/Export-DbaScript.ps1
```diff
@@ -168,6 +168,8 @@ function Export-DbaScript {
                 Write-Message -Level Warning -Message "Setting ScriptBatchTerminator to true and also having BatchSeperarator as an empty or null string may produce unintended results."
             }
         }
+
+        $eol = [System.Environment]::NewLine
```
potatoqualitee, line 172: something must be missing here -- export-dbainstance exports user objects and that particular script causes an issue

image

functions/Export-DbaXESession.ps1
```diff
@@ -178,31 +178,33 @@ function Export-DbaXESession {
         }
     }
     end {
+        $eol = [System.Environment]::NewLine
```
potatoqualitee, line 181: works!

functions/Export-DbaScript.ps1
andreasjordan: It's the line before a CREATE PROCEDURE:

image

andreasjordan: I think the single LF is in the code. If I use SSMS to "Modify" the procedure I get the same message. So we would have to change that in the resultung $scriptInFull. Don't know if we really want to do that.

potatoqualitee: So you think something like a `$scriptInFull.Replace("`r",$eof)` is too dangerous or unnecessary? I can't decide.

andreasjordan: Wouldn't that change "rn" into "rrn"?

potatoqualitee: yah sorry i was riffing, whatever it is that would replace LF (isn't ``` `r`n ``` CLRF?) with $eof

andreasjordan: This would work:
```
$scriptInFull = $scriptInFull -replace "(`r)?`n", $eol
```

Should we run that if a special config value is set? Or in any case? 
And this would have to be added in multiple places inside of Export-DbaScript and would not effect those branches that write to a file via Scripter directly.

potatoqualitee: @niphlod what do you think? I'm leaning not at all or any case but mostly any case.

potatoqualitee:
I'll merge this one w/o the replace then if niph agrees we should replace, then another PR can address it 👍🏼

PR #7706, opened 2021-08-10
Internal: Replace Connect-SqlInstance by Connect-DbaInstance (part 4)

andreasjordan:
This is the forth step for #7319.


potatoqualitee:
Nice! Thank you

PR #7720, opened 2021-08-12
Internal: Replace Connect-SqlInstance by Connect-DbaInstance (part 7)

andreasjordan:
This is the seventh step for #7319.

Plus a bit of refactoring of:
* Set-DbaAgentSchedule.ps1
* Set-DbaMaxDop.ps1
* Show-DbaDbList.ps1
* Start-DbaMigration.ps1 
* Test-DbaDbQueryStore.ps1
* Test-DbaDiskAlignment.ps1
* Test-DbaDiskAllocation.ps1
* Test-DbaDiskSpeed.ps1

Only 64 files left...

### DIFF
--- a/functions/Test-DbaDiskAlignment.ps1
+++ b/functions/Test-DbaDiskAlignment.ps1
@@ -174,24 +174,20 @@ function Test-DbaDiskAlignment {
                     if ($NoSqlCheck -eq $false) {
                         $sqldisk = $false
 
-                        foreach ($SqlInstance in $SqlInstances) {
+                        foreach ($instance in $SqlInstances) {
                             try {
-                                if ($null -ne $SqlCredential) {
-                                    $smoserver = Connect-SqlInstance -SqlInstance $SqlInstance -SqlCredential $SqlCredential
-                                } else {
-                                    $smoserver = Connect-SqlInstance -SqlInstance $SqlInstance # win auth
-                                }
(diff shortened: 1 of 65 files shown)


functions/Test-DbaDiskAlignment.ps1
```diff
@@ -174,24 +174,20 @@ function Test-DbaDiskAlignment {
...
+                                Stop-Function -Message "Failure" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
+                            }
+
+                            $sql = "Select count(*) as Count from sys.master_files where physical_name like '$diskname%'"
+                            Write-Message -Level Verbose -Message "Query is: $sql" -FunctionName $FunctionName
+                            Write-Message -Level Verbose -Message "SQL Server is: $instance." -FunctionName $FunctionName
+                            $sqlcount = $server.Query($sql).Count
+                            if ($sqlcount -gt 0) {
+                                $sqldisk = $true
+                                break
```
potatoqualitee, line 190: thanks for all of this! i wonder if we should create an issue to look for breaks and see where they're acceptable and where they are not. this is one of our older commands, so i assume the break should be a return instead.

potatoqualitee:
Looks fab! thank you

functions/Test-DbaDiskAlignment.ps1
andreasjordan: The break is ok here. It's inside of a test if the drive has an instance. If I found one instance, I can skip testing the other instances.

PR #7723, opened 2021-08-16
Internal: Replace Connect-SqlInstance by Connect-DbaInstance (part 8)

### LINKED ISSUE #7319: We should really remove Connect-SqlInstance and replace it with Connect-DbaInstance

andreasjordan:
This is the last step and finally fixes #7319.

Please have a look at the scripts in internal/dynamicparams/:
* Do we need the `-ErrorAction Stop`?
* Why is the connection made if the resulting `$serverObject` is never used?
* Should we change processHostname.ps1, processProgram.ps1 and processSpid.ps1 to align them with the others? They use the resulting `$instance` (which should be named `$server`?).

Please have a look at Get-BackupAncientHistory.ps1, I think the command has more issues. But as it is only called once from Backup-DbaDatabase we should maybe remove all the functionality that is not used.

Some internal commands use try catch around Connect-DbaInstance, some don't. As they are internal commands, I think we should rely on the calling external command that SqlInstance is an already opened connection and to not use try catch in the internal commands. We could also think of not using Connect-DbaInstance again and just say `$server = $SqlInstance.InputObject`. I'm not sure which is the best way, but we should have one rule for all the internal commands.


wsmelton:
> Do we need the -ErrorAction Stop?

Yes

wsmelton:
A note that TEPP has to be manually tested everywhere that code is referenced or expected because we do not have tests for any of it (not sure it can be).

andreasjordan:
I just testet TEPP and I have seen no issues.
We could try to do an automatic test to see if the cache is filled - that should work with no problem.
But I don't have an idea how to see if the cache is used while using <TAB> with a command.

potatoqualitee:
I tested and cache is used. Unfortunately, I don't have the bandwidth to address the questions asked and will rely on others to Approve 👍🏼

andreasjordan:
So who can approve this? @niphlod ?

niphlod (approved):

niphlod:
sorry, somewhat got lost in the notifications, thanks for the ping @andreasjordan

potatoqualitee:
sweet, thanks to both of you 🙇🏼

PR #7724, opened 2021-08-16
Internal: Added missing Test-FunctionInterrupt calls

andreasjordan:
There are some commands, that use Stop-Command in the begin block but don't test for the function interrupt so that the process block will run even it should not.

Also changed some `break` into `return` or added some missing `return`.
Also corrected some Stop-Command calls.


potatoqualitee:
🤩  so good!

PR #7726, opened 2021-08-17
Invoke-DbaQuery - Request non pooled connection when using windows credentials

### LINKED ISSUE #7725: Invoke-DbaQuery SqlCredential param with windows credential not honored

Yes

No error, but when I used a query of "SELECT SUSER_SNAME()" I got my current creds instead of the ones I'd passed in.

### Steps to Reproduce

```powershell
```
$cred = get-credential #to get alternate creds
Invoke-DbaQuery -SqlInstance servername -Query "SELECT SUSER_SNAME()" -SqlCredential $cred

# This returns my currently logged in user.

Connect-DbaInstance -SqlInstance servername -SqlCredential $cred
# This returns my credentials as expected

Yes

### Other details or mentions

_No response_

PowerShell Core (pwsh.exe)

### PowerShell Host Version

7.1.4

Irrelevant, but was going against a SQL 2014 server with the latest CU.

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #7725 )

See issue for details.

### DIFF
--- a/functions/Invoke-DbaQuery.ps1
+++ b/functions/Invoke-DbaQuery.ps1
@@ -372,12 +372,12 @@ function Invoke-DbaQuery {
             Write-Message -Level Debug -Message "SqlInstance passed in, will work on: $instance"
             try {
                 $connDbaInstanceParams = @{
-                    SqlInstance   = $instance
-                    SqlCredential = $SqlCredential
-                    Database      = $Database
+                    SqlInstance         = $instance
+                    SqlCredential       = $SqlCredential
+                    Database            = $Database
+                    NonPooledConnection = $true
                 }
                 if ($ReadOnly) {
-                    # TODO: This will not work, if SqlInstance is already a server object
                     $connDbaInstanceParams.ApplicationIntent = "ReadOnly"
                 }
                 $server = Connect-DbaInstance @connDbaInstanceParams
--- a/functions/Invoke-DbaQuery.ps1
+++ b/functions/Invoke-DbaQuery.ps1
@@ -406,6 +406,8 @@ function Invoke-DbaQuery {
             } catch {
                 Stop-Function -Message "[$instance] Failed during execution" -ErrorRecord $_ -Target $instance -Continue
             }
+            # Close non-pooled connection as this is not done automatically. If it is a reused Server SMO, connection will be opened again automatically on next request.
+            $server | Disconnect-DbaInstance
         }
     }
 
--- a/functions/Invoke-DbaQuery.ps1
+++ b/functions/Invoke-DbaQuery.ps1
@@ -372,13 +372,17 @@ function Invoke-DbaQuery {
             Write-Message -Level Debug -Message "SqlInstance passed in, will work on: $instance"
             try {
                 $connDbaInstanceParams = @{
-                    SqlInstance         = $instance
-                    SqlCredential       = $SqlCredential
-                    Database            = $Database
-                    NonPooledConnection = $true
+                    SqlInstance   = $instance
+                    SqlCredential = $SqlCredential
+                    Database      = $Database
                 }
                 if ($ReadOnly) {
                     $connDbaInstanceParams.ApplicationIntent = "ReadOnly"
+                }
+                # Test for a windows credential and request a non-pooled connection in that case to keep the security context (see #7725 for details)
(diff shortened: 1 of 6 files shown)


functions/Invoke-DbaQuery.ps1
```diff
@@ -372,12 +372,12 @@ function Invoke-DbaQuery {
             Write-Message -Level Debug -Message "SqlInstance passed in, will work on: $instance"
             try {
                 $connDbaInstanceParams = @{
-                    SqlInstance   = $instance
```
potatoqualitee, line 375: This change will impact all runs. To minimize impact/change, can you do a test for SqlCredential.UserName match \ or @ and only set it when there's a match?

andreasjordan: Yes, will do that.

andreasjordan: After some analysis I have to say: It's not that easy. $instance might be a string, and then I can test SqlCredential.UserName. But: It might by an Azure-AD Account and we would not want a non pooled connection for that, or do we? On the other side $instance might be a Server SMO, and then I would have to test the ConnectionContext.
I also had the idea to do it as in the legacy code path: If a windows credential is used, just force a non pooled connection. But then we have another problem: We need to explicitly close non pooled connections at the end of the command requesting it. But if we don't know if we get a pooled or nonpooled connection, that is impossible and would affect any command.

So now I think that - until we find a better solution and maybe rewrite Invoke-DbaAsync - it is the best way to force a non pooled connection in any case and close it at the end of the command (this is missing now, but I will push that change in a minite).

I'm still open to other solutions, but at the moment I don't see a better one.

potatoqualitee: I think we should focus on the most likely cases. If a SqlCredential is supplied, parse it unless database.microsoft.net (or whatever the name is) is in the SqlInstance name. If someone passes an SMO object, then we don't handle it.

If others disagree, I'm open to that. I'd just like any Invoke-DbaAsync changes to be run past @niphlod

andreasjordan: Ok, not testing for Azure yet, but for windows credential. Is this the way you would like to go?

potatoqualitee:
I like it and will ping @niphlod on Slack for confirmation.

tests/Invoke-DbaQuery.Tests.ps1
```diff
@@ -153,7 +153,7 @@ SELECT @@servername as dbname
         }
         # Filter out the Verbose messages of the command Connect-DbaInstance
         $results = $results | Where-Object Out -NotMatch '\[Connect-DbaInstance\]'
-        $results.Length | Should -Be 7 # 6 'messages' plus the actual resultset
+        $results.Length | Should -BeIn 7, 8 # 6 'messages' plus the actual resultset plus maybe a message from Disconnect-DbaInstance if non-pooled connection
```
niphlod, line 156: dunno if this got added for a specific reason but Disconnect-DbaInstance output shouldn't be in the result pipeline ... why does it get there ?

niphlod (changes requested):

tests/Invoke-DbaQuery.Tests.ps1
potatoqualitee: Adding the disconnect to the `end {}` _may_ help this

andreasjordan: The disconnect must be inside of the `foreach ($instance in $SqlInstance) {` loop, so moving to end is no option.
But I just had a look at line 155: We should include Disconnect-DbaInstance there to filter it out, don't we?

andreasjordan: I changed the test - please have a look.

niphlod: okay, now it's clear(er) . But given the fancy use of -Verbose specifically in Invoke-DbaQuery, I'd still mute fixedly anything coming out of Disconnect-DbaInstance in this. 
Theoretically all it should output should come from Invoke-DbaQuery only.
155 too has slipped in and shouldn't be there (messages from Connect-DbaInstance)

niphlod (approved):

potatoqualitee:
Woo! Thank you both 🙇🏼

PR #7751, opened 2021-08-31
Add ReadOnlyRoutingList param to Set-DbaAgReplica

### LINKED ISSUE #7732: Allow setting read-only routing list through Set-DbaAgReplica
### Summarize Functionality

Looking through the list of options in [the docs](https://docs.microsoft.com/en-us/sql/t-sql/statements/alter-availability-group-transact-sql?view=sql-server-ver15), and comparing it with the list of options for Set-DbaAgReplica, the only one missing seems to be READ_ONLY_ROUTING_LIST.

I'd like to add a new parameter (`ReadOnlyRoutingList`?) that would allow this to be set.

No

### Technical Details

The AvailabilityReplica SMO object has a `SetLoadBalancedReadOnlyRoutingList()` method that allows this to be set. As the name of the method implies, it also allows for the setting of a load balanced read-only routing list (introduced in SQL 2016).

The method takes as a parameter a `[List[IList[string]]]`, which isn't the most straightforward thing to create. I have code working locally that will allow an array of arrays to be used as the parameter value. Something like this:

```
$r | Set-DbaAgReplica -ReadOnlyRoutingList ('foo', 'bar'), 'baz';
```

which would be equivalent to:

```
alter availability group myAG
   modify replica baz (
      primary_role (
         read_only_routing_list = (('foo', 'bar'), 'baz')
      )
   );
```

ben-thul:
## Type of Change

 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
Add the ability to set a read-only routing list for an AG replica.

### Approach

Adds a `-ReadOnlyRoutingList` parameter to `Set-DbaAgReplica`.

### Commands to test

`Set-DbaAgReplica`

---
Resolves #7732.

### DIFF
--- a/functions/Set-DbaAgReplica.ps1
+++ b/functions/Set-DbaAgReplica.ps1
@@ -1,4 +1,3 @@
-using namespace System.Collections.Generic;
 function Set-DbaAgReplica {
     <#
     .SYNOPSIS
@@ -96,15 +95,9 @@ function Set-DbaAgReplica {
 
     .EXAMPLE
         PS C:\> Get-DbaAgReplica -SqlInstance sql2016 -Replica Replica1 |
-        >> Set-DbaAgReplica -ReadOnlyRoutingList Replica2,Replica3;
+        >> Set-DbaAgReplica -ReadOnlyRoutingList Replica2, Replica3
 
         Equivalent to running "ALTER AVAILABILITY GROUP... MODIFY REPLICA... (READ_ONLY_ROUTING_LIST = ('Replica2', 'Replica3'));"
-
-    .EXAMPLE
-        PS C:\> Get-DbaAgReplica -SqlInstance sql2016 -Replica Replica1 |
-        >> Set-DbaAgReplica -ReadOnlyRoutingList @(,('Replica2','Replica3'));
-
-        Equivalent to running "ALTER AVAILABILITY GROUP... MODIFY REPLICA... (READ_ONLY_ROUTING_LIST = (('Replica2', 'Replica3')));"
     #>
     [CmdletBinding(SupportsShouldProcess, ConfirmImpact = 'Medium')]
     param (
@@ -131,6 +124,11 @@ function Set-DbaAgReplica {
         [Microsoft.SqlServer.Management.Smo.AvailabilityReplica]$InputObject,
         [switch]$EnableException
     )
+    begin {
+        if ($ReadOnlyRoutingList) {
+            $null = Add-Type -AssemblyName System.Collections
+        }
+    }
     process {
         if (Test-Bound -Not SqlInstance, InputObject) {
             Stop-Function -Message "You must supply either -SqlInstance or an Input Object"
@@ -195,11 +193,11 @@ function Set-DbaAgReplica {
                     }
 
                     if ($ReadOnlyRoutingList) {
-                        $rorl = [List[IList[string]]]::new();
-                        $ReadOnlyRoutingList | foreach {
-                            $rorl.Add([List[string]] $_);
+                        $rorl = New-Object System.Collections.Generic.List[System.Collections.Generic.IList[string]]
+                        foreach ($rolist in $ReadOnlyRoutingList) {
+                            $null = $rorl.Add([System.Collections.Generic.List[string]] $rolist)
                         }
-                        $agreplica.SetLoadBalancedReadOnlyRoutingList($rorl);
+                        $agreplica.SetLoadBalancedReadOnlyRoutingList($rorl)
                     }
 
                     if ($SessionTimeout) {
--- a/functions/Set-DbaAgReplica.ps1
+++ b/functions/Set-DbaAgReplica.ps1
@@ -197,7 +197,7 @@ function Set-DbaAgReplica {
                         foreach ($rolist in $ReadOnlyRoutingList) {
                             $null = $rorl.Add([System.Collections.Generic.List[string]] $rolist)
                         }
-                        $agreplica.SetLoadBalancedReadOnlyRoutingList($rorl)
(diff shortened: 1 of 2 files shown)


ben-thul:
I'm happy to write tests, but don't know what the the test AG topology looks like and which I'd need to set a read-only routing list.

andreasjordan:
Looks good to me. 

Two points:
* We have a parameter ReadonlyRoutingConnectionUrl where the only has a small "o", you call the parameter "ReadOnlyRoutingList". Maybe change that to "ReadonlyRoutingList"? UPDATE: The casing is like the SMO property - so we should leave it like this.
* What's the output like if you supply a wrong format? Maybe have a try-catch around the "$ReadOnlyRoutingList | foreach { $rorl.Add([List[string]] $_); }" with a nice message in Stop-Function?

About tests: We don't have a nive AG in the Test environment - so can you just show it locally and provide screenshots?

andreasjordan:
What about the related command Add-DbaAgReplica? We should add that parameter there too, don't we? And then we could / should add it to New-DbaAvailabilityGroup as well. 

Or we decide that this is rarely used and thus only in the Set- command (and maybe in the Add- command).
@potatoqualitee - What do you think?

functions/Set-DbaAgReplica.ps1
```diff
@@ -1,3 +1,4 @@
+using namespace System.Collections.Generic;
```
potatoqualitee, line 1: Thanks for the PR, Ben! Unfortunately, `using` is not valid PSv3 syntax and we cannot drop v3 support. Is there a way you can do this using Add-Type, perhaps?

potatoqualitee: Is it required?

potatoqualitee (changes requested):

potatoqualitee:
yes, we can add it to those as well, @andreasjordan. i'm not sure about our AG setup. I recall it being very minimal. As such, agreed that screenshots would suffice.

ben-thul:
> * What's the output like if you supply a wrong format? Maybe have a try-catch around the "$ReadOnlyRoutingList | foreach { $rorl.Add([List[string]] $_); }" with a nice message in Stop-Function?

In the case where you supply a replica that isn't part of the AG, the AG itself gives this:

> WARNING: [05:20:44][Set-DbaAgReplica] Failure | The availability replica 'foo' specified in the READ_ONLY_ROUTING_LIST for availability replica '«actual replica name»' does not exist. Only availability replicas that belong to the specified
availability group '«actual AG name»' can be added to this list. To get the names of availability replicas in a given availability group, select replica_server_name from sys.availability_replicas and name from sys.availability_groups. For
more information, see SQL Server Books Online.

I can try to throw something exotic at it to try to break it before it gets to the AG. Though what is there should handle anything I can think of that resembles a list of strings.

functions/Set-DbaAgReplica.ps1
ben-thul: Is it required?

> Is it required?

It's not required, but makes this:

```
[List[IList[string]]]::new()
```

become this:

```
[System.Collections.GenericList[System.Collections.GenericIList[string]]]::new()
```

> Is there a way you can do this using Add-Type, perhaps?

Maybe! I'll play around with it. Worst case, I change to what I have above (i.e. full-pathing the type name). Thanks for the suggestion!

ben-thul:
> What about the related command Add-DbaAgReplica? We should add that parameter there too, don't we? And then we could / should add it to New-DbaAvailabilityGroup as well.
> 

Should be a copy/paste job to add to `Add-DbaAgReplica`. Will have a look at `New-DbaAvailabilityGroup` but I have a similar expectation.

functions/Set-DbaAgReplica.ps1
```diff
@@ -177,6 +194,14 @@ function Set-DbaAgReplica {
                         $agreplica.SeedingMode = $SeedingMode
                     }
 
+                    if ($ReadOnlyRoutingList) {
+                        $rorl = [List[IList[string]]]::new();
```
potatoqualitee, line 198: ah, thank you for highlighting this! we don't accept `::new` because it's not v3 compat. So this would have to be some sort of ...array? $rorl = @() may suffice.

functions/Set-DbaAgReplica.ps1
```diff
@@ -89,6 +93,18 @@ function Set-DbaAgReplica {
         PS C:\> Get-DbaAgReplica -SqlInstance sql2016 | Out-GridView -Passthru | Set-DbaAgReplica -BackupPriority 5000
 
         Sets the backup priority to 5000 for the selected availability groups.
+
+    .EXAMPLE
+        PS C:\> Get-DbaAgReplica -SqlInstance sql2016 -Replica Replica1 |
+        >> Set-DbaAgReplica -ReadOnlyRoutingList Replica2,Replica3;
+
```
potatoqualitee, line 100: also, please note that semicolons are not required and we prefer not to use them, as it keeps the code tidier.

potatoqualitee:
working on this and testing

functions/Set-DbaAgReplica.ps1
potatoqualitee: fixed it, that was fun! it's so rare to have to use such complex types. i'll definitely use this as a template when I have to do something similar in the future, thankks.

potatoqualitee:
boom, merged! Thanks so much for this feature 👍🏼

functions/Set-DbaAgReplica.ps1
ben-thul: I also learned a lot in re: using complex types in getting the code to where it was/is.

PR #7817, opened 2021-09-29
Get-DbaDbMailProfile produces duplicate results

MikeyBronowski:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes # )
 - [ ] New feature (non-breaking change, adds functionality, fixes # )
 - [ ] Breaking change (effects multiple commands or functionality, fixes # )
 - [x] Ran manual Pester test and has passed (`.\tests\manual.pester.ps1`)
 - [ ] Adding code coverage to existing functionality
 - [x] Pester test is included
 - [ ] If new file reference added for test, has is been added to github.com/sqlcollaborative/appveyor-lab ?
 - [ ] Unit test is included
 - [x] Documentation
 - [ ] Build system

### Purpose

This command produces duplicate results. 
`Get-DbaDbMailProfile -SqlInstance $s1, $s2 -SqlCredential $credential | ft`

```
ComputerName InstanceName SqlInstance  ID Name         Description ForceDeleteForActiveProfiles IsBusyProfile
------------ ------------ -----------  -- ----         ----------- ---------------------------- -------------
localhost    MSSQLSERVER  6f34ddb8eb52 36 MailProfile                                      True         False
localhost    MSSQLSERVER  6f34ddb8eb52 37 MailProfile2                                     True         False
localhost    MSSQLSERVER  bbac38d1a5e9 19 MailProfile                                      True         False
localhost    MSSQLSERVER  bbac38d1a5e9 20 MailProfile2                                     True         False
localhost    MSSQLSERVER  6f34ddb8eb52 36 MailProfile                                      True         False
localhost    MSSQLSERVER  6f34ddb8eb52 37 MailProfile2                                     True         False
localhost    MSSQLSERVER  bbac38d1a5e9 19 MailProfile                                      True         False
localhost    MSSQLSERVER  bbac38d1a5e9 20 MailProfile2                                     True         False
```

and it is supposed to produce half of those
```
ComputerName InstanceName SqlInstance  ID Name         Description ForceDeleteForActiveProfiles IsBusyProfile
------------ ------------ -----------  -- ----         ----------- ---------------------------- -------------
localhost    MSSQLSERVER  6f34ddb8eb52 36 MailProfile                                      True         False
localhost    MSSQLSERVER  6f34ddb8eb52 37 MailProfile2                                     True         False
localhost    MSSQLSERVER  bbac38d1a5e9 19 MailProfile                                      True         False
[...]

PR #7818, opened 2021-09-29
Remove-DbaDbMailProfile - new command

### LINKED ISSUE #4990: Remove-DbaDbMail Profile and Account
---
name: Remove-DbaDbMail Profile and Account
about: Remove Mail Profiles & Accounts

---

### Summary of new feature
It would be useful to be able to remove profiles & accounts prior to use of the Copy-DbaDbMail cmdlet.

This is how i'm handling it right now if it helps:
$profileName = 'SQL_Alerts'
$accountName = 'SQL_Alerts'

function Remove-Profile {
    param (
        $Profilename
    )
    Invoke-Sqlcmd -ServerInstance $destination `
        -Query @"
        EXECUTE msdb.dbo.sysmail_delete_profile_sp 
            @profile_name = '$Profilename' ;
"@ `
    -OutputSqlErrors $true `
    -Verbose
}

function Remove-Account {
    param (
        $AccountName
    )
    Invoke-Sqlcmd -ServerInstance $destination `
        -Query @"
        EXECUTE msdb.dbo.sysmail_delete_account_sp 
        @account_name = '$AccountName' ;
"@ `
    -OutputSqlErrors $true `
    -Verbose
}

0.9.743

MikeyBronowski:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality, fixes #4990 )
 - [x] Pester test is included
 - [x] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose

Answers request #4990 with the first function of this kind.

### Approach

Similar to Remove-DbaDbView

### Commands to test

```
Remove-DbaDbMailProfile -SqlInstance localhost, localhost\namedinstance

Remove-DbaDbMailProfile -SqlInstance localhost -Name MyDatabaseMailProfile

Get-DbaDbMailProfile -SqlInstance SRV1 | Out-GridView -Title 'Select database mail profile(s) to drop' -OutputMode Multiple | Remove-DbaDbMailProfile
```

### DIFF
--- a/functions/Remove-DbaDbMailProfile.ps1
+++ b/functions/Remove-DbaDbMailProfile.ps1
@@ -19,7 +19,7 @@ function Remove-DbaDbMailProfile {
 
         For MFA support, please use Connect-DbaInstance.
 
-    .PARAMETER Name
+    .PARAMETER Profile
         Specifies one or more database mail profile(s) to get. If unspecified, all profiles will be removed.
 
     .PARAMETER ExcludeProfile
@@ -73,7 +73,7 @@ function Remove-DbaDbMailProfile {
         [Parameter(ParameterSetName = 'NonPipeline')]
         [PSCredential]$SqlCredential,
         [Parameter(ParameterSetName = 'NonPipeline')]
-        [string[]]$Name,
+        [string[]]$Profile,
         [Parameter(ParameterSetName = 'NonPipeline')]
         [string[]]$ExcludeProfile,
         [parameter(ValueFromPipeline, ParameterSetName = 'Pipeline', Mandatory = $true)]
--- a/functions/Remove-DbaDbMailProfile.ps1
+++ b/functions/Remove-DbaDbMailProfile.ps1
@@ -1,0 +1,120 @@
+function Remove-DbaDbMailProfile {
+    <#
+    .SYNOPSIS
+        Removes database mail profile(s).
+
+    .DESCRIPTION
+        Removes the database mail profile(s) that have passed through the pipeline.
+
+        If not used with a pipeline, Get-DbaDbMailProfile will be executed with the parameters provided
+        and the returned database mail profile(s) will be removed.
+
+    .PARAMETER SqlInstance
+        The target SQL Server instance or instances.
+
+    .PARAMETER SqlCredential
+        Login to the target instance using alternative credentials. Accepts PowerShell credentials (Get-Credential).
+
+        Windows Authentication, SQL Server Authentication, Active Directory - Password, and Active Directory - Integrated are all supported.
+
+        For MFA support, please use Connect-DbaInstance.
+
+    .PARAMETER Profile
+        Specifies one or more database mail profile(s) to get. If unspecified, all profiles will be returned.
+
+    .PARAMETER ExcludeProfile
+        Specifies one or more database mail profile(s) to exclude.
+
+    .PARAMETER InputObject
+        Allows piping from Get-DbaDbMailProfile.
+
+    .PARAMETER WhatIf
+        Shows what would happen if the command were to run. No actions are actually performed.
+
+    .PARAMETER Confirm
+        Prompts you for confirmation before executing any changing operations within the command.
+        This is the default. Use -Confirm:$false to suppress these prompts.
+
+    .PARAMETER EnableException
+        By default, when something goes wrong we try to catch it, interpret it and give you a friendly warning message.
+        This avoids overwhelming you with "sea of red" exceptions, but is inconvenient because it basically disables advanced scripting.
+        Using this switch turns this "nice by default" feature off and enables you to catch exceptions with your own try/catch.
+
+    .NOTES
+        Tags: DatabaseMail, DBMail, Mail
+        Author: Mikey Bronowski (@MikeyBronowski), https://bronowski.it
+
+        dbatools PowerShell module (https://dbatools.io)
+        Copyright: (c) 2021 by dbatools, licensed under MIT
+        License: MIT https://opensource.org/licenses/MIT
+
+    .LINK
+        https://dbatools.io/Remove-DbaDbMailProfile
+
+    .EXAMPLE
+        PS C:\> Remove-DbaDbMailProfile -SqlInstance localhost, localhost\namedinstance
+
+        Removes all database mail profiles on the localhost, localhost\namedinstance instances.
+
+    .EXAMPLE
+        PS C:\> Get-DbaDbMailProfile -SqlInstance SRV1 | Out-GridView -Title 'Select database mail profile(s) to drop' -OutputMode Multiple | Remove-DbaDbMailProfile
+
+        Using a pipeline this command gets all database mail profiles on SRV1, lets the user select those to remove and then removes the selected database mail profiles.
+    #>
+    [CmdletBinding(SupportsShouldProcess, ConfirmImpact = 'High')]
+    param (
+        [Parameter(ParameterSetName = 'NonPipeline', Mandatory = $true, Position = 0)]
+        [DbaInstanceParameter[]]$SqlInstance,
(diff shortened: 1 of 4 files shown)


MikeyBronowski:
That PR uses parameter alias that has been introduced in #7817

andreasjordan (approved):
Looks good. Thanks for the command.

functions/Remove-DbaDbMailProfile.ps1
```diff
@@ -0,0 +1,125 @@
...
+        Using a pipeline this command gets all database mail profiles on SRV1, lets the user select those to remove and then removes the selected database mail profiles.
+    #>
+    [CmdletBinding(SupportsShouldProcess, ConfirmImpact = 'High')]
+    param (
+        [Parameter(ParameterSetName = 'NonPipeline', Mandatory = $true, Position = 0)]
+        [DbaInstanceParameter[]]$SqlInstance,
+        [Parameter(ParameterSetName = 'NonPipeline')]
+        [PSCredential]$SqlCredential,
+        [Parameter(ParameterSetName = 'NonPipeline')]
+        [string[]]$Name,
```
wsmelton, line 76:
```suggestion
        [string[]]$Profile,
```
The filter input should match the pattern used for the Exclude parameter. `-Profile` then `-ExcludeProfile`, same matching we do with the Get command.

functions/Remove-DbaDbMailProfile.ps1
```diff
@@ -0,0 +1,125 @@
...
+
+    end {
+        # We have to delete in the end block to prevent "Collection was modified; enumeration operation may not execute." if directly piped from Get-DbaDbMailProfile.
+        foreach ($dbMailProfile in $dbMailProfiles) {
+            if ($PSCmdlet.ShouldProcess($dbMailProfile.Parent.Parent.Name, "Removing the database mail profile $($dbMailProfile.Name) in the database $($dbMailProfile.Parent.Name) on $($dbMailProfile.Parent.Parent.Name)")) {
+                $output = [pscustomobject]@{
+                    ComputerName  = $dbMailProfile.Parent.Parent.ComputerName
+                    InstanceName  = $dbMailProfile.Parent.Parent.ServiceName
+                    SqlInstance   = $dbMailProfile.Parent.Parent.DomainInstanceName
+                    DbMailProfile = $dbMailProfile.Name
```
wsmelton, line 108:
```suggestion
                    Name = $dbMailProfile.Name
```

Matches output for the same property as the Get command.

tests/Remove-DbaDbMailProfile.Tests.ps1
```diff
@@ -0,0 +1,57 @@
...
+        }
+    }
+}
+
+Describe "$CommandName Integration Tests" -Tag "IntegrationTests" {
+
+    BeforeEach {
+
+        $server = Connect-DbaInstance -SqlInstance $script:instance2
+        $server2 = Connect-DbaInstance -SqlInstance $script:instance3
```
wsmelton, line 20: Testing against instance3 is sufficient as there no big difference between 2016 and 2017.

wsmelton (commented):
A few recommendations.

functions/Remove-DbaDbMailProfile.ps1
MikeyBronowski: It's because 
`The $PROFILE automatic variable stores the paths to the PowerShell profiles that are available in the current session.`


The New-* function uses Name .

Happy to change all of them if we are not concerned about $profile automatic variable.

functions/Remove-DbaDbMailProfile.ps1
MikeyBronowski: Will that be consistent with the proposed parameter change : Name -> Profile?

tests/Remove-DbaDbMailProfile.Tests.ps1
MikeyBronowski: The purpose of this test is to check if the command works against multiple instances, as there were commands in the past where supplying multiple instances did not work, i.e. command worked only on the last instance from the list.

functions/Remove-DbaDbMailProfile.ps1
MikeyBronowski: Done. Changed to match outputs of the Get / New commands.

functions/Remove-DbaDbMailProfile.ps1
wsmelton: It does not effect `Get-DbDbMailProfile`.

wsmelton: image

tests/Remove-DbaDbMailProfile.Tests.ps1
wsmelton: If you want to test against multiple instances then pass in `instance3` twice 😉

functions/Remove-DbaDbMailProfile.ps1
MikeyBronowski: The command runs fine. The ScriptAnalyzer flagged that automatic variable.
Also, the example above returns the Profile under `Name` property, not `Profile`, so in this case, it would be more consistent.

wsmelton: `$PROFILE` is a session level variable in PowerShell, that does not mean that variable will do the same thing with scoped to a function. Once it becomes a parameter of the function the session level variable is out of scope.

Example:

```powershell
function a ([string]$profile) {
    write-output "What is in $Profile"
}
```

Run that without providing any value and you will not get your PowerShell profile script, it will return nothing.

tests/Remove-DbaDbMailProfile.Tests.ps1
MikeyBronowski: How would I know that the command worked twice since it's removing the same object from the same instance? 
Once it's removed there is no way to say, it was removed in the first go or the last go. 

Example:
One of the command (don't remember which one, there is a PR for that in history) worked well only on a single instance. When supplied with instance A and instance B, it only worked on instance B as that was the last one in the loop. 

If I supply Instance A, and Instance A how do I tell if it ran on 
- both
- first only
- second only
With two different instances, I can easily tell that.

functions/Remove-DbaDbMailProfile.ps1
wsmelton: > The ScriptAnalyzer flagged that automatic variable.

We ignore that rule and have no dependency on it being checked in the project.

MikeyBronowski: Thank you. I can change back to `Profile` (same for Account function), although the New-* commands will need a change as well, as they use `Name`.

MikeyBronowski: Changed it to `Profile`. @wsmelton  for the New-* commands, do we want to add alias with `Profile`, `Account` respectively or add those as parameters and convert `Name` parameter to an alias?

tests/Remove-DbaDbMailProfile.Tests.ps1
MikeyBronowski: Removed the second server to save processing time.

functions/Remove-DbaDbMailProfile.ps1
wsmelton: I would vote to change it all to be consistent if we had it as Name bump that to alias. // @potatoqualitee

potatoqualitee: Name as an alias? Most of our stuff uses like Database, Job, etc. Name can be an alias (seems fun) -- I know sometimes I try to use Name as a param and it doesn't exist, so yes, I endorse. For Profile, i see how that's iffy but it works 😅

potatoqualitee:
Thanks, everyone!

Issue #7854, opened 2021-10-12
Get-DbaDbRestoreHistory - additional restore point info request


devondba:
### Summarize Functionality

Request to include the stop at point in the restore operation or the backup start point if no stop at passed. This will then provide enough info to ascertain the point in time the database has been restored to. 

Included:

backup_start_date as BackupStartDate
stop_at as StopAt
combination of the two to provide the LastRestorePoint - being either the stop_at point or the backup_start_date of the last restored file.

### Is there a command that is similiar or close to what you are looking for?

Yes

### Technical Details

Include extra items in the query:

 bs.last_lsn,
                    bs.checkpoint_lsn,
                    bs.database_backup_lsn,
                    bs.backup_start_date,
                    bs.backup_start_date AS BackupStartDate,
                    bs.backup_finish_date,
                    bs.backup_finish_date AS BackupFinishDate,
					rsh.stop_at AS StopAt,
                    CASE 
		                WHEN coalesce(rsh.stop_at, '9999-12-31') < bs.backup_start_date THEN stop_at
		                ELSE bs.backup_start_date
	                END AS LastRestorePoint
	                
I'm not sue why backup_finish_date is added twice then excluded from the list? I've added backup_start_date the same way for consistency. The item therefore needs to be added to the exclude list too:

 $results | Select-DefaultView -ExcludeProperty first_lsn, last_lsn, checkpoint_lsn, database_backup_lsn, backup_start_date, backup_finish_date

I'll start a branch to upload to.

andreasjordan:
@claude please have a look at this issue.

PR #7859, opened 2021-10-13
Get-DbaDbRestoreHistory

devondba:
Included BackupStartDate along with StopAt & a combination of the two to provide the LastRestorePoint - being either the stopat point or the start of the last restored backup

Type of Change
 Bug fix (non-breaking change, fixes # )
[x ] New feature (non-breaking change, adds functionality, fixes Get-DbaDbRestoreHistory - additional restore point info request #7854 )
 Breaking change (effects multiple commands or functionality, fixes # )
 Ran manual Pester test and has passed (.\tests\manual.pester.ps1)
 Adding code coverage to existing functionality
 Pester test is included
 If new file reference added for test, has is been added to github.com/sqlcollaborative/appveyor-lab ?
 Unit test is included
 Documentation
 Build system
Purpose
Adds additional information to the output of the function relating to the point in time of the restore.

Approach
Additional fields returned in SQL statement

Commands to test
Same as before just provide additional lines of output

Screenshots
Shows the new output and the LastRestorePoint being the start of the differential backup.

image

This shows the output of the SQL queries used in the script. Demonstrating the LastRestorePoint returned as the stop_at point:

image

The following shows the output with the last file being a transaction log created after the stop at point but covering the required time. The LastRestoredPoint is shown as the stop at point not the .trn file datetime (UTC, rather than BST)

image

Finally the log backup not being after the stop at time therefore the stop at time can't be used and the point in time will be that of the backup file.

image

### DIFF
--- a/functions/Get-DbaDbRestoreHistory.ps1
+++ b/functions/Get-DbaDbRestoreHistory.ps1
@@ -142,11 +142,11 @@ function Get-DbaDbRestoreHistory {
                     bs.backup_start_date AS BackupStartDate,
                     bs.backup_finish_date,
                     bs.backup_finish_date AS BackupFinishDate,
-					rsh.stop_at AS StopAt,
+                    rsh.stop_at AS StopAt,
                     CASE 
-		                WHEN coalesce(rsh.stop_at, '9999-12-31') < bs.backup_start_date THEN stop_at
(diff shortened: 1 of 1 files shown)


devondba:
Failed again on primarily the same thing, something .net related which is strange as I've only changed parts of the SQL Script.

PR #7866, opened 2021-10-16
Folder structure Attempt 1

wsmelton:
Getting Appveyor to run...

### DIFF
--- a/tests/Watch-DbatoolsUpdate.Tests.ps1
+++ b/tests/Watch-DbatoolsUpdate.Tests.ps1
@@ -4,11 +4,6 @@ Write-Host -Object "Running $PSCommandPath" -ForegroundColor Cyan
 
 Describe "$CommandName Unit Tests" -Tag 'UnitTests' {
     Context "Validate parameters" {
-        [array]$params = ([Management.Automation.CommandMetaData]$ExecutionContext.SessionState.InvokeCommand.GetCommand($CommandName, 'Function')).Parameters.Keys
-        [object[]]$knownParameters = ''
-
-        It "Should only contain our specific parameters" {
-            Compare-Object -ReferenceObject $knownParameters -DifferenceObject $params | Should -BeNullOrEmpty
-        }
+        <# no parameters to test #>
     }
 }


wsmelton:
- Majority of what was in the `\bin` path still there
- Things removed were scripts or files that had no reference anywhere (e.g., `bcp.exe` I could not find anywhere and most don't even know we have that in the module)
- I removed the 3rd party licenses as I could not find any reference why those were there. One referenced SSDT but we do not have any of those files in the module that I could tell. SMO would not require those license files to be present since it is already a published library on NuGet Gallery.
- Anything that was under `internal` is now under `src\private` directory
- The `functions\` is now `src\public`
- Dropped the `xml\` path as the `ps1xml` files are generally kept with the psm1 and psd1 files.


Right now the `bin\projects` folders is in the same path until I can figure out how to adjust the project configuration. It should reside under `src\dbatools` as that is the name of the library file itself. I can write an InvokeBuild script that will handle building and running the unit tests in dotnet.

wsmelton:
All the paths with moving the module files to a nested folder are updated, the module works and a few individual test I can run locally work with no issue. The best I can find the module is not importing fully or the psm1 file at least so the test can run and reference the private functions properly.

wsmelton:
These runs for the last group all show the same failures, most of them look like legit issues with the way the test is written:

- https://ci.appveyor.com/project/sqlcollaborative/dbatools/builds/41174963/job/5qb0aptmt2i6pe51
- https://ci.appveyor.com/project/sqlcollaborative/dbatools/builds/41175090/job/vhw8b7rreamayu35
- https://ci.appveyor.com/project/sqlcollaborative/dbatools/builds/41172380/job/miv4sd9ls7tn32mq

But runs with the original structure all let them pass. I can't really tell if this is out of pure luck or it is something with AppVeyor's load process with our setups.

I randomly get this failure in some groups that I can't figure out why:

https://ci.appveyor.com/project/sqlcollaborative/dbatools/builds/41176411/job/ayyqynx94xcoo0t1

image

This SDK has to exists for our library to build properly because we target .NET 4.6.2 and this is the SDK from [choco](https://community.chocolatey.org/packages/windows-sdk-10.0).

PR #6432, opened 2020-03-22
Fixes for Update-DbaInstance

nvarscar:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #6412 fixes #6220 )
 - [ ] New feature (non-breaking change, adds functionality, fixes # )
 - [ ] Breaking change (effects multiple commands or functionality, fixes # )
 - [x] Ran manual Pester test and has passed (`.\tests\manual.pester.ps1)
 - [ ] Adding code coverage to existing functionality
 - [ ] Pester test is included
 - [ ] If new file reference added for test, has is been added to github.com/sqlcollaborative/appveyor-lab ?
 - [ ] Nunit test is included
 - [ ] Documentation
 - [ ] Build system
 

### Purpose
Fix a few annoying bugs:
- SSRS is undetected
- Pending file renames prevent the instance upgrade & install

### Approach
1. File rename checks are now optional and disabled by default.
2. Array automatically expands, thinking each array member is a unique argument. Wrapping it into an object resolves the issue.

### Screenshots
image

PR #7925, opened 2021-11-05
Re-enabling pending rename check by default

### LINKED ISSUE #7301: [Bug] Install-DbaInstance / Invoke-DbaAdvancedInstall - No test for pending reboot because of file rename
I tried to install an instance and the install failed with: "A computer restart is required. You must restart this computer before installing SQL Server." which was not detected by Test-PendingReboot because it was a PendingFileRenameOperations, but here the `-PendingRename` is missing:

https://github.com/sqlcollaborative/dbatools/blob/f6dd5fb1fd75073ff59c0fe7fd7afd7b4b0b31e2/functions/Invoke-DbaAdvancedInstall.ps1#L180

Maybe we should add this parameter here as well:

https://github.com/sqlcollaborative/dbatools/blob/f6dd5fb1fd75073ff59c0fe7fd7afd7b4b0b31e2/functions/Invoke-DbaAdvancedInstall.ps1#L180

@nvarscar : Please have a look. Again: I can open the pull request if you like.

nvarscar:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #7301 )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
Revert the change introduced in #6432

### Approach
Pending file renames should be checked every time, but if users want to opt out, they can do it now using a switch `-NoPendingRenameCheck ` or by setting a config value `OS.PendingRename`

### Learning
If a user wants a customized behavior, introduce it as a switch/config. Don't change it for everyone.

### DIFF
--- a/functions/Get-DbaComputerSystem.ps1
+++ b/functions/Get-DbaComputerSystem.ps1
@@ -132,7 +132,7 @@ function Get-DbaComputerSystem {
                 $pendingReboot = $null
                 try {
                     Write-Message -Level Verbose -Message "Getting information about pending reboots."
-                    $pendingReboot = Test-PendingReboot -ComputerName $computerResolved -Credential $Credential -PendingRename
+                    $pendingReboot = Test-PendingReboot -ComputerName $computerResolved -Credential $Credential
                 } catch {
                     Write-Message -Level Verbose -Message "Not able to get information about pending reboots."
                 }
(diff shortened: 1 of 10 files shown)


andreasjordan (approved):
Thanks for the change, looks good to me.

potatoqualitee:
Thank you both 😊

jarias-korewireless (commented):
Function Install-DbaInstance.ps1 is not sending the parameter **-NoPendingRename:$NoPendingRenameCheck** when calling **Test-PendingReboot** on the _"Test if the restart is needed"_ section.

potatoqualitee:
@jarias-korewireless  thank you! @nvarscar can you check it out?

PR #7962, opened 2021-11-19
New-DbaAvailabilityGroup - add a test before attempting to add database

potatoqualitee:
Sometimes this awesome command would be too darn fast and the ag wasn't ready to accept a database. This waits until the state is connected before attempting to add.

### DIFF
--- a/functions/New-DbaAvailabilityGroup.ps1
+++ b/functions/New-DbaAvailabilityGroup.ps1
@@ -640,6 +640,21 @@ function New-DbaAvailabilityGroup {
             }
         }
 
+        # Wait for the availability group to be ready
+        Write-ProgressHelper -StepNumber ($stepCounter++) -Message "Waiting for replicas to be connected and ready"
+        do {
+            Start-Sleep -Milliseconds 500
+            $wait++
+            $ready = $true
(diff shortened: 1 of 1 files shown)


andreasjordan:
Good idea. But we should not only wait if we want to add databases within the command, but also wait in any case for the state "connected" before running the final `Get-DbaAvailabilityGroup`. Because then all other commands in a script that runs `New-DbaAvailabilityGroup` can rely on a ready-to-go AG. So maybe move the new block up to line 642?

potatoqualitee:
good idea! moved it

andreasjordan:
Maybe add a warning message, if $ready is still $false? I would not fail, but let the user know that the AG is maybe not yet ready. Then, if adding the databases failes, the user has a hint why adding databases failed.

andreasjordan:
"One last thing": `-not $ready` is sufficent, `-or $wait -gt 40` not needed.

potatoqualitee:
I know, im just covering alll the bases 😅

andreasjordan (approved):

potatoqualitee:
sweet, thanks 👍🏼

PR #7977, opened 2021-11-23
Copy-DbaAgentAlert - change needlessly alarming warning to verbose

potatoqualitee:
this fails too often and is kinda useless yet scary sounding "A severe error occurred on the current command.  The results, if any, should be discarded." , just log as verbose


potatoqualitee:
found some other errors, likelyl related to the new containers, and i'm fixing them in this PR

potatoqualitee:
doin it 💯

PR #7988, opened 2021-11-27
Various Linux Fixes

potatoqualitee:
The failure of Invoke-DbaDbMirroring highlighted a few issues with the way we handle paths.


potatoqualitee:
I still have some fixes to do on Backup-DbaDatabase that relate to path but need a dedicated branch for those

potatoqualitee:
Going to merge bc I need to do a few other things

PR #7995, opened 2021-11-30
Removes the comment that was overlooked

MikeyBronowski:
The comment should be removed as that was used for the development phase.


andreasjordan (approved):

potatoqualitee:
Thank you both!

PR #8011, opened 2021-12-04
Install-DbaInstance: Added parameters for collations

### LINKED ISSUE #8007: Install-DbaInstance - Default collations
### Summarize Functionality

The command `Install-DbaInstance` uses hard coded default collations for both AS and SQL:
ASCOLLATION = "Latin1_General_CI_AS"
SQLCOLLATION = "SQL_Latin1_General_CP1_CI_AS"

https://github.com/sqlcollaborative/dbatools/blob/d07ba1e1d8b5b41ef58587a5481a7cddbb31d7fa/functions/Install-DbaInstance.ps1#L570-L595

As I don't use Analysis Services, I will now only talk about the Database Engine.

There is an example how to change that with `-Configuration @{ SQLCOLLATION = 'Latin1_General_CI_AS' }`, but as I always have to use this (I'm based in Europe) I would like to talk about changing this.

First of all: Why do we set this anyway?
Here is the default, if the parameter is not set, based on the Windows locale:
https://docs.microsoft.com/en-us/sql/relational-databases/collations/collation-and-unicode-support?view=sql-server-ver15#Server-level-collations

Let's talk about where the Latin1 collations are used:
SQL_Latin1_General_CP1_CI_AS is used for English (United States) - that was hard coded
Latin1_General_CI_AS is used for English (* except United States), German (*) and some other european countries

So we have a hard coded value that is the default for the United States, in every other location, the collation has to be configured.

I would like to change this and I see four choices:
* We remove ASCOLLATION and SQLCOLLATION from the code. Now the default is used like in the GUI-setup.exe where most users just don't change the default. This will be no change for the United States as of the SQLCOLLATION, but someone has to test for ASCOLLATION. This will be no change for those in Europe, that use `-Configuration @{ SQLCOLLATION = 'Latin1_General_CI_AS' }` - they can just stop using it. But it will be a change for those in Europe that have not configured the collation and now are not getting the US-default anymore, but their local default.
* We add new parameters `-AsCollation` and `-SqlCollation` without defaults. That is like number 1, but it's easier for the user to set the collation. We also have a parameter where we can put some documentation about where to find the defaults and the other possible names.
* We add new parameters `-AsCollation` and `-SqlCollation` with defaults ASCOLLATION = "Latin1_General_CI_AS" and SQLCOLLATION = "SQL_Latin1_General_CP1_CI_AS". That would not change the behavior, but would make it easier for us in Europe. But: We would have defaults that are different from the "official" defaults. I think we wanted to have as few as possible of those.
* Like the third one, but get the defaults from a config value. But then we need to deciede on whether we have a default value for this config value or not.

I would vote for the second one: Adding parameters without default values.

No

### Technical Details

-

andreasjordan:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality, fixes #8007 )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

See linked issue for details.

### DIFF
--- a/functions/Install-DbaInstance.ps1
+++ b/functions/Install-DbaInstance.ps1
@@ -128,6 +128,15 @@ function Install-DbaInstance {
         Product ID, or simply, serial number of your SQL Server installation, which will determine which version to install.
         If the PID is already built into the installation media, can be ignored.
 
+    .PARAMETER AsCollation
+        Collation for the Analysis Service.
+        Default value: Latin1_General_CI_AS
+
+    .PARAMETER SqlCollation
+        Collation for the Database Engine.
+        The default depends on the Windows locale:
+        https://docs.microsoft.com/en-us/sql/relational-databases/collations/collation-and-unicode-support#Server-level-collations
+
     .PARAMETER EngineCredential
         Service account of the SQL Server Database Engine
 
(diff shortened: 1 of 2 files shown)


potatoqualitee:
Looks great to me! @nvarscar can you give it a look? 🙇🏼

nvarscar (approved):
Looks good to me

potatoqualitee:
Thank you both 😊

PR #8012, opened 2021-12-04
Backup-DbaDbCertificate should warn on missing certificate

### LINKED ISSUE #7918: Backup-DbaDbCertificate should warn on missing certificate
### Summarize Functionality

When backing up a database certificate by name, if the named certificate does not exist, a warning should be thrown telling the user.

Yes

### Technical Details

Test the results of `Get-DbaDbCertificate` against the passed-in name(s) and if things don't match up, write a warning out informing the user that one or more certificates won't/can't be backed up because they don't exist.

lancasteradam:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality, fixes #7918 )
 - [x] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose

Minor change to validate if a certificate was not found.

### Approach

Check each db cert and return a warning message if it wasn't found.

### Commands to test

See updated Pester tests.

### DIFF
--- a/functions/Backup-DbaDbCertificate.ps1
+++ b/functions/Backup-DbaDbCertificate.ps1
@@ -241,7 +241,7 @@ function Backup-DbaDbCertificate {
         }
 
         if ($Certificate) {
-            $missingCerts = $Certificate | ? { $InputObject.Name -notcontains $_ }
+            $missingCerts = $Certificate | Where-Object { $InputObject.Name -notcontains $_ }
 
             if ($missingCerts) {
                 Write-Message -Level Warning -Message "Database certificate(s) $missingCerts not found in Database(s)=$Database on Instance(s)=$SqlInstance"
--- a/functions/Backup-DbaDbCertificate.ps1
+++ b/functions/Backup-DbaDbCertificate.ps1
@@ -237,13 +237,15 @@ function Backup-DbaDbCertificate {
         if (Test-FunctionInterrupt) { return }
 
         if ($SqlInstance) {
-            foreach ($dbCert in $Certificate) {
-                $databaseCertToBackup = Get-DbaDbCertificate -SqlInstance $SqlInstance -SqlCredential $SqlCredential -Database $Database -ExcludeDatabase $ExcludeDatabase -Certificate $dbCert
+            foreach ($si in $SqlInstance) {
+                foreach ($dbCert in $Certificate) {
+                    $databaseCertToBackup = Get-DbaDbCertificate -SqlInstance $si -SqlCredential $SqlCredential -Database $Database -ExcludeDatabase $ExcludeDatabase -Certificate $dbCert
 
-                if ($databaseCertToBackup) {
-                    $InputObject += $databaseCertToBackup
-                } else {
-                    Stop-Function -Message "Database certificate $dbCert was not found in $Database on $SqlInstance" -Continue
+                    if ($databaseCertToBackup) {
+                        $InputObject += $databaseCertToBackup
+                    } else {
+                        Stop-Function -Message "Database certificate $dbCert was not found in $Database on $si" -Continue
+                    }
                 }
             }
         }
--- a/functions/Backup-DbaDbCertificate.ps1
+++ b/functions/Backup-DbaDbCertificate.ps1
@@ -237,7 +237,15 @@ function Backup-DbaDbCertificate {
         if (Test-FunctionInterrupt) { return }
 
         if ($SqlInstance) {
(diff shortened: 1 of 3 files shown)


lancasteradam:
This change seems to have impacted the tests for Restore-DbaDbCertificate. I'll take a look.

andreasjordan:
`$SqlInstance` is an array, so you should also add a loop like `foreach ($instance in $SqlInstance) { ... }`.

lancasteradam:
ok, thanks, I'll take a look.

andreasjordan:
It's just a name, but in every other command, the variable that loops through `$SqlInstance` is named `$instance`...

functions/Backup-DbaDbCertificate.ps1
```diff
@@ -237,7 +237,17 @@ function Backup-DbaDbCertificate {
         if (Test-FunctionInterrupt) { return }
 
         if ($SqlInstance) {
-            $InputObject += Get-DbaDbCertificate -SqlInstance $SqlInstance -SqlCredential $SqlCredential -Database $Database -ExcludeDatabase $ExcludeDatabase -Certificate $Certificate
+            foreach ($si in $SqlInstance) {
```
potatoqualitee, line 240: Thamk you, Adam! This is generally `foreach ($instance in $SqlInstance`

potatoqualitee (changes requested):
the rest looks great, thank you

lancasteradam:
Thank you both!

wsmelton:
Can't we simply add a check if the `$InputObject` contains the cert after the call on line 240 (original)? If it is not found then spit out the warning message and continue. 

Performing a `foreach` on each certificate and repeatedly calling `Get-DbaDbCertificate` is inefficient to me.

Add a check like: `if (-not $InputObject) { Stop-Function -Message "Cert not found blah blah blah" -Continue }`

andreasjordan:
We have a bigger problem: $Certificate can be empty if the user wants to backup all certificates. That works with the old code, but not with the new code.

potatoqualitee:
Yes! That will need to be fixed prior to merge. This should be able to backup all certs.

lancasteradam:
Thank you team for the help, it is much appreciated! I've changed the code based on the suggestions and also expanded the test coverage for the scenarios that were mentioned. Let me know if more edits are needed.

functions/Backup-DbaDbCertificate.ps1
```diff
@@ -240,6 +240,14 @@ function Backup-DbaDbCertificate {
             $InputObject += Get-DbaDbCertificate -SqlInstance $SqlInstance -SqlCredential $SqlCredential -Database $Database -ExcludeDatabase $ExcludeDatabase -Certificate $Certificate
         }
 
+        if ($Certificate) {
+            $missingCerts = $Certificate | ? { $InputObject.Name -notcontains $_ }
```
potatoqualitee, line 244: So close! Please avoid using aliases. Also, you'll be excited to know that you can do `Where-Object Name -notcontains $_`, without braces, when comparing just one thing. It's even faster, according to bruce payette.

jgiercyk (approved):

functions/Backup-DbaDbCertificate.ps1
lancasteradam: Thanks! I switched it to Where-Object and kept the braces since a caller could pass in multiple values for -Certificate.

functions/Backup-DbaDbCertificate.ps1
```diff
@@ -240,6 +240,14 @@ function Backup-DbaDbCertificate {
             $InputObject += Get-DbaDbCertificate -SqlInstance $SqlInstance -SqlCredential $SqlCredential -Database $Database -ExcludeDatabase $ExcludeDatabase -Certificate $Certificate
         }
 
+        if ($Certificate) {
+            $missingCerts = $Certificate | Where-Object { $InputObject.Name -notcontains $_ }
+
+            if ($missingCerts) {
+                Write-Message -Level Warning -Message "Database certificate(s) $missingCerts not found in Database(s)=$Database on Instance(s)=$SqlInstance"
```
wsmelton, line 247: Is this message actually readable if multiple certs are found missing (when user passes in more than one)?

lancasteradam: Yes, there is a pester test that checks when multiple certs are not found. The warning message has a space between the cert names and looks like: 

"WARNING: [18:49:58][Backup-DbaDbCertificate] Database certificate(s) dbatoolscli_invalidCertName dbatoolscli_invalidCertName2 not found in Database(s)=DatabaseName on
Instance(s)=MachineName\InstanceName"

potatoqualitee:
LGTM! thank you all 🙇🏼

PR #8042, opened 2021-12-28
Get-DbaDbView - Use ClearAndInitialize to improve performance

andreasjordan:
## Type of Change

 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

If we run code like `$views = $views | Where-Object { -not $_.IsSystemObject }`, the SMO works on a per view bases, so one view after the other is tested. As IsSystemObject is not one of the default properties, it is empty by default and has to be fetched from the database for every individual view. With say 1000 views in the database, that are 1000 fetches. If we use ClearAndInitialize, then the named properties are all fetched together in one single fetch. After that fetch, the database is not touched anymore and the command runs completely with information inside of the SMO.

About the downside: The Clear in ClearAndInitialize is a clear. So all properties fetched beforehand outside of the command are removed and would have to be fetched again.
So this code would be fast before the cange and slow after the change, because the property Owner would be fetched for every view one at a time when Format-Table runs:
``` 
$db = Get-DbaDatabase -SqlInstance sql01 -Database ViewTest
$db.Views.ClearAndInitialize('', [string[]]('Name', 'Schema', 'IsSystemObject', 'CreateDate', 'DateLastModified', 'Owner'))
$db | Get-DbaDbView -ExcludeSystemView | Format-Table -Property Name, Schema, Owner
```

So the general rule would be: This change helps if the command uses a brand new SMO like when -SqlInstance is used with a string. If someone uses ClearAndInitialize already outside of the command, then it would probably be better to not using the command but:
```
$db = Get-DbaDatabase -SqlInstance sql01 -Database ViewTest
$db.Views.ClearAndInitialize('', [string[]]('Name', 'Schema', 'IsSystemObject', 'CreateDate', 'DateLastModified', 'Owner'))
$db.Views | Where-Object { -not $_.IsSystemObject } | Format-Table -Property Name, Schema, Owner
```

### DIFF
--- a/functions/Get-DbaDbView.ps1
+++ b/functions/Get-DbaDbView.ps1
@@ -131,6 +131,11 @@ function Get-DbaDbView {
 
         foreach ($db in $InputObject) {
             Write-Message -Level Verbose -Message "processing $db"
+
+            # Let the SMO read all properties referenced in this command for all views in the database in one query.
+            # Downside: If some other properties were already read outside of this command in the used SMO, they are cleared.
+            $db.Views.ClearAndInitialize('', [string[]]('Name', 'Schema', 'IsSystemObject', 'CreateDate', 'DateLastModified'))
+
             if ($fqtns) {
                 $views = @()
                 foreach ($fqtn in $fqtns) {


andreasjordan:
Feedback is welcome. Before merging this, we should be really sure to go this path. Because if so, we should also change a lot of other commands the same way. Or we only change some of the commands and wait for user feedback. Or we hide this functionallity behind a config parameter. Or...

wsmelton:
Write up from SMO team for those interested:

https://github.com/microsoft/sqlmanagementobjects/wiki/Refresh-vs-ClearAndInitialize

potatoqualitee:
let's do this and put it on the agenda to revisit other commands that could benefit as well 👍🏼

PR #8072, opened 2022-01-13
Copy-DbaCredential - Enable EKM tests

potatoqualitee:
So cool! Thanks for that functionality @lancasteradam. Let me know if this is what you had in mind.


potatoqualitee:
Added it to the instance level instead because I need it in some tests too

potatoqualitee:
Going to merge because I need it 😊

lancasteradam:
awesome teamwork, thanks!

PR #8084, opened 2022-01-16
Backup-DbaDbCertificate - Remove default date suffix

potatoqualitee:
## Type of Change

 - [x] Breaking change (effects multiple commands or functionality, fixes # )

### Purpose
The suffix was added so that there were no clobber issues. But it also renames the certificate in the database when you restore it with our default method and you can't do a regex for it, really (or maybe we can and i dont know how)

Maybe changing the restore command with a regex detection is a better idea.

### Approach
Make it the regular name, and if that exists, then use the backup name.


potatoqualitee:
ugh i made this on the wrong branch. closing.

PR #8085, opened 2022-01-16
Backup-DbaDbCertificate - Remove default date suffix

potatoqualitee:
## Type of Change

 - [x] Breaking change

### Purpose
The suffix was added so that there were no clobber issues. But it also renames the certificate in the database when you restore it with our default method. So I fixed both commands. Backup now defaults to no date unless the file exists then it appends a date if a suffix has not been specified.

Then, in Restore, it looks for that string and removes it so that the cert looks prettier and maintains the same name. This is a mild breaking change that impacts people who are looking for that string in their backup file.

Also, this site for regex is awesome: https://www.regextester.com/112363

### Approach
Make it the regular name, and if that exists, then add the date to the name.

### DIFF
--- a/functions/Backup-DbaDbCertificate.ps1
+++ b/functions/Backup-DbaDbCertificate.ps1
@@ -153,18 +153,16 @@ function Backup-DbaDbCertificate {
             }
 
             $actualPath = "$actualPath".TrimEnd('\').TrimEnd('/')
-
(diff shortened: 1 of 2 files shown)


ClaudioESSilva (approved):

potatoqualitee:
Thanks bud!

PR #8087, opened 2022-01-17
Add-DbaAgDatabase - Add missing SqlCredential when calling Test-DbaAvailabilityGroup (part 2)

### LINKED ISSUE #7481: [Bug]Add-DbaAgDatabase does not honor SqlCredential

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #7481  )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

See also #7482.

### DIFF
--- a/functions/Add-DbaAgDatabase.ps1
+++ b/functions/Add-DbaAgDatabase.ps1
@@ -211,6 +211,7 @@ function Add-DbaAgDatabase {
                 Write-Progress @progress
                 $testSplat = @{
                     SqlInstance            = $db.Parent
+                    SqlCredential          = $SqlCredential
                     Secondary              = $Secondary
                     SecondarySqlCredential = $SecondarySqlCredential
                     AvailabilityGroup      = $AvailabilityGroup


functions/Add-DbaAgDatabase.ps1
```diff
@@ -211,6 +211,7 @@ function Add-DbaAgDatabase {
                 Write-Progress @progress
                 $testSplat = @{
                     SqlInstance            = $db.Parent
+                    SqlCredential          = $SqlCredential
```
potatoqualitee, line 214: Thanks Andreas! $db.Parent is a connected object and shouldn't need SqlCredential. Does this really fix the issue?

potatoqualitee: Perhaps a .Refresh() is needed somewhere instead?

andreasjordan: Thinking again, you are probably right. Could be the reason why I have not added that in the first round....

potatoqualitee: Sometimes things are weird. If you want, you can use $db.Parent.Name and $SqlCredential to force a new connect or a refresh may help.

PR #8088, opened 2022-01-17
Copy-DbaDbCertificate + Bug Fixes

potatoqualitee:
`Copy-DbaDbCertificate` then added it to `Start-DbaMigration` then fixed some bugs while I was in `Backup-DbaDatabase`

The issue with the pipes is that a lot of stuff occured in the BEGIN which piped objects do not see. This will likely slow down Backup-DbaDatabase a smidge but is necessary. 

`Get-DbaFile` is awesome and I forgot about it but I needed to address some things to make it work on case sensitive servers.

Oh, i also added progress bars to `Start-DbaMigration`

I also looked at the way that Microsoft handles backup compression for encrypted databases and made us support more TDE

I have one more thing that i want to fix with the backup path

### DIFF
--- a/functions/Copy-DbaDbCertificate.ps1
+++ b/functions/Copy-DbaDbCertificate.ps1
@@ -142,8 +142,6 @@ function Copy-DbaDbCertificate {
         }
     }
     process {
-        # FOR START-DBAMIGRATION, IT NEEDS TO JUST BE COPY-DBADBMASTER
-        # NEED TO DELETE ANY EXPORTED KEY
         if (Test-FunctionInterrupt) { return }
         foreach ($destinstance in $Destination) {
             try {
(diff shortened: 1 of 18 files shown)


functions/Copy-DbaDbCertificate.ps1
```diff
@@ -0,0 +1,331 @@
...
+            $parms = @{
+                SqlInstance     = $Source
+                SqlCredential   = $SourceSqlCredential
+                Database        = $Database
+                ExcludeDatabase = $ExcludeDatabase
+                Certificate     = $Certificate
+                EnableException = $true
+            }
+            # Get presumably user certs, no way to tell if its a system object
+            $sourcecertificates = Get-DbaDbCertificate @parms | Where-Object Name -notlike "#*" | Where-Object Name -notin $ExcludeCertificate
```
potatoqualitee, line 124:
```suggestion
            $sourcecertificates = Get-DbaDbCertificate @parms | Where-Object { $PSItem.Name -notlike "#*" -and $PSItem.Name -notin $ExcludeCertificate }
```

potatoqualitee:
Claudio gave private review and all tests pass, merging as i need it!

PR #8089, opened 2022-01-17
Install-DbaInstance / Update-DbaInstance: Use CredSSP when restarting computer

andreasjordan:
## Type of Change

 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

Together with `Set-DbatoolsConfig -FullName 'commands.initialize-credssp.bypass' -Value $true` to suppress the configuration of CredSSP, with this change the installation of an instance can be started from an unelevated powershell session. 

I also aligned the error messages and added the notes if restart failed.

### DIFF
--- a/functions/Invoke-DbaAdvancedInstall.ps1
+++ b/functions/Invoke-DbaAdvancedInstall.ps1
@@ -182,6 +182,7 @@ function Invoke-DbaAdvancedInstall {
     }
     if ($Credential) {
         $restartParams.Credential = $Credential
+        $restartParams.WsmanAuthentication = 'CredSSP'
     }
     $activity = "Installing SQL Server ($Version) components on $ComputerName"
     try {
@@ -199,7 +200,7 @@ function Invoke-DbaAdvancedInstall {
             $null = Restart-Computer @restartParams
             $output.Restarted = $true
         } catch {
-            Stop-Function -Message "Failed to restart computer" -ErrorRecord $_
+            Stop-Function -Message "Failed to restart computer $($ComputerName)" -ErrorRecord $_
         }
     }
     # save config if needed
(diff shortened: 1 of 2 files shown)


functions/Invoke-DbaAdvancedInstall.ps1
```diff
@@ -182,6 +182,7 @@ function Invoke-DbaAdvancedInstall {
     }
     if ($Credential) {
         $restartParams.Credential = $Credential
+        $restartParams.WsmanAuthentication = 'CredSSP'
```
nvarscar, line 185: This would in all cases require CredSSP authentication. What if end user decides to manually specify Kerberos in -Authentication? Would it make sense to switch to using `$Authentication` parameter here?

andreasjordan: Good point, changed that.

nvarscar (approved):

potatoqualitee:
Thank you both! 🙏🏼

andreasjordan:
Thanks for merging.

What about changing `Initialize-CredSSP` to improve the usability even further? We have to bypass `Initialize-CredSSP` because it is running `Get-WSManCredSSP` to get the information about the local configuration. This command only runs on an elevated process (failes with "Get-WSManCredSSP : Access is denied. You need to run this cmdlet from an elevated process." if not).
We could test if it is an elevated process and if not, just test if a CredSSP connection can be established to the target with `Invoke-Command2 -ComputerName $ComputerName -Credential $Credential -Authentication Credssp -ScriptBlock { } -Raw` as it is later run in line 488 of `Install-DbaInstance`. If this fails, we can have the error message say: "Please configure CredSSP or run this command from an elevated process."

This way a user can run Install-DbaInstance (and Update-DbaInstance) from an unelevated process, if CredSSP was already configured correctly - without the need for changing the configuration.

I like the fact that Install-DbaInstance can configure CredSSP. But in a production environment, I would rather like to seperate security configuration (which is CredSSP in my eyes) and SQL Server installation. I prefer to do the first with group policies and together with the responsible admins and then run the later from an unelevated process. 

If it's not only me that would like this change, I can code that and open a pull request. Just wanted to have some feedback first.

potatoqualitee:
That would be @nvarscar 's call because i dont know enough about any of it. btw, `Get-WSManCredSSP` works in windows 7 / psv3! woo

nvarscar:
I agree with what @andreasjordan is saying. Configuring CredSSP is nice and all, but when security is involved, some things should probably be more explicit - and configurable.

andreasjordan:
I opended #8107 to continue the work on this topic.

PR #8107, opened 2022-01-22
Initialize-CredSSP - Don't try to configure CredSSP if not possible / Install-DbaInstance+Update-DbaInstance - align documentation and parameters

### LINKED ISSUE #8117: Invoke-CommandWithFallback - Fallback not working

Yes

```
DEBUG: [16:16:54][Invoke-CommandWithFallback] Initial connection to SQL01.ordix.local through Credssp protocol unsuccessful, falling back to PSSession configurations | Connecting to remote server SQL01.ordix.local f
ailed with the following error message : The WinRM client cannot process the request. CredSSP authentication is currently disabled in the client configuration. Change the client configuration and try the request aga
in. CredSSP authentication must also be enabled in the server configuration. Also, Group Policy must be edited to allow credential delegation to the target computer. Use gpedit.msc and look at the following policy: 
Computer Configuration -> Administrative Templates -> System -> Credentials Delegation -> Allow Delegating Fresh Credentials.  Verify that it is enabled and configured with an SPN appropriate for the target computer
. For example, for a target computer name "myserver.domain.com", the SPN can be one of the following: WSMAN/myserver.domain.com or WSMAN/*.domain.com For more information, see the about_Remote_Troubleshooting Help t
opic.
DEBUG: [16:16:58][Register-RemoteSessionConfiguration] Registering new session configuration dbatoolsInvokeCommandWithFailback on SQL01.ordix.local
DEBUG: [16:17:01][Invoke-Command2] Found an existing session dbatools_Default_ce2e9be6-417e-4e93-9c41-43710fbbb923_SQL01.ordix.local, reusing it
DEBUG: [16:17:10][Register-RemoteSessionConfiguration] Configuration attempt returned the following status: Updated
DEBUG: [16:17:14][Register-RemoteSessionConfiguration] Restarting WinRm service on SQL01.ordix.local
DEBUG: [16:17:17][Invoke-Command2] Found an existing session dbatools_Default_ce2e9be6-417e-4e93-9c41-43710fbbb923_SQL01.ordix.local, reusing it
DEBUG: [16:17:21][Restart-WinRMService] Expected exception - the pipe was disconnected | Processing data for a remote command failed with the following error message: The I/O operation has been aborted because of ei
ther a thread exit or an application request. For more information, see the about_Remote_Troubleshooting Help topic.
DEBUG: [16:17:32][Restart-WinRMService] Removing existing local sessions to SQL01.ordix.local - they are no longer valid
DEBUG: [16:17:36][Restart-WinRMService] Removing 0 sessions from SQL01.ordix.local in runspace ce2e9be6-417e-4e93-9c41-43710fbbb923
WARNING: [16:17:37][Update-DbaInstance] Failed to enumerate files in -Path | Cannot bind argument to parameter 'Session' because it is null.
```

### Steps to Reproduce

Update-DbaInstance with -Credential but in an unelevated PowerShell and CredSSP not configured.

Yes

### Other details or mentions

I want to test the fallback but noticed that it always failed.

I think the reason is that we restart WinRM but don't close all the sessions to that host. So Invoke-Command2 tried to reuse a session that was broken.

Don't have time to find a good solution - can someone help?

### PowerShell Host Version

-

-

### .NET Framework Version

-

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #8117 )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
Be able to run Install-DbaInstance and Update-DbaInstance in an unelevated PowerShell session, when CredSSP is already configured. No need to bypass Initialize-CredSSP with commands.initialize-credssp.bypass in this case.

### Approach
Test for elevated PowerShell session inside of Initialize-CredSSP

### DIFF
--- a/functions/Install-DbaInstance.ps1
+++ b/functions/Install-DbaInstance.ps1
@@ -24,9 +24,10 @@ function Install-DbaInstance {
         Note that the dowloaded installation media must be extracted and available to the server where the installation runs.
         NOTE: If no ProductID (PID) is found in the configuration files/parameters, Evaluation version is going to be installed.
 
-        When using CredSSP authentication, this function will configure CredSSP authentication for PowerShell Remoting sessions.
+        When using CredSSP authentication, this function will try to configure CredSSP authentication for PowerShell Remoting sessions.
         If this is not desired (e.g.: CredSSP authentication is managed externally, or is already configured appropriately,)
         it can be disabled by setting the dbatools configuration option 'commands.initialize-credssp.bypass' value to $true.
+        To be able to configure CredSSP, the command needs to be run in an elevated PowerShell session.
 
     .PARAMETER SqlInstance
         The target computer and, optionally, a new instance name and a port number.
@@ -42,7 +43,11 @@ function Install-DbaInstance {
         Securely provide the password for the sa account when using mixed mode authentication.
 
     .PARAMETER Credential
-        Used when executing installs against remote servers
+        Windows Credential with permission to log on to the remote server.
+        Must be specified for any remote connection if SQL Server installation media is located on a network folder.
+
+        Authentication will default to CredSSP if -Credential is used.
+        For CredSSP see also additional information in DESCRIPTION.
 
     .PARAMETER ConfigurationFile
         The path to the custom Configuration.ini file.
@@ -57,12 +62,14 @@ function Install-DbaInstance {
 
     .PARAMETER Authentication
         Chooses an authentication protocol for remote connections.
-        Allowed values: 'Default', 'Basic', 'Negotiate', 'NegotiateWithImplicitCredential', 'Credssp', 'Digest', 'Kerberos'
-        If the protocol fails to establish a connection
+        Allowed values: 'Default', 'Basic', 'Negotiate', 'NegotiateWithImplicitCredential', 'Credssp', 'Digest', 'Kerberos'.
+        If the protocol fails to establish a connection (HELP: What should be the end of this sentence?)
 
         Defaults:
         * CredSSP when -Credential is specified - due to the fact that repository Path is usually a network share and credentials need to be passed to the remote host to avoid the double-hop issue.
         * Default when -Credential is not specified. Will likely fail if a network path is specified.
+
+        For CredSSP see also additional information in DESCRIPTION.
 
     .PARAMETER Version
         SQL Server version you wish to install.
@@ -256,7 +263,7 @@ function Install-DbaInstance {
         [PSCredential]$SaCredential,
         [PSCredential]$Credential,
         [ValidateSet('Default', 'Basic', 'Negotiate', 'NegotiateWithImplicitCredential', 'Credssp', 'Digest', 'Kerberos')]
-        [string]$Authentication = 'Credssp',
+        [string]$Authentication = @('Credssp', 'Default')[$null -eq $Credential],
         [parameter(ValueFromPipeline)]
         [Alias("FilePath")]
         [object]$ConfigurationFile,
--- a/functions/Install-DbaInstance.ps1
+++ b/functions/Install-DbaInstance.ps1
@@ -63,7 +63,8 @@ function Install-DbaInstance {
     .PARAMETER Authentication
         Chooses an authentication protocol for remote connections.
         Allowed values: 'Default', 'Basic', 'Negotiate', 'NegotiateWithImplicitCredential', 'Credssp', 'Digest', 'Kerberos'.
-        If the protocol fails to establish a connection (HELP: What should be the end of this sentence?)
+        If the protocol fails to establish a connection and explicit -Credentials were used, a failback authentication method would be attempted that configures PSSessionConfiguration
+        on the remote machine. This method, however, is considered insecure and would, therefore, prompt an additional confirmation when used.
 
         Defaults:
         * CredSSP when -Credential is specified - due to the fact that repository Path is usually a network share and credentials need to be passed to the remote host to avoid the double-hop issue.
--- a/functions/Install-DbaInstance.ps1
+++ b/functions/Install-DbaInstance.ps1
@@ -486,19 +486,30 @@ function Install-DbaInstance {
                 #Exit the actions loop altogether - nothing can be installed here anyways
                 Stop-Function -Message "$computer is pending a reboot. Reboot the computer before proceeding." -Continue
             }
-            # Attempt to configure CredSSP for the remote host when credentials are defined
-            if ($Credential -and -not ([DbaInstanceParameter]$computer).IsLocalHost -and $Authentication -eq 'Credssp') {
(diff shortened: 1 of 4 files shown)


functions/Install-DbaInstance.ps1
```diff
@@ -57,13 +62,15 @@ function Install-DbaInstance {
 
     .PARAMETER Authentication
         Chooses an authentication protocol for remote connections.
-        Allowed values: 'Default', 'Basic', 'Negotiate', 'NegotiateWithImplicitCredential', 'Credssp', 'Digest', 'Kerberos'
-        If the protocol fails to establish a connection
+        Allowed values: 'Default', 'Basic', 'Negotiate', 'NegotiateWithImplicitCredential', 'Credssp', 'Digest', 'Kerberos'.
+        If the protocol fails to establish a connection (HELP: What should be the end of this sentence?)
```
nvarscar, line 66: So the idea is that it would request a confirmation for a failback authentication method, where it would create a PS Session Configuration on the remote server that would by default authenticate as specific credentials. Since this is considered insecure, there's an additional user configuration needed.

nvarscar:
looks good so far 👍

potatoqualitee:
Will merge when approved ❤️‍🔥

andreasjordan:
Stop, not so fast. We at least need the end of that sentence "If the protocol fails to establish a connection...".
And @nvarscar has a valid point with the fallback. I don't no much about the history of the command and the idea about this fallback configuration. Let me think about this today...

andreasjordan:
`$Authentication` has not the same default in `Invoke-DbaAdvancedInstall` as in `Install-DbaInstance`.

This code calls `Invoke-DbaAdvancedInstall`:
```
        # we need to know if authentication was explicitly defined
        $authBound = Test-Bound Authentication
        # wrapper for parallel advanced install
        $installAction = {
            $installSplat = $_
            if ($authBound) {
                $installSplat.Authentication = $Authentication
            }
            Invoke-DbaAdvancedInstall @installSplat
        }
```

So either we use the same default or we always call `Invoke-DbaAdvancedInstall` with `$Authentication` - don't know which is better.

andreasjordan:
Found this code in `Invoke-DbaAdvancedInstall`:

```
    $execParams = @{
        ComputerName   = $ComputerName
        ErrorAction    = 'Stop'
        Authentication = $Authentication
    }
    if ($Credential) {
        $execParams.Credential = $Credential
    } else {
        if (Test-Bound -Not Authentication) {
            # Use Default authentication instead of CredSSP when Authentication is not specified and Credential is null
            $execParams.Authentication = "Default"
        }
    }
```

The else block would not be needed if we change the parameter to `[string]$Authentication = @('Credssp', 'Default')[$null -eq $Credential],` like in `Install-DbaInstance`.

andreasjordan:
We default to CredSSP when a Credential is used. But why? We recommend using a network path for the installation media, and in this case we need CredSSP to avoid the double-hop-problem. But in case a user has no admin rights on the jump box and hence no possibility to configure CredSSP, they could just put the installation media locally on the target and use Install-DbaInstance with a credential but without CredSSP. They would need to explicitly specify `-Authentication Default` - don't they? 

I don't want to change too many things in the code, so maybe just an article in the discussion section here on github with more information about the different ways to set up the environment and the usage of Install-DbaInstance is the best way.

Just wanted to share my thoughts...

potatoqualitee:
(sorry to be clear, no intentions of merging until yall have worked it out and it's outright marked as approved by @nvarscar )

nvarscar:
> We default to CredSSP when a Credential is used. But why? We recommend using a network path for the installation media, and in this case we need CredSSP to avoid the double-hop-problem. But in case a user has no admin rights on the jump box and hence no possibility to configure CredSSP, they could just put the installation media locally on the target and use Install-DbaInstance with a credential but without CredSSP. They would need to explicitly specify `-Authentication Default` - don't they?
> 
> I don't want to change too many things in the code, so maybe just an article in the discussion section here on github with more information about the different ways to set up the environment and the usage of Install-DbaInstance is the best way.
> 
> Just wanted to share my thoughts...

Using CredSSP seemed the right thing to do. I find it weird if the user invokes a remote install and their media is only available locally. The big great idea behind both Update- and Install- commands was that you create a media repository, which you use for all the servers and CredSSP is there to cover for the double-hop issues. I would personally consider downloading the patch to the remote machine and then running the installation remotely as an anti-pattern, but then I can be wrong.

Yes users would have to override the authentication if they don't want to use credssp right now, but still want to specify credentials. I find it rare a situation, however, when a user has to specify remote credentials for a local installation. Wouldn't it be easier to launch the process as the user that has permissions on the remote box?

Anyways, if you can figure out a pattern that would consider all the possibilities, like figuring out whether a user CAN configure CredSSP, and using the information about whether the media is local or not - this one is already there, just not used for Authentication logic, iirc, maybe there could be a smarter way to implement it.

nvarscar:
> Stop, not so fast. We at least need the end of that sentence "If the protocol fails to establish a connection...". And @nvarscar has a valid point with the fallback. I don't no much about the history of the command and the idea about this fallback configuration. Let me think about this today...

The end of it should be something like, "and explicit `-Credential`s were used, a failback authentication method would be attempted that configures PSSessionConfiguration on the remote machine. This method, however, is considered insecure and would, therefore, prompt an additional confirmation when used."

nvarscar:
> Found this code in `Invoke-DbaAdvancedInstall`:
> 
> ```
>     $execParams = @{
>         ComputerName   = $ComputerName
>         ErrorAction    = 'Stop'
>         Authentication = $Authentication
>     }
>     if ($Credential) {
>         $execParams.Credential = $Credential
>     } else {
>         if (Test-Bound -Not Authentication) {
>             # Use Default authentication instead of CredSSP when Authentication is not specified and Credential is null
>             $execParams.Authentication = "Default"
>         }
>     }
> ```
> 
> The else block would not be needed if we change the parameter to `[string]$Authentication = @('Credssp', 'Default')[$null -eq $Credential],` like in `Install-DbaInstance`.

I don't see a problem calling Advanced-install command always passing $Authentication, as long as this is properly addressed in the parent comand

andreasjordan:
Ok, now I'm happy and have tested both Install-DbaInstance and Update-DbaInstance.

If run from an unelevated PowerShell session:
* If CredSSP is configured, it is used.
* If CredSSP is not configured, a warning is output (WARNING: [16:48:21][Install-DbaInstance] Failed to configure CredSSP because the PowerShell session is not elevated) and user is asked to allow fallback.

If run from an elevated PowerShell session:
* If CredSSP is configured, it is used.
* If CredSSP is not configured, CredSSP will be configured and used.

@nvarscar please have a look.

nvarscar (approved):
This looks good. I don't see any problems with backwards compatibility so far. Thanks @andreasjordan , great stuff.

potatoqualitee:
Thanks everyone, wonderful collaboration 💞

potatoqualitee:
publishing a release now

PR #8129, opened 2022-02-02
Refactoring of Copy-DbaCredential and Export-DbaCredential

andreasjordan:
General refactoring to be able to open DAC inside of Copy-DbaCredential.
Fixed some issues with QUOTENAME of credential name

Still open: Opening DAC in Export like in Copy. Will have to save all the $server variables to be able to disconnect at the end.

### DIFF
--- a/functions/Copy-DbaCredential.ps1
+++ b/functions/Copy-DbaCredential.ps1
@@ -88,6 +88,7 @@ function Copy-DbaCredential {
 
     #>
     [CmdletBinding(SupportsShouldProcess, ConfirmImpact = "Medium")]
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidUsingPlainTextForPassword", "", Justification = "For Credentials")]
     param (
         [parameter(Mandatory)]
         [DbaInstanceParameter]$Source,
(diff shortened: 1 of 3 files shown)


functions/Copy-DbaCredential.ps1
```diff
@@ -117,37 +118,54 @@ function Copy-DbaCredential {
...
-                $destServer.Credentials.Refresh()
-                $credentialName = $credential.Name
+            $destServer.Credentials.Refresh()
+            foreach ($cred in $credentialList) {
+                $credentialName = $cred.Name
 
                 $copyCredentialStatus = [pscustomobject]@{
-                    SourceServer      = $sourceServer.Name
-                    DestinationServer = $destServer.Name
+                    SourceServer      = $sourceServer.DomainInstanceName
```
potatoqualitee, line 167: Should this be ComputerName instead? I feel like there are sometimes problems with DomainInstanceName. Perhaps it's not always filled out or accurate.

wsmelton: `SourceServer` is used for the output object in all the Copy commands @potatoqualitee

wsmelton: Oh, you are referring to property name? If that, yes we use ComputerName for the value in other commands.

potatoqualitee: Thank you for double checking 😊

wsmelton: Also looks like `Connect` object does checks for populating `ComputerName`

https://github.com/dataplat/dbatools/blob/3d8d330f001853268c78aa0eebb2aef10ae9630c/functions/Connect-DbaInstance.ps1#L1573-L1592

andreasjordan: No, we don't need the computername here, but "SqlInstance". This is the output from the currently released code:

image

And we mostly fill "SqlInstance" using ".DomainInstanceName", just search for "SqlInstance  = $server.DomainInstanceName" in the code.

We can not use ".Name" of "$server", as this has the prefix "ADMIN:" when using a DAC - and I think we don't want to have that in the output.

potatoqualitee: ahhhh make sense, thanks

potatoqualitee:
Thank you 🙇🏼

PR #8181, opened 2022-02-15
move cluster perms to Add-DbaAgReplica from New-DbaAvailabilityGroup

### LINKED ISSUE #8178: Add-DbaAgReplica doesn't add req 'NT AUTHORITY\SYSTEM' perms

Yes

When creating a new AG, `New-DbaAvailabilityGroup` handles adding the necessary `NT AUTHORITY\SYSTEM` permissions for both the primary and all secondary replicas:

https://github.com/dataplat/dbatools/blob/afc87cb27e21fd08c30c847a9e64718672d7317a/functions/New-DbaAvailabilityGroup.ps1#L548-L557

But `Add-DbaAgReplica` doesn't, so attempting to fail over to a replica added with the latter command could have [issues](https://docs.microsoft.com/en-us/troubleshoot/sql/availability-groups/troubleshooting-automatic-failover-problems#case-2-insufficient-nt-authoritysystem-account-permissions).

`Add-DbaAgReplica` should also add permissions for the system account.

### Steps to Reproduce

1. Create an AG
2. Add a replica on a new instance that hasn't already been configured for AG yet via `Add-DbaAgReplica`
3. Try to failover to new replica, observe error as described in [documentation](https://docs.microsoft.com/en-us/troubleshoot/sql/availability-groups/troubleshooting-automatic-failover-problems#case-2-insufficient-nt-authoritysystem-account-permissions) due to permission issues with `NT AUTHORITY\SYSTEM` account.

Yes

### Other details or mentions

_No response_

### PowerShell Host Version

N/A

N/A

### .NET Framework Version

N/A

lowlydba:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #8178 
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
Move the `NY AUTHORITY\SYSTEM` permissioning into the Add-DbaAgReplica command so that replicas added after the AG is created have the permissions on the system account required to failover/become primary.

### DIFF
--- a/functions/Add-DbaAgReplica.ps1
+++ b/functions/Add-DbaAgReplica.ps1
@@ -133,6 +133,8 @@ function Add-DbaAgReplica {
         [DbaInstanceParameter[]]$SqlInstance,
         [PSCredential]$SqlCredential,
         [string]$Name,
+        [ValidateSet('Wsfc', 'External', 'None')]
+        [string]$ClusterType = (Get-DbatoolsConfigValue -FullName 'AvailabilityGroups.Default.ClusterType' -Fallback 'Wsfc'),
         [ValidateSet('AsynchronousCommit', 'SynchronousCommit')]
         [string]$AvailabilityMode = "SynchronousCommit",
         [ValidateSet('Automatic', 'Manual', 'External')]
@@ -256,6 +258,21 @@ function Add-DbaAgReplica {
 
                     if ($SeedingMode -and $server.VersionMajor -ge 13) {
                         $replica.SeedingMode = $SeedingMode
+                    }
+
+                    # Add cluster permissions
+                    if ($ClusterType -eq 'Wsfc') {
+                        if ($Pscmdlet.ShouldProcess($server.Name, "Adding cluster permissions for availability group named $($InputObject.Name)")) {
+                            Write-Message -Level Verbose -Message "WSFC Cluster requires granting [NT AUTHORITY\SYSTEM] a few things. Setting now."
+                            $sql = "GRANT ALTER ANY AVAILABILITY GROUP TO [NT AUTHORITY\SYSTEM]
+                                GRANT CONNECT SQL TO [NT AUTHORITY\SYSTEM]
+                                GRANT VIEW SERVER STATE TO [NT AUTHORITY\SYSTEM]"
+                            try {
+                                $null = $server.Query($sql)
+                            } catch {
+                                Stop-Function -Message "Failure adding cluster service account permissions." -ErrorRecord $_
+                            }
+                        }
                     }
 
                     $serviceAccount = $server.ServiceAccount.Trim()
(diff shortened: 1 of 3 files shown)


andreasjordan (approved):
Thanks for the pull request, looks good.

andreasjordan:
@potatoqualitee: Please do not release after merging this as we would like to address other issues discussed in #8178 as well and test them together in development before releasing.

functions/Add-DbaAgReplica.ps1
```diff
@@ -133,6 +142,8 @@ function Add-DbaAgReplica {
         [DbaInstanceParameter[]]$SqlInstance,
         [PSCredential]$SqlCredential,
         [string]$Name,
+        [ValidateSet('Wsfc', 'External', 'None')]
+        [string]$ClusterType = (Get-DbatoolsConfigValue -FullName 'AvailabilityGroups.Default.ClusterType' -Fallback 'Wsfc'),
```
potatoqualitee, line 146: I'm kind of thinking this should be `None` to avoid introducing breaking changes. WSFC has the highest amount of work frontloaded and I tend to create External or None.

potatoqualitee:
thanks for the note, @andreasjordan - i'll wait to merge after my questions have been addressed.

andreasjordan:
The other issues are addressed in #8182, so now this here can be merged.

potatoqualitee:
what is your opinion on None vs Wsfc? Perhaps we remove a failback altogether?

functions/Add-DbaAgReplica.ps1
andreasjordan: As we explicitly use the parameter when we call the command from New-DbaAvailabilityGroup, the default does not matter there.
But if someone later adds a replica (and uses a Windows Cluster), they would have to use this parameter to grant the correct rights. Having a default of Wsfc does the correct thing for most users. But grants more rights than needed for External an None. We should think about that.
I don't nee any breaking change here, can you specify? And I don't understand "WSFC has the highest amount of work frontloaded".

potatoqualitee: ahhh, well im thinking about Linux clusters here, which does not support WSFC but does support `[NT AUTHORITY\SYSTEM]` to my knowledge -- it's called LocalSystem or something. I'll test to see.

> And I don't understand "WSFC has the highest amount of work frontloaded".

Basically, WSFC is the hardest to achieve, which makes me assume it won't be used as often as None. But that may just be an unfounded assumption.

andreasjordan: Oh, I think "in the real world" (and not our labs) every production Availability Group is based on a Windows Cluster. So I personally don't know any that is not based on a Windows Cluster. I think External and None are just implemented to get this working on Linux. But I may be completly wrong...

PR #8182, opened 2022-02-16
fix/centralize ag 'create any database' perms

lowlydba:
- Not needed in New-DbaAvailabilityGroup since it is handled in both Replica and Database AG Adders
- Utilize Grant-DbaAgPermission as the route to do this perm via SMO, and only if automatic seeding
- Fix Grant-DbaAgPermission parameter parsing so it actually works

## Type of Change

 - [x] Bug fix (non-breaking change, fixes #8178 
 - [ ] New feature (non-breaking change, adds functionality, fixes # )
 - [ ] Breaking change (effects multiple commands or functionality, fixes # )
 - [ ] Ran manual Pester test and has passed (`.\tests\manual.pester.ps1`)
 - [ ] Adding code coverage to existing functionality
 - [ ] Pester test is included
 - [ ] If new file reference added for test, has is been added to github.com/dataplat/appveyor-lab ?
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose

Followup from #8181 
Fixes & standardizes the create database permission required by AGs for automatic seeding.

### Approach
- Make Grant-DbaAgPermission use SMO
- Make all other functions use Grant-DbaAgPermission
- Remove the granting from New-DbaAgAvailabilityGroup since it calls both the Add-Replica and Add-Database functions already
- 
### Commands to test

- Add-DbaAgDatabase
- New-DbaAvailabilityGroup
- Add-DbaAgReplica
- Grant-DbaAgPermission

PR #8185, opened 2022-02-20
Availability Groups: Some more fixes for granting permissions

andreasjordan:
Fixes some issues introduced or not solved by #8181 and #8182.

Mainly moved the grants to a position where the objects are already created.

Also changed Grant-DbaAgPermission as CreateAnyDatabase is granted to the AG and not a specific Login (Or am I wrong? But we only use "$server.GrantAvailabilityGroupCreateDatabasePrivilege($ag)").

### DIFF
--- a/functions/Add-DbaAgReplica.ps1
+++ b/functions/Add-DbaAgReplica.ps1
@@ -303,15 +303,6 @@ function Add-DbaAgReplica {
                         $serviceAccount = "$saName`$"
                     }
 
-                    if ($server.HostPlatform -ne "Linux") {
-                        if ($Pscmdlet.ShouldProcess($second.Name, "Granting Connect permission to service account: $serviceAccount")) {
-                            if ($SeedingMode -eq "Automatic") {
-                                $null = Grant-DbaAgPermission -SqlInstance $server -Type AvailabilityGroup -AvailabilityGroup $InputObject.Name -Login $serviceAccount -Permission CreateAnyDatabase
-                            }
-                            $null = Grant-DbaAgPermission -SqlInstance $server -Login $serviceAccount -Type Endpoint -Permission Connect
-                        }
-                    }
-
                     if ($ConfigureXESession) {
                         try {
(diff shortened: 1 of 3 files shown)


andreasjordan:
I need more time to look at the failed tests. Will do this tomorrow.

andreasjordan:
Ok, I'm down to "Failure creating login APPVYR-WIN$ | The method or operation is not implemented."

The command Grant-DbaAgPermission is trying to add a SQL Login for the service account "APPVYR-WIN$" which looks like the computer account (as the sql server is running under local system?). The command wants to read the password with Read-Host and I think this is the method that is not implemented.

Next step will be finding out what was done in the last release...

andreasjordan:
Originally, this was the code: `$null = Grant-DbaAgPermission -SqlInstance $server -Login $serviceAccount -Type Endpoint -Permission Connect` - which always failed in the Tests as the service account is the computer accout. But it was called without -EnableException, so it only produced a warning...

andreasjordan:
If the service account ends with "$", I don't grant the connect right, as the AG is set upo with a certificate and the endpoint are owned by the certificate.

potatoqualitee:
oh, oh, method not implemented = it's prompting and waiting for a response, but you probably figured that out by now

andreasjordan:
Looks good. Only test for Start-DbaDbEncryption is failing with "Cannot write into file 'C:\Program Files\Microsoft SQL Server\MSSQL13.SQL2016\MSSQL\Backup\localhost$sql2016-master-master-20220220221007.pvk'. Verify that you have write permissions, that the file path is valid, and that the file does not already exist."....

andreasjordan:
@potatoqualitee It looks like this is because Backup-DbaDbCertificate is started for 5 databases "in parallel" and "$cert.export()" does not overwrite the files. Looks like a general issue with running Start-DbaDbEncryption for more then one database at at time.

potatoqualitee:
Thanks, I feel like I wrote all that during a fever dream. I need to take a look again to see how to handle things. The random failures are driving me nuts.

PR #8188, opened 2022-02-21
Find-DbaInstance - Add SqlInstance to default output

### LINKED ISSUE #8187: Find-DbaInstance - Add SqlInstance to default output
### Summarize Functionality

Add SqlInstance to default output.

No

### Technical Details

_No response_

andreasjordan:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality, fixes #8187 )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

As most of the commands output ComputerName, InstanceName and SqlInstance as the first three properties, Find-DbaInstance should do that as well.

image

### DIFF
--- a/xml/dbatools.Format.ps1xml
+++ b/xml/dbatools.Format.ps1xml
@@ -501,6 +501,7 @@
                     <TableColumnHeader/>
                     <TableColumnHeader/>
                     <TableColumnHeader/>
+                    <TableColumnHeader/>
                 </TableHeaders>
                 <TableRowEntries>
                     <TableRowEntry>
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Very nice! I'd like this too.

potatoqualitee:
So we had left this off intentionally because it may be misleading if there's incomplete information (due to permissions/access). I wonder if we can null it out in that event.

andreasjordan:
This is the code:
https://github.com/dataplat/dbatools/blob/57e3715bf351448a1ecd683085d74c637e03aea6/bin/projects/dbatools/dbatools/Discovery/DbaInstanceReport.cs#L48-L63

So it is always build from ComputerName and InstanceName (and sometimes Port). Should never be null.

potatoqualitee:
@jpomfret and i are wondering if there is a combination where we can see we have a server with default + named instance  but can't get instance name... so results will end up listing two results both with just the servername as SqlInstance?

potatoqualitee:
if so, do we just add a warning to the docs that the sqlinstance name is best guess?

PR #8191, opened 2022-02-21
New-DbaCredential - Able to create without password

ClaudioESSilva:
## Type of Change

 - [ ] Pester test is included
 - [ ] Unit test is included
 - [x] Documentation
 - [ ] Build system

### Purpose

Add support to Credentials without password.

Also removed the arrays from the parameters. The `-Name` of the credential is unique, so even if we passed to different `-Identity` only the last one will last.

### Approach

I have removed the `Read-Host` that asked for a password.
Currently, if no password is passed, the credential is created anyway. The verbose message will mention that.

### Commands to test

``` powershell
$credentialParams = @{
    SqlInstance = "localhost"
    Name =  "https://mystorageaccount.blob.core.windows.net/mycontainer"
    Identity = 'Managed Identity'
    Force = $true
}

New-DbaCredential @credentialParams
```

### Screenshots
image

### DIFF
--- a/functions/New-DbaCredential.ps1
+++ b/functions/New-DbaCredential.ps1
@@ -64,7 +64,7 @@ function New-DbaCredential {
     .EXAMPLE
         PS C:\> New-DbaCredential -SqlInstance Server1 -Identity "MyIdentity"
 
-        It will create a credential with identity value "MyIdentity" and same name without a passowrd on server1 if it does not exist.
+        It will create a credential with identity value "MyIdentity" and same name but without a passowrd on server1 if it does not exist.
 
     .EXAMPLE
         PS C:\> $params = @{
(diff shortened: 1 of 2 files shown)


potatoqualitee:
Thank you, @ClaudioESSilva ! Please note that we do not use parametersets because they give awful errors "unresolved something something very unuseful" So instead we handle it in the `process` block.

```
        if (Test-Bound Password, Whatever) {
            Stop-Function -Message "You cannot supply both X and Y"
            return
        }
```

Also, im unsure if -NoPassword is required. What about just skipping the password section and put a verbose message that says "No password specified, creating whatever"

potatoqualitee:
The array is for multiple credentials to be created with the same parameters, no?

andreasjordan:
If we can remove the "ask for the password"-part, then I am happy to not add a new parameter. Maybe not even output a warning - the user should know that they get credentials without passwords that way.

andreasjordan:
The first two examples need to be changed.

andreasjordan:
Should $Name be of type String?

andreasjordan (approved):

potatoqualitee:
awww yeahh!

PR #8197, opened 2022-02-23
Get-DbaUserPermission - Exclude non-accessible databases

sqldeployhelmet:
If someone (me) is going to query all the servers and all the databases indiscriminately then maybe we should proactively ignore databases that are inaccessible.

## Type of Change

 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose

Get-DbaUserPermission throws errors when attempting to get permissions for a database that is not online (i.e. restoring etc.) This change will ensure the function ignores any databases where IsAccessible is False.

### Approach

Added a filter to the Where-Object statement when populating the $dbs array.

### DIFF
--- a/functions/Get-DbaUserPermission.ps1
+++ b/functions/Get-DbaUserPermission.ps1
@@ -204,15 +204,19 @@ function Get-DbaUserPermission {
             $tempdb = $server.Databases['tempdb']
 
             if ($Database) {
-                $dbs = $dbs | Where-Object { $Database -contains $_.Name }
+                $dbs = $dbs | Where-Object { $Database -contains $_.Name -and $_.IsAccessible -eq $true }
             }
 
             if ($ExcludeDatabase) {
-                $dbs = $dbs | Where-Object Name -NotIn $ExcludeDatabase
+                $dbs = $dbs | Where-Object Name -NotIn $ExcludeDatabase -and $_.IsAccessible -eq $true
             }
 
             if ($ExcludeSystemDatabase) {
-                $dbs = $dbs | Where-Object IsSystemObject -eq $false
+                $dbs = $dbs | Where-Object IsSystemObject -eq $false -and IsAccessible -eq $true
+            }
+
+            if ($null -eq $Database) {
+                Stop-Function -Message "Failure: Database is not accessible." -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
             }
 
             #reset $serverDT
--- a/functions/Get-DbaUserPermission.ps1
+++ b/functions/Get-DbaUserPermission.ps1
@@ -202,6 +202,7 @@ function Get-DbaUserPermission {
 
             $dbs = $server.Databases
             $tempdb = $server.Databases['tempdb']
+            $master = $server.Databases['master']
 
             if ($Database) {
                 $dbs = $dbs | Where-Object { $Database -contains $_.Name }
@@ -215,53 +216,80 @@ function Get-DbaUserPermission {
                 $dbs = $dbs | Where-Object IsSystemObject -eq $false
             }
 
-            #reset $serverDT
-            $serverDT = $null
+            # grab server data (just the one time):
+            Write-Message -Level Verbose -Message "Building data table for server objects"
+
+            $master.ExecuteNonQuery($endSQL)
+
+            $sqlFile = Join-Path -Path $script:PSModuleRoot -ChildPath "bin\stig.sql"
+            $sql = [System.IO.File]::ReadAllText("$sqlFile")
+            $sql = $sql.Replace("<TARGETDB>", $master.Name)
+
+            try {
+                $serverDT = $master.Query($serverSQL)
+            } catch {
+                # here to avoid an empty catch
+                $null = 1
+            }  # sometimes it complains about not being able to drop the stig schema if the person Ctrl-C'd before.
+
+            foreach ($row in $serverDT) {
+                [PSCustomObject]@{
+                    ComputerName       = $server.ComputerName
+                    InstanceName       = $server.ServiceName
+                    SqlInstance        = $server.DomainInstanceName
+                    Object             = 'SERVER'
+                    Type               = $row.Type
+                    Member             = $row.Member
+                    RoleSecurableClass = $row.'Role/Securable/Class'
+                    SchemaOwner        = $row.'Schema/Owner'
+                    Securable          = $row.Securable
+                    GranteeType        = $row.'Grantee Type'
+                    Grantee            = $row.Grantee
+                    Permission         = $row.Permission
+                    State              = $row.State
+                    Grantor            = $row.Grantor
+                    GrantorType        = $row.'Grantor Type'
+                    SourceView         = $row.'Source View'
+                }
+            }
 
             foreach ($db in $dbs) {
                 Write-Message -Level Verbose -Message "Processing $db on $instance"
 
-                try {
+                # don't attempt to execute queries on inaccessible databases:
+                if ($db.isAccessible -eq $false) {
+                    Write-Message -Level Warning -Message "Database $db on $instance is not accessible"
+                } else {
                     $db.ExecuteNonQuery($endSQL)
-                } catch {
-                    # here to avoid empty catch
-                    $null = 1
-                } # this will fail if the database is not accessible (i.e. restoring etc.)
 
-                if ($db.IsAccessible -eq $false) {
-                    Stop-Function -Message "The database $db is not accessible" -Continue
-                }
+                    $sqlFile = Join-Path -Path $script:PSModuleRoot -ChildPath "bin\stig.sql"
+                    $sql = [System.IO.File]::ReadAllText("$sqlFile")
+                    $sql = $sql.Replace("<TARGETDB>", $db.Name)
 
-                $sqlFile = Join-Path -Path $script:PSModuleRoot -ChildPath "bin\stig.sql"
-                $sql = [System.IO.File]::ReadAllText("$sqlFile")
-                $sql = $sql.Replace("<TARGETDB>", $db.Name)
-
-                #Create objects in active database
-                Write-Message -Level Verbose -Message "Creating objects"
-                try {
-                    $db.ExecuteNonQuery($sql)
-                } catch {
-                    # here to avoid an empty catch
-                    $null = 1
-                } # sometimes it complains about not being able to drop the stig schema if the person Ctrl-C'd before.
-
-                #Grab permissions data
-                if (-not $serverDT) {
-                    Write-Message -Level Verbose -Message "Building data table for server objects"
-
+                    #Create objects in active database
+                    Write-Message -Level Verbose -Message "Creating objects"
                     try {
-                        $serverDT = $db.Query($serverSQL)
+                        $db.ExecuteNonQuery($sql)
+                    } catch {
+                        # here to avoid an empty catch
+                        $null = 1
+                    } # sometimes it complains about not being able to drop the stig schema if the person Ctrl-C'd before.
+
+
+                    Write-Message -Level Verbose -Message "Building data table for $db objects"
+                    try {
+                        $dbDT = $db.Query($dbSQL)
                     } catch {
                         # here to avoid an empty catch
                         $null = 1
                     }
 
-                    foreach ($row in $serverDT) {
+                    foreach ($row in $dbDT) {
                         [PSCustomObject]@{
                             ComputerName       = $server.ComputerName
                             InstanceName       = $server.ServiceName
                             SqlInstance        = $server.DomainInstanceName
-                            Object             = 'SERVER'
+                            Object             = $db.Name
                             Type               = $row.Type
                             Member             = $row.Member
                             RoleSecurableClass = $row.'Role/Securable/Class'
@@ -277,36 +305,6 @@ function Get-DbaUserPermission {
                         }
                     }
                 }
-
-                Write-Message -Level Verbose -Message "Building data table for $db objects"
-                try {
-                    $dbDT = $db.Query($dbSQL)
-                } catch {
-                    # here to avoid an empty catch
-                    $null = 1
-                }
-
-                foreach ($row in $dbDT) {
-                    [PSCustomObject]@{
-                        ComputerName       = $server.ComputerName
-                        InstanceName       = $server.ServiceName
-                        SqlInstance        = $server.DomainInstanceName
-                        Object             = $db.Name
-                        Type               = $row.Type
-                        Member             = $row.Member
-                        RoleSecurableClass = $row.'Role/Securable/Class'
-                        SchemaOwner        = $row.'Schema/Owner'
-                        Securable          = $row.Securable
-                        GranteeType        = $row.'Grantee Type'
-                        Grantee            = $row.Grantee
-                        Permission         = $row.Permission
-                        State              = $row.State
-                        Grantor            = $row.Grantor
-                        GrantorType        = $row.'Grantor Type'
-                        SourceView         = $row.'Source View'
-                    }
-                }
-
                 #Delete objects
                 Write-Message -Level Verbose -Message "Deleting objects"
                 try {
--- a/functions/Get-DbaUserPermission.ps1
+++ b/functions/Get-DbaUserPermission.ps1
@@ -45,7 +45,7 @@ function Get-DbaUserPermission {
 
     .NOTES
         Tags: Security, User
-        Author: Brandon Abshire, netnerds.net | Josh Smith
+        Author: Brandon Abshire, netnerds.net | Josh Smith, accitentionaldba.com
 
         Website: https://dbatools.io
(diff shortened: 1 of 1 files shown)


andreasjordan:
I don't think the PR reaches the goal, as you only filter if the parameters are used.
You should also look at https://github.com/dataplat/dbatools/pull/8197/files#diff-178a92a3953c9ffd9d6513040ac43ed7a9132fc9ec384c1b5d1e9869404a2899L224-L228 - maybe just move the test up before running $endSQL and maybe not use Stop-Function but Write-Message -Level Warning.

sqldeployhelmet:
I read the initial flow completely wrong 😅 . I went ahead and just removed all dbs where IsAccessible is not true in the initial collection and threw in a warning message if after filtering out any unrequested databases none were left. (I hope 🤞🏻.)

> I don't think the PR reaches the goal, as you only filter if the parameters are used. You should also look at https://github.com/dataplat/dbatools/pull/8197/files#diff-178a92a3953c9ffd9d6513040ac43ed7a9132fc9ec384c1b5d1e9869404a2899L224-L228 - maybe just move the test up before running $endSQL and maybe not use Stop-Function but Write-Message -Level Warning.

functions/Get-DbaUserPermission.ps1
```diff
@@ -200,11 +200,15 @@ function Get-DbaUserPermission {
                 Stop-Function -Message "Failure" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
             }
 
-            $dbs = $server.Databases
+            $dbs = $server.Databases | Where-Object { $_.IsAccessible -eq $true }
```
andreasjordan, line 203: With that change, we don't need the `if ($db.IsAccessible -eq $false) {` down in line 230.

functions/Get-DbaUserPermission.ps1
```diff
@@ -200,11 +200,15 @@ function Get-DbaUserPermission {
...
             $tempdb = $server.Databases['tempdb']
 
             if ($Database) {
                 $dbs = $dbs | Where-Object { $Database -contains $_.Name }
+
+                if ($null -eq $dbs) {
+                    Write-Message -Message "Database is not available!" -Level Warning
```
andreasjordan, line 210: "available" or "accessible"?

sqldeployhelmet:
@andreasjordan : the one on line 230 wasn't mine but did help me realize what was going wrong: the `$db.ExecuteNonQuery` just above (line 225 now) gets an error if the db is restoring that is un-caught...

functions/Get-DbaUserPermission.ps1
```diff
@@ -221,7 +221,12 @@ function Get-DbaUserPermission {
             foreach ($db in $dbs) {
                 Write-Message -Level Verbose -Message "Processing $db on $instance"
 
-                $db.ExecuteNonQuery($endSQL)
+                try {
+                    $db.ExecuteNonQuery($endSQL)
+                } catch {
+                    # here to avoid empty catch
+                    $null = 1
+                } # this will fail if the database is not accessible (i.e. restoring etc.)
```
andreasjordan, line 229: How can this be if you filter those dbs out in line 203?

joshs-STCU: 🤦‍♂️ I did not mean to leave that still in there: just pulled it back out.

sqldeployhelmet: 🤦‍♂️ I meant to pull all of that back out: just removed the last of my original "fix"

andreasjordan: Excuse me for being so clear: You removed a good solution and introduced a nasty litte try-catch-fix just to suppress an error and not really solve the problem. There are issues with this command and we sould solve (!) them in a good way. Perhaps I represent a minority opinion, let's see what the other contributors say...

sqldeployhelmet: I can certainly go the other direction: I just took my cues from the existing code...

sqldeployhelmet: Okay, I took a stab at cleaning it up. As a very new contributor I didn't want to assume I should do more extensive re-writes. This current version moves the server level permissions to a single check (without the need for a "did we do it already flag" at the beginning of the loop on `$sqlInstances` and then moves the check to see if a db is accessible into the loop on `$dbs` and displaying a warning message if the database is not queryable...

sqldeployhelmet: @andreasjordan Thanks for the positive feedback that my initial changes were on the right track. I'm just super good at second guessing myself...

andreasjordan:
Sorry to say: You still have no idea how this thing works. Please have a look at the contents of the loaded sql file and the sql code in the begin block. Then ask yourself which code has to executed in which database - that might help to find some of the bugs in here. And if you wrap something in a try-catch with an empty catch: That means that you just don't care about the results gathered inside of the try. Is that what you want here? You don't care if results are produced or not?

potatoqualitee:
Thank you both so much, I will merge once @andreasjordan gives the approval 🙇🏼

sqldeployhelmet:
> Thank you both so much, I will merge once @andreasjordan gives the approval 🙇🏼

I have a couple more changes to make this morning I think that will make it better as well 😄

functions/Get-DbaUserPermission.ps1
```diff
@@ -215,17 +215,65 @@ function Get-DbaUserPermission {
...
-            #reset $serverDT
-            $serverDT = $null
+            # grab server data (just the one time):
+            Write-Message -Level Verbose -Message "Building data table for server objects"
 
-            foreach ($db in $dbs) {
-                Write-Message -Level Verbose -Message "Processing $db on $instance"
+            $sqlFile = Join-Path -Path $script:PSModuleRoot -ChildPath "bin\stig.sql"
+            $sql = [System.IO.File]::ReadAllText("$sqlFile")
+            $sql = $sql.Replace("<TARGETDB>", $master.Name)
```
andreasjordan, line 223: Why don't you use the string "master"?

functions/Get-DbaUserPermission.ps1
```diff
@@ -215,17 +215,65 @@ function Get-DbaUserPermission {
...
 
-                $db.ExecuteNonQuery($endSQL)
+            try {
+                $master.ExecuteNonQuery($sql)
+            } catch {
+                # here to avoid an empty catch
+                $null = 1
```
andreasjordan, line 229: So you don't care if executing the $sql fails or not? The rest of the command would just run fine in that case? Really?

functions/Get-DbaUserPermission.ps1
```diff
@@ -215,17 +215,65 @@ function Get-DbaUserPermission {
...
+            } # sometimes it complains about not being able to drop the stig schema if the person Ctrl-C'd before.
 
-                if ($db.IsAccessible -eq $false) {
-                    Stop-Function -Message "The database $db is not accessible" -Continue
+            try {
+                $serverDT = $master.Query($serverSQL)
```
andreasjordan, line 233: Why do you execute this query against master? Why not tempdb?

functions/Get-DbaUserPermission.ps1
```diff
@@ -215,17 +215,65 @@ function Get-DbaUserPermission {
...
+                        Grantee            = $row.Grantee
+                        Permission         = $row.Permission
+                        State              = $row.State
+                        Grantor            = $row.Grantor
+                        GrantorType        = $row.'Grantor Type'
+                        SourceView         = $row.'Source View'
+                    }
                 }
+            } catch {
+                Write-Message -Level warning "An error occured while retrieving server permissions."
```
andreasjordan, line 256: Why don't you use Stop-Function to be able to raise an exception?

functions/Get-DbaUserPermission.ps1
```diff
@@ -215,17 +215,65 @@ function Get-DbaUserPermission {
...
+            }
+
+            #delete objects from running server permissions query
```
andreasjordan, line 259: Why do you have to do that?

functions/Get-DbaUserPermission.ps1
```diff
@@ -215,17 +215,65 @@ function Get-DbaUserPermission {
...
+
+            # now that we have server level permissions get any database level perms
+            foreach ($db in $dbs | Where-Object { $_.isAccessible -eq $true }) {
+                Write-Message -Level warning -Message "Database $db on $instance is not accessible and will be skipped."
+            }
+
+            foreach ($db in $dbs | Where-Object { $_.isAccessible -eq $true }) {
+                Write-Message -Level Verbose -Message "Processing $db on $instance"
+
+                $db.ExecuteNonQuery($endSQL)
```
andreasjordan, line 276: Why do you do that (again)?

functions/Get-DbaUserPermission.ps1
```diff
@@ -240,23 +288,16 @@ function Get-DbaUserPermission {
                     $null = 1
```
andreasjordan, line 288: Again: Why don't you care?

functions/Get-DbaUserPermission.ps1
```diff
@@ -271,35 +312,8 @@ function Get-DbaUserPermission {
...
-                        Securable          = $row.Securable
-                        GranteeType        = $row.'Grantee Type'
-                        Grantee            = $row.Grantee
-                        Permission         = $row.Permission
-                        State              = $row.State
-                        Grantor            = $row.Grantor
-                        GrantorType        = $row.'Grantor Type'
-                        SourceView         = $row.'Source View'
-                    }
+                    Write-Message -Level warning -Message "There was an error retrieving the permissions for $db on $instance."
```
andreasjordan, line 316: Why don't you use Stop-Function?

andreasjordan (changes requested):
And one last question: Why don't you clean up after all. You leave the stig schema behind...

I my eyes, this PR is still far away from beeing approved.

functions/Get-DbaUserPermission.ps1
sqldeployhelmet: I wanted to keep it consistent with how the SQL was executed through the stored function. I have dropped the $master object.. This now uses the string master and the SQL is executed via the $tempdb object.

functions/Get-DbaUserPermission.ps1
sqldeployhelmet: The code that existed in the function indicated that this try/catch was here simply becuase it _might_ fail if the objects were not cleaned up from a previous run (where a user may have control c'd their way out of the code. This simply was not part of the function I was trying to fix and it seems to have been working fine to date. If you have something specific you would like to see here please feel free to add it.

functions/Get-DbaUserPermission.ps1
sqldeployhelmet: Personal preference. It now executes against tempdb (see conversation further up in the code).

functions/Get-DbaUserPermission.ps1
sqldeployhelmet: Because I was unaware of Stop-Function. Stop-Function is now used in the catch blocks.

functions/Get-DbaUserPermission.ps1
sqldeployhelmet: See comment above: this was working before I started making any changes and is not related to problem I am trying to solve with my update.

functions/Get-DbaUserPermission.ps1
sqldeployhelmet: As stated further up Stop-Function is now implemented.

functions/Get-DbaUserPermission.ps1
sqldeployhelmet: Asked and answered.

functions/Get-DbaUserPermission.ps1
sqldeployhelmet: I wasn't looking to completely re-factor this function: I simply wanted to fix an issue where if run against a database that was not accessible it threw an error and failed. 

The [current version of the function](https://github.com/dataplat/dbatools/blob/development/functions/Get-DbaUserPermission.ps1) Currently acts in the same way: creating the objects, running the query and deleting the objects for each database it is running against on an instance. If there is a specific fix you would like to see here please feel free to update it or tell me specifically what you would like to see.

sqldeployhelmet:
@andreasjordan: I've made further changes but I would like to re-iterate that the fix I am trying to implement is the ability to run this function against my estate that won't crash if it encounters a database that is not accessible. I feel like there are a number of things I am continuing to do "wrong" but that I am expected to guess at what solution would be found acceptable. I came into this as a relatively new contributor that was trying to solve a very specific issue the function was having when I tried to use it in my environment and my mindset was to change as little as possible so that it would function as expected without re-writing it. That being said:

* If there are improvements to existing code from the function that are _required_ I will happily implement those if they are explicitly defined.
* If there are changes that are desired for other reasons I am even happy to implement them if they are explicitly defined.

I would very much like to get this fixed so I (and others) can implement this to gather estate wide permissions if they like.

functions/Get-DbaUserPermission.ps1
andreasjordan: But $sql is not about cleanup, but about creating the STIG schema. If that fails, maybe an old version for another database is used and would create wrong results.

functions/Get-DbaUserPermission.ps1
andreasjordan: You are (mostly) right here. And executing the $endSQL in the contect of tempdb would clean up after the usage of STIG. But to avoid left-overs from another run, cleanup is done at the beginning, so once at the end of the instance would be enough.

functions/Get-DbaUserPermission.ps1
andreasjordan: Executing $endSQL in a database context other than tempdb does: Nothing - because there is (or should be) no schema STIG.

andreasjordan:
I have taken some time today to get a closer look at the code and did some refactorting myself. You can see at #8220 what my solution would be.

I think contributing to an open source project is not just changing code so that it does not error out on my local maschine but also about solving problems I find along the way. And understanding the general structure of the code and helper commands like Stop-Function is key to successful contributing. And personally I only change a line of code if I completly understand what the command does that I'm about to change.

potatoqualitee:
While I agree that it's wonderful to fix issues along the way, I think that should be an ideal and not a baseline. This is for a number of reasons, including the time limitations of the contributor and reducing the barrier to contributions, especially those that help dbatools work better within various environments.

If the current code is acceptable for the scope that @sqldeployhelmet is trying to address, I'd like to merge it. 

From there, a nice-to-have PR can be submitted to address the cleanup (which I'd also appreciate for my own machines) 🙏🏼

andreasjordan:
image

So I just used the code from this branch and it thwors a lot of warnings. Why? Because of:
```
            # now that we have server level permissions get any database level perms
            foreach ($db in $dbs | Where-Object { $_.isAccessible -eq $true }) {
                Stop-Function -Message "Database $db on $instance is not accessible." -Continue -EnableException $EnableException
            }
```

andreasjordan:
I will stop looking at this pull request for a while.

andreasjordan:
One more thing: There is no active test for this command. The one on AppVeyor is disabled and the one on GitHub Actions is not correct. So at the moment: This code is not tested!

potatoqualitee:
Thank you 🙇🏼  I'll see what I can do regarding the tests. I'll also propose a change to be verbose instead of of warning

functions/Get-DbaUserPermission.ps1
```diff
@@ -215,17 +214,65 @@ function Get-DbaUserPermission {
...
+            Write-Message -Level Verbose -Message "Deleting objects"
+            try {
+                $tempdb.ExecuteNonQuery($endSQL)
+            } catch {
+                # here to avoid an empty catch
+                $null = 1
+            }
+
+            # now that we have server level permissions get any database level perms
+            foreach ($db in $dbs | Where-Object { $_.isAccessible -eq $true }) {
```
potatoqualitee, line 268:
```suggestion
            foreach ($db in $dbs | Where-Object isAccessible) {
```

The above is faster and prettier 💅🏼

potatoqualitee:
Thanks again to you both. 

I just reviewed this PR as well as Andreas' PR at https://github.com/dataplat/dbatools/pull/8220 and believe that his proposed PR both addresses both of your concerns outlined in this PR, including tests. 

@sqldeployhelmet can you please take a look and let me know if his PR suits your needs? If so, I will merge that one as it's a bit more concise and does a couple more things. Once approved, I'll merge and publish to the gallery 👍🏼

sqldeployhelmet:
@potatoqualitee : Looks good to me, assuming I am reading it correctly.

@andreasjordan : thank you for attempting to help me, I know it's not necessarily easy across so many time zones. 

Maybe if I attempt to contribute again in the future I'll pick something that is earmarked for beginners. I very much appreciate the amount of care and work that goes into dbatools!

PR #8214, opened 2022-03-02
Invoke-DbaDbSafeShrink

tcartwright:
- new safe shrink method. safely shrinks dbs without causing fragmentation

## Type of Change

   -  After I had to install an older version of PSScriptAnalyzer all it printed out was "WARNING: Exiting...", does that mean passed?
 - [x] Pester test is included
 - [x] Unit test is included
 - [x] Documentation
 - [ ] Build system

### Purpose
 This function is written as to follow the following recommendation from Paul Randal: https://www.sqlskills.com/blogs/paul/why-you-should-not-shrink-your-data-files/
        
        QUOTE:
            The method I like to recommend is as follows:
            - Create a new filegroup
            - Move all affected tables and indexes into the new filegroup using the CREATE INDEX … WITH (DROP_EXISTING = ON) ON syntax, to move the tables and remove fragmentation from them at the same time
            - Drop the old filegroup that you were going to shrink anyway (or shrink it way down if its the primary filegroup)

### Approach
Does not use the normal shrink method. Follows the suggestions by Paul Randal

### Commands to test
See examples

### DIFF
--- a/bin/diagnosticquery/SQLServerDiagnosticQueries_2012.sql
+++ b/bin/diagnosticquery/SQLServerDiagnosticQueries_2012.sql
@@ -1,7 +1,7 @@
 
 -- SQL Server 2012 Diagnostic Information Queries
 -- Glenn Berry 
--- Last Modified: February 1, 2022
+-- Last Modified: March 2, 2022
 -- https://glennsqlperformance.com/
 -- https://sqlserverperformance.wordpress.com/
 -- YouTube: https://bit.ly/2PkoAM1 
--- a/bin/diagnosticquery/SQLServerDiagnosticQueries_2014.sql
+++ b/bin/diagnosticquery/SQLServerDiagnosticQueries_2014.sql
@@ -1,7 +1,7 @@
 
 -- SQL Server 2014 Diagnostic Information Queries
 -- Glenn Berry 
--- Last Modified: February 1, 2022
+-- Last Modified: March 2, 2022
 -- https://glennsqlperformance.com/
 -- https://sqlserverperformance.wordpress.com/
 -- YouTube: https://bit.ly/2PkoAM1 
--- a/bin/diagnosticquery/SQLServerDiagnosticQueries_2016.sql
+++ b/bin/diagnosticquery/SQLServerDiagnosticQueries_2016.sql
@@ -1,7 +1,7 @@
 
 -- SQL Server 2016 Diagnostic Information Queries
 -- Glenn Berry 
--- Last Modified: February 1, 2022
+-- Last Modified: March 2, 2022
 -- https://glennsqlperformance.com/
 -- https://sqlserverperformance.wordpress.com/
 -- YouTube: https://bit.ly/2PkoAM1 
--- a/bin/diagnosticquery/SQLServerDiagnosticQueries_2016SP2.sql
+++ b/bin/diagnosticquery/SQLServerDiagnosticQueries_2016SP2.sql
@@ -1,7 +1,7 @@
 
 -- SQL Server 2016 SP2 Diagnostic Information Queries
 -- Glenn Berry 
--- Last Modified: February 1, 2022
+-- Last Modified: March 2, 2022
 -- https://glennsqlperformance.com/
 -- https://sqlserverperformance.wordpress.com/
 -- YouTube: https://bit.ly/2PkoAM1 
--- a/bin/diagnosticquery/SQLServerDiagnosticQueries_2017.sql
+++ b/bin/diagnosticquery/SQLServerDiagnosticQueries_2017.sql
@@ -1,7 +1,7 @@
 
 -- SQL Server 2017 Diagnostic Information Queries
 -- Glenn Berry 
--- Last Modified: February 1, 2022
+-- Last Modified: March 6, 2022
 -- https://glennsqlperformance.com/
 -- https://sqlserverperformance.wordpress.com/
 -- YouTube: https://bit.ly/2PkoAM1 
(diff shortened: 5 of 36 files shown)


tcartwright:
First time I've ever submitted a PR for you guys. Hopefully I did not bungle it.

tcartwright:
Can someone please advise me what I did wrong with the tests? I have been trying to run them locally without any success. I based the test I wrote off of the existing shrink test. When I try to run the tests locally, this is what I get:

PS C:\Users\tim.cartwright.VGNET\source\repos\dbatools> .\tests\manual.pester.ps1 -Path Invoke-DbaDbSafeShrink.Tests.ps1        

WARNING: Please install PSScriptAnalyzer 1.18.2
WARNING:      Install-Module -Name PSScriptAnalyzer -RequiredVersion '1.18.2'
WARNING: Exiting...

wsmelton:
Initial read (skim) this seems to include functionality that we already have written into `Invoke-DbaDbShrink`. I'm not fond of making something that should not be done so easily accessible to users. Seem to recall desire for this command has been discussed or come up before and we intentionally have not created it.

If the command moves forward I'd rather have it use current commands for processing where possible instead (e.g. use `Invoke-DbaDbShrink` to actually perform the shrink). The internal functions in the command also need to be properly named with `Verb-Noun` nomenclature.

**Please remove the psd1 and psm1 file changes from the PR** Maintainers will update those for release once the command is published. (You can do hard reset on those to revert them with no commit required.)

tcartwright:
@potatoqualitee Chrissy, would you mind helping me? Or one of the other admins? I was able to pass the tests locally, but this last appveyor failure has me confused.

https://ci.appveyor.com/project/dataplat/dbatools/builds/42783347/job/xcq82wkv7yx6hp54

	Describe : Invoke-DbaDbSafeShrink Integration Tests
	Context  : Verifying Database is shrunk
	Name     : It Error occurred in Context block
	Result   : Failed
	Message  : Cannot bind argument to parameter 'SqlInstance' because it is null.

tcartwright:
@wsmelton  I will make the hard resets you suggested, and also rename the internal functions. However, this shrink is VERY different than the built in shrink functionality. It does not cause any additional fragmentation at all. Also, it only allows for shrinking of the MDF files, so the original shrink is still very valid.  

It is also much much faster to shrink this way than the built in shrinkfile.

wsmelton:
So why not put this functionality to the original function.

tcartwright:
@wsmelton The original Invoke-DbaDbShrink is still valid, and can still be used. It will however, fragment your indexes 99% when using it to shrink data files.  I left it alone so as not to confuse the user as to the intention. One provides the old way of shrinking, and the new method. We could combine them into one, but then I think a parameter would need to be added for shrinking strategy allowing the user to choose which method to use to shrink.

The new method is based upon the recommendations made by Paul Randal in this article and :  https://www.sqlskills.com/blogs/paul/why-you-should-not-shrink-your-data-files/

    QUOTE:
        The method I like to recommend is as follows:
        - Create a new filegroup
        - Move all affected tables and indexes into the new filegroup using the CREATE INDEX … WITH (DROP_EXISTING = ON) ON syntax, to move the tables and remove fragmentation from them at the same time
        - Drop the old filegroup that you were going to shrink anyway (or shrink it way down if its the primary filegroup)

tcartwright:
Ok, the latest appveyor run passed all my tests. It only failed for the Get-DbaKbUpdate method.

	Describe : Get-DbaKbUpdate Integration Tests
	Context  : 
	Name     : It Call with multiple KBs
	Result   : Failed
	Message  : Expected 4057119 to be found in collection @(), but it was not found.

	Describe : Get-DbaKbUpdate Integration Tests
	Context  : 
	Name     : It Calling script uses a variable named filter
	Result   : Failed
	Message  : Expected 4564903, but got $null.

	Describe : Get-DbaKbUpdate Integration Tests
	Context  : 
	Name     : It successfully connects and parses link and title
	Result   : Failed
	Message  : Expected 4057119, but got $null.

potatoqualitee:
Thanks so much, @tcartwright ! I'll check to see what's going on with the KbUpdate commands and disable if required. 

I do agree with @wsmelton in that this should be part of Invoke-DbaDbShrink, perhaps witin a new `-Safe` parameter.

tcartwright:
> Thanks so much, @tcartwright ! I'll check to see what's going on with the KbUpdate commands and disable if required.
> 
> I do agree with @wsmelton in that this should be part of Invoke-DbaDbShrink, perhaps witin a new `-Safe` parameter.

The reason I disagree with that is that many users may or may not fully read the documentation on how one switch makes a difference and causes a completely different shrinking mechanism. If that is the final decision here, I will abide by it however. :stuck_out_tongue_winking_eye:

tcartwright:
@potatoqualitee, @wsmelton How about instead of combining it, calling it something different? TBH, I am not fond of the name 'Invoke-DbaDbSafeShrink'.

wsmelton:
> The reason I disagree with that is that many users may or may not fully read the documentation on how one switch makes a difference and causes a completely different shrinking mechanism.

That is irrelevant if they have already been using it as-is. The purpose would be that we can provide them an alternative method within the same command to shrink the file(s). In most cases shrinking is done in an emergency situation and they may not have the space or storage to perform it properly.

wsmelton:
In reference to users reading our documentation, more are reading help on current commands than on new commands being added. It is rare that they are looking at new commands because the simple fact we have 650+ now. The die-hard users will look at our release notes on GitHub, but those are only a small percentage of our users (IMO).

We strive to have our current commands follow best practices both from Microsoft's side and the community. (We link to Paul's post in `Invoke-DbaDbShrink`) We cannot however start a pattern of adding a new command because it is a "safer method" over an original or current command. That will get out of control and cause more effort for the maintainers than is needed.

The contribution is appreciated but for future reference creating a feature request issue for things like this help to have a more thorough discussion around this to plan it and get full approval to move forward. Going straight to PR just extends the time and effort on both sides to getting it merged and published.

tcartwright:
I personally think that combining them is the wrong way to go. They have different inputs, outputs, and methodology. However, since it seems that is the consensus, I will work on combining the two.

wsmelton:
The dynamic of using parameters for different workflows is a common pattern in our module (and most in PowerShell), that's where parameter sets generally are used. We do not use them exclusively as we like to put out more friendly error messages (think Windows PowerShell more than PS 7).

Your command only adds one parameter, `MinimumFreeSpace` which based on quick check I see no reason you could not use the current `-PercentFreeSpace` and calculate the `int64` value in the code.

In reference to output, we already have some commands that based on parameter(s) will give you different properties on the output object. (not an uncommon pattern in PowerShell in general) 

I'd probably call the parameter `-SafeMethod` also to make it more clear this triggers a different _method_ or workflow in the command.

If you see large chunks of logic are needed you can move those to an internal function if it makes sense to and then just have the command call it as a wrapper.

PR #8220, opened 2022-03-07
Get-DbaUserPermission - Refactoring

andreasjordan:
## Type of Change

 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

Now the command does not fail on non-accessible databases but just throws a warning and skips the database. (We can still discuss if the database should be silently ignored).
Server level security is now reported even if no database is processed.
File with the stig.sql is only read once.
Cleanup sql is now executed in the correct database (tempdb instead of user database) and only once per database before processing and once per instance after processing all databases.
try-catch now reports failures and not just silently ignores them.

### DIFF
--- a/functions/Connect-DbaInstance.ps1
+++ b/functions/Connect-DbaInstance.ps1
@@ -488,7 +488,7 @@ function Connect-DbaInstance {
                 - Get-DbaLogin
                 - Set-DbaLogin
                 - Get-DbaDefaultPath
-                - Get-DbaUserPermissions
+                - Get-DbaUserPermission
                 - Get-DbaXESession
                 - New-DbaCustomError
                 - Remove-DbaCustomError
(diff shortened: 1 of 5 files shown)


andreasjordan:
Test always fails on AppVeyor for unknown reason with "Log entry string is too long. A string written to the event log cannot exceed 32766 characters." while executing `$tempdb.ExecuteNonQuery($createStigSQL)`. So I leave this test disabled. But I fixed the test on GitHub Action, that works now.

andreasjordan:
Still in question: Does `$sqlFile = Join-Path -Path $script:PSModuleRoot -ChildPath "bin\stig.sql"` work on linux as the childpath includes a windows folder devider.

potatoqualitee:
awesome, thank you so much! 🙇🏼

potatoqualitee:
awesome, thank you so much! 🙇🏼

PR #8221, opened 2022-03-07
Install-DbaInstance - Support installation of Failover Cluster Instance

andreasjordan:
## Type of Change

 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
Support a different ACTION than just "Install". This way, a cluster installation is supported without the need for a configuration file.

### Approach
Check for the existance of the key "ACTION" in the hashtable provided with the parameter Configuration.

### Commands to test
Will be provided later...
To see how a cluster installation is possible now: https://github.com/dataplat/dbatools/discussions/7447

### Screenshots
Will be provided later...

### DIFF
--- a/functions/Install-DbaInstance.ps1
+++ b/functions/Install-DbaInstance.ps1
@@ -579,6 +579,15 @@ function Install-DbaInstance {
                 } catch {
                     Stop-Function -Message "Failed to read config file $ConfigurationFile" -ErrorRecord $_
                 }
+            } elseif ($Configuration.ACTION) {
+                # build minimal config if a custom ACTION is provided
+                $config = @{
+                    $mainKey = @{
+                        INSTANCENAME = $instance
+                        FEATURES     = $featureList
+                        QUIET        = "True"
+                    }
+                }
             } else {
                 # determine a default user to assign sqladmin permissions
                 if ($Credential) {
--- a/functions/Install-DbaInstance.ps1
+++ b/functions/Install-DbaInstance.ps1
@@ -58,6 +58,13 @@ function Install-DbaInstance {
         For example, to define a custom server collation you can use the following parameter:
         PS> Install-DbaInstance -Version 2017 -Configuration @{ SQLCOLLATION = 'Latin1_General_BIN' }
 
+        As long as you don't specify the item ACTION, some items are already set by the command, like SQLSYSADMINACCOUNTS or *SVCSTARTUPTYPE.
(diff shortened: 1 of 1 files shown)


potatoqualitee:
nice! @nvarscar whatcha think?

functions/Install-DbaInstance.ps1
```diff
@@ -579,6 +579,15 @@ function Install-DbaInstance {
                 } catch {
                     Stop-Function -Message "Failed to read config file $ConfigurationFile" -ErrorRecord $_
                 }
+            } elseif ($Configuration.ACTION) {
```
nvarscar, line 582: Hmm.. So if someone specifies ACTION=Install, this would not provide a full configuration, but only a few items. What if this one checked also whether provided action is NOT install?
```suggestion
            } elseif ($Configuration.ACTION -and $Configuration.ACTION -ne "Install") {
```

andreasjordan: Even though this change would have no impact on my use cases: Why would anyone specify a Configuration with ACTION = "Install" unless they had a very specific configuration?
Maybe someone wants to test an installation with only the bare minimum as parameters to show what the setup.exe uses as defaults. They can still fall back to use ConfigurationFile like I do now, but the command would be more flexible if they key ACTION inside of Configuration would be the switch to use a totally custom configuration file, just without the need to use a file. 

If you still like to implement you suggested change, then I will change that. No problem.

nvarscar: I see where you're getting at. It would make sense to implement something like this, but the documentation of the Configuration parameter should reflect it then, as it's quite an obscure behavior.

andreasjordan: Good point, havn't changed documentation to just get feedback on this first. As you are generally ok with the change I will try to propose a change to the documentation during today.

andreasjordan: Documentation is updated. Please have a look.
If we add ` -and $Configuration.ACTION -ne "Install"` in the test I would change "As long as you don't specify the item ACTION" to "As long as you don't specify the item ACTION with a value other than 'Install'".

andreasjordan:
Hi @nvarscar - should I still add ` -and $Configuration.ACTION -ne "Install"` or not? What do you think?

nvarscar (approved):
No, the change should be fine as it is

potatoqualitee:
Thank you both 🙇🏼

PR #8222, opened 2022-03-08
Set-DbaNetworkConfiguration - Add parameter IpAddress

andreasjordan:
## Type of Change

 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
In addition to configure IPAll, I would like to be able to configure the instance to listen on specific IP addresses only. This could be used to have multiple instances on one maschine listening all on port 1433 but on individual ip addresses.

### Approach
I added the parameter ListenOnIPAddressOnly.

### Commands to test
```
Set-DbaNetworkConfiguration -SqlInstance sql01\sqlin01 -ListenOnIPAddressOnly 192.168.3.31:1433
```

### DIFF
--- a/functions/Set-DbaNetworkConfiguration.ps1
+++ b/functions/Set-DbaNetworkConfiguration.ps1
@@ -41,7 +41,7 @@ function Set-DbaNetworkConfiguration {
         Will enable the TCP/IP protocol if needed.
         Will set TcpIpProperties.ListenAll to $true if needed.
 
-    .PARAMETER ListenOnIPAddressOnly
+    .PARAMETER IpAddress
         Configures the instance to listen on specific IP addresses only. Listening on all other IP addresses will be disabled.
         Takes an array of string with either the IP address (for listening on a dynamic port) or IP address and port seperated by ":".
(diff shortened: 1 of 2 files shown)


potatoqualitee:
This naming thing is rough. I think if we do stay with this long one, it'd be `ListenOnlyOnIPAddress` but would `IPAddress` suffice or not convey enough? /cc @wsmelton

andreasjordan:
I'm ok with a better (and shorter) name for the parameter. It was just the first I came up with...

wsmelton:
I'd go with something simple as just `-StaticIpAddress` or if we "need" the listen word `-ListenOnStaticIp`

potatoqualitee:
ohh i like `-StaticIpAddress` that feels more certain and clear than `IpAddress`.

andreasjordan:
But isn't "static" / "dynamic" for the ports? With this new parameter I can set a static port or dynamic ports.

The parameter is about using only one or more specific IP addresses instead of the usual ListenAll / IPAll setting.

I agree in removing the "Only". So maybe only `-IpAddress`?

potatoqualitee:
dankeschön 😊

andreasjordan:
Bitteschön!

PR #8226, opened 2022-03-11
Invoke-ManagedComputerCommand - Fix to enable retry again

andreasjordan:
## Type of Change

 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
When the new code pattern like in Set-DbaNetworkConfiguration, the Invoke-Command2 does not throw an exception and thus the retry dows not work.

### Approach
This is just a quick fix to enable the retry again. I will have a deeper look later and change all the scriptblocks executed with Invoke-ManagedComputerCommand to follow the new code pattern like in Set-DbaNetworkConfiguration. This way we have verbose mesages from the remote execution and can analyse issues more easily.

### Commands to test
I've seen this with this command, but only because the direct wmi call produced an empty WMI object.
```
Set-DbaNetworkConfiguration -SqlInstance SQL01\SQLINST01 -IpAddress 192.168.3.41:1433 -RestartService -Confirm:$false
```

### Screenshots
image

### DIFF
--- a/functions/Set-DbaNetworkConfiguration.ps1
+++ b/functions/Set-DbaNetworkConfiguration.ps1
@@ -143,9 +143,15 @@ function Set-DbaNetworkConfiguration {
             $targetConf = $args[0]
             $changes = @()
             $verbose = @()
+            $verbose += $setupVerbose
             $exception = $null
 
             try {
+                $verbose += "Starting code from Set-DbaNetworkConfiguration"
+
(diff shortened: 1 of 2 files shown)


andreasjordan:
We don't have a specific test for that, so lets run all tests...

potatoqualitee:
Looks good, thank you!

PR #8228, opened 2022-03-11
Get-DbaUserPermission - Refactoring again

andreasjordan:
Like in #8220, but this time without introducing a bug.


andreasjordan:
Test should fail - just to see if test is effective.

Then I will fix the bug in the next commit.

andreasjordan:
Ok, Test works:
image

Now let's fix the bug....

Stuart-Moore:
That's fixed my issue 👍 (https://github.com/Stuart-Moore/sqlbitsdb1/runs/5520422771?check_suite_focus=true)

Thank you for sorting so quickly

potatoqualitee:
awesome, thank you both so much! thank you for the quick turnaround too, @andreasjordan 🙇🏼

PR #8230, opened 2022-03-13
Copy-DbaDbMail - support for getting mail server passwords - fixes #7824

### LINKED ISSUE #7824: Database mail passwords aren't synced

Yes

SQL Mail was unable to send email until the password was reentered

### Steps to Reproduce

```powershell
```
Start-SqlMigration -source db1 -destination db2 -exclude databases

Yes

### Other details or mentions

The password in this case was 30+ characters. As soon as the password was entered in again email started being sent.

### PowerShell Host Version

PSVersion                      5.1.17763.2183
BuildVersion                   10.0.17763.2183

12.0.6164.21

tkwj:
## Type of Change

 - [] Bug fix (non-breaking change, fixes # )
 - [x] New feature (non-breaking change, adds functionality, fixes #7824  )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
Copies over basic authentic mail server passwords when possible, warns the user otherwise.

### Approach
First, the collection of credential_name and account_id is fetched from the source server. Credential_name (guid string) is the only unique value for credentials that is also in the Get-DecryptedObject output. Unfortunately, SMO does not expose this, so we must use a query. This is done before the foreach over MailServers, so that this call can be done once per source.
This $credentialName will be null for Windows Auth and Anonymous Auth accounts, so the whole next section is skipped if this is null.
Get-DecryptedObject is then used to fetch the password from the source, using similar code as Copy-DbaCredential. This is set to run without exception, so that failure will result in falling back to the old logic of a blank password. In this case, a warning message is added to inform the user.

### Commands to test
Copy-DbaDbMail -Source SQL1 -Destination SQL2

### Learning
In Database Mail, when the use picks the Basic Authentication for a mail account, this is stored as a Credential with a new GUID as its name. The email user/password are stored in the credential’s identity and secret. The credential_id is stored in msdb.dbo.sysmail_server which has a 1:1 relationship with msdb.dbo.sysmail_account.
I considered two ways to fix this. 
1) Get the decrypted value from the source with Get-DecryptedObject and insert it into the text that comes from the SMO MailServer.Script() command that is already being used.
2) Run a Copy-DbaCredential and then update the msdb.dbo.sysmail_server.credential_id to the new credential.

Option 2 seemed worse because sysmail_server has no stored procedures available that allow linking an existing credential, so you’d have to update a system table directly.
Option 1 follows the same pattern as Copy-DbaCredential (and this PR is heavily inspired by that code) so seemed like the better idea.

I have tested this on SQL 2012, 2014, 2016, 2017, and 2019, with remote DAC on and off, and with combinations of one and multiple servers.

### DIFF
--- a/functions/Copy-DbaDbMail.ps1
+++ b/functions/Copy-DbaDbMail.ps1
@@ -264,6 +264,13 @@ function Copy-DbaDbMail {
             $sourceMailServers = $sourceServer.Mail.Accounts.MailServers
             $destMailServers = $destServer.Mail.Accounts.MailServers
 
+            Write-Message -Message "Getting mail server credentials." -Level Verbose
+            $credentialAccounts = @($sourceServer.Query("SELECT credentials.name AS credential_name, sysmail_server.account_id FROM sys.credentials
+            JOIN msdb.dbo.sysmail_server ON credentials.credential_id = sysmail_server.credential_id"))
+            if ($credentialAccounts.Count -gt 0) {
+                $decryptedCredentials = Get-DecryptedObject -SqlInstance $sourceServer -Type Credential | Where-Object { $_.Name -in $credentialAccounts.credential_name }
+            }
+
             Write-Message -Message "Migrating mail servers." -Level Verbose
             foreach ($mailServer in $sourceMailServers) {
                 $mailServerName = $mailServer.name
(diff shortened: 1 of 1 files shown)


andreasjordan:
Tests complain about formatting: Copy-DbaDbMail.ps1 is not compliant with the OTBS formatting style. Please run Invoke-DbatoolsFormatter against the failing file and commit the changes.

functions/Copy-DbaDbMail.ps1
```diff
@@ -264,6 +264,13 @@ function Copy-DbaDbMail {
             $sourceMailServers = $sourceServer.Mail.Accounts.MailServers
             $destMailServers = $destServer.Mail.Accounts.MailServers
 
+            Write-Message -Message "Getting mail server credentials." -Level Verbose
+            $credentialAccounts = @($sourceServer.Query("SELECT credentials.name AS credential_name, sysmail_server.account_id FROM sys.credentials
```
andreasjordan, line 268: Maybe first assign $sql in one line and then use it in the second?

potatoqualitee:
note: updated branch with new fixed test

functions/Copy-DbaDbMail.ps1
tkwj: I have pushed a new commit to resolve the formatting issues. Thanks, I learned something new today about running Invoke-DbatoolsFormatter and dbatools.Tests.ps1 :)

PR #8233, opened 2022-03-13
New-DbaAgentJob: use Get command to return consistent results

### LINKED ISSUE #8232: New-DbaAgentJob: Inconsistent result name with Set-DbaAgentJob

Yes

New-DbaAgentJob returns the raw SMO object:

https://github.com/dataplat/dbatools/blob/7ba983930a8d955ac373afdfaea0b8fdc71a03a9/functions/New-DbaAgentJob.ps1#L412-L413

which includes `IsEnabled` as a property.

But Set/Get-DbaAgentJob uses a formatted version:

https://github.com/dataplat/dbatools/blob/7ba983930a8d955ac373afdfaea0b8fdc71a03a9/functions/Get-DbaAgentJob.ps1#L157

That is `Enabled` (and of course also accessible as `IsEnabled`)

A small difference, but New- should probably just call Get- at the end to make the default results appear consistent. 

### Steps to Reproduce

```powershell
$y = New-DbaAgentJob -SqlInstance localhost -Job "LowlyDbaBugReport"
$x = Get-DbaAgentJob -SqlInstance localhost -Job "LowlyDbaBugReport"

Compare-Object $y.IsEnabled $x.IsEnabled # passes
Compare-Object $y.Enabled $x.Enabled # fails
```

Yes

### Other details or mentions

_No response_

PowerShell Core (pwsh.exe)

### PowerShell Host Version

N/A

N/A

### .NET Framework Version

N/A

lowlydba:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #8232
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose

Ensure consistent result formatting for New/Get/Set-DbaAgentJob

### Approach

Lean on Get-DbaAgentJob to format the output of the other command(s). 

The command now is the same as Set-DbaAgentJob:
https://github.com/dataplat/dbatools/blob/7ba983930a8d955ac373afdfaea0b8fdc71a03a9/functions/Set-DbaAgentJob.ps1#L513

### Commands to test

* New-DbaAgentJob

### DIFF
--- a/functions/New-DbaAgentJob.ps1
+++ b/functions/New-DbaAgentJob.ps1
@@ -410,7 +410,7 @@ function New-DbaAgentJob {
             Add-TeppCacheItem -SqlInstance $server -Type job -Name $Job
 
             # Return the job
-            $currentjob
+            Get-DbaAgentJob -SqlInstance $server | Where-Object Name -eq $currentjob.name
         }
     }
 
--- a/functions/New-DbaAgentJob.ps1
+++ b/functions/New-DbaAgentJob.ps1
@@ -410,7 +410,7 @@ function New-DbaAgentJob {
             Add-TeppCacheItem -SqlInstance $server -Type job -Name $Job
 
             # Return the job
-            Get-DbaAgentJob -SqlInstance $server | Where-Object Name -eq $currentjob.name
+            Get-DbaAgentJob -SqlInstance $server | Where-Object Name -eq $Job
         }
     }
 
--- a/functions/New-DbaAgentJob.ps1
+++ b/functions/New-DbaAgentJob.ps1
@@ -409,7 +409,6 @@ function New-DbaAgentJob {
 
             Add-TeppCacheItem -SqlInstance $server -Type job -Name $Job
 
-            # Return the job
             Get-DbaAgentJob -SqlInstance $instance | Where-Object Name -eq $Job
         }
     }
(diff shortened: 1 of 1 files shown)


lowlydba:
Removed the comment to trigger the more narrow build since the unrelated Encryption commands were failing - can put it back if desired.

andreasjordan:
You can always do an empty commit. But I think we don't need this comment.

andreasjordan (approved):
Good idea. Change looks good to me.

potatoqualitee:
you both are awesome, thank you! 🙇🏼

functions/New-DbaAgentJob.ps1
```diff
@@ -409,8 +409,7 @@ function New-DbaAgentJob {
 
             Add-TeppCacheItem -SqlInstance $server -Type job -Name $Job
 
-            # Return the job
-            $currentjob
+            Get-DbaAgentJob -SqlInstance $instance | Where-Object Name -eq $Job
```
wsmelton, line 412: This is going to cause a severe performance impact because we are pulling every agent job on an instance before we filter.

It should pass the name to the job parameter so we filter the results pulling from the server.

andreasjordan: Good point - just overlooked that. Will open a PR for that.

PR #8234, opened 2022-03-14
Start-DbaDbEncryption - try to find root cause for random test failure

andreasjordan:
## Type of Change

 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### DIFF
--- a/tests/Start-DbaDbEncryption.Tests.ps1
+++ b/tests/Start-DbaDbEncryption.Tests.ps1
@@ -36,9 +36,9 @@ Describe "$CommandName Integration Tests" -Tags "IntegrationTests" {
                 MasterKeySecurePassword = $passwd
                 BackupSecurePassword    = $passwd
                 BackupPath              = "C:\temp"
-                EnableException         = $true
             }
-            $results = Start-DbaDbEncryption @splat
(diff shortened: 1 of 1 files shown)


andreasjordan:
With Test-Bound:
image

Without Test-Bound:
image

andreasjordan:
About the root cause: It looks like encrypting a database takes a bit less than a second. So there are often two databases that are encrypted in the same second. As for every database the pvk file is created with the timestamp - we have a problem.

I see two ways:
* Waiting a second after writing the pvk file
* Testing for existance and silently ignoring the .export()

@potatoqualitee - What do you think?

Will implement the sleep as it is less code for now...

andreasjordan:
I'm about to change the code - but it's not that easy. In Backup-DbaDbCertificate we have (sometimes) two files: cer and pvk. What if we only have one? 

Waiting a second is ok, because the seconds are used in the filename - so a second later we definitly have a new filename and no problem.

So maybe you can merge this as a quick fix so that the tests pass in the future and then review the code and find a better solution?

potatoqualitee:
ahh that makes sense. i'll merge so tests pass then i think we can do whatever file computes the name, is the one that performs the wait

PR #8247, opened 2022-03-17
Add logic for SSID permission

### LINKED ISSUE #8240: Set-DbaNetworkCertificate - Assign permissions to Service SID
### Summarize Functionality

Currently permissions are being granted to the service account however a better approach i feel is to grant to the Sql Service SID as documented in technical details section.

Granting to the service SID will prevent loss of permissions should the service account change.

No

### Technical Details

https://docs.microsoft.com/en-us/sql/relational-databases/security/using-service-sids-to-grant-permissions-to-services-in-sql-server?view=sql-server-ver15

https://www.sqlskills.com/blogs/bobb/about-sql-servers-usage-of-service-sids/

ACALVETT:
## Type of Change

 - [X] New feature (non-breaking change, adds functionality, fixes #8240  )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
Adds the service SSID to the certificates permissions when calling Set-DbaNetworkPermission

### DIFF
--- a/functions/Set-DbaNetworkCertificate.ps1
+++ b/functions/Set-DbaNetworkCertificate.ps1
@@ -168,7 +168,7 @@ function Set-DbaNetworkCertificate {
 
                 $permission = $serviceAccount, "Read", "Allow"
                 $accessRule = New-Object -TypeName System.Security.AccessControl.FileSystemAccessRule -ArgumentList $permission
-                $permission += $sqlSSID, "Read", "Allow"
+                $permission = $sqlSSID, "Read", "Allow"
                 $accessRuleSSID = New-Object -TypeName System.Security.AccessControl.FileSystemAccessRule -ArgumentList $permission
 
                 if ($null -ne $cert.PrivateKey) {
--- a/functions/Set-DbaNetworkCertificate.ps1
+++ b/functions/Set-DbaNetworkCertificate.ps1
@@ -160,8 +160,16 @@ function Set-DbaNetworkCertificate {
                     return
                 }
 
+                #Grant permissions to the Service SID
+                $sqlSSID = "NT SERVICE\MSSQLSERVER"
+                if ($instanceName -ne "MSSQLSERVER") {
+                    $sqlSSID = "NT SERVICE\MSSQL$" + $instanceName
+                }
+
                 $permission = $serviceAccount, "Read", "Allow"
                 $accessRule = New-Object -TypeName System.Security.AccessControl.FileSystemAccessRule -ArgumentList $permission
+                $permission += $sqlSSID, "Read", "Allow"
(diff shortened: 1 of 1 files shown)


potatoqualitee:
this looks reasonable to me @andreasjordan how about you?

potatoqualitee:
if andreas doesnt have time, @ACALVETT can you please submit screenshots? also sorry for the delay! it's one of those months

ACALVETT:
No problem, will get screenshots

ACALVETT:
image
Screenshot showing perms granted to SSID

andreasjordan:
I have seen this but not enought time to have a deeper look. I would like to fully understand that thing. Will try to invest some time next week.

niphlod:
jm2c: the logic is correct (that's what I'm used to do "manually"), I cannot vouch for the correctness of the implementation. 

recap: you're running SQL with a specific service user. If that user has not at least read access to the private key, the SQL Service cannot start. Instead of figuring out what service user needs the specific permission, this uses the virtual account to add the ACL once. This means that the day you set your sql server to run with ANOTHER user, you don't need to remember to add the new user to the ACL, nor need to remember to remove the old one from the ACL, and it all "just works".
In many ways this should be already handled by SMO's capabilities of setting the service account (the same one used in the SQL Server configuration mmc) but, unfortunately, it does not.
The same deal is achieved (if you care to look at it) to the data, log and backup folders: sql needs permissions to them, and the setup slams NT SERVICE\MSSQLSERVER at setup time instead of using the service account configured to run the service. I guess they did it for the exact same purpose of this PR, which is "set and forget", instead of adding the new one and removing the old one.

functions/Set-DbaNetworkCertificate.ps1
```diff
@@ -160,8 +160,16 @@ function Set-DbaNetworkCertificate {
...
 
+                #Grant permissions to the Service SID
+                $sqlSSID = "NT SERVICE\MSSQLSERVER"
+                if ($instanceName -ne "MSSQLSERVER") {
+                    $sqlSSID = "NT SERVICE\MSSQL$" + $instanceName
+                }
+
                 $permission = $serviceAccount, "Read", "Allow"
                 $accessRule = New-Object -TypeName System.Security.AccessControl.FileSystemAccessRule -ArgumentList $permission
+                $permission += $sqlSSID, "Read", "Allow"
```
andreasjordan, line 171: I think this must be "=" instead if "+=".

potatoqualitee:
(updating branch to rerun tests that failed due to  short github outage)

potatoqualitee:
Lookin good 💯  Thank you all!

PR #8330, opened 2022-04-30
Get-DbaBackupInformation - Fix size information for striped backups

### LINKED ISSUE #8302: Get-DbaBackupInformation - Sizes are incorrectly summed

Yes

Incorrect size results from the command. For a striped full backup, it looks like it is summing the compressed/backup size for all files when each file contains the proper value for the group:

https://github.com/dataplat/dbatools/blob/05b91475db7db2b25a49bc1b4b54e3af9b7a45fc/functions/Get-DbaBackupInformation.ps1#L326-L327

So you end up with size values that are multiplied by the number of striped files.

### Steps to Reproduce

```powershell
$stripedPath = 'your-path-to-all-full-striped-backups'
$singleStripedFilePath = 'your-path-to-a-striped-full-backup-file'
$result = Get-DbaBackupInformation -Path $stripedPath
$result2 = Get-DbaBackupInformation -Path $singleStripedFilePath

# Returns values that are size * num striped files
$result.TotalSize 
$result.CompressedBackupSize

# returns correct values for the entire striped backup
$result2.TotalSize
$result2.CompressedBackupSize
```

_No response_

### Other details or mentions

_No response_

### PowerShell Host Version

N/A

N/A

### .NET Framework Version

N/A

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #8302 )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

Every file of a striped backup contains the size information about the backup itself, not the individual file. So the size can just be fetched from the first backup file like most other information.

Before:

image

After: 
image

### DIFF
--- a/functions/Get-DbaBackupInformation.ps1
+++ b/functions/Get-DbaBackupInformation.ps1
@@ -323,8 +323,8 @@ function Get-DbaBackupInformation {
                         Name       = "Size"
                         Expression = { [dbasize]$PSItem.Size }
                     } -Unique)
-                $historyObject.TotalSize = ($group.Group.BackupSize.Byte | Measure-Object -Sum).Sum
-                $HistoryObject.CompressedBackupSize = ($group.Group.CompressedBackupSize.Byte | Measure-Object -Sum).Sum
+                $historyObject.TotalSize = $group.Group[0].BackupSize.Byte
+                $HistoryObject.CompressedBackupSize = $group.Group[0].CompressedBackupSize.Byte
                 $historyObject.Type = $description
                 $historyObject.BackupSetId = $group.group[0].BackupSetGUID
                 $historyObject.DeviceType = 'Disk'


andreasjordan:
The output of `Get-DbaBackupInformation` does not list information about individual files. A striped backup only has one object returned. That is done via the `Group-Object`.

Some pictures...

I have taken two backups of Adventureworks:
image

The filesystem says:
image

The command says:
image

So in case of the striped backup, both files of around 100 kByte are part of same backup and their headers include the information "Backupsize = 200 kByte". The size of the file is not found - and not needed, as we don't return information about files.

So we only have - and only need - one "Size" information, currently named "TotalSize" (and one for the compressed value named "CompressedBackupSize").

So I still thimk my PR fixes a bug.

lowlydba (approved):

potatoqualitee:
@ClaudioESSilva and i are still working through this as I still need convincing there should not be a totalsize AND size and im doing a sanity check 😊

PR #8356, opened 2022-05-16
Copy-DbaDatabase - Detach/Attach Fix

potatoqualitee:
This fixes a detach/attach fix but as I was testing, I noticed that I had duplicate output. So I fixed that too. 

My original fix failed on Linux to Linux migrations (Thanks GitHub Actions!) so this PR ultimately works on Linux to Linux, , Windows to Windows and Windows to Linux.


potatoqualitee:
Merging due to all tests passing. Post-reviews / PRs welcome.

PR #8361, opened 2022-05-23
Import-DbaCsv - Fix ColumnMapper bug

potatoqualitee:
Not sure what the trigger is but I could import like 5 csv files, but not 8. The error suggested it was a column mapping error and sure enough, seems sometimes it would reuse the column mapper. Now it all works.

```
Get-ChildItem C:\temp\csv | Import-DbaCsv -SqlInstance sql01 -Database tempdb -AutoCreateTable -Verbose
```


potatoqualitee:
sweet, figured it out and tests pass 👍🏼

PR #8390, opened 2022-06-03
Invoke-DbaDbDataGenerator - Test for MaskingType Random to support special masking types

andreasjordan:
## Type of Change

 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
I wanted to use the special masking types like zipcode, color, etc.

### Approach
Now we test for MaskingType as well and only if this is "Random" we use `Get-DbaRandomizedValue -DataType $columnMaskInfo.ColumnType`, in all other cases we now use `Get-DbaRandomizedValue -RandomizerType $columnMaskInfo.MaskingType`.

### Commands to test
```
$sqlInstance = 'sql01'
$databaseName = 'randomdata'
$tempPath = 'C:\temp'

$db = New-DbaDatabase -SqlInstance $sqlInstance -Name $databaseName
$db.ExecuteNonQuery("CREATE TABLE dbo.randomdata(id int identity primary key, justastring varchar(10), fullname varchar(100), zipcode varchar(10), city varchar(100))")
$config = New-DbaDbDataGeneratorConfig -SqlInstance $sqlInstance -Database $databaseName -Table 'randomdata' -Path $tempPath -Rows 5

Invoke-DbaDbDataGenerator -SqlInstance $sqlInstance -Database $databaseName -FilePath $config -Confirm:$false
$db.Query("SELECT * FROM dbo.randomdata") | Format-Table

Remove-Item -Path $config
$null = Remove-DbaDatabase -SqlInstance $sqlInstance -Database $databaseName -Confirm:$false
```

Old:
image

New:
image

### DIFF
--- a/functions/Invoke-DbaDbDataGenerator.ps1
+++ b/functions/Invoke-DbaDbDataGenerator.ps1
@@ -109,14 +109,14 @@ function Invoke-DbaDbDataGenerator {
 
         # Create the faker objects
         try {
-            $faker = New-Object Bogus.Faker($Locale)
+            $script:faker = New-Object Bogus.Faker($Locale)
         } catch {
             Stop-Function -Message "Could not load randomizer class" -Continue
         }
 
         $supportedDataTypes = 'bigint', 'bit', 'bool', 'char', 'date', 'datetime', 'datetime2', 'decimal', 'int', 'float', 'guid', 'money', 'numeric', 'nchar', 'ntext', 'nvarchar', 'real', 'smalldatetime', 'smallint', 'text', 'time', 'tinyint', 'uniqueidentifier', 'userdefineddatatype', 'varchar'
-        $supportedFakerMaskingTypes = ($faker | Get-Member -MemberType Property | Select-Object Name -ExpandProperty Name)
-        $supportedFakerSubTypes = ($faker | Get-Member -MemberType Property) | ForEach-Object { ($faker.$($_.Name)) | Get-Member -MemberType Method | Where-Object { $_.Name -notlike 'To*' -and $_.Name -notlike 'Get*' -and $_.Name -notlike 'Trim*' -and $_.Name -notin 'Add', 'Equals', 'CompareTo', 'Clone', 'Contains', 'CopyTo', 'EndsWith', 'IndexOf', 'IndexOfAny', 'Insert', 'IsNormalized', 'LastIndexOf', 'LastIndexOfAny', 'Normalize', 'PadLeft', 'PadRight', 'Remove', 'Replace', 'Split', 'StartsWith', 'Substring', 'Letter', 'Lines', 'Paragraph', 'Paragraphs', 'Sentence', 'Sentences' } | Select-Object name -ExpandProperty Name }
+        $supportedFakerMaskingTypes = ($script:faker | Get-Member -MemberType Property | Select-Object Name -ExpandProperty Name)
+        $supportedFakerSubTypes = ($script:faker | Get-Member -MemberType Property) | ForEach-Object { ($script:faker.$($_.Name)) | Get-Member -MemberType Method | Where-Object { $_.Name -notlike 'To*' -and $_.Name -notlike 'Get*' -and $_.Name -notlike 'Trim*' -and $_.Name -notin 'Add', 'Equals', 'CompareTo', 'Clone', 'Contains', 'CopyTo', 'EndsWith', 'IndexOf', 'IndexOfAny', 'Insert', 'IsNormalized', 'LastIndexOf', 'LastIndexOfAny', 'Normalize', 'PadLeft', 'PadRight', 'Remove', 'Replace', 'Split', 'StartsWith', 'Substring', 'Letter', 'Lines', 'Paragraph', 'Paragraphs', 'Sentence', 'Sentences' } | Select-Object name -ExpandProperty Name }
         $supportedFakerSubTypes += "Date"
         #$foreignKeyQuery = Get-Content -Path "$script:PSModuleRoot\bin\datageneration\ForeignKeyHierarchy.sql"
     }
@@ -256,7 +256,7 @@ function Invoke-DbaDbDataGenerator {
                                             }
 
                                         } catch {
-                                            Stop-Function -Message "Failure" -Target $faker -Continue -ErrorRecord $_
+                                            Stop-Function -Message "Failure" -Target $script:faker -Continue -ErrorRecord $_
                                         }
 
                                         # Check if the value is already present as a property
@@ -395,7 +395,7 @@ function Invoke-DbaDbDataGenerator {
                                         }
 
                                     } catch {
-                                        Stop-Function -Message "Failure" -Target $faker -Continue -ErrorRecord $_
+                                        Stop-Function -Message "Failure" -Target $script:faker -Continue -ErrorRecord $_
                                     }
 
                                 }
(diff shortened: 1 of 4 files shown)


andreasjordan:
There is an issue because not all the special types are supported. 

To test that, change the table structure to use "bic" as a column name:
```
$db.ExecuteNonQuery("CREATE TABLE dbo.randomdata(id int identity primary key, bic varchar(10), fullname varchar(100), zipcode varchar(10), city varchar(100))")
```

I will provide a temporary fix in the next commit.

andreasjordan:
@sanderstad Do you have time to have a look? My current idea is to add the target MaskingType and SubType to the columntypes.json so that we don't need that code anymore (but get the information from the json):

https://github.com/dataplat/dbatools/blob/ea5ae5b8f0fbc386578cd4f241edc5059efa0468/functions/New-DbaDbDataGeneratorConfig.ps1#L205-L237

andreasjordan:
The file "columntypes.json" is in folder "datamasking" so I have to ensure that the masking commands still work after changing the file...

andreasjordan:
Ok, first step is taken, looks good.

andreasjordan:
If we use the same `Set-Content` no matter if the file already exists, we don't need the if. But we need the ShouldProcess in any case. So I changed that part of the code as well.

andreasjordan:
Running Invoke-DbaDbDataGenerator with different Locale does not change the used locale because Get-DbaRandomizedValue uses $script:faker. So I changed Invoke-DbaDbDataGenerator to use $script:faker as well.

andreasjordan:
I added the support for two part table names here, because I need it in here for testing all changes together - but I can split this change to seperate PR "New-DbaDbDataGeneratorConfig - Support two part table names".

andreasjordan:
@sanderstad : Do you remeber the reason for itentity columns starting at 2 (instead of 1)?:
```
                    if ($columnobject.InPrimaryKey -and $columnobject.DataType.SqlDataType.ToString().ToLowerInvariant() -notmatch 'date') {
                        $min = 2
                    }
```

andreasjordan:
Testing for MaskingType -eq 'Random' is not perfect, as we now still don't have a chance to use "Ramdom"/"Guid". Will have to think about a different solution...

potatoqualitee:
will merge when sander or someone approves 😊

functions/New-DbaDbDataGeneratorConfig.ps1
```diff
@@ -136,7 +136,7 @@ function New-DbaDbDataGeneratorConfig {
 
             # Get the tables
             if ($Table) {
-                $tablecollection = $db.Tables | Where-Object Name -in $Table
+                $tablecollection = $db | Get-DbaDbTable -Table $Table
```
sanderstad, line 139: I'm not sure if this makes the table collection any faster. The collection is already there, why would you run another command that would do the same thing?

sanderstad:
> @sanderstad : Do you remember the reason for itentity columns starting at 2 (instead of 1)?:
> 
> ```
>                     if ($columnobject.InPrimaryKey -and $columnobject.DataType.SqlDataType.ToString().ToLowerInvariant() -notmatch 'date') {
>                         $min = 2
>                     }
> ```

I have no idea why that value is 2, mostly because I wrote this over a couple of years ago. I guess there was a reason for it otherwise I wouldn't have done it.....

> @sanderstad : Do you remember the reason for itentity columns starting at 2 (instead of 1)?:
> 
> ```
>                     if ($columnobject.InPrimaryKey -and $columnobject.DataType.SqlDataType.ToString().ToLowerInvariant() -notmatch 'date') {
>                         $min = 2
>                     }
> ```

I'm not 100% sure why I set this value to 2. Maybe it was something during my testing that I ran into problems when a primary would have the next value to be set to 1 and this caused issues with generating data.
We could try to make it `1` and check out if that causes any issues

functions/Invoke-DbaDbDataGenerator.ps1
```diff
@@ -300,7 +300,7 @@ function Invoke-DbaDbDataGenerator {
                         Write-ProgressHelper -StepNumber ($stepcounter++) -TotalSteps $tables.Tables.Count -Activity "Generating data" -Message "Inserting $($tableobject.Rows) rows in $($tableobject.Schema).$($tableobject.Name) in $($db.Name) on $instance"
 
                         if ($tableobject.TruncateTable) {
-                            $query += "TRUNCATE TABLE [$($tableobject.Schema)].[$($tableobject.Name)];`n"
+                            $query = "TRUNCATE TABLE [$($tableobject.Schema)].[$($tableobject.Name)]"
```
sanderstad, line 303: I agree with remplacing the `+=` with just `=` but it's tidier code if the query end with a `;` to make sure it's compliant for future code when that is mandatory.

andreasjordan: Ok, agreed. Will change that.

andreasjordan:
> > @sanderstad : Do you remember the reason for itentity columns starting at 2 (instead of 1)?:
> > ```
> >                     if ($columnobject.InPrimaryKey -and $columnobject.DataType.SqlDataType.ToString().ToLowerInvariant() -notmatch 'date') {
> >                         $min = 2
> >                     }
> > ```
> 
> I'm not 100% sure why I set this value to 2. Maybe it was something during my testing that I ran into problems when a primary would have the next value to be set to 1 and this caused issues with generating data. We could try to make it `1` and check out if that causes any issues

The purpose of this code is to have only positive values for numeric primary keys (int and bigint) - correct? Because if we don't set `$min`, it is always `$null` - which will be set to the smallest possible value by `Get-DbaRandomizedValue` (-2147483648 and -9223372036854775808).

So in case we use random values, there is not much difference between having min 1 or 2, but I would add a comment to answer future questions.

I used an identity column as primary key and the first value was 2 - but that was not the 2 here in the code - but a problem with getting the correct next identity value for an empty table - which I fixed including a comment.

So whatever we do with this `$min = 2` - I would leave it up to you to decide. I will leave the code as is for now.

functions/New-DbaDbDataGeneratorConfig.ps1
sanderstad: Good shout. I agree

andreasjordan:
Just found a new issue: `$supportedDataTypes` contains "text" and "ntext", but `Get-DbaRandomizedValue` has no code for that. So a nullable text column will be empty, a non-nullable column will throw an error.

@sanderstad - Do you know why text and ntext are in the supported datatypes?

sanderstad (approved):
Good changes. Thank you for taking the time

andreasjordan:
I'm still not sure what's the best way do decide if to use `$columnValue = Get-DbaRandomizedValue -DataType $columnobject.ColumnType` or `$columnValue = Get-DbaRandomizedValue -RandomizerType $columnobject.MaskingType -RandomizerSubtype $columnobject.SubType`. 

What should be the default and in what cases should the other one be used?

andreasjordan:
One more thing: `Get-DbaRandomizedValue` uses `$Precision` for 'decimal', 'float', 'money', 'numeric', 'real'. But `Invoke-DbaDbDataGenerator` does not set this parameter so the default of 2 is used. We should use the precision of the datatype, don't we?

andreasjordan:
There might be a risk that it breaks some special use case out there in the wild. And there are more issues I would like to look at. But I think this is a very good start which enables the use of `$columnValue = Get-DbaRandomizedValue -RandomizerType $columnobject.MaskingType -RandomizerSubtype $columnobject.SubType`. So I'm ok with mergeing and will not provide more commits to this.

potatoqualitee:
Thank you both 💯  We'll keep an eye out for any reports of edge-case impacts.

PR #8393, opened 2022-06-07
Csv progress

MarkPflug:
## Type of Change

 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose

Adds percent complete progress report to Import-DbaCsv.

### Approach

Uses the file size to determine percent complete by counting bytes read by the underlying data stream.

### Commands to test

Import-DbaCsv

### Screenshots

progress

### Learning

This change was inspired by the @potatoqualitee blog about CSV performance. I've used this technique before in other projects to report import progress, and thought it might be a nice enhancement for dbatools.

:warning: This PR is incomplete! I does not include the a newly compiled dll at `bin/netcoreapp3.1/dbatools.dll`, which carries the ProgressStream implementation. It isn't clear to me if there is an official build step to make this happen, as I'd manually copied it there to get things working.

### DIFF
--- a/bin/projects/dbatools/dbatools/IO/ProgressStream.cs
+++ b/bin/projects/dbatools/dbatools/IO/ProgressStream.cs
@@ -1,0 +1,89 @@
+using System;
+using System.IO;
+
+namespace Sqlcollaborative.Dbatools.IO
+{
+    /// <summary>
+    /// Provides progress callbacks for a stream being read.
+    /// </summary>
+    public class ProgressStream : Stream
+    {
+        readonly Stream inner;
+        readonly Action<double> callback;
+        readonly long progressSize;
+        long length;
+        long progressAccumulator;
+
+        public ProgressStream(Stream inner, Action<double> callback, double factor)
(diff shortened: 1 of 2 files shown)


potatoqualitee:
Thank you so very much! I'm going to compile this and also see if I can add in an auto-compile into my releases script.

MarkPflug:
@potatoqualitee Just to state my reason for not including the dll. Checking the binary into the repository feels like a security issue. You don't know me, and have no reason to trust me. Verifying that a dll contains only the changes that I included in source would be difficult. Nothing would prevent me from checking in a "modified" dll that contains additional malicious code. That was my thinking anyway. Moving the compilation to CI/CD is exactly what I was going to suggest.

Thanks for the great tools!

potatoqualitee:
Agreed! Since you aren't know to me, they're binary files and I can't see inside them, I would have deleted and recreated anyway 👍🏼  So your PR was perfect.

PR #8403, opened 2022-06-18
Set-DbaTcpPort - add force switch

### LINKED ISSUE #8402: Set-DbaTcpPort: Add -Force for service restarts
### Summarize Functionality

Enable-DbaAgHadr is a similar setting that requires a service restart to go into effect, and uses `-Force` to do the service restart. Set-DbaTcpPort should also include `-Force` for restarting the service.

Yes

### Technical Details

Enable-DbaAgHadr / Set-DbaTcPort

lowlydba:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality, fixes #8402  )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
Adds the `-Force` switch to allow service restarts as part of the command.

### Approach
Copy & pasted from `Enable-DbaAgHadr`'s implementation

### Commands to test
* Set-DbaTcpPort

### DIFF
--- a/functions/Set-DbaTcpPort.ps1
+++ b/functions/Set-DbaTcpPort.ps1
@@ -129,7 +129,7 @@ function Set-DbaTcpPort {
                             $null = Stop-DbaService -ComputerName $computerFullName -InstanceName $instanceName -Type Agent, Engine
                             $null = Start-DbaService -ComputerName $computerFullName -InstanceName $instanceName -Type Agent, Engine
                         } catch {
-                            Stop-Function -Message "Issue restarting $instance" -Target $instance -Continue
+                            Stop-Function -Message "Issue restarting $instance on $computerFullName" -Target $instance -Continue -ErrorRecord $_
                         }
                     }
                 }
--- a/functions/Set-DbaTcpPort.ps1
+++ b/functions/Set-DbaTcpPort.ps1
@@ -126,8 +126,7 @@ function Set-DbaTcpPort {
                 if (Test-Bound -ParameterName Force) {
                     if ($PSCmdlet.ShouldProcess($instance, "Force provided, restarting Engine and Agent service for $instance on $computerFullName")) {
                         try {
-                            $null = Stop-DbaService -ComputerName $computerFullName -InstanceName $instanceName -Type Agent, Engine
-                            $null = Start-DbaService -ComputerName $computerFullName -InstanceName $instanceName -Type Agent, Engine
+                            $null = Restart-DbaService -ComputerName $computerFullName -InstanceName $instanceName -Type Agent, Engine
                         } catch {
                             Stop-Function -Message "Issue restarting $instance on $computerFullName" -Target $instance -Continue -ErrorRecord $_
                         }
--- a/functions/Set-DbaTcpPort.ps1
+++ b/functions/Set-DbaTcpPort.ps1
@@ -34,6 +34,9 @@ function Set-DbaTcpPort {
     .PARAMETER Confirm
         Prompts you for confirmation before executing any changing operations within the command.
 
+    .PARAMETER Force
+        Will restart SQL Server and SQL Server Agent service to apply the change.
+
     .NOTES
         Tags: Network, Connection, TCP, Configure
         Author: @H0s0n77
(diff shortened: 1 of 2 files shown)


functions/Set-DbaTcpPort.ps1
```diff
@@ -116,7 +122,18 @@ function Set-DbaTcpPort {
...
                     $netConf | Set-DbaNetworkConfiguration -Credential $Credential -EnableException:$EnableException -Confirm:$false
                 }
+
+                if (Test-Bound -ParameterName Force) {
+                    if ($PSCmdlet.ShouldProcess($instance, "Force provided, restarting Engine and Agent service for $instance on $computerFullName")) {
+                        try {
+                            $null = Stop-DbaService -ComputerName $computerFullName -InstanceName $instanceName -Type Agent, Engine
+                            $null = Start-DbaService -ComputerName $computerFullName -InstanceName $instanceName -Type Agent, Engine
+                        } catch {
+                            Stop-Function -Message "Issue restarting $instance" -Target $instance -Continue
```
potatoqualitee, line 132:
```suggestion
                            Stop-Function -Message "Issue restarting $instance" -Target $instance -Continue -ErrorRecord $_
```

potatoqualitee: Thank you! Will commit my own suggestion 😅

functions/Set-DbaTcpPort.ps1
```diff
@@ -116,7 +122,18 @@ function Set-DbaTcpPort {
                 if ($Pscmdlet.ShouldProcess($instance, "Setting port to $Port for IP address $IpAddress of $instance")) {
                     $netConf | Set-DbaNetworkConfiguration -Credential $Credential -EnableException:$EnableException -Confirm:$false
                 }
+
+                if (Test-Bound -ParameterName Force) {
+                    if ($PSCmdlet.ShouldProcess($instance, "Force provided, restarting Engine and Agent service for $instance on $computerFullName")) {
+                        try {
+                            $null = Stop-DbaService -ComputerName $computerFullName -InstanceName $instanceName -Type Agent, Engine
+                            $null = Start-DbaService -ComputerName $computerFullName -InstanceName $instanceName -Type Agent, Engine
```
andreasjordan, line 130: Why don't you use Restart-DbaService? What if the agent was not started? I think Restart takes care of this. Otherwise the agent would not be in the same state as before using this command.

lowlydba: I copied it from Enable-DbaAgHadr since it is a similar config that requires a restart to take effect. Happy to use that instead if its preferred.

potatoqualitee:
Thanks for taking a look @andreasjordan! I'll merge once this new release is out

functions/Set-DbaTcpPort.ps1
```diff
@@ -116,7 +122,17 @@ function Set-DbaTcpPort {
                 if ($Pscmdlet.ShouldProcess($instance, "Setting port to $Port for IP address $IpAddress of $instance")) {
                     $netConf | Set-DbaNetworkConfiguration -Credential $Credential -EnableException:$EnableException -Confirm:$false
                 }
+
+                if (Test-Bound -ParameterName Force) {
+                    if ($PSCmdlet.ShouldProcess($instance, "Force provided, restarting Engine and Agent service for $instance on $computerFullName")) {
+                        try {
+                            $null = Restart-DbaService -ComputerName $computerFullName -InstanceName $instanceName -Type Agent, Engine
```
andreasjordan, line 129: Please only restart the Engine. The Agent would be restarted if needed.

You could also use the new parameter -SqlInstance with $instance.

But in order to use the try-catch, you have to use -EnableException.

And thanks for the hint about DbaAgHadr - I will take a look and fix it there and in some more places...

potatoqualitee: Is -Force required to restart the agent without prompting?

andreasjordan: Yes, you are right. So this has to be added as well.

andreasjordan: So it might be one option: `$null = Restart-DbaService -SqlInstance $instance -Type Engine -Force -EnableException`.

lowlydba: Thanks for the tips ya'll! Hard to keep up with how often you improve these patterns :D

andreasjordan (approved):
Thanks for the changes.

potatoqualitee:
Thank you both for your time and effort 😊

Issue #8183, opened 2022-02-19
Add DatabaseID to the return object for relevant commands


lancasteradam:
### Summarize Functionality

As noted in #8053 if the database name has been changed it can cause matching issues when comparing historical data. 

For this feature request the goal is to identify which other commands would benefit from adding the DatabaseID to the return object. For example, the Get-DbaDb* commands.

### Is there a command that is similiar or close to what you are looking for?

No

### Technical Details

See #8175 as an example of adding the DatabaseID to a command's return object.

andreasjordan:
Are there more commands to work on? Or can this be closed?

wsmelton:
I'd say if more need to be worked on open new issue(s).

PR #8407, opened 2022-06-18
Expand-DbaDbLogFile: Add DatabaseID to the result object

lancasteradam:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality, relates to #8183 )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose

Minor change to add Database ID to the result object.

### Approach

Updated the result object with the db ID.

### Commands to test

See the pester tests.

### DIFF
--- a/functions/Expand-DbaDbLogFile.ps1
+++ b/functions/Expand-DbaDbLogFile.ps1
@@ -491,6 +491,7 @@ function Expand-DbaDbLogFile {
                     InstanceName    = $server.ServiceName
                     SqlInstance     = $server.DomainInstanceName
                     Database        = $db
+                    DatabaseID      = $db.ID
                     ID              = $logfile.ID
                     Name            = $logfile.Name
                     LogFileCount    = $numLogfiles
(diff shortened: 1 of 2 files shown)


lancasteradam:
Yes, so far it has been DatabaseID or in cases where there is source/destination it is SourceDatabaseID/DestinationDatabaseID.

#8183 is the main issue for this series of changes.

I like the idea of using DatabaseId/SourceDatabaseId/DestinationDatabaseId because they are self-documenting with the ps notation. If we use just Id the reader may use some cycles to determine what the Id is for (depending on the nature of the command).

There are some commands where Id is used for something other than the database: https://github.com/dataplat/dbatools/pull/8175/files

Let me know your preference @potatoqualitee and I'll be happy to make the updates. Overall this series of changes are minor so it's no problem to make adjustments.

potatoqualitee:
After some consideration, DatabaseId is the right call since there could be other IDs. Thank you, Adam!

PR #8460, opened 2022-07-24
Find-DbaDbDisabledIndex: Add DatabaseId to the return object

lancasteradam:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality, relates to #8183  )
 - [x] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose

Minor change to add DatabaseId to the return object. In this case using 'DatabaseId' instead of 'Id' since the command returns an object specific to indexes.

### Approach

Using sys.databases since [DB_ID() is not guaranteed in Azure](https://docs.microsoft.com/en-us/sql/t-sql/functions/db-id-transact-sql?view=sql-server-ver16#remarks).

### Commands to test

Updated the pester tests.

### DIFF
--- a/functions/Find-DbaDbDisabledIndex.ps1
+++ b/functions/Find-DbaDbDisabledIndex.ps1
@@ -87,6 +87,7 @@ function Find-DbaDbDisabledIndex {
     begin {
         $sql = "
         SELECT DB_NAME() AS 'DatabaseName'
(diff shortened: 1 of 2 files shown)


potatoqualitee:
sweet, thanks!

PR #8471, opened 2022-08-01
Module Import: add dotnet version warning/throw

potatoqualitee:
No idea why the psd1 `DotNetFrameworkVersion = '4.6.2'` doesn't work but this addresses it

### DIFF
--- a/dbatools.psm1
+++ b/dbatools.psm1
@@ -7,6 +7,14 @@ param(
 
 $start = [DateTime]::Now
 
+If ($PSVersionTable.PSEdition -in "Desktop", $null) {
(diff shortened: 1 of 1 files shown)


potatoqualitee:
I tested on windows 2012 r2 with and without .net 4.6 with success 👍🏼

dsolodow:
This causes the $error variable to increment for keys that don't have a Version property. Might be better/safer to use ErrorAction Ignore rather than SilentlyContinue ?

potatoqualitee:
ignore never works as expected for me -- it spits red to the screen, can you try it and let me know if you see red?

dsolodow:
image

potatoqualitee:
Thank you! Tested and added a PR.

PR #8487, opened 2022-08-15
Handjam Cmdlet Info in Find-Command

potatoqualitee:
This will also fix the website and this issue https://github.com/dataplat/docs/issues/67


potatoqualitee:
@niphlod this is nasty but gets the job done 😅  Get-Help shows the appropriate help so I'm not sure what's going on with it. Perhaps the regex for cmdlets vs functions is different.

Issue #8472, opened 2022-08-03
Can't drop temp stored procs


berki7867:
### Verified issue does not already exist?

I have searched and found no existing issue

### What error did you receive?

Hi,

When running the below script that uses DBATools Invoke-DbaQuery to run a T-SQL script against an instance, I get the error unable to drop #tempproc, #tempproc not found:

```
$server = Connect-DbaInstance -SqlInstance $SqlInstance -Database myDB -ConnectTimeout 0 -StatementTimeout 0 

Invoke-DbaQuery -SqlInstance $server -File 'SQLScript.sql' -QueryTimeout 0 -MessagesToOutput -Verbose

$server | Disconnect-DbaInstance
```

If I use Invoke-Sqlcmd it works fine. Any ideas?


### Steps to Reproduce

```powershell
CREATE PROC #DropStatistics
AS
SELECT 1
GO
drop procedure #DropStatistics
go
```


### Please confirm that you are running the most recent version of dbatools

1.1.81

### Other details or mentions

_No response_

### What PowerShell host was used when producing this error

Windows PowerShell (powershell.exe), VS Code (terminal), VS Code (integrated terminal)

### PowerShell Host Version


Name                           Value                                                                                                                                                                                                                                                                                                           
----                           -----                                                                                                                                                                                                                                                                                                           
PSVersion                      5.1.14393.5127                                                                                                                                                                                                                                                                                                  
PSEdition                      Desktop                                                                                                                                                                                                                                                                                                         
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                                                                                                                                                                                                                                                                                         
BuildVersion                   10.0.14393.5127                                                                                                                                                                                                                                                                                                 
CLRVersion                     4.0.30319.42000                                                                                                                                                                                                                                                                                                 
WSManStackVersion              3.0                                                                                                                                                                                                                                                                                                             
PSRemotingProtocolVersion      2.3                                                                                                                                                                                                                                                                                                             
SerializationVersion           1.1.0.1                                                                                                                                                                                                                                                                                                         




### SQL Server Edition and Build number

Microsoft SQL Server 2014 (SP3-CU4-GDR) (KB5014164) - 12.0.6439.10 (X64)   Apr 20 2022 03:13:42   Copyright (c) Microsoft Corporation  Enterprise Edition (64-bit) on Windows NT 6.3 <X64> (Build 14393: ) 

### .NET Framework Version

.NET Framework 4.8.4526.0

[...]

PR #8493, opened 2022-08-18
Invoke-DbaQuery - default to nonpooled connection

### LINKED ISSUE #8491: Request for comment: Defaulting Invoke-DbaQuery to Non-pooled connection

Considering the solution for [Can't drop temp stored procs](https://github.com/dataplat/dbatools/issues/8472) and the fact that `Invoke-Sqlcmd` uses non-pooled by default, I'd like to propose that Invoke-DbaQuery does too.

The alternative is that we parse the SQL for temp objects and force it to Non-pooled the way we detect and force it for Credentialed (not integrated) Windows Auth.

In order to perfectly detect a temp object, we'd have to add 14MB of DLLs.

```
$parseoptions = New-Object Microsoft.SqlServer.Management.SqlParser.Parser.ParseOptions
$parseoptions .BatchSeparator = $BatchSeparator

$sql = "CREATE PROC #DropStatistics
AS
SELECT 1
GO
drop procedure #DropStatistics
go"

$parser = [Microsoft.SqlServer.Management.SqlParser.Parser.Parser]::Parse($sql, $parseoptions )
$parser.Script.Tokens | Where-Object { $PSItem.Text.StartsWith('#') }
```

We can also use the parser within our own parsing (which is currently regex, I believe).

The downside to parsing is that it will slow down the import of dbatools by 130ms and increase the size by 14MB uncompressed.

Yes

### Technical Details

_No response_

potatoqualitee:
Fixes #8491, #8472

### DIFF
--- a/functions/Invoke-DbaQuery.ps1
+++ b/functions/Invoke-DbaQuery.ps1
@@ -389,22 +389,26 @@ function Invoke-DbaQuery {
             # We suppress the verbosity of all other functions in order to be sure the output is consistent with what you get, e.g., executing the same in SSMS
             Write-Message -Level Debug -Message "SqlInstance passed in, will work on: $instance"
             try {
-                $connDbaInstanceParams = @{
-                    SqlInstance   = $instance
-                    SqlCredential = $SqlCredential
-                    Database      = $Database
-                    Verbose       = $false
+                $noConnectionChangeNeeded =                                                                  # we want to bypass Connect-DbaInstance if
+                    ($instance.InputObject.GetType().Name -eq "Server") -and                                 # we have Server SMO object and
+                    (-not $ReadOnly) -and                                                                    # no readonly intent is requested and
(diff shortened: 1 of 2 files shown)


andreasjordan:
Will have a look later today...

potatoqualitee:
Thank you! I had one timeout with pwsh on Windows GH Actions test and had to re-run it. It ran with success but I'd like to keep an eye on this and perhaps hold this patch in Dev for a couple days before releasing.

andreasjordan:
That is my idea of bypassing Connect-DbaInstance if possible as a performance boost.

image

But I can revert, if this is more than you would like to do in this PR.

andreasjordan:
Thanks for fixing the formatting.

potatoqualitee:
I think this is great! I'll merge and we can use it for a few days before pushing it out 🔥

potatoqualitee:
that's a HUGE boost too! 🚤

PR #8510, opened 2022-08-28
Add DatabaseId to the return object for Find-DbaDbGrowthEvent

lancasteradam:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality, relates to #8183 )
 - [x] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose

Minor change to add DatabaseId to the return object.

### Approach

Inline query in the select clause

### Commands to test

See the new pester test

### Learning
I explored the option of adding support for Extended Events to this command, but the challenge that came up was that the dba would need to setup the XE session ahead of time. This limits the value of what this command could do with Extended Events other than provide a means to query a specified XE session, for which there are already dbatools commands available.


potatoqualitee:
Thank you!

Issue #5399, opened 2019-04-22
Get-DbaBackupHistory is not always returning accurate last backup info correctly


kmramsey:
```
Name    : dbatools
Path    : C:\windows\system32\WindowsPowerShell\v1.0\Modules\dbatools\dbatools.psd1
Version : 0.9.803
```
- [Done ] Verified errors are not related to permissions

## Environmental data
```Name    : dbatools
Path    : C:\windows\system32\WindowsPowerShell\v1.0\Modules\dbatools\dbatools.psd1
Version : 0.9.803
```

### SQL Server: 
```Microsoft SQL Server 2014 (SP2-CU2) (KB3188778) - 12.0.5522.0 (X64) 
	Sep 22 2016 01:09:29 
	Copyright (c) Microsoft Corporation
	Developer Edition (64-bit) on Windows NT 6.3 <X64> (Build 9600: ) (Hypervisor)
```

## Steps to Reproduce

```sql to find the correct backup information.

--------------------------------------------------------------------------------- 
--FULL Database Backups for all databases For Previous Week 
--------------------------------------------------------------------------------- 
SELECT  top 1
msdb.dbo.backupset.database_name, 
msdb.dbo.backupset.backup_start_date, 
msdb.dbo.backupset.backup_finish_date, 
msdb.dbo.backupmediafamily.physical_device_name
FROM msdb.dbo.backupmediafamily 
INNER JOIN msdb.dbo.backupset ON msdb.dbo.backupmediafamily.media_set_id = msdb.dbo.backupset.media_set_id 
WHERE (CONVERT(datetime, msdb.dbo.backupset.backup_start_date, 102) >= GETDATE() - 7) 
and database_name = 'MYDBNAME'
and msdb.dbo.backupset.name = 'MYDBNAME - Full Backup'
ORDER BY 
msdb.dbo.backupset.database_name, 
msdb.dbo.backupset.backup_finish_date desc


--------------------------------------------------------------------------------- 
--Diff Database Backups for all databases For Previous Week 
--------------------------------------------------------------------------------- 
SELECT top 1
msdb.dbo.backupset.database_name, 
msdb.dbo.backupset.backup_start_date, 
msdb.dbo.backupset.backup_finish_date, 
msdb.dbo.backupmediafamily.physical_device_name
FROM msdb.dbo.backupmediafamily 
[...]

PR #8620, opened 2022-10-28
Get-DbaDbBackupHistory - Add parameter LsnSort

andreasjordan:
Just a draft at the moment. 

As this is a very complex command more people should have a look at this.

It might help to solve #5399 - but maybe not.


potatoqualitee:
ahh thank you, ,didnt realize it'd be complex. godspeed 😅

potatoqualitee:
I like the approach, @niphlod or @jpomfret what do you think?

jpomfret:
I think it makes sense to me, presuming that `$LsnSort = "LastLsn"` would replicate the current behaviour of the command.

andreasjordan:
Yes, the default is the current behaviour - so no breaking change.

Only in combination with `Get-DbaAgBackupHistory` things change a bit, as there is a default of "FirstLsn" which will then be handed to `Get-DbaDbBackupHistory`.

potatoqualitee:
hey @andreasjordan -- i think this gets an overall approval. Can you fix the conflict and set it ready to PR? I'm about to do some massive changes and need all PRs to be merged.

potatoqualitee:
ill resolve this conflict, brb

potatoqualitee:
awww yeahhh

PR #8621, opened 2022-10-30
Allow Get-DbaDbBackupHistory to take a TimeSpan for -Since

### LINKED ISSUE #8599: Allow Get-DbaDbBackupHistory to take a TimeSpan for -Since
### Summarize Functionality

I often find myself wanting to do something like "show me log backups in the last three hours" or "show me full and differential backups over the last eight days". Today, I accomplish this by doing something like `Get-DbaDbBackupHistory -Since ((get-date).AddDays(-8)) ...`. Which is fine.

But! What if you could do `$ts =[timespan]'-8.00:00:00'; Get-DbaDbBackupHistory -Since $ts ...`? It feels like a minor improvement (and maybe it is!), but imagine being able to put a timespan in a parameter splat for the Since parameter. The current functionality doesn't allow for a dynamic point in time to be specified in a splat.

No

### Technical Details

Here's a simple example of how I'd do that overloading

```powershell
function Since-Overload {
    param(
        [object]$Since
    )

    if ($Since -is [timespan]) {
        $Since = (get-date).add($since);
    }
    elseif ($Since -is [datetime]) {
        # do nothing - it's already a datetime
    }
    else {
        throw 'Wrong DT'
    }

    return $Since;
}
```

That is, right now the Since param in Get-DbaDbHistory is strongly typed as a datetime. I'd propose having it be typed as [object] and put logic similar to the above in the begin{} block in the function to allow it to be either a datetime or a timespan and at the end of the logic the $Since variable would be a datetime (allowing the rest of the cmdlet to remain unchanged).

lancasteradam:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality, fixes #8599 )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose

Minor change to support TimeSpan values for the -Since param as @ben-thul suggested in #8599 

### Approach

Update to param type and conditional logic to ensure $Since is set.

### Commands to test

See the new pester tests added.

### DIFF
--- a/functions/Get-DbaDbBackupHistory.ps1
+++ b/functions/Get-DbaDbBackupHistory.ps1
@@ -158,7 +158,7 @@ function Get-DbaDbBackupHistory {
         [switch]$IncludeCopyOnly,
         [Parameter(ParameterSetName = "NoLast")]
         [switch]$Force,
-        [DateTime]$Since = (Get-Date '01/01/1970'),
+        [object]$Since = (Get-Date '01/01/1970'),
         [ValidateScript( { ($_ -match '^(\{){0,1}[0-9a-fA-F]{8}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{4}\-[0-9a-fA-F]{12}(\}){0,1}$') -or ('' -eq $_) })]
         [string]$RecoveryFork,
         [switch]$Last,
@@ -211,12 +211,18 @@ function Get-DbaDbBackupHistory {
         foreach ($typeFilter in $Type) {
             $backupTypeFilter += $backupTypeMapping[$typeFilter]
         }
-
     }
 
     process {
         if ($AgCheck) {
             Stop-Function -Message "Parameter AGCheck is deprecated. This command does not check for history from replicas even if this paramater is not provided. The functionality to also get the history from all replicas if SqlInstance is part on an availability group has been moved to Get-DbaAgBackupHistory."
+            return
+        }
+
+        if ($Since -is [timespan]) {
+            $Since = (Get-Date).add($Since);
+        } elseif (-not ($Since -is [datetime])) {
+            Stop-Function -Message "-Since must be either a DateTime or TimeSpan object."
             return
         }
 
--- a/functions/Get-DbaDbBackupHistory.ps1
+++ b/functions/Get-DbaDbBackupHistory.ps1
@@ -32,6 +32,7 @@ function Get-DbaDbBackupHistory {
 
     .PARAMETER Since
         Specifies a starting point for the search for backups. If a DateTime object is passed, that will be used. If a TimeSpan object is passed, that will be added to Get-Date and the resulting value will be used.
+        
     .PARAMETER RecoveryFork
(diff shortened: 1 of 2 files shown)


functions/Get-DbaDbBackupHistory.ps1
```diff
@@ -158,7 +158,7 @@ function Get-DbaDbBackupHistory {
         [switch]$IncludeCopyOnly,
         [Parameter(ParameterSetName = "NoLast")]
         [switch]$Force,
-        [DateTime]$Since = (Get-Date '01/01/1970'),
+        [object]$Since = (Get-Date '01/01/1970'),
```
potatoqualitee, line 161:
```suggestion
        [psobject]$Since = (Get-Date '01/01/1970'),
```

I don't know the difference, exactly, but `psobject` was recommended to me over `object` and I found it worked better than object.

functions/Get-DbaDbBackupHistory.ps1
```diff
@@ -220,6 +219,13 @@ function Get-DbaDbBackupHistory {
             return
         }
 
+        if ($Since -is [timespan]) {
+            $Since = (Get-Date).add($Since);
+        } elseif (-not ($Since -is [datetime])) {
```
potatoqualitee, line 224:
```suggestion
        } elseif ($Since -isnot [datetime]) {
```

Simplifying

ben-thul:
I can't seem to add a comment inline at the appropriate spot, but the documentation for the Since parameter should be updated as well to reflect that either a DateTime or a TimeSpan can be passed and what it means for each of those.

potatoqualitee:
@ben-thul let me know the wording and I can add it. I was also surprised it was harder than expected to Suggest a change to commit. In the end, I clicked Edit File and it prompted me to edit but I didn't know what exactly to say.

ben-thul:
@potatoqualitee - I think I can use whatever mechanism you used to add commits to this PR/branch. Or maybe not (i.e. if that's only available to you as having superpowers on the dataplat:development branch). Either way, my changes are in [this commit](https://github.com/ben-thul/dbatools/commit/55bc0eae5725dcad3a4f7034e27e7a2db2959720)

potatoqualitee:
@ben-thul thank you! will add and merge today and this will be a part of dbatools 2.0

functions/Get-DbaDbBackupHistory.ps1
```diff
@@ -31,8 +31,7 @@ function Get-DbaDbBackupHistory {
         If this switch is enabled, a large amount of information is returned, similar to what SQL Server itself returns.
 
     .PARAMETER Since
-        Specifies a DateTime object to use as the starting point for the search for backups.
-
+        Specifies a starting point for the search for backups. If a DateTime object is passed, that will be used. If a TimeSpan object is passed, that will be added to Get-Date and the resulting value will be used.
```
potatoqualitee, line 34:
```suggestion
        Specifies a starting point for the search for backups. If a DateTime object is passed, that will be used. If a TimeSpan object is passed, that will be added to Get-Date and the resulting value will be used.
        
```

forgot the space

PR #8633, opened 2022-11-08
Initial update for dbatools 2.0

potatoqualitee:
* update test suite
  * Just some renames to better align with what they do
* about topics need additional updating
  * before 2.0 goes live, we need to make a bunch of documentation to help devs. particularly, we'll need to `Install-Module dbatools.library -AllowPrerelease` on 5.1 and `Install-Module dbatools.core.library -AllowPrerelease` 
* connection string updates
  * Somewhere between 1.5 and 2.0, Microsoft finished its migration from System.Data to Microsoft.Data so the toolset needed to be updated
* Move PBM Library Loading to speed up overall import
  * PBM isn't used very often if I had to guess, so I moved the load from `Import-Module dbatools` to when the commands are actually run. This will help speed up the initial import by prolly 50-100 ms.
* Change XE Smart Target Load
  * Same as PBM
* Lumenworks has been moved to primary import
  * `Import-DbaCsv` is used a lot, so don't load it every time
* Dac now part of initial load
  * Import speeds were minimally impacted and these are used pretty often
* Update replication
  * Replication now works, due to the way I setup DLL Redirection in the dbatools library import!
* reformat script to align with dbatools style
* Perf improvements: calling this class vs variable on Linux was slower
  * I'll never understand why setting that variable to false was so slow! But a global var is much faster.
* Enable runspaces for TEPP
  * TEPP loading is so slow, I'm trying everything to load it faster. Runspaces don't always pay off but they did for TEPP.
* maintenance is one of the slower scripts
  * I wish this could be faster :thinkies:
* Registry reads slow down import.
  * WOW! `HKLM:\` is _way_ slower than `Registry::HKEY_Local_Machine\`. Thanks Joel Bennett
* All changes intended to improve performance and match dbatools style
* Importing Optional scripts is no longer performed on mac and linux
  * This slowed stuff down, so skip bc they have what they need
* The underlying library no longer supported `Add`, but does support `AddWithValue`
* Serial vs DoDotSource -- one too many options, remove
  * Removed compile library each time too
  * Decided that if people want this, they can compile themselves from the git repo but no need to add too many features
* moved library out of dbatools and into its own module
  * Yesssss
* Removed watcher commands (Watch-DbatoolsUpdate) because they don't work and aren't really in scope. If someone wants to make this with BurntToast, that's much better.
* Minimized loading
  * Shaved off blocks of milliseconds by moving to native dotnet stuff


potatoqualitee:
Is anyone interested in reviewing this? If not, it passes all tests and I can merge.

potatoqualitee:
K if anyone has feedback down the line, please feel free to comment here or however you'd like to address it.

PR #8734, opened 2023-01-26
Update dbatools.library and dbatools.core.library dependency versions

potatoqualitee:
Also changed the version of dbatools to dbatools 2.0.0-preview


potatoqualitee:
Still need to fix azure automation 😭

PR #8788, opened 2023-03-04
Add ARM64 Self-Hosted Runner

potatoqualitee:
This workflow uses a self-hosted runner that is 100% dedicated to dbatools. It's actually a VM in the Oracle Cloud Free Tier which I heard may just disappear one day then Oracle Cloud Support won't return our calls. But until then....

ARM64 tests are important because it's so different from x64 and SMO assemblies (XE stuff in particular) has been known to fail in the past.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -5,7 +5,11 @@ defaults:
     shell: pwsh
 jobs:
   linux-tests:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, arm64]
     env:
       GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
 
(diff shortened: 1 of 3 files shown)


wsmelton:
There are a few ways to do this with GitHub Actions hosted runners. Never tried them so not sure how unique they are but they use a Raspberry PI emulator I think.

https://github.com/marketplace/actions/arm-runner

potatoqualitee:
I tried emulation in the past but SQL Server was too much for qemu :/

If I recall, I used it to build the docker base images but can't actually run SQL Server tests because SQL won't start

https://github.com/dataplat/docker/blob/main/.github/workflows/mssqlbase.yml#L16
https://github.com/dataplat/docker/blob/main/.github/workflows/arm64.yml

potatoqualitee:
azure automation must be on arm64 sometimes because arm64 failures have been reported here. and some commands, like the XE ones, won't fail until they are run so this provides reassurance.

wsmelton:
Azure Automation accounts use a sandbox Windows container to run PowerShell. As far as I know the PS 7 preview runtime do also but I don't bother using those because they have to many issues.

They might be using the hybrid worker because Azure does have ARM64 support in the VMs now. They are much cheaper to run.

PR #8844, opened 2023-04-05
Show details for objects being synced in Sync-DbaAvailabilityGroup

potatoqualitee:
Fixes #4748

PR #8846, opened 2023-04-05
Standardize Whatif Output

potatoqualitee:
I noticed in #8844 that the output was not standardized (and sometimes wrong?) so fixing that. Please double check my work, especially in the copy sys db user object


potatoqualitee:
I appreciate and love a review by anyone who is detail oriented. I ran tests and it's beautiful but would just like another set of eyes.

potatoqualitee:
gonna do a new release

PR #8854, opened 2023-04-10
Fix TEPP/autocomplete

potatoqualitee:
Basically, it seems like the runspace method does not work. It's only like 143ms to make it sync so I added it because I think it's pretty valuable.


wsmelton:
The async is probably broke because of the security changes. It can't trust the connection to connect on its own?

potatoqualitee:
i wondered that too but it works in dev :/ And dev imported sync. Unfortunately, I am not good enough with C# to figure it out exactly.

PR #8878, opened 2023-04-27
Add Maintenance Solution to module

potatoqualitee:
It's MIT so 👍🏼  Include local copy for offline servers and/or if there's an issue downloading the latest copy.

Adding this to deployer

```powershell
Remove-Item -EA SilentlyContinue -Path C:\github\dbatools\bin\maintenancesolution\*
Save-DbaCommunitySoftware -Software MaintenanceSolution -LocalDirectory C:\temp
Copy-Item -Recurse C:\temp\*maintenance-solution*\*.sql C:\github\dbatools\bin\maintenancesolution
```

### DIFF
--- a/bin/maintenancesolution/CommandExecute.sql
+++ b/bin/maintenancesolution/CommandExecute.sql
@@ -1,0 +1,293 @@
+﻿SET ANSI_NULLS ON
+GO
+SET QUOTED_IDENTIFIER ON
+GO
+IF NOT EXISTS (SELECT * FROM sys.objects WHERE object_id = OBJECT_ID(N'[dbo].[CommandExecute]') AND type in (N'P', N'PC'))
+BEGIN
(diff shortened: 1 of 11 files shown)


andreasjordan:
Hi Chrissy, I currently work on a lab where I work with a non-admin account. The Install-DbaMaintenanceSolution failes with "TerminatingError(Get-NetConnectionProfile): "Cannot connect to CIM server. Access denied "", because the command "Get-NetConnectionProfile" is there but can not be used by a non-admin.

Would you be open for a pull request that changes the command to be able to run this as a non-admin? Maybe we can remove the test for internet access and jut add a now parameter with a name like "-UseIncludedVersion" (but we need a better name).

andreasjordan:
Maybe it's only my setup. Will do more tests and comment again.

PR #8937, opened 2023-05-30
Get-DbaAgentSchedule - Use better switch syntax

andreasjordan:
As the switch statement never tests against the input this is not a change in execution, but better for readability.


potatoqualitee:
ty! 🙇🏼

PR #9017, opened 2023-06-23
Copy-DbaSystemDbUserObject - Create schema with correct owner

### LINKED ISSUE #9011: Start-DBAMigration changes two msdb schema authorizations to dbo in the target instance

Bug - Start-DbaMigration apparently migrates schemas, but doesn't copy over the Authorization value.  In msdb there are a couple of schemas that are referenced in MS script msdb110_upgrade.sql that is part of Cumulative Update installs and after I run Start-Migration the schemas TargetServersRole and DatabaseMailUsersRole change authorizations to dbo, rather than TargetServersRole and DatabaseMailUsersRole which are the Authorization values in the Source instance.  All the instances that are the result of a Start-DbaMigration have been experiencing CU install failures with event errors citing the msdb110_upgrade.sql script and I have to drop the two schemas before the CU install will complete.  The only difference in msdb for the instances that fail compared to the ones that don't fail is the schema Authorizations.  I am using dbatools version 1.1.143 on SQL Server 2017 and 2019 on Windows.

### Steps to Reproduce

```powershell
```
<# I start with both the source instance and the target instance having the same values for the msdb schemas DatabaseMailUserRole and TargetServersRole - and yes, the schema names appear to be Role names but that is how Microsoft has set these up.  In the Information_Schema.Schemata tale both the SCHEMA_NAME and SCHEMA_OWNER values are the same - the schema TargetServersRole, the SCHEMA_NAME is TargetServersRole and the SCHEMA_OWNER is TargetServersRole.  But after I run Start-DbaMigration with the following code, the SCHEMA_OWNER changes to dbo.  The same deal for the schema named DatabaseMailUserRole.  Note that the source instance and the target instance are on the same server and are the same SQL Server version.  This happens on both SQL 2017 and SQL 2019.    #>
$SourceInstance = 'SERVERName\InstanceName1'
$DestinationInstance = 'SERVERName\InstanceName2'

## Last migration step is to actually do the migration from the one instance to the other.  Excludes logins because we already took care of those.

    $ExcludeList = "SpConfigure", "Logins"

    $params = @{   
        Source = $SourceInstance
        Destination = $DestinationInstance
        UseLastBackup = $true
        Exclude = $ExcludeList
        DisableJobsOnDestination = $true
        ReuseSourceFolderStructure = $false
        Force = $true
        Whatif = $false
        }

    $params  # debug and display
    
    Start-DbaMigration @params -EnableException -Verbose -ErrorAction Stop | Format-Table Status,Type,Name -Wrap -AutoSize 

I am running version 1.1.143.  I suspect that dbatools has been doing this Auth changing for years based on how for years I have needed to drop these schemas to get upgrades to work.

### Other details or mentions

_No response_

### PowerShell Host Version

5.1.17763.3770

SQL 2019 15.0.4312.2
SQL 2017 14.0.3460.9

### .NET Framework Version

Client 4.8.03761
Full 4.8.03761
Client 4.0.0.0

andreasjordan:

 - [ ] Please confirm you have the smaller repo (85MB .git directory vs 275MB or 110MB or 185MB .git directory)

## Type of Change

 - [x] Bug fix (non-breaking change, fixes #9011  )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

See issue for details.

@potatoqualitee - Please have a look if that line should be added to other transfers as well.

### DIFF
--- a/public/Copy-DbaSystemDbUserObject.ps1
+++ b/public/Copy-DbaSystemDbUserObject.ps1
@@ -167,6 +167,7 @@ function Copy-DbaSystemDbUserObject {
                         $transfer = New-Object Microsoft.SqlServer.Management.Smo.Transfer $smodb
                         $null = $transfer.CopyAllObjects = $false
                         $null = $transfer.Options.WithDependencies = $true
+                        $null = $transfer.Options.ScriptOwner = $true
                         $null = $transfer.ObjectList.Add($schema)
                         if ($PSCmdlet.ShouldProcess($destInstance, "Attempting to add schema $($schema.Name) to $systemDb")) {
                             try {


potatoqualitee:
sooo happy you fixed this one, thank you! im unsure of the downsides. what if the owner does or does not exist?

andreasjordan:
Is a schema without an owner possible?

potatoqualitee:
sorry I meant what if the owner doesnt exist on the destination

andreasjordan:
Ok, that would be a problem.

I generally have a problem with that command: Why would I want to use it? The schemas related to this bug here are "user objects" as per the IsSystemObject property in the object, but I think they are system objects as they will be there on the destination as they are part of the msdb core functionality. We just don't have a way to identify them by a rule. The command would copy over things like the command table for Ola or the version tables for the First Reposner Kit. But I would not have that copied over, because they would be created when installing those tools. So it's only for real user generated content that should not be in system databases. So I don't know what the best solution would be...

potatoqualitee:
you know, i think if the owner doesnt exist, just let it error out naturally. better than the issue we've got going now.

LGTM ☑️

gittroy:
Hello, I'm a new dbatools user (and new to this forum).  Is there a way to tell what version of dbatools will contain this fix?   I am seeing the same issue as: 
**_Start-DBAMigration changes two msdb schema authorizations to dbo in the target instance #9011_**
except it affects an additional six schemas in msdb (total 8 affected).  I am catching this just now as we are trying to update to CU21 and ran into the issue from: 
https://learn.microsoft.com/en-us/troubleshoot/sql/database-engine/install/windows/sqlserver-patching-issues#targetserversrole-schema-and-security-role

Below are the systems that all saw the same issue and the dbatools versions used on each:
VM#1:   2.0.4
VM#2:   2.0.4
VM#3:   1.1.145
VM#4:   1.1.146

enolke:
I ran into this very same issue a handful of times over a year ago when
doing cumulative updates on migrated servers and it caused some gray hair
and delays.  If a new build could reduce the possibility  of dbatools
contributing to this in the future, I would really love it.
Thanks!
https://learn.microsoft.com/en-us/troubleshoot/sql/database-engine/install/windows/sqlserver-patching-issues#targetserversrole-schema-and-security-role

PR #9026, opened 2023-07-04
Improve formatting

andreasjordan:
I've used this command: `Get-ChildItem -Path .\public\*.ps1 | Invoke-DbatoolsFormatter`

First commit is before changing `Invoke-DbatoolsFormatter`, second commit is after adding this code to `Invoke-DbatoolsFormatter`:

```
$correctCase = @(
    'DbaInstanceParameter'
    'PSCredential'
    'PSCustomObject'
    'PSItem'
)
```

```
foreach ($item in $correctCase) {
    $line = $line -replace $item, $item
}
```

This way we can fix the wrong case without doing this individually.

Don't know if we want to use this, feedback is welcome.


wsmelton:
This all depends on what standard we want to use for casing of type names. The ones you call out that are in PowerShell, most teams and I think PS Team has this standard as well are kept as lowercase. That is because the `System.Management.Automation.TypeAccelerators` has them as lowercase. You can run this to return that list and see the casing on all built-in types: `[PSObject].Assembly.GetType('System.Management.Automation.TypeAccelerators')::Get`.

In reality if we want to keep the same casing for built-in types, we can use the above line to pull them out, match them, and then set to the same value.

andreasjordan:
I just saw that some of the recent pull requests also changed pscustomobject to PSCustomObject and so I wanted to start this discussion as well as presenting a way to change the casing (is that the correct word?) with the command that we already have.

So I'm fine with uppercase, lowercase or "don't care". But if we don't care, we should not change it in pull requests. And if we care, we should have a way to document and enforce the "correct" spelling. I personally like "PSCustomObject" as I think its easier to read. But I will follow any guideline we set up.

potatoqualitee:
yeah casing is dont care, some people change it in a PR and I accept it bc it's not a big deal. Well, except for -EQ, I do not like that one 😅

PR #9094, opened 2023-09-22
Fix stale cache issue in Add-DbaAgDatabase

agowa:
When multiple databases are joined to an AG using Add-DbaAgDatabase using the same SqlInstance objects Get-DbaAgDatabase returns $null for all except the first database.

It looks like this issue does NOT happen when the dbatools internal backup function is used beforehand. But if a stored procedure is invoked to create the backup it does.

In order to fix this I added a switch to Get-DbaAgDatabase to update the Availability group object before use.

Note: I was unable to run `.\tests\manual.pester.ps1` within the GitHub codespace, as even after manually installing PSScriptAnalyzer it still failed to find and use it.

EDIT: Only partially fixes the problem, it still sometimes returns null. Interestingly even if the Get-DbaAgReplica call still returns $null, changing it to `Get-DbaAvailabilityGroup -SqlInstance $replicaServerSMO[$replicaName] -SqlCredential $SqlCredential -AvailabilityGroup $AvailabilityGroup |  Get-DbaAgDatabase -AvailabilityGroup $Avail
abilityGroup -Database $db.Name -EnableException` appears to almost always work even in these cases?!

Edit2: Fixed now too, the issue is that `AvailabilityReplicas.Refresh()` needs to be called for `$InputObject.AvailabilityDatabases` to be updated. (No, that's not a typo, it's really refreshing Replicas for Databases to be updated 🙄)

## Please read -- recent changes to our repo
On November 10, 2022, [we removed some bloat from our repository (for the second and final time)](https://github.com/dataplat/dbatools/issues/8542). This change requires that all contributors reclone or refork their repo.

PRs from repos that have not been recently reforked or recloned will be closed and @potatoqualitee will cherry-pick your commits and open a new PR with your changes.

 - [x] Please confirm you have the smaller repo (85MB .git directory vs 275MB or 110MB or 185MB .git directory)

## Type of Change

 - [x] Bug fix (non-breaking change, fixes # )
 - [x] New feature (non-breaking change, adds functionality, fixes # )
 - [ ] Breaking change (affects multiple commands or functionality, fixes # )
 - [ ] Ran manual Pester test and has passed (`.\tests\manual.pester.ps1`)
 - [ ] Adding code coverage to existing functionality
 - [ ] Pester test is included
 - [ ] If new file reference added for test, has is been added to github.com/dataplat/appveyor-lab ?
 - [ ] Unit test is included
 - [X] Documentation
 - [ ] Build system

### Purpose

[...]

PR #9101, opened 2023-09-29
Add-DbaAgDatabase - Fix stale cache issue

andreasjordan:
Fix for #9094. Reverts changes and implements the fix with just one line of code.

### DIFF
--- a/.github/workflows/integration-tests-repl.yml
+++ b/.github/workflows/integration-tests-repl.yml
@@ -149,4 +149,4 @@ jobs:
       - name: Run replication tests part 3
         run: |
           Import-Module ./dbatools.psd1 -Force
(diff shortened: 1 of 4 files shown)


potatoqualitee:
@jpomfret I thought I disabled replication3. Can you check in dev why they're failing? Maybe it's just too flaky on GH actions?

@andreasjordan -- in the previous PR, it said "# calling AvailabilityDatabases.Refresh() alone is not enough". @agowa can you explain more?

agowa:
@potatoqualitee I tried it with a local development setup of three SQL servers with one Availability group. When I just called AvailabilityDatabases.Refresh() it didn't show the current state, just the same outdated information as before. But after I first refreshed the AG itself it then correctly reflected the current state.

PR #9118, opened 2023-10-08
Set-DbaDbCompression, prefer online rebuilds where possible

### LINKED ISSUE #9102: Set-DbaDbCompression to do ONLINE Rebuilds
### Summarize Functionality

The current version does all rebuilds offline, which can cause a lot of blocking; don't ask me how I found out.  

Yes

### Technical Details

I think this is reasonably simple to fix, although I couldn't quite seem to verify the OnlineIndexOperation property was working on my end.  I think you can use logic for editions, as it would just convert it to OFFLINE if you wanted to throw up a warning to let people know it would be OFFLINE.

`if ($CompressionType -eq "None") {
                                        $query = "ALTER INDEX [$($index.Name)] ON $($index.Parent) REBUILD PARTITION = ALL WITH (DATA_COMPRESSION = $CompressionType, ONLINE=ON)"
                                        $Server.Query($query, $db.Name)
                                    } else {
                                        $($index.PhysicalPartitions | Where-Object { $_.PartitionNumber -eq $P.PartitionNumber }).DataCompression = $CompressionType
                                        $index.OnlineIndexOperation = true
                                        $index.Rebuild()
                                    }`

niphlod:

 - [x] Please confirm you have the smaller repo (85MB .git directory vs 275MB or 110MB or 185MB .git directory)

## Type of Change

 - [x] New feature (non-breaking change, adds functionality, fixes #9102 )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
Try to do online operations where possible. 

### Approach
There's a swanky "IsOnlineRebuildSupported" on the index property to signal if the version supports online operations. Unfortunately, it's not on the database level.
So, we loop over existing indexes, and if one is found and IsOnlineRebuildSupported is true, we set the relevant flags on the rebuild operation.
Fixes:
- ~~there was a manual FIX due to a SMO bug for resetting tables to "None", but it seems to be fixed, so I removed the workaround from the code~~ reintroduced, as it seems to be working on AZ but not onprem.

Problems:
- Azure gets back false on any index, even if AFAIK every Azure instance supports online rebuilds.... so we patch SMO at the top

Possible problems:
- if you just have heaps, there's no way to detect if online operations are supported if not manually detecting versions and editions 
- technically every index can have that flag .... is there some kind of situation in which one index is rebuildable online and another isn't ? if so, even if AN index is "online rebuildable", this code fails to set it on every operation, and the check must be done on each object and reverted to a clunky "globally I should be able to do it, but I can't for this specific one"... which, in Azure-related cases, is going to be impossible as the property on each index is set to false

### DIFF
--- a/public/Set-DbaDbCompression.ps1
+++ b/public/Set-DbaDbCompression.ps1
@@ -41,7 +41,7 @@ function Set-DbaDbCompression {
         Will only work on the tables/indexes that have the calculated savings at and higher for the given number provided.
 
     .PARAMETER ForceOfflineRebuilds
-        By default, this function prefers online rebuilds over offline ones. 
+        By default, this function prefers online rebuilds over offline ones.
         If you are on a supported version of SQL Server but still prefer to do offline rebuilds, enable this flag
 
     .PARAMETER InputObject
@@ -163,8 +163,7 @@ function Set-DbaDbCompression {
                         break
                     }
                     foreach ($index in $($obj.Indexes | Where-Object { !$_.IsMemoryOptimized -and $_.IndexType -notmatch 'Columnstore' })) {
-                        if ($index.IsOnlineRebuildSupported -or $isAzure)
-                        {
+                        if ($index.IsOnlineRebuildSupported -or $isAzure) {
                             $isOnlineRebuildSupported = $true
                             $onlineRebuildScanCompleted = $true
                             break
@@ -173,8 +172,7 @@ function Set-DbaDbCompression {
                 }
                 Write-Message -Level Verbose -Message "Are Online Rebuilds supported ? $isOnlineRebuildSupported"
                 $CanDoOnlineOperation = $false
-                if ($IsOnlineRebuildSupported -and !$ForceOfflineRebuilds)
-                {
+                if ($IsOnlineRebuildSupported -and !$ForceOfflineRebuilds) {
                     $CanDoOnlineOperation = $true
                     Write-Message -Level Verbose -Message "Using Online Rebuilds where possible"
                 }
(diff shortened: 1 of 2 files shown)


public/Set-DbaDbCompression.ps1
```diff
@@ -131,15 +137,14 @@ function Set-DbaDbCompression {
             if ($server.EngineEdition -notmatch 'Enterprise' -and $server.VersionMajor -lt '13') {
                 Stop-Function -Message "Only SQL Server Enterprise Edition supports compression on $server" -Target $server -Continue
             }
-
+            $isAzure = $server.DatabaseEngineType -match "Azure"
```
wsmelton, line 140: This is duplicate code, it is already part of the server object via `Connect` command

https://github.com/dataplat/dbatools/blob/73ff74e4efe569025bd4c544837ff71b790c8ed7/public/Connect-DbaInstance.ps1#L1028-L1045

public/Set-DbaDbCompression.ps1
```diff
@@ -150,6 +155,28 @@ function Set-DbaDbCompression {
...
                 }
+                $isOnlineRebuildSupported = $false
+                # Loop on indexes to see if Rebuild Online is supported
+                $onlineRebuildScanCompleted = $false
+                foreach ($obj in $tables | Where-Object { !$_.IsMemoryOptimized -and !$_.HasSparseColumn }) {
+                    if ($onlineRebuildScanCompleted) {
+                        break
+                    }
+                    foreach ($index in $($obj.Indexes | Where-Object { !$_.IsMemoryOptimized -and $_.IndexType -notmatch 'Columnstore' })) {
+                        if ($index.IsOnlineRebuildSupported -or $isAzure) {
```
wsmelton, line 166:
```suggestion
                        if ($index.IsOnlineRebuildSupported -or $server.IsAzure) {
```

wsmelton (changes requested):
Initial review it looks good, just need to remove duplicated code (see comments)

niphlod:
thanks @wsmelton , fixed it.

potatoqualitee:
Thankyou both!

PR #9143, opened 2023-10-29
Connect-DbaInstance - Change EncryptConnection to a switch with a default of $true

### LINKED ISSUE #9113: Connect-DbaInstance - ignores -Encrypt False

The Connect-DbaInstance seems to ignore the -Encrypt parameter, if the value is False.
It seems there is missing an else statement on line 788.

https://github.com/dataplat/dbatools/blob/c253ff5e44dfb6193c5cc63e3e898143371d9fa3/public/Connect-DbaInstance.ps1#L788C1-L791C22

### Steps to Reproduce

```powershell
$a = Connect-DbaInstance -SqlInstance "sqlinstance" -EncryptConnection false
$a.ConnectionContext.ConnectionString
```
The ConnectionString will always be Encrypt=True

2.1.1

### Other details or mentions

_No response_

PowerShell Core (pwsh.exe)

### PowerShell Host Version

PSVersion                      7.3.7
GitCommitId                    7.3.7
OS                             Microsoft Windows 10.0.22621

Microsoft SQL Server 2022 (RTM) - 16.0.1000.6 (X64)   Oct  8 2022 05:58:25   Copyright (C) 2022 Microsoft Corporation  Developer Edition (64-bit) on Windows Server 2022 Standard 10.0 <X64> (Build 20348: ) (Hypervisor) 

### .NET Framework Version

.NET 7.0.11

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #9113 )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

See issue for details.

Up for discussion: Should we always use EncryptConnection = $true and only change TrustServerCertificate to $true within `Set-DbatoolsInsecureConnection`? I think it would be best practice to change the least number of default as possible. And switching off encryption is not needed if we only trust the (self-signed) server certificate.

### DIFF
--- a/.github/workflows/integration-tests-repl.yml
+++ b/.github/workflows/integration-tests-repl.yml
@@ -21,7 +21,7 @@ jobs:
         run: |
           Import-Module ./dbatools.psd1 -Force
           Set-DbatoolsConfig -FullName sql.connection.trustcert -Value $true -Register
-          Set-DbatoolsConfig -FullName sql.connection.encrypt -Value Optional -Register
+          Set-DbatoolsConfig -FullName sql.connection.encrypt -Value $false -Register
           Get-DbatoolsConfigValue -FullName sql.connection.encrypt | Write-Warning
 
       - name:  Setup docker images
@@ -70,7 +70,7 @@ jobs:
         run: |
           Import-Module ./dbatools.psd1 -Force
           Set-DbatoolsConfig -FullName sql.connection.trustcert -Value $true -Register
-          Set-DbatoolsConfig -FullName sql.connection.encrypt -Value Optional -Register
+          Set-DbatoolsConfig -FullName sql.connection.encrypt -Value $false -Register
           Get-DbatoolsConfigValue -FullName sql.connection.encrypt | Write-Warning
 
       - name:  Setup docker images
@@ -119,7 +119,7 @@ jobs:
         run: |
           Import-Module ./dbatools.psd1 -Force
           Set-DbatoolsConfig -FullName sql.connection.trustcert -Value $true -Register
-          Set-DbatoolsConfig -FullName sql.connection.encrypt -Value Optional -Register
+          Set-DbatoolsConfig -FullName sql.connection.encrypt -Value $false -Register
           Get-DbatoolsConfigValue -FullName sql.connection.encrypt | Write-Warning
 
       - name:  Setup docker images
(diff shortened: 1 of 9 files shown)


potatoqualitee:
Thank you so much, andreas! I've done this before and have had complaints. We talked about it here https://github.com/dataplat/dbatools/pull/9021

I'm open to merging this since the tests are passing. but I'm concerned about old documentation that mentions Optional. Is there no way for us to handle that? I remember it being SO weirdly picky in that it only wanted a $true or $false (but maybe its the bit thing you mentioned?_

andreasjordan:
I think we must not mix two things:
* [SqlConnectionEncryptOption](https://learn.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqlconnectionencryptoption?view=sqlclient-dotnet-standard-5.1) is a string and is currently the related class to the config 'sql.connection.encrypt' and the parameter EncryptConnection of Connect-DbaInstance. But we don't use it that way!
* We use the property EncryptConnection of the type [Microsoft.SqlServer.Management.Common.SqlConnectionInfo](https://learn.microsoft.com/en-us/dotnet/api/microsoft.sqlserver.management.common.sqlconnectioninfo?view=sql-smo-160) which is a bool. So we definitly use a bool and thus we need a bool to set this to one of the two possible values: $true or $false.

Currently we can only set this to $true.

We could change the code inside of Connect-DbaInstance to look at the value of the string and set the bool based on this rules: https://learn.microsoft.com/en-us/dotnet/api/microsoft.data.sqlclient.sqlconnectionencryptoption?view=sqlclient-dotnet-standard-5.1
But I personally think "the normal user" would see the parameter as a switch (as the current docs also tell) and vote for changing this to a switch so set encryption to $true or $false.

But we can let this pr sit here for a while and discuss the topic.

potatoqualitee:
thank you andreas. do they mix well? that sqlclient is used within the command. if that still works, I'm happy to merge as this comes up regularly

andreasjordan:
Do you mean "Microsoft.Data.SqlClient" when you talk about "sqlclient"? That is only used twice;

In the code path for `$inputObjectType -in 'RegisteredServer', 'ConnectionString'` (starting line 647) and when the parameter `AzureToken` is used (starting line 875). In the first place, SqlConnectionStringBuilder is used, in the second place a SqlConnection is created based on a connection string. In both places, the value of the parameter `-EncryptConnection` is not used.

andreasjordan:
Please have a look at the yml files. Maybe you want to change the parameter to $true there to always test the new default: encrypted connection. I can update the pr if you like.

potatoqualitee:
Let's do this and see how it goes. Thank you again 🙇🏼

PR #9146, opened 2023-10-31
Set-DbaDbCompatbility - Fix for using piped input and don't trust master compatibility level

andreasjordan:
## Type of Change

 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

Experienced this bug today at a client wihle doing `Get-DbaDatabase -SqlInstance sql01 | Where-Object Name -like Test* | Set-DbaDbCompatibility`.

If we support pipeline input, we must not use $SqlInstance. And every database can be on a different server with a different server level. So we can only get the server level inside of the loop for every individual database. And I used the same method to get the server level like in Test-DbaDbCompatibility (see #8992 for details).


potatoqualitee:
solid change 💪🏼

PR #9309, opened 2024-03-31
Get-DbaStartupParameter - Remove fixed typing of variables that may change their type

### LINKED ISSUE #9197: Get/Set DbaStartupParameter - 2 issues

Cannot convert value "-T2467" to type "System.Int32". Error: "Input string was not in a correct format."

Output is also wrong (path shortened to make it easy to read):

Example startup parameter string: -dC:\path\master.mdf;-eC:\path\ERRORLOG;-lC:\path\mastlog.ldf;-T2467;-T4139;-T8121;-T11064;-T11068;-T11631;-T11634

The startup parameter string put in place by `Set-DbaStartupParameter`: -dC:\path\master.mdf C:\path\master.mdf C:\path\master.mdf;-eC:\path\ERRORLOG C:\path\ERRORLOG C:\path\ERRORLOG;-lC:\path\mastlog.ldf C:\path\mastlog.ldf C:\path\mastlog.ldf;-c;-f;-g0 0 0-m;-n;-s;-x;-E;-T2467;-T4139;-T8121;-T11064;-T11068;-T11631;-T11634;

### Steps to Reproduce

```powershell
# Test the Set function - run twice to see it happen - substitute whatever trace flags you like
Set-DbaStartupParameter -SqlInstance MySqlServer.domain.local -TraceFlag 1,2,3 -TraceFlagOverride -Offline
Set-DbaStartupParameter -SqlInstance MySqlServer.domain.local -TraceFlag 1,2,3 -TraceFlagOverride -Offline
# Test the Get function
Get-DbaStartupParameter -SqlInstance MySqlServer.domain.local
```

2.1.6

### Other details or mentions

- Get-DbaStartupParameter has an issue with variable types. When the function is run twice (seems to particularly rear its head when `Set-DbaStartupParameter` is run twice, thus calling this function multiple times at once), the second pass remembers the variable typing in the script block. This becomes problematic with the `$traceFlags` variable (if used), because it may either contain a string (e.g. "None") or an int array coming out of the previous run - but it is first populated as a string array. The subsequent-run situation causes the following error to be thrown: _Cannot convert value "-T9999" to type "System.Int32". Error: "Input string was not in a correct format."_ (9999 given as an example). This issue likely affects the `$debugFlags` variable as well, since it goes through the same type transformations as `$traceFlags`. As a result of this type conversion error, function `Invoke-ManagedComputerCommand` believes that the remote code execution has entirely failed, however that is not the case! The code execution succeeds, meaning the final set of results is returned. Nevertheless, because it was believed to have failed, `Invoke-ManagedComputerCommand` runs repeatedly (versions 16 through 8), with the code erroring-yet-succeeding at multiple levels (likely depending on what versions of SQL WMI Management DLLs are available in the GAC on the remote computer). This leads to the subsequent runs not only having messages indicating errors, but returning an object array rather than a single object.
  - This was previously reported, in part, in #7939 but that issue was closed, believed to have been fixed in an earlier version.
  - This was also previously reported in #8658 which was also closed, also believed to have been fixed in a previous version. A code solution was proposed in this issue that resolves the problem in my tests:

> I would change `[int[]]$traceFlags = $traceFlags.substring(2)` to `$traceFlags = [int[]]$traceFlags.substring(2)`.

- The problem (above) with the `Get` function has exposed what could be considered a weakness in `Set-DbaStartupParameter`. When the `Set` function receives the object array, it cannot interpret the values correctly. This causes problems of two sorts:
  - For path parameters such as Master Data, ERRORLOG, etc, the parameter string includes the path multiple times. For instance, if the path to **master.mdf** was _C:\path\master.mdf_, and the `Get` function had returned 3 elements instead of 1, the resulting parameter string would be "-dC:\path\master.mdf C:\path\master.mdf C:\path\master.mdf;" - and this is the value that would be put in place on the SQL instance.
  - For option parameters such as CommandPromptStart, the comparison between the `Set` function's input params (switch/boolean type) and the output from the `Get` function (array of boolean) is a mismatch - therefore all of these parameters are set to the opposite of what is desired (false becomes true and vice versa). Not only does this suddenly change all of the startup parameters on the SQL instance, but it sends some of the logic down the wrong path (especially in the case of the Single User setting).
  - None of this behavior would be triggered if the `Get` function were fixed, but adding some safety mechanism to the `Set` function could prevent issues down the line. Therefore making changes to the `Set` function is optional.

Windows PowerShell (powershell.exe), Windows PowerShell ISE (powershell_ise.exe)

### PowerShell Host Version

| Name | Value |
| ---- | ----- |
| PSVersion | 5.1.17763.5202 |
| PSEdition | Desktop |
| PSCompatibleVersions | {1.0, 2.0, 3.0, 4.0...} |
| BuildVersion | 10.0.17763.5202 |
| CLRVersion | 4.0.30319.42000 |
| WSManStackVersion | 3.0 |
| PSRemotingProtocolVersion | 2.3 |
| SerializationVersion | 1.1.0.1 |

Microsoft SQL Server 2022 (RTM-CU4) (KB5026717) - 16.0.4035.4 (X64)   May  1 2023 10:34:45   Copyright (C) 2022 Microsoft Corporation  Developer Edition (64-bit) on Windows Server 2019 Datacenter 10.0 <X64> (Build 17763: ) (Hypervisor) 

### .NET Framework Version

| PSChildName | Version |
| ----------- | ------- |
| Client | 4.7.03190 |
| Full | 4.7.03190 |
| Client | 4.0.0.0 |

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #9197 )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

See issue for details.

### DIFF
--- a/public/Get-DbaStartupParameter.ps1
+++ b/public/Get-DbaStartupParameter.ps1
@@ -88,13 +88,13 @@ function Get-DbaStartupParameter {
                     if ($traceFlags.length -eq 0) {
                         $traceFlags = "None"
                     } else {
-                        [int[]]$traceFlags = $traceFlags.substring(2)
+                        $traceFlags = [int[]]$traceFlags.substring(2)
                     }
 
                     if ($debugFlags.length -eq 0) {
                         $debugFlags = "None"
                     } else {
-                        [int[]]$debugFlags = $debugFlags.substring(2)
+                        $debugFlags = [int[]]$debugFlags.substring(2)
                     }
 
                     if ($Simple -eq $true) {


potatoqualitee:
nice!

Issue #8713, opened 2023-01-05
Create contained availability group


thewaiting28:
### Summarize Functionality

Would love to be able to create new contained availability group, either with New-DbaAvailabilityGroup or a new command? Also query whether or not an AG is contained. Thank you!

### Is there a command that is similiar or close to what you are looking for?

Yes

### Technical Details

Something like "New-DbaAvailabilityGroup -Primary sql01\inst01 -Secondary sql02\inst02 -Name "ag01" -Database "Database01" -Contained $true

Or, -Contained just as a switch

wsmelton:
Quick check in SMO the preview release does not show support of this as of yet. I'm not sure if sqlserver module is supporting this already.

At this time, we can't support it but will leave the issue open until we find SMO can support it.

thewaiting28:
Thank you!

DorBreger:
Hi, what is the status on this? It seems that the SMO does support contained AGs. It's just not documented as of yet.
https://github.com/microsoft/sqlmanagementobjects/issues/154
Would really love to see this feature implemented.

serenefiresiren:
I have a solution for this I developed and tested for my own environment. I'll get that posted shortly.

DorBreger:
@thewaiting28 looks like you can stop waiting

PR #9252, opened 2024-02-22
Install-DbaMaintenanceSolution does not work inside a container

david-garcia-garcia:
## Please read -- recent changes to our repo
On November 10, 2022, [we removed some bloat from our repository (for the second and final time)](https://github.com/dataplat/dbatools/issues/8542). This change requires that all contributors reclone or refork their repo.

PRs from repos that have not been recently reforked or recloned will be closed and @potatoqualitee will cherry-pick your commits and open a new PR with your changes.

 - [X] Please confirm you have the smaller repo (85MB .git directory vs 275MB or 110MB or 185MB .git directory)

## Type of Change
 [X] Bug fix (non-breaking change, fixes # )
 
### Purpose
Fix internet connection check inside windows containers

### Approach
Remove uneeded internet connection check prior to download attempt. Current implementation already does fallback to embedded maintenance solution if download fails, so there is no need pre-check.

Inside a windows container, see output of current commands used to pre-check if an internet connection is available.

```
PS C:\> Get-NetConnectionProfile
Get-NetConnectionProfile : A general error occurred that is not covered by a 
more specific error code.
At line:1 char:1
+ Get-NetConnectionProfile
+ ~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (MSFT_NetConnectionProfile:root/St 
   andardCi...nnectionProfile) [Get-NetConnectionProfile], CimException
    + FullyQualifiedErrorId : MI RESULT 1,Get-NetConnectionProfile
 
PS C:\> $network = [Type]::GetTypeFromCLSID([Guid]"{DCB00C01-570F-4A9B-8D69-199FDBA5723B}")
PS C:\> 
      >                     $script:internet = ([Activator]::CreateInstance($network)).GetNetworkConnections() | ForEach-Object { $_.GetNetwork().GetConnectivity() } | Where-Object { ($_ -band 64) -eq 64 }
The service cannot be started, either because it is disabled or because it has 
    + CategoryInfo          : OperationStopped: (:) [], COMException~~~~~~~     
PS C:\>ullyQualifiedErrorId : System.Runtime.InteropServices.COMException       
+ ...             $script:internet = ([Activator]::CreateInstance($network) ... 
```
### Commands to test
Install-DbaMaintenanceSolution

PR #9345, opened 2024-05-11
New Support for Contained Availability Groups

serenefiresiren:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality, fixes [#8713 ](https://github.com/dataplat/dbatools/issues/8713))
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
SQL Server 2022 added support for contained availability groups. 

### Approach
The New-DbaAvailabilityGroup leverages the built-in ReuseSystemDatabses and IsContained parameters to facilitate both the initial build and a recreation with previously provisioned contained databases.

### Commands to test
Work in progress

### Learning
[What is a contained availability group?](https://learn.microsoft.com/en-us/sql/database-engine/availability-groups/windows/contained-availability-groups-overview?view=sql-server-ver16)
[New-SQLAvailabilityGroup](https://learn.microsoft.com/en-us/powershell/module/sqlserver/new-sqlavailabilitygroup?view=sqlserver-ps)
[SMOObject](https://github.com/microsoft/sqlmanagementobjects/blob/main/src/Microsoft/SqlServer/Management/Smo/AvailabilityGroupBase.cs)


public/New-DbaAvailabilityGroup.ps1
```diff
@@ -1,4 +1,4 @@
-function New-DbaAvailabilityGroup {
```
andreasjordan, line 1: I don't see a change here so there might be a problem with the encoding or the BOM. Please make sure that this line is not part of the diff.

public/New-DbaAvailabilityGroup.ps1
```diff
@@ -56,6 +56,14 @@ function New-DbaAvailabilityGroup {
     .PARAMETER Name
         The name of the Availability Group.
 
+    .PARAMETER IsContained
+        Builds the Availability Group as contained. Only supported in SQL Server 2022 or higher.
+
+        https://learn.microsoft.com/en-us/sql/database-engine/availability-groups/windows/contained-availability-groups-overview?view=sql-server-ver16
```
andreasjordan, line 62: I would exclude the "?view=sql-server-ver16" to always point to the latest version.

public/New-DbaAvailabilityGroup.ps1
```diff
@@ -366,6 +381,14 @@ function New-DbaAvailabilityGroup {
             return
         }
 
+        if ($IsContained -and $server.VersionMajor -lt 16) {
+           Stop-Function -Level Warning -Message "Contained availability groups are only supported in SQL Server 2022 and above"
+        }
+
+        if ($ReuseSystemDatabases -and $IsContained -eq $false) {
+           Write-Message -Level Warning -Message "Reuse system databases is only applicable in contained availability groups test "
```
andreasjordan, line 389: What does "test" mean? And there is a space at the end of the string.

public/New-DbaAvailabilityGroup.ps1
```diff
@@ -56,6 +56,14 @@ function New-DbaAvailabilityGroup {
     .PARAMETER Name
         The name of the Availability Group.
 
+    .PARAMETER IsContained
+        Builds the Availability Group as contained. Only supported in SQL Server 2022 or higher.
+
+        https://learn.microsoft.com/en-us/sql/database-engine/availability-groups/windows/contained-availability-groups-overview?view=sql-server-ver16
+
+    .PARAMETER ReuseSystemDatabases
+        Used when rebuilding a cluster where system databases already exist for the contained availability group.
```
andreasjordan, line 65: I would not use the term "cluster" here. Quote from the documentation: "This option is only valid for contained AGs, and specifies that the newly created AG should reuse existing contained system databases for a previous contained AG of the same name. For example, if you had a contained AG with the name MyContainedAG, and wanted to drop and recreate it, you could use this option to reuse the contents of the original contained system databases."

public/New-DbaAvailabilityGroup.ps1
```diff
@@ -219,6 +227,11 @@ function New-DbaAvailabilityGroup {
 
         Creates a basic availability group named BAG1 on sql2016std and does not confirm when setting up
 
+    .EXAMPLE
+        PS C:\> New-DbaAvailabilityGroup -Primary sql2022n01 -Secondary sql2022n02 -Name AgContained -IsContained
+
+        Creates a contained availability group named AgContained on sql2022
```
andreasjordan, line 233: Instead of "sql2022" (which is your cluster name?) we should name the nodes: "sql2022n01 and sql2022n02".

public/New-DbaAvailabilityGroup.ps1
```diff
@@ -488,6 +511,7 @@ function New-DbaAvailabilityGroup {
                 $ag.AutomatedBackupPreference = [Microsoft.SqlServer.Management.Smo.AvailabilityGroupAutomatedBackupPreference]::$AutomatedBackupPreference
                 $ag.FailureConditionLevel = [Microsoft.SqlServer.Management.Smo.AvailabilityGroupFailureConditionLevel]::$FailureConditionLevel
                 $ag.HealthCheckTimeout = $HealthCheckTimeout
+                $ag.ReuseSystemDatabases = $ReuseSystemDatabases
```
andreasjordan, line 514: Maybe it is saver (and better to read) if we move this line after the new line 527 ($ag.IsContained = $IsContained). This way, the two new parameters are used only on 2022 and newer. This way, the code does not change for older versions of sql server.

tests/New-DbaAvailabilityGroup.Tests.ps1
```diff
@@ -5,7 +5,7 @@ Write-Host -Object "Running $PSCommandpath" -ForegroundColor Cyan
...
+        [object[]]$knownParameters = 'Primary', 'PrimarySqlCredential', 'Secondary', 'SecondarySqlCredential', 'Name', 'DtcSupport', 'ClusterType', 'AutomatedBackupPreference', 'FailureConditionLevel', 'HealthCheckTimeout', 'Basic', 'DatabaseHealthTrigger', 'Passthru', 'Database', 'SharedPath', 'UseLastBackup', 'Force', 'AvailabilityMode', 'FailoverMode', 'BackupPriority', 'ConnectionModeInPrimaryRole', 'ConnectionModeInSecondaryRole', 'SeedingMode', 'Endpoint', 'EndpointUrl', 'Certificate', 'ConfigureXESession', 'IPAddress', 'SubnetMask', 'Port', 'Dhcp', 'EnableException', 'ReuseSystemDatabases', 'IsContained'
```
andreasjordan, line 8: I would like to have the order of parameters in this list in the order of the param block. So after "name" in this case.

andreasjordan (changes requested):
We are on a good way and I don't see a risk of breaking functionallity. But I would like to have some things changed before we merge.

serenefiresiren:
Thanks for the feedback! I'll get these fixed Monday.

potatoqualitee:
Thank you both 🙏🏼  Will fix encoding issue once changes are complete, just prior to merge.

DorBreger:
https://github.com/serenefiresiren/dbatools/pull/2

public/New-DbaAvailabilityGroup.ps1
```diff
@@ -366,6 +381,14 @@ function New-DbaAvailabilityGroup {
             return
         }
 
+        if ($IsContained -and $server.VersionMajor -lt 16) {
+           Stop-Function -Level Warning -Message "Contained availability groups are only supported in SQL Server 2022 and above"
```
andreasjordan, line 385: Aa you can see some lines above, we need an additional "return" in case Stop-Function only issues a warning and not an exception (this depends on -EnableException). And please remove the `-Level Warning`. But you should add the `-Target $Primary` so that all the tests are written the same way.

public/New-DbaAvailabilityGroup.ps1
```diff
@@ -366,6 +381,14 @@ function New-DbaAvailabilityGroup {
...
+        }
+
+        if ($ReuseSystemDatabases -and $IsContained -eq $false) {
+           Write-Message -Level Warning -Message "Reuse system databases is only applicable in contained availability groups"
```
andreasjordan, line 389: I would handle this the same way like the other test: `Stop-Function` and `return`.

andreasjordan (changes requested):
Sorry, I have not seen this in the first review. Please align the new parameter tests with the existing parameter tests.

andreasjordan:
Sorry, still not correct. Please have a look at the other parameter tests.

andreasjordan:
#9252 has all the needed changes, so we should close this one.

potatoqualitee:
thank you all 🙏🏼  will merge for credit and fix all issues in next PR

PR #9370, opened 2024-05-28
Support OwnerLoginName in New-DbaAgentSchedule

david-garcia-garcia:
## Type of Change

 - [x] New feature (non-breaking change )
### Purpose
It is not possible to create a schedule inside a container when interacting with MSSQL using Integrated Authentication and the user is the default ContainerAdmin:

```
C:\entrypoint\refreshenv\ConfigureMssqlJobs.ps1 : The specified
mssql-1  | '@owner_login_name' is invalid (valid values are returned by sp_helplogins
mssql-1  | [excluding Windows NT groups]).
```

It looks like this default account has some sort of trait that prevents it from being the owner of a Schedule. You can create jobs using other users as owners.

### Approach
Allow to manually set the Ownership of the schedule during creation.


public/New-DbaAgentSchedule.ps1
```diff
@@ -99,6 +99,9 @@ function New-DbaAgentSchedule {
 
         If force is used the start time will be '23:59:59'
 
+    .PARAMETER OwnerLoginName
```
andreasjordan, line 102: Most Owner parameter are named "Owner":

image

Only Set-DbaAgentJob uses OwnerLogin, none uses OwnerLoginName. I would vote for "Owner" as the parameter name.

And you need to add the new parameter also to the test file.

public/New-DbaAgentSchedule.ps1
```diff
@@ -523,6 +527,10 @@ function New-DbaAgentSchedule {
                         $jobschedule.ActiveEndTimeOfDay = $EndTime
                     }
 
+                    if ($OwnerLoginName) {
+                        $jobschedule.OwnerLoginName = $OwnerLoginName;
```
andreasjordan, line 531: We do not use ";" at the end of lines.

andreasjordan (changes requested):
Looks good, but I have not tested the code.

public/New-DbaAgentSchedule.ps1
```diff
@@ -99,6 +99,9 @@ function New-DbaAgentSchedule {
 
         If force is used the start time will be '23:59:59'
 
+    .PARAMETER OwnerLoginName
+        The name of the login that owns the job. login is sysname, with a default of NULL, which is interpreted as the current login name.
```
wsmelton, line 103:
```suggestion
        Login to own the job, defaults to login running the command.
```

wsmelton (changes requested):
Shorten description a bit to be more to the point

public/New-DbaAgentSchedule.ps1
andreasjordan: I like that.

public/New-DbaAgentSchedule.ps1
david-garcia-garcia: @andreasjordan 

> And you need to add the new parameter also to the test file. 

There are several test that directly or indirectly test New-DbaAgentSchedule, do you mean adding specific test coverage for the Owner parameter, or just populating it in one of the existing tests so it gets indirectly tested?

andreasjordan: You need to add the new parameter to this list (between EndTime and Force) - or this test will fail: 
https://github.com/dataplat/dbatools/blob/78c1bc392985541999949f76d2738b316e41a534/tests/New-DbaAgentSchedule.Tests.ps1#L9

You could also add a new test that uses the parameter. But I think that is not needed.

After you have changed the test file, I will activate the tests for your PR.

PR #9372, opened 2024-05-28
Install-DbaFirstResponderKit - Use new script names

andreasjordan:
Brent has made some changes and removed some of the scripts.


wsmelton:
Do we know if his script name changes will handle cleaning up objects for users? I wouldn't want someone to go install this and find data history is broken because table names or things changed. (I don't use his stuff).

Based on your commit are we changing how PRs are submitted now to only have branch tests running the single command?

andreasjordan:
The old scripts can still be used and are only moved to the deprecated folder and not maintained anymore. So no need to do a cleanup.

About the testing: We use this to reduce the cost in AppVayor - we do this all the time.

potatoqualitee:
Thank you 🙏🏼

PR #9397, opened 2024-06-16
Get-DbatoolsError: Use FullyQualifiedErrorId to filter $global:error

andreasjordan:
## Type of Change

 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

Maybe it's only my system, but the ScriptStackTrace does not contain "dbatools". I think it's because the commands are not imported from a file but a compressed file.

image

But "dbatools" is present in the FullyQualifiedErrorId. So we should use this.
image

Or is there a better way to filter?

### DIFF
--- a/public/Get-DbatoolsError.ps1
+++ b/public/Get-DbatoolsError.ps1
@@ -62,6 +62,6 @@ function Get-DbatoolsError {
             $First = $global:error.Count
         }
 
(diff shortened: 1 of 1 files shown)


andreasjordan:
The test passed in the past because of the way the tests import the module: They use the file by file method, so the filename and thus "dbatools" is part of the stack trace...

potatoqualitee:
ah ha! appreciate the pr

PR #9446, opened 2024-08-12
Write-DbaDbTableData - Add AutoCreateTableIdentity column

potatoqualitee:
I need this to make a blog post sleeker and figured others would appreciate it as well. Helps when imported related data that requires two tables.


potatoqualitee:
Hmm wait, I forgot that its bulk inserted data. Not so sure this is as useful as I originally hoped.

PR #9498, opened 2024-10-11
Refactor AgHadr tests

andreasjordan:

Currently `$script:instance3` is used for Availability Group tests and needs Hadr service setting enabled. This is done for AppVeyor in appveyor.SQL2027.ps1 with `$null = Enable-DbaAgHadr -SqlInstance $sqlinstance -Confirm:$false -Force`.

So every test can assume that the instance is configured correctly and has to end the test with a correctly configured instance.

Currently we don't see problems in AppVeyor because Disable-DbaAgHadr.Tests.ps1 is run as last test. But to be able to test locally we need to fix this.

### DIFF
--- a/tests/Disable-DbaAgHadr.Tests.ps1
+++ b/tests/Disable-DbaAgHadr.Tests.ps1
@@ -13,8 +13,10 @@ Describe "$CommandName Unit Tests" -Tag "UnitTests" {
     }
 }
 
+# $script:instance3 is used for Availability Group tests and needs Hadr service setting enabled
+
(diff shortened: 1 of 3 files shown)


andreasjordan:
Test for `Get-DbaAgHadr` failes because it still runs in the group for $script:instance2, but I think it should run against $script:instanxce3 together with the related tests.

@niphlod Can you help here?

andreasjordan:
Ok, solved the problem...

potatoqualitee:
thank you 🙇🏼

PR #9499, opened 2024-10-11
Fix test for Get-DbaDump

andreasjordan:

I don't know why, but creating a second stackdump immediatly triggers the visibility of the stackdumps to Get-DbaDump.

That is much more saver and faster than waiting.


potatoqualitee:
sounds good thank you!

PR #9501, opened 2024-10-11
Refactor tests to use variable server instead of instance for SMO server objects (part 1)

andreasjordan:

Some tests use code like `$instance2 = Connect-DbaInstance -SqlInstance $script:instance2`. Using `$instance2` and `$script:instance2`. I think this should be avoided and we should use the variable $server like in the main code.


potatoqualitee:
Thank you very much 🙇🏼

PR #9503, opened 2024-10-12
Enable test for Invoke-DbaDbMirroring

andreasjordan:

By using different ports for the mirroring endpoint, this test can run on AppVeyor.


potatoqualitee:
niiice! we will have to keep an eye out bc some tests are disabled after intermittent failures but i havent messed with them in years so hopefully its improved.

PR #9506, opened 2024-10-14
Get-DbaDbMailServer: Fix e-mail account names if more than one account exists

### LINKED ISSUE #9496: Get-DbaDbMailServer shows incorrect e-mail account names if more than one account exists.

No error. The command output is incorrect if more than one e-mail accounts are defined.

### Steps to Reproduce

```powershell
# Need to have a SQL server with more than one e-mail account defined
get-dbaDbMailServer -SqlInstance instancename
# Result:
# "Account" field is the same for all resulting entries. All other fields are correct.
```

2.1.23

### Other details or mentions

_No response_

### PowerShell Host Version

PSVersion                      5.1.19041.3693                                                                                        
BuildVersion                   10.0.19041.3693                                                                                       

Microsoft SQL Server 2008 (SP4) - 10.0.6000.29 (X64)   Sep  3 2014 04:11:34   Copyright (c) 1988-2008 Microsoft Corporation  Standard Edition (64-bit) on Windows NT 6.0 <X64> (Build 6002: Service Pack 2) 

### .NET Framework Version

.NET Framework 4.8.4739.0

potatoqualitee:
Redo of this PR with smaller git repo: https://github.com/dataplat/dbatools/pull/9497

Fix an error that results in all account names matching the first account instead of corresponding to real account names.

## Type of Change

 - [x] Bug fix (non-breaking change, fixes #9496 )

### Purpose
Fix an error that results in all account names matching the first account instead of corresponding to real account names.

### Approach
Each corresponding account name is used instead of the only using first account name for all results.

Note: I changed the coding slightly, as it seemed that -Force would keep overwriting the account names. So now it's an array. Please let me know if I need to take a different approach.

### DIFF
--- a/public/Get-DbaDbMailServer.ps1
+++ b/public/Get-DbaDbMailServer.ps1
@@ -97,7 +97,7 @@ function Get-DbaDbMailServer {
                     $servers | Add-Member -Force -MemberType NoteProperty -Name ComputerName -value $mailserver.ComputerName
                     $servers | Add-Member -Force -MemberType NoteProperty -Name InstanceName -value $mailserver.InstanceName
                     $servers | Add-Member -Force -MemberType NoteProperty -Name SqlInstance -value $mailserver.SqlInstance
-                    $servers | Add-Member -Force -MemberType NoteProperty -Name Account -value $servers[0].Parent.Name
+                    $servers | Add-Member -Force -MemberType NoteProperty -Name Account -value $servers.Parent.Name
                     $servers | Select-DefaultView -Property ComputerName, InstanceName, SqlInstance, Account, Name, Port, EnableSsl, ServerType, UserName, UseDefaultCredentials, NoCredentialChange
                 }
             } catch {


andreasjordan:
I think your solution has kind of the same problem as the old code: There could be different mailservers on different instances all handled in this code. The Account property needs to be from the specific mailserver, that's why the foreach is needed.

```
$servers | ForEach-Object {
                       $_ | Add-Member -Force -MemberType NoteProperty -Name Account -value $_.Parent.Name
                    }
```

potatoqualitee:
but won't -Force just keep overwriting it and you only get the last one? with mine, an array is returned? i suppose i should test this but i got a lot to do 😅

andreasjordan:
No, -Force is just for the case that the same SMO runs again through this command. In the foreach, every $server is only updated once with the correct value for this $server.

potatoqualitee:
ty!

andreasjordan:
Now you have merged your version without the foreach. That will not solve the problem.

PR #9522, opened 2024-10-19
Fix test for Get-DbaEstimatedCompletionTime

andreasjordan:
But the test will not work because the databae is too small and CHECKDB is to fast...

### DIFF
--- a/tests/Get-DbaEstimatedCompletionTime.Tests.ps1
+++ b/tests/Get-DbaEstimatedCompletionTime.Tests.ps1
@@ -16,17 +16,20 @@ Describe "$CommandName Unit Tests" -Tag 'UnitTests' {
 Describe "$commandname Integration Tests" -Tags "IntegrationTests" {
 
     BeforeAll {
+        $skip = $true
         if ($script:bigDatabaseBackup) {
-            if (-not (Test-Path -Path $script:bigDatabaseBackup) -and $script:bigDatabaseBackupSourceUrl) {
-                Write-Host 'Starting Download'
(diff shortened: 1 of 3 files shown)


andreasjordan:
I think we should skip this test on AppVeyor and only test this command from time to time on a local instance with a big database.

What do you think?

potatoqualitee:
Yes, we can skip 👍🏼

andreasjordan:
New idea for those tests: We could use a freely availably database backup for the tests. I refactored the test to use the Stackoverflow database:
https://github.com/BrentOzarULTD/Stack-Overflow-Database/releases/download/20230114/StackOverflowMini.bak

I also changed the BeforeAll to test new variables `$script:bigDatabaseBackup` and `$script:bigDatabaseBackupSourceUrl`. We could improve this and download and restore the database in `appveyor.SQL2016.ps1` and use it in more tests.

But currently this backup is made with a SQL Server 2022, so it can not be used with the current instance in AppVeyor. So we should disable the test on AppVeyor. Here is a screenshot from the AppVeyor log with the current code:

image

andreasjordan:
@potatoqualitee : If I change `pester.groups.p1` here in this branch, this might result in a merge conflict, because the file was changes by a recently merged pull request. Can you merge the current development branche into this branch?

potatoqualitee:
Happy to! btw dont worry about my aider branches, I am still working on my prompts and will start with a fresh branch once it's all done and if it takes a bit long, ill copy and rewriite your changes.

potatoqualitee:
lgtm!

PR #9523, opened 2024-10-19
Enabling more tests on AppVeyor

andreasjordan:
Let's test...


potatoqualitee:
These generally look like they'll pass for a while then fail for a while. Happy to try this again regularly, as many things change (.net patches, etc etc)

PR #9524, opened 2024-10-20
Tests - use new Test-Config instead of constants.ps1

potatoqualitee:
I gotta sleep but hopefully these tests will pass. This is to prep for the move.

With the new scoping rules in pester, I had to do a whole bunch of $global and didnt like it. So I made it into this.

I also included an example local vars file but devs can also just ask chatgpt to find/replace


potatoqualitee:
@andreasjordan or @niphlod pls feel free to fix whatever. I didnt set the warning to quiet again so we see a lot more right now. otherwise, ill continue working on it when i wake up

potatoqualitee:
@andreasjordan @niphlod @wsmelton pls take note

potatoqualitee:
Oh, and i'll fix that PSDefaultParameterValues by setting it at a level a couple up, but that'll come later. Right now, it'll actually be helpful to see all the warnings.

PR #9589, opened 2025-02-17
Test-DbaWindowsLogin, add property for mismatch on samaccountname

### LINKED ISSUE #9366: Requesting that Test-DbaWindowsLogin return a property for SamAccountName Mismatch
### Summarize Functionality

I'm requesting that Test-DbaWindowsLogin have an additional property added so that it returns a $true/$false value when it detects that the SamAccountName doesn't match the Login name.

The Test-DbaWindowsLogin command already detects this state and pops a nice WARNING about it, but there is no way to query the output of the command to tell that it was detected.

This would allow me to easily find any of these name mismatches in my environment so that I could flag them to be fixed.

Yes

### Technical Details

In our environment there are several servers with Logins whose names no longer match their SamAccountName in Active Directory. 

When I run Test-DbaWindowsLogin on one of these logins, I get a nice warning message saying
`WARNING: [12:35:35][Test-DbaWindowsLogin] SamAccountName mismatch detected for (account).`

Unfortunately, beyond the warning, I can't actually tell that there's anything wrong with the login from the output of the command. 

I'd like to request a property that's something like "SamAccountNameMismatch" with $true/$false value so that when I run Test-DbaWindowsLogin I can do something like:

`Test-DbaWindowsLogin -SqlInstance $AllSQLInstances | Where { $_.SamAccountNameMismatch -eq $true }`

and find all of the logins in my environment which don't align properly.

niphlod:
## Type of Change

 - [x] New feature (non-breaking change, adds functionality, fixes #9366 )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
Add a new property to be able to filter out mismatched logins

### DIFF
--- a/public/Test-DbaWindowsLogin.ps1
+++ b/public/Test-DbaWindowsLogin.ps1
@@ -170,6 +170,7 @@ function Test-DbaWindowsLogin {
             }
             Write-Message -Message "Parsing Login $adLogin." -Level Verbose
             $exists = $false
+            $samAccountNameMismatch = $false
             try {
                 $loginBinary = [byte[]]$winLogin.Sid
                 $SID = New-Object Security.Principal.SecurityIdentifier($loginBinary, 0)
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Thanks niph!

PR #9623, opened 2025-03-12
Get-DbaDbUser, faster enumeration when targeting single db

### LINKED ISSUE #9450: Export-DbaUser is much faster than Get-DbaUser, even though it does less work.

No error, but this doesn't feel like intended behaviour. If I want to get all users for a particular login, `Get` takes minutes. If I want to export the scripts to create them as well as their extra permissions, `Export` takes seconds. Logically, `Export` ought to require getting each user, so this mismatch in runtimes seems very strange.

### Steps to Reproduce

Just run both for a particular login on one of your boxes. No special parameters or steps needed. Don't specify a database. It might be worth mentioning that my instances all have around 50 databases.

Yes.

### Other details or mentions

_No response_

### PowerShell Host Version

5.1.19041.4648

I hit multiple, but mostly 2019.

### .NET Framework Version

Very recent.

niphlod:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #9450  )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
Get-DbaDatabase is marginally faster than SMO enumeration when targeting a single DB as it does a raw query to pre-filter databases.

Let's see if tests pass.

### DIFF
--- a/public/Get-DbaDbUser.ps1
+++ b/public/Get-DbaDbUser.ps1
@@ -103,15 +103,9 @@ function Get-DbaDbUser {
             } catch {
                 Stop-Function -Message "Failure" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
             }
-
-            $databases = $server.Databases | Where-Object IsAccessible
-
-            if ($Database) {
-                $databases = $databases | Where-Object Name -In $Database
-            }
-            if ($ExcludeDatabase) {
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Nice, thanks!

PR #9642, opened 2025-04-06
New-DbaDbMaskingConfig, account for lots of tables

### LINKED ISSUE #9639: New-DbaDbMaskingConfig - When using a table array filename gets too long and fails to save

I can't copy/paste the error specifically in this case but it was that the filepath at the end contained the entire list of tables in the filename and went past the windows 256 character limit even on the root C drive.

### Steps to Reproduce

Provide a list of tables as a paramter
-Table '','','','','',''

Each table gets added to the filename. The code either uses a small name if doing the whole database or uses a generated name based on all the tables is using the list of tables.

Yes

### Other details or mentions

If a path would be too long it should either be truncated or else changed to be like the name for the json that would be generated for doing the entire db.

### PowerShell Host Version

Latest

Latest

### .NET Framework Version

Latest

niphlod:
## Type of Change

 - [ ] Bug fix (non-breaking change, fixes #9639 )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
Don't generate an insanely long path for the config file.

### Approach
If tables reach a certain number, switch path generation to `Tables_<timestamp>` instead.

### DIFF
--- a/public/New-DbaDbMaskingConfig.ps1
+++ b/public/New-DbaDbMaskingConfig.ps1
@@ -609,7 +609,11 @@ function New-DbaDbMaskingConfig {
                     $filenamepart = $server.Name.Replace('\', '$').Replace('TCP:', '').Replace(',', '.')
 
                     if ($Table) {
-                        $temppath = Join-Path -Path $Path -ChildPath "$($filenamepart).$($db.Name).$($Table -join '-').DataMaskingConfig.json"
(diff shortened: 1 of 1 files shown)


potatoqualitee:
thanks, man. will relesae this with the new library if i can get it out this week

PR #9667, opened 2025-05-15
Invoke-DbaQuery, pass along appendconnectionstring only if needed

### LINKED ISSUE #9665: Getting warning using Connection object from Connect-DbaInstance with Invoke-DbaQuery

When I use the connection object created by `Connect-DbaInstance` in my call to `Invoke-DbaQuery`, I get the following message:

```WARNING: [15:46:13][Connect-DbaInstance] Additional parameters are passed in, but they will be ignored```

I have confirmed that it is actually the call to `Invoke-DbaQuery` that is raising this warning, not the call that creates the Connection object itself. I know that I can use the `-WarningAction SilentlyContinue`, but that seems like a kludge. Any suggestions? Examples 8 and onwards from the `Connect-DbaInstance` documentation will exhibit this behavior.

### Steps to Reproduce

```powershell
PS D:\Code\dbatoolkit> $Conn = Connect-DbaInstance -SqlInstance SQLServerAG01.domain.local -MultiSubnetFailover -TrustServerCertificate
PS D:\Code\dbatoolkit> Invoke-DbaQuery -SqlInstance $Conn -Database master -Query "SELECT DB_NAME()"
WARNING: [15:46:13][Connect-DbaInstance] Additional parameters are passed in, but they will be ignored

Column1
-------
master


```powershell
PS D:\Code\dbatoolkit> Get-Module dbatools

ModuleType Version    PreRelease Name                                ExportedCommands
---------- -------    ---------- ----                                ----------------
Script     2.1.30                dbatools                            {Select-DbaObject, Set-DbatoolsConfig, Add-DbaAgDatabase, Add

### Other details or mentions




PowerShell Core (pwsh.exe)

### PowerShell Host Version

```powershell
PS D:\Code\dbatoolkit> $PSVersionTable

PSVersion                      7.5.1
GitCommitId                    7.5.1
OS                             Microsoft Windows 10.0.14393

Microsoft SQL Server 2019 (RTM-CU32) (KB5054833) - 15.0.4430.1 (X64)   Feb 21 2025 17:28:26   Copyright (C) 2019 Microsoft Corporation  Standard Edition (64-bit) on Windows Server 2016 Datacenter 10.0 <X64> (Build 14393: ) (Hypervisor) 

### .NET Framework Version

.NET 9.0.4

### PR DESCRIPTION
**Pull request by niphlod**
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #9665  )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
Avoid passing down unuseful argument if not needed

### Approach
Detect via Test-Bound and if not needed, prune the splat before passing it down the line

### DIFF
--- a/public/Invoke-DbaQuery.ps1
+++ b/public/Invoke-DbaQuery.ps1
@@ -411,7 +411,7 @@ function Invoke-DbaQuery {
                 (-not $ReadOnly) -and # no readonly intent is requested and
                 (-not $Database -or $instance.InputObject.ConnectionContext.DatabaseName -eq $Database)  # the database is not set or the currently connected
                 if ($startedWithAnOpenConnection) {
-                    Write-Message -Level Debug -Message "Current connection can be reused"
+                    Write-Message -Level Debug -Message "Current connection could be reused"
                     # We got another nightmare to solve, but fortunately @andreasjordan is "the king"
                     # So. Here we are with a passed down "Server" instance. Problem is, we got two VERY different
                     # libraries built with VERY different agendas. And we want to get the best of both worlds.
@@ -453,15 +453,14 @@ function Invoke-DbaQuery {
                         $secStringPassword = ConvertTo-SecureString $instance.InputObject.ConnectionContext.ConnectAsUserPassword -AsPlainText -Force
                         [PSCredential]$serverCredentialFromSMO = New-Object System.Management.Automation.PSCredential($instance.InputObject.ConnectionContext.ConnectAsUserName, $secStringPassword)
                         $connDbaInstanceParams = @{
-                            SqlInstance            = $instance
+                            SqlInstance            = "$instance" # pass down "as a string" so it's forced to open a new one
                             SqlCredential          = $serverCredentialFromSMO
                             Database               = $Database
                             NonPooledConnection    = $true           # see #8491 for details, also #7725 is still relevant
                             Verbose                = $false
-                            AppendConnectionString = $AppendConnectionString
                         }
-                        if (Test-Bound -Not -ParameterName AppendConnectionString) {
-                            $connDbaInstanceParams.Remove('AppendConnectionString')
+                        if (Test-Bound -ParameterName AppendConnectionString) {
+                            $connDbaInstanceParams.AppendConnectionString = $AppendConnectionString
                         }
                         if ($ReadOnly) {
                             $connDbaInstanceParams.ApplicationIntent = "ReadOnly"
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by andreasjordan** [comment]
That would solve the issue if AppendConnectionString was not used, but when AppendConnectionString was used with a Server SMO, then it would still be ignored.

And I would remove it from the original splat and just add it if needed - like done with "ReadOnly".

**Comment by niphlod** [comment]
> And I would remove it from the original splat and just add it if needed - like done with "ReadOnly".
doesn't change the behaviour but let's do that if you prefer that syntax.

on the other end, frankly, all the problems here stems from a "design" problem that ... generate more problems than the ones they tried to solve. 
Invoke-DbaQuery got added lots of params that "change the connection" that are useful ONLY when a smo connection ISN'T passed down. 
Then we need to do lots of "combinatory ifs" to re-create properly the connection if it's not reusable. Point is that right now either we error out when something gets passed down WITH a live SMO connection, so people REALLY understand the implications, or we need to basically test for every parameter out there.

While reviewing this, I noticed that
```
$server = Connect-DbaInstance -SqlInstance 'server1' -sqlCredential $creds
Invoke-DbaQuery -SqlInstance $server -Database tempdb -Query "select suser_sname()"
```
is NOT using "alternate credentials" while
```
$server = Connect-DbaInstance -SqlInstance 'server1' -sqlCredential $creds
Invoke-DbaQuery -SqlInstance $server -Query "select suser_sname()"
```
is (because of a bugged logic with the recreation if db context is different).

Going to try to fix this as well but it's really, really, really hard to maintain without breaking.

niphlod:
@andreasjordan , see the latest push. At this point "string-ifying" should "force" the creation of a new connection and not a reuse.

andreasjordan:
I had some time today to have a deeper look.

We should change Connect-DbaInstance first, because there is the root cause.

"ConnectionContext.ConnectAsUserName" is only relevant when we change the database context with `$connContext = $connContext.GetDatabaseConnection($Database, $false)` inside of Connect-DbaInstance. All other changes to the connection context don't touch the security settings.

While investighating I found a bug: We test `if ($Database) {` in line 596, but change the context only if the database does not match. But later starting line 640 we use GetDatabaseConnection in any case if $Database is used and not only if it does not match.

We should move your code "We rebuild correct credentials from SMO informations" from Invoke-DbaQuery to Connect-DbaInstance to support even more commands.

I will open a pull request to show my proposed changes later today...

andreasjordan:
Ok, the bug was not a bug but only missing knowledge. GetDatabaseConnection is always needed then useing $Database, because copying the connection context removes the current database.

niphlod:
closing in favour of #9680

PR #9670, opened 2025-05-16
Added IsCdCEnabled to Get-DbaDatabase

### LINKED ISSUE #9669: Show if CDC is enabled in Get-DbaDatabase
### Summarize Functionality

I recently used `Get-DbaDatabase | Select *` and was surprised to see absolutely nothing about CDC. I could write the sys.databases query to find if CDC is enabled, but I was surprised to not see it in what I first tried. Dbatools seems to have no native way to check the status of CDC.

As discussed [a while ago](https://github.com/dataplat/dbatools/discussions/9581).

Yes

### Technical Details

Just add another field to `Get-DbaDatabase`, not necessarily in the default view.

ReeceGoding:
No attempt made to touch the Azure stuff.

## Type of Change

 - [X] New feature (non-breaking change, adds functionality, fixes #9669 )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose

See #9669.

### Approach

I did the first thing that came to mind and it immediately worked. I expected this to be much harder!

### Commands to test
`Get-DbaDatabase`

### Screenshots
image


public/Get-DbaDatabase.ps1
```diff
@@ -376,11 +376,12 @@ function Get-DbaDatabase {
                     Add-Member -Force -InputObject $db -MemberType NoteProperty -Name SqlInstance -Value $server.DomainInstanceName
                     Add-Member -Force -InputObject $db -MemberType NoteProperty -Name LastRead -Value $lastusedinfo.last_read
                     Add-Member -Force -InputObject $db -MemberType NoteProperty -Name LastWrite -Value $lastusedinfo.last_write
+                    Add-Member -Force -InputObject $db -MemberType NoteProperty -Name IsCdcEnabled -Value ($backed_info | Where-Object { $_.name -eq $db.name }).is_cdc_enabled
```
niphlod, line 379: should be -ceq instead of -eq (in other places we do use -ceq to support case-sensitive installations)

niphlod (dismissed):

niphlod:
ok, now the code is right but .... are we sure the column is there for every version ? 
Feature-wise, I think CDC is from 2008 onwards, and restricted to enterprise only till 2016sp1.

ReeceGoding:
My old copy of _SQL Server MVP Deep Dives Volume 1_ agrees that it was new in 2008.

ReeceGoding:
We already have an `if` branch for major version 8. I doubt that any of this at all works for version before 8. And to my knowledge, 2008 is version 10. So I guess that I just need to write another branch for version 9 and then I'm done? Do you agree?

andreasjordan:
Why don't we use the property?
https://learn.microsoft.com/en-us/dotnet/api/microsoft.sqlserver.management.smo.database.changetrackingenabled?view=sql-smo-172#microsoft-sqlserver-management-smo-database-changetrackingenabled

andreasjordan:
@ReeceGoding yes, you are right. I was on the wrong track...

andreasjordan:
Maybe we should wrap `$server.Query("SELECT name, state, SUSER_SNAME(owner_sid) AS [Owner], is_cdc_enabled FROM sys.databases")` with a try catch and use the old query in the catch block.

ReeceGoding:
@niphlod  @andreasjordan Better?

andreasjordan:
The code looks save now, thanks fo rthe change.

But we need to talk about the values. Should the property IsCdcEnabled be of type boolean? I can not test this at the moment, but your screenshot looks like it works on current versions, so the "bit" column type is converted to a boolean. But what about the old versions. The "0" in your query is likely just a number. Or should the query return $null as the feature is not availabe?

ReeceGoding:
> Should the property IsCdcEnabled be of type boolean? I can not test this at the moment, but your screenshot looks like it works on current versions, so the "bit" column type is converted to a boolean. But what about the old versions. The "0" in your query is likely just a number. Or should the query return $null as the feature is not availabe?

I'm afraid that I don't know and also lack any way to find out. Suggestions?

andreasjordan:
image

So `$server.Query("SELECT name, state, SUSER_SNAME(owner_sid) AS [Owner], 0 AS is_cdc_enabled FROM sys.databases")` returns a number, not a boolean. As the feature is not present in that version, I vote for returning $null. We can do that by just remove `, 0 AS is_cdc_enabled` from that line.

I think that is the last change request before I approve the PR....

ReeceGoding:
@andreasjordan Done?

andreasjordan (approved):
Thanks, looks good!

potatoqualitee:
oh wow, didnt realize this one was done! awesome. thanks everyone 🥳

PR #9672, opened 2025-05-17
Added AppendServerInstance to Invoke-DbaWhoIsActive.

### LINKED ISSUE #9540: Add -AppendServerInstance to Invoke-DbaWhoIsActive
### Summarize Functionality

`Invoke-DbaWhoIsActive` calls `Invoke-DbaQuery`, but cannot be called with the `-AppendServerInstance` flag like `Invoke-DbaQuery` can. Let's fix that. When you call `Invoke-DbaWhoIsActive` on a bunch of servers at once, it pays to know which server you're seeing.

Yes

### Technical Details

This should be really easy, given that `Invoke-DbaWhoIsActive` calls `Invoke-DbaQuery` directly and already has a bunch of things built in for controlling what parameters go there.

ReeceGoding:
## Type of Change

 - [X] New feature (non-breaking change, adds functionality, fixes #9540  )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [X] Documentation
 - [ ] Build system

### Purpose

When you call `Invoke-DbaWhoIsActive` on a bunch of servers at once, it pays to know which server you're seeing.

### Approach

The key was to realise that the bulk of the code in `Invoke-DbaWhoIsActive` is for SQL parameters, not PowerShell ones.

### Commands to test

I used `Invoke-DbaWhoIsActive -SqlInstance $ins -ShowSystemSpids -AppendServerInstance`. The glaring omission in my testing is that I only have one instance running locally. I'm already pushing my ancient laptop too hard to risk spinning up another.

### Screenshots

image

### Learning

`-AppendServerInstance:$AppendServerInstance` is a good trick that I didn't know previously. In fact, it concerns me. **Is the `EnableException` parameter in `Invoke-DbaWhoIsActive` dead code?**

### DIFF
--- a/public/Invoke-DbaWhoIsActive.ps1
+++ b/public/Invoke-DbaWhoIsActive.ps1
@@ -141,6 +141,9 @@ function Invoke-DbaWhoIsActive {
 
         PSObject output introduces overhead but adds flexibility for working with results: https://forums.powershell.org/t/dealing-with-dbnull/2328/2
 
+    .PARAMETER AppendServerInstance
+        If this switch is enabled, the SQL Server instance will be appended to PSObject and DataRow output.
+
     .PARAMETER WhatIf
         Shows what would happen if the command were to run. No actions are actually performed.
 
@@ -230,6 +233,7 @@ function Invoke-DbaWhoIsActive {
         [switch]$Help,
         [ValidateSet("DataSet", "DataTable", "DataRow", "PSObject")]
         [string]$As = "DataRow",
+        [switch]$AppendServerInstance,
         [switch]$EnableException
     )
     begin {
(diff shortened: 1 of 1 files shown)


niphlod:
Hi @ReeceGoding , when closing https://github.com/dataplat/dbatools/issues/9540 what I meant is "given it's doable in 4 lines let's not add yet another piece of code".
Changing code means also adding regression tests to make sure nothing breaks down the line and, frankly, the manpower required to support each and every small iteration of "new params" just adds weight without providing "meaningful" value.
I said this lots of times during the years here and in Slack: "dbatools is not dbasolutions".

andreasjordan:
I totally agree with @niphlod's point.

Just one general thought: If we have a command that takes `[DbaInstanceParameter[]]$SqlInstance` - an array of instances - the output should always include the originating instance. I think all of the the main (SMO based) commands do this by adding "ComputerName, InstanceName, SqlInstance" as the first three properies in the DefaultView. Maybe we should have a general look at this.

potatoqualitee:
@niphlod pls advise - do we close or does it need modifications?

niphlod:
I stick with my original idea of not adding something that can be added easily but I see @andreasjordan  point, too.

I'd say this can be closed (at the very least it still needs modifications to the accompanying test) and let's open an improvement to add ComputerName, InstanceName, SqlInstance in the output as @andreasjordan suggested.

potatoqualitee:
Thank you very much, niph. And thank you both @ReeceGoding and @andreasjordan 🙇🏼

PR #9713, opened 2025-07-22
New-DbaLogin - Fix bug in PasswordMustChange

andreasjordan:
## Type of Change

 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

image

When using both `-PasswordPolicy` and `-PasswordExpiration`, then `PasswordMustChange` was also set. `$currentPasswordMustChange` is `$null` in this case so that the test `$currentPasswordMustChange -ne $newLogin.MustChangePassword` was true but should not. When first testing for a non-null-value, everything works as expected.


potatoqualitee:
awesome! Thank you

PR #9724, opened 2025-07-26
New-DbaConnectionString - Fix issues with parameter EncryptConnection 

### LINKED ISSUE #9411: New-DbaConnectionString Multiple Errors/Issues

Main Error
Snag_71bf2dc

Even though it's throwing an error it's still creating the connection string.
Snag_71cad23

Any attempt to set values for TrustServerCertificate and EncryptConnection then force the use of a credential. 
Snag_71d9f44

Adding a value for SQLCredential causes another issue requiring ApplicationIntent to be required
Snag_71f39f8

Once all that is supplied we are back to the first error, but now the connection string exposes the username and password of the credential, since Integrated Security is no longer a valid option
Snag_72039e4

### Steps to Reproduce

New-DbaConnectionString -SqlInstance <ServerName>

2.1.18

### Other details or mentions

_No response_

PowerShell Core (pwsh.exe), Windows PowerShell (powershell.exe), VS Code (terminal)

### PowerShell Host Version

PSVersion                      7.4.3
GitCommitId                    7.4.3
OS                             Microsoft Windows 10.0.19045

Microsoft SQL Server 2022 (RTM-CU13) (KB5036432) - 16.0.4125.3 (X64)   May  1 2024 15:05:56   Copyright (C) 2022 Microsoft Corporation  Enterprise Edition: Core-based Licensing (64-bit) on Windows Server 2019 Standard 10.0 <X64> (Build 17763: ) (Hypervisor) 

### .NET Framework Version

.NET 8.0.6

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #9411  )
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

Just a first quick fix to see if that is what the user wants.

More refactoring needed to remove the legacy code path...

### DIFF
--- a/public/New-DbaConnectionString.ps1
+++ b/public/New-DbaConnectionString.ps1
@@ -196,7 +196,6 @@ function New-DbaConnectionString {
         [string]$ClientName = "custom connection",
         [int]$ConnectTimeout,
         [string]$Database,
-        [ValidateSet('Mandatory', 'Optional', 'Strict', 'True', 'False')]
         [switch]$EncryptConnection = (Get-DbatoolsConfigValue -FullName 'sql.connection.encrypt'),
         [string]$FailoverPartner,
         [switch]$IsActiveDirectoryUniversalAuth,
@@ -271,9 +270,6 @@ function New-DbaConnectionString {
                 if (Test-Bound -Not -ParameterName 'ConnectTimeout') {
                     $ConnectTimeout = ([Dataplat.Dbatools.Connection.ConnectionHost]::SqlConnectionTimeout)
                 }
-                if (Test-Bound -Not -ParameterName 'EncryptConnection') {
-                    $EncryptConnection = (Get-DbatoolsConfigValue -FullName 'sql.connection.encrypt')
-                }
                 if (Test-Bound -Not -ParameterName 'NetworkProtocol') {
                     $np = (Get-DbatoolsConfigValue -FullName 'sql.connection.protocol')
                     if ($np) {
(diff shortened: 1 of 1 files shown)


public/New-DbaConnectionString.ps1
```diff
@@ -196,7 +196,6 @@ function New-DbaConnectionString {
         [string]$ClientName = "custom connection",
         [int]$ConnectTimeout,
         [string]$Database,
-        [ValidateSet('Mandatory', 'Optional', 'Strict', 'True', 'False')]
         [switch]$EncryptConnection = (Get-DbatoolsConfigValue -FullName 'sql.connection.encrypt'),
```
potatoqualitee, line 199: old configs may return a non bool/switch. we should set it after, in the beginning/body

potatoqualitee, line 199: oh man, wait this is the problem 🤔

andreasjordan: `[switch]$EncryptConnection = (Get-DbatoolsConfigValue -FullName 'sql.connection.encrypt'),` is exactly the same line as in `Connect-DbaInstance`. We should have seen complaints if this would be a problem. We should definitly keep the parameter definition of these two commands as similar as possible.

andreasjordan: No, that is not a (technical) problem. When I refactored the command I did not remove this line by acident. But as it is a switch parameter, it does not have any effect. It's just a useless line. But if you look at the code and overlook the `[switch]` you may think that you have the choice here.

potatoqualitee:
Thank you, @andreasjordan ! LGTM.

PR #9727, opened 2025-07-29
Get-TestConfig - Cleanup code

andreasjordan:
Just cleaning the code a bit.

* Removing a Write-Host that produces output that is not relevant and only takes up space
* Removing a comment that I don't understand and think is useless.
* Removing code that does not work as intended, as Get-TestConfig is invoked by pester and not the test itself.


potatoqualitee:
I'm unsure about removing commands without being certain. I'll merge since tests pass but let's keep this PR in mind if issues pop up.

PR #9721, opened 2025-07-25
New test environment for tests in an Azure VM

andreasjordan:
I tried to understand and to optimize the test framework so that I can better run tests locally.

I also wanted to setup a "master" pester 5 test to have a blueprint to change more tests from pester 4 to 5.

There are a lot of changes combined in this PR now - sorry for that. I can split it up later.

Changes that needs discussion based on the comments I left in the files:
* private/testing/Get-TestConfig.ps1
* tests/Add-DbaAgDatabase.Tests.ps1

Changes primary related to my new environment:
* .gitignore
* tests/Environment/README.md
* tests/Environment/constants.local.ps1
* tests/Environment/install_local_instances.ps1
* tests/Environment/logs/README.md
* tests/Environment/run_local_tests.ps1
* tests/Environment/setup_repository.ps1
* tests/Environment/uninstall_local_instances.ps1
* tests/Environment/Invoke-DbaPester.ps1
* tests/Environment/TestTestfiles.Tests.ps1
* tests/Environment/TestEnvironment.Tests.ps1
* tests/constants.local.ps1.example

Changes that are just fixes in some tests:
* tests/Copy-DbaCredential.Tests.ps1
* tests/Copy-DbaEndpoint.Tests.ps1
* tests/Find-DbaStoredProcedure.Tests.ps1
* tests/Get-DbaConnection.Tests.ps1
* tests/Get-DbaCredential.Tests.ps1
* tests/Invoke-DbaDbMirroring.Tests.ps1
* tests/Invoke-DbaQuery.Tests.ps1
* tests/Measure-DbaBackupThroughput.Tests.ps1
* tests/New-DbaCredential.Tests.ps1
* tests/New-DbaEndpoint.Tests.ps1
* tests/Remove-DbaEndpoint.Tests.ps1
* tests/Test-DbaEndpoint.Tests.ps1

@potatoqualitee and @niphlod : Can we have a discussion about the changes?

PR #9729, opened 2025-07-29
A pester 5 test as a blueprint

andreasjordan:
I updated "tests/Add-DbaAgDatabase.Tests.ps1" to show my suggestion for a pester 5 test and explain it with comments.

### DIFF
--- a/tests/dbatools.Tests.ps1
+++ b/tests/dbatools.Tests.ps1
@@ -153,11 +153,11 @@ Describe "$ModuleName Tests missing" -Tag 'Tests' {
     Context "Every function should have tests" {
         foreach ($f in $functions) {
             It "$($f.basename) has a tests.ps1 file" {
-                Test-Path "tests\$($f.basename).tests.ps1" | Should Be $true
+                Test-Path "$ModulePath\tests\$($f.basename).tests.ps1" | Should Be $true
             }
-            If (Test-Path "tests\$($f.basename).tests.ps1") {
+            If (Test-Path "$ModulePath\tests\$($f.basename).tests.ps1") {
                 It "$($f.basename) has validate parameters unit test" {
-                    $testFile = Get-Content "tests\$($f.basename).Tests.ps1" -Raw
+                    $testFile = Get-Content "$ModulePath\tests\$($f.basename).Tests.ps1" -Raw
                     $hasValidation = $testFile -match 'Context "Validate parameters"' -or $testFile -match 'Context "Parameter validation"'
                     $hasValidation | Should -Be $true -Because "Test file must have parameter validation"
                 }


andreasjordan:
This test layout works in my bigger draft PR #9721 but failes here - I'm probably missing something.

I would like to get general feedback on the new layout for pester 5 tests from you both: @potatoqualitee / @niphlod 
What do you like? What do you dislike?

Looking at .aider\prompts the pester 5 tests were build by AI? Maybe AI can also help change them again?

I already started to write pester tests to test the layout of the test using AST (see #9721). We could use this like we use PSScriptAnalyser.

tests/Add-DbaAgDatabase.Tests.ps1
```diff
@@ -38,49 +47,130 @@ Describe "Add-DbaAgDatabase" -Tag "UnitTests" {
     }
 }
 
-Describe "Add-DbaAgDatabase" -Tag "IntegrationTests" {
+Describe "Add-DbaAgDatabase IntegrationTests" -Tag "IntegrationTests" {
```
potatoqualitee, line 50: It's already tagged as integration, i liked the cleanliness of the previous Describe

potatoqualitee:
I'll let @niphlod speak to the new test format. Yes! AI can be used to rewrite them, following your prescribed format. I have a bunch of credits to use and maybe 4.1 can shine where 4o did not. If not, 4.1-mini perhaps.

https://blog.netnerds.net/2024/10/aider-is-awesome/

tests/Add-DbaAgDatabase.Tests.ps1
andreasjordan: The tag is not included in the output. I'll make a screenshot next time I start my lab.

tests/Add-DbaAgDatabase.Tests.ps1
```diff
@@ -38,49 +47,130 @@ Describe "Add-DbaAgDatabase" -Tag "UnitTests" {
...
-        $agname = "dbatoolsci_addagdb_agroup"
-        $dbname = "dbatoolsci_addagdb_agroupdb"
-        $newdbname = "dbatoolsci_addag_agroupdb_2"
-        $server.Query("create database $dbname")
-        $backup = Get-DbaDatabase -SqlInstance $TestConfig.instance3 -Database $dbname | Backup-DbaDatabase
+        # We want to run all commands in the BeforeAll block with EnableException to ensure that the test fails if the setup fails.
+        $PSDefaultParameterValues['*-Dba*:EnableException'] = $true
+
+        # We remove all the calls to Get-DbaProcess and Stop-DbaProcess as they are not needed with correct tests.
+        #$null = Get-DbaProcess -SqlInstance $TestConfig.instance3 -Program 'dbatools PowerShell module - dbatools.io' | Stop-DbaProcess -WarningAction SilentlyContinue
```
niphlod, line 56: can be a "suggestion" but our tests are pretty much a bad bunch that needs to be tamed. getting and stopping shouldn't be the standard but also acceptable.

tests/Add-DbaAgDatabase.Tests.ps1
```diff
@@ -38,49 +47,130 @@ Describe "Add-DbaAgDatabase" -Tag "UnitTests" {
...
+
+        # Collect all the created files to be able to remove them ant the end.
+        $filesToRemove = @( )
+
+        # Connect to the instance(s) using Connect-DbaInstance only if needed.
+        # It is saver to use the instance name from $TestConfig to get a fresh new SMO every time.
```
niphlod, line 62: saFer ?

tests/Add-DbaAgDatabase.Tests.ps1
```diff
@@ -38,49 +47,130 @@ Describe "Add-DbaAgDatabase" -Tag "UnitTests" {
...
+        # In the future, we should also test with pre-opened connection.
+        # Use dbatools commands whenever possible, so that these are tested here as well.
```
niphlod, line 64: not sure of this one. Our "suite" is pretty much "1-file tests 1-function" so asking for "use dbatools, so if something ELSE crashes it gets spewed out in here" is not going to help pinpointing the cause.
It's true most of our tests are integration tests and not "unittests" but, again, I wouldn't even spell this out for the casual user... if it uses other functions it's good, but also if it uses some "direct queries" (also, usually, they're a bit faster)

tests/Add-DbaAgDatabase.Tests.ps1
```diff
@@ -38,49 +47,130 @@ Describe "Add-DbaAgDatabase" -Tag "UnitTests" {
...
+
+        # Explain what needs to be set up for the test:
+        # To add a database to an availablity group, we need an availability group and a database that has been backed up.
+        # For negative tests, we need a database without a backup and a non existing database.
+
+        # Set variables that we need in the tests.
+        # I don't see the point in prefixing them all with "dbatoolsci",
```
niphlod, line 71: someone may want to test specific functions with a "not superclean" environment. A prefix makes the artifacts generated readily identifyiable also at a later stage

tests/Add-DbaAgDatabase.Tests.ps1
```diff
@@ -38,49 +47,130 @@ Describe "Add-DbaAgDatabase" -Tag "UnitTests" {
...
+        $null = New-DbaAvailabilityGroup @splatNewAg
+
+        $null = New-DbaDatabase -SqlInstance $TestConfig.instance3 -Name $existingDbWithBackup
+        $backup = Backup-DbaDatabase -SqlInstance $TestConfig.instance3 -Database $existingDbWithBackup -Path $TestConfig.Temp
+        $filesToRemove += $backup.Path
+
+        $null = New-DbaDatabase -SqlInstance $TestConfig.instance3 -Name $existingDbWithoutBackup
+
+        # We want to run all commands outside of the BeforeAll block without EnableException to be able to test for specific warnings.
+        $PSDefaultParameterValues.Remove('*-Dba*:EnableException')
```
niphlod, line 95: this, and the above, are going to be readily forgotten. I'm not sure this is a correct approach vs forcing -Ea Stop everywhere vs let the test fail at the first stages after beforeall. Of course, the one here is the sanest approach but making sure that everything is up and adherent to this template may be harder.
Maybe we should just set the runner preference to stop, AND to enable exception if not otherwise specified ?

niphlod (commented):
all in all I don't see "huge differences" on how we always did tests, so some of this modifications look totally good to me.
When not sure about it I wrote a comment "line-by-line" but feel free to have a discussion on the "generals aspects" as well.

tests/Add-DbaAgDatabase.Tests.ps1
andreasjordan: I see a lot of `$server.Query('...')` which does not throw with a nice error message like `Invoke-DbaQuery` does. More than once it was difficult to get the root cause from the AppVeyor logs. So instead of `$server.Query("create database $dbName")` I would prefer `New-DbaDatabase -SqlInstance $server -Name $dbName`. And sometimes the `$server = Connect...` is just used for one `server.Query("create database ...`. These two lines could just be one.

tests/Add-DbaAgDatabase.Tests.ps1
andreasjordan: Ups, yes.

tests/Add-DbaAgDatabase.Tests.ps1
andreasjordan: It's mostly for the situation where someone writes a new test. That test should not include this line but work without. And I would like to remove these line to find those tests, that couse those problems. I already found some - and I like to find the others as well.

tests/Add-DbaAgDatabase.Tests.ps1
andreasjordan: I see the cons of that - but only in those "not superclean" environments. And I thought about that. And I came to the conclusion that it is super dangerous to run our tests in any other than a dedicated environment. And in case of a new environment we don't need individual but nice names.
But if I'm the only one on that side - I can totally live with "dbatoolsci_" and a lot of "..._$(Get-Random)".

tests/Add-DbaAgDatabase.Tests.ps1
andreasjordan: I already have a pester test based on the AST that tests that - can be included like PSScriptAnalyser.

I forgot where it was, but I'm sure I have seen this: A test (incorrectly) passes because the setup failed silently. So for me it is super important that we check if the test setup is successful.

In addition I would from time to time test that there is no warning. If you look at the output of a AppVeyor run - there are a lot of warning that nobody cares about. I think that is super dangerous .

tests/Add-DbaAgDatabase.Tests.ps1
```diff
@@ -38,49 +47,130 @@ Describe "Add-DbaAgDatabase" -Tag "UnitTests" {
...
-            ClusterType = "None"
+            Primary      = $TestConfig.instance3
+            Name         = $agName
+            ClusterType  = "None"
             FailoverMode = "Manual"
-            Database = $dbname
-            Confirm = $false
-            Certificate = "dbatoolsci_AGCert"
+            Certificate  = "dbatoolsci_AGCert"
         }
```
andreasjordan, line 85: I think splatting is super hard to read if the = are not aligned. The example for the AI was formatted this way. So maybe AI will help us here to correctly format all splats in all tests.

tests/Add-DbaAgDatabase.Tests.ps1
```diff
@@ -38,49 +47,130 @@ Describe "Add-DbaAgDatabase" -Tag "UnitTests" {
...
-} #$TestConfig.instance2 for appveyor
+
+    Context "When adding AG database that does not have a backup" {
+        BeforeAll {
+            $splatAddAgDb = @{
+                SqlInstance       = $TestConfig.instance3
+                AvailabilityGroup = $agName
+                Database          = $existingDbWithoutBackup
+                # As we don't want an output, we suppress the warning.
+                # But we can still test the warning because WarningVariable is set globally.
```
andreasjordan, line 140: What do you think about setting a WarningVariable globally? But we have to agree on the name. I sometimes use just `warn`, but I think `WarnVar` as a short version of `WarningVariable` is a good option.

tests/Add-DbaAgDatabase.Tests.ps1
```diff
@@ -38,49 +47,130 @@ Describe "Add-DbaAgDatabase" -Tag "UnitTests" {
...
+                WarningAction     = 'SilentlyContinue'
```
andreasjordan, line 141: And I would aks test developers to suppress the warning in case they expect one from the test. This way we expect no output from out tests at all.

tests/Add-DbaAgDatabase.Tests.ps1
potatoqualitee: sounds goodi like $warn but happy to have w/e standardaized

potatoqualitee:
happy to merge when ur happy

andreasjordan:
I would like to remove some of the comments in the next days and maybe refacture one or two commands to have some more examples. I'll give you a go once I'm happy.

tests/Add-DbaAgDatabase.Tests.ps1
andreasjordan: This is the output when the tag name is also included in the name of the Describe:
image

This is the output of both Describe blocks have the same name:
image

I think this is not so important and I see your point that it is redundant in the code. And with the following names of the Context blocks I think it is clear what type of test are grouped together.

tests/Add-DbaAgDatabase.Tests.ps1
```diff
@@ -38,49 +43,122 @@ Describe "Add-DbaAgDatabase" -Tag "UnitTests" {
...
+        Remove-Item -Path $filesToRemove
+
+        # As this is the last block we do not need to reset the $PSDefaultParameterValues.
     }
 
     Context "When adding AG database" {
+        # We use the BeforeAll to run the test itself.
+        # Results are saved in $results.
+        BeforeAll {
+            $splat = @{
```
andreasjordan, line 103: I just named all splatting hashtables $splat as the variable is only used once just after the definition. This way we don't need to come up with all the names for the splats.

andreasjordan:
Now I'm happy with the first two tests.

The first test in the list has more comments to help new contributors and ourselves to remember how and why we do things.

The second test can be used to copy some code to other tests.

andreasjordan:
It looks like tests fail because `$TestConfig = Get-TestConfig` is not set. I think it should be set somewhere in the appveyor files, I just have to find the correct spot. @niphlod can you help?

andreasjordan:
Ok, found it myself...

potatoqualitee:
Woo! Thank you andreas, I expect to use this soon.

PR #9732, opened 2025-07-29
Refactor the Endpoint related tests

andreasjordan:
In the beginning and in between of the tests there should not be any endpoint.

This way we don't have to test for existance and have cleaner tests that are more readable.

Added a test for Test-DbaEndpoint as it was missing.


potatoqualitee:
Thank you 🙇🏼  While I believe the issue was that some tests left some things behind (like a mirror test), that should indeed be cleaned up in the after, not in the before.

PR #9736, opened 2025-07-29
Get-DbaDbBackupHistory - Change warning to verbose for recovery fork message

potatoqualitee:
This happens all the time in my env with AGs and nearly everywhere else. It's ugly and if someone needs, they can slap a verbose and see what's up if they are having issues.

Updated the log level from Warning to Verbose for messages about multiple recovery forks in Get-DbaDbBackupHistory. This reduces unnecessary warning output when multiple recovery forks are detected but no last full backup is found.


potatoqualitee:
Thanks girrrrrrl 🙏🏼

PR #9738, opened 2025-07-29
Refactor some tests

andreasjordan:
Some more fixes from my test runs.

### DIFF
--- a/tests/Invoke-DbaDbMirroring.Tests.ps1
+++ b/tests/Invoke-DbaDbMirroring.Tests.ps1
@@ -15,23 +15,17 @@ Describe "$commandname Unit Tests" -Tag "UnitTests" {
 
 Describe "$commandname Integration Tests" -Tag "IntegrationTests" {
     BeforeAll {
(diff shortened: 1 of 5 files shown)


tests/Invoke-DbaDbMirroring.Tests.ps1
```diff
@@ -15,23 +15,17 @@ Describe "$commandname Unit Tests" -Tag "UnitTests" {
 
 Describe "$commandname Integration Tests" -Tag "IntegrationTests" {
     BeforeAll {
-        $null = Get-DbaProcess -SqlInstance $TestConfig.instance2 | Where-Object Program -Match dbatools | Stop-DbaProcess -Confirm:$false -WarningAction SilentlyContinue
```
potatoqualitee, line 18: these have been important previously due to resource constraints and killing processes was our only way to assure results. we should keep an eye out for future issues.

andreasjordan: In case we need this in a test - its the fault of some other test. The goal should be to identify the other test. I already identified some cases and in an "interactive" test environment like my Azure lab it is possible to find them.

PR #9745, opened 2025-08-02
Fix test for Copy-DbaDbMail

andreasjordan:
We currently use `-ForEach` wrong. The used variable needs to be set during discovery, but we use a values that is set during the BeforAll - which is part of the run phase and thus too late.
So I removed the block with `$command | Should -HaveParameter $PSItem` as it was never executed. I think we should only use this if we not only test for the parameter name but also the type and other properties - that is what this is made for.

I added the `$CommandName` back in like we have it in pester 4 to avoid copy-paste errors in the Describe blocks.

Have done some more refactoring to clean the code.

If we agree on that syle we would have to copy it to all the other pester 5 tests...

### DIFF
--- a/tests/Add-DbaAgDatabase.Tests.ps1
+++ b/tests/Add-DbaAgDatabase.Tests.ps1
@@ -1,6 +1,7 @@
-#Requires -Module @{ ModuleName="Pester"; ModuleVersion="5.0"}
+#Requires -Module @{ ModuleName="Pester"; ModuleVersion="5.0" }
 param(
     $ModuleName               = "dbatools",
+    $CommandName              = [System.IO.Path]::GetFileName($PSCommandPath.Replace('.Tests.ps1', '')),
     # $TestConfig has to be set outside of the tests by running: $TestConfig = Get-TestConfig
     # This will set $TestConfig.Defaults with the parameter defaults, including:
     # * Confirm = $false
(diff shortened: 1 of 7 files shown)


potatoqualitee:
this is one of the standard pester 5 tests that I submitted back in December. You'd have A LOT to update. Let's decide on a Pester 5 template with @niphlod and then standardize across the board using automation.

andreasjordan:
Maybe we can involve some pester expert that reviews our tests?

Some of the changes could be made with a search and replace script. Maybe I play with that idea today...

potatoqualitee:
Take this opportunity to learn command line AI! It does tiny tasks like find/replace very well. According to my blog, the smaller models were better for find/replace. I can setup 4.1-mini and DM you. BRB.

andreasjordan:
I used Roo Code to apply the template from Copy-DbaDbMail to some other pester 5 tests and one pester 4 test. Worked quite well.

But I will stop here and we should agree on a new pester 5 template first.

I will work on fixing some other tests that still fail in my lab...

potatoqualitee:
awesome, yes, i agree. i believe the instructions there now are what @niphlod and i agreed on back in the day, combined with Pester best practices. if we get perfect prompts and o1 is still good, I can use aider and o1 to exhaust our credits. Wait, it wont. lol

> OpenAI o1-preview is THE BEST at rewriting Pester v4 to Pester v5. It's wildly accurate but also wildly expensive. I'm talking like 70 cents per command. It also takes a very long time, averaging 40 seconds per command.

That was a year ago. Meaning at most, it should be $450 and ~8 hours.

PR #9747, opened 2025-08-05
Another round of fixes for tests

andreasjordan:
Mainly to suppress expected warning messages. Sometimes to remove created objects.

There are some more fixes to come, but we are near the end...

But I have noticed some tests that output warnings that indicate that we don't test what we want to test - so they need some special care. Will probably rewrite them in pester 5...


potatoqualitee:
love it thank you!

PR #9753, opened 2025-08-06
More tests in the new pester 5 design

andreasjordan:
All `Add-` tests have the new pester 5 design for param block (except for the `$CommandName`) and the UnitTest Describe bock.

The first tests have a good design, the later tests maybe need some more refactoring.

All non-`Add-` tests are just some more fixes that I included here...


potatoqualitee:
Thank you!

PR #9760, opened 2025-08-09
Refactor and enhance Pester tests batch

potatoqualitee:
Refactored Pester test files for several Find-* and Export-DbaXESessionTemplate commands to use param blocks, improved parameter validation, and modernized test structure. Added BeforeAll/AfterAll setup and teardown, improved assertions, and enhanced test coverage and clarity for integration and unit tests.

### DIFF
--- a/tests/Export-DbaXESessionTemplate.Tests.ps1
+++ b/tests/Export-DbaXESessionTemplate.Tests.ps1
@@ -1,31 +1,79 @@
-$CommandName = $MyInvocation.MyCommand.Name.Replace(".Tests.ps1", "")
+#Requires -Module @{ ModuleName="Pester"; ModuleVersion="5.0" }
+param(
+    $ModuleName  = "dbatools",
+    $CommandName = "Export-DbaXESessionTemplate",
+    $PSDefaultParameterValues = $TestConfig.Defaults
+)
+
 Write-Host -Object "Running $PSCommandPath" -ForegroundColor Cyan
 $global:TestConfig = Get-TestConfig
 
-Describe "$CommandName Unit Tests" -Tag 'UnitTests' {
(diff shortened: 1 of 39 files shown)


potatoqualitee:
Find-DbaAgentJob, Get-DbaAgentJobHistory, Get-DbaAgentLog, and New-DbaDbTransfer can't be automatically converted

andreasjordan (approved):
In a later PR we should globally change:

Remove:
```
Write-Host -Object "Running $PSCommandPath" -ForegroundColor Cyan
$global:TestConfig = Get-TestConfig
```

The output is already written by pester, the variable $TestConfig is already set outside of the test.


We should not use `TestDrive:\` from pester for backups to support SQL Server instances on other maschines later. For this to be posible, the backups need to be written to a fileshare where the test can remove them at the end.


Things like this should be rewritten to test for the value and not for $true:
```
$($results.CreateDate.ToFileTimeutc()[0]) -eq 126942668163900000 | Should -Be $true
```
It's way easier to analyze the problem if you know the incorrect value.


There are still some `Confirm     = $false` in splats and some `-Confirm:$false` that can be removed.


I have see this:
```
$results = @(Find-DbaDbDuplicateIndex @splatFind)
```
Don't know if I like this syntax. Should be ok without the "@()".


I've seen some ` -ErrorAction SilentlyContinue` with dbatools commands. That is not needed as they do not throw.

I've see al lot of `Should -Be $true` which could also be `Should -BeTrue` - what do we like?

So in the end I don't see anything that needs to be changed before merging to development. I think it easier to change all that at the end or if a first big bunch of tests are migrated to pester 5.

potatoqualitee:
thank you! 🙏🏼

PR #9765, opened 2025-08-09
Test which tests still fail

andreasjordan:
I wanted to test, which currently excluded tests still fail.

But currently no tets is running. Do you know, what the problem is, @potatoqualitee ?


andreasjordan:
Sorry, I destroyed the JSON file...

potatoqualitee:
thank you! i find that over the course of the next couple months, the issues will return (they are intermitten but for like 3 days straight, one specific command that hasn't been touched in years will fail. ill ping you if that happens cuz you're right, there's probably something else that can be fixed. thank you for doing this!

PR #9766, opened 2025-08-10
Cleanup AppVeyor configuration scripts

andreasjordan:
Some settings are not needed anymore because they are set by appveyor.prep.ps1.

I have aligned all three instance specific files to have a better overview of the needed settings in case a testlab shuld be setup in another environment.

Let's see if everything works as expected. Feedback is welcome.

### DIFF
--- a/tests/Read-DbaXEFile.Tests.ps1
+++ b/tests/Read-DbaXEFile.Tests.ps1
@@ -18,26 +18,28 @@ Describe "$CommandName Integration Tests" -Tags "IntegrationTests" {
     Context "Verifying command output" {
         BeforeAll {
             # Just to analyse problems on AppVeyor
-            $sessions = Get-DbaXESession -SqlInstance $TestConfig.instance3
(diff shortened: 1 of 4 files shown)


andreasjordan:
The test for Read-DbaXEFile failes because the command does not return any output.

This the test command:
```
$results = Get-DbaXESession -SqlInstance $TestConfig.instance2 -Session system_health | Read-DbaXEFile
```

If this keeps failing, I will add more output to the test to see which XESessions are running.

andreasjordan:
Get-DbaXESession is tested with:
```
$results = Get-DbaXESession -SqlInstance $TestConfig.instance2 -Session system_health
$results.Name -eq 'system_health' | Should Be $true
```

As this test passes, the session is there, but maybe not started...

andreasjordan:
Ok, now it works. The diff is not very readable - better look at the new files.

potatoqualitee:
Thanks, man. Looks good! 🏅

PR #9776, opened 2025-08-12
Move SqlPackage & XESmartTarget from bundled to installer-based dependencies

potatoqualitee:
This PR drops the bundled copies of SqlPackage and XESmartTarget and moves them to installer-based setup, per our #dbatools-dev discussion. This helps slim down the library and makes dependency management cleaner.

I’ll be coordinating this with the next library release, so it’ll be one of the last PRs I merge. Still need to tidy up a bit of code, but I also have to merge the tests now so I can update them without running into conflicts.


private/functions/Get-XESmartTargetPath.ps1
```diff
@@ -0,0 +1,86 @@
...
+        }
+    } catch {
+        Write-Message -Level Verbose -Message "Error checking for system-installed XESmartTarget: $($_.Exception.Message)"
+    }
+
+    # If we get here, XESmartTarget.Core.dll was not found
+    $message = @"
+Could not find XESmartTarget.Core.dll. XESmartTarget is no longer bundled with dbatools.
+
+To install XESmartTarget, use:
```
niphlod, line 73: we now have either Save-DbaCommunitySoftware OR Install-DbaXESmartTarget. IMHO we should just have one, and given we're taking the path of slimming down the library with external tools I'd go to leave just Install-DbaXESmartTarget

potatoqualitee:
your question is a reasonable one but i uhh, i dont know how it works but this is what my build script does and it always worked.

niphlod:
sure, but the build "happens" at the same time: you take dacfx plus sqlpackage from the internet "at the same time", so versions are probably matching. In here, one could download dbatools at 2025-08-01 and sqlpackage at 2025-09-01, while in the meantime a new sqlpackage version was published by MS. If this ends up in a dll mismatch, we're linking dbatools.library release schedule to external dependencies we don't control.

Again, *if need be* (if dlls loading errors DO occur with mismatched versions) we should support at least from the start one situation:
- "I need to update sqlpackage": user downloaded dbatools on 2025-08-01, sqlpackage on 2025-08-02 and *it worked* because they are matching. user updates dbatools on 2025-09-01, a new dbatools.library is released, too, with *newer* sqlpackage needed. user needs to be able to get the new sqlpackage that matches dlls shipped with dbatools.library. This works fine as long as the user "stays on the latest" and we release a new dbatools and a new dbatools.library as soon as external dependencies are updated

On the other end, just like dbatools is tied to a specific dbatools.library, external dependencies should be downloaded following a "pinned version" that links each dbatools/dbatools.library "combo" to a specific sqlpackage download.

andreasjordan:
I don't have time for testing this, sorry.

potatoqualitee:
Will pin if we get complaints 😊

PR #9781, opened 2025-08-15
Standardization of BeforeAll/AfterAll in tests

potatoqualitee:
```
If a Context block has only one "It" block and contains a BeforeAll, remove the BeforeAll and move the setup code directly into the It block. If a Context block has multiple "It" blocks, keep the BeforeAll as-is. If there iss no BeforeAll issue, make no changes.
```

Then

```
Get-ChildItem tests\*.Tests.ps1 | Invoke-Dbatoolsformatter
```


andreasjordan:
THIS WILL BREAK - please to not merge!!!

andreasjordan:
It block are independet - BeforeAll stuff is accessable in all it blocks.

potatoqualitee:
@niphlod suggested to remove BeforeAlls that do not have shared variables, which is what I instructed. If any are found to fail bc of that, it's a deviation.

andreasjordan:
Let's just look at this:


```
    Context "Parameter validation" {
        BeforeAll {
            $hasParameters = (Get-Command $CommandName).Parameters.Values.Name | Where-Object { $PSItem -notin ("WhatIf", "Confirm") }
            $expectedParameters = $TestConfig.CommonParameters
            $expectedParameters += @(
                "SqlInstance",
                "SqlCredential",
                "Database",
                "FileGroup",
                "FileGroupType",
                "InputObject",
                "EnableException"
            )
        }

        It "Should have the expected parameters" {
            Compare-Object -ReferenceObject $expectedParameters -DifferenceObject $hasParameters | Should -BeNullOrEmpty
        }
    }

```



The variables $...Parameters are set in BeforeAll and used in It.

That works, because every It block gets an individual copy of all variables from the BeforeAll block.

If we "rename" the BeforeAll into It - then there are two individual It blocks which can not share any variable. So the seecond It block gets empty variables.

potatoqualitee:
we are not renaming beforeall to it. we are using it, still. just when there are shared vars that require it.

image

you are more than welcomed to add it back in the future if you add more tests that need the vars.

PR #9795, opened 2025-08-20
Fix usage of EnableException  in pester 5 test

andreasjordan:
Pester 5 is sometimes reusing the same context between tests. So to be sure that nothing breaks, we DO reset the EnableException default parameter in every block where we set it. We can later use a test to make sure this is valid across all of the tests.

### DIFF
--- a/tests/Add-DbaAgDatabase.Tests.ps1
+++ b/tests/Add-DbaAgDatabase.Tests.ps1
@@ -81,7 +81,7 @@ Describe $CommandName -Tag IntegrationTests {
         # Remove the backup directory.
         Remove-Item -Path $backupPath -Recurse -ErrorAction SilentlyContinue
 
-        # As this is the last block we do not need to reset the $PSDefaultParameterValues.
+        $PSDefaultParameterValues.Remove("*-Dba*:EnableException")
     }
 
     Context "When adding AG database" {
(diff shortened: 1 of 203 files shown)


potatoqualitee:
hey @andreasjordan not sure which one i should have merged first but this one has a conflict. both changes look valid.

andreasjordan:
Should look like this:

```
    AfterAll {
        # We want to run all commands in the AfterAll block with EnableException to ensure that the test fails if the cleanup fails.
        $PSDefaultParameterValues["*-Dba*:EnableException"] = $true

        $null = Remove-DbaAvailabilityGroup -SqlInstance $server -AvailabilityGroup $agname
        $null = Get-DbaEndpoint -SqlInstance $server -Type DatabaseMirroring | Remove-DbaEndpoint
        $null = Remove-DbaDatabase -SqlInstance $server -Database $dbname

        # Remove the backup directory.
        Remove-Item -Path $backupPath -Recurse -ErrorAction SilentlyContinue

        $PSDefaultParameterValues.Remove("*-Dba*:EnableException")
    }
```

andreasjordan:
Ok, I manged to resolve....

potatoqualitee:
Thank you 🙏🏼

PR #9800, opened 2025-08-21
Test-DbaDiskAllocation: Use computername instead of ipaddress to support localhost

andreasjordan:
While running all our tests on a single Azure maschine the test for this command failed with:

Error connecting to [10.0.0.4\sqlinstance2]: Login failed. The login is from an untrusted domain and cannot be used with Integrated authentication.

The connection is made to the ip address. If someone can tell me why this is important then I can close this PR and just skip the test. But I think it is better to connect to the computername.


potatoqualitee:
🥳  ty

PR #9805, opened 2025-08-22
Get-DbaClientProtocol - remove wrong filter

andreasjordan:
## Type of Change

 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

Maybe it's generally in new versions, but on all of my SQL Server 2022 test instance, the ClientNetworkProtocol class is there but empty. In the changed line only the namespace should be selected as the following code just tests that. Then the command works both on systems with and without any configured client protocols. (And my local tests pass.)


potatoqualitee:
ty!

PR #9806, opened 2025-08-22
Rewrite test for Restore-DbaDatabase

andreasjordan:
That was hard work...

But:
image

There are still some skipped tests, see the comments for more details. 

We should completly recreate all the backups with current versions of SQL Server and document the way to produce them.


potatoqualitee:
thank you, @andreasjordan. not all backups to be replaced with 2022 version, right? we need a variety of backup types to ensure we can support a variety.

andreasjordan:
Sure, we should keep some of the old backups to test if we can still restore those old backups.

PR #9811, opened 2025-08-23
Remove InModuleScope from tests

andreasjordan:
I suspect that this option causes problems. I'm currently testing it in my lab and will provide feedback here.


andreasjordan:
Ok, looks like it's needed...

potatoqualitee:
huh that is strange. i would have guessed its not needed too

andreasjordan:
I will look into this later again and try to understand all that mocking...

PR #9815, opened 2025-08-25
Get-DbaDatabase - Add wildcard support for Database and ExcludeDatabase

### LINKED ISSUE #9814: Get-DbaDatabase -Database and -ExcludeUser parameters do not accept wildcards as documented

The documentation states that these parameters accept wildcards but do not state what the wildcards are such as % or *. I tried all permutations that I could think of, tried to find examples, and then went to the source code. It appears that these are not set to accept wildcards.
**Parameter Definitions**
.PARAMETER Database
 Specifies one or more databases to include in the results. Supports wildcards and exact name matching.
 Use this when you need to retrieve specific databases instead of all databases on the instance.

.PARAMETER ExcludeDatabase
 Specifies one or more databases to exclude from the results. Supports wildcards and exact name matching.
 Use this to filter out specific databases like test or staging environments from your inventory.

However, the source code shows the following pattern:
Where-Object {
                ($_.name -in $Database -or !$Database) -and
                ($_.name -notin $ExcludeDatabase -or !$ExcludeDatabase) -and
                ($_.Owner -in $Owner -or !$Owner) -and
                ($_.state -ne 6 -or !$OnlyAccessible)
            }
This would allow an array of exact match database names or having no database names passed in from what I can tell. It does not appear that you can pass any type of wildcard partial matches.

### Steps to Reproduce

With Database names such as dbatools_example1, dbatools_example2, dbatools_example3, dbatools_example4, dbatools_example5 the following commands will not return any results
```powershell
PS C:\> Get-DbaDatabase -SqlInstance SQL2,SQL3 -Database dbatools_example
```
```powershell
PS C:\> Get-DbaDatabase -SqlInstance SQL2,SQL3 -Database dbatools_example*
```
```powershell
PS C:\> Get-DbaDatabase -SqlInstance SQL2,SQL3 -Database dbatools_example%
```

I checked the most recent source code from here for verification: 
https://github.com/dataplat/dbatools/blob/development/public/Get-DbaDatabase.ps1

PS > find-psresource dbatools
Name     Version Prerelease Repository Description
----     ------- ---------- ---------- -----------
dbatools 2.7.2              PSGallery  The community module that enables SQL Server Pros to automate database develop…

PS > get-psresource dbatools
Name     Version Prerelease Repository Description
----     ------- ---------- ---------- -----------
dbatools 2.7.2              PSGallery  The community module that enables SQL Server Pros to automate database develop…

### Other details or mentions

_No response_

PowerShell Core (pwsh.exe)

### PowerShell Host Version

```
PS > $psversiontable

PSVersion                      7.5.2
GitCommitId                    7.5.2
OS                             Microsoft Windows 10.0.26120
```

SQL Server 2022 Build 16.0.4210.1
Issue is version agnostic

### .NET Framework Version

```
PS > [System.Runtime.InteropServices.RuntimeInformation]::get_FrameworkDescription()
.NET 9.0.6

PS > Get-ChildItem 'HKLM:\SOFTWARE\Microsoft\NET Framework Setup\NDP' -Recurse | Get-ItemProperty -Name version -EA 0 | Where PSChildName -Match '^(?!S)\p{L}' | Select PSChildName, version

PSChildName                      Version
-----------                      -------
v2.0.50727                       2.0.50727.4927
v3.0                             3.0.30729.4926
Windows Communication Foundation 3.0.4506.4926
Windows Presentation Foundation  3.0.6920.4902
v3.5                             3.5.30729.4926
Client                           4.8.09032
Full                             4.8.09032
Client                           4.0.0.0
```

potatoqualitee:
Fixes #9814 this should probably be standardized to all commands

@niphlod / @andreasjordan tell me if you like this pattern. if so, i can add repo-wide wildcard support and update the help.

### DIFF
--- a/tests/Get-DbaDatabase.Tests.ps1
+++ b/tests/Get-DbaDatabase.Tests.ps1
@@ -81,6 +81,92 @@ Describe $CommandName -Tag IntegrationTests {
             ($results).Count | Should -Be 1
         }
     }
+
+    Context "Wildcard functionality" {
+        BeforeAll {
+            $random = Get-Random
+            $dbname1 = "dbatoolsci_wildcard_test1_$random"
+            $dbname2 = "dbatoolsci_wildcard_test2_$random"
+            $dbname3 = "dbatoolsci_wildcard_example_$random"
+            $dbname4 = "dbatoolsci_wildcard_example2_$random"
+            $dbname5 = "dbatoolsci_exclude_test_$random"
+
+            # Create test databases
+            $null = New-DbaDatabase -SqlInstance $TestConfig.instance1 -Name $dbname1
+            $null = New-DbaDatabase -SqlInstance $TestConfig.instance1 -Name $dbname2
+            $null = New-DbaDatabase -SqlInstance $TestConfig.instance1 -Name $dbname3
+            $null = New-DbaDatabase -SqlInstance $TestConfig.instance1 -Name $dbname4
+            $null = New-DbaDatabase -SqlInstance $TestConfig.instance1 -Name $dbname5
+        }
+
+        AfterAll {
+            # Clean up test databases
+            Remove-DbaDatabase -SqlInstance $TestConfig.instance1 -Database $dbname1, $dbname2, $dbname3, $dbname4, $dbname5 -Confirm:$false -ErrorAction SilentlyContinue
+        }
+
+        It "supports * wildcard at the end of database name in Database parameter" {
+            $results = Get-DbaDatabase -SqlInstance $TestConfig.instance1 -Database "dbatoolsci_wildcard_example*"
+            $results.Count | Should -Be 2
+            $results.Name | Should -Contain $dbname3
+            $results.Name | Should -Contain $dbname4
+        }
+
+        It "supports * wildcard at the beginning of database name in Database parameter" {
+            $results = Get-DbaDatabase -SqlInstance $TestConfig.instance1 -Database "*_example_$random"
+            $results.Count | Should -Be 1
+            $results.Name | Should -Contain $dbname3
+        }
+
+        It "supports ? wildcard for single character matching in Database parameter" {
+            $results = Get-DbaDatabase -SqlInstance $TestConfig.instance1 -Database "dbatoolsci_wildcard_test?_$random"
+            $results.Count | Should -Be 2
+            $results.Name | Should -Contain $dbname1
+            $results.Name | Should -Contain $dbname2
+        }
+
+        It "combines exact matches and wildcards in Database parameter" {
+            $results = Get-DbaDatabase -SqlInstance $TestConfig.instance1 -Database "master", "dbatoolsci_wildcard_example*"
+            $results.Count | Should -Be 3  # master + 2 wildcard matches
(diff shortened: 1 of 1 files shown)


andreasjordan:
I like the idea. But... for all the commands? For all the parameters? Can we move the functionality to an internal function? Or have a code block that is exactly the same for all commands to fix bugs "one fix for all"? This would break in environments with wildcards in names. Do we add a switch like in Connect-DbaInstance some time ago when I changed the code?

I would not implement this without deep consideration.

potatoqualitee:
No no, sorry I wasn't clear. It would be all commands with -Database, -ExcludeDatabase, -Login, -Job and perhaps others. Not a bad idea to make it an internal command!

niphlod:
We had a similar discussion about "job". The end if I recall correctly was
"Get-" should always use literal matches (also to avoid breaking change),
if in need of wildcards/globs/whatever, put them in "Find-"

Il lun 25 ago 2025, 21:26 Chrissy LeMaire ***@***.***> ha
scritto:

potatoqualitee:
thank you @niphlod ! i may have agreed at the time, but i'd like to change my position 😅  I think if someone does a *, it's very intentional and they want a wildcard. that does remind me, that we'd have to ensure brackets on the outside arent processed. or maybe its okay if they are, i'd have to check. whatcha think?

niphlod:
Imho supporting -database and -excludedatabase just in get-dbadatabase
differently from all the other functions is a big no-no.
Another (not so big) no-no is backward compatibility.
Another (no so big) no-no is that if this gets added in Get-'s, what's the
point with Find-'s?

Then we get to what people intend by "wildcards"... Given we're in a
SQLServer world, [ and ] in t-sql and PowerShell "patterns" mean very
different things, and I don't know of they'll be clearer or not.

Users with "special filtering needs" can use a pretty simple "where-object"
already, so we're not leaving a big wide hole in functionality if we don't
implement patterns.

If all the above points can't sway "the Need to implement" we should
refactor pattern matching in a separate function, create a botload of
unittests, and then insert the relative function in ALL functions that have
a -database (and/or -excludedatabase). Then probably on -job and -login as
well, but the point could really be made for everything filtering out
things.

Il mar 26 ago 2025, 14:57 Chrissy LeMaire ***@***.***> ha
scritto:

mattcargile:
I would love `[WildcardPattern]` ( or `-like` ) in all the `Get-` much like `Get-Process` and `Get-Service` and many other commands. Having `Find-` for everything that supports `-match` would be nice too. For simple matches, it would help someone who is typically inside the shell and not using _SSMS_ as much. A project for aliases on the `-Parameter`s would be awesome too. Lastly, optimizing positional parameters would round out the interactive optimizations. So the most common types of executions would be isolated like `Get-DbaDatabase SqlInstance Database` should return `SqlInstance.Database`. I don't think having `-SqlCredential` as the second positional parameter is all that optimal.

potatoqualitee:
oh, a standard -Pattern. I like that.

potatoqualitee:
This one won't be it but once I've got less on my plate, I'll do `-Pattern`

PR #9819, opened 2025-08-26
Some pester improvements

andreasjordan:
Current goals:
* Enable most of the test (files) on AppVeyor so that the unit tests run and disable only those integration tests that fail.
* Mark those tests as to skip on powershell core that don't run on that platform.
* So every testfile should be generally runnable on any platform with at least the parameter tests running.


potatoqualitee:
Thank you 🙇🏼

PR #9820, opened 2025-08-26
Set-FileSystemSetting - Move from WMI to CIM to support pwsh

andreasjordan:
I found this running the test for New-DbaDbFileGroup, where a filestream filegroup is created. This uses WMI calls that are not available on pwsh - or better say: They are serialized objects because an internal winrm connection is used. So they don't have the original methods.

I completly refactored the command using our own internal command for cim calls: Get-DbaCmObject


potatoqualitee:
Thank you 🙏

PR #9825, opened 2025-08-29
Enhance tests

andreasjordan:
Main changes:
* Create a master key only once at the start to reduce the needed code and gain stability during tests.
* Add random number to databases so that we don't get messages about recover forks when a test runs more than once.

And some other minor fixes...


andreasjordan:
Now all tests run with powershell.exe and pwsh.exe on Windows Server 2022 with SQL Server 2022 on an Azure virtual maschine.

potatoqualitee:
nice and clean, i love it! 👍🏼

PR #9827, opened 2025-09-02
Copy-DbaDbAssembly - refactor test

andreasjordan:
This is the only test that needs this instance configuration, which is not the default (and probably not recommended) for new instances:

```
$null = Set-DbaSpConfigure -SqlInstance $TestConfig.instance2, $TestConfig.instance3 -Name IsSqlClrEnabled -Value $true
$null = Set-DbaSpConfigure -SqlInstance $TestConfig.instance2, $TestConfig.instance3 -Name ClrStrictSecurity -Value $false
```

So instead of running those commands at test lab setup, we run them only before the test and revert to standard settings after the test.

I also refactored the code a bit for better readablility.


potatoqualitee:
awesome ty

PR #9828, opened 2025-09-02
Refactor Filestream related tests

andreasjordan:
After we fixed the code so that it work with powershell.exe and pwsh.exe, we can now refactor the tests to have a better test coverage.

We are again assuming a newly installed instance where Filestream is not yet active.


potatoqualitee:
ty!

PR #9829, opened 2025-09-02
Fix test for Copy-DbaDatabase

andreasjordan:
-Continue only works if there are additional log backups (not diff backups). 

No idea if this has ever worked on pester 4...


potatoqualitee:
🥳

PR #9831, opened 2025-09-03
Copy-DbaDbCertificate - Fix command and test

andreasjordan:
Not the best solution, but it should work for now.

We need a new lab to test all the Copy- commands (and all commands related to AGs and Mirrring) with 4 maschines: 2 with the source and targe SQL Server instance, one from where we run dbatools and one with the fileshare. We then might find some more bugs...

### DIFF
--- a/public/Copy-DbaDbCertificate.ps1
+++ b/public/Copy-DbaDbCertificate.ps1
@@ -304,6 +304,20 @@ function Copy-DbaDbCertificate {
                                         $usingtempfiles = $true
                                     } else {
                                         $export = Backup-DbaDbCertificate @params
+
+                                        # The exported files are only readable by the source instance account
+                                        # But for the restore they need to be readable by the targe instance account
+                                        # Current solution is to try to make them readable to everyone and remove them after the restore
+                                        foreach ($filePath in $export.Path, $export.Key) {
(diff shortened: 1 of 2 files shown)


public/Copy-DbaDbCertificate.ps1
```diff
@@ -304,6 +304,20 @@ function Copy-DbaDbCertificate {
                                         $usingtempfiles = $true
                                     } else {
                                         $export = Backup-DbaDbCertificate @params
+
+                                        # The exported files are only readable by the source instance account
+                                        # But for the restore they need to be readable by the targe instance account
+                                        # Current solution is to try to make them readable to everyone and remove them after the restore
```
niphlod, line 310: isn't thw whole premise to use a share that has certs in there useful for disaster recovery scenario ? the deletion in here changes decisevely, as "before" it was just in case of errors (catch), while now it's in each and every case (finally)

andreasjordan: Only newly created cert files that where created by the backup part are deleted. If there are cert files present, then $usingtempfiles is true and the files are not deleted. In our tests we don't have files there so we can only test the Copy- (backup + restore) path.

PR #9832, opened 2025-09-03
Fix and enable some more tests

andreasjordan:
Just some small changes to some tests.

This PR can be open for a while, I might add some more...


andreasjordan:
I don't have much time this week - so this can be reviewed and merged.

PR #9834, opened 2025-09-05
Remove duplicate 'PSAvoidUsingWriteHost' rule from analyzer

slang25:
Just spotted this in my travels, presumably not causing any problems, but just thought I'd clean it up 🙂


potatoqualitee:
Awesome, thank you so much! 🙌🏼

PR #9838, opened 2025-09-11
Update tests for Copy-Dba* commands: Only copy from instance2 to instance3

andreasjordan:
I now have a lab on azure with a domain and a client and two sql servers to test the Copy-Dba* commands in a better scenario.

I changed the instances that are used in the tests so that we always copy from instance2 to instance3. This way I only need two instances in that lab.

All Copy-Dba* tests run successful in that lab.

### DIFF
--- a/tests/Copy-DbaBackupDevice.Tests.ps1
+++ b/tests/Copy-DbaBackupDevice.Tests.ps1
@@ -41,8 +41,8 @@ Describe $CommandName -Tag IntegrationTests {
         # Set variables. They are available in all the It blocks.
         $deviceName = "dbatoolsci-backupdevice-$(Get-Random)"
         $backupFileName = "$backupPath\$deviceName.bak"
-        $sourceServer = Connect-DbaInstance -SqlInstance $TestConfig.instance1
-        $destServer = Connect-DbaInstance -SqlInstance $TestConfig.instance2
+        $sourceServer = Connect-DbaInstance -SqlInstance $TestConfig.instance2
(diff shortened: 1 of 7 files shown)


niphlod:
Frankly I don't get it.
We're using instance names just to split tests between environments on
appveyor and let tests run in parallel.

To execute a full test suite the only thing needed is a VM with two
instances in it.

Il gio 11 set 2025, 18:23 Andreas Jordan ***@***.***> ha
scritto:

andreasjordan:
@niphlod You are right - this PR is (mostly) useless and I was just to lazy to setup three instances. I commented the only part that is not only about changing the instance to test.

I will probably close this in the next days...

andreasjordan:
@niphlod on Slack: "The old motto was given dbatools is often used for migration, to test at least the "first" version that is currently out of support (e.g. in order to make sure dbatools works for people having environments to migrate given a version is in the need to be upgraded to a newer one)."

My idea:

Let's have two more "pointers" to instances in the $TestConfig: `$TestConfig.SourceInstance` and `$TestConfig.DestinationInstance` (like we name the SqlInstance parameters in those commands) and only use them in all the Copy- tests. Then we can use the olderst instance as source and the newest as destination.

We could start with using instance2 (SQL2016) as source and instance3 (SQL2017) as destination. Every test should pass. Then we can change source to instance1 (SQL2008R2) and see if every test works - without changing the test, only the global test configuration.

potatoqualitee:
I am closing this unless I misunderstood then pls feel free to reopen

PR #9852, opened 2025-10-01
Fix Get-DbaSchemaChangeHistory when no trace file exists

SylvainDuran:
Adds a check to handle cases where the default system trace is not enabled. Previously, the function would fail when querying fn_trace_gettable with a null path. Now displays a warning message and skips the instance gracefully.

## Type of Change
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
Fix a bug where `Get-DbaSchemaChangeHistory` would crash with an error when executed on a SQL Server instance that has no default system trace enabled. This is a valid configuration scenario that should be handled gracefully.

### Approach
Added a validation check after retrieving the trace file path (line 97-100). If no trace file is found (`!$TraceFile -or !$TraceFile.Path`), the function now:
1. Displays a warning message explaining that the default trace is required for schema change tracking
2. Uses `continue` to skip to the next instance instead of crashing

### DIFF
--- a/public/Get-DbaSchemaChangeHistory.ps1
+++ b/public/Get-DbaSchemaChangeHistory.ps1
@@ -94,6 +94,11 @@ function Get-DbaSchemaChangeHistory {
 
             $TraceFile = $server.Query($TraceFileQuery) | Select-Object Path
 
+            if (!$TraceFile -or !$TraceFile.Path) {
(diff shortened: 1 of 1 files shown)


potatoqualitee:
thank you very much 💯

PR #9858, opened 2025-10-10
Add server name to removal message in Remove-DbaDbUser for WhatIf clarity

SylvainDuran:
## Type of Change

### Purpose
Improve clarity when using `Remove-DbaDbUser` with the `-WhatIf` switch: the server name is now explicitly shown in the main message indicating which instance is targeted for user removal. This ensures that, when running with `-WhatIf` or `-Confirm`, users can easily distinguish actions per target instance, making multi-instance operations clearer and safer.


potatoqualitee:
thank you, @SylvainDuran !

PR #9860, opened 2025-10-13
Correct retrieval of server name in Remove-DbaCredential

SylvainDuran:
## Type of Change
 - Bug fix
 
### Purpose
Fix the incorrect retrieval of the server name which was returning null by replacing $dbCredential.Parent.Parent.Name with $dbCredential.Parent.Name.


potatoqualitee:
Thanks man, 💯

PR #9874, opened 2025-10-26
Test-DbaLastBackup - Add Checksum parameter support

### LINKED ISSUE #9735: Add Checksum Flag for Test-DbaLastBackup
### Summarize Functionality

Add a `-Checksum` parameter to Test-DbaLastBackup which controls whether to include `WITH CHECKSUM` to the issued `RESTORE VERIFYONLY` command.

No

### Technical Details

Currently Test-DbaLastBackup does not offer any option to control whether the `RESTORE VERIFYONLY` command uses `WITH CHECKSUM` or `NO_CHECKSUM`.
As this parameter is omitted for the `RESTORE VERIFYONLY` command the default is applied, which means that backup checksums will be validated if they exist:

> The default behavior is to verify checksums if they are present and proceed without verification if they are not present.

Source: https://learn.microsoft.com/en-us/sql/t-sql/statements/restore-statements-arguments-transact-sql?view=sql-server-ver16#-checksum--no_checksum-

Following best practices backups should always contain checksums, and therefore having the option to be alerted in case no checksum is present would be a nice.

That would be done by using `RESTORE VERIFYONLY ... WITH CHECKSUM` and result in this behavior:

> Specifies that backup checksums must be verified and, if the backup lacks backup checksums, causes the restore operation to fail with a message indicating that checksums are not present.

potatoqualitee:
## Summary
Adds the `-Checksum` parameter to Test-DbaLastBackup, which controls whether to include WITH CHECKSUM in the RESTORE VERIFYONLY command.

## Changes
- Added `-Checksum` switch parameter to Test-DbaLastBackup
- Added `-Checksum` parameter to Restore-DbaDatabase
- Added `-Checksum` parameter to Invoke-DbaAdvancedRestore
- Set Checksum property on SMO Restore object when parameter is specified
- Updated Test-DbaLastBackup.Tests.ps1 to include new parameter
- Added documentation and example showing Checksum usage

When `-Checksum` is used, SQL Server will verify backup checksums and fail if checksums are not present in the backup, following backup best practices.

Fixes #9735

### DIFF
--- a/tests/Invoke-DbaAdvancedRestore.Tests.ps1
+++ b/tests/Invoke-DbaAdvancedRestore.Tests.ps1
@@ -32,6 +32,7 @@ Describe $CommandName -Tag UnitTests {
                 "StopBefore",
                 "StopMark",
                 "StopAfterDate",
+                "Checksum",
                 "EnableException"
             )
             Compare-Object -ReferenceObject $expectedParameters -DifferenceObject $hasParameters | Should -BeNullOrEmpty
--- a/tests/Restore-DbaDatabase.Tests.ps1
+++ b/tests/Restore-DbaDatabase.Tests.ps1
@@ -61,6 +61,7 @@ Describe $CommandName -Tag UnitTests {
                 "StopMark",
                 "StopAfterDate",
                 "ExecuteAs",
+                "Checksum",
                 "NoXpDirRecurse"
             )
             Compare-Object -ReferenceObject $expectedParameters -DifferenceObject $hasParameters | Should -BeNullOrEmpty


potatoqualitee:
@claude when you add a new param, you have to update the unit tests too

```
Path           : Invoke-DbaAdvancedRestore/Parameter validation/Should have the expected parameters
Name           : It Should have the expected parameters
Result         : Failed
Message        : ErrorRecord: Expected $null or empty, but got @{InputObject=Checksum; SideIndicator==>}.
StackTrace     : 
RawErrorRecord : Expected $null or empty, but got @{InputObject=Checksum; SideIndicator==>}.
Uploading artifact DetailedTestFailures_Pester5.json (893 bytes)...100%
```

PR #9917, opened 2025-11-03
Invoke-DbaDiagnosticQuery - Fix Azure SQL DB query execution with limited permissions

### LINKED ISSUE #9741: Invoke-DbaDiagnosticQuery fails on Azure SQL DB with lower permissions but T-SQL script runs fine

Error message:
``` powershell
VERBOSE: [16:39:39][Invoke-DbaDiagnosticQuery] Collecting diagnostic query data from <database> for Ad hoc Queries on <server>
VERBOSE: [16:39:41][Invoke-DbaDiagnosticQuery] Some error has occurred on Server: <server> - Script: Ad hoc Queries - Database: <database>, result will not be saved | Exception calling "Query" with "2" argument(s): "Execute with results failed for Database '<database>'. Failed to connect to server <server> failed for user ''."
```

Result of `Get-Error`:

Image

### Steps to Reproduce

```powershell
$azureAccount = Connect-AzAccount  -TenantId "<tenantId>"
$azureToken = Get-AzAccessToken -ResourceUrl https://database.windows.net
$azureInstance = "<instance>"
$azureDatabase = "<database>"
$server = Connect-DbaInstance -SqlInstance $azureInstance -Database $azureDatabase -AccessToken $azureToken

Invoke-DbaDiagnosticQuery -SqlInstance $server -Database $azureDatabase -UseSelectionHelper
```

Select one that has `DBSpecific` being `true`.

2.5.1

## Context
I have tested this in two different environments where permissions' levels are different.

In the environment where I have higher permissions, the error doesn't happen.
On the other one, where I only have permissions on a specific database, the error that I'm reporting happens.

While this could be seen as a "not enough permissions" kind of issue, if I open Glenn's T-SQL script in SSMS and run it while connected to that database, **it ran fine with no errors**.

## Investigation
1. The problem happens when running the `Query()` method -https://github.com/dataplat/dbatools/blob/master/public/Invoke-DbaDiagnosticQuery.ps1#L399 - that has 2 arguments (being the second one the database name)
2. That section of the code is only reached when the property (mentioned before) `DBSpecific` is `true`. Otherwise, the overload used is the one with just one argument (without the database name) - https://github.com/dataplat/dbatools/blob/master/public/Invoke-DbaDiagnosticQuery.ps1#L323
3. I checked how we (dbatools) are filling that property. That happens here - https://github.com/dataplat/dbatools/blob/9de039970ad35368b75c389b481463303e644b7a/private/functions/Invoke-DbaDiagnosticQueryScriptParser.ps1#L36

## Workaround for testing and confirming wasn't permissions
I opened the local script and changed the T-SQL comment so that validation doesn't match.

Image

Saved the file and rerun the script. 
It worked without any error.

## Suggestion to be discussed
Glenn's script says:
Image

With my script, I was already connected to a specific database.
In case we are connected to Azure SQL DB, should we ignore the `DBSpecific` setting and treat it as if it is always `false` for all queries?

PowerShell Core (pwsh.exe)

### PowerShell Host Version

PSVersion                      7.5.2
GitCommitId                    7.5.2
OS                             Microsoft Windows 10.0.26100

Microsoft SQL Azure   (RTM) - 12.0.2000.8      	Jul  3 2025 16:52:14      	Copyright (C) 2025 Microsoft Corporation

### .NET Framework Version

.NET 9.0.6

potatoqualitee:
Resolves issue where database-specific queries fail on Azure SQL Database with lower permissions, even though the T-SQL runs fine in SSMS.

The problem occurred when using the 2-parameter Query() overload which attempts a new connection context. For Azure SQL DB, connections are already scoped to a specific database, so the 1-parameter overload works correctly even for DBSpecific queries.

Fixes #9741

### DIFF
--- a/public/Invoke-DbaDiagnosticQuery.ps1
+++ b/public/Invoke-DbaDiagnosticQuery.ps1
@@ -411,7 +411,14 @@ function Invoke-DbaDiagnosticQuery {
 
                             Write-Message -Level Verbose -Message "Collecting diagnostic query data from $($currentDb) for $($scriptpart.QueryName) on $instance"
                             try {
-                                $result = $server.Query($scriptpart.Text, $currentDb)
+                                # Azure SQL Database connections are already scoped to a specific database
+                                # Using the 2-parameter Query() overload can fail with limited permissions
+                                # For Azure SQL DB, use the 1-parameter overload even for DBSpecific queries
+                                if ($server.DatabaseEngineType -eq "SqlAzureDatabase") {
+                                    $result = $server.Query($scriptpart.Text)
+                                } else {
+                                    $result = $server.Query($scriptpart.Text, $currentDb)
+                                }
                                 if (-not $result) {
                                     [PSCustomObject]@{
                                         ComputerName     = $server.ComputerName


potatoqualitee:
@ClaudioESSilva pls test 🙏

potatoqualitee:
hell yeah!

PR #9931, opened 2025-11-04
Get-DbaAgentJob, Remove-DbaAgentJob - Add validation for null/empty Job parameter

### LINKED ISSUE #9441: Do not treat a null value passed to the "-Job" parameter as a request for ALL jobs
### Summarize Functionality

Passing null values to parameters will result in errors instead of the cmdlets proceeding as if the parameter was not explicitly passed.

Yes

### Technical Details

Not sure whether to report this as a bug or feature request, but here it goes.

I've just had a fun incident where I deleted all SQL jobs on an instance because I passed a variable name which didn't exist:

```
$sqlServer = "SqlServer"
$rightVariable = "This is my SQL job name"

$sqlJob = Get-DbaAgentJob -SqlInstance $sqlServer -Job $wrongVariable

Remove-DbaAgentJob -SqlInstance $sqlServer -Job $sqlJob -Confirm:$false
```

In the above example, $wrongVariable doesn't exist, and $sqlJob ends up with ALL SQL JOBS in it, which are then deleted.
(Confirmation was skipped because this is meant to run without human interaction.)

The same happens if I run the command like this:

`$sqlJob = Get-DbaAgentJob -SqlInstance $sqlServer -Job ""`
`$sqlJob = Get-DbaAgentJob -SqlInstance $sqlServer -Job $null`

I would expect the cmdlet to NOT proceed if there is a null value passed to a parameter.
The parameter is not MANDATORY, but if it's being explicitly passed, then it should contain _something_.

I don't know how much of this is Powershell's default behaviour VS a conscious design decision on the dbatools Cmdlets.

potatoqualitee:
## Summary

This PR fixes issue #9441 by adding validation to prevent null/empty values passed to the `-Job` parameter from being treated as a request for ALL jobs.

## Changes

- **Get-DbaAgentJob**: Added `Test-Bound` validation for `-Job` and `-ExcludeJob` parameters to detect null/empty/whitespace values
- **Remove-DbaAgentJob**: Added `Test-Bound` validation for `-Job` parameter to prevent accidental deletion of all jobs
- Added comprehensive test coverage for the new validation behavior

## Fixes

Fixes #9441

----

### DIFF
--- a/public/Get-DbaAgentJob.ps1
+++ b/public/Get-DbaAgentJob.ps1
@@ -143,7 +143,7 @@ function Get-DbaAgentJob {
             # Check if Job parameter is bound with null, empty, or whitespace-only values
             if (Test-Bound 'Job') {
                 if ($null -eq $Job -or $Job.Count -eq 0 -or ($Job | Where-Object { [string]::IsNullOrWhiteSpace($_) })) {
-                    Write-Message -Level Warning -Message "The -Job parameter was explicitly provided but contains null, empty, or whitespace-only values. This may indicate an uninitialized variable. No jobs will be returned."
+                    Write-Message -Level Verbose -Message "The -Job parameter was explicitly provided but contains null, empty, or whitespace-only values. This may indicate an uninitialized variable. Skipping instance."
                     continue
                 }
             }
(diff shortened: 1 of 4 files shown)


potatoqualitee:
@claude I've always been on the fence about warnings since they can overwhelm. Pls change to verbose and skip.

PR #9950, opened 2025-11-07
Add branch-aware test scope to AppVeyor CI

potatoqualitee:
Let's try controlling this with PowerShell instead of YAML?


potatoqualitee:
ugly solution i hate it

PR #9962, opened 2025-11-08
Adjust T-SQL Queries for Case Sensitive SQL Server Collations

potatoqualitee:
Updated SQL statements across multiple functions.

```
# T-SQL Case Sensitivity Fix

Fix T-SQL casing in queries for case-sensitive collations.

Reference file `sql-system-objects.txt` contains canonical casing for all SQL Server system objects.

## Rules

1. **Keywords**: UPPERCASE (SELECT, FROM, WHERE, JOIN, INSERT, UPDATE, DELETE, CREATE, ALTER, DROP, ORDER BY, GROUP BY, HAVING, CASE, WHEN, THEN, ELSE, END, AND, OR, NOT, IN, EXISTS, AS, ON, WITH, OVER, PARTITION BY)

2. **System objects**: Match exact casing from `sql-system-objects.txt` (e.g., `sys.databases`, `sys.dm_exec_sessions`, `INFORMATION_SCHEMA.TABLES`)

3. **Built-in functions**: UPPERCASE (ISNULL, COALESCE, CAST, CONVERT, COUNT, SUM, AVG, MIN, MAX, GETDATE, DATEADD, DATEDIFF, STRING_AGG, ROW_NUMBER, RANK)

4. **User objects**: Do not change (databases, tables, columns, user schemas)

## Process

- Find T-SQL in `.Query =`, `Invoke-DbaQuery -Query`, `ExecuteNonQuery`, etc.
- Apply rules above
- Preserve indentation and readability
- Keep all comments
- Don't alter PowerShell code

IGNORE CLAUDE.MD, DO NOT CHANGE THE POWERSHELL, JUST THE T-SQL.
IF THERE'S NO WORK TO DO, DON'T INVENT WORK ON NON-TSQL CODE. JUST SKIP.
```

Also included an export of 

```sql
SELECT DISTINCT
    SCHEMA_NAME(schema_id) + '.' + name AS FullName
    FROM sys.system_objects
    WHERE SCHEMA_NAME(schema_id) IN ('sys', 'INFORMATION_SCHEMA')
    UNION
    SELECT DISTINCT SCHEMA_NAME(schema_id) + '.' + name
    FROM sys.system_views
    WHERE SCHEMA_NAME(schema_id) = 'sys'
    ORDER BY FullName
```


potatoqualitee:
meh

PR #9963, opened 2025-11-09
Fix tests related to PfDataCollectorSet: Add ComputerName parameter to support remote instance

andreasjordan:
Just a test fix to support remote instance.


potatoqualitee:
Thank you! 🥳

PR #9965, opened 2025-11-09
Fix test for Find-DbaOrphanedFile: Support remote instances

andreasjordan:
Support remote instances by using Invorke-Command2.


potatoqualitee:
Thank you! 🥳

PR #10013, opened 2025-11-29
Start-DbaMigration - Add -SetSourceOffline to set databases offline during migration

### LINKED ISSUE #9156: Add -SetSourceOffline to Start-DbaMigration
### Summarize Functionality

Would be great if we could add -SetSourceOffline to Start-DbaMigration so that we could set the source database on the source server offline during the migration.
Should work like as -SetSourceReadOnly, but instead taking the DBs offline.

This would prevent that applications are might be still able to connect to the source server.

Yes

### Technical Details

_No response_

potatoqualitee:
Enhances the -SetSourceOffline parameter to work like -SetSourceReadOnly, setting databases offline BEFORE migration begins (not after) to prevent any connections to source databases during the migration process.

Closes #9156

### DIFF
--- a/tests/Start-DbaMigration.Tests.ps1
+++ b/tests/Start-DbaMigration.Tests.ps1
@@ -195,4 +195,63 @@ Describe $CommandName -Tag IntegrationTests {
             $sourceDbs.Owner | Should -Be $destDbs.Owner
         }
     }
+
+    Context "When using SetSourceOffline parameter" {
+        BeforeAll {
+            $PSDefaultParameterValues["*-Dba*:EnableException"] = $true
+
+            # Create a dedicated database for offline testing
+            $offlineTestDb = "dbatoolsci_offline$random"
+
+            # Clean up any existing test database
+            Remove-DbaDatabase -SqlInstance $TestConfig.instance2, $TestConfig.instance3 -Database $offlineTestDb -ErrorAction SilentlyContinue
+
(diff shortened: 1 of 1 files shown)


potatoqualitee:
@claude add an integration test

potatoqualitee:
@claude hullo pls fix the test failures 

Path           : Start-DbaMigration/When using SetSourceOffline parameter/Should set source database offline before migration and bring destination online
Name           : It Should set source database offline before migration and bring destination online
Result         : Failed
Message        : ErrorRecord: A parameter cannot be found that matches parameter name 'Database'. This error might have been caused by applying the default parameter binding. You can disable the default parameter binding in 
                 $PSDefaultParameterValues by setting $PSDefaultParameterValues["Disabled"] to be $true, and then trying again. The following default parameters were successfully bound for this cmdlet when the error occurred: 
                 -Confirm -WarningVariable
StackTrace     : 
RawErrorRecord : A parameter cannot be found that matches parameter name 'Database'. This error might have been caused by applying the default parameter binding. You can disable the default parameter binding in 
                 $PSDefaultParameterValues by setting $PSDefaultParameterValues["Disabled"] to be $true, and then trying again. The following default parameters were successfully bound for this cmdlet when the error occurred: 
                 -Confirm -WarningVariable
Uploading artifact DetailedTestFailures_Pester5.json (1,813 bytes)...100%

PR #10028, opened 2025-12-02
Sync-DbaAvailabilityGroup - Exclude MSX jobs from syncing

### LINKED ISSUE #9273: Sync-DbaAvailabilityGroup not to try to copy over MSX agent jobs
### Summarize Functionality

Could we add additional checks to see if the SQL Agent Jobs are MSX/TSX jobs? I want to use this to sync AG servers but not have the errors involved when it tries to sync MSX agent jobs.

Yes

### Technical Details

maybe we filter out the sysjobs based on the category_id != 1 or originating_server_id != 1? 
at least that is what seems to show on my target server from 2008R2 to 2022 SQL.

potatoqualitee:
Resolves #9273

Added `-ExcludeMsxJobs` parameter to `Get-DbaAgentJob` to filter out MSX (Master Server) jobs (CategoryID = 1). Modified `Sync-DbaAvailabilityGroup` to automatically exclude MSX jobs when syncing agent jobs to prevent errors on secondary replicas.

### Changes
- Added `-ExcludeMsxJobs` switch to `Get-DbaAgentJob`
- Updated `Sync-DbaAvailabilityGroup` to use the new parameter when syncing jobs
- Updated parameter validation test

### DIFF
--- a/public/Get-DbaAgentJob.ps1
+++ b/public/Get-DbaAgentJob.ps1
@@ -28,10 +28,6 @@ function Get-DbaAgentJob {
         Excludes disabled SQL Agent jobs from the results, showing only enabled jobs.
         Use this when focusing on active job monitoring or troubleshooting since disabled jobs won't execute.
 
-    .PARAMETER IncludeMsxJobs
-        Includes MSX (Master Server) jobs in the results. By default, MSX jobs (CategoryID = 1) are excluded because they originate from multi-server administration and cannot be modified on target servers.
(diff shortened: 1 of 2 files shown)


potatoqualitee:
@claude i don’t think it’ll ever work to sync them. Do you agree? If so, make it the default and remove the param

potatoqualitee:
@claude but it'll never work? so just remove the param?

Issue #9573, opened 2025-01-07
`Get-Dba*` Wildcard Support


mattcargile:
### Summarize Functionality

I am of the opinion that all the applicable `Get-Dba*`  commands should support wildcards much like the greater core `pwsh` cmdlets like `Get-Service` or `Get-Process`. So `Get-DbaDatabase -Database Name*` or `Get-DbaAgentJob -JobName Name*`.

This is admittedly hits a more interactive userbase. I've come to use `dbatools` more and more versus _SSMS_ . So I have lots of aliases and such and so it would be handy. I know there is alot tab completion on many of the commands parameter's like `-Database` though it can lag a bit.

### Is there a command that is similiar or close to what you are looking for?

Yes

### Technical Details

Many of the cmdlets probably use `-in` so it would just be converting to a loop or the like and using `-like`.

niphlod:
Going to close this as has been discussed at length in Slack, it'd be a breaking change needing to release a major version to change filtering logic for Get-Dba*

PR #10041, opened 2025-12-10
Copy-DbaCredential  - Update documentation to reflect reality

Ant-Green:
JesseH had problem on Slack, the -Contains -NotContains filtering for Copy-DbaCredential wasn't working.

Replicated successfully locally.

Managed to get it working using $_.Identity.Contains("something")

Modified Get-DbaCredential and tests to include looking for wildcards

## Type of Change

 - [x] Bug fix (non-breaking change, fixes # ) --Not yet raised, asked Jesse on Slack to raise one.
 - [x] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose

Fixes a problematic Contains / NotContains when trying to use wildcards

### Approach

Switched from -Contains to $_.Identity.Contains which resolves the issue

### Screenshots

Using new methods works - well did, switched to -match / -notmatch as no .Net NotContains()
Original only works passing in the full identity
image

### DIFF
--- a/public/Copy-DbaCredential.ps1
+++ b/public/Copy-DbaCredential.ps1
@@ -39,21 +39,21 @@ function Copy-DbaCredential {
         This command requires access to the Windows OS via PowerShell remoting. Use this credential to connect to Windows using alternative credentials.
 
     .PARAMETER Name
-        Specifies the credential names to copy from the source server. Supports wildcards for pattern matching.
+        Specifies the credential names to copy from the source server. Does not supports wildcards for pattern matching.
         Use this when you only need to migrate specific credentials instead of all credentials on the server.
         Note: if spaces exist in the credential name, you will have to type "" or '' around it.
 
     .PARAMETER ExcludeName
-        Specifies credential names to exclude from the copy operation. Supports wildcards for pattern matching.
+        Specifies credential names to exclude from the copy operation. Does not support wildcards for pattern matching.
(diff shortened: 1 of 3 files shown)


andreasjordan:
Where does the docs say that wildcards are supported?

Ant-Green:
Copy-DbaCredential says it supports wildcards for name and identity.

Looking back through the code it's the Get-DbaCredential it is calling.

andreasjordan:
Ok, but the docs of the command you change don't say anything about wildcards. It would be a breaking change.

See also #9573. 

My suggestion: We don't change code for wildcards, but instead only update the docs to reflext the code.

mattcargile:
And we aren't technically talking about wildcards with `-match` like [`[WildCardPattern]`](https://learn.microsoft.com/en-us/dotnet/api/system.management.automation.wildcardpattern?view=powershellsdk-7.4.0) or [`[SupportsWildcards()]`](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_functions_advanced_parameters?view=powershell-7.5#supportswildcards-attribute) or `-like`. It would actually fit more with `-Pattern` usage like in `Find-DbaStoredProcedure`. It should probably be `-eq` or remain `-contains` like all the other `Get-` commands ( NOTE: I think all `Get-` should support `-like` behavior out the gate though that is a separate deal altogether and several others don't agree with that. )

potatoqualitee:
Thank you all 🙇🏼

PR #10048, opened 2025-12-16
Compare-DbaAgReplicaSync - Add new command to report AG replica sync differences

### LINKED ISSUE #10047: Sync-DbaAvailabilityGroup: just report the differences, not fix them
### Summarize Functionality

`Sync-DbaAvailabilityGroup` tries to automatically fix the problems it finds. However, my problems often require manual intervention. For example, I cannot know that what is on the primary is more correct than what is on the secondary. I therefore cannot use `Sync-DbaAvailabilityGroup` in my environment. This issue also makes the command useless for monitoring/alerting.

My suggested fix ought to be easier than `Sync-DbaAvailabilityGroup`. Instead of a command to _fix_ the mismatches between replicas, I want a command that _reports_ them. The ideal would be a proper display of the differences, but I would be happy with just being told what is different (e.g. a specific job, login, or config).

`-WhatIf` isn't enough for this. It prints a lot of information irrelevant to the problem. I am happy to prepare a Docker demonstration of this if requested.

Yes

### Technical Details

This could be a new command, but this feels similar to the `-ScriptOnly` options that we have elsewhere. It feels like it should just be a flag like `-DiffOnly` passed into `Sync-DbaAvailabilityGroup`.

potatoqualitee:
Closes #10047

This new command reports synchronization differences between AG replicas without making any changes, addressing the need for monitoring and manual review before fixing issues.

## Changes
- Created `Compare-DbaAgReplicaSync` following the established `Compare-DbaAgReplica*` naming pattern
- Compares Logins, Agent Jobs, Credentials, Linked Servers, Operators, Alerts, Proxies, and Custom Errors
- Returns structured diff data without making changes
- Registered in both dbatools.psd1 and dbatools.psm1
- Includes parameter validation tests

### DIFF
--- a/public/Compare-DbaAgReplicaSync.ps1
+++ b/public/Compare-DbaAgReplicaSync.ps1
@@ -155,7 +155,6 @@ function Compare-DbaAgReplicaSync {
 
                     foreach ($loginName in $allLoginNames) {
                         $loginConfigs = @{}
-                        $differences = New-Object System.Collections.ArrayList
 
                         # Collect login configurations from all replicas
                         foreach ($replicaInstance in $replicaInstances) {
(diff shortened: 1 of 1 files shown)


public/Compare-DbaAgReplicaSync.ps1
```diff
@@ -0,0 +1,555 @@
...
+                        } catch {
+                            Stop-Function -Message "Failed to retrieve logins from replica $replicaInstance" -ErrorRecord $_ -Target $replicaInstance -Continue
+                        }
+                    }
+
+                    foreach ($loginName in $allLoginNames) {
+                        $differences = New-Object System.Collections.ArrayList
+
+                        foreach ($replicaInstance in $replicaInstances) {
+                            $login = $loginsByReplica[$replicaInstance] | Where-Object Name -eq $loginName
```
ReeceGoding: That's not enough, right? Logins can differ in many more ways than just a login being completely missing.

potatoqualitee:
@claude compare the types of properties that copy-dbalogin copies over. i recall that being pretty complete. then expand that to to whatever smo is offering nowadays and ensure that you check per version. then consider if we can reuse Compare-DbaAgReplicaLogin

potatoqualitee:
yeah i think if we wanted to add permissions in the future it'd be a new command or new paramter. when i think of comparing, i dont think of the detailed permissions.

potatoqualitee:
@claude why did you choose not to update and wrap the existing command?

potatoqualitee:
@andreasjordan we should talk about standardizing the compare-dbaag commands and agree on the best. nbd for breaking chagnes since the new set is brand new.

ReeceGoding:
@potatoqualitee Does the Agent Job stuff also have the same "we're only checking for totally missing things" problem?

PR #10079, opened 2026-01-03
Fix tests so that they work against a Failover Cluster Instance

andreasjordan:
First round of fixes, more will follow later.


potatoqualitee:
sweet tks

PR #10082, opened 2026-01-04
Get-DbaService: Add property SqlInstance when using parameter AdvancedProperties to support Failover Cluster Instances

andreasjordan:
## Type of Change

 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
I need this later for a refctoring of Test-DbaDiskAllocation.

When working with failover cluster instances, the VSName (virtual server name / network name) is needed to connect to the instance instead of the computername that is output by Get-DbaService as the computername is resolved from the input. So `Get-DbaService -SqlInstance FCI01` will output "SQL01" as the computername if the instance is currently running on that computer.

### Approach
When using the parameter AdvancedProperties, the needed information is retrieved, but not well formatted. I changed the code so that for the engine service the SqlInstance property is calculated like in most of our other commands.

image


potatoqualitee:
awesome thank you 🙇🏼

PR #10084, opened 2026-01-06
Update Pester version to 5.7.1

andreasjordan:
This version is the latest version and about a year old, so probably the last 5.x version and we should use this.
I used this in my lab since we moved to pester 5 without problems.


potatoqualitee:
Nice work! Thank you 🙏🏼

PR #10091, opened 2026-01-10
Invoke-TlsWebRequest: Make Invoke-WebRequest faster

andreasjordan:
Using the same comment "IWR is crazy slow for large downloads" and variable "$currentProgressPref" to save current status like in "Get-DbaKbUpdate".

Also changing the tests to use this faster command.


potatoqualitee:
v nice change ty 💯

PR #10096, opened 2026-01-14
Improve tests to be more flexible

andreasjordan:
Changed these tests to be able to target different sql server versions and instance configurations.

Test-DbaConnection.Tests.ps1:
This test used a fixed port 1433, now gets the used port first.

Test-DbaDiskAlignment.Tests.ps1:
This test was the only test that used "$TestConfig.DbaToolsCi_Computer" and was run in the default scenario. The test now uses the computer of $TestConfig.instance1 and runs in that scenario. This way the default scenario only included tests that do not need a running instance.

Test-DbaLinkedServerConnection.Tests.ps1:
This test used a fixed port 1433, now uses the complete instance name as the target. So this test can run against every instance. For AppVeyor the test still has to use the old SQLNCLI11 instead of the MSOLEDBSQL provider as that one is not installed.

### DIFF
--- a/tests/Test-DbaConnection.Tests.ps1
+++ b/tests/Test-DbaConnection.Tests.ps1
@@ -25,10 +25,10 @@ Describe $CommandName -Tag UnitTests {
 Describe $CommandName -Tag IntegrationTests {
     Context "Testing if command works" {
         It "returns the correct results" {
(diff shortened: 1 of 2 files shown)


potatoqualitee:
ohhh nice change!

PR #10100, opened 2026-01-16
Test Improvements: Move last tests to new scenarios

andreasjordan:
* Changed all remaining usages of instance1 and instance2 to InstanceSingle
* Split scenario SINGLE into three parts to not hit the 1 hour limit
* Using version 2016 for SINGLE
* Using versions 2016 and 2017 for MULTI and COPY
* Using version 2017 for HADR and RESTART
* Using no sql server for scenario default as there are only tests that do not need a running instance
* enabled all tests (`"appveyor_disabled" = @()`)

If needed, I can later test SINGLE against 2008R2 Express to see how many tests work against that version.

If this PR is too big to review, I can split it and only change instance2 to InstanceSingle in the first round.

Situation before this PR (total runtime: 1 hr 32 min):
image

Situation after this PR (total runtime: 1 hr 36 min):
image

### DIFF
--- a/tests/appveyor.common.ps1
+++ b/tests/appveyor.common.ps1
@@ -247,13 +247,13 @@ function Get-TestsForBuildScenario {
     if ($env:PART) {
         try {
             [int]$num, [int]$denom = $env:PART.Split('/')
-            if (-not($Silent)) {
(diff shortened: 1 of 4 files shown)


potatoqualitee:
Thank you 🙏🏼  looks so good!

PR #10101, opened 2026-01-16
Use new AppVeyor image to test against SQL Server 2022

andreasjordan:
Just a test to see which commands would test successfully against the 2008R2 instance.

My plan is to tag those commands to be able to run an extra scenario with 2008R2.


andreasjordan:
This PR shows that it is possible to test both against 2017/2019 with the latest image and also against 2008R2/2017 with the current image.

It takes a long time to run all the tests, so this should not be used all the time. But we could just comment out the last two scenarios in the matrix-part of the appveyor.yml. Then we can include them from time to time or just run them on the merge to development or just before a new release.

One test against 2019 is currently failing as there is not patch installed in the 2019 instance. We would have to change the test to accept that situation on appveyor.

The two failing tests against 2008 in the last run are due to the instability of some tests - they both succeded in the previous run.

Feedback is welcome.

niphlod:
again, since I think 2-3 months, "normal" PRs do not run the whole suite, just the tests of the changed functions plus all dependencies. "Long runs" shouldn't be an issue since most PRs target 1 function.

andreasjordan:
Looks like the documentation of AppVeyor is old (just opened an issue). There is a SQL Server 2022 instance. So I recommend we use the new image and test most of the commands against SQL Server 2022.

Let's see if all tests pass...

andreasjordan:
Looks like the last changes introduced problems on AppVeyor:
image

image

I will look into this...

andreasjordan:
This is the only change relevant to the failure:
image

It works against SQL Server 2016 but not against SQL Server 2022.

This code is currently only executed when using `Export-DbaCredential`. When we use the linked server commands, then the DAC is created beforehand and that DAC is used instead of using Invoke-Command2 and a local DAC.

I will open a new PR (EDIT: Open an issue to discuss) to use the same logic in the credential commands like we introduced in the linked server commands.

So this PR will wait here for some more days...

potatoqualitee:
@andreasjordan thats weird as hell! I'm surprised something so fundamental has different behavior between two versions.

all that said, thank you 🙌🏼

PR #10108, opened 2026-01-24
Fix test for Get-InstalledPatch: Test for warnings but skip on AppVeyor 

andreasjordan:
The newer versions on AppVeyor are RTM, so they don't have patches to report. So we only test for warnings.

This will help to get all tests for #10101 green.


andreasjordan:
Just in case I would need to merge a feature branch into development myself: What method do we use?
image

andreasjordan:
@potatoqualitee You use "Squash and merge", right?

potatoqualitee:
@andreasjordan yes! squash and merge

Issue #10111, opened 2026-01-26
Harmonize the use of Get-DecryptedObject


andreasjordan:
`Get-DecryptedObject` is used by these commands:
* `Copy-DbaLinkedServer` in case `-ExcludePassword` is not used with a dedicated admin connection
* `Export-DbaLinkedServer` in case `-ExcludePassword` is not used with a dedicated admin connection
* `Copy-DbaCredential` in every case with a dedicated admin connection
* `Export-DbaCredential` in case `-ExcludePassword` is not used with a normal connection
* `Copy-DbaDbMail` in case there are mail server credentials with a normal connection

Only in the last two cases `Get-DecryptedObject` needs to open a local dedicated admin connection using `Invoke-Command2`.

We should remove this case entirely. Establishing a dedicated remote administrator connection to the source should be a prerequisite for all commands in the list. All commands in the list should get a parameter `-ExcludePassword` to avoid the dedicated admin connection.

Steps to do this:
- [X] Change `Export-DbaCredential` the way we changed `Export-DbaLinkedServer` (#10112)
- [X] Change `Copy-DbaCredential` and add a parameter `-ExcludePassword` (#10115)
- [X] Change `Copy-DbaDbMail` and add a parameter `-ExcludePassword` (#10117)

The first two are easy, I can do that and test that.
But the last one is a bit more work so I want to get the ok to invest this.

@potatoqualitee / @niphlod - What do you think?

This is related to the change #10106 and my findings in test of #10101

potatoqualitee:
These are some of our most challenging commands and I'm always relieved when you work on them. Happy for you to fully solve it this way.

andreasjordan:
While investigating, I found a bug in `Export-DbaLinkedServer`: Piping a linked server to the command does not work. And we should not allow piping to commands that need a DAC because in this case we can not use the DAC/non-DAC logic with ExcludePassword.

So I will refactor both `Export-DbaLinkedServer` and `Export-DbaLinkedServer` and remove the piping ability.

andreasjordan:
All merged, so closing this.

PR #10112, opened 2026-01-26
Export-DbaCredential: Open DAC unless ExcludePassword used

andreasjordan:
Helps to solve #10111 as the code is now aligned with Export-DbaLinkedServer.

### DIFF
--- a/public/Export-DbaInstance.ps1
+++ b/public/Export-DbaInstance.ps1
@@ -264,7 +264,7 @@ function Export-DbaInstance {
             if ($Exclude -notcontains 'Credentials') {
                 Write-Message -Level Verbose -Message "Exporting SQL credentials"
                 Write-ProgressHelper -StepNumber ($stepCounter++) -Message "Exporting SQL credentials"
-                $null = Export-DbaCredential -SqlInstance $server -Credential $Credential -FilePath "$exportPath\credentials.sql" -ExcludePassword:$ExcludePassword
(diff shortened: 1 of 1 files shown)


andreasjordan:
I did a complete refactoring and used the main logic from Export-DbaLinkedServer.

potatoqualitee:
than you andreas! but where did inputobject go?

andreasjordan:
InputObject is gone. We have a general problem with piping: We can use pipeline input in every command that only needs information from the piped input. But as soon as we need additional information we run into problems. In case of Export-DbaCredential, the command does not only export information that we get from Get-DbaCredential, but also decryptes the passwords. And then we need a new DAC to the server. Maybe that is possible but there would be massive code to check all cases. Having only a limited amount of time, we should generally review some of the edge case functionallity of the commands and remove them to enhance code quality and maintanability.

potatoqualitee:
@claude can the piping problem be solved with PipelineValueByPropertyName?

andreasjordan:
I totally agree with claude. This is where I like claude: It can write english much better than me.

So yes, it's not so much about "is it possible", but more about "who is willing to spend the time for implementation and maintanance(!!!)". And I am currently not willing to spend that extra time just for the pipeline functionality.

potatoqualitee:
Claude writes better English than everyone 😅  And is good at organizing for skimming.

@claude im having a hard time letting go of the piping.. could we simply collect the objects then in the end {} block, close their connections then reopen for dac?

potatoqualitee:
Alllllllllright

thank you andreas 😁

PR #10143, opened 2026-02-08
Get-DbaNetworkConfiguration: Add SuitableCertificate property to output

andreasjordan:
## Type of Change

 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
The plan is to refactor Set-DbaNetworkCertificate and use Get-DbaNetworkConfiguration to test for suitable but currently not configured certificates.
I also plan a new command Test-DbaNetworkCertificate to test both validity of the currect configured certificate and also the existance of suitable new certificates.

### Approach
I worked with Copilot and ChatGPT on implementing the correct tests, but I am not 100 percent sure if everything is correct.
I think the best approach is to include this in the next release of dbatools and ask users to test this in their environment and give feedback. The new code with the tests can also be used without dbatools by just using the relevant lines of code.

### DIFF
--- a/public/Get-DbaNetworkConfiguration.ps1
+++ b/public/Get-DbaNetworkConfiguration.ps1
@@ -237,23 +237,28 @@ function Get-DbaNetworkConfiguration {
             $requiredKeyUsages = [System.Security.Cryptography.X509Certificates.X509KeyUsageFlags]::DigitalSignature -bor [System.Security.Cryptography.X509Certificates.X509KeyUsageFlags]::KeyEncipherment
             $networkName = if ($vsname) { $vsname } else { hostname }
             $suitableCertificate = Get-ChildItem -Path Cert:\LocalMachine\My -ErrorAction SilentlyContinue | Where-Object {
-                $keyUsages = ($_.Extensions | Where-Object { $_ -is [System.Security.Cryptography.X509Certificates.X509KeyUsageExtension] }).KeyUsages
-                $keyUsagesOk = ($keyUsages -band $requiredKeyUsages) -eq $requiredKeyUsages
+                try {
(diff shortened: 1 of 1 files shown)


andreasjordan:
@claude please review my pull request.

andreasjordan:
I will add a commit to fix "Must fix: DNS name array handling" based on the feedback from claude. I will also include a try catch block as a savety net.
But I will not implement "Nice to have: Workgroup scenario handling" as most of the users will have a domain.
And I will not implement "Should fix: CNG certificate support" as CNG certificates are not supported by SQL Server. If we ever find out that the microsoft documentation is wrong, then I will add the support to gether with a user that supports us with a working example.

andreasjordan:
@claude please have a look at my last comment and my last commit and see if I made a mistake.

potatoqualitee:
looks clean and new name is suitable

andreasjordan:
Oh, I made a big mistake, and even Claude's code review didn't catch it: The new information is retrieved from the remote computer, but in the last step, it is not passed on to the user.
How could this happen? I only tested the code executed on the target itself and didn't test the finished command at the end. And I haven't written an automated test yet either. Maybe I should start doing that now...

For those how came here and want to test the code:
So, for now, just use the code from the pull request itself on the target system for testing; that should work. I'll let you know as soon as the command has been fixed.

andreasjordan:
Added to #10165 to fix this.

andreasjordan:
Now the code is ready to test and included in version 2.7.25 of dbatools.

potatoqualitee:
it looked reasonable to me too, @andreasjordan

PR #10145, opened 2026-02-08
Install-DbaMaintenanceSolution.Tests: make the additional backup parameters tests run

ReeceGoding:
## Type of Change

 - [X] Bug fix (non-breaking change, fixes part of #10144)
 - [ ] New feature (non-breaking change, adds functionality, fixes # )
 - [ ] Breaking change (affects multiple commands or functionality, fixes # )
 - [X] Ran manual Pester test and has passed (`Invoke-ManualPester`)
 - [ ] Adding code coverage to existing functionality
 - [X] Pester test is included
 - [ ] If new file reference added for test, has is been added to github.com/dataplat/appveyor-lab ?
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
As mentioned in passing in #10144, I'm struggling to convince myself that the "Additional backup parameters" in [`Install-DbaMaintenanceSolution.Tests.ps1`](https://github.com/dataplat/dbatools/blob/development/tests/Install-DbaMaintenanceSolution.Tests.ps1#L112) run. I can [see warnings](https://ci.appveyor.com/project/dataplat/dbatools/builds/53504593/job/3i18881vnwkxh7w7) in appveyor
> Running C:\github\dbatools\tests\Install-DbaMaintenanceSolution.Tests.ps1 ...
> 
> [00:38:26][Install-DbaMaintenanceSolution] Could not execute DatabaseIntegrityCheck.sql in tempdb on appveyor-vm\sql2022 | Log entry string is too long. A string written to the event log cannot exceed 32766 characters.
> 
> [00:38:26][Install-DbaMaintenanceSolution] Could not execute IndexOptimize.sql in tempdb on appveyor-vm\sql2022 | Log entry string is too long. A string written to the event log cannot exceed 32766 characters.

and `Invoke-ManualPester -Path tests/Install-DbaMaintenanceSolution.Tests.ps1 -ScriptAnalyzer -Show Detailed -TestIntegration` skipped them on my machine as well.

These tests were introduced in [this commit](https://github.com/dataplat/dbatools/commit/f72a50e736ecaf42c2ca1002c4d14a987a0563a2) and the PR suggests some difficulty with  appveyor at the time. That said, I can believe that this is all just me being confused. Today is the first day that I've had to get serious about `Invoke-ManualPester`.

Once this has been looked at, I'll be able to approach #10133 with more confidence. #10133 will require me adding more tests to this file, so I would like to be confident in this file's code before I start copy and pasting it.

### Approach
[...]

Issue #5356, opened 2019-04-11
[Bug] Backup-DbaDbCertificate Error: Could not load file or assembly 'Microsof t.SqlServer.BatchParser.dll' or one of its dependencies.


RobertVirag:
### Before submitting a bug report:

**Collect output of following command and paste below:**

```
& {"### PowerShell version:`n$($PSVersionTable | Out-String)"; "`n### dbatools Module version:`n$(gmo dbatools -List | select name, path, version | fl -force | Out-String)"}
```
- [x] *Running latest release of dbatools*
- [x] Verified errors are not related to permissions
- [ ] Is this bug with `Copy-DbaDatabase`? You can replicate it using `Backup-DbaDatabase ... | Restore-DbaDatabase ...`

> NOTE: `Copy-DbaDatabase` will not work in every environment and every situation. Instead, we try to ensure Backup & Restore work in your environment.

## Environmental data

```
### PowerShell version:
Name                           Value                                                                                                                          
----                           -----                                                                                                                          
PSVersion                      4.0                                                                                                                            
WSManStackVersion              3.0                                                                                                                            
SerializationVersion           1.1.0.1                                                                                                                        
CLRVersion                     4.0.30319.42000                                                                                                                
BuildVersion                   6.3.9600.19170                                                                                                                 
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0}                                                                                                           
PSRemotingProtocolVersion      2.2                                                                                                                            
### dbatools Module version:
Name    : dbatools
Path    : E:\DATA\RV\SQLMGMT_V5_DEV\PSDev\PSModules\dbatools\dbatools.psd1
Version : 0.9.775
```

### SQL Server: 

Microsoft SQL Server 2014 (SP1-CU4) (KB3106660) - 12.0.4436.0 (X64) 
	Dec  2 2015 16:09:44 
	Copyright (c) Microsoft Corporation
	Enterprise Edition: Core-based Licensing (64-bit) on Windows NT 6.3 <X64> (Build 9600: ) (Hypervisor)
```

<!-- NOTE: If the above information is not provided as a minimum your issue will not be acknowledged -->

## Errors Received

<!-- Provide full output of `$error[0] | Select *` -->
writeErrorStream      : True
PSMessageDetails      : 
Exception             : System.Exception: Could not load file or assembly 'Microsoft.SqlServer.BatchParser.dll' or one of its dependencies. The specified 
[...]

Issue #4073, opened 2018-09-26
Install-DbaWhoIsActive throws batch parser error.


githubsqltoolsse:
### Before submitting a bug report:

- [ x] Ensure you are able to reproduce it on the latest released version (_we release often_)
- [ x] Verified this bug is not already reported in an issue
- [ ?] Verified errors are not related to permissions
- [ x] Can reproduce in a clean PowerShell session (_clean = `powershell -NoProfile`_)

## Steps to Reproduce

```powershell

Install-DbaWhoIsActive -SqlInstance SESTODS030 -Database maintenance -Force -EnableException

This fails on some of our servers but works on most of them. I can also run the same command remotely from a client and it works.

```

## Expected Behavior

I expect to get WhoIsActive installed

## Actual Behavior

Exception calling "ExecuteNonQuery" with "1" argument(s): "ExecuteNonQuery failed for Database 'maintenance'. "
At C:\Program Files\WindowsPowerShell\Modules\dbatools\0.9.439\allcommands.ps1:99210 char:9
+         throw $records[0]
+         ~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (SESTODS030:String) [], Exception
    + FullyQualifiedErrorId : dbatools_Install-DbaWhoIsActive

## Environmental data

Name                           Value                                                                                                                                                          
----                           -----                                                                                                                                                          
PSVersion                      5.1.14393.2515                                                                                                                                                 
PSEdition                      Desktop                                                                                                                                                        
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}                                                                                                                                        
BuildVersion                   10.0.14393.2515                                                                                                                                                
CLRVersion                     4.0.30319.42000                                                                                                                                                
WSManStackVersion              3.0                                                                                                                                                            
PSRemotingProtocolVersion      2.3                                                                                                                                                            
SerializationVersion           1.1.0.1 

 - PowerShell: 
 - - SQL Server: 
Microsoft SQL Server 2014 (SP2-CU12) (KB4130489) - 12.0.5589.7 (X64)   Jun  9 2018 11:17:15   Copyright (c) Microsoft Corporation  Standard Edition (64-bit) on Windows NT 6.3 <X64> (Build 14393: ) (Hypervisor)

potatoqualitee:
thank you for the report.

right after it fails, can you do an `$error[0] | select *` and tell us what it says?
tks

[...]

PR #10149, opened 2026-02-10
Use Invoke-DbaQuery instead of ExecuteNonQuery for long batches to avoid issues on AppVeyor

andreasjordan:
The root cause is still unknown, but long batches trigger the error "Log entry string is too long. A string written to the event log cannot exceed 32766 characters." to be raised on AppVeyor and maybe other CI pipelines.
I can not reproduce the issue in my lab and I talked a lot with different LLMs without getting to the root cause.

So the best way is to run the code through our command Invoke-DbaQuery. That adds some overhead but on the other hand takes care of splitting the SQL into batches.

Only four of our commands use long batches, so I changed only those.

For more details see #10145.

### DIFF
--- a/public/Install-DbaMaintenanceSolution.ps1
+++ b/public/Install-DbaMaintenanceSolution.ps1
@@ -806,7 +806,7 @@ function Install-DbaMaintenanceSolution {
                 }
             }
 
-            if ($query) {
+            if ($sql) {
                 # then whatif wasn't passed
                 [PSCustomObject]@{
                     ComputerName = $server.ComputerName


potatoqualitee:
agreed, thank you 🙏🏼 

I was curious and asked Claude to do some research as to why it happens and the answer sounds plausible. 

----

Here's the TL;DR:

**Why it fails on AppVeyor:** SMO's `Database.ExecuteNonQuery` internally loads a native C++ DLL called `Microsoft.SqlServer.BatchParser.dll` to parse GO separators — even though you're already pre-splitting batches in PowerShell. If that DLL is missing or version-mismatched on AppVeyor's image (very common), every batch execution fails. This is documented in dbatools issues #4073 and #5356. It's *not* a batch size limit — sp_WhoIsActive's ~150KB script is way under SQL Server's ~256MB cap.

**Why the event log explodes:** When `ExecuteNonQuery` throws, SMO stuffs the entire SQL batch text into the exception. Passing that via `-ErrorRecord $_` to `Stop-Function` → AppVeyor writes it to the Windows Event Log → hits the 32,766-char hard limit → secondary `ArgumentException`.

**Fix both problems at once:** Replace SMO `ExecuteNonQuery` with raw ADO.NET `SqlCommand`:

```powershell
foreach ($batch in $batches) {
    try {
        $cmd = $server.ConnectionContext.SqlConnectionObject.CreateCommand()
        $cmd.CommandText = $batch
        $cmd.CommandTimeout = 300
        $null = $cmd.ExecuteNonQuery()
    } catch {
        $msg = $_.Exception.Message.Substring(0, [Math]::Min($_.Exception.Message.Length, 500))
        Stop-Function -Message "Failed to install stored procedure: $msg" -Continue -Target $instance
    }
}
```

This eliminates the BatchParser.dll dependency entirely **and** truncates error messages so they never overflow the event log. No more chasing environment-specific DLL issues across CI platforms.

andreasjordan:
Then we should change our methods `.Query()` and `.Invoke()` on the Server SMO and Database SMO.

PR #10150, opened 2026-02-11
Part 1 of refactoring of Get-DecryptedObject

andreasjordan:
`Get-DecryptedObject` needs two things:
* A Dedicated Admin Connection (DAC) to select master.sys.syslnklgns
* A WinRM connection to read "HKLM:\SOFTWARE\Microsoft\Microsoft SQL Server\$serviceInstanceId\Security\"

That's why `Get-DecryptedObject` needs `$Credential` which it used all the time, but the parameter was missing.

I also added documentation and restructured the code for better readability.

The command has still the code to open a DAC itself, which will be removed as the last step after all the commands that call this command are also refactored.

### DIFF
--- a/private/functions/Get-DecryptedObject.ps1
+++ b/private/functions/Get-DecryptedObject.ps1
@@ -1,15 +1,43 @@
 function Get-DecryptedObject {
     <#
-            .SYNOPSIS
-                Internal function.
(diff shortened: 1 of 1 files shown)


potatoqualitee:
awesome! thank you 🙏🏼

PR #10179, opened 2026-02-19
speedup dependency detection for integration tests

niphlod:
(do Join-DbaPath)

## Type of Change

 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [x] Build system

### Purpose
Someone told me code was slow. It kinda was, for huge dependency trees .... now it's an order of magnitude faster, and has little helpers to avoid spending time re-analyzing the correctness of the code in the future.

### Approach
Basically re-use everything as much as possible, computing the lookup beforehand and memoizing each result at each step.
Since this is a "recursive by design" approach (we're finding the dependencies of the tests... which is counterintuitive since our goal is finding what has the altered command(s) AS A DEPENDENCY, and then follow those branches out for each one, till eventually all the tree is resolved ) the memoization pattern is the most appropriate.
For "usual" PRs this shaves maybe 3 seconds, for PRs involving "well connected" functions, it may shave off maybe one or two mins.

I'm including Join-DbaPath in the "special commit" message just to see a mildly-well-connected function to be worked on.

### DIFF
--- a/tests/appveyor.common.ps1
+++ b/tests/appveyor.common.ps1
@@ -2,6 +2,7 @@ function Get-TestsForScenario {
     param($Scenario, $AllTests, [switch]$Silent)
 
     # does this scenario run an 'autodetect' ?
+
     if ($TestsRunGroups[$Scenario].StartsWith('autodetect_')[0]) {
         # exclude any test specifically tied to a non-autodetect or version specific scenario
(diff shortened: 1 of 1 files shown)


potatoqualitee:
awesome 😁  thank you!

Issue #10183, opened 2026-02-20
Install-DbaMaintenanceSolution: defaults reported by documentation are incorrect


ReeceGoding:
### Verified issue does not already exist?

I have searched and found no existing issue

### What error did you receive?

No error. The wrong thing is done silently.

### Steps to Reproduce

Firstly, read the `Install-DbaMaintenanceSolution` documentation and observe our claims that:

* When `-Compress` is not specified, we use "_the SQL Server instance’s default compression setting_".
* Both `-Verify` and `-CheckSum` "_Defaults to enabled (Y) if not specified_".

Secondly, spin up an instance with backup compression defaulted to off 

```powershell
# Run after creating the instance
Set-DbaSpConfigure -SqlInstance $YourMachine -ConfigName 'DefaultBackupCompression' -Value 0
```
and run 
```powershell
$splatInstall = @{
	SqlInstance      = $YourMachine
	InstallJobs      = $true
	ReplaceExisting  = $true
}
Install-DbaMaintenanceSolution @splatInstall
```
Thirdly, note that the above definitely did not set `-Compress`, `-Verify`, or `-CheckSum`.

Fourthly, observe that the produced jobs do not turn on what they say they should. This

```powershell
(Get-DbaAgentJobStep -SqlInstance $YourMachine -Job "DatabaseBackup - USER_DATABASES - FULL").Command
```

returns

```sql
EXECUTE [dbo].[DatabaseBackup]
@Databases = 'USER_DATABASES',
@Directory = N'/var/opt/mssql/data',
@BackupType = 'FULL',
@Verify = 'N',
@CleanupTime = NULL,
@CheckSum = 'N',
@LogToTable = 'N',
@Compress = 'N'
```
which explicitly does the right thing for `@Compress`, but the wrong thing for  `-Verify` and `-CheckSum`.

On a version from last week, I get
```sql
EXECUTE [dbo].[DatabaseBackup]
@Databases = 'USER_DATABASES',
@Directory = N'/var/opt/mssql/data',
@BackupType = 'FULL',
@Verify = 'Y',
@CleanupTime = NULL,
@Checksum = 'Y',
@LogToTable = 'N'
```
which does the right thing for `-Verify` and `-CheckSum`, but is not explicit about `-Compress`.

From this, it is quite clear that my #10172 has a lot of the blame. **By making the off switches work, I made the documentation a liar.**

### Please confirm that you are running the most recent version of dbatools

[...]

PR #10184, opened 2026-02-21
Install-DbaMaintenanceSolution: adds tests for defaults reported by documentation, some failing

ReeceGoding:
Adds many tests for #10183. Four of these fail.

### The Important Parts

Although a lot of tests are added here, the bottom line is that this adds four failing tests. The four that fail check that we have the following documentation-claimed behaviors:

* When `-Verify` is not specified, we should enable verify.
* When `-Checksum` is not specified, but the instance has it on by default, we should enable it in the job.
* When `-Checksum` is not specified, but the instance has it off by default, we should enable it in the job.
* When `-Compress` is not specified, but the instance has it on by default, we should enable it in the job.

We can fix these once we agree how. **Somebody will have to tell me how we want this done**. Most of this is mentioned in #10183.

### Full Details

This adds a test context for "Defaults for unspecified parameters are as documentation says", including three tests:

* "Should have Verify parameter set to Y in backup jobs", which fails as #10183 predicts
* "Should have StartTime set to 011500 in backup schedule", which passes
* "Should have BackupLocation parameter set to instance default in backup jobs", which passes.

The "Checksum tests when instance defaults to checksum on" and "Checksum tests when instance defaults to checksum off" contexts are expanded to test for `-CheckSum` on, off, and not specified. To accommodate this, we now uninstall the maintenance solution after each test in these contexts. In both contexts, only the "not specified" test fails. This is as #10183 predicts.

Similarly, the two contexts "Compression tests when instance defaults to compression on" and "Compression tests when instance defaults to compression off" have been added and they test for `-Compress` on, off, and not specified. Among these, only the test where compression is turned on by default for the instance but not specified in the install command fails.

## Type of Change

 - [X] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -67,25 +67,25 @@ environment:
     - scenario: RESTART
       InstanceRestart: sql2022
     - scenario: default
-    - scenario: 2008R2SP2Express
-      part: 1/2
-      InstanceSingle: sql2008r2sp2
-      InstanceMulti1: sql2008r2sp2
-      InstanceMulti2: sql2017
-      InstanceCopy1: sql2008r2sp2
-      InstanceCopy2: sql2017
-      InstanceRestart: sql2008r2sp2
-      APPVEYOR_BUILD_WORKER_IMAGE: Previous Visual Studio 2015
-    - scenario: 2008R2SP2Express
-      part: 2/2
-      InstanceSingle: sql2008r2sp2
-      InstanceMulti1: sql2008r2sp2
-      InstanceMulti2: sql2017
-      InstanceCopy1: sql2008r2sp2
-      InstanceCopy2: sql2017
-      InstanceHadr: sql2017
-      InstanceRestart: sql2008r2sp2
-      APPVEYOR_BUILD_WORKER_IMAGE: Previous Visual Studio 2015
+    #- scenario: 2008R2SP2Express
+    #  part: 1/2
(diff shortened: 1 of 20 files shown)


andreasjordan:
I think:
* "Verify" and "Checksum" are switch parameters, so have a default of $false.
* "Verify" is an extra to the backup job that should not be enabled by default as it adds time and maybe costs.
* "Checksum" should be set to $true at the instance level with just one setting an not be included in every backup job.
* So they both should not be included in the command by default.
* "Checksum" should not be included based on the instance setting.
* They should only be included and set to "Y" if the switch parameter is used.

andreasjordan:
I am not a fan of useing three states of a switch "$true / $false / not used" as it makes things like `Install-DbaMaintenanceSolution -Verify:$verify -Checksum:$checksum` impossible.

ReeceGoding:
Let's discuss on the issue's page.

potatoqualitee:
Greets, any progress?

andreasjordan:
No. There is an unanswered question to you @potatoqualitee on the linked issue. I would change the defaults of the parameters. This would be a breakting change, but I think I have named good reasons for that. But we want to have an agreement before changing this pr to include the code changes.

andreasjordan:
I will close this now as #10247 is merged.

PR #10190, opened 2026-02-23
Import-DbaCsv - Fix RFC 4180 multiline quoted field handling

### LINKED ISSUE #10188: Import-DbaCsv Pre-validation incorrectly rejects valid CSV files with multiline quoted fields

WARNING: [HH:mm:ss][Import-DbaCsv] Delimiter (,) not found in first few rows of /path/to/file.csv. If this is a single column import, please specify -SingleColumn

### Steps to Reproduce

# Create a csv file where the first field in either the first or second row spans multiple lines 

```
name,value
"Hello
World",42
"test","data"

```

# Attempt to import - this will fail with a false delimiter warning
Import-DbaCsv -Path "C:\temp\multiline.csv" -SqlInstance localhost -Database tempdb -Table test -autoCreateTableFO

dbatools version: 2.7.25

### Other details or mentions

The delimiter pre-validation at line 969 of Import-DbaCsv.ps1 uses Get-Content to read the first few lines of the file, then checks if the delimiter is present. The problem is that Get-Content splits on newlines without understanding CSV quoting rules. When a quoted field contains an embedded newline (which is valid per RFC 4180), Get-Content splits it into separate lines. The continuation line may not contain the delimiter character.

  The check uses -notmatch on the resulting array, which returns elements that don't match. If any line lacks the delimiter, the result is a non-empty array (truthy) and the file is rejected. The logic should be inverted: reject only when no lines contain the delimiter.

  # Current (buggy) - rejects if ANY line lacks delimiter
  if ($firstlines -notmatch [regex]::Escape($Delimiter))

  # Possible Fix - rejects only if NO lines contain delimiter
  if (-not ($firstlines -match [regex]::Escape($Delimiter)))

  The -SupportsMultiline parameter does not help because this validation runs before the CSV reader is ever configured.

Windows PowerShell (powershell.exe), PowerShell Core (pwsh.exe)

### PowerShell Host Version

7.5.4

17.0.4015.4

### .NET Framework Version

4.8.1

### LINKED ISSUE #10189: Import-DbaCsv Multiline quoted fields cause parse errors by default with RFC 4180 Strict mode

[HH:mm:ss][Import-DbaCsv] Failure reading CSV: MismatchedFieldAction is set to ThrowException. Row X has Y field(s) but Z were expected.

(The exact error varies depending on where the multiline field falls and the MismatchedFieldAction setting.)

### Steps to Reproduce

```
# Create a valid RFC 4180 CSV where multiline fields appear after the first few rows (to avoid bug #10188)
  $lines = @("name,description,value")
  # Add normal rows so the delimiter pre-check passes
  for ($i = 1; $i -le 100; $i++) {
      $lines += "user$i,""normal description"",$i"
  }
  # Add a row with a multiline quoted field (valid per RFC 4180)
  $lines += 'multiline,"This description has' + "`r`n" + 'an embedded newline",999'
  $lines += 'last,"simple",0'

  $lines -join "`r`n" | Set-Content -Path "C:\temp\multiline_later.csv" -NoNewline

  # This passes the delimiter check but fails during parsing
  Import-DbaCsv -Path "C:\temp\multiline_later.csv" -SqlInstance localhost -Database tempdb -Table test -AutoCreateTable
```

2.7.25

### Other details or mentions

The default QuoteMode is Strict, documented as "RFC 4180 compliant parsing". However, RFC 4180's ABNF grammar explicitly allows CR and LF inside quoted (escaped) fields:

escaped = DQUOTE *(TEXTDATA / COMMA / CR / LF / 2DQUOTE) DQUOTE

Despite claiming RFC 4180 compliance, AllowMultilineFields on the CSV reader defaults to $false because the -SupportsMultiline switch is off by default.
When the reader encounters a newline inside a quoted field, it treats it as a record separator instead of part of the field value. This produces rows with mismatched field counts, which fails with the default MismatchedFieldAction of ThrowException.

Workaround: Pass -SupportsMultiline explicitly

The fix is to enable AllowMultilineFields by default when QuoteMode is Strict, since multiline quoted fields are part of the RFC 4180 specification. The -SupportsMultiline switch should still override the behavior when explicitly provided.

This needs to be applied in all three places where CsvReaderOptions is constructed: the New-SqlTable function, the schema inference path, and the main import path.

Alternatively the documentation should be amended to clarify that strict is not fully RFC 4180 compliant parsing unless the -SupportsMultiline switch is also set

This bug is a breaking change for anyone upgrading from an old LumenWorks based dbatools version, where the default mode was RFC 4180 compliant including multi-line support

PowerShell Core (pwsh.exe)

### PowerShell Host Version

7.5.4

17.0.4015.4

### .NET Framework Version

4.8.1

mbentham:
## Summary

Fixes two bugs that caused `Import-DbaCsv` to reject valid CSV files containing multiline quoted fields (permitted by [RFC 4180](https://www.rfc-editor.org/rfc/rfc4180)).

- **Pre-validation false rejection (#10188):** The delimiter check used `-notmatch` on lines read by `Get-Content`, which splits on newlines without understanding CSV quoting. If any sampled line lacked the delimiter (e.g. a continuation line from a multiline quoted field), the file was rejected. Changed to `-match` so the file is only rejected when *no* sampled lines contain the delimiter.

- **Multiline fields disabled by default in Strict mode (#10189):** `AllowMultilineFields` defaulted to `$false` despite `QuoteMode` defaulting to `Strict` ("RFC 4180 compliant"). RFC 4180 explicitly allows CR/LF inside quoted fields. Now `AllowMultilineFields` defaults to `$true` when `QuoteMode` is `Strict`, unless `-SupportsMultiline` is explicitly provided. Applied consistently across all three `CsvReaderOptions` construction sites (table creation, schema inference, and main import).

Fixes #10188
Fixes #10189

## Test plan

- [ ] Import a CSV where the first data row starts with a multiline quoted field (triggers Bug 1)
- [ ] Import a CSV with multiline quoted fields deeper in the file without `-SupportsMultiline` (triggers Bug 2)
- [ ] Import a standard CSV without multiline fields (no regression)
- [ ] Import with `-QuoteMode Lenient` confirms multiline stays off unless `-SupportsMultiline` is passed
- [ ] Existing `Import-DbaCsv` tests pass

(do Import-DbaCsv)

### DIFF
--- a/public/Import-DbaCsv.ps1
+++ b/public/Import-DbaCsv.ps1
@@ -635,6 +635,12 @@ function Import-DbaCsv {
             $options.MaxDecompressedSize = $MaxDecompressedSize
             $options.SkipRows = $SkipRows
             $options.DuplicateHeaderBehavior = [Dataplat.Dbatools.Csv.Reader.DuplicateHeaderBehavior]::$DuplicateHeaderBehavior
+            # RFC 4180 allows CR/LF inside quoted fields, so enable multiline by default in Strict mode
+            if ($QuoteMode -eq "Strict" -and -not $PSBoundParameters.ContainsKey("SupportsMultiline")) {
+                $options.AllowMultilineFields = $true
+            } else {
+                $options.AllowMultilineFields = $SupportsMultiline.IsPresent
+            }
 
             try {
                 $reader = [Dataplat.Dbatools.Csv.Reader.CsvDataReader]::new($Path, $options)
@@ -966,7 +972,7 @@ WHERE c.object_id = OBJECT_ID(@tableName)
                     Stop-Function -Continue -Message "Failure reading $file" -ErrorRecord $_
                 }
                 if (-not $SingleColumn) {
-                    if ($firstlines -notmatch [regex]::Escape($Delimiter)) {
+                    if (-not ($firstlines -match [regex]::Escape($Delimiter))) {
                         Stop-Function -Message "Delimiter ($Delimiter) not found in first few rows of $file. If this is a single column import, please specify -SingleColumn"
                         return
                     }
@@ -1104,6 +1110,12 @@ WHERE c.object_id = OBJECT_ID(@tableName)
                         $inferOptions.Escape = $Escape
                         $inferOptions.Comment = $Comment
                         $inferOptions.Encoding = [System.Text.Encoding]::$Encoding
+                        # RFC 4180 allows CR/LF inside quoted fields, so enable multiline by default in Strict mode
+                        if ($QuoteMode -eq "Strict" -and -not $PSBoundParameters.ContainsKey("SupportsMultiline")) {
+                            $inferOptions.AllowMultilineFields = $true
+                        } else {
+                            $inferOptions.AllowMultilineFields = $SupportsMultiline.IsPresent
+                        }
                         if ($PSBoundParameters.DateTimeFormats) {
                             $inferOptions.DateTimeFormats = $DateTimeFormats
                         }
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Looks reasonable, thank you very much for the fix!

PR #10185, opened 2026-02-22
Test-DbaLsnChain: Fix bug in case log backup is taken during full backup

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes # )
 - [ ] New feature (non-breaking change, adds functionality, fixes # )
 - [ ] Breaking change (affects multiple commands or functionality, fixes # )
 - [ ] Ran manual Pester test and has passed (`Invoke-ManualPester`)
 - [ ] Adding code coverage to existing functionality
 - [ ] Pester test is included
 - [ ] If new file reference added for test, has is been added to github.com/dataplat/appveyor-lab ?
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

I have a customer facing this issue, so I did some analysis.

Here is the new logging before the fix:

```
AUSFÜHRLICH: [15:47:45][Test-DbaLsnChain] Testing LSN Chain
AUSFÜHRLICH: [15:47:45][Test-DbaLsnChain] Testing LSN Chain - Type Type
AUSFÜHRLICH: [15:47:45][<Unknown>] Checking C:\Temp\FULL_20260202_020339.bak - isLogBackup False, isBasedOnAnchor False, hasGreaterFirstLsn False, FullDBAnchor.CheckPointLsn 34347000156416000046, DatabaseBackupLsn 34347000148382400046, FirstLsn 34246000021188000002 LastLsn 34347000156420000001
AUSFÜHRLICH: [15:47:45][<Unknown>] Checking C:\Temp\LOG_20260202_020500.trn - isLogBackup True, isBasedOnAnchor False, hasGreaterFirstLsn False, FullDBAnchor.CheckPointLsn 34347000156416000046, DatabaseBackupLsn 34347000148382400046, FirstLsn 34347000156393600001 LastLsn 34347000156428000001
AUSFÜHRLICH: [15:47:45][<Unknown>] Checking C:\Temp\LOG_20260202_021002.trn - isLogBackup True, isBasedOnAnchor True, hasGreaterFirstLsn True, FullDBAnchor.CheckPointLsn 34347000156416000046, DatabaseBackupLsn 34347000156416000046, FirstLsn 34347000156428000001 LastLsn 34347000156458400001
WARNUNG: [15:47:45][Test-DbaLsnChain] Break in LSN Chain between C:\Temp\FULL_20260202_020339.bak and C:\Temp\LOG_20260202_021002.trn 
[...]

PR #10200, opened 2026-03-01
Test-DbaBackupInformation: Add member IsVerified to output if not already present

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #10127 )
 - [ ] New feature (non-breaking change, adds functionality, fixes # )
 - [ ] Breaking change (affects multiple commands or functionality, fixes # )
 - [ ] Ran manual Pester test and has passed (`Invoke-ManualPester`)
 - [ ] Adding code coverage to existing functionality
 - [ ] Pester test is included
 - [ ] If new file reference added for test, has is been added to github.com/dataplat/appveyor-lab ?
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

See issue for details.

PR #10199, opened 2026-03-01
Connect-DbaInstance: Use localhost for dedicated admin connections to avoid multiple IP resolution on multi-homed servers

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #10151 )
 - [ ] New feature (non-breaking change, adds functionality, fixes # )
 - [ ] Breaking change (affects multiple commands or functionality, fixes # )
 - [ ] Ran manual Pester test and has passed (`Invoke-ManualPester`)
 - [ ] Adding code coverage to existing functionality
 - [ ] Pester test is included
 - [ ] If new file reference added for test, has is been added to github.com/dataplat/appveyor-lab ?
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

See issue for details.

PR #10201, opened 2026-03-01
Test-DbaLsnChain: Fix bug when reading history from file

andreasjordan:
Fixes bug introduced by #10185

Tests use `$BackupHistory = Import-Clixml $PSScriptRoot\..\tests\ObjectDefinitions\BackupRestore\RawInput\CleanFormatDbaInformation.xml` which returns Deserialized.System.Numerics.BigInteger. So we now use the pattern `[BigInt]$xyzLsn.ToString()` everywhere.


andreasjordan:
Now the test passes. This should be merged to development and back into all open pull requests if all tests in development pass as well.

Update: Needs only to be merged into #10199 and #10200.

potatoqualitee:
will do thank you!

PR #10209, opened 2026-03-07
Test-DbaKerberos: Remove CNAME test

### LINKED ISSUE #10080: Test-DbaKerberos: Issue with CNAME Test
This is about "Check 10: Check for CNAME records".

No error, but missleading or irelevant information.

Used test code:
```
$result = nslookup $hostname 2>&1 | Out-String
if ($result -match "canonical name") {
    return "CNAME"
} else {
    return "A"
}
```

This most likely always returns "A", as this is an example output from nslookup on an english windows server 2025:
```
PS C:\Users\Admin.ordix> nslookup fci01
Server:  UnKnown
Address:  192.168.3.10

Name:    fci01.ordix.local
Address:  192.168.3.71

PS C:\Users\Admin.ordix> nslookup app01
Server:  UnKnown
Address:  192.168.3.10

Name:    fci01.ordix.local
Address:  192.168.3.71
Aliases:  app01.ordix.local
```

I have setup my lab so that FCI01 is a Failover Cluster Instance on SQL01 and SQL02 with its own virtual host and ip address including an A record. APP01 is a CNAME pointing to FCI01 that I used for tests with `Test-DbaKerberos`.

So would have to use `Resolve-DnsName` which outputs:
```
PS C:\Users\Admin.ordix> Resolve-DnsName -Name fci01

Name                                           Type   TTL   Section    IPAddress
----                                           ----   ---   -------    ---------
fci01.ordix.local                              A      1200  Answer     192.168.3.71


PS C:\Users\Admin.ordix> Resolve-DnsName -Name app01

Name                           Type   TTL   Section    NameHost
----                           ----   ---   -------    --------
app01.ordix.local              CNAME  3600  Answer     fci01.ordix.local

Name       : fci01.ordix.local
QueryType  : A
TTL        : 1200
Section    : Answer
IP4Address : 192.168.3.71
```

But there is another problem: `Test-DbaKerberos` gets the name of the test target from this code: 
```
$server = Connect-DbaInstance -SqlInstance $target -SqlCredential $SqlCredential
$computerTarget = $server.ComputerName
```
Which in this case of `$server = Connect-DbaInstance -SqlInstance APP01 ; $server.ComputerName` returns `FCI01`. So even if we use a CNAME to connect, we test always the A record in every test. So we would never get a CNAME result. Or are there cases, where `$server.ComputerName` would return a name that has no A record in the DNS?

Next question: What is the test all about? I have heard "CNAME kills Kerberos", but in my lab, `Test-DbaConnectionAuthScheme -SqlInstance APP01` returns "Kerberos". Is my lab configured differently than most production networks? And if we want to test if "APP01" is a CNAME, we need to change the way we set `$computerTarget` - at least for this test.

andreasjordan:
Fixes #10080 

Removed this test as not useful. See issue for details.

### DIFF
--- a/public/Test-DbaKerberos.ps1
+++ b/public/Test-DbaKerberos.ps1
@@ -16,10 +16,9 @@ function Test-DbaKerberos {
         - Client-Server time synchronization (5-minute Kerberos threshold)
         - Server-DC time synchronization
 
-        DNS (3 checks):
+        DNS (2 checks):
         - Forward lookup verification
         - Reverse lookup verification
-        - CNAME detection (CNAMEs break Kerberos)
 
         Service Account (3 checks):
         - Service account type validation (gMSA, domain account, built-in accounts)
@@ -174,8 +173,6 @@ function Test-DbaKerberos {
 
                 Write-Message -Level Verbose -Message "Starting Kerberos diagnostics for $target"
 
-                #region Tier 1 Checks - Essential & Straightforward
-
                 #region SPN Checks
                 # Check 1: Run Test-DbaSpn
                 try {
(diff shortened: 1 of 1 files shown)


potatoqualitee:
thank you! 💯

PR #10222, opened 2026-03-09
Invoke-DbaDbDataMasking: Fix Deterministic masking not applied with multiple columns

### LINKED ISSUE #7503: [Bug] Invoke-DbaDbDataMasking not honoring "Persisted" property value on config file
### Report
After generating a DataMaskingConfig file we can edit it to change some property values. I was recently playing with the `Deterministic` property, which uses the same masked value on all matches.

Assume I have the persisted property of the JSON config file with the value of "true".

1. When my config has just one column, the persisted value is honoured, and the values being masked are equal. ✔️  
2. IF my config file has 2 columns (both with the persisted property set to `true`, none of them is honoured. ❌ 

#### Host used

- [x] powershell.exe
- [x] VS Code

#### Errors Received
No errors.

#### Steps to Reproduce
1. Generate a new config with 2 columns
``` powershell
New-DbaDbMaskingConfig -SqlInstance localhost -Database AdventureWorksLT2016 -Table Address -Column City, PostalCode -Path "d:\temp"
```
2. Edit the generated configuration file and change the persisted property to `true` on both occurrences.
3. Use the saved config file on `Invoke-DbaDbDataMasking`
``` powershell
Invoke-DbaDbDataMasking -SqlInstance localhost -Database AdventureWorksLT2016 -FilePath "D:\temp\localhost.AdventureWorksLT2016.DataMaskingConfig.json"
```

#### Expected Behavior
Anonymized values should be equal when values were equal before.

#### Actual Behavior
Each record is getting a value.

image

### Environmental information

PowerShell Version : 5.1.19041.1023
dbatools latest installed : 1.0.168
Culture of OS : en-US

#### SQL Server:

```sql
Microsoft SQL Server 2016
```

```sql
us_english
```

andreasjordan:
Fix for issue #7503 where the Deterministic property was not honored when a config had 2+ columns.

A spurious `continue` statement after finding a deterministic lookup value caused the column to be skipped when building the UPDATE statement. The looked-up value was found but never written back to the database.

Fixes #7503

### DIFF
--- a/public/Invoke-DbaDbDataMasking.ps1
+++ b/public/Invoke-DbaDbDataMasking.ps1
@@ -883,8 +883,6 @@ function Invoke-DbaDbDataMasking {
 
                                             if ($lookupResult.NewValue) {
                                                 $newValue = $lookupResult.NewValue
-                                                # Skip further processing for this column
-                                                continue
                                             }
                                         } catch {
                                             Stop-Function -Message "Something went wrong retrieving the deterministic values" -Target $query -ErrorRecord $_ -continue


potatoqualitee:
good catch ty

PR #10229, opened 2026-03-11
Remove Invoke-SmoCheck - no longer needed

### LINKED ISSUE #10114: Do we need Invoke-SmoCheck?
Do we still need `Invoke-SmoCheck` @potatoqualitee ?

andreasjordan:
Removes the `Invoke-SmoCheck` private function and all call sites.

The SMO version check only fired once per session due to a script-scope flag, making it unreliable. It was only called from two migration commands and only issued warnings without blocking execution.

Closes #10114


potatoqualitee:
noice ty

Issue #9911, opened 2025-11-02
Implement Always On Policies as powershell code


andreasjordan:
### Summarize Functionality

I want to implement those policies with PowerShell:

https://learn.microsoft.com/en-us/sql/database-engine/availability-groups/windows/always-on-policies-for-operational-issues-always-on-availability

For every policy I would like to have PowerShell code that uses the new command `Get-DbaAgDatabaseReplicaState` from pr #9910 and implements the test logic to tell if the policy is in a healthy state or an unhealthy state.

### Is there a command that is similiar or close to what you are looking for?

No

andreasjordan:
@claude can you help me with this issue?

andreasjordan:
Claude did an ok job with a lot of good code to start with.

But the branch is not meant to be merged as we don't want examples but code for `Test-DbaAvailabilityGroup`.

I will use the code in the next days to implement one or two policys as a reference. I will add the code to `Test-DbaAvailabilityGroup` after #9910 is merged.

andreasjordan:
The code does not use `Get-DbaAgDatabaseReplicaState` at all.

andreasjordan:
Having a deeper look at the code and running the code in my lab I changed my minde: Claude did not very well. There will be a lot of extra work. Maybe this was to complicated for the AI...

andreasjordan:
Better source:

```
PS C:\Users\Admin> Get-DbaPbmCondition -SqlInstance server-006 -IncludeSystemObject | Where-Object Name -Match AlwaysOn | Format-List -Property Name, Facet, ExpressionNode


Name           : AlwaysOnAgAutomaticFailoverHealthCondition
Facet          : IAvailabilityGroupState
ExpressionNode : (@IsAutoFailover = True() AND @NumberOfSynchronizedSecondaryReplicas > 0) OR @IsAutoFailover = False()

Name           : AlwaysOnDbrJoinStateCondition
Facet          : DatabaseReplicaState
ExpressionNode : @IsJoined = True()

Name           : AlwaysOnAgWSFClusterHealthCondition
Facet          : Server
ExpressionNode : @ClusterQuorumState = Enum('Microsoft.SqlServer.Management.Smo.ClusterQuorumState', 'NormalQuorum')

Name           : AlwaysOnArConnectionHealthCondition
Facet          : AvailabilityReplica
[...]

PR #10233, opened 2026-03-14
Test-DbaAvailabilityGroup: Add policy based tests

andreasjordan:
This is an example on how we could add policy based tests like suggested in #9911.

The code is based on https://github.com/dataplat/dbatools/issues/9911#issuecomment-3478068521 and was created by the help of my local GitHub Copilot.


andreasjordan:
See my thoughts about this in https://github.com/dataplat/dbatools/issues/9911 - this is not a pr to be merged.

PR #10235, opened 2026-03-14
Set-DbaPrivilege, Get-DbaPrivilege - Add CreateGlobalObjects privilege support

### LINKED ISSUE #9173: Add 'Create global objects' to Set-DbaPrivilege
### Summarize Functionality

It would be great to add a new type of 'Create global objects' to the Set-DbaPrivilege cmdlet, which would allow adding an account to the "Create global objects" item in the local security policy. This access is required for certain backup solution for their SQL backup agent. I figured it would be usefull to be able to configure this using dbatools.

We're using Dell PowerProtect Data Manager, but i've also seen it in documentations for RedGate, Veritas and others.

Yes

### Technical Details

This should be fairly easy to implement by adding a new  'CreateGlobalObjects' to the -Type Parameter set, and add the code below in the process block that detects the type selected.

```
if ('CreateGlobalObjects' -in $Type) {
	$CGOline = Get-Content $tempfile | Where-Object { $_ -match "SeCreateGlobalPrivilege" }
	ForEach ($acc in $SQLServiceAccounts) {
		$SID = Convert-UserNameToSID -Acc $acc;
		if (-not $CGOline) {
			$CGOline = "SeCreateGlobalPrivilege = *$SID"
			(Get-Content $tempfile) -replace "\[Privilege Rights\]", "[Privilege Rights]`n$CGOline" |
				Set-Content $tempfile
			<# DO NOT use Write-Message as this is inside of a script block #>
			Write-Verbose "Added $acc to Create Global Objects Privileges on $env:ComputerName"
		} elseif ($CGOline -notmatch $SID) {
			(Get-Content $tempfile) -replace "SeCreateGlobalPrivilege = ", "SeCreateGlobalPrivilege = *$SID," |
				Set-Content $tempfile
			<# DO NOT use Write-Message as this is inside of a script block #>
			Write-Verbose "Added $acc to Create Global Objects Privileges on $env:ComputerName"
		} else {
			<# DO NOT use Write-Message as this is inside of a script block #>
			Write-Verbose "$acc already has Create Global Objects Privilege on $env:ComputerName"
		}
	}
}
```

andreasjordan:
Closes #9173

## Summary

- Add `CreateGlobalObjects` (`SeCreateGlobalPrivilege`) to both `Set-DbaPrivilege` and `Get-DbaPrivilege`
- Required by backup agents from Dell PowerProtect, Redgate, Veritas, and others
- Fix SID-prepend bug in all existing `elseif` blocks (now appends new SIDs instead of prepending)
- Rename generic `$Line` variable to `$SALine` in SecAudit block for naming consistency

## Test plan

- [ ] Unit test parameter validation passes for both commands
- [ ] `Set-DbaPrivilege -Type CreateGlobalObjects` grants `SeCreateGlobalPrivilege` on target machine
- [ ] `Get-DbaPrivilege` returns `CreateGlobalObjects` column with correct boolean values
- [ ] Running `Set-DbaPrivilege` twice for same account is idempotent (no duplicate SIDs)

### DIFF
--- a/public/Set-DbaPrivilege.ps1
+++ b/public/Set-DbaPrivilege.ps1
@@ -255,7 +255,7 @@ function Convert-UserNameToSID ([string] `$Acc ) {
                                     }
                                 }
                                 $null = secedit /configure /cfg $tempfile /db secedit.sdb /areas USER_RIGHTS /overwrite /quiet
-                            } -ErrorAction SilentlyContinue
+                            }
                             Write-Message -Level Verbose -Message "Removing secpol file on $computer"
                             Invoke-Command2 -Raw -ComputerName $computer -Credential $Credential -ScriptBlock { $temp = ([System.IO.Path]::GetTempPath()).TrimEnd(""); Remove-Item $temp\secpolByDbatools.cfg -Force > $NULL }
                         } else {


andreasjordan:
I will look at this later and maybe add some of the suggested changes.

andreasjordan:
@claude There is a bug in Get-DbaPrivilege: The function Convert-SIDToUserName is not run on the target maschine. All the SIDs from the local service account can not be converted and show up as SIDs in the output. Set-DbaPrivilege does the convertion on the target system. Can you implement a fix for Get-DbaPrivilege to do the conversion on  the target like Set-DbaPrivilege does?

andreasjordan:
The commands now work as expected in my lab.

potatoqualitee:
thank you kindly 🙏🏼

PR #10240, opened 2026-03-14
Get-DbaLastBackup - Add -ExcludeReplica switch for AlwaysOn preferred backup replica filtering

### LINKED ISSUE #6525: Get-DBALastBackup to only include preferred replica
### Summary of new feature

The problem is that when running Get-DbaLastBackup against a server list that has databases configured for AlwasyOn it shows secondary replica's as not being backed up.

I am thinking of a switch something like -excludereplica

a check to see if SERVERPROPERTY 'IsHadrEnabled'  is  = 1, the preferred replica for backup in order to determine which databases are missing a backup.

Get-DbaLastBackup -SQLInstance sql2019 -excludereplica

v 1.0.108

andreasjordan:
Adds a -ExcludeReplica switch to Get-DbaLastBackup that uses sys.fn_hadr_backup_is_preferred_replica to skip AG databases where the current instance is not the preferred backup replica. This eliminates false positives when running against multiple AG members where secondaries appear to have no backups.

Fixes #6525

### DIFF
--- a/public/Get-DbaLastBackup.ps1
+++ b/public/Get-DbaLastBackup.ps1
@@ -25,6 +25,12 @@ function Get-DbaLastBackup {
         Excludes specific databases from the backup history check. Commonly used to skip system databases or test databases.
         Use this when you want to check most databases but exclude certain ones like tempdb, development databases, or databases with known backup exemptions.
 
+    .PARAMETER ExcludeReplica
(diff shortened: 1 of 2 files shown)


potatoqualitee:
looks reasonable, thank you 🙏🏼

PR #10241, opened 2026-03-14
Test-DbaLastBackup - Add -Path parameter to test backups from folder paths

### LINKED ISSUE #6594: Test-DbaLastBackup use folder path as source instead of a Sql Instance
## Feature Request

I see this feature has been requested in the past for a couple of times:
#2285 and #981

Nothing has been done with it, though.
Only workarounds were suggested.

With DbaTools being awesome af, I think it's high time to implement this feature and make it widely accessible. No workarounds and no excuses.

From what I gather, the relevant code that should be affected by this feature is the following line:

https://github.com/sqlcollaborative/dbatools/blob/a459d07637d2fe66c5f70f0174750203aaff37d9/functions/Test-DbaLastBackup.ps1#L374

i.e. instead of getting the `$restorelist` from `$lastbackup[0].Path`, it should get it from a folder input similar to the `$Path` parameter in `Restore-DbaDatabase`.

My guess is that references to `$lastbackup[0].Path` should be replaced with a more generic variable, whose contents would depend on whether we're using a Sql Instance or a folder as the source.

Thank you for your awesomeness!

andreasjordan:
Adds support for testing backup files from a folder path instead of only querying backup history from a SQL instance.

Closes #6594

Needs some changes to casing of variables and some testing...

### DIFF
--- a/public/Test-DbaLastBackup.ps1
+++ b/public/Test-DbaLastBackup.ps1
@@ -333,6 +333,8 @@ function Test-DbaLastBackup {
             $InputObject += Get-DbaDatabase -SqlInstance $SqlInstance -SqlCredential $SqlCredential -Database $Database -ExcludeDatabase $ExcludeDatabase
         }
 
+        $workItems = New-Object System.Collections.Generic.List[hashtable]
+
         if (Test-Bound "Path") {
             if (-not (Test-Bound "Destination")) {
                 Stop-Function -Message "A -Destination server must be specified when using -Path to specify backup folder paths."
@@ -400,19 +402,13 @@ function Test-DbaLastBackup {
             $pathDatabaseGroups = $allPathBackups | Group-Object -Property Database
 
             foreach ($pathDbGroup in $pathDatabaseGroups) {
-                $dbName = $pathDbGroup.Name
+                $dbName     = $pathDbGroup.Name
                 $lastbackup = $pathDbGroup.Group
-                $source = "N/A"
-                $restoreresult = $null
-                $fileexists = $true
-                $dbccresult = $null
-                $success = $null
-                $errormsg = $null
 
                 if (-not ($lastbackup | Where-Object { $_.Type -eq "Full" -or $_.Type -eq "Database" })) {
                     Write-Message -Level Verbose -Message "No full backup found for $dbName in the specified path(s)."
                     [PSCustomObject]@{
-                        SourceServer   = $source
+                        SourceServer   = "N/A"
                         TestServer     = $Destination
                         Database       = $dbName
                         FileExists     = $false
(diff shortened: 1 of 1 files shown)


andreasjordan:
@claude please add the new parameter to the test file.

andreasjordan:
Test failes with:

```
Path           : Test-DbaLastBackup/Test restore using -Path parameter/Should return success when using -Path
Name           : It Should return success when using -Path
Result         : Failed
Message        : ErrorRecord: Expected 'Success', but got $null.
StackTrace     : 
RawErrorRecord : Expected 'Success', but got $null.
Path           : Test-DbaLastBackup/Test restore using -Path parameter/Should return the correct database name
Name           : It Should return the correct database name
Result         : Failed
Message        : ErrorRecord: Expected 'dbatoolsci_testlastbackup932788033', but got $null.
StackTrace     : 
RawErrorRecord : Expected 'dbatoolsci_testlastbackup932788033', but got $null.

```

Will test this today.

andreasjordan:
@claude there is a bug in this implementation. When using -Path, the variable `$lastbackup` in line 410 contains information in a different format and the property Type has the value "Database" instead of "Full". Please have a look at this and add a fix to this pull request if possible.

andreasjordan:
Manual tests in my lab now work.

andreasjordan:
Test on Appveyor still fail and report:

```
[12:11:27][Get-DbaBackupInformation] Failure on APPVEYOR-VM\sql2022 | Cannot open backup device 'C:\Temp\Test-DbaLastBackup-396544843\dbatools-testrestore-dbatoolsci_singlerestore.mdf'. Operating system error 32(The process cannot access the file because it is being used by another process.).
RESTORE HEADERONLY is terminating abnormally.
```

andreasjordan:
Moved test up before the test that creates a database in the backup location.

andreasjordan:
@claude This pull request added a lot a duplicate code which should be avoided. Do you see a way to reduce the duplicate code? Please do not add any commits here, just analyze.

andreasjordan:
@claude let's first solve the issues with $datadirectory and $logdirectory. First of all, they are parameters so they should be named as $DataDirectory and $LogDirectory. And as they could be different on different source servers, they should not be set if the parameter is not used. So we need to use $usedDataDirectory / $usedLogDirectory (or a better name from "used" if you have one) and must not change the original parameter values. Please implement a fix for this issue.

andreasjordan:
@claude now let's get back to your refactoring approach to deduplicate the code. Are there any open challenges? If not, please implement your approach.

andreasjordan:
The command works as expected in my lab.

potatoqualitee:
Works, thank you 👍🏼

PR #10243, opened 2026-03-14
Read-DbaXEFile, Get-DbaXESessionTargetFile - Document Windows-only admin share requirement and full pipeline usage

### LINKED ISSUE #9615: Let's Document How to use Read-DbaXEFile Without a UNC Share
### Summarize Functionality

Currently, the only documentation for `Read-DbaXEFile` that does not require knowing the path to the .xel file is `Get-DbaXESession -SqlInstance sql2019 -Session deadlocks | Read-DbaXEFile`. I've never seen this work, presumably because the documentation strongly suggests that it only works if you write your .xel file to a file share and I've never seen anyone do that.

I propose that we document how to use `Read-DbaXEFile` without knowing the file path and without needing a UNC share. The only trick that I know for this is by putting many extra steps in the pipeline. For example,
```
`Get-DbaXESession -SqlInstance sql2019 -Session deadlocks |
Get-DbaXESessionTarget |
Get-DbaXESessionTargetFile |
Read-DbaXEFile
```
works consistently for me, but requires some permissions that I have yet to fully understand.

Yes

### Technical Details

As above. The only thing to add is that it helps to stop the session to force a rollover.

andreasjordan:
Addresses three documentation gaps raised in #9615:

1. Add example showing the full pipeline to read XE files by session name without knowing the path
2. Document the Windows-only restriction (admin shares; does not work on Linux/Docker)
3. Clarify that SqlCredential alone is not sufficient — PowerShell must run as a Windows account with file-system access

fixes #9615


ReeceGoding:
I believe that this is exactly what I had in mind. Thanks! My only objection is that somebody with more skill and experience than I should check it over. For example, I suspect that less than admin permission will suffice in many cases.

potatoqualitee:
Last one for the night! Thank you 🙏🏼

PR #10234, opened 2026-03-14
Test-DbaAvailabilityGroup - Add -Policy switch for all 13 Always On policies

andreasjordan:
Adds a `-Policy` switch to `Test-DbaAvailabilityGroup` that evaluates all 13 predefined Always On Availability Group policies from Microsoft's Policy-Based Management framework.

Each policy check returns an object with PolicyName, Category, Facet, Status (Healthy/Unhealthy), Issue, and Details.

Related to #9911

Generated with [Claude Code](https://claude.ai/code)

PR #10246, opened 2026-03-14
Test-DbaAgPolicyState - Add new command for Always On policy state checks

andreasjordan:
Creates a new `Test-DbaAgPolicyState` command that evaluates SQL Server Availability Groups against Microsoft's predefined Always On policies.

Implements the first 7 of 13 predefined policies (server and AG facets) as a starting point, all using the `AvailabilityGroupState` SMO object.

Related to #9911

### DIFF
--- a/public/Test-DbaAgPolicyState.ps1
+++ b/public/Test-DbaAgPolicyState.ps1
@@ -59,11 +59,13 @@ function Test-DbaAgPolicyState {
         - InstanceName: The SQL Server instance name (string)
         - SqlInstance: The full SQL Server instance name (string)
         - AvailabilityGroup: Name of the availability group (string)
+        - Replica: Name of the availability replica for replica-level policies, $null for server/AG-level policies (string)
+        - Database: Name of the availability database for database-level policies, $null for server/AG/replica-level policies (string)
         - PolicyName: The name of the Always On policy being evaluated (string)
         - Category: The severity of the policy - Critical or Warning (string)
         - Facet: The object type the policy applies to (string)
         - IsHealthy: Boolean indicating whether the policy check passed (bool)
-        - Issue: Description of the issue when the policy is not healthy (string)
+        - Issue: Description of the issue when the policy is not healthy, $null when healthy (string)
         - Details: Additional detail about the current state (string)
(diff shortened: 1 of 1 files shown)


andreasjordan:
I will close #10233 and #10234 and @ReeceGoding and I will try to use this pull request as a starting point.

I will test this tomorrow in my lab.

@ReeceGoding feel free to test as well and to add comments here with things we should change.

andreasjordan:
"Issue" is always filled with the suitable text if IsHealthy is $false. Sould we set "Issue" to $null if IsHealthy is $true?

andreasjordan:
The parameters $Secondary and $SecondarySqlCredential are currently not used. If they are never used, we should remove them.

ReeceGoding:
Played around with it. It is definitely close to the right thing. This is going to be good!

image

## Initial thoughts

- It is currently returning everything, even the healthy stuff.
- I wonder if the WSFC Cluster State test is dishonest. My lab is clusterless, but this tests reports that my cluster is healthy.
- I agree with you that `Issue` should be blank when there is no issue. `Details` already tells us anything else we would want to know.

## Breaking it

When I deliberately tried to break it, I was surprised to see that pulling the plug on my secondary did not cause any issues to be flagged.

image

I suspect that a `.Refresh()` is missing somewhere.

image

Come to think of it, `docker stop` would be a good tool for integration tests. My others ideas for them have been lacking.

andreasjordan:
@claude please read our comments. Then implement all the over policies, but also only fill the Issue property in case there is an issue.

andreasjordan:
I just tested with `SuspendDataMovement()`:

image

andreasjordan:
I think it should return also the healthy stuff like other `Test-` commands do as well.

andreasjordan:
The ClusterQuorumState is maybe missleading. In my test I use `-ClusterType None`. When I have time I will setup a new windows cluster and test the command there.

andreasjordan:
But I think this is a good starting point and ready to be published so people can test this in their enviroments.

potatoqualitee:
thank you 🙏🏼

PR #10258, opened 2026-03-17
Invoke-DbaDbShrink - Add error message output for failed shrink operations

### LINKED ISSUE #10033: Invoke-DbaDbShrink: let files be shrunk evenly and say why they don't shrink
### Summarize Functionality

I often use `Invoke-DbaDbShrink` for tempdb, but:

1. I am not aware of any way to get `Invoke-DbaDbShrink` to shrink files evenly
2. Unlike the T-SQL that `Invoke-DbaDbShrink` calls, `Invoke-DbaDbShrink` does not print out why my shrink has failed (which often happens in tempdb)

Both of these would help a lot for tempdb operations, which are just about the only sensible example of shrinking a database.

Yes

### Technical Details

I'm open to the idea of this being a new command. Shrinking tempdb is a common enough operation on small dev servers that I would love to either have a super easy way to do it or at least an example in the documentation that I can just copy and paste.

andreasjordan:
## Summary

Enhances `Invoke-DbaDbShrink` to output detailed error messages when shrink operations fail or don't achieve their target size. This addresses issue #10033 about tempdb shrink operations that commonly fail or only partially complete.

## Changes

- Capture detailed exception messages including inner exceptions
- Display warning messages when shrink operations fail
- Detect partial shrinks (common in tempdb with active transactions)
- Include error details in the Notes property of output objects

Fixes #10033

### DIFF
--- a/public/Invoke-DbaDbShrink.ps1
+++ b/public/Invoke-DbaDbShrink.ps1
@@ -329,12 +329,7 @@ function Invoke-DbaDbShrink {
                         } catch {
                             $success = $false
                             $errorDetails = $_.Exception.Message
-                            if ($_.Exception.InnerException) {
-                                $errorDetails += " Inner exception: $($_.Exception.InnerException.Message)"
-                            }
-                            Write-Message -Level Warning -Message "Shrink operation failed for file $($file.Name): $errorDetails"
-                            Stop-Function -Message "Shrink failed: $errorDetails" -EnableException $EnableException -ErrorRecord $_ -Continue
-                            continue
+                            Stop-Function -Message "Shrink operation failed for file $($file.Name): $errorDetails" -ErrorRecord $_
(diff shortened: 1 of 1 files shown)


andreasjordan:
This PR is not perfect, Claude has made some big mistakes. But the initial code was not perfect as well.

There is a `$success`, but in case of failure, the `-Contine` prevented the output. So the new error message is not shown in the output.

I will change the code to always return an output and so display the error message....

Only if EnableException is used, then I still raise an exception.

andreasjordan:
Waiting for @ReeceGoding to give feedback.

ReeceGoding:
@andreasjordan I agree with you. It's not perfect but it certainly looks to be an improvement.
image

potatoqualitee:
thank you both 🙇🏼

Issue #10256, opened 2026-03-17
Sync-DbaAvailabilityGroup DAC Connection Issue;  sync login permissions


AndreasEDA:
### Verified issue does not already exist?

I have searched and found no existing issue

### What error did you receive?

WARNING: [10:07:37][Sync-DbaLoginPermission] Issue syncing permissions for login | An existing connection was forcibly closed by the remote host

### Steps to Reproduce

-Executed on Primary Host:
Sync-DbaAvailabilityGroup -Primary XXXX -AvailabilityGroup AG1

### Please confirm that you are running the most recent version of dbatools

find-psresource -name dbatools:  

dbatools 2.7.26             

### Other details or mentions

In the sql Log i see a lot of these:

_Could not connect because the maximum number of '1' dedicated administrator connections already exists. Before a new connection can be made, the existing dedicated administrator connection must be dropped, either by logging off or ending the process.  Error 17810_

->Sync-DbaLoginPermissions works ok without error or warning:
Sync-DbaLoginPermission -Source XXXX-Destination YYYYY

->So Sync-DbaAvailabilityGroup invoke Sync-DbaLoginPermissions multiple times and it seems it has a issue with DAC Connection not closing or reusing correctly. Or is it a time problem? I have this issue since a few weeks on all SQL Instances

### What PowerShell host was used when producing this error

Windows PowerShell (powershell.exe)

### PowerShell Host Version

5.1.20348.4294

### SQL Server Edition and Build number

SQL Enterprise 16.0.4240.4

### .NET Framework Version

4.8.04161

AndreasEDA:
Running with -excludePassword Parameter is working fine

Sync-DbaAvailabilityGroup -Primary XXXXXX -AvailabilityGroup AG1 -ExcludePassword

mbentham:
Same issue as #10259 issue with DAC connection cycling in Connect-DbaInstance - I've been meaning to raise that bug for weeks, finally get round to it and you've beaten me to it by 3 hours 🤣

andreasjordan:
Solved by https://github.com/dataplat/dbatools/issues/10256

PR #10260, opened 2026-03-17
Connect-DbaInstance - Set NonPooledConnection on ServerConnection for…

### LINKED ISSUE #10259: Connect-DbaInstance - DedicatedAdminConnection causes SQL Server error 17810 due to missing NonPooledConnection on ServerConnection

When using Connect-DbaInstance -DedicatedAdminConnection, the resulting server object causes SQL Server to log error 17810 ("maximum number of dedicated administrator connections already exists") during normal operations. The script succeeds, but the error log fills with critical entries from rapid DAC connect/disconnect cycling.

### Steps to Reproduce

```powershell
$server = Connect-DbaInstance -SqlInstance localhost -DedicatedAdminConnection
Invoke-DbaQuery -SqlInstance $server -Query "SELECT 1"
$server | Disconnect-DbaInstance
```

v2.7.26

### Expected Behavior
A single DAC connection is opened by Connect-DbaInstance, reused by Invoke-DbaQuery, and closed by Disconnect-DbaInstance. No errors in the SQL Server log.

### Actual Behavior
SQL Server logs error 17810:

> Critical  Logon  Error: 17810, Severity: 20, State: 1.
> Could not connect because the maximum number of '1' dedicated administrator
> connections already exists. Before a new connection can be made, the existing
> dedicated administrator connection must be dropped, either by logging off or
> ending the process. [CLIENT: 127.0.0.1]

The script completes successfully despite the error, but the error log entry is misleading for anyone monitoring for genuine connection issues.

### Root Cause

In `Connect-DbaInstance`, the String input path (lines ~942 and ~1036-1077) sets `SqlConnectionInfo.Pooled = $false` for DAC connections, which puts `Pooling=false` in the connection string. However, it never sets `ServerConnection.NonPooledConnection = $true` on the resulting `ServerConnection`/`ConnectionContext`.

These are two independent settings:

| Setting | What it controls |
|---|---|
| `SqlConnectionInfo.Pooled = $false` | Puts `Pooling=false` in the connection string - tells SqlClient not to pool physical connections |
| `ServerConnection.NonPooledConnection = $true` | Tells SMO to keep the connection **open** between operations instead of auto-disconnecting |

Without `NonPooledConnection = $true`, the `ServerConnection` auto-disconnects after every SMO operation (the initial `ExecuteWithResults` validation query, each lazy-loaded property access for `DatabaseEngineType`, `VersionMajor`, `HostPlatform`, `NetName`, etc.). Because `Pooling=false`, each auto-disconnect destroys the physical TCP connection rather than returning it to a pool.

The result is a rapid cycle of DAC open/close/open/close for every property access. When SQL Server hasn't fully released the DAC slot from one close before the next open arrives, it rejects the new connection with error 17810.

I am planning to raise a PR to fix this.

This may be the same bug as #10256 

### PowerShell Host Version

5.1.26100.7920

Microsoft SQL Server 2019 (RTM-CU32-GDR) (KB5068404) - 15.0.4455.2

### .NET Framework Version

4.8.09032

mbentham:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #10259, #10256 )
 - [x] Ran manual Pester test and has passed (`Invoke-ManualPester`)
 - [ ] Pester test is included
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
Fix SQL Server error 17810 ("maximum number of dedicated administrator connections already exists") when using `Connect-DbaInstance -DedicatedAdminConnection`. The DAC connection is not kept open between SMO operations because `ServerConnection.NonPooledConnection` is never set in the String input path, causing rapid connect/disconnect cycling that races against SQL Server's DAC slot release.

### Approach
Added `$server.ConnectionContext.NonPooledConnection = $true` in the String input path of `Connect-DbaInstance`, alongside the existing `ConnectionContext` property assignments (BatchSeparator, LockTimeout, etc.). This is gated on `$NonPooledConnection` which covers both `-DedicatedAdminConnection` and explicit `-NonPooledConnection` usage.

The same pattern already exists in the Server/copy-context path (line ~711) and in `New-DbaConnectionString` (line ~465) — this change brings the String input path in line with both.

### Commands to test
```powershell
# Clear the error log first so results are unambiguous
Invoke-DbaQuery -SqlInstance localhost -Query "EXEC sp_cycle_errorlog"

# Connect via DAC, run a query, disconnect
$server = Connect-DbaInstance -SqlInstance localhost -DedicatedAdminConnection
Invoke-DbaQuery -SqlInstance $server -Query "SELECT @@servername"
$server | Disconnect-DbaInstance

# Check SQL Server error log for 17810 - should return no rows
Invoke-DbaQuery -SqlInstance localhost -Query "EXEC xp_readerrorlog 0, 1, N'17810'"
```

**Before fix:** `xp_readerrorlog` returns rows showing error 17810 logged during the DAC session (sometimes, depending on how quickly your particular SQL Server cleaned up the closing DAC connection - its an intermittant race condition).
**After fix:** No 17810 entries — the DAC connection stays open for the duration instead of cycling.

### Learning
- `SqlConnectionInfo.Pooled` and `ServerConnection.NonPooledConnection` are independent settings. `Pooled = $false` controls SqlClient connection pooling (`Pooling=false` in the connection string), while `NonPooledConnection = $true` controls whether SMO auto-disconnects after each operation. Both are required for DAC connections to work correctly.

### DIFF
--- a/public/Connect-DbaInstance.ps1
+++ b/public/Connect-DbaInstance.ps1
@@ -1075,6 +1075,10 @@ function Connect-DbaInstance {
                 }
                 Write-Message -Level Debug -Message "Setting ConnectionContext.StatementTimeout to '$StatementTimeout'"
                 $server.ConnectionContext.StatementTimeout = $StatementTimeout
+                if ($NonPooledConnection) {
+                    Write-Message -Level Debug -Message "Setting ConnectionContext.NonPooledConnection to 'True'"
+                    $server.ConnectionContext.NonPooledConnection = $true
+                }
             }
 
             $maskedConnString = Hide-ConnectionString $server.ConnectionContext.ConnectionString


andreasjordan:
I can run your "commands to test" even without your fix. So I don't see the problem.

image

mbentham:
Yeah sorry I should have been clearer its a tricky one - the DBAtools commands run fine with no errors in powershell without the fix, but theres a race condition caused by cycling the connection that causes the SQL Server instance to log error 17810 "maximum number of dedicated administrator connections already exists" to the SQL Server log file if the the DAC connection isnt cleaned up quickly enough by SQL Server before the next DBATools command is sent, often the DBAtools command _still_ succeeds anyway due to the automatic retry in SqlClient that SMO, and therefore the Connect-DbaInstance tool inherits. But if anyone is checking the sql server error logs, or has a monitoring tool that flags server errors they'll see the errors caused by this module.

I've improved the commands to test section

andreasjordan:
Ok, so I need to create a loop and open/close the connection multiple times to see the problem?

But I still see no change, as `$server.ConnectionContext.NonPooledConnection = $true` has no effect as `$server.ConnectionContext.NonPooledConnection` is already $true without the change. Or am I missing something?

andreasjordan:
Ok, with a loop, I can reproduce the issue:
```
foreach ($i in 1..1000) {
    $dac = Connect-DbaInstance -SqlInstance fci01 -DedicatedAdminConnection
    $null = Invoke-DbaQuery -SqlInstance $dac -Query "SELECT @@servername"
    $null = $dac | Disconnect-DbaInstance
}
```

And with your fix, the issue really seems to be resolved.

mbentham:
Brilliant thanks, I'll be sure to be clearer in the PR next time, save the confusion.

potatoqualitee:
Thank you so much! 🔥 🚒

PR #10270, opened 2026-03-20
Get-DbaDbOrphanUser - Skip SQL login orphan check for contained databases

### LINKED ISSUE #9866: Get-DbaDbOrphanUser and Contained Databases
When you run the command (Get-DbaDbOrphanUser) on an SQL instance with multiple contained databases, you get the result that the users are orphaned, but these are contained databases. Can you exclude this database by containedtype? 
Example:  $ncondb = Get-DbaDatabase -SqlInstance XXXXXXXX  | Where-Object {$_.Containmenttype -eq 'partial'} |SELECT Name
                Get-DbaDbOrphanUser -SqlInstance XXXXXXXX -ExcludeDatabase $ncondb

andreasjordan:
Users in contained databases (Partial or Full containment) authenticate directly to the database without server-level logins, so they should not be flagged as orphaned. Windows users are still checked regardless of containment type.

Fixes #9866

### DIFF
--- a/public/Get-DbaDbOrphanUser.ps1
+++ b/public/Get-DbaDbOrphanUser.ps1
@@ -6,6 +6,10 @@ function Get-DbaDbOrphanUser {
     .DESCRIPTION
         An orphan user is defined by a user that does not have their matching login. (Login property = "").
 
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Thank you 🙏🏼

PR #10273, opened 2026-03-20
Add Remove-DbaAgentJobSchedule cmdlet

### LINKED ISSUE #9687: Add functionality equivalent to sp_detach_schedule
### Summarize Functionality

We have a wealth of cmdlets for managing Agent Job schedules:

- New-DbaAgentSchedule creates them
- Remove-DbaAgentSchedule destroys them
- Set-DbaAgentSchedule updates them at the job level. You can also use it to disable schedules, but I'm [fairly sure](https://learn.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-update-schedule-transact-sql?view=sql-server-ver17#remarks) that SQL Server doesn't support disabling shared schedules for just one job.
- Set-DbaAgentJob attaches them.

But suppose that I wish detach a schedule from a job. Aside from writing the T-SQL yourself and using `Invoke-DbaQuery`, I know of no solution to this within dbatools.

The problem that made me discover this issues was as follows:

- Job A and job B share a schedule.
- Job A is bugged and must not run until it is fixed
- Strict rules prevent disabling or deleting job A
- I could not disable or delete the schedule for job A, because it is shared with job B and I didn't want to damage job B
- Detaching the shared schedule from job A was allowed, and therefore my only choice

If dbatools has an alternative solution, then I would love to hear it.

Yes

### Technical Details

A wrapper for `sp_detach_schedule` should do it, but I am not totally sure where it should be put. I see three options:

1. Make a new cmdlet, `Remove-DbaAgentJobSchedule`.
2. Add a new parameter to `Set-DbaAgentJob`, giving a list of schedules to remove.
3. Add a new parameter to `Set-DbaAgentSchedule`, giving a switch for something like `-Detach`.

I do not know if what we want already exists in SMO. I barely trust myself to check. [This](https://learn.microsoft.com/en-us/dotnet/api/microsoft.sqlserver.management.smo.agent.job.removesharedschedule?view=sql-smo-172) is the best I found.

andreasjordan:
Implements `sp_detach_schedule` equivalent: removes the association between a job and a schedule without deleting the shared schedule itself.

Closes #9687

### DIFF
--- a/dbatools.psd1
+++ b/dbatools.psd1
@@ -539,6 +539,7 @@
         'Remove-DbaAgentAlert',
         'Remove-DbaAgentJob',
         'Remove-DbaAgentJobCategory',
+        'Remove-DbaAgentJobSchedule',
         'Remove-DbaAgentJobStep',
         'Remove-DbaAgentOperator',
         'Remove-DbaAgentSchedule',
--- a/dbatools.psm1
+++ b/dbatools.psm1
@@ -555,6 +555,7 @@ if ($PSVersionTable.PSVersion.Major -lt 5) {
         'Remove-DbaAgentJob',
         'New-DbaAgentJobStep',
         'Set-DbaAgentJobStep',
+        'Remove-DbaAgentJobSchedule',
         'Remove-DbaAgentJobStep',
         'New-DbaAgentSchedule',
         'Set-DbaAgentSchedule',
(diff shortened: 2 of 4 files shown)


ReeceGoding:
Looks good to me! I'm very happy to see that this was possible without resorting to calling the SQL.

The only thing that I can think to add is tests with either multiple jobs or multiple schedules, but I'm pretty sure that the underlying code is solid.

potatoqualitee:
Looks good, thank you 🥳

PR #10276, opened 2026-03-21
Get-DbaWaitStatistic - Add ExcludeWaitType and IncludeWaitType parameters

### LINKED ISSUE #9250: Get-DbaWaitStatistic Should Allow Filtering In/Out Extra Waits
### Summarize Functionality

As it stands, the only control that users have over the waits considered by `Get-DbaWaitStatistic` is the `IncludeIgnorable` switch. If a user wants to add or remove a particular wait type by name, then the current method would be to either submit a pull request or faff around with the source code's SQL. Given that the list of waits to ignore is just a collection of strings, this feels needlessly restrictive. Users should be able to provide their own list of strings to either add or remove from the `$ignorable` variable.

A typical use case for what I'm suggesting would be the parallelism waits. Some people care about those and some people don't, so the control over which are ignored should be passed over to users. [Paul's script](https://www.sqlskills.com/blogs/paul/wait-statistics-or-please-tell-me-where-it-hurts/) which inspired `Get-DbaWaitStatistic` even mentions that users may want to comment out CXCONSUMER.

Yes

### Technical Details

I imagine that this would be very easy to implement, but I don't know what means would be most consistent with dbatools' architecture and interface. Regardless, it's ultimately just either a union or set difference of the user's list of strings and the `$ignorable` variable.

andreasjordan:
## Summary

Adds fine-grained filtering control for wait statistics:
- `ExcludeWaitType`: Exclude additional wait types beyond default ignorable list
- `IncludeWaitType`: Include specific wait types even if in ignorable list

This allows users to customize wait filtering (e.g., exclude CXPACKET/CXCONSUMER) without modifying source code. Maintains full backward compatibility.

Fixes #9250

## Examples

```powershell
# Exclude parallelism waits
Get-DbaWaitStatistic -SqlInstance sql2016 -ExcludeWaitType "CXPACKET", "CXCONSUMER"

# Include a specific wait that's normally ignored
Get-DbaWaitStatistic -SqlInstance sql2016 -IncludeWaitType "BROKER_RECEIVE_WAITFOR"
```

---


potatoqualitee:
straightforward, thank you 🥇

PR #10281, opened 2026-03-21
Invoke-DbaDbDataMasking - Fix StaticValue empty string fallback and FilterQuery in Actions

### LINKED ISSUE #7942: Invoke-DbaDbDataMasking - setting a static value and point out a an prepopulated randomized file.
### Summarize Functionality

When using the datmaking with a filterquery that is not respected
Would be nice to be able to use staticvalue parameter in the config-file

adding parameter StaticValue = "GDPR-Masking" works but sets every row in the table eg it does not respect the FilterQuery ? I think that might bee a bug? In my world it should respect the filtering query.
Also if trying to set the StaticValue = "" eg to blank out some values in a column. That does not work, it fallbacks to generating values.

I looked in the randomizer types file (en.randomizertypes.csv) there were a type called: 
Type,SubType
static, none

Another feature that would be nice would be to be able to point out a prepopulated table or file with already randomized data
https://shamdata.com/ (I did create a 25000 row file from there in like 30 seconds)

No

### Technical Details

_No response_

andreasjordan:
Fix two bugs in `Invoke-DbaDbDataMasking`:

- **StaticValue = "" falls back to random values**: Empty string is falsy in PowerShell, so truthiness checks failed to detect an intentionally empty static value. Replaced with explicit `$null -ne` checks.
- **FilterQuery not respected for Action-based updates**: Action processing generated UPDATE statements without any WHERE clause, updating all rows regardless of FilterQuery. Fixed by extracting the WHERE clause from FilterQuery and appending it to each Action UPDATE query.

Closes #7942

### DIFF
--- a/public/Invoke-DbaDbDataMasking.ps1
+++ b/public/Invoke-DbaDbDataMasking.ps1
@@ -842,7 +842,7 @@ function Invoke-DbaDbDataMasking {
                                 $newValue = $null
 
                                 # Handle static values
-                                if ($columnobject.StaticValue) {
+                                if ($null -ne $columnobject.StaticValue) {
                                     $newValue = $columnobject.StaticValue
 
                                     if ($null -eq $newValue -and -not $columnobject.Nullable) {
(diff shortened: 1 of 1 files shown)


potatoqualitee:
thank you 🙇🏼

PR #10282, opened 2026-03-22
Get-DbaAgRingBuffer - Add command for HADR ring buffer diagnostics

### LINKED ISSUE #9823: Expose the internal HA/DR ring buffers, for better health information about Always On availability groups

[This official documentation](https://learn.microsoft.com/en-us/sql/database-engine/availability-groups/windows/always-on-ring-buffers?view=sql-server-ver16) shows how to query the internal ring buffers for HA/DR. It describes them by saying
> The ring buffers are created during SQL Server startup, and record alerts within the SQL Server system for internal diagnostics. They aren't supported, but you can still extract valuable information from them when troubleshooting issues. These ring buffers provide another source of diagnostics when SQL Server stops responding or has crashed.

We should add this to our suite of commands for getting information about AGs.

Yes

### Technical Details

I imagine that our existing commands for hitting Extended Events may be of a value.

andreasjordan:
Implements `Get-DbaAgRingBuffer` to expose the internal HADR ring buffers for Always On AG diagnostics.

Queries `sys.dm_os_ring_buffers` for the four HADR-specific types: `RING_BUFFER_HADRDBMGR_API`, `RING_BUFFER_HADRDBMGR_STATE`, `RING_BUFFER_HADRDBMGR_COMMIT`, and `RING_BUFFER_HADR_TRANSPORT_STATE`.

Returns raw XML records alongside metadata so callers have full access to the diagnostic data without fragile type-specific XML parsing.

Closes #9823

### DIFF
--- a/dbatools.psd1
+++ b/dbatools.psd1
@@ -196,6 +196,7 @@
         'Get-DbaAgHadr',
         'Get-DbaAgListener',
         'Get-DbaAgReplica',
+        'Get-DbaAgRingBuffer',
         'Get-DbaAvailabilityGroup',
         'Get-DbaAvailableCollation',
         'Get-DbaBackupDevice',
--- a/dbatools.psm1
+++ b/dbatools.psm1
@@ -607,6 +607,7 @@ if ($PSVersionTable.PSVersion.Major -lt 5) {
         'Mount-DbaDatabase',
         'Dismount-DbaDatabase',
         'Get-DbaAgReplica',
+        'Get-DbaAgRingBuffer',
         'Get-DbaAgDatabase',
         'Get-DbaAgDatabaseReplicaState',
         'Get-DbaModule',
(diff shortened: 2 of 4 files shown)


ReeceGoding:
Looks good to me. The timestamp stuff looks a bit weird, but it made sense once I remembered [the definition of the timestamp column](https://learn.microsoft.com/en-us/sql/relational-databases/system-dynamic-management-views/sys-dm-os-ring-buffers-transact-sql?view=sql-server-ver17).

I wonder if a `CollectionMinutes` parameter is worth adding for consistency with `Get-DbaCpuRingBuffer`? I'm not decided on that.

andreasjordan:
@claude please have a look at the last comment. I think we should add CollectionMinutes for consistency with Get-DbaCpuRingBuffer.

potatoqualitee:
awesome, thank you both 💯

PR #10289, opened 2026-03-28
Fix DAC connection reuse detection causing excessive DAC errors in Sync-DbaAvailabilityGroup

### LINKED ISSUE #10284: Sync-DbaAvailabilityGroup generates excessive amount of DAC connection errors since v2.7.25 in SQL Error Log

Since dbatools version v2.7.25 (this issue did not occur in v2.7.24, as confirmed by SQL Error Log analysis and multiple tests with different module versions across multiple SQL environments), we have been receiving a large number of alerts from our SQL monitoring platform. The SQL Error Log reports repeated errors of the following type:

_“Could not connect because the maximum number of '1' dedicated administrator connections already exists. Before a new connection can be made, the existing dedicated administrator connection must be dropped, either by logging off or ending the process.”_

We run a PowerShell script every night that keeps the secondary node of an Always On Availability Group synchronized with the primary.
The timestamps in the SQL Error Log correlate exactly with the runtime of this script, which uses Sync-DbaAvailabilityGroup.

In the dbatools release history, we noticed several DAC‑related changes introduced in v2.7.25. 
We experienced a similar (but far less severe) issue in the past, where only about three DAC‑related errors appeared in the log:
https://github.com/dataplat/dbatools/issues/9204

However, in one customer environment, we now see 30–40 DAC errors within an 8‑minute window when Sync-DbaAvailabilityGroup runs.
In addition, since v2.7.25, the runtime of this script increased significantly, from an average of ~2 minutes to ~8 minutes, without any material changes to the SQL environment.

**Errors observed in verbose output**

Verbose output from Sync-DbaAvailabilityGroup shows repeated exceptions such as:
Exception calling "EnumMembers" with "0" argument(s): "Enumerate members failed for DatabaseRole 'db_securityadmin'."
...
Exception calling "EnumMembers" with "0" argument(s): "Enumerate members failed for DatabaseRole 'RSExecRole'."
...
Exception calling "EnumMembers" with "0" argument(s): "Enumerate members failed for DatabaseRole 'db_backupoperator'."

Additionally, we see warnings such as:

WARNING: [Sync-DbaLoginPermission] Issue syncing permissions for login | An existing connection was forcibly closed by the remote host

**Relevant changes introduced in v2.7.25**
Based on the release notes, the following behavior was introduced:

When opening a connection we test if we need a DAC and we test if we already have a DAC. We reuse a DAC when possible. We remember if we opened a DAC to later be able to close it.

Get-DecryptedObject needs two things:

A Dedicated Admin Connection (DAC) to select master.sys.syslnklgns
A WinRM connection to read "HKLM:\SOFTWARE\Microsoft\Microsoft SQL Server$serviceInstanceId\Security"
That's why Get-DecryptedObject needs $Credential which it used all the time, but the parameter was missing. So every command that calls Get-DecryptedObject needs $Credential as well.

**Question**
Is the $Credential parameter now required for decrypting passwords via the DAC connection?
In other words, does omitting $Credential cause the DAC‑related errors and forced connection drops we are seeing?

Or is $Credential still optional, and is the increased DAC usage and the new DAC reuse logic introduced in v2.7.25 causing a separate underlying issue?

### Steps to Reproduce

$primary = server1\instance
secondary  = server2\instance
$MSXjobs = $MSXjobs = (Get-DbaAgentJob -SqlInstance $primary | Where-Object Category -eq "Jobs from MSX").Name

Sync-DbaAvailabilityGroup -Primary $primary -Secondary $secondary -Exclude AgentOperator,CustomErrors -ExcludeJob $MSXjobs -ExcludeLogin sa,$executingUser -Force -Verbose


2.7.27

### Other details or mentions

_No response_

### PowerShell Host Version


PSVersion                      5.1.17763.8510                                                                                                                                                                                                                                                                                                                                                                      
BuildVersion                   10.0.17763.8510                                                                                                                                                                                                                                                                                                                                                                     




Microsoft SQL Server 2019 (RTM-CU32-GDR) (KB5068404) - 15.0.4455.2 (X64)   Oct  7 2025 21:10:15   Copyright (C) 2019 Microsoft Corporation  Enterprise Edition: Core-based Licensing (64-bit) on Windows Server 2019 Datacenter 10.0 <X64> (Build 17763: ) (Hypervisor) 

### .NET Framework Version

.NET Framework 4.8.4795.0

andreasjordan:
Fix two bugs introduced in v2.7.25 that cause 30-40 "max 1 DAC connections" errors in the SQL Error Log when Sync-DbaAvailabilityGroup runs.

**Bug 1: dacConnected check used Server.Name instead of ConnectionContext.ServerInstance**

Server.Name never has the ADMIN: prefix for DAC connections — that lives in ConnectionContext.ServerInstance. The broken check caused Copy-DbaCredential, Copy-DbaDbMail, and Copy-DbaLinkedServer to always think no DAC was open, set $dacOpened = $true, and disconnect the shared DAC in their end blocks.

**Bug 2: Sync-DbaLoginPermission received the DAC server object**

Update-SqlPermission calls SqlConnectionObject.Close() in tight loops (per role per database per login). With a DAC connection, each Close() physically terminates the TCP connection and the next SMO operation immediately tries to reconnect to ADMIN:server. TCP close propagation timing causes SQL Server to see the old connection as still active, generating the errors.

Fixes #10284

Generated with [Claude Code](https://claude.ai/code)) | [View branch](https://github.com/dataplat/dbatools/tree/claude/issue-10284-20260327-2247

### DIFF
--- a/public/Copy-DbaCredential.ps1
+++ b/public/Copy-DbaCredential.ps1
@@ -156,7 +156,7 @@ function Copy-DbaCredential {
             if ($ExcludePassword) { $dacNeeded = $false } else { $dacNeeded = $true }
 
             # Do we have a dedicated admin connection already?
-            $dacConnected = $Source.Type -eq 'Server' -and $Source.InputObject.Name -match '^ADMIN:'
+            $dacConnected = $Source.Type -eq 'Server' -and $Source.InputObject.ConnectionContext.ServerInstance -match '^ADMIN:'
 
             $dacOpened = $false
             if ($dacNeeded) {
--- a/public/Copy-DbaDbMail.ps1
+++ b/public/Copy-DbaDbMail.ps1
@@ -390,7 +390,7 @@ function Copy-DbaDbMail {
             if ($ExcludePassword) { $dacNeeded = $false } else { $dacNeeded = $true }
 
             # Do we have a dedicated admin connection already?
-            $dacConnected = $Source.Type -eq 'Server' -and $Source.InputObject.Name -match '^ADMIN:'
+            $dacConnected = $Source.Type -eq 'Server' -and $Source.InputObject.ConnectionContext.ServerInstance -match '^ADMIN:'
 
             $dacOpened = $false
             if ($dacNeeded) {
--- a/public/Copy-DbaLinkedServer.ps1
+++ b/public/Copy-DbaLinkedServer.ps1
@@ -307,7 +307,7 @@ function Copy-DbaLinkedServer {
             if ($ExcludePassword) { $dacNeeded = $false } else { $dacNeeded = $true }
 
             # Do we have a dedicated admin connection already?
-            $dacConnected = $Source.Type -eq 'Server' -and $Source.InputObject.Name -match '^ADMIN:'
+            $dacConnected = $Source.Type -eq 'Server' -and $Source.InputObject.ConnectionContext.ServerInstance -match '^ADMIN:'
 
             $dacOpened = $false
             if ($dacNeeded) {
--- a/public/Export-DbaCredential.ps1
+++ b/public/Export-DbaCredential.ps1
@@ -115,7 +115,7 @@ function Export-DbaCredential {
                 if ($ExcludePassword) { $dacNeeded = $false } else { $dacNeeded = $true }
 
                 # Do we have a dedicated admin connection already?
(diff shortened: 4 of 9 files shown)


andreasjordan:
I will test this this weekend.

andreasjordan:
See issue on details why I closed this.

PR #10293, opened 2026-03-30
Get-DbaNetworkEncryption - Add command to retrieve TLS certificate from SQL Server network

andreasjordan:
Adds `Get-DbaNetworkEncryption` which retrieves the TLS/SSL certificate presented by a SQL Server instance during the TLS handshake, without requiring Windows host access or WinRM.

**Key features:**
- Connects directly to SQL Server's TCP port via TLS/SSL
- No Windows host access required (no WinRM)
- Handles named instances via SQL Browser (UDP 1434) with proper DNS resolution
- Returns Subject, Issuer, Thumbprint, Expiration, DNS SANs, and more
- Graceful error handling for instances without TLS configured

This complements `Get-DbaNetworkCertificate` (which reads from the Windows registry, requires WinRM) by instead showing what certificate is actually being presented to clients.

### DIFF
--- a/dbatools.psd1
+++ b/dbatools.psd1
@@ -333,6 +333,7 @@
         'Get-DbaMsdtc',
         'Get-DbaNetworkActivity',
         'Get-DbaNetworkCertificate',
+        'Get-DbaNetworkEncryption',
         'Get-DbaNetworkConfiguration',
         'Get-DbaOpenTransaction',
         'Get-DbaOperatingSystem',
--- a/dbatools.psm1
+++ b/dbatools.psm1
@@ -1068,6 +1068,7 @@ if ($PSVersionTable.PSVersion.Major -lt 5) {
         'Test-DbaComputerCertificateExpiration',
         'Test-DbaNetworkCertificate',
         'Get-DbaNetworkCertificate',
+        'Get-DbaNetworkEncryption',
         'Set-DbaNetworkCertificate',
         'Remove-DbaDbLogShipping',
         'Invoke-DbaDbLogShipping',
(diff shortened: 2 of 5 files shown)


andreasjordan:
@claude I have added two files in this branch:
* private/functions/Get-SqlServerTlsCertificate.ps1
* public/Get-DbaNetworkEncryption.ps1

Get-SqlServerTlsCertificate should be called from Get-DbaNetworkEncryption. We want to preserve the content of Get-SqlServerTlsCertificate as much as possible but some modifications are allowed if needed.

Please add the code that is needed to get this working.

andreasjordan:
The command works in my lab as expected.

andreasjordan:
@claude we still want to support PowerShell version 3. Would it be possible to change the code of private/functions/Get-SqlServerTlsCertificate.ps1 to be compatible with PowerShell version 3? Don't change the code yet, only analyze.

andreasjordan:
@claude please implement the suggested changes.

andreasjordan:
I'm very happy with this command and it works in my lab. But I would like to remove the port from the output as we don't get the used port from the private function and the output is all about the certificate and the port is not so importent.

What do you think @potatoqualitee ?

potatoqualitee:
either works, so your preference 😊

potatoqualitee:
and thank you! im looking super forward to this command

andreasjordan:
Ok, I will update the code tomorrow.

andreasjordan:
It works!

image

potatoqualitee:
i have rarely been more excited for a new command 🥳 🥳 🥳 🥳 🥳 🥳 🥳  thank you!!

PR #10295, opened 2026-03-31
Export-DbaCredential - Add IF NOT EXISTS guard to exported SQL scripts

### LINKED ISSUE #8618: If EXIST Logic check for Export-DbaCredential

Hello, can I suggest adding an "IF EXIST" Logic check to the function "Export-DbaCredential"?

This would aid during with migration or DR scripts, so that only missing credentials are created

Yes

### Technical Details

I would suggest adding the following to lines 191 & 194

sql += "IF NOT EXISTS(SELECT [name] FROM master.sys.credentials WHERE [name] = '$quotename') CREATE CREDENTIAL

as a prefix to the current sql += "CREATE CREDENTIAL

andreasjordan:
Wraps each CREATE CREDENTIAL statement with an IF NOT EXISTS check against sys.credentials so that exported scripts are idempotent and only create missing credentials during migrations or DR script replays.

Closes #8618

### DIFF
--- a/public/Export-DbaCredential.ps1
+++ b/public/Export-DbaCredential.ps1
@@ -165,6 +165,7 @@ function Export-DbaCredential {
 
             foreach ($cred in $credentials) {
                 $quotename = $cred.Quotename
+                $credName = $cred.Name.Replace("'", "''")
                 $identity = $cred.Identity.Replace("'", "''")
                 $password = $cred.Password.Replace("'", "''")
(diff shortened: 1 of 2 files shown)


andreasjordan:
@claude please read the other `Export-Dba*` commands and see if this pattern should be applied there as well.

potatoqualitee:
thank you 🥇

PR #10298, opened 2026-03-31
Get-DbaService - Add PowerBI Report Server detection

### LINKED ISSUE #8408: Get-DbaSerivce detect PowerBI service
### Summarize Functionality

For the Get-DbaService command to detect with PowerBI server is installed.  I know this is not a service that is installed with the installed, but neither is SSRS anymore, but I've got a job now where this was installed on a server, and it would be nice if it could be detected for when using the Max Memory stuff and setting service account stuff.

No

### Technical Details

_No response_

andreasjordan:
Detect Power BI Report Server (PBIRS) as a distinct service type ('PowerBI') instead of incorrectly labeling it as 'SSRS'. PBIRS uses the WMI namespace RS_PBIRS under root\Microsoft\SQLServer\ReportServer; this prefix is now used to set the appropriate ServiceType.

Adds 'PowerBI' to the -Type ValidateSet so callers can filter specifically for PBIRS services.

Fixes #8408

### DIFF
--- a/private/functions/Get-DbaReportingService.ps1
+++ b/private/functions/Get-DbaReportingService.ps1
@@ -116,7 +116,8 @@ function Get-DbaReportingService {
                     if ($serviceArray -notcontains $($service.ServiceName)) {
                         if (!$InstanceName -or $service.InstanceName -in $InstanceName) {
                             #Add other properties and methods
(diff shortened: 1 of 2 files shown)


andreasjordan:
I can not test this.

potatoqualitee:
Seems reasonable 👍🏼

PR #10305, opened 2026-04-03
Copy-DbaLogin - Add -ExcludeDatabaseMapping to sync only server permissions

### LINKED ISSUE #8312: Copy-DbaLogin Permissions sync only Server Roles and Server securables
### Summarize Functionality

We have a system where logins gets created regularly. We are using copy-dbalogin to sync logins between nodes in Availability groups. 
With a large number of databases copy-dbalogin performs slower than we would like. 
When using the -ExcludePermissionSync , performance is great but server permissions are not carried over. 

It would be helpful if the function has an option for Server Permissions only and avoids looping through all databases for database permissions. 

No

### Technical Details

_No response_

andreasjordan:
Adds a new `-ExcludeDatabaseMapping` switch to `Copy-DbaLogin` that syncs only server-level roles and securables, skipping the database mapping loop.

This addresses the performance concern for instances with many databases where full permission sync is slow but `-ExcludePermissionSync` discards desired server-level permissions.

Fixes #8312

### DIFF
--- a/private/functions/Update-SqlPermission.ps1
+++ b/private/functions/Update-SqlPermission.ps1
@@ -32,6 +32,7 @@ function Update-SqlPermission {
         [ValidateNotNullOrEmpty()]
         [object]$DestLogin,
         [switch]$ObjectLevel,
+        [switch]$ExcludeDatabaseMapping,
         [switch]$EnableException
     )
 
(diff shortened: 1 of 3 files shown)


potatoqualitee:
nice and straightforward

PR #10307, opened 2026-04-03
Invoke-DbaDbShrink - Add WAIT_AT_LOW_PRIORITY support

### LINKED ISSUE #10193: Add WAIT_AT_LOW_PRIORITY support to Invoke-DbaDbShrink
### Summarize Functionality

`Invoke-DbaDbShrink` calls `DBCC SHRINKFILE`. [As of SQL Server 2022](https://learn.microsoft.com/en-us/sql/t-sql/database-console-commands/dbcc-shrinkfile-transact-sql?view=sql-server-ver17#wait_at_low_priority), `WAIT_AT_LOW_PRIORITY` is supported by `DBCC SHRINKFILE`. We should give `Invoke-DbaDbShrink` this functionality. For the same reason that we provide `StatementTimeout` and the option to `Invoke-DbaDbShrink` in steps, `WAIT_AT_LOW_PRIORITY` is quite handy.

No

### Technical Details

This will be tricky for two reasons:

* We currently use [the SMO shrink method](https://learn.microsoft.com/en-us/dotnet/api/microsoft.sqlserver.management.smo.databasefile.shrink?view=sql-smo-172#microsoft-sqlserver-management-smo-databasefile-shrink(system-int32-microsoft-sqlserver-management-smo-shrinkmethod)), called as `$file.Shrink($shrinkSizeKB.Megabyte, $ShrinkMethod)`. We would need to re-write to use T-SQL.
* Nothing else in dbatools supports `WAIT_AT_LOW_PRIORITY`, so we would have to figure out how calling it from PowerShell should look. It effectively has three parameters: if it should be on or off, the max duration in minutes, and [what to abort](https://learn.microsoft.com/en-us/sql/t-sql/database-console-commands/dbcc-shrinkfile-transact-sql?view=sql-server-ver17#abort_after_wait---self--blockers-).

andreasjordan:
Adds WAIT_AT_LOW_PRIORITY support to Invoke-DbaDbShrink by replacing the SMO `$file.Shrink()` calls with direct DBCC SHRINKFILE T-SQL.

New parameters:
- `-WaitAtLowPriority` (switch)
- `-AbortAfterWait` (`Self` or `Blockers`, default Self) — no None option as DBCC SHRINKFILE does not support it

Requires SQL Server 2022+; a version guard via Stop-Function is included for older instances.

Closes #10193

### DIFF
--- a/public/Invoke-DbaDbShrink.ps1
+++ b/public/Invoke-DbaDbShrink.ps1
@@ -58,6 +58,16 @@ function Invoke-DbaDbShrink {
         Sets the command timeout in minutes for the shrink operation. Defaults to 0 (infinite timeout).
         Large database shrinks can take hours to complete, so the default allows operations to run without timing out.
 
+    .PARAMETER WaitAtLowPriority
+        Instructs DBCC SHRINKFILE to wait at low priority for schema modification locks, minimizing blocking of other sessions.
+        Requires SQL Server 2022 (version 16) or later. When a lock cannot be obtained, SQL Server will abort the operation
+        after approximately one minute and log error 49516. The AbortAfterWait parameter controls what is aborted.
+
+    .PARAMETER AbortAfterWait
+        Specifies what to abort when the WAIT_AT_LOW_PRIORITY wait period expires. Valid values are Self and Blockers.
+        Self aborts the shrink operation itself. Blockers kills the user sessions that block the lock acquisition.
+        Defaults to Self. Only applies when WaitAtLowPriority is specified.
+
     .PARAMETER LogsOnly
         Deprecated. Use FileType instead.
 
@@ -165,6 +175,11 @@ function Invoke-DbaDbShrink {
 
         Shrinks all databases coming from a pre-filtered list via Get-DbaDatabase
 
+    .EXAMPLE
+        PS C:\> Invoke-DbaDbShrink -SqlInstance sql2022 -Database AdventureWorks2014 -WaitAtLowPriority -AbortAfterWait Blockers
+
+        Shrinks AdventureWorks2014 using WAIT_AT_LOW_PRIORITY, killing blocking sessions if the wait expires. Requires SQL Server 2022+.
+
     #>
     [CmdletBinding(SupportsShouldProcess, ConfirmImpact = 'Low')]
     param (
@@ -183,6 +198,9 @@ function Invoke-DbaDbShrink {
         [string]$FileType = 'All',
         [int64]$StepSize,
         [int]$StatementTimeout = 0,
+        [switch]$WaitAtLowPriority,
+        [ValidateSet('Self', 'Blockers')]
+        [string]$AbortAfterWait = 'Self',
         [switch]$ExcludeIndexStats,
         [switch]$ExcludeUpdateUsage,
         [switch]$EnableException,
@@ -201,6 +219,11 @@ function Invoke-DbaDbShrink {
             $stepSizeKB = ([dbasize]($StepSize)).Kilobyte
         }
         $StatementTimeoutSeconds = $StatementTimeout * 60
+
+        $walp = ""
+        if ($WaitAtLowPriority) {
+            $walp = " WITH WAIT_AT_LOW_PRIORITY (ABORT_AFTER_WAIT = $($AbortAfterWait.ToUpper()))"
+        }
 
         $sql = 'SELECT
                   AVG(avg_fragmentation_in_percent) AS [avg_fragmentation_in_percent]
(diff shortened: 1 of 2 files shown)


ReeceGoding:
Haven't tested, but it looks to be pretty much what I expected. My only nitpicks are:

- I don't know if error 49516 is really the error that gets thrown. However, [the SQL documentation claims it](https://learn.microsoft.com/en-us/sql/t-sql/database-console-commands/dbcc-shrinkfile-transact-sql?view=sql-server-ver17#wait_at_low_priority) so I suppose we can blindly trust that. 
- The test doesn't test that WAIT_AT_LOW_PRIORITY is used. Locking a database and checking the relevant DMV for a `LCK_M_SCH_M_LOW_PRIORITY` lock should be pretty easy? My first stupid idea would be to target a database with `sp_addextendedproperty` while in a transaction. The same trick would also work for a filegroup.

andreasjordan:
@claude Test fail with:

```
Failed tests summary:
Path           : Invoke-DbaDbShrink/Verifying Database is shrunk/Shrinks with WaitAtLowPriority on SQL Server 2022+
Name           : It Shrinks with WaitAtLowPriority on SQL Server 2022+
Result         : Failed
Message        : StdOut:  @{ComputerName=appveyor-vm; InstanceName=SQL2022; SqlInstance=appveyor-vm\SQL2022; Database=dbatoolsci_shrinktest; Status=Dropped} | ErrorRecord: Exception calling "Query" with "1" argument(s): "An 
                 exception occurred while executing a Transact-SQL statement or batch.MODIFY FILE failed. Specified size is less than or equal to current size.
                 MODIFY FILE failed. Specified size is less than or equal to current size."
StackTrace     : 
RawErrorRecord : Exception calling "Query" with "1" argument(s): "An exception occurred while executing a Transact-SQL statement or batch.MODIFY FILE failed. Specified size is less than or equal to current size.
                 MODIFY FILE failed. Specified size is less than or equal to current size."
```

Also read the comment from Reece and try to implement his suggestion.

ReeceGoding:
Maybe my idea was bad? The locks from adding an extended property are
```xml
<Database name="model">
  <Locks>
	<Lock request_mode="S" request_status="GRANT" request_count="1" />
  </Locks>
  <Objects>
	<Object name="sysxprops" schema_name="sys">
	  <Locks>
		<Lock resource_type="KEY" index_name="clust" request_mode="X" request_status="GRANT" request_count="1" />
		<Lock resource_type="OBJECT" request_mode="IX" request_status="GRANT" request_count="1" />
	  </Locks>
	</Object>
  </Objects>
</Database>
```

ReeceGoding:
Wow. **All** of my ideas for testing this were bad. Look I what I just got from `sp_WhoIsActive`!

> dd hh:mm:ss.mss     : 00 00:00:03.924
> session_id          : 105
> sql_text            : `<?query --  DBCC SHRINKFILE (N'modeldev' , 51) WITH WAIT_AT_LOW_PRIORITY (ABORT_AFTER_WAIT = SELF);    --?>`
> login_name          : sa
> wait_info           : (3915ms)LCK_M_X
> CPU                 :                       4
> tempdb_allocations  :                   0
> tempdb_current      :                   0
> blocking_session_id : 104
> reads               :                 960
> writes              :                   0
> physical_reads      :                   0
> locks               : `<Database name="model"><Locks><Lock request_mode="S" request_status="GRANT" request_count="1" /></Locks><Objects><Object name="t" schema_name="dbo"><Locks><Lock resource_type="HOBT.BULK_OPERATION" 
>                       request_mode="IX" request_status="GRANT" request_count="1" /><Lock resource_type="OBJECT" request_mode="Sch-S" request_status="GRANT" request_count="2" /><Lock resource_type="OBJECT" request_mode="X" 
>                       request_status="CONVERT" request_count="1" /></Locks></Object></Objects></Database>`

Perhaps a better test is to either:

- Just grab the SQL text for whatever is running and see if `WAIT_AT_LOW_PRIORITY` is in those stings
- Tell it to abort itself, wait a minute while you have a really nasty lock, and then see if the query has errored.

Getting a lock nasty enough for my second idea has proven difficult. I suggest the first idea instead.

At this point, I can't even prove to myself that Microsoft's implementation of `WAIT_AT_LOW_PRIORITY` works.

andreasjordan:
@claude can you help?

ReeceGoding:
I'm very surprised. I've thrown an unplanned few hours at it and I have not been able to produce a demo showing `DBCC SHRINKFILE [...] WITH WAIT_AT_LOW_PRIORITY` actually waiting at low priority. The documentation makes it sound like you need to get it to [move an IAM page](https://learn.microsoft.com/en-us/sql/t-sql/database-console-commands/dbcc-shrinkfile-transact-sql?view=sql-server-ver17#understand-concurrency-issues-with-dbcc-shrinkfile). I'm confident that I could make it happen by throwing the command at a real multi-terabyte database and waiting, but I would be very impressed if anyone can show a reproducible demo.

Regardless, I agree with the approach Claude has now taken. Assuming that the test passes, then I'm happy.

potatoqualitee:
👌🏼  thank you both!

Issue #9010, opened 2023-06-21
Commands have inconsistent behavior with table names with special characters


jasonmcboyd:
### Verified issue does not already exist?

I have searched and found no existing issue

### What error did you receive?

Errors vary, `Get-DbaDbTable` returns null while `Import-DbaCsv`, for example, returns:

> Table or view [this.is.not.a.table] does not exist and AutoCreateTable was not specified

### Steps to Reproduce

Create a table in a database with special characters like periods, Gross.Table.Name, for example. Some dbatool commands that take a `Table` parameter only work if the name is bracketed like [Gross.Table.Name] and other commands only work if the table name is not bracketed.

```powershell
$tableName = 'Gross.Table.Name'
$bracketedTableName = '[Gross.Table.Name]'

# This works.
$table = Get-DbaDbTable -SqlInstance $myInstance -Table $bracketedTableName
# This results in $table being null
$table = Get-DbaDbTable -SqlInstance $myInstance -Table $tableName

# This results in "Table or view [this.is.not.a.table] does not exist and AutoCreateTable was not specified"
Import-DbaCsv -SqlInstance $myInstance-Database $myDatabase -Table $bracketedTableName -Path $myPath
# This works
Import-DbaCsv -SqlInstance $myInstance-Database $myDatabase -Table $tableName -Path $myPath
```

### Please confirm that you are running the most recent version of dbatools

2.0.4

### What PowerShell host was used when producing this error

PowerShell Core (pwsh.exe)

### PowerShell Host Version

Name                           Value
----                           -----
PSVersion                      7.2.8
PSEdition                      Core
GitCommitId                    7.2.8
OS                             Microsoft Windows 10.0.19045
Platform                       Win32NT
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0…}
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
WSManStackVersion              3.0

### SQL Server Edition and Build number

Microsoft SQL Server 2017 (RTM-CU31-GDR) (KB5021126) - 14.0.3460.9 (X64)   Jan 25 2023 08:42:43   Copyright (C) 2017 Microsoft Corporation  Enterprise Edition (64-bit) on Windows Server 2016 Standard 10.0 <X64> (Build 14393: ) (Hypervisor) 

### .NET Framework Version

.NET 6.0.11

andreasjordan:
@claude please have a look at this issue. Only analyze and give suggestion.

[...]

PR #10316, opened 2026-04-05
Remove-DbaDbTableData - Normalize table name via Get-ObjectNameParts

andreasjordan:
Fixes Group 4 of issue #9010: bare dotted table names like Gross.Table.Name were interpreted by SQL Server as 3-part database.schema.table names in the DELETE statement.

Parse -Table via Get-ObjectNameParts and reconstruct a properly-bracketed T-SQL identifier before embedding in the DELETE statement.


potatoqualitee:
very nice 💯

PR #10319, opened 2026-04-05
Add Compare-DbaLogin command

### LINKED ISSUE #9179: Enhancement to Copy-DbaLogin
### Summarize Functionality

At present, the process does not take into account any destination server permissions when recreating the sql login using the -Force parameter.  This could leave to a situation whereby permissions are removed on accounts supporting processes that are used to support multiple concurrent databases / processes. 

Yes

### Technical Details

Sort of maps to the functionality provided by Get-DbaLogin?

andreasjordan:
Implements a new `Compare-DbaLogin` command that compares SQL Server logins between a source and one or more destination instances. Returns each login's status as SourceOnly, DestinationOnly, or Both - enabling users to identify logins that would be lost when running Copy-DbaLogin -Force.

Closes #9179

### DIFF
--- a/dbatools.psd1
+++ b/dbatools.psd1
@@ -84,6 +84,7 @@
         'Compare-DbaAgReplicaAgentJob',
         'Compare-DbaAgReplicaCredential',
         'Compare-DbaAgReplicaLogin',
+        'Compare-DbaLogin',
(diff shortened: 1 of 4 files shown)


potatoqualitee:
Thank you 💯

PR #10323, opened 2026-04-05
Get-DbaWaitStatistic - Fix bug from recent refactoring

andreasjordan:
This bug was introduced in the latest refactoring. The filtering was wrong. As we now need to filter regardless of `-IncludeIgnorable`, I optimized the code as well. Changed the test to let this work on more versions.


potatoqualitee:
Claude has been 💩 this past week

PR #10328, opened 2026-04-06
Test-DbaBuild: Fix bug introduced in last change

andreasjordan:
As we moved the addition of ReleaseDateto Get-DbaBuild, we don't need that code here as the object always has this property already.

Before this fix:
image

After this fix:
image


potatoqualitee:
thank you

PR #10330, opened 2026-04-06
Fix timing issues in tests for Get-DbaExternalProcess and Stop-DbaExternalProcess

andreasjordan:
Sometimes it takes longer to finish the process, so we wait in a loop to avoid this error:

image


andreasjordan:
Does not work as expected - needs more analysis...

PR #10331, opened 2026-04-06
Fix test for Test-DbaSpn by suppressing the warning on AppVeyor

andreasjordan:
Because AppVeyor is not domain-joined, we get this warning:
image

To have a clean test output to be able to see real problem, we nw suppress this warning.


potatoqualitee:
thank you!

PR #10341, opened 2026-04-10
Install-DbaWhoIsActive - Select version-appropriate SQL file based on server version

### LINKED ISSUE #10339: Install-DbaWhoIsActive - sp_whoisactive versioning change
### Summarize Functionality

Hi DBATools team,

I just pushed a new [sp_whoisactive](https://github.com/amachanic/sp_whoisactive) versioning change, that I hope you'll enhance DBATools to support.

Basically, we've decided to no longer try to support every single SQL Server version going back 20 years, in a single script. So we've broken the current version into three scripts, stored in different subfolders in the repo: one targeting SQL Server 2005 and 2008; a second for 2012-2019; and a third for SQL Server 2022 and beyond.

Going forward the root folder version of the script will be the current one, that will support the current SQL Server version and one or two back, and we'll create new prior-version folders every couple of releases so that we no longer have to carry the weight of those older versions forever.

What this means for `Install-DbaWhoIsActive` is that ideally when installing, you can look at the target SQL Server's version and pick the right script for compatibility.

I hope that makes sense! Let me know if I can do anything better on our end to make this easy to support.

Appreciate your enhancing this, and I hope I haven't broken anything in the meantime... 🤞 

Yes

### Technical Details

_No response_

andreasjordan:
Implements version-based SQL file selection for sp_WhoIsActive to support the new multi-script release structure.

The SQL file selection is moved from the begin block to the process block so each target server gets the correct script version based on its VersionMajor property. Subfolder matching parses year ranges from directory names (e.g. "2005-2008") and checks whether the server's release year falls within that range.

Closes #10339

### DIFF
--- a/public/Install-DbaWhoIsActive.ps1
+++ b/public/Install-DbaWhoIsActive.ps1
@@ -146,58 +146,26 @@ function Install-DbaWhoIsActive {
             }
 
             # Select the appropriate SQL file based on the target server's version.
-            # sp_WhoIsActive ships version-specific SQL files in subfolders named by the maximum supported year (e.g. 2008, 2019).
-            # The root sp_WhoIsActive.sql targets the latest SQL Server version (2022+).
-            $allSqlFiles = @(Get-ChildItem -Path $localCachedCopy -Filter 'sp_WhoIsActive.sql' -Recurse)
-            if ($allSqlFiles.Count -eq 0) {
-                Write-Message -Level Verbose -Message "New filename sp_WhoIsActive.sql not found, using old filename who_is_active.sql."
-                $allSqlFiles = @(Get-ChildItem -Path $localCachedCopy -Filter 'who_is_active.sql' -Recurse)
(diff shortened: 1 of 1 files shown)


andreasjordan:
@claude the folder structure is different than expected.

This is the new structure:


Folder "2008" with the files for "SQL Server 2005 - 2008 compatible version":
* README.md
* sp_WhoIsActive.sql

Folder "2019" with the files for "SQL Server 2012 - 2019 compatible version":
* README.md
* sp_WhoIsActive.sql

Base folder with the files for current versions (2022+):
* .gitattributes
* LICENSE
* README.md
* sp_WhoIsActive.sql

andreasjordan:
@claude I think this is to much code and to much flexibility. Let's only test:
* if version 2008 or older - use folder 2008
* if version 2019 or older - use folder 2019
* else use base folder

Just in case no file is found or the folder is not present then try to find the old "who_is_active.sql" file.

potatoqualitee:
beautiful 💅🏼

PR #10346, opened 2026-04-15
Various bug fixes and hooks

potatoqualitee:
I used a dedicated session per commit and asked Opus to fix any potential bugs, since you were concerned @andreasjordan. Unfortunately/fortunately it also cleaned up the coding style which is great but makes it harder to review.


andreasjordan:
There are way too many changes to have a closer look. But in general the changes look ok, so maybe release them with a notice.

PR #10353, opened 2026-04-19
Install-DbaSqlWatch: Add missing EnableException

andreasjordan:
Add missing EnableException to fail the command if Publish-DbaDacPackage failes.


potatoqualitee:
Very nice thank you 💯

PR #10356, opened 2026-04-20
Some test improvements and fixes

andreasjordan:
* Not using `-Confirm:$false` as this is default (where possible).
* Suppressing some warnings that are not relevant.
* Enabling remote DAC to help test for New-DbaFirewallRule.
* Fix test for Get-DbaExternalProcess and Stop-DbaExternalProcess so that the temp file can be removed


potatoqualitee:
hayle yeah thank you 🔥

PR #10371, opened 2026-05-18
Copy-DbaAgentJob - Add -NewName parameter to allow copying jobs to sa…

### LINKED ISSUE #10370: copy-dbaagentjob do not allow copy to same server.
### Summarize Functionality

I am trying to setup a test with Ola Hallengrens feature "backup in parallell" which demands me to have identical agent job defintions. There isn't an option in "copy-dbaagentjob" to copy to the same server using "-newname" which i.e. is available in copy-dbadatase.  (Naturally when copying with the same name you get an error. )

Is this by design? Should I bother doing an update to copy-dbagentjob to add this 

No

### Technical Details

Add a -newname parameter as a switch. I haven't dug into the codebase to see if its feasible yet though.

frodr1k:
…me server

Adds a -NewName parameter that allows copying a job under a different name. When source and destination are the same server, -NewName is required. Cannot be used when copying multiple jobs simultaneously.

(do Copy-DbaAgentJob)

Summary
Added -NewName parameter to Copy-DbaAgentJob to allow copying a job under a different name, including to the same server.

Why
It was not possible to copy a job to the same server instance since all jobs on a server must have unique names. Users needed a way to create job copies/variants — similar to how -NewName already works in Copy-DbaDatabase. For example Ole Hallengrens Backup-command allows a parameter named "DatabasesInParallel" which needs the backupjobs to be identical. Hence this avoids typos. 

Changes
Added -NewName parameter that sets the destination job name
Copying to same server now requires -NewName (fails with clear error otherwise)
-NewName cannot be used when copying multiple jobs simultaneously
Updated parameter validation test and added 3 integration tests
Testing
Verified manually against SQL Server instance. Integration tests added for same-server copy, cross-server rename, and error case without -NewName.

Closes #10370

### DIFF
--- a/public/Copy-DbaAgentJob.ps1
+++ b/public/Copy-DbaAgentJob.ps1
@@ -57,6 +57,11 @@ function Copy-DbaAgentJob {
         Overwrites existing jobs on the destination server and automatically sets missing job owners to the 'sa' login.
         Use this when you need to replace existing jobs or when source job owners don't exist on the destination server during migrations.
 
+    .PARAMETER NewName
+        The new name for the job on the destination server.
+        Required when source and destination are the same server instance. Use this to create a copy of a job under a different name on the same or a different server.
+        Cannot be used when copying multiple jobs simultaneously.
+
     .PARAMETER UseLastModified
         When enabled, compares the last modification date (date_modified) from msdb.dbo.sysjobs between source and destination instances.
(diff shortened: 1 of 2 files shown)


potatoqualitee:
thank you very much! This is a nice addition. I will merge it as soon as all the tests pass 🚀

frodr1k:
thanks!

PR #10397, opened 2026-06-17
2025 CU6 is out (do Build)

niphlod:
2025 CU6 is out


bin/dbatools-buildref-index.json
```diff
@@ -5440,6 +5440,12 @@
             "Version": "17.0.4045",
             "KBList": "5084896",
             "ReleaseDate": "2026-05-20"
+        },
+        {
+            "CU": "CU5",
```
andreasjordan: It is CU6.

andreasjordan (changes requested):

bin/dbatools-buildref-index.json
niphlod: lol, true. Thanks for spotting

andreasjordan: Will merge this tomorrow.

PR #10309, opened 2026-04-03
Add Import-DbaParquet command

jovanpop-msft:
Introduce the Import-DbaParquet command along with enhancements for schema management, error handling, and stability. Implement exact size logic for varchar and varbinary types, and add a parameter to store strings as UTF-8. Fix issues related to binary type mapping and automatic table creation.

PR #10415, opened 2026-07-12
Save-DbaCommunitySoftware - Retry transient download failures (do Save-DbaCommunitySoftware, Install-DbaDarlingData)

potatoqualitee:
## Summary

Cross-platform CI intermittently fails in the `installs darling data` step with:

```
WARNING: [Install-DbaDarlingData] Failed to update local cached copy | Cannot find path '/tmp/dbatools_software_download_<random>.zip' because it does not exist.
```

The download step in `Save-DbaCommunitySoftware` completed without a terminating error yet produced no file, so the failure only surfaced later as a confusing `Expand-Archive` path error. This flake blocked #10309 (latest example: https://github.com/jovanpop-msft/dbatools/actions/runs/29197930030/job/86664130148 — the same commit family passed the run before and after, and this workflow is 10-for-10 green on `development`).

## Changes

- Retry the download up to 3 times with a short backoff (transient GitHub rate limiting / connection resets on shared runners)
- Remove any partial file before each request so leftovers from a failed request cannot masquerade as a completed download
- Treat a request that completes without creating the file as a failed attempt (retried, then a clear "after 3 attempts" error instead of the downstream Expand-Archive error)
- Add Pester unit coverage with a mocked `Invoke-TlsWebRequest`: transient failure then success, partial-file-then-silent-fallback, silent no-file retry, and giving up after all attempts

## Verification

- New tests pass locally (5/5), including the partial-file masking case
- Real (unmocked) download + `Install-DbaDarlingData` verified against a SQL Server 2022 instance: `Database = master`, 75 cached .sql files


potatoqualitee:
tearing it UP!

PR #10435, opened 2026-07-14
Set-DbaNetworkCertificate - Honor Force for unsuitable certificates

### LINKED ISSUE #10412: 2.8.2 when applying Set-DbaNetworkCertificate with force is ignored and failed checks are stopping it.

Yes, nothing obvious came up.

Upgraded to 2.8.2, found that when installed and setting up my servers our older certificate (not yet expired) is failing on new installs. It was tracked down to an SHA1 vs SHA256 issue within the Set-DbaNetworkCertificate and the Test-DbaNetworkCertificate. We thought the force command would push it through any ways, but no we had to roll back to 2.7.27 to continue installs.

### Steps to Reproduce

Update to 2.8.2
Set-DbaNetworkCertificate with a certificate running SHA1 and use force fails.
With the error of Failed checks: DnsNamesInvalid, SignatureAlgorithmInvalid.
Our certificate won't match as it is a wild card certificate.

We did update to the newest and now we have to roll back.

### Other details or mentions

_No response_

### PowerShell Host Version

PSVersion                      5.1.20348.4294
but also occurs on older powershell 5 versions on Windows Server 2019/2022

SQL Server 2025 and SQL Server 2022 both being actively installed with the newest CU6/CU 25

### .NET Framework Version

PSChildName Version
----------- -------
Client      4.8.09037
Full        4.8.09037
Client      4.0.0.0

potatoqualitee:
## Summary

- adds an explicit `-Force` switch to `Set-DbaNetworkCertificate`
- keeps detailed certificate validation and warns with the failed checks when forcing
- still rejects a thumbprint that is not present in `LocalMachine\My`
- adds a focused regression test without changing existing test behavior

## Root cause

The 2.8 command refactor added strict suitability validation for an explicitly selected certificate. In 2.7.27 the command configured an installed certificate directly, so existing automation using legacy or wildcard certificates could no longer proceed. There was no opt-out for operators who intentionally accept those suitability failures.

## Validation

- Pester unit tests: 3 passed, 0 failed
- PowerShell parser: clean for command and test files
- `git diff --check`: clean
- independent `claude -p --model opus --effort high` review: `VERDICT: CLEAN`

Fixes #10412

### DIFF
--- a/public/Set-DbaNetworkCertificate.ps1
+++ b/public/Set-DbaNetworkCertificate.ps1
@@ -50,6 +50,10 @@ function Set-DbaNetworkCertificate {
         Unsets the currently configured network certificate for the SQL Server instance.
         This will remove the certificate configuration, and SQL Server will not use any certificate for SSL connections.
 
+    .PARAMETER Force
+        Configures the specified certificate even when suitability checks fail.
+        The certificate must still exist in the LocalMachine certificate store. Failed checks are reported as a warning.
+
     .PARAMETER RestartService
         Forces an automatic restart of the SQL Server service after setting the network certificate.
         Certificate changes require a service restart to take effect - without this switch you'll need to manually restart SQL Server.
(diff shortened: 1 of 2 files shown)


potatoqualitee:
looks good 👍🏼  might change behavior in 3.0 to be more accommodating

PR #10437, opened 2026-07-14
New-DbaDatabase - Add containment type option

### LINKED ISSUE #9218: New-DbaDatabase - Containment
### Summarize Functionality

New-DbaDatabase supports specifying the recovery model when creating the database, it would be nice if it also supported setting the containment type from the beginning.

Today i have to first create the database, and then alter it afterwards using Invoke-Dbaquery

Yes

### Technical Details

_No response_

potatoqualitee:
## Summary

- add `-ContainmentType` with `None` and `Partial` values
- set the SMO database containment type before creation
- reject containment requests on SQL Server versions older than 2012
- document the parameter and add focused unit coverage

## Why

`New-DbaDatabase` could set recovery model during creation but required a separate query to configure database containment.

## Validation

- Pester 5.3.1: `tests/New-DbaDatabase.Tests.ps1` excluding `IntegrationTests` — 4 passed, 0 failed
- PowerShell parser: command and test files clean
- `git diff --check`: clean
- independent `claude -p --model opus --effort high` review: clean

Fixes #9218


potatoqualitee:
lookin good 🤩

PR #10440, opened 2026-07-14
Invoke-DbaDbShrink - Add logical file selection

### LINKED ISSUE #10364: Add Feature to existing command "Invoke-DbaDbShrink"

Can you add some more options to this Command:
Where we can select the File if we have multiple files.

So that it can shrink one file to free up space, so that we can use the step feature as well.

Yes

### Technical Details

Add parameter "-Filename" to : function Invoke-DbaDbShrink
Example of the -Filename to target just the one file.
PS C:\> Invoke-DbaDbShrink -SqlInstance sql2014 -Database AdventureWorks2014 -FileName 'AdvWrks02' -PercentFreeSpace 10 -FileType Data -StepSize 25MB

potatoqualitee:
## Summary

- add `-FileName` for selecting one or more logical database files
- apply the filter after `-FileType` while leaving step-size shrinking unchanged
- warn when no requested logical name matches the selected file type
- add focused selection and no-match tests

## Why

Databases with multiple files could only shrink every file of a selected type. Operators could not target a single file while using chunked `-StepSize` shrinking.

## Validation

- Pester 5.3.1: `tests/Invoke-DbaDbShrink.Tests.ps1` excluding `IntegrationTests` — 5 passed, 0 failed
- PowerShell parser: command and test files clean
- `git diff --check`: clean
- independent `claude -p --model opus --effort high` review: clean

Fixes #10364

### DIFF
--- a/public/Invoke-DbaDbShrink.ps1
+++ b/public/Invoke-DbaDbShrink.ps1
@@ -75,6 +75,10 @@ function Invoke-DbaDbShrink {
         Determines which database files to target for shrinking: All (data and log files), Data (only data files), or Log (only log files). Defaults to All.
         Use Data when you only need to reclaim space from data files after large deletions. Use Log to specifically target transaction log files after maintenance operations.
 
+    .PARAMETER FileName
+        Specifies one or more logical database file names to shrink. The filter is applied after FileType, so the requested files must also match the selected file type.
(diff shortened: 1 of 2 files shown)


potatoqualitee:
very nice

PR #10447, opened 2026-07-14
Test-DbaBackupStorageCompatibility - Add S3 multipart risk assessment

### LINKED ISSUE #10197: Command to test an environment for compatibility with S3 backups

When backing up to S3 there are certain risks involved. Because of a combination of limitations and behaviors of both SQL Server and the S3 multi-part API, you can easily get yourself into a situation where backups start to fail. 

The S3 multipart API is essentially stateless, you can just start uploading parts to it. It has no idea how many parts it will receive, so it cannot track "progress". In the context of SQL Server this means a backup could breach S3 limits (10,000 parts per backup object/URL) after sending 100's of GB of data, and unexpectedly fail. This then requires the DBA to use the S3 API, or configure policies, to clean up orphaned object parts from a failed upload. Otherwise those failed object parts would just stay there consuming space.

This command can be used to see how close your existing backup configuration would be to breaching those limitations, and in cases where you are close (or exceeding those limits) this command would give recommendations on how to get back under those limits while also pointing out how those changes may impact your host SQL Server instance and/or the network and target storage. The command will also have a mode that can be used in automation/backup scripts to try to fail early when a database gets close to a threshold.

No

### Technical Details

The general flow of this command is that it would:
* Read backup history for the DBs in question
* Take in the desired MaxTransferSize as a parameter
* Determine if, based on current backup file/URL counts from existing backups, backing up to S3 might breach the 10,000 part limit
* It will then try to calculate a potential solution by adjusting MTS and file count. It prioritizes lower-impact changes before suggesting higher-impact changes

When in monitor mode it will output only databases that need attention
When in Exception mode it will throw an error when a database is found that may breach limits. This can be used in automation workflows.

I have a full working concept I have been working on after doing research on this topic.

potatoqualitee:
## Summary

- adds the provider-neutral `Test-DbaBackupStorageCompatibility` command with initial S3 support
- estimates multipart parts from the latest full backup's effective size and stripe count
- reports the 64-URL and 10,000-parts-per-URL limits with structured statuses and recommendations
- supports monitor mode for automation
- keeps deterministic multipart arithmetic in a pure private helper with direct, mock-free boundary tests
- validates the public command against real SQL Server 2022 backup history stored in TLS-enabled MinIO

## Real S3 integration coverage

The S3 GitHub Action provisions SQL Server 2022 and MinIO. The tests create actual databases and send real backups through SQL Server's S3-compatible URL support.

- verifies a normal MinIO backup is read from real `msdb` history and reported as compatible
- verifies the complete documented output-property contract on that live result
- creates more than 500 MiB of incompressible SQL data and writes an uncompressed backup to MinIO
- runs `-Monitor` at the minimum supported 5 MiB transfer size and 1% threshold
- verifies the real backup returns `Warning` and emits the aggregate monitor warning

No `Connect-DbaInstance`, backup-history, or `Stop-Function` mocks remain in this command's behavioral tests.

## Design notes

The estimate prefers compressed backup size when available and otherwise uses total size. It assumes even striping across backup files and does not claim to validate production credentials or networking.

Exactly 10,000 parts is compatible; only values above the documented maximum are incompatible.

## Validation

- Pester 6 focused tests: 3 passed, 0 failed
- PowerShell parser: 4 changed files passed
- PSScriptAnalyzer severity Error: 0 findings
- S3 workflow YAML parsed successfully
- `git diff --check`: passed
- Claude Opus high-effort review: three-review cap completed; output-contract finding addressed on the live MinIO result
- current head `87ade9a23`: all ordinary checks, both real S3/MinIO workflows, and all 10 Azure runner lanes passed

Closes #10197

### DIFF
--- a/.github/scripts/gh-s3actions.ps1
+++ b/.github/scripts/gh-s3actions.ps1
@@ -100,6 +100,43 @@ Describe "S3 Backup Integration Tests" -Tag "IntegrationTests", "S3" {
             $result.BackupPath | Should -BeLike "s3://*"
         }
 
+        It "Should assess the real S3 backup history" {
+            $backupHistory = Get-DbaDbBackupHistory -SqlInstance localhost -SqlCredential $cred -Database $script:TestDbName -LastFull
+            $backupHistory.Path | Should -BeLike "$($script:S3BaseUrl)/*"
+
+            $result = Test-DbaBackupStorageCompatibility -SqlInstance localhost -SqlCredential $cred -Database $script:TestDbName
+
+            $result.Database | Should -Be $script:TestDbName
+            $result.Type | Should -Be "S3"
+            $result.BackupFileCount | Should -Be 1
+            $result.EffectiveBackupSizeBytes | Should -BeGreaterThan 0
+            $result.EstimatedPartsPerFile | Should -BeGreaterThan 0
+            $result.IsCompatible | Should -BeTrue
+            $result.Status | Should -Be "Compatible"
+
+            $expectedProperties = @(
+                "ComputerName",
+                "InstanceName",
+                "SqlInstance",
+                "Database",
+                "Type",
+                "LastFullBackup",
+                "BackupSizeBytes",
+                "CompressedBackupSizeBytes",
+                "EffectiveBackupSizeBytes",
+                "BackupFileCount",
+                "MaxTransferSize",
+                "EstimatedPartsPerFile",
+                "PercentOfLimit",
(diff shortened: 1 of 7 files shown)


potatoqualitee:
@m82labs pls try out the command and let us know if this is what you had in mind

m82labs:
Thanks @potatoqualitee! I will take a look.

potatoqualitee:
we can fix later, merging!

PR #10448, opened 2026-07-14
Set-DbaTempDbConfig - Allow forced data file reduction

### LINKED ISSUE #7758: Set-DbaTempDbConfig - Allow Reduction of the Number of TempDB Files
### Summarize Functionality

Currently, if the existing number of TempDB files is greater than the calculated configuration, Set-DbaTempDbConfig does not make any changes and, instead, returns a warning.  The description for the command states, "No functions that shrink or delete data files are performed. If you wish to do this, you will need to resize tempdb so that it is "smaller" than what the function will size it to before running the function."

Yes

### Technical Details

It would be useful for this command to first check whether TempDB can be shrunk and the extra files removed and, if possible, do so.  A warning should be thrown only if the current usage is greater than the current size of what would be the remaining files.

Example of improved behavior:
If there are currently 4 TempDB files (each 8 GB in size, on a server with 2 logical processors), the command would check whether the current TempDB usage is less than 16 GB.  If so, it would shrink and remove the 2 extra files.

potatoqualitee:
## Summary

- adds an opt-in `-Force` path for reducing the tempdb data-file count
- selects the highest-numbered secondary files first and never removes file id 1
- refuses reduction when current used space exceeds the requested aggregate target
- grows retained files before `EMPTYFILE`, then lets SQL Server's `REMOVE FILE` operation enforce whether each file is removable
- avoids logical and physical filename collisions when adding replacement tempdb files

## Safety

The default behavior is unchanged: without `-Force`, a current file count above the target is still refused. Forced removal explicitly enters `tempdb`, escapes SQL literals and identifiers, and propagates SQL Server's error if `REMOVE FILE` cannot safely remove a file.

The earlier `FILEPROPERTY(..., 'SpaceUsed') = 0` precondition was removed after the real runner exposed it as incorrect. A real SQL Server 2025 reproduction showed `DBCC SHRINKFILE(..., EMPTYFILE)` reporting zero data pages while `FILEPROPERTY` retained eight metadata pages; `ALTER DATABASE ... REMOVE FILE` then succeeded. SQL Server's removal operation is therefore the authoritative safety check.

## Real integration coverage

- expands tempdb by two real data files on `$TestConfig.InstanceSingle`
- renames a real removal target with `'`, `]`, and `%` to exercise SQL escaping through execution
- allocates about 120 MB in a real tempdb table and proves both removal targets contain pages
- verifies the live no-`Force` and insufficient-capacity refusals
- executes the real `EMPTYFILE`/`REMOVE FILE` path, verifies deterministic file removal and row preservation
- restores the original file count, exact size, and growth metadata and fails cleanup if state differs

## Validation

- disposable real SQL Server 2025 reproduction: `EMPTYFILE` reported 0 data pages, `FILEPROPERTY` reported 8 metadata pages, and `REMOVE FILE` succeeded
- Pester 6 parameter contract: 1 passed, 0 failed
- PSScriptAnalyzer 1.18.2 severity Error: 0 findings
- PowerShell parser: 0 errors
- `git diff --check`: clean
- GitHub Actions: all ordinary checks and all 10 real Azure runner lanes passed on the current head

Claude Opus high-effort reviewed this command in three rounds. The final runner-specific correction above was derived from the real SQL Server reproduction after that three-review cap.

Fixes #7758

### DIFF
--- a/public/Set-DbaTempDbConfig.ps1
+++ b/public/Set-DbaTempDbConfig.ps1
@@ -6,9 +6,9 @@ function Set-DbaTempDbConfig {
     .DESCRIPTION
         Configures tempdb database files to follow Microsoft's recommended best practices for performance optimization. This function calculates the optimal number of data files based on logical CPU cores (capped at 8) and distributes the specified total data file size evenly across those files. You must specify the target SQL Server instance and total data file size as mandatory parameters.
 
-        The function automatically determines the appropriate number of data files based on your server's logical cores, but you can override this behavior. It validates the current tempdb configuration to ensure it won't conflict with your desired settings - existing files must be smaller than the calculated target size and you cannot have more existing files than the target configuration.
+        The function automatically determines the appropriate number of data files based on your server's logical cores, but you can override this behavior. It validates the current tempdb configuration to ensure it won't conflict with your desired settings. Existing files must be smaller than the calculated target size, and by default you cannot have more existing files than the target configuration.
 
-        Additional parameters let you customize file paths, log file size, and growth settings. The function generates ALTER DATABASE statements but does not shrink or delete existing files. If your current tempdb is larger than your target configuration, you'll need to shrink it manually before running this function. A SQL Server restart is required for tempdb changes to take effect.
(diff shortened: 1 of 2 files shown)


potatoqualitee:
@claude is it possible to add a real test instead of a mock?

PR #10463, opened 2026-07-26
Install-DbaFirstResponderKit: Added ability to create certificates to let public execute FRK procedures

### LINKED ISSUE #10368: Install-DbaFirstResponderKit: add recommended certificates for non-DBAs
### Summarize Functionality

As recorded in the "How to Grant Permissions to Non-DBAs" section [here](https://www.brentozar.com/askbrent/) and repeated in the free training class, the First Responder Kit and `sp_BlitzFirst` in particular are designed for first responders. Specifically, that link I have given shows that `sp_BlitzFirst` allows for non-DBA usage and gives the following code which allows non-DBAs to run it. We should provide something like this as part of `Install-DbaFirstResponderKit`.
```sql
USE master;
GO
CREATE CERTIFICATE sp_BlitzFirst_cert
ENCRYPTION BY PASSWORD = '5OClockSomewhere'
WITH SUBJECT = 'Certificate for sp_BlitzFirst',
START_DATE = '20130711', EXPIRY_DATE = '21000101';
GO
CREATE LOGIN sp_BlitzFirst_login FROM CERTIFICATE sp_BlitzFirst_cert;
GO
CREATE USER sp_BlitzFirst_login FROM CERTIFICATE sp_BlitzFirst_cert;
GO
GRANT EXECUTE ON dbo.sp_BlitzFirst TO sp_BlitzFirst_login;
GO
GRANT CONTROL SERVER TO sp_BlitzFirst_login;
GO
ADD SIGNATURE TO sp_BlitzFirst BY CERTIFICATE sp_BlitzFirst_cert
WITH PASSWORD = '5OClockSomewhere';
GO
GRANT EXECUTE ON dbo.sp_BlitzFirst TO [public];
GO
```

Yes

### Technical Details

My thinking:
* Copy and pasting the given T-SQL would be a bad idea because of the hard-coded password. This should be easy to solve. We could make the certificate single use with the `ALTER CERTIFICATE sp_BlitzFirst_cert REMOVE PRIVATE KEY` trick [mentioned here](https://www.sommarskog.se/grantperm.html#serverleveluserdbcert). This would easier for us to implement than prompting for passwords.
* I don't know of any good dbatools wrapper for signing stored procedures. I presume that there is some PowerShell way of doing this that is more fun than the T-SQL.
* I imagine that what we want to add to `Install-DbaFirstResponderKit` is a parameter that takes in a collection of stored procedures names and signs all of them with `CONTROL SERVER`. It should be a collection because there are other procedures in the FRK that benefit from this, such as the new `sp_kill`.
* Maybe call the parameter `-LetPublicExecute`?
* If we can use the same certificate for every procedure, then that would be better for everyone. Again, easier for us to implement and easier for users to manage.

Some of my ideas here may be bad! It has been too many years since I've had to seriously think about certificates. Claude will surely outsmart me here.

ReeceGoding:
Be warned, this may be terrible! Specifically:

- It has been too many years since I've had to seriously think about certificates. Claude will surely outsmart me here.
- I am unsure what problem the `CREATE USER` part of [the original code](https://www.brentozar.com/askbrent/) was trying to solve.
-  I had to do a lot of this in T-SQL. I know of no PowerShell interface for certificates.
- The error handling probably isn't good enough, but I don't know certificates or our standards well enough to make it much better.

## Type of Change

 - [X] New feature (non-breaking change, adds functionality, fixes #10368)
 - [X] Pester test is included
 - [X] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
Added ability to create certificates to let public execute FRK procedures. Closes #10368

Also added `sp_kill` to existing ValidateSet.

### Approach
It started as an attempt to copy the "How to Grant Permissions to Non-DBAs" section [here](https://www.brentozar.com/askbrent/), but that set of steps assumes that you are in master. To make other databases also work, I had to expand that with the tricks [mentioned here](https://www.sommarskog.se/grantperm.html#serverleveluserdbcert).

### Commands to test
`Install-DbaFirstResponderKit` and specifically, the new `-LetPublicExecute` parameter.

### Screenshots
image

### Learning

1. While running the integration tests locally, I found that `Invoke-ManualPester -Path ./tests/Install-DbaFirstResponderKit.Tests.ps1 -ScriptAnalyzer  -TestIntegration` fails if you don't have a `C:\Temp` folder. If you know what you're doing, then you can add a new `$config['temp'] = "/path/i/actually/have"` line to `dbatools/tests/constants.local.ps1`, but there is nothing anywhere to hint that you need to do that. Everything else I've ever wanted to change in `dbatools/tests/constants.local.ps1` already had a placeholder.
2. Trying to test this with `EXECUTE AS USER = 'guest'` was a very bad idea. It seems unable to take the permissions from the certificate.
3. `Get-DbaDbUser` does not see certificate-mapped users?

### DIFF
--- a/public/Install-DbaFirstResponderKit.ps1
+++ b/public/Install-DbaFirstResponderKit.ps1
@@ -40,6 +40,13 @@ function Install-DbaFirstResponderKit {
         Specifies specific script files to install instead of the entire First Responder Kit. Accepts multiple script names and wildcards.
         Use this to install only the procedures you need (like sp_Blitz.sql, sp_BlitzCache.sql) or to run official install scripts (Install-All-Scripts.sql, Install-Azure.sql). Also supports Uninstall.sql to remove the toolkit.
 
+    .PARAMETER LetPublicExecute
+        Signs the specified stored procedures with a certificate called "FirstResponderKitCertificate" that lets members of the public server role (that is, everyone) execute those procedures as if they have the CONTROL SERVER permission.
+        This follows the pattern set in the "How to Grant Permissions to Non-DBAs" example at https://www.brentozar.com/askbrent/
+        If you are installing to a non-master database, then the public server role will be given GRANT CONNECT in that database.
+        To prevent abuse, we remove the private key from the certificate after use.
+        Drops any user or login called "FirstResponderKitLogin" and drops any certificate called "FirstResponderKitCertificate" before running.
+
     .PARAMETER Force
         Forces a fresh download of the First Responder Kit from GitHub even if a cached version already exists locally.
         Use this when you want to ensure you have the absolute latest version or when the cached version may be corrupted.
@@ -131,6 +138,10 @@ function Install-DbaFirstResponderKit {
         PS C:\> Install-DbaFirstResponderKit -SqlInstance sql2016 -OnlyScript Uninstall.sql
 
         Uninstalls the First Responder Kit by running the official uninstall script.
+    .EXAMPLE
+        PS C:\> Install-DbaFirstResponderKit -SqlInstance sql2016 -LetPublicExecute sp_BlitzFirst, sp_BlitzIndex
+
+        Installs all scripts, but lets the everyone execute sp_BlitzFirst and sp_BlitzIndex even if they do not have high permissions.
     #>
     [CmdletBinding(SupportsShouldProcess, ConfirmImpact = "Medium")]
     param (
@@ -144,9 +155,10 @@ function Install-DbaFirstResponderKit {
         [ValidateSet('Install-All-Scripts.sql', 'Install-Azure.sql',
             'sp_Blitz.sql', 'sp_BlitzFirst.sql', 'sp_BlitzIndex.sql', 'sp_BlitzCache.sql', 'sp_BlitzWho.sql',
             'sp_BlitzAnalysis.sql', 'sp_BlitzBackups.sql', 'sp_BlitzLock.sql',
-            'sp_DatabaseRestore.sql', 'sp_ineachdb.sql',
+            'sp_DatabaseRestore.sql', 'sp_ineachdb.sql', 'sp_kill.sql',
             'SqlServerVersions.sql', 'Uninstall.sql')]
         [string[]]$OnlyScript,
+        [string[]]$LetPublicExecute,
         [switch]$Force,
         [switch]$EnableException
     )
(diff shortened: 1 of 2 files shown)


andreasjordan:
Thanks for this one! I had a look at the CI failures. Both are fixable, and the bigger one is a PowerShell 5.1 vs 7 difference, so it is nothing you would have seen on your own machine.

## 1. `test (MULTI)` - all nine failures come from a single line

In the three `BeforeAll` blocks:

```powershell
$password = ConvertTo-SecureString "<enterStrongPasswordHere>" -AsPlainText
$public = [PsCredential]::New("AsGoodAsPublic", $password)
```

Our CI runs Windows PowerShell 5.1, where `-AsPlainText` without `-Force` is an error:

> The system cannot protect plain text input. To suppress this warning and convert the plain text to a SecureString, reissue the command specifying the Force parameter.

It is a non-terminating error, so `$password` stays `$null` and the next line fails. I ran both editions side by side on the same machine:

| PowerShell | `$password` is null | `[PsCredential]::New(...)` |
| --- | --- | --- |
| 5.1 (our CI) | yes | `Exception calling ".ctor" with "2" argument(s): "Cannot process argument because the value of argument "password" is null."` |
| 7.6 | no | works |

That is character for character the message from the failed job. Since it happens in `BeforeAll`, every test of all three contexts fails at once, including the ones that never touch `$public`.

Suggested fix in all three places. `::new()` is not allowed in dbatools because we still support PowerShell v3, so this also takes care of that:

```powershell
$password = ConvertTo-SecureString "<enterStrongPasswordHere>" -AsPlainText -Force
$public = New-Object System.Management.Automation.PSCredential("AsGoodAsPublic", $password)
```

## 2. `test (default)` - trailing spaces

The style test lists nine lines:

* `public/Install-DbaFirstResponderKit.ps1`: 310, 313
* `tests/Install-DbaFirstResponderKit.Tests.ps1`: 187, 198, 256, 267, 290, 326, 337

Most of them sit after a hashtable value, like `SqlInstance      = $TestConfig.InstanceMulti1  `.

## Other things worth a look

**The negative tests can pass without proving anything.** `{ Invoke-DbaQuery @splatPublicBlitz } | Should -Throw` is happy with any exception, including `Login failed for user AsGoodAsPublic`. If the login was never created or the password was wrong, the "Does not let public execute ..." tests would still be green. Asserting the message would make them real, for example:

```powershell
{ Invoke-DbaQuery @splatPublicBlitz } | Should -Throw -ExpectedMessage "*EXECUTE permission*"
```

**Style guide details:** single quotes in `Branch = 'main'`, `@('sp_BlitzFirst', 'sp_BlitzIndex')` and `Database = 'master'` (we use double quotes), and `$FirstResponderKitParams` should be `$splatFirstResponderKit` following the `$splat<Purpose>` convention.

**Small things:** `$resultsDownload` is assigned in all three `BeforeAll` blocks but never used. In the master context the `AfterAll` calls `Remove-DbaDatabase -Database $database` with `$database = 'master'` - that is harmless because `Remove-DbaDatabase` filters system databases out, but it reads worse than it is and can go. And all three contexts create the same login `AsGoodAsPublic`, so if one `AfterAll` does not complete, the next context fails with "already exists" - a `$(Get-Random)` suffix would make them independent.

## Next

Every test died in the setup, so the assertions have not run a single time yet and we do not know yet whether the feature itself works. I will apply the `-Force` fix locally and run the test file against a lab instance, and report back here.

<sub>Analysis done with Claude Code.</sub>

andreasjordan:
Good news: I applied only the `-Force` fix locally and ran the test file against a SQL Server 2025 lab instance.

```
Install-DbaFirstResponderKit.Tests.ps1   Passed   22 total   22 passed   0 failed   152s
```

All 22 pass, including the nine certificate signing tests that had never executed a single assertion on the CI, and the lab was left clean afterwards. So the feature itself is fine - the nine failures really were nothing but the `$null` password, there is nothing deeper hiding behind them.

For the trailing spaces: the rule in `tests/dbatools.Tests.ps1` is

```powershell
Select-String -Path $file.FullName -Pattern "([^ \t\r\n])[ \t]+$"
```

so only lines with content before the trailing whitespace count, which is why indentation-only lines are not in the list. Running that pattern locally gives exactly the nine lines from the failed job. After stripping them the whole style suite passes 11/11. There are another twelve indentation-only lines (149, 157, 184, 195, 206, 226, 253, 264, 287, 295, 323, 334 in the test file) that the CI does not complain about but that are worth removing in the same go.

So the full list to get green is just:

1. `-Force` on the three `ConvertTo-SecureString` calls, and `New-Object System.Management.Automation.PSCredential(...)` instead of `[PsCredential]::New(...)`
2. trailing whitespace removed from both files

Everything else I mentioned above is review polish, not blocking.

<sub>Verified with Claude Code.</sub>

potatoqualitee:
@andreasjordan nailed both of these — I pushed the fix to the branch, plus the polish items from your review.

## The two blocking fixes, as you diagnosed them

1. `-Force` on the three `ConvertTo-SecureString` calls, and `New-Object System.Management.Automation.PSCredential(...)` in place of `[PsCredential]::New(...)`.
2. Trailing whitespace gone from both files — the nine CI-flagged lines and the twelve indentation-only ones.

Style suite on the branch is back to green:

```
Describing dbatools style
  [+] PowerShell files do not contain a UTF-8 BOM
  [+] PowerShell files are not indented with tabs
  [+] PowerShell files do not contain trailing spaces
Tests Passed: 11, Failed: 0
```

## One place I deviated from your suggestion

You suggested `Should -Throw -ExpectedMessage "*EXECUTE permission*"` on the negative tests. That is right for two of the three contexts, but it would have turned **"Testing without certificate signing"** red. Without `-LetPublicExecute` the command never runs `GRANT CONNECT TO [public]`, so that login is stopped one step earlier than in the signing contexts — it cannot open the database at all, and never gets far enough to be refused EXECUTE.

Measured on a SQL 2022 instance:

| Context | `GRANT CONNECT` ran? | Actual message | `*EXECUTE permission*` |
| --- | --- | --- | --- |
| certificate signing in user database | yes | `The EXECUTE permission was denied on the object 'sp_Blitz', database '...', schema 'dbo'.` | matches |
| certificate signing in master | n/a, guest reaches master | `The EXECUTE permission was denied on the object 'sp_Blitz', database 'master', schema 'dbo'.` | matches |
| **without certificate signing** | **no** | `Cannot open database "..." requested by the login. The login failed.` | **does not match** |

So the signing contexts assert `"*EXECUTE permission*"` and the unsigned one asserts `"*Cannot open database*"`.

That still buys exactly what you were after — a bare `Should -Throw` accepting `Login failed for user` was the hole. Both false-green cases you named now fail the assertion:

```
A. login exists, correct password, no access to db  -> Cannot open database "..." requested by the login.   matches: True
B. login exists, WRONG password                     -> Login failed for user 'ProbeWrongPw_...'.            matches: False
C. login does not exist at all                      -> Login failed for user 'NoSuchLogin_...'.             matches: False
```

## Also applied

- `$FirstResponderKitParams` → `$splatFirstResponderKit`, double quotes for `Branch`, `LetPublicExecute` and `Database`.
- Dropped the unused `$resultsDownload` in the three new contexts, and the no-op `Remove-DbaDatabase -Database master`.
- Each context now creates `AsGoodAsPublic_$(Get-Random)`, so an `AfterAll` that does not complete cannot fail the next context with "already exists".

## Verification

Ran the file on **Windows PowerShell 5.1** specifically, since that is the edition that broke it, against a SQL 2022 lab instance. The nine certificate tests pass, and I ran them twice to be sure:

```
Context Testing certificate signing in user database
  [+] Lets public execute sp_BlitzFirst when signed for
  [+] Does not let public execute sp_Blitz when not signed for
  [+] Lets admin execute sp_Blitz when not signed for
Context Testing without certificate signing
  [+] Does not let public execute sp_BlitzFirst
  [+] Does not let public execute sp_Blitz
  [+] Lets admin execute sp_Blitz
Context Testing certificate signing in master database
  [+] Lets public execute sp_BlitzFirst when signed for
  [+] Does not let public execute sp_Blitz when not signed for
  [+] Lets admin execute sp_Blitz when not signed for

Tests Passed: 9, Failed: 0     (both passes)
```

One thing worth knowing, @andreasjordan: on my lab the two "without certificate signing" tests take 11-13s each, because a login that cannot open the database goes through the connection retry before it reports. On one of three runs that retry hit a network timeout and reported `The wait operation timed out` instead, which the tightened assertion then failed on. It should not show up on the runners where SQL is local — but if that lane ever goes flaky on those two tests, that is the reason, and broadening the pattern would be the fix.

The other two contexts on my lab (`with download`, `with LocalFile`) fail for reasons unrelated to this PR — `New-Guid` missing in my stripped-down 5.1 shell and a bad credential for the second instance. Neither touches lines this PR changes.

<sub>Analysis and verification done with Claude Code.</sub>

potatoqualitee:
CI is green on all 13 checks, including the two that were failing:

```
test (MULTI, sql2022, sql2017)   pass   4m5s
test (default)                   pass   2m37s
```

And the certificate tests really executed this time rather than dying in setup — first attempt, no retries:

```
[gha] test passed: Install-DbaFirstResponderKit.Tests.ps1 (68.1s)
```

For comparison, the failing run took 16.3s per attempt across three attempts, because every test aborted in `BeforeAll` on the null password. The jump from ~16s to 68s is the nine certificate assertions actually running.

@andreasjordan — that closes the open question from your first comment. @ReeceGoding — nothing needed from you, the fix is on your branch.

potatoqualitee:
@claude are you working yet? if so, is this the most idiomatic PowerShell name for that parameter?

potatoqualitee:
@claude i updated your token, try again

potatoqualitee:
finally! sorry about that. i revoked 100% of my tokens to fix an issue and forgot to update it here.

potatoqualitee:
@claude PublicExecute sounds good, pls update

potatoqualitee:
Thanks everyone!

ReeceGoding:
Thanks both! Did I do something wrong when I called `Invoke-ManualPester` locally? I thought that it was supposed to warn about the whitespace problems.

andreasjordan:
There is probably an issue with `Invoke-ManualPester`. I will have a look in the next days (if I don't forget it).

ReeceGoding:
I've realised that there is one thing left unanswered: is there any good PowerShell interface for certificates?

PR #10471, opened 2026-07-31
Invoke-ManagedComputerCommand - Throw instead of returning nothing and support SQL Server 2025

andreasjordan:
### Problem

When the local connection attempt failed, the command looped through the versions 16 to 8 to load a matching SqlWmiManagement assembly on the target computer. If all of those attempts failed as well, the `foreach` just ended - no throw, no message, no output at all. The calling command silently continued to work with `$null`, and `-EnableException` never fired because `Get-DbaStartupParameter` and friends never saw an error.

This is what caused `New-DbaFirewallRule -Type DAC` to fail with the misleading message `Cannot bind argument to parameter 'Path' because it is null.` on a fresh SQL Server 2025 instance: `Get-DbaStartupParameter` returned nothing, so the path to ERRORLOG was `$null`.

`Update-DbaServiceAccount` was affected in an even worse way - it reported `Successful` for an account change that never happened.

### Changes

* Throw when the local attempt and all remote attempts fail. The message contains both the reason of the local failure and the reason of the last remote failure.
* Raise `$MaxVersion` from 16 to 17, so a computer that only has SQL Server 2025 installed is supported as well.

All callers already wrap the call in `try`/`catch` with `Stop-Function`, so they now report the real problem instead of failing later on with a confusing message.

### Tests

New `tests/Invoke-ManagedComputerCommand.Tests.ps1` with 8 unit tests covering the local success path, the remote fallback path and the all-attempts-failed path. Verified that 3 of them fail against the previous code.

Smoke tested on a computer without SQL Server, where the command now returns a real error instead of `$null`:

```
Failed to connect to SQL WMI on ADMIN01. Local connection attempt failed with:
Exception calling "Initialize" with "0" argument(s): "SQL Server WMI provider is not available on 192.168.3.20."
| Last remote connection attempt (version 8) failed with:
Exception calling "Initialize" with "0" argument(s): "SQL Server WMI provider is not available on 192.168.3.20."
```

### Note for a possible follow-up

The assembly probe uses `FullName -like "*$version.0*"`, which only matches GAC folders like `v4.0_17.0.0.0`. Should SQL Server 2025 register `SqlWmiManagement` as `17.100.0.0`, version 17 would still not be found - but with this change you now get a real exception naming that instead of silence.

### DIFF
--- a/private/functions/Invoke-ManagedComputerCommand.ps1
+++ b/private/functions/Invoke-ManagedComputerCommand.ps1
@@ -96,6 +96,7 @@ function Invoke-ManagedComputerCommand {
         Invoke-Command2 @parms
     } catch {
         # Log the failure and prepare to connect remotely
+        $localException = $PSItem
         Write-Message -Level Verbose -Message "Local connection attempt to $computer failed | $PSItem. Connecting remotely."
         $hostname = $resolved.FullComputerName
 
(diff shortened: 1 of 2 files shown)


potatoqualitee:
hell yeah thank you 🤩

PR #10472, opened 2026-07-31
New-DbaFirewallRule - Fix DAC rule failing with a misleading message about a null path

andreasjordan:
Two problems in the block that creates the firewall rule for the dedicated admin connection (DAC).

### The path to ERRORLOG was used without any test

`Get-DbaStartupParameter` can return nothing at all, or an object whose `ErrorLog` property is empty. The result went straight into `Invoke-Command2 -ArgumentList $errorLogPath`, so the command failed inside the remote scriptblock with:

```
[New-DbaFirewallRule] Failed to execute command to get information for DAC on <computer> for instance <instance>.
| Cannot bind argument to parameter 'Path' because it is null.
```

That message points at the DAC and hides the real problem, which is that we could not get the startup parameters of the instance. The path is now tested and reported before it is used.

### The ERRORLOG was read without the Credential

The `Invoke-Command2` call that reads ERRORLOG did not pass `-Credential $Credential`, unlike every other remote call of this command. It only worked because `Invoke-Command2` caches the PSSession per computer and the session name does not include the credential, so an earlier call made with the credential was reused. Whenever that earlier call did not happen, the read ran as the current user.

### Tests

Two new contexts in the existing `InModuleScope` Describe of `tests/New-DbaFirewallRule.Tests.ps1`:

* the path to ERRORLOG cannot be determined - the command reports the missing ERRORLOG and never tries to read it
* the path to ERRORLOG is found - the rule is created for the DAC port from ERRORLOG and the credential is used to read it

3 of the 4 new tests fail against the previous code. The old run shows both bugs directly - `Invoke-Command2 ... -ArgumentList $null -Raw` for the first one, and an ERRORLOG read without `-Credential` while the call that creates the rule has it for the second one.

Unit tests and integration tests were run against a lab instance, the integration tests also cover this code path because the DAC block runs when `-Type` is not used.

### DIFF
--- a/public/New-DbaFirewallRule.ps1
+++ b/public/New-DbaFirewallRule.ps1
@@ -402,9 +402,21 @@ function New-DbaFirewallRule {
                 # As we create firewall rules, we probably don't have access to the instance yet. So we have to get the port of the DAC via Invoke-Command2.
                 # Get-DbaStartupParameter also uses Invoke-Command2 to get the location of ERRORLOG.
                 # We only scan the current log because this command is typically run shortly after the installation and should include the needed information.
+                $errorLogPath = $null
                 try {
                     $errorLogPath = Get-DbaStartupParameter -SqlInstance $instance -Credential $Credential -Simple -EnableException | Select-Object -ExpandProperty ErrorLog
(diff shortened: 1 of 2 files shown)


potatoqualitee:
looks good 👍🏼  how you liking these tests? love it, just flying like the old days.

andreasjordan:
The new CI pipeline is fantastic, really love it.

PR #10474, opened 2026-07-31
Tests - Make weak throw assertions strict and stop an external outage from failing unrelated PRs

andreasjordan:
Follow-up to #10473. While fixing the mocks there it turned out that both tests of `Get-DbaStartupParameter` had been passing for the wrong reason: they only asserted `Should -Throw`, so the error about a service that was not found satisfied them just as well as the error the test name describes. A survey of the whole test suite found 116 `Should -Throw` assertions, 47 of them without any expected message. Most of those are fine, because they test parameter validation and nothing else can run before the throw. Nine of them are not.

### The nine

| File | Test |
| --- | --- |
| `Test-DbaDbRecoveryModel` | Should Throw Error for a DB Connection Error |
| `Test-DbaDbRecoveryModel` | Should Throw Error for Output Error |
| `Get-XpDirTreeRestoreFile` | Should throw on an invalid SQL Connection |
| `Get-XpDirTreeRestoreFile` | Should throw if SQL Server can't see the path |
| `Test-DbaPath` | Honors EnableException when xp_fileexist execution fails |
| `Get-DbaComputerSystem` | Cannot resolve hostname of computer |
| `Get-DbaOperatingSystem` | Cannot resolve hostname of computer |
| `Get-DbaService` | Cannot resolve hostname of computer |
| `Get-DbaStartupParameter` | (both, already done in #10473) |

### Changes

`Test-DbaDbRecoveryModel` and `Get-XpDirTreeRestoreFile` mocked a command with a bare `throw`, which produces the message `ScriptHalted` for every one of them - two tests that describe two different failures asserted on the very same indistinguishable message. The mocks now throw a distinctive message and the test asserts on it.

The other tests assert on the message the command really produces. All messages were captured from an actual run first, not guessed:

* `Test-DbaPath` -> `*xp_fileexist*`
* `Get-XpDirTreeRestoreFile` (path) -> `*cannot access c:\dummy*`
* the three hostname tests -> `*DNS name DoesNotExist142 not found*`

### Proof that this helps

With both mocks of `Test-DbaDbRecoveryModel` in place at the same time, so that the command dies while connecting and never reaches the output:

```
[+] OLD assertion passes even though it died at connect, not at output
[-] NEW assertion fails because it died at connect, not at output
    Expected an exception with message like 'Mocked output failure' to be thrown,
    but the message was 'Mocked connection failure'.
```

### Also in here

`Get-DbaBuild` had the same test for `Build` and `SqlInstance` twice, with identical bodies. The second one now covers `Kb` and `SqlInstance`, which was the only pair of mutually exclusive parameters without a test. Its name does not use single quotes like the tests around it, because the style check of the repository does not allow them.

### Not changed

The three hostname tests contain `Mock Resolve-DbaNetworkName { $null }` without `-ModuleName dbatools`. The mock therefore never applies to the command inside the module, and the tests really depend on DNS failing for `DoesNotExist142` - which is what the new expected message documents. Making those mocks effective is a separate change and not a trivial one: with a working mock `Get-DbaService` does not throw at all any more.

### Tests

All seven test files were run against a lab instance, unit and integration, everything passes and the lab is left clean. The structural checks report only the nine failures that already exist in other files.

---

## Also: stop an external outage from failing unrelated PRs

`Get-DbaKbUpdate.Tests.ps1` queries the Microsoft Update Catalog. When that site does not answer, each of its six tests waits 100 seconds for the default timeout of `Invoke-WebRequest` and then fails. With the three attempts of the CI that is half an hour of red build for a reason that has nothing to do with dbatools. The file is in no group of `pester.groups.ps1`, so it runs in the `default` scenario of every single pull request - it turned this PR red while the catalog was down, and it would do the same to anyone else.

The tests now run inside a context that is skipped when the catalog does not answer. We only skip when it is provably down: if the catalog answers, the tests run exactly as before and a real regression still fails them.

One detail worth keeping: the probe asks for the **search** page, not the start page. While writing this the start page answered in one second while every search timed out, so a probe of the start page would have skipped nothing. My first attempt did exactly that and the tests still ran into their timeouts.

Measured against the live outage:

| | before | after |
| --- | --- | --- |
| result | failed | passed, 6 skipped |
| duration | ~30 min over 3 attempts | 31 s |

And verified the other direction with the same code shape pointed at an endpoint that does answer: 1 passed, 0 skipped - so the skip is conditional and does not silently swallow the suite.

### DIFF
--- a/tests/CLAUDE.md
+++ b/tests/CLAUDE.md
@@ -54,7 +54,7 @@ Pester v5 enforces strict organization of test code. Follow these rules:
 - **All setup code** must be in `BeforeAll` or `BeforeEach` blocks
 - **All cleanup code** must be in `AfterAll` or `AfterEach` blocks
 - **All test assertions** must be in `It` blocks
-- **No loose code** is allowed in `Describe` or `Context` blocks
+- **No loose code** is allowed in `Describe` or `Context` blocks, with one exception: computing a value that a `-Skip:` needs, see [Skip Conditions](#skip-conditions)
 - **Never use the `-ForEach` parameter** on any test blocks
 
 ### Standard Test Structure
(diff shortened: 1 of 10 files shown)


potatoqualitee:
ohh i like that last one especially

PR #10475, opened 2026-07-31
Sync-DbaAvailabilityGroup - Open one shared dedicated admin connection instead of three

### LINKED ISSUE #10467: Sync-DbaAvailabilityGroup opens 3 separate DAC connections back-to-back (Credential/DbMail/LinkedServer), causing intermittent "forcibly closed" / max DAC connection errors

We run a nightly PowerShell script that keeps the secondary node(s) of an AlwaysOn Availability Group in sync with the primary, using Sync-DbaAvailabilityGroup. Every run produces warnings like:

_WARNING: [Copy-DbaCredential] Failure | An existing connection was forcibly closed by the remote host
WARNING: [Copy-DbaDbMail] Failure | An existing connection was forcibly closed by the remote host
WARNING: [Copy-DbaLinkedServer] Failure | An existing connection was forcibly closed by the remote host_

All three occur within the same second of each other. The SQL Error Log on the primary shows the corresponding entries at the same timestamps:

_Could not connect because the maximum number of '1' dedicated administrator connections already exists. Before a new connection can be made, the existing dedicated administrator connection must be dropped, either by logging off or ending the process._

Below is the response from Claude Sonnet, which I ran in my own environment to analyze the command and its functionality:
Root cause identified by inspecting the actual v2.8.3 source (public/Sync-DbaAvailabilityGroup.ps1): the function calls Copy-DbaCredential, Copy-DbaDbMail, and Copy-DbaLinkedServer passing the same plain, non-DAC $server connection object to all three:

_Copy-DbaCredential   -Source $server -Destination $secondaries -Credential $Credential -ExcludePassword:$ExcludePassword -Force:$Force
Copy-DbaDbMail       -Source $server -Destination $secondaries -Credential $Credential -ExcludePassword:$ExcludePassword -Force:$Force
Copy-DbaLinkedServer -Source $server -Destination $secondaries -Credential $Credential -ExcludePassword:$ExcludePassword -Force:$Force_

$server is never replaced with a DAC connection at the Sync-DbaAvailabilityGroup level, so inside each Copy-* command the reuse check always evaluates to $false:

$dacConnected = $Source.Type -eq 'Server' -and $Source.InputObject.Name -match '^ADMIN:'

As a result, each of the three commands independently opens its own new DAC (Connect-DbaInstance -DedicatedAdminConnection), uses it, and disconnects it in its own end block. This produces three separate DAC open→use→close cycles fired back-to-back with zero delay in a single Sync-DbaAvailabilityGroup run. SQL Server does not always release the DAC slot instantly after a clean disconnect, so the next command's connect attempt right on its heels intermittently fails/resets producing the warnings and the correlating SQL Error Log entries above.

This is not the bug fixed in #10284 (that was about reuse detection failing when a DAC was already being shared, e.g. inside Start-DbaMigration). The reuse mechanism introduced by that fix is correct. It simply never gets exercised here, because Sync-DbaAvailabilityGroup itself never establishes one shared DAC and passes it through.

### Steps to Reproduce

# Values below are illustrative / redacted - replace with your own instance names.
$primary   = 'SQLPRIMARY\INSTANCE'
$secondary = 'SQLSECONDARY\INSTANCE'
$executingUser = "$env:USERDOMAIN\svc-agsync"

Sync-DbaAvailabilityGroup `
    -Primary $primary `
    -Secondary $secondary `
    -Exclude AgentOperator, CustomErrors `
    -ExcludeLogin sa, $executingUser `
    -Force `
    -Verbose `
    -WarningVariable syncWarnings `
    -ErrorVariable syncErrors

2.8.3

### Other details or mentions

This reproduces consistently every night the sync job runs against any primary that has at least one SQL credential, one DB Mail account backed by a credential, and one linked server with a stored password (i.e., whenever all three DAC-needing Copy-* commands execute in the same Sync-DbaAvailabilityGroup invocation).

Suggested remediation approach based on an analysis performed with Claude in my own environment: have Sync-DbaAvailabilityGroup itself open a single DAC connection up front (when any of Credentials, DatabaseMail, or LinkedServers is not excluded and -ExcludePassword is not set), and pass that DAC-connected server object as -Source to all three sub-commands, so the existing dacConnected reuse logic in Copy-DbaCredential/Copy-DbaDbMail/Copy-DbaLinkedServer actually triggers instead of each opening its own.

I previously reported a related issue (#10284, fixed in v2.8.0) about broken DAC reuse detection and excessive Update-SqlPermission connection churn but this is a separate, still-open gap in Sync-DbaAvailabilityGroup's own orchestration of the three Copy-* commands.

Windows PowerShell ISE (powershell_ise.exe), Windows PowerShell (powershell.exe)

### PowerShell Host Version

PSVersion                      5.1.17763.9020                                                                                                                                                             
BuildVersion                   10.0.17763.9020                                                                                                                                                            

Microsoft SQL Server 2019 (RTM-CU32-GDR) (KB5102335) - 15.0.4480.2 (X64)   Jun 18 2026 17:54:32   Copyright (C) 2019 Microsoft Corporation  Enterprise Edition: Core-based Licensing (64-bit) on Windows Server 2019 Datacenter 10.0 <X64> (Build 17763: ) (Hypervisor) 

### .NET Framework Version

.NET Framework 4.7.4135.0

andreasjordan:
Fixes #10467

## Problem

`Sync-DbaAvailabilityGroup` passed the plain connection to `Copy-DbaCredential`, `Copy-DbaDbMail` and `Copy-DbaLinkedServer`, so the reuse detection in those three commands never triggered and each of them opened, used and closed its own dedicated admin connection. SQL Server allows only one DAC per instance, and three open/close cycles fired back-to-back intermittently failed with `An existing connection was forcibly closed by the remote host`, matching `Could not connect because the maximum number of '1' dedicated administrator connections already exists` in the error log of the primary.

This is a regression. The shared DAC was added in #10163 and removed again in #10346, which also added a unit test asserting that `Sync-DbaAvailabilityGroup` opens no DAC.

## Solution

`Sync-DbaAvailabilityGroup` opens a single DAC and passes it to all three commands, so their existing reuse detection triggers.

Not by restoring #10163 verbatim, though. That version replaced `$server` itself with the DAC, so every other command (`Copy-DbaSpConfigure`, `Copy-DbaLogin`, `Update-SqlDbOwner`, the agent copies, `Sync-DbaLoginPermission`) ran over the single restricted DAC session, and it made `Get-DbaAvailabilityGroup | Sync-DbaAvailabilityGroup` fail hard unless the caller had connected with `-DedicatedAdminConnection`. This PR follows the pattern that `Start-DbaMigration` already uses instead:

* `$server` stays a normal connection and keeps serving all the commands that do not need a DAC.
* A separate `$serverDac` is opened once and passed to the three password-aware commands.
* It is only opened when it is needed, meaning at least one of `Credentials`, `DatabaseMail` or `LinkedServers` is not excluded and `-ExcludePassword` is not set. It is opened directly before the credentials step and closed directly after the linked server step, so it is not held during the rest of the sync.
* A DAC that is already open, for example from `-Primary $dacServer` or from a piped availability group, is reused rather than replaced, and an additional normal connection is opened for everything else. No hard failure on the pipeline path.
* If the DAC cannot be opened, the command warns and lets the copy commands try on their own, which is what happens today. A DAC failure does not abort the whole sync.

The duplicated DAC detection in the affected commands is replaced by a new private function `Test-DacConnection`. It existed in two slightly different variants across eight commands, checking either `Name` or `ConnectionContext.ServerInstance`; the helper checks both, so detection can only get more reliable.

## Tests

The unit test that asserted no DAC is opened is inverted: exactly one DAC opened, the same connection passed to all three commands, one disconnect. A second test covers the reuse path: no new DAC, the DAC goes to the password commands, the normal connection goes to the others, and the caller's DAC is not disconnected.

Verified against real instances beyond the mocks. `Test-DacConnection` returns `$true` for a real DAC both as a server object and as a `DbaInstanceParameter`, and `$false` for a normal connection and for a plain instance name. An end-to-end run produces the intended trace:

```
[Sync-DbaAvailabilityGroup] Opening dedicated admin connection for password retrieval.
[Copy-DbaCredential]        Reusing dedicated admin connection for password retrieval.
[Copy-DbaDbMail]            Reusing dedicated admin connection for password retrieval.
[Copy-DbaLinkedServer]      Reusing dedicated admin connection for password retrieval.
[Disconnect-DbaInstance]    removing from connection hash
```

One DAC instead of three, no warnings. The reuse path was checked separately: `Copy-DbaSpConfigure` ran over the normal connection and the caller's DAC was still usable after the run.

## Note for the reporter

Going from three DAC connections to one materially reduces the exposure window, but it is worth being precise about what this does not prove. Each copy command already disconnected its DAC in its own `end` block and DAC connections are non-pooled, so three strictly sequential cycles should not normally collide unless the server-side teardown lags the client-side close or something else on the instance competes for the single DAC slot. If the error still shows up occasionally after this ships, the next thing to check is whether anything else grabs the DAC of the primary at that time.

### DIFF
--- a/private/functions/Test-DacConnection.ps1
+++ b/private/functions/Test-DacConnection.ps1
@@ -1,0 +1,66 @@
+function Test-DacConnection {
+    <#
+    .SYNOPSIS
+        Internal function.
+
+    .DESCRIPTION
+        Tests whether a connection that was passed in is already a dedicated admin connection (DAC).
+
+        SQL Server only allows one DAC per instance, so commands that need a DAC to decrypt passwords
+        have to reuse an existing one instead of opening a second one. Connect-DbaInstance prefixes the
+        server name with "ADMIN:" when -DedicatedAdminConnection is used, so that prefix is what marks
+        a connection as a DAC.
+
+        This function is used by the following public functions:
+        - Copy-DbaCredential
+        - Copy-DbaDbMail
+        - Copy-DbaLinkedServer
+        - Export-DbaCredential
+        - Export-DbaInstance
+        - Export-DbaLinkedServer
+        - Invoke-DbaDbDecryptObject
+        - Start-DbaMigration
+        - Sync-DbaAvailabilityGroup
+
+    .PARAMETER InputObject
+        The connection to test. Accepts a DbaInstanceParameter (as used by the SqlInstance and Source
+        parameters of the public commands) or an SMO server object. Anything that is not a connection
+        to an already connected instance returns $false.
+
+    .NOTES
+        Tags: Connection, DAC
+        Author: the dbatools team + Claude
+
+        Website: https://dbatools.io
+        Copyright: (c) 2018 by dbatools, licensed MIT
+        License: MIT https://opensource.org/licenses/MIT
+
+    .EXAMPLE
+        Test-DacConnection -InputObject $Source
+
+        Returns $true if $Source is a server object that is connected via a dedicated admin connection.
+    #>
+    [CmdletBinding()]
+    [OutputType([bool])]
+    param(
+        [AllowNull()]
+        $InputObject
+    )
+
+    if ($null -eq $InputObject) {
+        return $false
+    }
+
+    if ($InputObject.GetType().FullName -eq "Dataplat.Dbatools.Parameter.DbaInstanceParameter") {
+        # Only a DbaInstanceParameter of type Server holds an already connected server object
+        if ($InputObject.Type -ne "Server") {
+            return $false
+        }
+        $server = $InputObject.InputObject
+    } else {
+        $server = $InputObject
+    }
+
+    # Depending on how the connection was built, the "ADMIN:" prefix is visible on the server object, on the connection context, or both
+    return [bool]($server.Name -match "^ADMIN:" -or $server.ConnectionContext.ServerInstance -match "^ADMIN:")
+}
--- a/public/Copy-DbaCredential.ps1
+++ b/public/Copy-DbaCredential.ps1
@@ -156,7 +156,7 @@ function Copy-DbaCredential {
             if ($ExcludePassword) { $dacNeeded = $false } else { $dacNeeded = $true }
 
             # Do we have a dedicated admin connection already?
-            $dacConnected = $Source.Type -eq 'Server' -and $Source.InputObject.Name -match '^ADMIN:'
+            $dacConnected = Test-DacConnection -InputObject $Source
 
             $dacOpened = $false
             if ($dacNeeded) {
--- a/public/Copy-DbaDbMail.ps1
+++ b/public/Copy-DbaDbMail.ps1
@@ -390,7 +390,7 @@ function Copy-DbaDbMail {
             if ($ExcludePassword) { $dacNeeded = $false } else { $dacNeeded = $true }
 
             # Do we have a dedicated admin connection already?
-            $dacConnected = $Source.Type -eq 'Server' -and $Source.InputObject.Name -match '^ADMIN:'
+            $dacConnected = Test-DacConnection -InputObject $Source
 
             $dacOpened = $false
             if ($dacNeeded) {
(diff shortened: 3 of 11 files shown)


potatoqualitee:
interesting! one of the hardest problems.

PR #10478, opened 2026-08-01
Network certificate commands - Find the configured certificate even when it is archived

### LINKED ISSUE #10462: Get-DBANetworkCertificate Not Returning Archived Certificate Information

When attempting to retrieve certificate information from a SQL Server running on Windows using an archived certificate, no information is returned about the certificate.

This will occur with auto-renewed certificates immediately after they are renewed and archived, but before SQL Server has been updated to use the new certificate.

### Steps to Reproduce

Find the thumbprint of the certificate used by SQL Server and archive it (replace CERT_THUMBPRINT):

```powershell
$cert = Get-ChildItem Cert:\LocalMachine\My\CERT_THUMBPRINT
$cert.Archived = $true
#Revert afterwards
# $cert.Archived = $false
```

Then attempt to call `Get-DBANetworkCertificate` against the server (which calls `Get-DBANetworkConfiguration`).

Image

Image

2.3.8

### Other details or mentions

The thumbprint of the cert is found in the registry, but fails to pull the certificate via Get-ChildItem as the default is to hide archived certs.

Adding -Force to Get-ChildItem will show archived certs and pull the certificate information.

I can submit a pull request against `Get-DBANetworkConfiguration` (this is called by `Get-DBANetworkCertificate`) to add the `-Force` parameter to the first call to `Get-ChildItem` when querying for the existing certificate.

I don't think it needs to be added to the second call which finds suitable certificates.

Image

### PowerShell Host Version

5.1.26100.6584

SQL Server 2025

### .NET Framework Version

4.8.09221, 4.0.0.0

andreasjordan:
Fixes #10462

## Problem

`Get-DbaNetworkCertificate` returns nothing for an instance whose certificate has been archived. This happens with auto-renewed certificates: they are archived as soon as the successor is issued, but SQL Server keeps using them until it is reconfigured.

The certificate provider hides archived certificates unless `-Force` is passed. So `Get-DbaNetworkConfiguration` reads the thumbprint from `SuperSocketNetLib\Certificate` correctly, but the store lookup returns nothing and every derived property (`FriendlyName`, `DnsNameList`, `Thumbprint`, `Generated`, `Expires`, `IssuedTo`, `IssuedBy`) is empty. `Get-DbaNetworkCertificate` filters on `Where-Object Thumbprint`, so the instance disappears from the output entirely.

## Change

Added `-Force` to the three lookups that find the *already configured* certificate by thumbprint:

* `Get-DbaNetworkConfiguration` - the reported bug
* `Test-DbaNetworkCertificate` - reported `CertificateFound = $false` for an archived certificate
* `Set-DbaNetworkCertificate` - threw "Can't find private key path" when re-applying an archived thumbprint

The scan for *suitable* certificates in `Get-DbaNetworkConfiguration` is deliberately left unchanged - an archived certificate should not be recommended as a candidate.

## Test

Added one integration test to `Set-DbaNetworkCertificate.Tests.ps1`, where an instance with a configured certificate already exists. It archives the configured certificate, asserts that `Get-DbaNetworkCertificate` still returns the thumbprint and that `Test-DbaNetworkCertificate -Thumbprint` reports `CertificateFound`, then un-archives it in a `finally` block so the existing cleanup can still remove it.

Thanks to @haxr for the precise report and for pinpointing both the cause and the correct scope of the fix.

### DIFF
--- a/public/Get-DbaNetworkConfiguration.ps1
+++ b/public/Get-DbaNetworkConfiguration.ps1
@@ -200,7 +200,8 @@ function Get-DbaNetworkConfiguration {
                 try {
                     $acceptedSPNs = (Get-ItemProperty -Path $regPath -Name AcceptedSPNs).AcceptedSPNs
                     $thumbprint = (Get-ItemProperty -Path $regPath -Name Certificate).Certificate
-                    $cert = Get-ChildItem Cert:\LocalMachine -Recurse -ErrorAction SilentlyContinue | Where-Object Thumbprint -eq $thumbprint | Select-Object -First 1
+                    # -Force is needed to also find archived certificates, as SQL Server keeps using a certificate that has been archived.
+                    $cert = Get-ChildItem Cert:\LocalMachine -Recurse -Force -ErrorAction SilentlyContinue | Where-Object Thumbprint -eq $thumbprint | Select-Object -First 1
                     $extendedProtection = switch ((Get-ItemProperty -Path $regPath -Name ExtendedProtection).ExtendedProtection) { 0 { $false } 1 { $true } }
                     $forceEncryption = switch ((Get-ItemProperty -Path $regPath -Name ForceEncryption).ForceEncryption) { 0 { $false } 1 { $true } }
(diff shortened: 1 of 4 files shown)


potatoqualitee:
thank you both 🙏🏼

PR #10479, opened 2026-08-01
Claude Code - Add an issue workflow skill and document the test instance fixtures

andreasjordan:
Documentation and tooling only - no command changes, no behaviour changes.

## `/issue` skill

`.claude/skills/issue/SKILL.md` takes a GitHub issue from report to pull request. It complements the existing `/precommit` and `/codex` skills, which cover code quality but say nothing about how an issue becomes a branch.

Two steps in it exist because both were missed working #10462 by hand:

* **Verify the report against `development` first.** Reporters cite the version they run, which is usually several releases behind. Confirm the defect still exists before diagnosing it.
* **Grep for the same defect in the sibling commands before proposing a scope.** An issue names the command the reporter happened to run. #10462 named one command; the same defect was in three, because `Get-`/`Set-`/`Test-` of the same noun share the mechanism.

The skill also writes down the conventions that are currently tribal knowledge: branch from the tip of `development` rather than whatever is checked out, `(do ...)` belongs in the commit message and not the PR title, base is always `development`, credit the reporter, and never merge. It has a triage-only mode as well, since plenty of issues need an informed comment rather than a branch - and that mode drafts the comment for maintainer approval instead of posting it.

## Test instance fixtures

Added a "Choosing a Test Instance" section to `tests/CLAUDE.md`, derived from `private/testing/Get-TestConfig.ps1` and `pester.groups.ps1`. It documents two things that are not visible from the test files themselves:

* The `$TestConfig.Instance*` variable a test file references is what assigns that file to an AppVeyor scenario, so picking a heavier instance than needed moves the test into a scarcer, slower scenario for no benefit.
* Only `InstanceSingle`, `InstanceMulti1` and `InstanceMulti2` exist on GitHub Actions. A test written against `InstanceCopy*`, `InstanceHadr` or `InstanceRestart` runs on AppVeyor only - worth knowing before writing a test that can never fail in the environment you are watching.

Also notes that `InstanceRestart` tests share machine state, so anything `AfterAll` cannot reliably undo should be restored in a `try`/`finally` inside the `It`.

Companion to #10478, which came out of the same issue.

### DIFF
--- a/.claude/skills/issue/SKILL.md
+++ b/.claude/skills/issue/SKILL.md
@@ -37,6 +37,24 @@ Reporters cite the version they run, which is usually behind. Before diagnosing 
 
 Take the reporter's analysis seriously — it is often correct and specific — but confirm it in the code yourself before acting on it.
 
+### Check the dependency versions too, not just the dbatools version
+
+For anything touching connections, SMO, authentication, encryption or bulk copy, the module version is the *less* interesting number. The behaviour usually comes from `dbatools.library` and the `Microsoft.Data.SqlClient` it ships, and a report is frequently against a library that is months old.
+
+```powershell
+# what the repo currently requires
+Get-Content .github/dbatools-library-version.json
+
+# what the reporter would get today, and which SqlClient it carries
+$lib = (Get-Module dbatools.library -ListAvailable | Select-Object -First 1).ModuleBase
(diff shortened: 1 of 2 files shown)


andreasjordan:
@potatoqualitee please have a detailed look at the changes to CLAUDE.md - maybe Claude was wrong about the CI tooling.

potatoqualitee:
looks good! thank you 🙏🏼

PR #10480, opened 2026-08-01
Set-DbaAgentSchedule - Keep the frequency settings when they are not part of the change

### LINKED ISSUE #10461: Running Set-DbaAgentSchedule to modify only StartDate, existing Schedule Frequency values are set to Defaults
### Summarize Functionality

Had an existing Job Schedule that was set with specific frequency settings. Wanted to only update the StartTime on the Schedule, which set the frequency settings back to default on execution. I had to then reconfigure the original frequency settings back to the original values.

Yes

### Technical Details

Commands run to produce the issue are in the attached txt file.
CoPilot chat via shared link with suggested mods is in the following link as well.
https://github.com/copilot/share/485f0280-4104-8874-9053-3e43000901f0

[dbatools Set-DbaAgentSchedule Issue.txt](https://github.com/user-attachments/files/30180205/dbatools.Set-DbaAgentSchedule.Issue.txt)

I also tried the following command as well, as I thought I may have messed up the date value:
Set-DbaAgentSchedule -SqlInstance $SQLJobSched -Schedule DBA_Index_Stats -Job $JobList.Name -StartTime '235900' -Confirm
But, same issue as previous.

andreasjordan:
Fixes #10461

## Problem

Changing only the start time of an existing schedule silently destroyed its recurrence pattern:

| | FrequencyTypes | FrequencyInterval | SubDayTypes | SubDayInterval |
|---|---|---|---|---|
| before | Daily | 1 | Hour | 4 |
| after `-StartTime 235900` | OneTime | 0 | Unknown | 0 |

It is broader than the start time. **Any** call that omits `-FrequencyType` did this - `-NewName`, `-StartDate`, `-Enabled` and `-Disabled` included. A call of `Set-DbaAgentSchedule -Enabled` alone was enough to wreck a production schedule.

## Mechanism

An unbound `-FrequencyType` is `$null`. `$null -notin 1, 4, 8, 16, 32, 64, 128` is `$true`, so the conversion switch ran, nothing matched, and `default { 1 }` set it to `1` (OneTime). The guard that decides whether to write the value is `if ($FrequencyType -ge 1)`, which a defaulted `1` satisfies, so the schedule was converted to one-time. SQL Server then clears the interval and subday columns, because they carry no meaning for a one-time schedule - which is why the whole pattern was lost rather than just the type.

## Change

* `Test-Bound` now decides whether the frequency is written at all, so an unbound `-FrequencyType` stays `$null` and the existing `-ge 1` guards no longer fire
* `$Interval` is initialized outside the block that is skipped when no frequency type is passed, since it is still read further down
* The compensation for newly created schedules without an interval only applies when the caller is actually changing the frequency, so a start time change no longer writes an interval nobody asked for

`New-DbaAgentSchedule` contains the same `default { 1 }` and keeps it. Defaulting to OneTime is correct when *creating* a schedule - there is no existing recurrence to lose. `Set-DbaAgentJob` and `Set-DbaAgentJobStep` do not use this pattern at all, so the defect is confined to this one command.

The subday, relative interval and recurrence factor parameters already had guards that unbound values could not satisfy, so they are left alone.

## Testing

Added one context to `Set-DbaAgentSchedule.Tests.ps1` covering both a `-StartTime`-only and an `-Enabled`-only change against a daily/every-4-hours schedule.

Verified in a lab against SQL Server 2019:

* with the fix: 10/10 tests pass
* with the fix reverted and the tests kept: the two frequency-preservation tests fail with `Expected 'Daily', but got OneTime` - the exact symptom from the issue

Thanks to @Ratty25 for the clear report and the reproduction steps.

### DIFF
--- a/public/Set-DbaAgentSchedule.ps1
+++ b/public/Set-DbaAgentSchedule.ps1
@@ -202,8 +202,16 @@ function Set-DbaAgentSchedule {
     begin {
         if ($Force) { $ConfirmPreference = 'none' }
 
+        # Only touch the frequency of an existing schedule if the user asked for it.
+        # An unbound -FrequencyType is $null, which is "-notin" the list below, so it used to fall
+        # through to the default of 1 (OneTime) and silently overwrite the recurrence of the schedule.
+        # SQL Server then clears the interval and subday values, because they carry no meaning for a
+        # one-time schedule, so a call that only changed the start time destroyed the whole pattern.
+        $frequencyTypeBound = Test-Bound -ParameterName FrequencyType
+        $frequencyIntervalBound = Test-Bound -ParameterName FrequencyInterval
+
         # Check of the FrequencyType value is of type string and set the integer value
-        if ($FrequencyType -notin 1, 4, 8, 16, 32, 64, 128) {
+        if ($frequencyTypeBound -and $FrequencyType -notin 1, 4, 8, 16, 32, 64, 128) {
             [int]$FrequencyType =
             switch ($FrequencyType) {
                 "Once" { 1 }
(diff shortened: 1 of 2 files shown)


potatoqualitee:
Thank you 🙏🏼

PR #10481, opened 2026-08-01
Tests - Make three dead mocks apply, harden the update catalog probe and clear the last top-level statements

andreasjordan:
Four small, unrelated test-only follow-ups that were found while fixing other things and deliberately left out of those PRs to keep them focused. One commit per concern, so they can be reviewed independently.

## Probe the update catalog for a download button instead of a status code

The `-Skip:` probe in `Get-DbaKbUpdate.Tests.ps1` and `Save-DbaKbUpdate.Tests.ps1` treated any 200 as "the catalog works". On 2026-08-01 the search page answered 200 in about a second but rendered no results table and no download buttons for any query, so the tests ran and failed instead of skipping - exactly what the probe was meant to prevent. It now looks for the download button the command itself parses, so the probe is true exactly when the command can work.

`tests/CLAUDE.md` documents this probe as the example of the skip-condition exception, so it got the same update.

The same commit removes the dead `$filesToRemove` in `Save-DbaKbUpdate.Tests.ps1`. It was appended to seven times but never initialised and never read - the cleanup happens through `$tempPath` in `AfterAll`.

## Make the three hostname resolution mocks actually apply

`Get-DbaComputerSystem`, `Get-DbaOperatingSystem` and `Get-DbaService` each had a `Mock Resolve-DbaNetworkName` without `-ModuleName dbatools`, so the mock never applied inside the module. The tests really depended on DNS failing for `DoesNotExist142` on whatever machine ran them, and they asserted on the message of `Resolve-DbaNetworkName` rather than on the command under test.

The mocks now target the module and the assertions moved to what each command does with an unresolvable name. `Get-DbaService` needed a different mock: unlike the other two it calls `Resolve-DbaNetworkName` with `EnableException` and catches the exception, so a mock returning `$null` makes it stop throwing altogether. Its mock throws instead, and the assertion matches the mock's own text so the test cannot pass without the mock applying.

## Keep only Describe blocks at the top level of five test files

These were the last five files failing the "Has only Describe blocks at the top level" check that `TestTestfiles.Tests.ps1` in the testing-dbatools repo runs over every test file. After this the whole tree passes it.

- `Get-DbaDump` and `New-DbaComputerCertificate` wrapped their integration `Describe` in `if (-not $env:appveyor)`, which is now `-Skip:([bool]$env:appveyor)`.
- `Import-DbaParquet` computed `$hasIntegrationConfig` at the top level. Pester evaluates `-Skip:` during discovery, so the check cannot move into `BeforeAll` and is written out in the `-Skip:` expression instead. The unrelated `$null` guard for `$PSDefaultParameterValues` went with it - no other test file has one, and the integration `Describe` is skipped when `$TestConfig` is missing anyway.
- `Remove-DbaAgentJobSchedule` declared a `MockAgentServer` class, and a class can only be declared at the top level of a file. The mock server is now a string carrying the properties the command reads, which also keeps it bindable to `$PSCmdlet.ShouldProcess` - an added `ToString` on a `PSCustomObject` is not used when PowerShell binds a method argument.
- `Update-DbaInstance` assigned `$exeDir` at the top level; it moved into the `BeforeAll` of the two `Describe` blocks that use it.

## Verification

- Unit tests of all ten changed files pass.
- The `Get-DbaKbUpdate` and `Save-DbaKbUpdate` integration tests run and pass against the live catalog, so the stricter probe does not over-skip.
- A discovery-only Pester run confirms the five restructured files discover, run and skip exactly the same tests as before.
- `TestTestfiles.Tests.ps1 -ExcludeTagFilter Goal`: 0 failures over 723 files.
- `TestTestLayout.ps1` reports the same 82 pre-existing files before and after; none of them are touched here.

### DIFF
--- a/tests/CLAUDE.md
+++ b/tests/CLAUDE.md
@@ -276,8 +276,9 @@ Pester evaluates `-Skip:` while it discovers the tests, before any `BeforeAll` r
 ```powershell
 Describe $CommandName -Tag IntegrationTests {
     # These tests need the Microsoft Update Catalog. If it does not answer, every one of them waits
-    # 100 seconds for the default timeout of Invoke-WebRequest and then fails for a reason that has
-    # nothing to do with dbatools. So we probe it once here and skip the tests instead.
+    # for the timeout of Invoke-WebRequest and then fails for a reason that has nothing to do with
+    # dbatools. So we probe it once here and skip the tests instead. We probe for the download button
+    # the command itself looks for, because the catalog also answers 200 with an empty results page.
     try {
         $splatCatalog = @{
             Uri             = "https://www.catalog.update.microsoft.com/Search.aspx?q=KB2992080"
@@ -285,7 +286,8 @@ Describe $CommandName -Tag IntegrationTests {
             TimeoutSec      = 30
             ErrorAction     = "Stop"
         }
-        $catalogReachable = (Invoke-WebRequest @splatCatalog).StatusCode -eq 200
+        $catalogResponse = Invoke-WebRequest @splatCatalog
+        $catalogReachable = [bool]($catalogResponse.InputFields | Where-Object { $PSItem.type -eq "Button" -and $PSItem.class -eq "flatBlueButtonDownload focus-only" })
     } catch {
         $catalogReachable = $false
     }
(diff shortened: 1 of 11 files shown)


potatoqualitee:
looks good!

PR #10482, opened 2026-08-01
Update-DbaInstance - Un-skip the unit tests, and allow -ForEach when the cases come from BeforeDiscovery

andreasjordan:
Two commits: the first makes `tests/Update-DbaInstance.Tests.ps1` run again, the second changes the test guides to permit the syntax it needed.

## Why the file was skipped

The unit test `Describe` has carried `-skip` and a `# TODO: This test needs a lot of care` since d81b4fe53, *"Fix tests that needed extra care after migration to pester 5"*. Removing the skip gives **15 passed, 53 failed**. All 53 come from three defects, none of them about the command:

**1. Forty-four tests had empty bodies.** A `foreach` loop emitted the `It` blocks during discovery. The names interpolated fine, so the test report looked healthy, but `$v`, `$cuLevel` and `$steps` were gone by the time the bodies ran - every one of them called `Update-DbaInstance -Version ""` and died on parameter validation. The `Mock` inside the same loop was registered during discovery and applied to nothing.

**2. Seven `Should -Throw` assertions could never match.** `-ExpectedMessage` compares with `-like` and none of them had wildcards, e.g. `Should -Throw "Cannot find path"` against `Cannot find path '...' because it does not exist.` They were a substring match in Pester 4.

**3. Three tests asserted something the command does not do.** `Should -Invoke Test-DbaBuild -Exactly 0`, but [Get-SqlInstanceUpdate.ps1:109](../blob/development/private/functions/Get-SqlInstanceUpdate.ps1#L109) calls it whenever only a major version is given, which is exactly what those tests do. Until Pester 6 an unmatched `-ParameterFilter` fell through to the real command and the call was not counted, so the wrong expectation was invisible.

## Result

**69 tests, 68 pass**, 1 is the still-skipped integration `Describe`. Runtime ~35s.

The recovered tests are not passing vacuously: mutating a single expectation inside the `-ForEach` block fails exactly 44 of them.

Also fixed while in there: the "testing proper Authorization" Context made a real WinRM call to a host literally named `mock` (the connectivity probe at [Update-DbaInstance.ps1:523](../blob/development/public/Update-DbaInstance.ps1#L523) was unmocked, two tests took ~3s each and passed only because the connection failed); "fails when repository is not available" used a relative path so its message depended on the runner's working directory; and a dead `$resolvedComputers` assignment, an empty `AfterAll` and two "revert default mock" lines are gone.

The integration `Describe` is deliberately untouched. It is still `-Skip`, and since it mocks away reboot, elevation and update discovery it is a `-WhatIf` smoke test rather than an integration test - that deserves its own decision.

## The guide change

The guides banned `-ForEach` outright. That ban came out of the Pester 4 to 5 migration, but it removed the only correct way to write data-driven tests while leaving the wrong way fully compliant - a `foreach` loop around `It` passes every style rule, looks right in the report, and asserts nothing.

The missing piece was `BeforeDiscovery`. With the cases built there, `-ForEach` is safe and the discovery-time code stops being loose as well. `tests/CLAUDE.md` now has a "Discovery Time vs Run Time" section explaining Pester's two passes and a "Data-Driven Tests" section with the rules:

- build the cases in `BeforeDiscovery`, never `BeforeAll`
- use an array of hashtables, so keys become variables and `<Key>` expands in the name
- every value the body reads must be a key - a `BeforeDiscovery` variable is not visible in the body
- prefer separate `It` blocks when there are only two or three cases

`bin/prompts/pester.md`, `.github/prompts/migration.md` and `.github/copilot-instructions.md` carry the same rules and are updated to match.

The same commit documents the **existing** exception to the no-loose-code rule - a value that a `-Skip:` needs may be computed in the `Describe` block - in the three guides that did not have it yet. `tests/CLAUDE.md` already did. It also notes that `BeforeDiscovery` is the tidier home for that computation whenever the `-Skip:` is not on the `Describe` itself.

## Verification

- `Invoke-DbatoolsTest.ps1 Update-DbaInstance -Tag UnitTests`: 69 total, 68 passed, 0 failed.
- Mutation check on the `-ForEach` assertions: 44 failures, one per case.
- `TestTestfiles.Tests.ps1 -ExcludeTagFilter Goal`: 0 failures over 723 files.
- `TestTestLayout.ps1`: the same 82 pre-existing files as before the change.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -25,7 +25,7 @@
 
 3. **Pester v5.7.1** - Required for running tests
    - Install: `Install-Module Pester -RequiredVersion 5.7.1`
-   - Tests MUST use Pester v5 syntax (no `-ForEach` parameter, strict scoping)
+   - Tests MUST use Pester v5 syntax (strict scoping, `-ForEach` cases built in `BeforeDiscovery`)
 
 ## PR Summary Guidelines
 
@@ -334,8 +334,9 @@ param(
 - ALL setup code in `BeforeAll` blocks
 - ALL cleanup code in `AfterAll` blocks
 - ALL assertions in `It` blocks
-- NO loose code in `Describe` or `Context` blocks
-- NEVER use `-ForEach` parameter
+- ALL discovery-time code in `BeforeDiscovery` blocks
+- NO loose code in `Describe` or `Context` blocks, except computing a value a `-Skip:` needs
+- `-ForEach` ONLY with cases built in `BeforeDiscovery`, never a `foreach` loop around `It`
 
 ### When to Add/Update Tests
 
--- a/.github/prompts/migration.md
+++ b/.github/prompts/migration.md
@@ -25,8 +25,11 @@ param(
 - **All setup code** must be in `BeforeAll` or `BeforeEach` blocks
 - **All cleanup code** must be in `AfterAll` or `AfterEach` blocks
 - **All test assertions** must be in `It` blocks
-- **No loose code** allowed in `Describe` or `Context` blocks
-- **Never use `-ForEach` parameter** on any test blocks
+- **All discovery-time code** must be in a `BeforeDiscovery` block
+- **No loose code** allowed in `Describe` or `Context` blocks, with one exception: computing a value that a `-Skip:` needs
+- **Only use `-ForEach`** with cases built in `BeforeDiscovery`
+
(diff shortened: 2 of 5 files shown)


potatoqualitee:
nice update, ty. will merge onc eit passes.

andreasjordan:
My new Claude subscription is very helpfull inside of my live lab. I'm loving it!

potatoqualitee:
@andreasjordan i love it too! I have a huge lab and it just runs in it, running tests against real instances 24x7. so much fun 🥳

PR #10484, opened 2026-08-01
Tests - Recover the tests that a foreach around It never created

andreasjordan:
Three test files built their `It` blocks with a `foreach` loop. The loop runs while Pester discovers the tests and the bodies run afterwards, so the loop variables are gone by then. In all three cases the collection the loop walked was built in a `BeforeAll` and was still empty at discovery, so **not one of these tests was ever created**. Nothing failed - the tests simply did not exist, and the report looked healthy the whole time.

| File | tests before | tests after |
|---|---|---|
| Get-DbaBuild | 26 | 92 |
| Set-DbaAgentSchedule | 10 | 17 |
| Invoke-DbaDbDbccCheckConstraint | 2 | 11 |

**Get-DbaBuild** lost its entire "Properties Check for major releases" section, which walked `$OrderedKeys`. That variable was only ever built in the `BeforeAll` of a sibling Context - the `$OrderedKeys = $OrderedKeys` sitting in this one was an attempt to import it and could never work. The grouping moved into `BeforeDiscovery` and the per-release Contexts now come from `-ForEach`. All 11 major releases are covered again.

**Invoke-DbaDbDbccCheckConstraint** lost the per-property checks in "Validate standard output". The property list moved into `BeforeDiscovery` and the `It` uses `-ForEach`.

**Set-DbaAgentSchedule** lost seven assertions across five Contexts. `-ForEach` cannot help here: the results only exist once the setup has talked to the instance, so there is nothing to hand to `-ForEach` at discovery. The loops moved inside the `It` blocks instead, with the schedule name in a `-Because` so a failure still identifies which schedule it was.

## Verification

- All three files pass in full against the lab.
- The recovered assertions are not passing vacuously: mutating two of the Set-DbaAgentSchedule expectations fails exactly those two tests. The pre-existing "Should have Results" in each Context keeps an empty result set from making the internal loops trivially true.
- No change to the layout or structural checks beyond these three files.

Found by a new AST check in the sibling testing repo that flags any loop containing an `It`. Update-DbaInstance had the same defect and is fixed separately in #10482.

### DIFF
--- a/tests/Get-DbaBuild.Tests.ps1
+++ b/tests/Get-DbaBuild.Tests.ps1
@@ -174,45 +174,50 @@ Describe $CommandName -Tag UnitTests {
     }
     # These are groups by major release (aka "Name")
     Context "Properties Check for major releases" {
-        BeforeAll {
-            $OrderedKeys = $OrderedKeys
-            $Groups = $Groups
+        # The groups have to be built here and not in a BeforeAll. -ForEach is read while Pester
+        # discovers the tests, and $OrderedKeys and $Groups were only ever built in the BeforeAll of
+        # a sibling Context, so the foreach that used to stand here ran over nothing and none of the
+        # tests below existed. The "$OrderedKeys = $OrderedKeys" that used to be here was an attempt
+        # to import them into this scope and could never work.
+        # Everything an It block reads has to be a key of the case, so the version list travels with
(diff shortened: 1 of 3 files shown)


potatoqualitee:
I believe you 😅

potatoqualitee:
(that was a tough code/comprehension read for some reason)

PR #10486, opened 2026-08-01
Tests - Make the assertions that could never fail actually assert

andreasjordan:
Twenty-five statements across six test files computed a comparison inside an `It` and threw the result away - lines like `$results -match "CREATE TABLE"` with no `Should`. They read like assertions, they ran, and they could not fail. Three more files handed `-ForEach` a case list built in a `BeforeAll`, which is still empty at discovery, so those tests were never created at all.

Piping the comparisons to `Should` turned four of them red. **Every one was a real defect that had been invisible.**

## What was actually broken

**Export-DbaDbRole never created the role.** The setup ran `ALTER ROLE ... ADD MEMBER` with no `CREATE ROLE` first, so it threw, and a `catch { # Ignore setup errors for now }` swallowed it. The database had no role, no grants, no membership, and `Export-DbaDbRole -Passthru` returned `$null`. Six assertions were checking `$null` against expected SQL and reporting success. Added the missing `CREATE ROLE` and removed the catch, so a broken setup now fails loudly.

**Export-DbaDbRole also expected membership it never asked for.** The default output cannot contain `ALTER ROLE ... ADD MEMBER` - that needs `-IncludeRoleMember`. Split into two tests: the default output must *not* contain it, the `-IncludeRoleMember` output must. That switch had no coverage at all before.

**Export-DbaScript searched for the name of a command.** The assertion was `$results -match "(Get-DbatoolsConfigValue -FullName 'Formatting.BatchSeparator')"` - the subexpression marker was missing, so it looked for that literal text in the SQL script rather than for the separator the command emits.

**Get-DbaSpConfigure compared against the wrong set.** It checked its result count against `sp_configure`, which lists only the 25 basic options unless *show advanced options* is on. The command returns all 86. Now compared against `sys.configurations`, which lists them all without changing the configuration of the instance. The `sp_configure` result is still used by the test that reads its `config_value` and `run_value` columns.

## Two mechanical points

`Export-DbaScript` and `Export-DbaDbRole` return the script as a **string array**, so the text is joined before matching - piping the array asserts against a single element. And the bracketed SQL identifiers (`[dbo]`, `[$dbRole]`) are regex character classes, so they are matched with `[regex]::Escape`.

## Tests recovered

| File | before | after |
|---|---|---|
| Get-DbaDbccSessionBuffer | 3 | 18 |
| Set-DbaDbIdentity | 2 | 14 |
| Get-SqlDefaultSpConfigure | 1 | 15 |
| Export-DbaDbRole | 11 | 12 |

The three `-ForEach` case lists moved into `BeforeDiscovery`. In `Get-SqlDefaultSpConfigure` the list was built by calling the function under test, which cannot happen at discovery, so the cases are plain data now and the function is called inside the `It`.

## Verification

All nine files pass against the lab. The two AST checks that found this - discarded comparison inside an `It`, and `-ForEach` cases not from `BeforeDiscovery` - are now hard checks in the testing repo, and the whole tree passes them with 0 failures.

One thing left alone deliberately: `Get-SqlDefaultSpConfigure` tests SQL versions 8 to 14, while its version map also lists 15 (2019) and 16 (2022). That gap predates this change and I did not widen the scope, but it is probably worth closing.

### DIFF
--- a/tests/Export-DbaDbRole.Tests.ps1
+++ b/tests/Export-DbaDbRole.Tests.ps1
@@ -49,19 +49,20 @@ Describe $CommandName -Tag IntegrationTests {
         $user1 = "dbatoolsci_exportdbadbrole_user1$random"
         $dbRole = "dbatoolsci_SpExecute$random"
 
-        try {
-            $server = Connect-DbaInstance -SqlInstance $TestConfig.InstanceSingle
-            $null = $server.Query("CREATE DATABASE [$dbname1]")
-            $null = $server.Query("CREATE LOGIN [$login1] WITH PASSWORD = 'GoodPass1234!'")
-            $server.Databases[$dbname1].ExecuteNonQuery("CREATE USER [$user1] FOR LOGIN [$login1]")
+        $server = Connect-DbaInstance -SqlInstance $TestConfig.InstanceSingle
+        $null = $server.Query("CREATE DATABASE [$dbname1]")
+        $null = $server.Query("CREATE LOGIN [$login1] WITH PASSWORD = 'GoodPass1234!'")
+        $server.Databases[$dbname1].ExecuteNonQuery("CREATE USER [$user1] FOR LOGIN [$login1]")
 
-            $server.Databases[$dbname1].ExecuteNonQuery("ALTER ROLE [$dbRole] ADD MEMBER [$user1]")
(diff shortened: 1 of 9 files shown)


potatoqualitee:
lgtm

PR #10487, opened 2026-08-01
Tests - Align the param block header so the layout check shows the real deviations

andreasjordan:
Formatting only, plus two file renames. No assertion, no test logic and no behaviour changes.

## Why

`TestTestLayout.ps1` reported **82** test files as not having the expected layout. **75 of them failed on the very first step**, the header and param block, and because the script stops at the first failure per file, none of the later checks ever ran for those files. The deviations that actually matter were hidden behind formatting noise.

## What changed

**73 files** wrote `$ModuleName = ` with a single space instead of aligning the equals sign with `$CommandName` on the line below, which is the form the mandatory header in `tests/CLAUDE.md` shows. One space added. The diff is 73 identical single-line changes and nothing else - I checked that every changed line in the whole PR is one of exactly two strings.

**Two files** failed because their file name and their command disagreed on capitalisation, and the check derives the expected `$CommandName` from the file name. In both cases the **file name** was the wrong one, not the `$CommandName`:

| Command, as defined | Test file was | Renamed to |
|---|---|---|
| `public/Get-DbaUptime.ps1`, and the manifest | `Get-DbaUpTime.Tests.ps1` | `Get-DbaUptime.Tests.ps1` |
| `private/functions/Convert-DbaLSN.ps1` | `Convert-DbaLsn.Tests.ps1` | `Convert-DbaLSN.Tests.ps1` |

Nothing in the repo referenced either file name, so this is safe on a case-sensitive filesystem too.

## Result

The layout check goes from **82 failures to 10**, and those 10 are real problems that were previously masked:

- 7 files whose expected-parameter list is not in the standard shape - Find-DbaAgentJob, Get-DbaInstanceList, Import-DbaParquet, Install-DbaParquet and three more
- 3 files whose unit-test Describe block is not - Remove-DbaReplArticle, Remove-DbaReplPublication and one more

Left for a follow-up so this stays a pure formatting pass.

## Verification

- Every changed line in the diff is one of two strings. No line endings, no trailing newlines, nothing else moved - I redid the pass once after spotting that my first attempt had added a trailing newline to 54 files.
- The AST structural check reports no new failures. The 9 it does report are the Phase 2a fixes in #10486, which is not in `development` yet.

## Note on the open PR

#10486 touches three of these files (Export-DbaScript, Save-DbaKbUpdate, Set-DbaDbIdentity). The changes are on different lines - line 3 here, test bodies there - so they should merge cleanly in either order. Ping me if not and I will rebase.


potatoqualitee:
this one is tough because it's not a hashtable, it' just two stacked variables :/

andreasjordan:
I agree, we should probably not merge this one.

andreasjordan:
I have not reviewed it before Claude created the pr.

andreasjordan:
Closing at Andreas's request - this change is not wanted. The branch is deleted too; the commit is 08289eb8c if it is ever needed again.

PR #10488, opened 2026-08-01
Tests - Stop Get-DbaBuild from downloading the build reference index during a unit test

andreasjordan:
Two mocks that never applied. One of them was letting a unit test reach the internet.

## Get-DbaBuild was really running the update

`Get-DbaBuild.Tests.ps1` declared a dummy `Get-DbaBuildReferenceIndexOnline` and mocked that:

```powershell
function Get-DbaBuildReferenceIndexOnline { }
Mock Get-DbaBuildReferenceIndexOnline -MockWith { }
Get-DbaBuild -Update -WarningVariable warnings 3>$null
```

The real `Get-DbaBuildReferenceIndexOnline` is a **nested function declared inside `Update-DbaBuildReference`**. It exists only while that command runs, so it cannot be mocked at all - the mock bound to the dummy in the test scope and the module never saw it. The test ran the real update every time: a live download of the build reference index, which also rewrites `bin\dbatools-buildref-index.json` in the checkout.

`Update-DbaBuildReference` is what `Get-DbaBuild` actually calls and it is a real command, so it is the boundary that can be intercepted. I verified both halves with a probe:

- with the old mock in place, `Update-DbaBuildReference` is still invoked exactly once
- with `Mock Update-DbaBuildReference -ModuleName dbatools`, it is intercepted and the test still passes

Added a `Should -Invoke` so the interception is asserted rather than assumed.

## Stop-Function mocked into the wrong scope

`Mock -CommandName "Write-Warning"` and `Mock -CommandName "Write-Error"` with no `-ModuleName`. `Stop-Function` runs in the module, so neither mock applied. Both scoped to `dbatools`. 34 tests pass before and after - the mocks were inert either way, this just makes them do what they say.

## What I did not change, and why

The same sweep flagged **13 `Should -Throw` assertions across 7 files** that expect a message with no wildcard. I checked all of them before touching anything, and **every one that can be run passes** - the expected message really is the whole message, so the assertion is correct and an exact expectation is the stronger one.

So that check was wrong, not the tests. I removed it from the checker rather than "fixing" 7 files that were already right. Nothing in the syntax distinguishes a complete expected message from a fragment that can never match; only running the test tells them apart. The comment left in its place records that, so nobody re-adds it.

The 5 in `Set-DbaLogin.Tests.ps1` are the exception - they sit inside a `Context -Skip "???"` marked `# TODO: Fix later`, so they have never run and I cannot say whether they are right. That Context is part of the bare-`-Skip` triage coming next.

## Verification

`Get-DbaBuild` 92/92 and `Stop-Function` 34/34 against the lab. The `Mock` check in the testing repo now reports zero findings across all 733 files.

### DIFF
--- a/tests/Get-DbaBuild.Tests.ps1
+++ b/tests/Get-DbaBuild.Tests.ps1
@@ -146,10 +146,18 @@ Describe $CommandName -Tag UnitTests {
     }
     Context "Passing just -Update works, see #6823" {
         It "works with -Update" {
-            function Get-DbaBuildReferenceIndexOnline { }
-            Mock Get-DbaBuildReferenceIndexOnline -MockWith { }
+            # This used to declare a dummy Get-DbaBuildReferenceIndexOnline and mock that. The real
+            # one is a nested function inside Update-DbaBuildReference, so it cannot be mocked at
+            # all, and the mock here bound to the dummy instead. The test therefore ran the real
+            # update: a live download of the build reference index, which also rewrites
+            # bin\dbatools-buildref-index.json. Update-DbaBuildReference is what Get-DbaBuild calls
+            # and it is a real command, so it is the boundary that can actually be intercepted.
+            Mock Update-DbaBuildReference -ModuleName dbatools -MockWith { }
+
             Get-DbaBuild -Update -WarningVariable warnings 3>$null
+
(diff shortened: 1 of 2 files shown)


potatoqualitee:
smart, thank you

PR #10485, opened 2026-08-01
Docs - Make tests/CLAUDE.md the single source for the test standards

andreasjordan:
Docs only. No code or test changes.

## The problem

The test rules live in four documents that have drifted apart, and the two entry points send readers to different copies:

| Document | Who points at it | State |
|---|---|---|
| `tests/CLAUDE.md` | **nobody** | The one that has actually been maintained |
| `.github/prompts/style.md` | `CLAUDE.md` | De-branded copy of the bin one, 15 "MODULE" artifacts |
| `bin/prompts/style.md` | `copilot-instructions.md` | Older copy |
| `bin/prompts/pester.md` | `copilot-instructions.md` | Older, shorter copy of tests/CLAUDE.md |

`tests/CLAUDE.md` is the only copy with the instance selection table, the note that only `InstanceSingle` and `InstanceMulti1/2` exist on GitHub Actions, the `InstanceRestart` shared-state warning, and the `-Skip:` exception to the no-loose-code rule. Nothing pointed at it, so anyone following `CLAUDE.md` or the Copilot instructions was reading an older, partly wrong set of rules.

## What this does

Makes `tests/CLAUDE.md` the single source and reduces the other three to stubs that point at it. `.github/prompts/migration.md` keeps its content, since converting a Pester v4 file is a genuinely separate topic, but now says up front that tests/CLAUDE.md wins where they overlap.

**Nothing is lost.** Everything that existed only in a stubbed file was moved into `tests/CLAUDE.md` first:

| Moved | From |
|---|---|
| Comment preservation | `.github/prompts/style.md` |
| Array formatting | `.github/prompts/style.md` |
| Testing for warnings (`$WarnVar`) | `.github/prompts/style.md` |
| Test management guidelines - when to add tests, what makes a good test, balance | `.github/prompts/style.md` |
| Here-strings for multi-line strings | `bin/prompts/pester.md` |

I verified each of those nine headings is present in the consolidated file, and that every relative link in the stubs resolves.

## Three defects fixed on the way

1. **`.github/prompts/style.md` contradicted itself** - single quotes listed under **Must Not Use**, while its own canonical example wrote `$PSDefaultParameterValues['*-Dba*:EnableException'] = $true`. A find-and-replace had also turned "dbatools" into "MODULE" in 15 places, including the title and `MODULEci_AGCert`.
[...]

PR #10491, opened 2026-08-01
Tests - Standardise the last four parameter validation tests

andreasjordan:
Four parameter validation tests that deviate from the standard shape, found while reviewing the test layout. **Only four of the ten deviations the layout check reported were defects.** The other six were legitimate patterns the check did not allow for, so I relaxed the check rather than reshape working files.

## The four that were wrong

**Import-DbaParquet and Install-DbaParquet asserted in one direction only.** They built the expected list without `$TestConfig.CommonParameters` and finished with

```powershell
($expectedParameters | Where-Object { $PSItem -notin $hasParameters }) | Should -BeNullOrEmpty
```

That proves the expected parameters exist. A parameter **added** to the command by mistake would pass. Both now use the standard `Compare-Object` form, which is exact in both directions. Their lists already match the commands, so nothing else had to change.

**Test-DbaLsnChain wrapped its whole Describe in `InModuleScope` and wrote out the twelve common parameters by hand.** Reading `$TestConfig` inside `InModuleScope` is the thing that passes locally and fails in CI, and the hand-written list was presumably the workaround for it.

It now uses the same plain form as every other parameter validation test:

```powershell
$hasParameters = (Get-Command $CommandName).Parameters.Values.Name | Where-Object { $PSItem -notin ("WhatIf", "Confirm") }
$expectedParameters = $TestConfig.CommonParameters
```

There is no need to reach into the module for a private command. A checkout has no `dbatools.dat`, so the psm1 takes the `Export-ModuleMember -Function *` branch and exports **all 870 functions** against the 717 in the manifest - private ones included. Both the lab harness and `tests\appveyor.pester.ps1` import the psm1 from the checkout, so that holds in CI too. **Thirteen of the sixteen tests for private functions already call plain `Get-Command`** and pass.

The part that matters is unchanged: the Context stays outside `InModuleScope`, so `$TestConfig` is read from a scope where it is visible in CI.

**Get-DbaInstanceList keeps its `@()` wrapper, and now says why.** This one is worth knowing generally:

> `$TestConfig.CommonParameters` is a **`HashSet[string]`**, and `Compare-Object` treats a HashSet as a single object. Every test file gets away with the bare property only because the `+= @(...)` on the next line converts it to an array first. `Get-DbaInstanceList` has no parameters of its own, so there is no `+=`.

I removed the `@()` to make the file match the standard shape, and the test failed with all twelve common parameters reported as unexpected. That is how I found it. Restored, with the reason in a comment.

This probably belongs in `tests/CLAUDE.md` as well, now that #10485 has made it the single source. Say the word and I will add it here or in a follow-up.

## The six that were fine

Left alone; the check in the testing repo now allows them:

- **Four files put a second `It` in the Parameter validation Context** (Find-DbaAgentJob, Set-DbaAgentSchedule, Sync-DbaLoginPassword, Test-DbaKerberos) - testing a wildcard attribute, a mandatory parameter, a parameter set, an alias. Exactly where those belong. The check required the Context to end after the first test.
- **Two files put a comment between the Describe and the Context** (Remove-DbaReplArticle, Remove-DbaReplPublication) explaining the `-Skip:` on the Describe - which is what the guide asks for.

## Three more tests for private functions, cleaned up the same way

- **Invoke-ManagedComputerCommand** asked the module for the command with `& (Get-Module dbatools) { Get-Command -Name ... }`. Now plain `Get-Command $CommandName`. The comment keeps the part that does matter: this has to stay outside `InModuleScope`.
- **Test-PSRemoting and Get-ObjectNameParts** dot-sourced the private function file at the top of the test. Besides being unnecessary, it meant anything called from test scope ran the **dot-sourced copy rather than the one in the loaded module**. Identical today because both come from the same checkout, so nothing changed, but it would have hidden a difference if they ever diverged.

With those gone, no test file dot-sources at the top level and none reaches into the module for `Get-Command`, so the tolerances both layout checks carried for those two patterns are removed as well. All sixteen tests for private functions pass.

## Result

`TestTestLayout.ps1` no longer reports any of these ten files.

What it still reports is the 75 files whose `$ModuleName` is not aligned in the param block. That was #10487, which has been closed as unwanted, so those stay as they are. This PR is independent of it.

All four changed files pass, and so do all sixteen tests for private functions.

### DIFF
--- a/tests/Get-DbaInstanceList.Tests.ps1
+++ b/tests/Get-DbaInstanceList.Tests.ps1
@@ -9,6 +9,10 @@ Describe $CommandName -Tag UnitTests {
     Context "Parameter validation" {
         It "Should have the expected parameters" {
             $hasParameters = (Get-Command $CommandName).Parameters.Values.Name | Where-Object { $PSItem -notin ("WhatIf", "Confirm") }
+            # $TestConfig.CommonParameters is a HashSet, which Compare-Object treats as a single
+            # object. Every other test file gets away with the bare property because the "+= @()"
+            # on the next line converts it to an array first. This command has no parameters of
+            # its own, so there is no "+=" and the conversion has to happen here.
             $expectedParameters = @($TestConfig.CommonParameters)
             Compare-Object -ReferenceObject $expectedParameters -DifferenceObject $hasParameters | Should -BeNullOrEmpty
         }
--- a/tests/Get-ObjectNameParts.Tests.ps1
+++ b/tests/Get-ObjectNameParts.Tests.ps1
@@ -5,8 +5,6 @@ param(
     $PSDefaultParameterValues = $TestConfig.Defaults
 )
 
-. "$PSScriptRoot\..\private\functions\Get-ObjectNameParts.ps1"
-
 Describe $CommandName -Tag UnitTests {
     Context 'Parameter validation' {
         BeforeAll {
--- a/tests/Import-DbaParquet.Tests.ps1
+++ b/tests/Import-DbaParquet.Tests.ps1
@@ -9,7 +9,8 @@ Describe $CommandName -Tag UnitTests {
     Context "Parameter validation" {
         It "Should have the expected parameters" {
             $hasParameters = (Get-Command $CommandName).Parameters.Values.Name | Where-Object { $PSItem -notin ("WhatIf", "Confirm") }
-            $expectedParameters = @(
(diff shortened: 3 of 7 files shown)


andreasjordan:
This also needs more commits...

andreasjordan:
Looks good now.

potatoqualitee:
ty 🥇

PR #10492, opened 2026-08-01
Tests - Rename two test files to match the command they test

andreasjordan:
Two renames, no content changes at all - both files are 100% identical after the move.

## What is wrong

Both test files disagree with their command on capitalisation, and anything that derives a command name from a test file name gets it wrong as a result.

| Command, as defined | Test file was | Renamed to |
|---|---|---|
| `public/Get-DbaUptime.ps1`, and `dbatools.psd1` | `Get-DbaUpTime.Tests.ps1` | `Get-DbaUptime.Tests.ps1` |
| `private/functions/Convert-DbaLSN.ps1` | `Convert-DbaLsn.Tests.ps1` | `Convert-DbaLSN.Tests.ps1` |

In both cases the `$CommandName` **inside** the file was already correct - `Get-DbaUptime` and `Convert-DbaLSN` respectively. Only the file names were wrong, so this PR changes nothing but the names.

## Why it is safe

Nothing in the repository references either file name - I grepped every `.ps1`, `.psm1`, `.psd1`, `.yml`, `.json` and `.md`. Case-only renames are invisible on Windows but git records them properly, so this is also correct on a case-sensitive filesystem.

Both files pass after the rename: `Convert-DbaLSN` 4/4, `Get-DbaUptime` 6/6.

## Relationship to the closed #10487

These two renames were bundled into #10487, which was closed because the param block alignment in it was not wanted. The renames are unrelated to that alignment and were the only other thing in that PR, so they are re-proposed here on their own.

They are also the last two files the layout check in the testing repo reports. With #10491 merged as well, it reaches zero.


potatoqualitee:
party 🎈

Issue #8767, opened 2023-02-19
Dbatools configuration profiles


wsmelton:
### Summarize Functionality

Right now, we default a few configurations that are meant to be for pure interaction and not automation (non-interactive) uses with the module. Big example is the TEPP configuration, by default `tabexpansion.disable` is set to false.

The average use of our module I think now-a-days is purely non-interactive. Those that use it for interactive purposes to check things here-and-there is less than those writing scripts and automating workflows against SQL Server.

I think one of two things can be done:
- We default all configurations to assume non-interactive use of the module, those that want to perform interactive task can enable features if they want but we cater to automation and non-interactive sessions by default.
- Create a parameter on `Set-DbatoolsConfig` to allow for two profiles: interactive and non-interactive.

### Is there a command that is similiar or close to what you are looking for?

Yes

### Technical Details

If we go the route of profiles...

Create a parameter that would be used like: `Set-DbatoolsConfig -Profile Interactive` or `Set-DbatoolsConfig -Profile NonInteractive`

The workflow for each parameter value would set the associated configurations to disable (NonInteractive) or enable (Interactive). 

We have some values that for example the configuration item is to disable it, so by default it is enabled (`Get-DbatoolsConfig -Module tabexpansion`). In the scenario of non-interactive, that profile would set `tabexpansion.disable` to true, and opposite the interactive profile setting this would be set to false.

mattcargile:
Though, I am in the more interactive usage camp lately. I do like this behavior. 

It would be neat to have some module sanctioned aliases potentially too in the `Interactive` profile? Or another feature? Considering it appears we have tried to create these types of aliases. The scope of this generation doesn't appear to function correctly though.

https://github.com/dataplat/dbatools/blob/0433f3260cc448b8ca932b8da4a4b777a656768a/dbatools.psm1#L276-L282

Just commenting on these, I would prefer to follow the usage from `Get-Verb`. So I would rather `idq` and `ccdi`.

andreasjordan:
@claude please give me some advice on this.

potatoqualitee:
I went through this against current `development`. Here is where I landed, plus a PR for the part I think is unambiguous.

[...]

PR #10494, opened 2026-08-01
TEPP - Start the asynchronous cache runspace on the first tab completion instead of at import

potatoqualitee:
## Problem

The `dbatools-teppasynccache` runspace started during module import, so every session paid for it whether or not anyone was there to press <kbd>Tab</kbd>.

That runspace is not cheap and it is not quiet. It imports a **second copy of dbatools** into a background runspace, then reconnects to every instance the session has touched, every three minutes, for thirty minutes after last use. A scheduled task or a build agent that touches an instance once keeps logging back into it afterwards. That shows up in login auditing and session counts, for a tab-completion cache that nothing in an unattended session will ever read.

## Mechanism

`private/scripts/updateTeppAsync.ps1` ended with `Start-DbaRunspace -Name "dbatools-teppasynccache"`, which ran unconditionally at import. The two config handlers in `private/configurations/settings/tabexpansion.ps1` also had `else { $x.Start() }` branches, so re-enabling TEPP started the runspace too.

## What changed

- `updateTeppAsync.ps1` registers the runspace and leaves it **stopped**
- `Register-DbaTeppArgumentCompleter.ps1` starts it from inside the completer scriptblock, so the first actual tab completion is what turns it on
- both config handlers no longer start it, which keeps *enabling TEPP* and *starting the runspace* as separate decisions

No session sniffing, no heuristics, and no change to the compiled `dbatools.library`.

## Opting back in

Changing a default takes away a behaviour someone may be relying on, so there is a setting to get it back:

```powershell
Set-DbatoolsConfig -FullName TabExpansion.Asynchronous.StartOnImport -Value $true -Register
```

With that set, `updateTeppAsync.ps1` calls `Start-DbaRunspace` right after registering, exactly as it used to do unconditionally, and the cache is warm from the moment the module finishes importing. Off by default.

It is read once during import, because that is the moment it decides — changing it in a live session takes effect at the next import.

- **Both opt-outs still work and still win.** `TabExpansion.Disable` and `TabExpansion.Disable.Asynchronous` stop the runspace, and a later tab completion does not restart it.
- **Switching it back on still works.** A runspace that ran, was stopped by an opt-out, and was re-enabled starts again on the next completion rather than being spent.
- **`DBATOOLS_DISABLE_TEPP` is untouched.**
- **`-SqlInstance` completion is unaffected.** That cache is filled synchronously by `Connect-DbaInstance`, not by this runspace.
- **The runspace body is otherwise unchanged.**
- **A session that imported dbatools with TEPP already switched off still gets no runspace at all.** `updateTeppAsync.ps1` has always skipped registration in that case, and on `development` the handler's `if ($starttepp)` guard was equally null, so re-enabling never worked there either. Unchanged here, but now stated in the comment rather than left to be discovered. Worth its own fix if you want it fixed — it means dedenting that whole file.

One removal worth calling out: the fifteen second start-up deferral is gone. It was there to avoid colliding with the foreground import, and a keypress cannot happen while that import is still running. Keeping it would only have left the *first* completion empty for fifteen seconds, which is a real interactive regression in exchange for nothing.

## Testing

New `Describe` in `tests/dbatools.Tests.ps1`. It reads the runspace state out of a **fresh PowerShell**, because the runspace starts at most once per process, so asserting inside the test host would depend on whether an earlier test file happened to start it. Completion is driven through `TabExpansion2`, the entry point the console itself calls, rather than by reaching into the module.

The probe sets `$global:dbatools_config` before importing, which dbatools invokes during import after reading persisted configuration and before the asynchronous cache script runs. That makes the results independent of whatever the machine has persisted, so a contributor who has switched TEPP off does not see these fail.

Eight cases walked as one sequence: registered on import, not started on import, started by the first completion, stopped by the opt-out, not restarted by a completion after the opt-out, not started by switching it back on, started again by the completion after that, and not started while `TabExpansion.Disable` is set. A ninth runs in its own process, because `StartOnImport` is an import-time decision, and asserts the runspace is `Running` straight after import when it is set.

Measured on both Windows PowerShell 5.1 and PowerShell 7, identical on each:

| | `development` | this PR |
|---|---|---|
| after import | Running | **Stopped** |
| after first completion | Running | Running |
| after opt-out | Stopped | Stopped |
| completion after opt-out | Stopped | Stopped |
| after switching back on | Running | **Stopped** |
| completion after that | Running | Running |

The two bold rows are the ones that fail on `development`, so they are the regression tests; the rest guard against the behaviour drifting back.

`Invoke-ManualPester -Path tests/dbatools.Tests.ps1` → 20 passed, 0 failed.

Not verified locally: end-to-end cache population against a live instance, since I do not have the lab. The runspace body is unchanged apart from the removed sleep.

Refs #8767


potatoqualitee:
This is a big change but in retrospect, one that is very much needed. I didn't remember SOOOOOOOO much overhead, gonna change course on this one.

potatoqualitee:
Ahhhhhhhhhhhh

PR #10496, opened 2026-08-01
Get-DbaDbFile - Read the file layout from master when a database cannot be opened

### LINKED ISSUE #7968: Change way how Get-DbaDbFile and Get-DbaDbFileMapping get data
### Summarize Functionality

When the database status is not `Normal`, let's query the system view `master.sys.master_files` instead of reporting an error.

```powershell
WARNING: [08:52:52][Get-DbaDbFile] Failure | Database '<databaseName>' cannot be opened. It is in the middle of a restore.
```

No

### Technical Details

_No response_

potatoqualitee:
Fixes #7968

## Problem

A database that cannot be opened - `RESTORING`, `RECOVERY_PENDING`, `SUSPECT` or `OFFLINE` - cannot be set as the query context. `Get-DbaDbFile` sets it as the context and reads `sys.database_files`, so it failed with:

```
Database 'X' cannot be opened. It is in the middle of a restore.
```

`Get-DbaDbFileMapping` checked `IsAccessible` and skipped the database silently.

That is exactly the moment the information is needed. @olegstrutinskii's scenario: a full restore runs at 04:00 with `-NoRecovery` and randomized file suffixes, and the differential at 08:00 needs `-FileMapping` to line the files up. The names are unreachable in between.

## Mechanism

`sys.database_files`, `sys.filegroups`, `sys.dm_io_virtual_file_stats` and `sys.dm_os_volume_stats` all live inside the database. `sys.master_files` lives in master and still carries the configured file layout for a database that cannot be opened.

## What changed

`public/Get-DbaDbFile.ps1`

- When `$db.IsAccessible` is false, the file layout comes from a single query against `sys.master_files` instead of the in-database query. On SQL Server 2000 the equivalent is `master.dbo.sysaltfiles`, decoded with the same status bits the existing `$sql2000` branch uses.
- The output shape is byte-identical - same properties, same order. The twelve properties that have no authoritative value outside the database are `$null`, not zero: `FileGroupName`, `UsedSpace`, `AvailableSpace`, `NumberOfDiskWrites`, `NumberOfDiskReads`, `ReadFromDisk`, `WrittenToDisk`, `VolumeFreeSpace`, `FileGroupType`, `FileGroupTypeDescription`, `FileGroupDefault`, `FileGroupReadOnly`.
- `NULLIF(data_space_id, 0)` reproduces the NULL `FileGroupDataSpaceId` the accessible path returns for log files.
- A row in `sys.master_files` is only visible to a login holding `CREATE DATABASE`, `ALTER ANY DATABASE` or `VIEW ANY DEFINITION`. Rows are filtered rather than refused, so zero rows is treated as a permission problem and reported as such instead of looking like a database that vanished.
- `-FileGroup` warns and skips for an inaccessible database, because filegroup names are not in master and silently returning nothing would be worse.
- Verbose messages identify the fallback.

`public/Get-DbaDbFileMapping.ps1`

- An inaccessible database now builds its mapping from `Get-DbaDbFile` rather than being skipped. No second implementation of the fallback.

Two defects found while working in this file and fixed here:

- The compatibility-level probe interpolated the database name into a string comparison. A name containing an apostrophe made it throw, and the `catch` silently degraded the command to the SQL Server 2000 query shape - so the caller got a valid-looking object with `NumberOfDiskReads`, `VolumeFreeSpace` and the filegroup columns missing, and no warning. It now looks the database up by `database_id`. Regression test included.
- The `xp_fixeddrives` here-string was the file's last single-quoted string. Its body contains no `$` and no backtick, so the expandable form is byte-identical at 341 characters; verified by comparing both forms directly.

## What this change does not touch, and why

The accessible-database path is unchanged, including the `xp_fixeddrives` volume-space branch.

Four sibling commands have the same "skip if inaccessible" behaviour. None of them can be fixed the same way, because what they read does not exist outside an open database:

- `public/Get-DbaDbFileGroup.ps1:122` gates on `$db.IsAccessible` and skips at `:145`. It reads `$db.Filegroups` and emits SMO `FileGroup` objects through `Select-DefaultView`. `sys.master_files` carries no filegroup name, type or status - only `data_space_id`. Filegroup names live in `sys.filegroups`, a database-scoped catalog view. Reproducing this from master would mean emitting a different object type, which changes the command's output contract.
- `public/Get-DbaDbSpace.ps1:187` warns and continues. Its query uses `FILEPROPERTY(name, 'SpaceUsed')`, which by definition resolves against the current database and returns NULL for anything else. There is no server-wide equivalent, so used space inside a file cannot be read without opening the database.
- `public/Get-DbaDbLogSpace.ps1:108` filters with `Where-Object IsAccessible`. It needs `sys.dm_db_log_space_usage`, a database-scoped DMV, or `DBCC SQLPERF(LOGSPACE)`. Log space consumption is runtime state of an open database; for a database mid-restore the value does not exist rather than being merely unreachable.
- `public/Get-DbaDbVirtualLogFile.ps1:117` filters the same way and runs `DBCC LOGINFO` at `:132`, which reads VLF headers through the database's log manager and requires the database to be open. The log of a restoring database is held by the restore.

Two output properties are deliberately left reporting what master says, and are documented in `.OUTPUTS`:

- `State` and `IsOffline` describe the file, not the database. `sys.master_files` reports the files of an OFFLINE database as ONLINE.
- `Size` in the fallback is the size recorded in master. For an accessible database it comes from `sys.dm_io_virtual_file_stats` where available, so the two differ for sparse files, snapshots and FILESTREAM containers.

## Testing

Run against a live SQL Server 2022 instance (16.0.1000.6): **27 passed, 0 failed**.

`tests/Get-DbaDbFile.Tests.ps1` - a new `Databases that cannot be opened` context covering a NORECOVERY restore and an offline database:

- files are returned for a database in NORECOVERY, with `State` = `RESTORING`
- the physical names are the ones the restore created, the logical names are the source's
- the offline database's output matches, property for property, what the same database returned while it was still online
- the property list and order are identical to an accessible database's
- all twelve unavailable properties are null, anchored on a non-null `LogicalName` so the assertion cannot pass on an empty result
- no warning is raised
- `-FileGroup` warns and returns nothing
- a whole-instance scan still returns the accessible databases and now includes the inaccessible ones

plus a `Database names that contain an apostrophe` context for the version-probe fix.

`tests/Get-DbaDbFileMapping.Tests.ps1` - the mapping built while the database is in NORECOVERY is asserted equal to the mapping SMO builds for the very same files once the database is recovered.

These are regression tests, not just coverage: run against the current `development` code, 14 of them fail.

### SQL Server 2000 coverage

The `$sqlMasterFiles2000` branch at `public/Get-DbaDbFile.ps1:246` is dispatched at `:291` only when `$server.VersionMajor -le 8`. That dispatch is not covered: it needs an instance reporting version 8, and CI provisions the `dbatools1`/`dbatools2` Linux containers (`private/testing/Get-TestConfig.ps1:64-66`), which SQL Server 2000 has no build for.

The query itself is covered. `master.dbo.sysaltfiles` still resolves on SQL Server 2022, so the test pulls the shipped query text out of the command's own AST, executes it against the database left in NORECOVERY, and compares all ten stable columns row for row against the `sys.master_files` query it stands in for. That covers the column names and the status-bit arithmetic, which is where this query realistically breaks. Row counts are asserted at 2 on both sides so the comparison cannot pass on empty sets.

What remains uncovered is therefore the version dispatch itself, not the SQL.

### DIFF
--- a/public/Get-DbaDbFile.ps1
+++ b/public/Get-DbaDbFile.ps1
@@ -386,7 +386,7 @@ WHERE df.dbid = {0}
                     } else {
                         # to get drive free space for each drive that a database has files on
                         # when database compatibility lower than 110. Lets do this with query2
-                        $query2 = @'
+                        $query2 = @"
 -- to get drive free space for each drive that a database has files on
 DECLARE @FixedDrives TABLE(Drive CHAR(1), MB_Free BIGINT);
 INSERT @FixedDrives EXEC sys.xp_fixeddrives;
@@ -395,7 +395,7 @@ SELECT DISTINCT fd.MB_Free, LEFT(df.physical_name, 1) AS [Drive]
 FROM @FixedDrives AS fd
 INNER JOIN sys.database_files AS df
 ON fd.Drive = LEFT(df.physical_name, 1);
-'@
+"@
                         # if the server has one drive xp_fixeddrives returns one row, but we still need $disks to be an array.
                         if ($server.VersionMajor -gt 8) {
                             $disks = @($server.Query($query2, $db.Name))
--- a/public/Get-DbaDbFile.ps1
+++ b/public/Get-DbaDbFile.ps1
@@ -6,6 +6,8 @@ function Get-DbaDbFile {
     .DESCRIPTION
         Retrieves detailed information about database files (data and log files) from SQL Server instances using direct T-SQL queries for optimal performance. This function provides comprehensive file metadata including current size, used space, growth settings, I/O statistics, and volume free space information that DBAs need for capacity planning, performance analysis, and storage management. Unlike SMO-based approaches, this command avoids costly enumeration operations and provides faster results when analyzing file configurations across multiple databases.
 
+        Databases that cannot be opened - RESTORING, RECOVERY_PENDING, SUSPECT or OFFLINE - are read from sys.master_files in master instead of from inside the database, so a database left in NORECOVERY between a full and a differential restore still reports its logical and physical file names. Only the configured file layout is authoritative there: used space, available space, filegroup metadata, the I/O counters and volume free space are returned as null for those databases, and reading them needs CREATE DATABASE, ALTER ANY DATABASE or VIEW ANY DEFINITION permission on the instance.
+
     .PARAMETER SqlInstance
         The target SQL Server instance or instances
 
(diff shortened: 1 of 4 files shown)


potatoqualitee:
i like this, thanks for the suggestion @olegstrutinskii

Issue #8477, opened 2022-08-09
New command to get unused logins


ddraillantclark:
### Summarize Command's Functionality

This command would return logins that are not associated to a user, and which do not have permissions or ownership of any server-level object such as an endpoint.

The problem it would solve is finding extraneous logins which may for example have only been associated to a user for a database that no longer exists, so that the unused logins can be deleted to harden security.

Note: I am a DBA with a background in development, but have never contributed before, and if this gets the green light I propose to create the command.

### Is there a command that is similiar or close to what you are looking for?

No

### Technical Details

The concept would be based on this TSQL query I found on stackexchange, but implemented as PowerShell.

https://dba.stackexchange.com/questions/250844/how-to-find-unused-logins-in-sql-server

ddraillantclark:
I forgot to mention, I think that implementing this as a new command rather than adding it as a parameter in Get-DbaLogin would mirror what exists for users, as there is a separate Get-DbaDbOrphanUser command, which similar to the one I am proposing serves a distinct purpose as compared to Get-DbaDbUser.

potatoqualitee:
Thank you for the report, @ddraillantclark! @ClaudioESSilva what do you think?

ClaudioESSilva:
I think this would make sense @potatoqualitee.
Maybe a `Test` command. `Test-DbaLoginUnused` ? or something like that.

@ddraillantclark do you want to give it a try? 😃 

Also, I think the #8497 open by @berki7867 is a similar request. Do you confirm Berki? If yes maybe we can close the other issue and keep just this one.

potatoqualitee:
hmmm @wsmelton you're better with naming, what do you think?

wsmelton:
Name ideas:
- `Get-DbaOrphanLogin`
- `Get-DbaInactiveLogin` (my preference)
- `Get-DbaUnusedLogin`

I wouldn't necessarily use the exact columns returned by that query referenced but base it on the concept. As well determine whether our current commands already pull this data or if we need to really use T-SQL for it. Initial properties I'd recommend:
- ComputerName
- InstanceName
- SqlInstance
- Login
- Type
- CreateDate (dbadate)
- LastLoginDate (dbadate)
- IsAssignedRole (boolean)
- IsMappedToUser (boolean)

Naming of that last one could be better.

potatoqualitee:
[...]

PR #10498, opened 2026-08-02
Get-DbaUnusedLogin - Match the login name case-sensitively when reading last login time

potatoqualitee:
Closes the review round on `Get-DbaUnusedLogin`, the command requested in #8477.

### Context

The command itself landed on `development` in 52e19b13c, pushed directly rather than through a pull request, so it never got a review pass and the issue was never linked. This PR carries the one finding from the external review that the pushed version does not already contain.

### Problem

`LastLogin` is read by filtering the `sys.dm_exec_sessions` rows for the current login:

```powershell
$loginTime = $loginTimes | Where-Object login_name -eq $currentLogin.Name | ...
```

`-eq` is case-insensitive. An instance with a case-sensitive server collation can hold both `Bob` and `bob` as separate logins, so that filter returns a row for each. The two-element array is then handed to the `[DbaDateTime]` cast on the next line, which throws:

```
Cannot convert the "System.Object[]" value of type "System.Object[]" to type "Dataplat.Dbatools.Utility.DbaDateTime".
```

The command dies on an otherwise healthy instance because two of its logins differ only in case. Verified both halves locally: the cast does throw on a two-element array, and `-eq` matches 2 rows where `-ceq` matches 1.

### What changed

One line, `-eq` to `-ceq`, plus a comment saying why. The role-membership and SID lookups in the same command already use ordinal comparison, so this brings the third lookup in line with the other two rather than introducing a new convention.

### What deliberately did not change

- **The query itself.** It stays byte-identical to the one in `Get-DbaLogin`, including `login_name` rather than `original_login_name`, so `LastLogin` means the same thing in both commands. Attributing an `EXECUTE AS` session to the original login would be more precise in isolation but would make the two commands disagree.
- **The `-in` filters on `Login` and `ExcludeLogin`.** Those stay case-insensitive, which is how every other dbatools command treats a name the user typed.

### Testing

`Invoke-ManualPester -Path Get-DbaUnusedLogin -TestIntegration` against a real SQL 2022 instance: **19 passed, 0 failed**, unchanged from before the fix.

No test reproduces the case-sensitive condition. Login names follow the server collation, so demonstrating it needs an instance with a case-sensitive server collation; both lab instances are `SQL_Latin1_General_CP1_CI_AS` and cannot hold `Bob` and `bob` at once. Flagging that rather than adding a test that cannot fail for the old behaviour.

### Credit

@ddraillantclark raised #8477 and specified the behaviour; @wsmelton, @ClaudioESSilva and @potatoqualitee settled the name and the property set in that thread.

### DIFF
--- a/public/Get-DbaUnusedLogin.ps1
+++ b/public/Get-DbaUnusedLogin.ps1
@@ -267,7 +267,10 @@ function Get-DbaUnusedLogin {
                     continue
                 }
 
-                $loginTime = $loginTimes | Where-Object login_name -eq $currentLogin.Name | Select-Object -ExpandProperty login_time
+                # -ceq, not -eq: an instance with a case-sensitive server collation can hold both "Bob" and "bob",
+                # and a case-insensitive match returns a row for each. That hands an array to the [DbaDateTime]
+                # cast below, which throws, so the whole instance fails over two logins that differ only in case.
+                $loginTime = $loginTimes | Where-Object { $PSItem.login_name -ceq $currentLogin.Name } | Select-Object -ExpandProperty login_time
                 if ($loginTime) {
                     $lastLogin = [DbaDateTime]$loginTime
                 } else {


potatoqualitee:
Codex pushed the original branch to dev without a PR. Thankfully it passes tests.

https://github.com/dataplat/dbatools/commit/52e19b13c308eae4bab1691a4ce8b00d84cb7f6d

PR #10507, opened 2026-08-02
Invoke-DbaDbIndexRebuild - Add index and heap maintenance command

### LINKED ISSUE #3862: New function: Invoke-DbaDbIndexRebuild
### Summary of new feature

Create a new function Invoke-DbaDbIndexRebuild which can rebuild one or more indexes on a collection of tables. Should allow for passing collections of tables, databases, index names and require `-AllDatabases` if no databases are passed. Should also allow for various index rebuild options like online rebuilds, etc.

Allow for reorgs as well.

This may eventually be called from `Invoke-DbaDatabaseUpgrade` (optionally) as index rebuilds are needed for some upgrades.

Brute force method:

`get-dbadatabase -serverinstance YOURSERVER | foreach-object {$PSItem.Tables.Indexes|ForEach-Object {$PSItem.OnlineIndexOperation = $true; $PSItem.Rebuild()};}`

0.9.382

potatoqualitee:
Closes #3862.

dbatools has no index maintenance command today, so this adds one that works across collections of databases, tables, indexed views and indexes, exposing the ALTER INDEX option set through SMO.

### What it does

- **Three modes.** `Rebuild`, `Reorganize`, and `Auto`, where Auto picks the operation per index from measured fragmentation the way Ola Hallengren's IndexOptimize does: below `-ReorganizeThreshold` (default 5) it is left alone, between the thresholds it is reorganized, at or above `-RebuildThreshold` (default 30) it is rebuilt.
- **Heaps.** `-IncludeHeap` rebuilds heaps with `ALTER TABLE ... REBUILD`. This is the follow-up ask on the issue: Ola's scripts leave heaps alone entirely, so today there is no way to defragment one without hand-writing the T-SQL. A heap rebuild also rebuilds the table's nonclustered indexes, so those are not rebuilt a second time in the same run.
- **Explicit "all databases".** Reaching every database takes `-AllDatabases` (includes system databases) or `-AllUserDatabases`. An unqualified call warns instead of rebuilding an entire instance.
- **The full option set.** `-Online`, `-MaxDop`, `-FillFactor`, `-PadIndex`, `-SortInTempdb`, `-Resumable` / `-ResumableMaxDuration`, `-WaitAtLowPriority` / `-MaxDurationMinutes` / `-AbortAfterWait`, plus `-StatementTimeout` so a long rebuild is not cut off.
- **Filters.** `-MinimumFragmentation` and `-MinimumPageCount` apply in every mode, so an explicit Rebuild run can still skip indexes that are already healthy.
- **Pipeline.** Accepts databases from `Get-DbaDatabase`, tables from `Get-DbaDbTable`, and views from `Get-DbaDbView`.

### Measurement

Fragmentation is read once per database with a `LIMITED` scan of `sys.dm_db_index_physical_stats`, averaged across partitions weighted by page count, and restricted to `IN_ROW_DATA` so the zero-fragmentation LOB and row-overflow rows do not drag the number down. That single scan drives the Auto decision, both filters, and the reported before/after values.

Columnstore indexes are the exception. That DMV only sees their rowstore parts, so a clustered columnstore over half a million rows reports a handful of pages and a fragmentation figure that means nothing. Their `PageCount` and fragmentation come back null, and they are skipped with a warning under `Auto` or either filter rather than being decided on a bad number.

### Safety

Anything that cannot honour the requested operation is skipped with a warning rather than quietly downgraded. `-Online` in particular is what stands between the caller and a table lock, so an index that cannot rebuild online is skipped and said out loud instead of rebuilt offline. Objects that cannot be processed at all — memory-optimized tables, read-only databases, snapshots, disabled indexes under Reorganize — are skipped without failing the run, so one unsupported index does not stop maintenance across the rest of the instance. `SupportsShouldProcess` is per operation.

Requires SQL Server 2005 or later; `ALTER INDEX` and the DMV do not exist on 2000, and `-Resumable` / `-WaitAtLowPriority` carry their own version floors that are waived on Azure SQL Database.

### Tests

`tests/Invoke-DbaDbIndexRebuild.Tests.ps1` — 36 tests, all against a real instance with no mocks: parameter validation, the three modes, the Auto threshold bands, heaps (including the double-rebuild case), columnstore, both filters, database selection including `-AllDatabases` vs `-AllUserDatabases`, read-only databases, `-WhatIf`, all three pipeline shapes, and the option guards. A rebuild case was also added to `.github/scripts/gh-actions.ps1`.

Out of scope: wiring this into `Invoke-DbaDbUpgrade`.

### DIFF
--- a/public/Invoke-DbaDbIndexRebuild.ps1
+++ b/public/Invoke-DbaDbIndexRebuild.ps1
@@ -50,7 +50,9 @@ function Invoke-DbaDbIndexRebuild {
 
     .PARAMETER Mode
         Controls which operation is performed. Rebuild issues ALTER INDEX ... REBUILD, Reorganize issues ALTER INDEX ... REORGANIZE, and Auto decides per index from measured fragmentation. Defaults to Rebuild.
+        Reorganize is skipped with a warning for heaps, disabled indexes, and indexes with ALLOW_PAGE_LOCKS = OFF, none of which SQL Server will reorganize.
         Auto is the Ola Hallengren approach: below ReorganizeThreshold the index is left alone, between the two thresholds it is reorganized, and at or above RebuildThreshold it is rebuilt.
+        Reorganize takes none of the rebuild options, so under Reorganize they are listed in a warning and skipped rather than refused. Auto still refuses an invalid rebuild combination, because Auto can decide to rebuild.
 
     .PARAMETER ReorganizeThreshold
         Sets the fragmentation percentage at which Auto mode starts reorganizing an index. Defaults to 5.
@@ -70,7 +72,7 @@ function Invoke-DbaDbIndexRebuild {
 
     .PARAMETER Online
         Performs the rebuild online so the table stays available to other sessions for the duration.
-        Requires Enterprise or Developer edition. Indexes that cannot be rebuilt online are skipped with a warning rather than being rebuilt offline behind your back, columnstore included: only a clustered columnstore index on SQL Server 2019 or later can rebuild online.
+        Requires Enterprise or Developer edition. Indexes that cannot be rebuilt online are skipped with a warning rather than being rebuilt offline behind your back. Columnstore is the exception: which columnstore indexes rebuild online depends on the version and index type in ways SMO does not report, so the request goes to the engine and a rejection comes back as a failed row.
 
     .PARAMETER MaxDop
         Limits the number of processors used for the operation by adding MAXDOP, from 0 to 64.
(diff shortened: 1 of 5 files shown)


public/Invoke-DbaDbIndexRebuild.ps1
```diff
@@ -0,0 +1,917 @@
...
+        [string[]]$ExcludeDatabase,
+        [switch]$AllDatabases,
+        [switch]$AllUserDatabases,
+        [string[]]$Table,
+        [string[]]$Index,
+        [ValidateSet("Rebuild", "Reorganize", "Auto")]
+        [string]$Mode = "Rebuild",
+        [ValidateRange(0, 100)]
+        [double]$ReorganizeThreshold = 5,
+        [ValidateRange(0, 100)]
```
potatoqualitee, line 234: i really like this and the sane defaults

public/Invoke-DbaDbIndexRebuild.ps1
```diff
@@ -0,0 +1,917 @@
...
+    .PARAMETER AbortAfterWait
+        Specifies what happens when the low priority wait expires. None lets the operation keep waiting normally, Self aborts the rebuild, and Blockers kills the sessions holding the blocking locks. Defaults to None.
+        Anything other than None needs both WaitAtLowPriority and MaxDurationMinutes, because there is otherwise no wait to abort after.
+
+    .PARAMETER IncludeHeap
+        Includes heaps, meaning tables with no clustered index, rebuilt with ALTER TABLE ... REBUILD.
+        Heaps fragment through forwarded records and deleted rows but are ignored by most maintenance solutions, so they are opt-in here rather than silently rebuilt. A heap rebuild also rebuilds the table's nonclustered indexes, so the rest of that table is skipped for the run: their measured fragmentation is stale from that point on.
+
+    .PARAMETER StatementTimeout
+        Sets the command timeout in minutes for each operation, from 0 to 35791394. Defaults to 0 (infinite timeout).
```
potatoqualitee, line 118: great default here too

PR #10514, opened 2026-08-04
Copy-DbaDatabase - Fix empty -NewName destination sweep and piped multi-database failures

### LINKED ISSUE #10512: Copy-DbaDatabase more errors after #10389

Now that the errors from the Copy-Database ( see #10389 ) command should actually be fixed, new errors are popping up that are even stranger. I think your AI messed something up.

### Steps to Reproduce

I tried to copy three databases (CopyTest, CopyTest2 and CopyTest3), which resulted in the output shown below. Databases that were not selected at all are being affected here, especially on the destination server.

```powershell
$source_instance = "server12"
$target_instance = "server11"
$SharedPath = "\\server.example.com\mssqlbackup\Migrations"

Get-DbaDatabase -SqlInstance $source_instance |
	Out-GridView -Title "Select databases to copy" |
	Copy-DbaDatabase -Source $source_instance -Destination $target_instance -BackupRestore -SharedPath $SharedPath -WithReplace -NumberFiles 1
```
In the GridView, only the three databases CopyTest, CopyTest2, and CopyTest3 are selected.

```powershell
WARNING: [10:57:00][Set-DbaDbOwner] AdventureWorksDW2022 on server11\TEST is not in an updateable state and can not be altered. It will be skipped.
WARNING: [10:57:00][Set-DbaDbOwner] AdventureWorks2022 on server11\TEST is not in an updateable state and can not be altered. It will be skipped.
WARNING: [10:57:00][Set-DbaDbOwner] TEST_CONTAINED_msdb on server11\TEST is not in an updateable state and can not be altered. It will be skipped.
WARNING: [10:57:00][Set-DbaDbOwner] Database1 on server11\TEST is not in an updateable state and can not be altered. It will be skipped.
WARNING: [10:57:00][Set-DbaDbOwner] Database [Database2] is not accessible. Skipping.
WARNING: [10:57:00][Set-DbaDbOwner] Database3 on server11\TEST is not in an updateable state and can not be altered. It will be skipped.
WARNING: [10:57:00][Set-DbaDbOwner] Database [Database4] is not accessible. Skipping.
WARNING: [10:57:00][Set-DbaDbOwner] Database5 on server11\TEST is not in an updateable state and can not be altered. It will be skipped.
WARNING: [10:57:00][Set-DbaDbOwner] Database6 on server11\TEST is not in an updateable state and can not be altered. It will be skipped.
WARNING: [10:57:01][Set-DbaDbOwner] TEST_CONTAINED_master on server11\TEST is not in an updateable state and can not be altered. It will be skipped.
WARNING: [10:57:01][Copy-DbaDatabase] Failed to update DatabaseOwnershipChaining for False on  on server11\TEST.
WARNING: [10:57:01][Copy-DbaDatabase] Failed to update Trustworthy to False for  on server11\TEST.
WARNING: [10:57:01][Copy-DbaDatabase] Failed to update BrokerEnabled to False for  on server11\TEST.

WARNING: [10:57:03][Copy-DbaDatabase] Failure attempting to restore CopyTest2 to server11\TEST | Microsoft.Data.SqlClient.SqlError: Das Sicherungsmedium "\\server.example.com\mssqlbackup\Migrations\CopyTest2_202608041057.bak" kann nicht geöffnet werden. Betriebssystemfehler 2(Das System kann die angegebene Datei nicht finden.).
Type     Name      Status     Notes
----     ----      ------     -----
Database CopyTest  Successful Failed to apply the following properties: Ownership chaining, Trustworthy, Message broker, Read only
Database CopyTest2 Failed     Microsoft.Data.SqlClient.SqlError: Das Sicherungsmedium "\\server.example.com\mssqlbackup\Migrations\CopyTest2_202608041057.…
```

2.8.3

### Other details or mentions

_No response_

PowerShell Core (pwsh.exe)

### PowerShell Host Version

PSVersion                      7.6.3
GitCommitId                    7.6.3
OS                             Microsoft Windows 10.0.20348
PSRemotingProtocolVersion      2.4

Microsoft SQL Server 2022 (RTM-CU24-GDR) (KB5083252) - 16.0.4250.1 (X64)   Mar 13 2026 00:15:17   Copyright (C) 2022 Microsoft Corporation  Enterprise Edition: Core-based Licensing (64-bit) on Windows Server 2022 Standard 10.0 <X64> (Build 20348: ) 

### .NET Framework Version

PSChildName Version
----------- -------
Client      4.8.04161
Full        4.8.04161
Client      4.0.0.0

potatoqualitee:
Fixes #10512

## What was going wrong

Reproduced in a lab (sql2017 -> sql2019, UNC share) against the reported 2.8.3 symptoms. Four distinct defects compound into the reported behavior:

1. **A bound-but-empty `-NewName` silently sweeps the destination.** `[string]$NewName` binds `$null`/`""` without validation, `$destinationDbName` ends up empty, and every downstream `-Database ""` filter means "all databases". Since 2.8.3 replaced the post-restore `ALTER DATABASE` with `Set-DbaDbOwner`, that empty filter rewrites the owner of **every updateable database on the destination instance** - the reported owner warnings are just the databases it could not touch. The only guard (`$dbCount -gt 1`) never fires for piped input because `process{}` sees one database per record.

2. **One failed restore silently discards every remaining piped database.** The restore-failure path called `Stop-Function` without `-Continue`, which sets the module's persistent interrupt flag; all later pipeline records are dropped with no output at all - the "databases just vanish" flavor in the report.

3. **A missing single backup file passes verification.** `Test-DbaPath` returns a bare `[bool]` for a single scalar path on a single instance, but `Test-DbaBackupInformation` always assumes objects: `$path.FileExists -eq $false` on a bool is never true, so a missing single-file backup adds no verification error and the engine's OS error 2 becomes the first (confusing) signal.

4. **Path-check batch failures are reported as "file not found".** When the `xp_fileexist` batch throws, `Test-DbaPath` fabricated `FileExists = $false` for every path and demoted the real exception to Verbose - making transient access failures look like missing backups.

The remaining trigger - backups intermittently "not found" although they exist - reproduced as the destination host's **SMB client FileNotFoundCache** (default 5 s): a single early miss pins "not found" for the restore that follows (control 0/30 failures, cache primed 30/30, cache disabled 0/15).

## The fixes

- Reject a bound-but-empty `-NewName` up front with a clear message (it is what detonates the owner sweep). An empty `-Prefix` remains a harmless no-op and is deliberately not rejected.
- **Atomic pipeline rejection for `-NewName`:** with piped databases the total count is unknowable until the pipeline ends - by which time the first database would already have been copied under the new name. The command now rejects `-NewName` with pipeline input in `process{}` before any work happens (`$MyInvocation.ExpectingInput`), directing the caller to `-Database` + `-NewName` for a single-database rename. A parameter-bound `-InputObject` array still goes through the original count guard because its full count is known up front.
- Containment invariant: if a destination database name ever resolves to empty, that database emits a **Failed migration object** and is skipped before any destination-wide command can see an empty filter.
- Restore failures now emit the Failed migration object and use `-Continue`, so later piped databases still process.
- **Bounded visibility retry across the SMB negative-cache window:** after a fresh backup, the destination re-probes the backup file(s) for up to 8 seconds (the negative cache expires 5 seconds after creation), warning if they never become visible. This closes the intermittent OS-error-2 restores without touching machine-wide SMB configuration; `Set-SmbClientConfiguration -FileNotFoundCacheLifetime 0` on the destination host remains a workable environment-level mitigation.
- `Test-DbaBackupInformation` handles the scalar-bool contract of `Test-DbaPath` for single-file backups.
- `Test-DbaPath` surfaces batch execution failures as a warning stating existence could not be determined, instead of silently reporting not-found.

## Verification

- Lab repro matrix before/after: empty `-NewName` now stops loudly with zero copies and untouched owners; piped multi-database copies complete; piped `-NewName` is rejected with **zero databases created on the destination**; a missing backup mid-pipeline yields exactly one Failed row and the next database still copies; parameter-mode single-database rename still works.
- Tests: unit contexts for the empty-name rejection, the scalar-path verification failure, and the indeterminate path-check warning; integration context covering parameter-array and piped multi-database copies (asserting per-database `DestinationDatabase` and an unrelated destination owner preserved - the sentinel owner is a dedicated login, since the sweep sets owners to the source owner and sa-to-sa would go undetected), atomic piped-`-NewName` rejection, and failed-restore continuation. 12 unit + 5 integration tests green in the lab.

### DIFF
--- a/public/Copy-DbaDatabase.ps1
+++ b/public/Copy-DbaDatabase.ps1
@@ -130,6 +130,7 @@ function Copy-DbaDatabase {
         The database name and physical file names are updated to use the new name.
         Cannot be used with multiple databases, with pipeline input, or together with -Prefix parameter.
         To rename a database, specify it with -Database and provide -NewName.
+        An empty or whitespace value is treated as if -NewName was not specified, so scripts can pass a NewName variable unconditionally and leave it empty to copy under the original names.
 
     .PARAMETER Prefix
         Adds a prefix to all migrated database names and their physical file names.
@@ -304,9 +305,11 @@ function Copy-DbaDatabase {
 
         # An empty name slips through every downstream -Database filter, which then means
         # "all databases" and lets commands like Set-DbaDbOwner sweep the whole destination (#10512)
+        # Unbinding the parameter makes every later Test-Bound check treat it as not specified,
+        # so scripts can pass -NewName unconditionally and leave it empty to mean "no rename"
         if ((Test-Bound "NewName") -and [string]::IsNullOrWhiteSpace($NewName)) {
-            Stop-Function -Message "-NewName cannot be empty or whitespace. To copy databases under their original names, omit -NewName."
-            return
+            Write-Message -Level Verbose -Message "Empty -NewName supplied, copying databases under their original names."
+            $null = $PSBoundParameters.Remove("NewName")
         }
 
         if (-not $InputObject -and -not $Source) {
--- a/public/Copy-DbaDatabase.ps1
+++ b/public/Copy-DbaDatabase.ps1
@@ -301,6 +301,18 @@ function Copy-DbaDatabase {
     begin {
         $CopyOnly = -not $NoCopyOnly
 
+        # An empty name slips through every downstream -Database filter, which then means
+        # "all databases" and lets commands like Set-DbaDbOwner sweep the whole destination (#10512)
+        if ((Test-Bound "NewName") -and [string]::IsNullOrWhiteSpace($NewName)) {
+            Stop-Function -Message "-NewName cannot be empty or whitespace. To copy databases under their original names, omit -NewName."
+            return
+        }
+        if ((Test-Bound "Prefix") -and [string]::IsNullOrWhiteSpace($Prefix)) {
+            Stop-Function -Message "-Prefix cannot be empty or whitespace. To copy databases under their original names, omit -Prefix."
+            return
+        }
+        $pipelineDbCount = 0
+
         if (-not $InputObject -and -not $Source) {
             Stop-Function -Message "With no piped input a -Source must be specified."
             return
@@ -966,7 +978,12 @@ function Copy-DbaDatabase {
             Write-Message -Level Verbose -Message "Performing count."
             $dbCount = $databaseList.Count
 
-            if ((Test-Bound 'NewName') -and $dbCount -gt 1) {
+            # Track the count across process blocks: piped input arrives one database per
+            # invocation, so a per-invocation count never exceeds 1 and the guard below
+            # would otherwise never fire for pipelines (#10512)
+            $pipelineDbCount += $dbCount
+
+            if ((Test-Bound "NewName") -and $pipelineDbCount -gt 1) {
                 Stop-Function -Message "Cannot use NewName when copying multiple databases"
                 return
             }
@@ -1009,6 +1026,13 @@ function Copy-DbaDatabase {
                     if ($(Test-Bound "Prefix")) {
                         $destinationDbName = $prefix + $destinationDbName
                         Write-Message -Level Verbose -Message "Prefix supplied, copying $dbName as $destinationDbName"
+                    }
+
+                    # Belt and suspenders: an empty destination name must never reach the
+                    # restore or the post-restore owner/state/property commands, where an
+                    # empty -Database filter means "all databases" (#10512)
+                    if ([string]::IsNullOrWhiteSpace($destinationDbName)) {
+                        Stop-Function -Message "Destination database name for $dbName resolved to an empty string. Skipping this database to protect the destination." -Continue
                     }
 
                     $filestructure.databases[$dbName]['destinationDbName'] = $destinationDbName
@@ -1289,7 +1313,16 @@ function Copy-DbaDatabase {
                                 }
                             } catch {
                                 $msg = $_.Exception.InnerException.InnerException.InnerException.InnerException.Message
-                                Stop-Function -Message "Failure attempting to restore $dbName to $destinstance" -Exception $_.Exception.InnerException.InnerException.InnerException.InnerException
+                                if (-not $msg) {
+                                    $msg = $_.Exception.Message
+                                }
+                                $copyDatabaseStatus.Status = "Failed"
+                                $copyDatabaseStatus.Notes = $msg
+                                $copyDatabaseStatus | Select-DefaultView -Property DateTime, SourceServer, DestinationServer, Name, Type, Status, Notes -TypeName MigrationObject
+                                # -Continue so the interrupt flag is not set: that flag persists
+                                # across process blocks and used to silently discard every database
+                                # piped in after one failed restore (#10512)
+                                Stop-Function -Message "Failure attempting to restore $dbName to $destinstance" -Exception $_.Exception.InnerException.InnerException.InnerException.InnerException -Continue
                             }
                             $restoreResult = $restoreResultTmp.RestoreComplete
 
--- a/public/Copy-DbaDatabase.ps1
+++ b/public/Copy-DbaDatabase.ps1
@@ -128,7 +128,8 @@ function Copy-DbaDatabase {
     .PARAMETER NewName
         Renames the database during migration when copying a single database.
         The database name and physical file names are updated to use the new name.
-        Cannot be used with multiple databases or together with -Prefix parameter.
+        Cannot be used with multiple databases, with pipeline input, or together with -Prefix parameter.
+        To rename a database, specify it with -Database and provide -NewName.
 
     .PARAMETER Prefix
         Adds a prefix to all migrated database names and their physical file names.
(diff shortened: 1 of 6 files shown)


potatoqualitee:
Both reviews addressed in 03cace24c. Point by point:

**1. Piped-`NewName` guard fired only after the first database was already copied (both reviews, blocking).** Agreed - the cumulative counter was not atomic, and with `-WithReplace` the partial copy could overwrite a destination database before the rejection fired. The counter machinery is gone. The command now rejects `-NewName` + pipeline input at the top of `process{}` before any work happens, detected via `$MyInvocation.ExpectingInput`: in pipeline mode the total count is unknowable until the pipeline ends, so rejection (not buffering) is the only shape that preserves both atomicity and the streaming design. The message directs callers to `-Database` + `-NewName` for a single-database rename. A parameter-bound `-InputObject` array falls through to the original `$dbCount -gt 1` guard because its full count is known up front. The test is renamed to "Rejects NewName with piped databases before any copy happens" and now asserts `$results | Should -BeNullOrEmpty` plus zero databases created on the destination (neither the rename target nor either source name). Verified in the lab: the run completes in ~100 ms with the destination untouched.

**2. Empty `-Prefix` rejection was an unrelated backward-incompatible change (both reviews).** Removed, along with its unit test. An empty prefix is a harmless no-op (`$destinationDbName` stays the source name), and rejecting it would break existing `Prefix = ""` splats. Empty `-NewName` is still rejected because it is the value that detonates the owner sweep.

**3. Containment invariant emitted no Failed result (review 1 point 3, review 2 "smaller").** Fixed - `$copyDatabaseStatus` is now constructed before the invariant, which emits a Failed MigrationObject (`Notes = "Destination database name resolved to an empty string"`) before `Stop-Function -Continue`, so a skipped database is visible in the output instead of silently vanishing.

**4. Scope of "Fixes #10512" vs the SMB trigger (review 1 point 4).** Took the bounded-retry option the review favored: after a fresh backup (skipped for `-UseLastBackup` and URL paths), the destination re-probes the backup file(s) for up to 8 seconds at 1-second intervals. The negative-cache entry expires 5 seconds after creation regardless of intervening probes, so the window is guaranteed to be crossed; if the files still are not visible the command warns and lets the restore surface the underlying error. The retry handles both return shapes of `Test-DbaPath` (bare bool for a single scalar path, objects otherwise). "Fixes #10512" stays.

**5. Failed-restore continuation test (review 1 "missing regression test", review 2 point 3).** Added: "Continues to the next piped database when one restore fails". It backs up both databases to the share, deletes the first backup file (history intact, so `-UseLastBackup` deterministically hands the restore a missing file - no timing race), pipes both databases, and asserts exactly one result row per database, first Failed, second Successful and present on the destination. This is also the reported no-`NewName` command shape: on 2.8.3 this scenario loses the second database with no output at all.

**6. Sentinel-owner false negative (review 2 point 4).** Fixed - the sentinel database owner is now a dedicated `dbatoolsci_owner*` login created for the test. The sweep sets destination owners to the source database owner, so with both at `sa` a sweep was undetectable; now any sweep flips the sentinel to `sa` and the assert catches it. The piped-success test also asserts nonblank `DestinationDatabase` per database, closing review 2's "empty-NewName test only exercises begin validation" gap: the happy-path pipe run now proves the resolved destination name per record.

Lab verification after the rework: 12 unit + 5 integration tests green, plus a live parameter-mode `-Database` + `-NewName` rename check (Successful, restored under the new name). The `Test-DbaPath` scalar-bool fix and the Verbose-to-Warning change both reviews endorsed are unchanged.

potatoqualitee:
Round-2 review feedback addressed in 42127627d. Point by point:

**Empty/incomplete visibility probe treated as success** — fixed. The retry now materializes the `Test-DbaPath` results and requires one conclusive result for every backup path (count match against the requested list, plus a non-empty list) before anything counts as visible. An indeterminate probe — `Test-DbaPath` emits nothing at all when its own connection attempt fails via `Stop-Function -Continue` — now keeps retrying until the negative-cache deadline and then emits the timeout warning.

**Retry state-machine coverage** — added `Keeps retrying when the visibility probe returns no results`: a module-scoped mock of `Test-DbaPath`, filtered to the backup file path so the `-SharedPath` folder checks still hit the real command, returns nothing; the test asserts the copy waits out the window and emits the negative-cache warning instead of proceeding. This one test exercises both the no-output case and persistent-not-visible-through-deadline (every probe is indeterminate until the deadline). The false→true recovery transition drives the identical loop machinery — not-visible, sleep, re-probe — and both exit conditions (visible, deadline) are covered between this test and the immediately-visible copies elsewhere in the suite, so I stopped at one focused test rather than three.

**Continuation test ordering** — fixed. The failing database is now piped explicitly first via two separate `Get-DbaDatabase` calls; the test no longer leans on `sys.databases` emission order, and a failure on the last record can no longer mask the old discard behavior.

**NewName help** — updated to state the pipeline-input restriction and the escape hatch: specify the database with `-Database` and provide `-NewName`.

**Single-database piped rename (the P1 in one review)** — kept the atomic rejection, deliberately. The other two reviews endorse it as the right implementation, and buffering pipeline input to allow the one-record case was considered and rejected in the previous round: it would restructure the streaming `process{}` flow of a very large command inside a regression fix, for a form the docs never promised. The guard fails fast before any destination work with an actionable message, and the help now documents the restriction. If the maintainers want the buffer-one variant, I'm happy to do it as a follow-up PR where the restructuring can be reviewed on its own.

potatoqualitee:
The CI failure on 42127627d was the new mock-based retry test — the module-scoped `Mock Test-DbaPath` intercepted the internal `-SharedPath` folder probe, and a Pester 5 filtered mock has no real-command fallback for non-matching calls. Removed the test in e5c56fa5a per project policy (integration tests only, no mocks). The no-output probe path can't be driven from real instances, so it stays untested; the conclusive-result guard itself remains in the production retry loop and the rest of the suite covers the visibility behavior.

potatoqualitee:
github actions are broken :(

PR #9716, opened 2025-07-23
Invoke-DbaDbDataGenerator - Add workaround for DateTimeReference

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes # )
 - [ ] New feature (non-breaking change, adds functionality, fixes # )
 - [ ] Breaking change (affects multiple commands or functionality, fixes # )
 - [ ] Ran manual Pester test and has passed (`Invoke-ManualPester`)
 - [ ] Adding code coverage to existing functionality
 - [ ] Pester test is included
 - [ ] If new file reference added for test, has is been added to github.com/dataplat/appveyor-lab ?
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

The value of the property DateTimeReference is currently $null, which causes the changed line to throw an exception.
We have to contact the developer of Bogus to solve the issue.

This is the original code to test if this throws on other installations as well:
```
$script:faker = New-Object Bogus.Faker('en')
$supportedFakerSubTypes = ($script:faker | Get-Member -MemberType Property) | ForEach-Object { ($script:faker.$($_.Name)) | Get-Member -MemberType Method | Where-Object { $_.Name -notlike 'To*' -and $_.Name -notlike 'Get*' -and $_.Name -notlike 'Trim*' -and $_.Name -notin 'Add', 'Equals', 'CompareTo', 'Clone', 'Contains', 'CopyTo', 'EndsWith', 'IndexOf', 'IndexOfAny', 'Insert', 'IsNormalized', 'LastIndexOf', 'LastIndexOfAny', 'Normalize', 'PadLeft', 'PadRight', 'Remove', 'Replace', 'Split', 'StartsWith', 'Substring', 'Letter', 'Lines', 'Paragraph', 'Paragraphs', 'Sentence', 'Sentences' } | Select-Object name -ExpandProperty Name }
```

So please test yourself and add a comment with the result.

PR #10517, opened 2026-08-06
Data generation - Make the supported randomizer types match what actually generates

andreasjordan:
## Type of Change

 - [x] Breaking change (affects multiple commands or functionality)
 - [x] Pester test is included

Follow-up to #9716. That PR worked around a `$null` `DateTimeReference` by excluding the property by name. bchavez/Bogus#612 has since been answered: `DateTimeReference` is **null by design**, an opt-in anchor the caller sets. So nothing is coming from upstream, and the real defect is further back — the commands disagreed with each other about what can be generated.

Five commits, reviewable in order. The fifth addresses the review below.

## The problem

Data generation had four vocabularies that all disagreed:

| Source | Used by |
|---|---|
| `bin/datamasking/columntypes.json` | `New-DbaDbDataGeneratorConfig` writes configs |
| `bin/randomizer/en.randomizertypes.csv` | `Test-Dba*Config`, `Invoke-DbaDbDataMasking` validate |
| Live reflection over `Bogus.Faker` | `Invoke-DbaDbDataGenerator` validated |
| A hardcoded `switch` | `Get-DbaRandomizedValue` actually generates |

A config could pass every validation step and still generate nothing, or be rejected for something that works.

## 1. Use the randomizer types as the gate

`Invoke-DbaDbDataGenerator` built its supported lists by reflecting over a `Bogus.Faker` object. That can never agree with `Get-DbaRandomizedValue`, which validates against the randomizer types. Measured against Bogus 35.6.5:

* **34 subtypes** were offered that no randomizer type knows, so the config was accepted and then rejected downstream, warning without naming the column.
* **11 working combinations were rejected.** Every subtype Bogus exposes as a *property* rather than a *method* was invisible, because the list was built from methods only — `Get-DbaRandomizedValue.ps1:587` reads them as a property. That is `Person/DateOfBirth`, `Gender`, `Website`, `Phone`, six working `Lorem` subtypes excluded by name, and `Date/Random`.
* `Locale` and `Context` are not generators at all, so their string and dictionary methods became subtypes: `ContainsKey`, `TryAdd`, `EnsureCapacity`.
* It broke outright whenever Bogus added a null property, which is what #9716 patched.

Using the randomizer types removes the coupling to the Bogus type system, so a dbatools.library bump can no longer break this command.

## 2. Bring the randomizer types back in line with Bogus

The gate is only as good as the list, and the list had drifted.

| | |
|---|---|
| Removed, gone from Bogus | `Image/Abstract, Animals, Business, Cats, City, Fashion, Food, Image, Nature, Nightlife, People, Sports, Technics, Transport` |
| Removed, `bool` predicates that throw when called as generators | `Name/HasFirstNameList, SupportsGenderFirstNames, SupportsGenderLastNames, SupportsGenderPrefixes` |
| Added, what the Image data set offers instead | `Image/LoremFlickrUrl, PicsumUrl, PlaceImgUrl` |

## 3. Every randomizer type returns a value or says why

Of the 191 types, **29 produced nothing usable**: 12 returned `$null` in silence because the `switch` has no branch for their type and no `default` either, 16 failed with a raw .NET overload error because they need an argument the command never passes, one returned an empty string.

Fixed outright:

* **The type resolved for a subtype was cached in the module scope.** The first subtype used in a session decided the type for every later call, so after one `-RandomizerSubType FirstName`, `-RandomizerSubType ZipCode` asked the Name data set for a zip code and threw. The zipcode test in `Get-DbaRandomizedValue.Tests.ps1` was skipped on AppVeyor for exactly this, with the error in the comment. It runs now.
* Both branches of the `Rant` type tested for `"reviews"`, so `Review` returned nothing and `Reviews` returned a single review.
* The `Shuffle` guard compared against `$null`, but `Value` is a string, so it is empty and never `$null` when omitted. The guard never fired.
* `Random/SByte` was called with the default min and max of 1 and 255, which do not fit in an `sbyte`.

Implemented: `Random/AlphaNumeric, ClampString, Digits, Replace, ReplaceNumbers, WordsArray`, `Date/BetweenOffset`, and the Hashids encoders. Added: `Music/Genre` and `Vehicle/Fuel, Manufacturer, Model, Type, Vin`, complete data sets that were never wired up.

Removed because they cannot produce a random value here — the decoders reverse a hash rather than generate one, the collection ones need a .NET list or type a JSON config cannot express, `ReplaceSymbols` needs a `Func` delegate, and `DataUri` emits a multi-kilobyte base64 blob:

`Hashids/Decode, DecodeHex, DecodeLong, Decrypt, DecryptHex` · `Random/CollectionItem, Enum, ListItem, ListItems, WeightedRandom, ReplaceSymbols` · `Image/DataUri`

A `default` branch now catches any type without a generator.

## 4. Check the columns before building the insert

The insert names every column and is built before the rows are generated, but the column checks ran inside the row loop and each skipped only its own column. The statement was left with fewer values than columns, so SQL Server answered `Incorrect syntax near ')'` — nothing about the config that caused it. The checks also ran once per row.

## 5. Validate the combination, not two lists (from the review)

The type and the subtype were still checked against two independent lists, so `Name/ZipCode` passed because `Name` is a type somewhere and `ZipCode` is a subtype somewhere. They are checked as combinations now, in `Get-DbaRandomizedValue`, both `Test-Dba*Config` commands and the generator preflight.

That found a config dbatools writes itself: `New-DbaDbDataGeneratorConfig` starts from masking type `Random` and for `smalldatetime` changed only the subtype to `Date`. `Random/Date` is not a combination, and the manual `+= "Date"` hid it. That column now gets `Date/Past`.

The catalog gained a **`RequiredParameter`** column. `Random/Replace` and `ReplaceNumbers` need a `Format`, `Random/ClampString` and `Shuffle` need a `Value`, `Static/None` marks a value the config supplies. A data generation config has nowhere to put any of those, so the generator rejects those columns up front, while direct `Get-DbaRandomizedValue` calls and masking configs, which do carry `Format`, still use them.

Also removed, because they have no scalar representation and wrote a type name into the column:

```
Person/Address   -> Bogus.Person+CardAddress
Person/Company   -> Bogus.Person+CardCompany
Person/Context   -> System.Collections.Generic.Dictionary
```

And two more from the review: the column checks now run before the unique-index values are generated rather than after, and `Person/DateOfBirth` is formatted with the invariant culture before it becomes a SQL literal instead of picking up the process culture.

## Net effect

| | Before | After |
|---|---|---|
| Randomizer types | 206 | 182 |
| ...that generate a value with no extra parameters | 158 | 175 |
| ...that return nothing without saying why | 12 | 0 |
| ...that fail with a raw .NET error | 16 | 0 |
| ...usable from a data generation config | unchecked | 177 |
| Working combinations the generator rejected | 11 | 0 |
| Subtypes the generator offered that cannot generate | 34 | 0 |
| Invalid combinations accepted (`Name/ZipCode`, `Random/Date`) | yes | 0 |

## Tests

* The generator rejects a non-generator property, an invalid combination and a combination needing a `Format`, each with the **column named**, and skips the table rather than running a broken insert.
* The same rejection happens for a table with a unique index, before any value is generated.
* `Person/DateOfBirth` generates, and generates the same dates under `de-DE`.
* Subtype-only calls resolve their type independently, covering the cache bug.
* Every randomizer type returns a value or a message, counting returned objects rather than testing them for truth so `Random/Bool` returning `$false` is not mistaken for silence.
* Every randomizer type still exists in the loaded Bogus library, so the next library bump that removes something fails a test instead of reaching a user.

Run against SQL Server 2019 in a lab, 66 tests over 11 files, all green:

```
Get-DbaRandomizedDataset          2/2    Invoke-DbaDbPiiScan              3/3 (9 skipped)
Get-DbaRandomizedDatasetTemplate  2/2    New-DbaDbDataGeneratorConfig     2/2
Get-DbaRandomizedType             5/5    New-DbaDbMaskingConfig           3/3
Get-DbaRandomizedValue            9/9    Test-DbaDbDataGeneratorConfig    3/3
Invoke-DbaDbDataGenerator       14/14    Test-DbaDbDataMaskingConfig      7/7
Invoke-DbaDbDataMasking           7/7
```

## Upgrade note

A config naming a removed type, an invalid combination, or a combination that needs a `Format` or `Value` now fails its `Test-Dba*Config` check instead of silently generating nothing or writing a type name into a column. Every case was already producing unusable data.


potatoqualitee:
wow nice, how'd you find this?

potatoqualitee:
ahhh an old issue filed in 2025

andreasjordan:
Currently working on some old stuff with my new Claude... (and in this case the answer in the Bogus repo)

potatoqualitee:
GPT Pro Sayz:

## Verdict

**Request changes.** I reviewed the current head, `a7dabe5`. The reflection removal and subtype-cache fix are directionally correct, but the new capability checks still accept configurations that cannot generate valid SQL values. I found five actionable issues before approval.

## 1. Medium: support is still validated as two independent lists, not valid Type/SubType pairs

`Invoke-DbaDbDataGenerator` builds one list of types and another list of subtypes. `Get-DbaRandomizedValue` performs the same independent checks. Consequently, `Name/ZipCode` passes validation because `Name` exists somewhere and `ZipCode` exists somewhere, even though the only valid pair is `Address/ZipCode`. Execution then reaches:

```powershell
$script:faker.Name.ZipCode()
```

which does not exist.

There is already a concrete dbatools-generated version of this problem. `New-DbaDbDataGeneratorConfig` initializes `$type = "Random"` and, for `smalldatetime`, changes only `$subType` to `"Date"`. The PR manually appends `"Date"` to the generator’s subtype list, so `Random/Date` passes the new preflight even though that pair does not exist in the CSV and `Get-DbaRandomizedValue` rejects it. The configuration checker reports a validation object, but `Invoke-DbaDbDataGenerator` only reacts to thrown exceptions, so it continues processing.

The checks should validate the exact pair:

```powershell
$supportedPair = Get-DbaRandomizedType |
    Where-Object {
        $_.Type -eq $columnobject.MaskingType -and
        $_.SubType -eq $columnobject.SubType
    }

if (-not $supportedPair) {
    # reject the column
}
```

That pair validation belongs in `Get-DbaRandomizedValue`, both `Test-Dba*Config` commands, and the generator preflight. The `smalldatetime` configuration should also be changed to a real pair, likely `Date/Past`, and the manual `+= "Date"` exception should be removed.

## 2. Medium: the randomizer catalog contains entries the data generator cannot invoke

The CSV is a list of things `Get-DbaRandomizedValue` knows about, not a complete capability contract for `Invoke-DbaDbDataGenerator`.

For example, this PR newly allows `Random/Replace`; the previous reflection-based gate explicitly excluded `Replace`. However:

* `Get-DbaRandomizedValue` requires `-Format` for `Replace` and `ReplaceNumbers`.
* It requires `-Value` for `Shuffle` and `ClampString`.
* The data-generator configuration schema contains neither `Format` nor `Value`.
* The generator’s call to `Get-DbaRandomizedValue` passes only type, subtype, character set, locale, minimum, and maximum.
* `Static/None` is also in the accepted catalog even though the PR intentionally routes it to the “no generator” warning.

These configurations therefore pass the new preflight but still cannot return the requested value.

The catalog needs command-specific capability metadata, for example:

```text
Type,SubType,GeneratorSupported,MaskingSupported,RequiredParameters,OutputKind
```

A smaller generator-specific allowlist would also work. At minimum, add an integration test that attempts `Random/Replace` through `Invoke-DbaDbDataGenerator`; testing the direct randomizer with missing parameters does not cover this path.

## 3. Medium: several newly accepted `Person` properties are not usable SQL scalar values

The PR describes `Person/Address`, `Person/Company`, and the other property-backed values as working pairs. `Get-DbaRandomizedValue` returns those properties directly. In Bogus, however:

* `Person.Address` is a `CardAddress` object.
* `Person.Company` is a `CardCompany` object.
* Neither class defines a scalar representation.
* `Invoke-DbaDbDataGenerator` later calls `.ToString()` on these values.

As a result, a varchar column configured with `Person/Address` or `Person/Company` will be populated with an object type name rather than an address or company name. This is silent incorrect output, not a rejected configuration.

The tested `Person/DateOfBirth` case has a related portability problem. It returns a `DateTime`, which is also converted using parameterless `.ToString()`. That creates a culture-dependent SQL literal. The current integration test verifies only that five dates were accepted under the CI environment; it does not exercise a non-US process culture or verify that month and day were preserved.

The generator should either exclude complex output types or explicitly map them to scalar fields. Date/time values should be formatted invariantly or passed through the same typed conversion machinery used by the masking command.

## 4. Medium: the new validation occurs after unique-index values are generated

The command performs the complete unique-index value generation loop before it selects `$tablecolumns` and runs the new unsupported-column checks. Therefore, an invalid pair, parameter-dependent randomizer, or complex output on an indexed column is invoked before the table can be skipped. It can reach the uniqueness retry logic with missing or stale values and produce the lower-level behavior the new preflight is intended to prevent.

All of the new invalid-configuration integration cases explicitly use:

```json
"HasUniqueIndex": false
```

so this ordering is not covered.

Column selection and full capability validation should occur immediately after confirming that the table exists, before `$uniqueValues` is initialized and before any indexed values are generated. Add one invalid-config test using a unique indexed column.

## 5. Medium: the all-types test is nondeterministic for `Random/Bool`

The new invariant test determines whether a value was returned using:

```powershell
if ($typeValue) {
    $typeResults[$typeKey] = "value"
}
```

`Random/Bool` is in the catalog, and Bogus legitimately returns either `$true` or `$false`. When it returns `$false`, PowerShell has received a valid output object, but the condition evaluates false. Since there is no warning, the test labels the type `"silent"` and fails. The current green run only means that Bool returned true during that invocation.

Capture and count success-stream objects instead of testing truthiness:

```powershell
$typeValue = @(Get-DbaRandomizedValue @splatRandomValue)

if ($typeValue.Count -gt 0) {
    $typeResults[$typeKey] = "value"
} elseif ($typeWarning) {
    $typeResults[$typeKey] = "message"
} else {
    $typeResults[$typeKey] = "silent"
}
```

An explicit empty-string check can remain separate if empty strings are considered unusable.

## Documentation

The comment-based help still omits `Music`, `Vehicle`, and `Static` from the available types. It also documents `Value` as required only for `Shuffle`, although `ClampString` now requires it, and does not explain that `Replace` and `ReplaceNumbers` require `Format`. Since the PR intentionally makes breaking catalog changes, the public help should be updated in the same change.

The subtype-cache correction and the `Rant/Review` versus `Rant/Reviews` dispatch fix look correct. I could not independently execute Pester locally because this environment does not have PowerShell installed; the findings above come from the current diff, connected repository sources, Bogus 35.6.5 source, and the visible GitHub Actions results.

andreasjordan:
I will work on this later today...

andreasjordan:
Thanks — that review was worth its length. I verified all five points against a live SQL Server 2019 lab before changing anything, and every one of them held up. Fixed in `ab7378e0`.

**1. Independent lists instead of combinations.** Correct, and worse than the example suggests. `Random/Date` is written by dbatools itself: `New-DbaDbDataGeneratorConfig` starts from `$type = "Random"` and for `smalldatetime` changed only `$subType` to `"Date"`, and the manual `+= "Date"` hid it from the preflight. Combination validation now lives in `Get-DbaRandomizedValue`, both `Test-Dba*Config` commands and the generator preflight; that column now gets `Date/Past` and the exception is gone.

**2. Catalog is not a generator capability contract.** Agreed, and this was a regression I introduced — the old reflection gate excluded `Replace` by name and my CSV gate let it back in. The catalog gained a `RequiredParameter` column (`Format`, `Value`, `StaticValue`). The generator rejects any combination that has one, since its config schema has nowhere to put it, while direct `Get-DbaRandomizedValue` calls and masking configs, which do carry `Format`, still work. `Get-DbaRandomizedValue` reads the same column instead of naming subtypes in code. Integration test added for `Random/Replace` through `Invoke-DbaDbDataGenerator`, as you asked.

**3. Person properties are not scalars.** Confirmed:

```
Address      type=CardAddress      ToString()=Bogus.Person+CardAddress
Company      type=CardCompany      ToString()=Bogus.Person+CardCompany
Context      type=Dictionary`2     ToString()=System.Collections.Generic.Dictionary`2[...]
```

My "14 working pairs" measurement counted "not null and not empty" as working, so an object passed. All three are removed. `DateOfBirth` is now formatted with the invariant culture before it reaches the SQL literal, with a test that runs the generator under `de-DE`.

**4. Validation after unique-index generation.** Column selection and all checks moved to directly after the table is found, before any value is generated. Test added with `HasUniqueIndex: true`.

**5. `Random/Bool` nondeterminism.** Confirmed empirically — 8 of 12 runs returned `False`. The test counts returned objects now rather than testing them for truth, which also covers `Random/SByte` returning 0. It had passed twice by luck.

Help updated for `Music`, `Vehicle`, `Static`, and the `Value`/`Format` requirements.

Two notes where I did not follow the suggestion exactly:

- I kept `Static/None` in the catalog rather than removing it. `Invoke-DbaDbDataMasking` tests `$columnComposite.Type -eq 'Static'` for composite items and `Test-DbaDbDataMaskingConfig` validates `MaskingType` against the catalog, so removing it would reject every valid `StaticValue` masking config. It carries `RequiredParameter = StaticValue`, so the generator rejects it and `Get-DbaRandomizedValue` explains that it marks a supplied value rather than a generated one.
- Combination validation went into `Invoke-DbaDbDataMasking`'s config test but not into its own two independent lists. `Get-DbaRandomizedValue` backstops it either way, and reworking the masking command's row loop is a bigger change than this PR should carry.

66 tests over 11 files, all green against SQL Server 2019.

potatoqualitee:
LGTM!

PR #10518, opened 2026-08-06
Invoke-ManualPester - Check the test file and the repository style gate

andreasjordan:
Follow-up to the question in https://github.com/dataplat/dbatools/pull/10463#issuecomment-5153094978: `Invoke-ManualPester` was supposed to warn about the whitespace problems and did not.

## Type of Change

 - [X] Pester test is included
 - [ ] Unit test is included
 - [X] Documentation
 - [ ] Build system

### Purpose

A local run of `Invoke-ManualPester` could be green while CI failed on style, because neither of the two checks that catch it is part of a command's test file. PR #10463 is the worked example: nine trailing space lines, and nothing locally reported any of them.

I checked out that PR's head commit and ran ScriptAnalyzer exactly the way `-ScriptAnalyzer` called it. It reported two of the nine, both in `public\Install-DbaFirstResponderKit.ps1`. The other seven were in the test file, and `-ScriptAnalyzer` never looked at the test file: it set `$HeadFunctionPath` to the first *coverage* file, so the file you passed in was discarded. The check that actually failed the build is `Describe "$ModuleName style" -Tag Compliance` in `tests\dbatools.Tests.ps1`, and the runner had no way to run it.

### Approach

**`Invoke-ManualPester`**

- `-ScriptAnalyzer` now checks the test file as well as the command under test.
- `-ScriptAnalyzer` now passes `tests\PSScriptAnalyzerRules.psd1`, which the help already claimed it used. That profile had no consumer anywhere in the repository, and it did not include `PSAvoidTrailingWhitespace` - the rule the style test fails the build on - so that is added. `CheckOpenBrace` is turned off: across the test tree it produced 180 findings, 178 of them the `{ ... } | Should -Throw` idiom, which buried everything else the rule found.
- New `-Compliance` switch runs the `Compliance` tagged tests over the whole repository (about half a minute). A run without it now says so at the end.
- The command under test is resolved from its own file name. Taking the first entry of `Get-CoverageIndications` measured and analysed a *dependency* for a private function, and returned nothing at all for 19 test files - where `-ScriptAnalyzer` then failed on a null path and `-Coverage` silently measured nothing.
- The Pester 4 branch is removed. All 742 test files declare the Pester 5 header, so it was unreachable; a file without the header is now reported instead of being run with something else. The version warnings that contradicted their own constants are corrected, and the `$config_TestIntegration` variable that is set nowhere in the repository is gone.
- A run over several files ends with a summary of which files failed, with the failing line and message, instead of leaving the reports to be read by scrolling back.
- It refuses a module base without a `tests` folder rather than creating a `.git` directory in it, and says so when `dbatools.psd1` cannot be found instead of failing later on a confusing path.

**Test file structure checks**

`tests\dbatools.Tests.ps1` gains `Describe "$ModuleName test file structure" -Tag Compliance`, ported from the `testing-dbatools` repository. They are aggregated one test per rule rather than one per rule and file - the per-file form is 14023 results and two and a half minutes, this is 19 results in 24 seconds, and a failure still names every file it found.

Thirteen rules are enforced and the tree passes them. Six describe a backlog the tree does not meet yet and are skipped unless `$env:DbatoolsTestFileGoals` is set.

Several of these are not style. A loop that emits `It` blocks, a `-ForEach` fed from a `BeforeAll` and an assertion never piped to `Should` all produce a green report while testing nothing.

**One finding worth its own look, deliberately not fixed here**

`tests\InModule.Help.Tests.ps1` builds its `Describe` blocks in `foreach ($command in $global:commandsWithHelp)`, and a `BeforeAll` fills that variable - long after discovery has read the loop. Verified in a fresh session: `Invoke-Pester` on that file reports `TotalCount 0`. The whole help suite has never run an assertion.

Rewriting it onto `BeforeDiscovery` and `-ForEach` switches several thousand help assertions on at once and nobody knows yet how many fail, so it is named in `$knownLoopExceptions` with the reasoning rather than fixed in this PR.

**Smaller things**

- `Get-TestInstanceUsage` reports which `$TestConfig` instances a test file uses, and therefore which CI lane it runs in. 395 test files use `InstanceSingle`, 21 need `Hadr`, 220 need no instance at all.
- `tests\constants.local.ps1.example` gains the missing `$config['Temp']` placeholder - the other Learning from #10463, where the run failed because `C:\Temp` did not exist and nothing hinted that the key could be set.
- Three test files needed a fix to pass the new rules: `New-DbaConnectionStringBuilder` (backtick continuation), `Test-DbaBackupEncrypted` (plain `$splat`), `Get-DbaUnusedLogin` (an `EnableException` enabled in `AfterAll` and never removed, which carries into whatever runs next in the same session).

### Commands to test

`Invoke-ManualPester` itself, and the `Compliance` tagged tests:

```powershell
Invoke-ManualPester -Path Get-DbaDatabase -ScriptAnalyzer -Compliance

# the backlog the Goal rules describe
$env:DbatoolsTestFileGoals = 1
Invoke-Pester -Path .\tests\dbatools.Tests.ps1 -TagFilter Compliance
```

### Learning

The `{ Get-DbaFoo } | Should -Throw` idiom trips `PSUseConsistentWhitespace`'s `CheckOpenBrace`, because the statement starts with the brace. 178 of 180 findings across the test tree were that; the two that were not are a nested `FindAll({` and one genuine double space before a brace.

### DIFF
--- a/.github/CONTRIBUTING-TESTING.md
+++ b/.github/CONTRIBUTING-TESTING.md
@@ -329,14 +329,31 @@ Invoke-ManualPester -Path Get-DbaDatabase -TestIntegration -Coverage
 # Run with full coverage including dependencies
 Invoke-ManualPester -Path Get-DbaDatabase -TestIntegration -Coverage -DependencyCoverage
 
-# Run with script analyzer check
+# Run with script analyzer check on the test file and the command
 Invoke-ManualPester -Path Get-DbaDatabase -TestIntegration -ScriptAnalyzer
+
+# Run the repository wide style checks that CI enforces
+Invoke-ManualPester -Path Get-DbaDatabase -TestIntegration -Compliance
 
 # Run multiple tests matching a pattern
 Invoke-ManualPester -Path "*Backup*" -TestIntegration
 
 # Show less output
 Invoke-ManualPester -Path Get-DbaDatabase -TestIntegration -Show None
+```
+
+### Before opening a pull request
+
+`Invoke-ManualPester -Path <command>` runs only that command's test file. Two of the checks CI runs
+belong to no command's test file at all, so a green run of it says nothing about them:
+
+| Switch | What it adds |
+|---|---|
+| `-ScriptAnalyzer` | Checks the test file and the command under test against [tests/PSScriptAnalyzerRules.psd1](../tests/PSScriptAnalyzerRules.psd1), the repository style profile. |
+| `-Compliance` | Runs the `Compliance` tagged tests from [tests/dbatools.Tests.ps1](../tests/dbatools.Tests.ps1) over the whole repository: no UTF-8 BOM, no tabs, no trailing spaces, no ScriptAnalyzer errors. Takes about half a minute. |
+
+```powershell
+Invoke-ManualPester -Path Get-DbaDatabase -TestIntegration -ScriptAnalyzer -Compliance
 ```
 
 ### Method 2: Direct Pester Invocation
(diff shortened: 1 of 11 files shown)


potatoqualitee:
LGTM!

PR #10519, opened 2026-08-06
InModule.Help - Make the help tests actually run

andreasjordan:
Follow-up to #10518, which found this while porting the test file structure checks.

## Type of Change

 - [X] Pester test is included
 - [X] Unit test is included
 - [X] Documentation
 - [ ] Build system

### Purpose

`tests\InModule.Help.Tests.ps1` has never run a single assertion.

It wrapped every `Describe` in `foreach ($command in $global:commandsWithHelp)` at the top level of the outer `Describe`. That loop runs while Pester *discovers* the tests, and the `BeforeAll` that fills `$global:commandsWithHelp` runs long afterwards - so the loop found nothing and built nothing. Verified in a fresh session: `Invoke-Pester` on that file reports `TotalCount 0`. The whole help suite - auto-generated synopsis, missing description, missing example, parameter help - has been reporting success by testing nothing.

### Approach

**Three bugs in the test file**

Two more were hiding behind the first, and both would have fired on all 720 commands the moment the file worked:

1. The inner `foreach ($parameter in $commandParameters)` reads a variable that a `BeforeAll` fills, so fixing only the outer loop still produces no parameter tests - two thirds of the suite.
2. The auto-generated check was `Should -Not -BeLike "*`+"`"+`[`+"`"+`<CommonParameters`+"`"+`>`+"`"+`]*"`. The backticks escape for PowerShell, not for the wildcard engine, so the pattern string is `*[<CommonParameters>]*` - and in a wildcard `[...]` is a **character class** matching any one of those letters. It matched almost every real synopsis:

   ```
   "Retrieves database objects from SQL Server instances." -like "*[<CommonParameters>]*"   ->  True
   ```

   It is a regex now, where the intent cannot be lost to quoting.

**Common parameters**

The hard coded `$Common` list was the Windows PowerShell 5.1 set. PowerShell 7.4 added `ProgressAction`, so on modern editions every command was measured against a parameter that comment based help never documents: **2151 failures across 717 commands**, none of which say anything about the help. It now comes from `[System.Management.Automation.PSCmdlet]::CommonParameters`, which is edition correct - it contains `ProgressAction` on 7.6 and leaves it out on 5.1, both verified. `Get-TestConfig` already builds `$TestConfig.CommonParameters` the same way.

`WhatIf` and `Confirm` are deliberately *not* excluded. They are `OptionalCommonParameters`, but 368 commands document them and three did not, so they are held to the same standard.

**Shape**

The assertions are aggregated one per rule rather than one per rule and command. The per-command form is 26823 test results and two and a half minutes; this is 11 results in 14 seconds, and a failure still names every command it applies to. Same shape as the test file structure checks in #10518.

There is also a new test that the command list is not empty, so the file cannot silently go back to testing nothing.

**A landmine worth knowing about**

Pester 6 aborts the *entire discovery run* on the first `-ForEach` handed an empty array, not just that block. `Measure-DbatoolsImport` has no non-common parameters, so a straightforward `-ForEach` rewrite silently produced only 432 of 720 commands - everything alphabetically past `Measure-` was gone, with nothing in the normal output to say so. The aggregated shape sidesteps it entirely.

**The eight real findings, all fixed here**

| Command | Finding |
|---|---|
| `Export-DbaCsv`, `Remove-DbaInstanceList`, `Start-DbaAzMigration` | Use `SupportsShouldProcess` but do not document `WhatIf` and `Confirm`. 368 other commands do. |
| `Set-DbatoolsInsecureConnection` | Two urls under one `.LINK`, so PowerShell folds them into a single navigation link whose `uri` is both lines joined and `Get-Help -Online` cannot use it. Split into two `.LINK` blocks. |
| `Measure-DbatoolsImport` | No blank line between the code and the description of the first `.EXAMPLE`, so the description became part of the code and the example had no remarks. |

That is a good outcome - the help itself is in far better shape than "2159 failures" suggests.

**One line removed from #10518**

With this file fixed, the `$knownLoopExceptions` entry that #10518 added to the test file structure checks is no longer needed, so the "no test file builds test blocks in a loop" check applies to every test file again.

### Commands to test

```powershell
Invoke-ManualPester -Path tests\InModule.Help.Tests.ps1
Invoke-ManualPester -Path Export-DbaCsv -Compliance
```

Verified green on **PowerShell 7.6** and **Windows PowerShell 5.1**: `InModule.Help.Tests.ps1` 11/11, `dbatools.Tests.ps1` 33 passed / 0 failed / 6 skipped.


potatoqualitee:
thank you 🙏🏼

PR #10520, opened 2026-08-06
Pester 6 - Drop stale v4/v5 references and take the version out of names

andreasjordan:
The suite moved to Pester 6 a while ago, but a lot of the tree still names 4 or 5. This covers everything **except** the 744 `#Requires` headers — see the open question at the bottom.

## Stale Pester 4

| Where | Was | Now |
|---|---|---|
| `.devcontainer/Dockerfile` | installed Pester `4.4.3` | the pinned `6.0.0` |
| `.github/prompts/prompt.md` | the "Pester v4 to v5 Migration" prompt | deleted — nothing has referenced it since the migration finished, and `.gitignore` already ignores stray `prompt.md` files |
| `Invoke-ManualPester.ps1`, `tests/CLAUDE.md` | prose about "no Pester 4 tests left" / "confusion left over from Pester v4" | reworded without the version archaeology |

## Version out of names

So that a future move to Pester 7 is a one-line change rather than another sweep:

- `$pester5Config` / `$pester5Result` → `$pesterConfig` / `$pesterResult` (both `Invoke-ManualPester.ps1` and `appveyor.pester.ps1`; the latter also had a `$pester5config` casing variant)
- `Pester5Coverage*.xml` → `PesterCoverage*.xml`, `Pester5Results*.xml` → `PesterResults*.xml`, `TestFailureSummary_Pester5_*.json` → `TestFailureSummary_*.json`, `DetailedTestFailures_Pester5.json` → `DetailedTestFailures.json`. `appveyor.post.ps1` reads the new coverage glob. Nothing outside these two scripts referenced the old names.
- `appveyor.pester.ps1` now names the version once, in `$PesterRequiredVersion`, which feeds both the `Import-Module` and the `PesterVersion` field it reports. That field was hardcoded to `"5"` in four places while the script imported 6.

## Two deliberate non-changes

**The codecov flag still says `pester5_`.** codecov keys its history off the flag name, so renaming it starts a new series and loses the trend on the existing dashboards. There is now a comment saying so.

**`Update-DbaInstance.Tests.ps1` still says "In Pester 4 the same strings were a substring match and did pass."** That is accurate history explaining why the assertions need wildcards — rewriting it to say 6 would make it false, and deleting it invites someone to "fix" the wildcards back out.

## One behavior change worth a look

`Invoke-ManualPester` gated on a minimum of Pester `5.0.0.0`; it now gates on `6.0.0`, so a local pass means what a CI pass means. If you would rather keep letting Pester 5 through locally, this is the line to revert.

The bound is `[Version] "6.0.0"` and **not** `"6.0.0.0"` on purpose. Pester reports itself as `6.0.0`, and `[Version]` treats an absent revision as `-1`, so `6.0.0 -lt 6.0.0.0` is `$true` and a four-component bound rejects the pinned build. I wrote `6.0.0.0` first and the runner refused to start.

## Open question — the 744 `#Requires` headers

Every test file (plus `gh-azmigration-actions.ps1`) carries:

```powershell
#Requires -Module @{ ModuleName="Pester"; ModuleVersion="5.0" }
```

Left alone on purpose, because it is not cosmetic. `#Requires` reads that as a **floor** ("5.0 or newer"), so 6.0.0 already satisfies it, and `tests/CLAUDE.md` documents the retention as intentional. Bumping it to `6.0` would start rejecting a Pester 5 install rather than silently running on it. That is a real policy decision, and it touches 744 files, so it belongs in its own commit on this branch just before merge — if the contributors want it at all.

Note that `Invoke-ManualPester`'s header check still matches `ModuleVersion="5\.`, so the headers and that regex have to move together.

## Testing

- Every edited PowerShell file parse-checked.
- `Invoke-ManualPester -Path tests\Get-DbaDatabase.Tests.ps1` runs green through the renamed variables and the new version gate.
- `Invoke-ManualPester -Compliance` clean.
- Repo swept for leftovers: the only remaining `pester4`/`pester5` strings are the three deliberate ones above plus the headers.


.github/runners/tests/runner-policy.Tests.ps1
```diff
@@ -344,7 +344,7 @@ Describe "Get-DesiredRunnerPools demand-driven sizing" {
 
 Describe "Get-PoolJobDemandFromJobs" {
     BeforeAll {
-        # Defined in BeforeAll, not the Describe body: Pester 5 runs the Describe body
+        # Defined in BeforeAll, not the Describe body: Pester 6 runs the Describe body
```
andreasjordan: Maybe remove the number "6" here.

tests/appveyor.pester.ps1
```diff
@@ -244,7 +248,7 @@ function Get-ComprehensiveErrorMessage {
             $errorMessages += "StdErr: $($TestResult.StandardError)"
         }
 
-        # Check Block.ErrorRecord for container-level errors (common in Pester 5)
+        # Check Block.ErrorRecord for container-level errors (common in Pester 6)
```
andreasjordan: Maybe remove " (common in Pester 6)"

tests/appveyor.pester.ps1
```diff
@@ -325,7 +329,7 @@ function Export-TestFailureSummary {
 
     $failedTests = @()
 
-    # Pester 5 format
+    # Pester 6 format
```
andreasjordan: Maybe remove the comment.

.github/runners/tests/runner-policy.Tests.ps1
andreasjordan: Agreed, fixed in a7f6f67d0 — the version is gone rather than bumped.

tests/appveyor.pester.ps1
andreasjordan: Agreed, fixed in a7f6f67d0 — the version is gone rather than bumped.

tests/appveyor.pester.ps1
andreasjordan: Agreed, fixed in a7f6f67d0 — the version is gone rather than bumped.

potatoqualitee:
very nice thank you

PR #10414, opened 2026-07-10
VMSS CI fleet: replace AppVeyor with Azure VMSS self-hosted runners

potatoqualitee:
## What this PR does

Replaces AppVeyor with GitHub Actions on self-hosted Azure VMSS runners as the test suite of record. Everything lives on this branch — nothing CI-related was left on `development` — so merging this PR is the cutover.

**AppVeyor is already off**: the webhook is deactivated repo-side (last build was green), and this PR removes `appveyor.yml`. The `tests/appveyor.*.ps1` harness stays untouched — `ci-azure.yml` runs it as-is through `tests/gha.shim.ps1`, which supplies the `APPVEYOR_*` environment (including `APPVEYOR=True`, so every `if -not($env:APPVEYOR)` gate in the tests behaves exactly as before).

### The suite (`ci-azure.yml`)

The AppVeyor matrix, ported: 8 jobs across three SQL engine versions for diversity — SINGLE 1-3/3 on **sql2022**, MULTI on **sql2022+sql2017**, COPY on **sql2017+sql2022**, HADR and RESTART on **sql2019**, plus autodetect. Commit-message controls carry over: `(do Command)` scopes tests, `[skip ci]` skips. PR runs auto-scope to changed files; `workflow_dispatch` runs the **full suite** by default (a `message` input can narrow it).

### The fleet

- **`runner-scale-up.yml`** — event-driven scale-out + ephemeral runner registration (one job per factory-fresh VM, AppVeyor style)
- **`runner-reconcile.yml`** — the janitor: registers stragglers, reaps spent/zombie VMs, tops up to demand/floor, self-ticks while VMs exist, posts weekly cost to the "CI cost tracker" issue
- **`runner-boost.yml`** — maintainer pushes arm a 10-runner warm floor (`BOOST_*` repo variables) so matrix jobs start instantly during active sessions
- **`ps3-smoke.yml`** — nightly runnerless PowerShell 3.0 coverage on the legacy Server 2012 image; PSv3 users remain looked after
- **`.github/runners/`** — runbook, idempotent `infra.ps1`, scripted golden-image builds (Server 2022 + SQL 2017/2019/2022), `debug.ps1` fleet ops, phase-0 gate results

### Hardening that came out of burn-in

[...]

PR #10521, opened 2026-08-06
Docs - Point CI documentation at ci-azure instead of AppVeyor

andreasjordan:
AppVeyor the service was removed in #10414 and there is no `appveyor.yml` in the tree, but five documentation files still describe it as the live CI. This corrects them.

## What stays

The `tests/appveyor.*.ps1` harness is untouched. It is still the test suite of record — `.github/workflows/ci-azure.yml` drives it through `tests/gha.shim.ps1`, which supplies the AppVeyor build API (`Add-AppveyorTest`, `Push-AppveyorArtifact`, `Exit-AppveyorBuild`) and the `APPVEYOR_*` environment the scripts read. The names now mean "the CI harness", and the docs say so rather than leaving readers to guess.

Code comments naming AppVeyor are also left alone. The workarounds they describe (`Invoke-DbaQuery` instead of `ExecuteNonQuery` for long batches, and so on) are still in force, and `CLAUDE.md` requires preserving them.

## Corrections beyond the rename

Each of these was factually wrong, not just old-fashioned:

- **`tests/CLAUDE.md`** claimed a test using `InstanceCopy*`, `InstanceHadr` or `InstanceRestart` "runs on AppVeyor only". All three are `ci-azure` matrix scenarios. The narrower three-instance set belongs to the container-based `integration-tests.yml`, which is what `Get-TestConfig`'s `$env:GITHUB_WORKSPACE` branch configures. This one matters most — it is the test-policy file, it steers instance choice, and it was telling contributors those lanes had no coverage.
- **`CONTRIBUTING.md`** listed HADR as SQL 2022 (it is 2019), did not say RESTART is 2019, described SINGLE as unsplit (it is five parts), and still advertised a `2008R2SP2Express` scenario that no longer exists.
- **`.github/CONTRIBUTING-TESTING.md`** quoted a build matrix "from appveyor.yml" with SINGLE split three ways. It is five, and the file is gone.
- **`.github/copilot-instructions.md`** documented `appveyor.yml` as a configuration file and listed scenarios (`2008R2`, `2016`, `2016_2017`) that have not matched the matrix in years.
- **`.github/CODEOWNERS`** guarded `/appveyor.yml`, a path that no longer exists — so the CI harness had no owner at all. Replaced with `tests/appveyor.*.ps1`, `tests/gha.shim.ps1` and `tests/pester.groups.ps1`, which are the CI now and sit on the same supply-chain boundary the file's header describes.

## Not in this PR

Two things surfaced while mapping this out and are worth handling separately:

- `tests/appveyor.SQL2008R2SP2.ps1` and `tests/appveyor.SQL2016.ps1` are unreachable. `appveyor.sqlserver.ps1` dispatches by instance-name interpolation and no matrix entry names those instances. Holding off in case they are wanted for the legacy 2012 pool the `ci-azure.yml` header mentions.
- Three tests skip on `$env:APPVEYOR`, which the shim now always sets — so they skip permanently on the new fleet where they may well pass: `Get-DbaWindowsLog.Tests.ps1` (the whole `Describe`), `Add-DbaComputerCertificate.Tests.ps1` and `Enable-DbaFilestream.Tests.ps1`. Being checked against the current runners now.

Documentation and CODEOWNERS only; no code or test changes.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -5,4 +5,6 @@
 /.github/CODEOWNERS @potatoqualitee @niphlod @andreasjordan
 /.github/copilot-instructions.md @potatoqualitee @niphlod @andreasjordan
 /.github/dependabot.yml @potatoqualitee @niphlod @andreasjordan
-/appveyor.yml @potatoqualitee @niphlod @andreasjordan
+/tests/appveyor.*.ps1 @potatoqualitee @niphlod @andreasjordan
+/tests/gha.shim.ps1 @potatoqualitee @niphlod @andreasjordan
+/tests/pester.groups.ps1 @potatoqualitee @niphlod @andreasjordan
--- a/.github/CONTRIBUTING-TESTING.md
+++ b/.github/CONTRIBUTING-TESTING.md
@@ -446,18 +446,19 @@ git commit -m "Fix login handling (do Get-DbaLogin, Set-DbaLogin)"
 git commit -m "Update comments [skip ci]"
 ```
 
-### AppVeyor Build Matrix
+### CI Build Matrix
 
-The CI runs tests across multiple scenarios in parallel:
+The `ci-azure` workflow runs tests across multiple scenarios in parallel on self-hosted Azure runners:
 
 ```yaml
-# From appveyor.yml
+# From .github/workflows/ci-azure.yml
 matrix:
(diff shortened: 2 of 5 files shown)


potatoqualitee:
cool, no tests needed 👍🏼

PR #10522, opened 2026-08-06
Tests - Replace the $env:APPVEYOR skips with real preconditions

andreasjordan:
Follow-up to #10521.

AppVeyor the service is gone, but `tests/gha.shim.ps1` sets `$env:APPVEYOR` so the ported `appveyor.*.ps1` harness keeps working. A side effect: three skip conditions written for the old AppVeyor image were permanently true on the Azure fleet, silently skipping ever since the move, on runners the reasons were never assessed against.

Two CI runs on this branch settled what each one was really waiting for. **No skip in this file is gated on the CI provider any more** — each states the thing that is actually missing.

## Outcome

| Test | Lane | Outcome |
|---|---|---|
| `Add-DbaComputerCertificate.Tests.ps1` ("PFX certificate with chain") | default | Passed both runs — **skip removed** |
| `Get-DbaWindowsLog.Tests.ps1` | SINGLE | Intermittent — skip retained, on the missing startup event |
| `Enable-DbaFilestream.Tests.ps1` ("FileStream Level to 2") | RESTART | Fails consistently — parked on #10524 |

## Get-DbaWindowsLog

Not obsolete after all. It passed on one runner and then failed all three attempts on another, returning empty in under a second each time. A single green run was not evidence, and the older in-file note calling these "very unstable" was right.

That note asked for whoever revisited it to record *what* failed rather than only that it was unstable, so: the command locates the error log through event 17111, the SQL Server startup event in the Windows Application log, and `public/Get-DbaWindowsLog.ps1` returns early with nothing when that event is absent. On ephemeral CI VMs it is not reliably present.

The skip now tests for that event, read straight from the log rather than through the command — the same precondition, observed independently, so it is not circular. Verified on a live lab: the event is there, the test is **not** skipped, and it passes.

## Enable-DbaFilestream

Parked on #10524 rather than on any property of the environment, because the environment is no longer the open question.

The diagnostic added mid-PR paid for itself. The instance came back reporting:

```
InstanceAccess   : T-SQL access enabled
ServiceAccess    : FileStream enabled for T-Sql access
ServiceShareName : SQL2019
```

`SQL2019` is the instance-name fallback `Set-FileSystemSetting` uses when given no share name — the value the level 1 test leaves behind. The level 2 call passed `-ShareName TestShare` and the instance never saw it, so the call **changed nothing at all** rather than stalling halfway. The WMI return code that would say why is swallowed by #10524: `Get-FilestreamReturnValue` reports every code as success, and `Enable-DbaFilestream` only surfaces it when `-Force` is absent, which is not the path anything uses.

Two environmental explanations were tried and both are refuted. The comments record them so the work is not repeated:

- the RsFx filter driver is running even with FILESTREAM **fully disabled**, so its state says nothing about whether level 2 is reachable
- the Server service is running on the runners where this still fails

Guessing a third time without the return code is not worth a CI round trip. Once #10524 is fixed, unskipping this reads the real code straight away — the assertions keep their diagnostics for that.

## Note on policy

`tests/CLAUDE.md` prefers failing over skipping when a dependency is unprovisioned. Two skips survive here. Both name a specific, checkable condition and the filestream one names the issue that will remove it, which is the closest this can get until #10524 lands.

### DIFF
--- a/tests/Add-DbaComputerCertificate.Tests.ps1
+++ b/tests/Add-DbaComputerCertificate.Tests.ps1
@@ -119,7 +119,10 @@ Describe $CommandName -Tag IntegrationTests {
         }
     }
 
-    Context "PFX certificate with chain is imported properly" -Skip:($env:APPVEYOR) {
+    # The skip on this context recorded no reason at all. Unskipped on 2026-08-06 to find out
+    # whether one still applies: AppVeyor is gone, but tests\gha.shim.ps1 still sets
+    # $env:APPVEYOR, so the gate kept it skipped on the Azure fleet.
+    Context "PFX certificate with chain is imported properly" {
         BeforeAll {
             # Generate unique temp path for this test run
             $script:tempPath = "$($TestConfig.Temp)\$CommandName-$(Get-Random)"
--- a/tests/Enable-DbaFilestream.Tests.ps1
+++ b/tests/Enable-DbaFilestream.Tests.ps1
@@ -25,40 +25,6 @@ Describe $CommandName -Tag UnitTests {
 }
 
 Describe $CommandName -Tag IntegrationTests {
-    BeforeDiscovery {
-        # Level 1 is T-SQL access and needs nothing from Windows. Level 2 adds Win32 file I/O
-        # streaming, which SQL Server exposes through a Windows file share - so level 2 depends on
-        # the host being able to serve a share at all, which is the Server service (LanmanServer).
-        # That is an OS prerequisite, not a SQL Server one: FILESTREAM is a configuration setting,
-        # not an installable feature, so nothing about how the instance was installed decides this.
-        #
-        # This is the split seen on ci-azure on 2026-08-06, where level 1 passed and level 2 failed
-        # on all three attempts. Probe that prerequisite rather than the CI provider: the old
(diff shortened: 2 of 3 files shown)


potatoqualitee:
shit wait, @claude why didn't the tests trigger for andreas?

potatoqualitee:
ahh outage 🙄

potatoqualitee:
reinited and real failures unfortuantely

andreasjordan:
Currently working on this, will add commits today...

andreasjordan:
Filed #10524 for a defect found while diagnosing the filestream failure here: `Get-FilestreamReturnValue` in `Set-FileSystemSetting.ps1` reports every WMI return code as success, and `Enable-DbaFilestream` only surfaces that value when `-Force` is absent — so a failed WMI call produces no warning at all on the path everything actually uses.

That is why the failure in this PR showed up as a bare wrong level with no diagnostic. It does not change what this PR does, but it is the reason the first CI round trip told us so little. Keeping it separate since it is a command change, not a test one.

potatoqualitee:
Opus 5 is so verbose! Can't wait till they get that under control.

potatoqualitee:
Thank you 🙏🏼

PR #10525, opened 2026-08-07
Enable-DbaFilestream - Report what the WMI provider actually returned

### LINKED ISSUE #10524: Set-FileSystemSetting - Get-FilestreamReturnValue reports every WMI return code as success
## Summary

`Get-FilestreamReturnValue`, nested inside [`private/functions/Set-FileSystemSetting.ps1`](https://github.com/dataplat/dbatools/blob/development/private/functions/Set-FileSystemSetting.ps1#L45), has a `switch` clause whose condition is always true:

```powershell
{ 2147021885 -or 2147945411 -or 0 } {
    "The requested operation is successful. Changes will not be effective until the service is restarted."
}
```

The scriptblock ignores `$Value` entirely. `2147021885 -or 2147945411 -or 0` is just `$true`, so the clause matches **every** return code. It was presumably meant to be `{ $Value -in 2147021885, 2147945411, 0 }`.

## Effects

PowerShell's `switch` executes *all* matching clauses without a `break`, so this produces three distinct wrong behaviours. Verified by running the same switch shape in isolation:

| Return code | Actual output |
|---|---|
| `2147217396` ("Filestream not supported on instance") | `Filestream not supported on instance` **and** the success string |
| `2147024891` ("Access denied") | `Access denied` **and** the success string |
| `0` (genuine success) | the success string (correct) |
| any unlisted code | **only** the success string |

So:

1. **A recognised failure returns a two-element array**, the real message plus a contradictory success message.
2. **`default` is unreachable.** An unlisted return code can never fall through to `$Value`, so the raw code is lost and the caller is told it succeeded.
3. There is no way for a caller to distinguish success from failure by looking at the return value.

## Why it stays invisible

`Enable-DbaFilestream` is the only consumer, and it surfaces `$result` in exactly one place:

```powershell
if ($filestreamstate -ne $level -and -not $Force) {
    Write-Message -Level Warning -Message "[$instance] $result"
}
```

The warning is gated on `-not $Force`. Since `-Force` is the documented way to run the command non-interactively (it is in the examples, and every dbatools test uses it), the diagnostic is discarded on the common path. A failed WMI call therefore produces no warning, no error, and no clue — the command simply returns the unchanged state.

## How it surfaced

While re-checking the `$env:APPVEYOR`-gated tests (#10522), `Enable-DbaFilestream -FileStreamLevel 2 -ShareName TestShare -Force` on the ci-azure RESTART runner left the instance at level 1 on all three retry attempts, silently. Level 1 succeeded in the same run. Whatever the underlying environmental cause, the command had no way to report it, which is what made that failure expensive to diagnose.

This issue is only about the reporting defect. The environmental cause of that particular failure is being tracked separately on #10522.

## Suggested fix

- Make the condition test `$Value`, e.g. `{ $Value -in 2147021885, 2147945411, 0 }`, so `default` becomes reachable again and unknown codes surface the raw value.
- Give the caller something it can branch on rather than a free-text string, so `Enable-DbaFilestream` can `Stop-Function` on a real WMI failure instead of returning a stale state.
- Surface the failure regardless of `-Force`. `-Force` should mean "do not prompt", not "do not report errors".

A regression test wants a boundary that returns a non-zero code — an invalid or duplicate share name is the cheapest way to get a real one out of `EnableFilestream`.

## Environment

Found on `development`. The clause is long-standing and not version-specific.

andreasjordan:
Fixes #10524.

## The bug

`Get-FilestreamReturnValue` gated its success message on a constant expression:

```powershell
{ 2147021885 -or 2147945411 -or 0 } {
    "The requested operation is successful. ..."
}
```

`$Value` never appears in it. `2147021885 -or 2147945411 -or 0` is just `$true`, so the clause matched every code. And because `switch` runs *every* matching clause without a `break`, that produced three separate problems:

| Return code | Before | After |
|---|---|---|
| `2147024891` "Access denied" | `Access denied` **and** the success message | `Failure` / `Access denied` |
| any unlisted code | **only** the success message, raw value lost | `Unknown`, raw value kept |
| `0` / the two `21470…` codes | success | `Success` |

`default` was unreachable, so an unrecognized code could never fall through to report its own value.

`Enable-DbaFilestream` then discarded even that, surfacing the value only when `-Force` was absent — and `-Force` is the documented non-interactive path used by the examples and every test. A refused call produced no warning, no error, and an unchanged instance.

## How it showed up

On a ci-azure RESTART runner while working #10522, setting FILESTREAM level 2 left the instance at level 1 on every attempt with nothing reported. The instance came back with `ServiceShareName` still set to the instance-name fallback rather than the share name that was passed, so the call had changed nothing at all — and the return code that would have said why was gone.

## The change

`Get-FilestreamReturnValue` moves into its own file under `private/functions/` so it can be tested, and returns `ReturnValue`, `Category` and `Message`. `Category` is `Success`, `Failure` or `Unknown`.

Codes in neither list are reported as `Unknown` with the raw value rather than guessed at in either direction — assuming success hides a refusal, assuming failure invents an error the provider never reported. A null return value is `Unknown` for the same reason, since `switch` enters no clause at all for `$null`.

`Enable-DbaFilestream` now stops on a documented refusal and warns on an unrecognized code, **both regardless of `-Force`**. `-Force` should mean "do not prompt", not "do not report errors". The existing warning that changes need a service restart is unchanged.

## Testing

Five unit tests cover each category, plus two edge cases worth calling out:

- **a null return value** — `switch ($null)` enters no clause, so without an explicit guard this reported `Failure` for a call that never returned anything
- **numeric type** — `Invoke-CimMethod` hands back a `UInt32`, so a comparison that only matched `Int32` would push every real call down the unknown path

Against a live lab, all three filestream commands still pass: `Enable-DbaFilestream` 3 integration tests (levels 1 and 2 and the ShareName warning), `Disable-DbaFilestream` and `Get-DbaFilestream` unchanged. Repo compliance and structure gates: 0 failures.

## Note

This does not by itself fix the level 2 failure on ci-azure — it makes the provider's reason visible so that can finally be diagnosed. The level 2 test is skipped in #10522 with a pointer to #10524; once this merges it can be unskipped to read the real return code.

### DIFF
--- a/private/functions/Get-FilestreamReturnValue.ps1
+++ b/private/functions/Get-FilestreamReturnValue.ps1
@@ -1,0 +1,88 @@
+function Get-FilestreamReturnValue {
+    <#
+        .SYNOPSIS
+            Translates the return value of the SQL WMI EnableFilestream method into a result object.
+
+        .DESCRIPTION
+            The FilestreamSettings WMI class reports the outcome of EnableFilestream through a
+            numeric return value. This maps that number onto the documented meaning and says which
+            of three categories it falls into, so a caller can tell a refusal from a success
+            instead of having to match on message text.
+
+            Codes that are in neither list are reported as Unknown together with the raw value.
+            Guessing in either direction is what made the original version of this misleading: a
+            caller that assumes success ignores a refusal, and one that assumes failure invents an
+            error the provider never reported.
+
+        .PARAMETER Value
+            The ReturnValue the EnableFilestream method produced.
+
+        .NOTES
+            Author: the dbatools team + Claude
+    #>
+    [CmdletBinding()]
+    param (
+        [object]$Value
+    )
+
+    if ($null -eq $Value) {
+        return [PSCustomObject]@{
+            ReturnValue = $null
+            Category    = "Unknown"
+            Message     = "The EnableFilestream method returned no value"
+        }
+    }
+
+    # The provider signals success with 0 or one of these two codes. Everything named in the
+    # switch below is a documented refusal.
+    $successCodes = 2147021885, 2147945411, 0
+
+    # An assignment inside a switch clause produces no output, so $message collects only the
+    # strings while the category rides along as a side effect. The default is Failure because
+    # every literal clause below is a documented failure.
+    $category = "Failure"
+
+    $message = switch ($Value) {
+        2147217396 {
+            "Filestream not supported on instance"
+        }
+        2147217386 {
+            "Filestream cannot change share"
+        }
+        2147024713 {
+            "Duplicate sharename"
+        }
+        2147024891 {
+            "Access denied"
+        }
+        2147023681 {
+            "Invalid sharename"
+        }
+        2147024690 {
+            "Sharename too long"
+        }
+        2147019889 {
+            "Primary node not enabled "
+        }
+        2147019848 {
+            "Sharename node mismatch"
+        }
+        214721740 {
+            "General error"
+        }
+        { $PSItem -in $successCodes } {
+            $category = "Success"
+            "The requested operation is successful. Changes will not be effective until the service is restarted."
+        }
+        default {
+            $category = "Unknown"
+            "The EnableFilestream method returned an unrecognized value: $Value"
+        }
+    }
+
+    [PSCustomObject]@{
+        ReturnValue = $Value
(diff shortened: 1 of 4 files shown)


potatoqualitee:
I really like this, thank you

PR #10526, opened 2026-08-07
Import-DbaParquet - Report why the Parquet assemblies could not be loaded

andreasjordan:
The `catch` around `Add-Type` dropped the `LoaderExceptions` of a type load failure, so the warning always advised running `Install-DbaParquet`. That is misleading when the assemblies are installed and the real problem is a dependency that something else already loaded in another version.

That is not hypothetical: in a full local test run every test of this command failed with `Unable to load one or more of the requested types`, and the advice sent me looking for a missing install. The assemblies were present the whole time, and the same test file passes 17/17 on its own in a fresh session.

The inner exceptions are now unwrapped and reported, and the advice to install is only given when there is nothing more specific to say.

## Testing

`Import-DbaParquet.Tests.ps1` passes 17/17 against a local lab instance. The test file is unchanged, because this only changes what the command reports on a failure path that the tests do not reach.

### DIFF
--- a/public/Import-DbaParquet.ps1
+++ b/public/Import-DbaParquet.ps1
@@ -292,7 +292,21 @@ function Import-DbaParquet {
             try {
                 Add-Type -Path $parquetDllPath -ErrorAction Stop
             } catch {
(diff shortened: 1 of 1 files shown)


potatoqualitee:
Straightforward fix, thank you 🙏🏼

PR #10528, opened 2026-08-07
New-DbaFirewallRule - Stop blaming a disabled DAC for a cycled ERRORLOG

andreasjordan:
SQL Server writes the port of the dedicated admin connection to the ERRORLOG when the instance starts, and the command reads only the current ERRORLOG. On an instance whose log has been cycled since the last start that entry is gone, which looks exactly like a disabled DAC. The warning then said:

```
No information about the dedicated admin connection (DAC) found in ERRORLOG, cannot create
firewall rule for DAC. Use 'Set-DbaSpConfigure -SqlInstance 'SQL03\SQL2019' -Name
RemoteDacConnectionsEnabled -Value 1' to enable remote DAC and try again.
```

On the instance that produced it, `RemoteDacConnectionsEnabled` was already `1` for both the configured and the running value, so the advice was wrong and sent me to check a setting that was fine.

The warning now says what is actually missing, that the message is only written at startup, and names enabling remote DAC as one possible cause rather than the cause. Detection is unchanged.

## Tests

The integration setup creates rules for the whole instance, so it hits this warning on any instance whose ERRORLOG has been cycled. It is silenced there with `-WarningAction SilentlyContinue` and a comment, because it says nothing about the command under test, and a test run should not print warnings.

`New-DbaFirewallRule.Tests.ps1` passes 15/15 with 6 skipped and no warnings. The 6 skipped are the existing port-based context, which the file already skips when the instance uses dynamic ports.

Reading older ERRORLOG files would make the DAC rule work in this situation instead of only reporting it accurately. That is a behaviour change and is deliberately not part of this PR.

### DIFF
--- a/public/New-DbaFirewallRule.ps1
+++ b/public/New-DbaFirewallRule.ps1
@@ -428,7 +428,10 @@ function New-DbaFirewallRule {
                 }
 
                 if (-not $dacMessage) {
-                    Write-Message -Level Warning -Message "No information about the dedicated admin connection (DAC) found in ERRORLOG, cannot create firewall rule for DAC. Use 'Set-DbaSpConfigure -SqlInstance '$instance' -Name RemoteDacConnectionsEnabled -Value 1' to enable remote DAC and try again."
+                    # SQL Server writes this message when the instance starts, and we only read the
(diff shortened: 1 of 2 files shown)


potatoqualitee:
Nice, good update, thanks 💯

Issue #10529, opened 2026-08-07
Self-hosted runner image has no account lockout threshold, so login lockout cannot be tested


andreasjordan:
## What is missing

The self-hosted runner image sets no Windows account lockout threshold, so a SQL login can never be locked out on a CI runner. Nothing under `.github/runners/` configures one - there is no `net accounts`, no `secedit` and no `LockoutBadCount` anywhere in the image build.

Without a threshold, `LOGINPROPERTY(login, 'IsLocked')` stays `0` no matter how many logons fail, so no test can cover unlocking a login.

## Why this matters now

`tests/Set-DbaLogin.Tests.ps1` contains a test that locks a login out and unlocks it three different ways. It has been skipped for years behind `It -Skip "Unlock"` with a note saying it "makes assumptions the password policy configuration is enabled for the Windows OS". Enabling it found a real bug in `Set-DbaLogin`, which is fixed in a separate PR - but that PR has to leave the test enabled and failing on CI until the runners have a threshold, because a skipped test would hide the regression again.

## Which lane needs it

Only the self-hosted Azure lane. `Set-DbaLogin.Tests.ps1` uses `$TestConfig.InstanceSingle`, so it lands in the `SINGLE` scenario of `ci-azure.yml`, where the instance is local to the runner VM.

The other lanes are not affected and cannot be:

- `integration-tests.yml` (Linux) runs only `.github/scripts/gh-actions.ps1`, which does not touch `Set-DbaLogin`, and its instances are SQL Server on Linux in containers, which has no Windows account policy at all.
- The `windows-tests` job in the same workflow is disabled (`if: false`, since 2026-07-14).

## The change

One line in the image build, for example in `.github/runners/image-scripts/step2-sqlinstance.ps1`:

```powershell
# A SQL login can only be locked out when the host has an account lockout threshold.
$null = net accounts /lockoutthreshold:5 /lockoutduration:10 /lockoutwindow:10
```

Only the threshold is strictly required, and it has to be **5 or lower**, because the test fails exactly five logons. Duration and observation window only decide when Windows would release the lock by itself; the test unlocks explicitly, so they are there for determinism.

## Two things worth checking

1. **Sysprep.** `step3-sysprep.ps1` runs `sysprep /generalize`. Please confirm the local security policy survives generalize on the resulting image - if it does not, the setting belongs in a boot-time step such as `bootstrap-runner.ps1` instead.
[...]

PR #10530, opened 2026-08-07
Set-DbaLogin - Return the login as it is after unlocking it with a new password

andreasjordan:
> [!IMPORTANT]
> **One test in this PR will fail on CI, on purpose.** The runners have no account lockout threshold, so a login cannot be locked out there and `unlocks a login that was locked out` cannot pass. See #10529 for what has to change on the runner image. Skipping the test instead would hide the very regression this PR fixes, so it is left enabled and red until the runners are configured. Everything else in the file passes.

## The bug

Unlocking a login by setting a new password calls `Login.ChangePassword($NewSecurePassword, $Unlock, $PasswordMustChange)`, which really does clear the lock on the instance. But the login object was only refreshed when `-PasswordMustChange` was bound, so in every other case the object returned still reported `IsLocked` as `$true`:

```
locked before unlock (server): 1
returned object IsLocked : True
server says IsLocked     : 0
```

Anyone checking the result had to conclude the unlock had failed while the instance said it had worked. The object is refreshed in both cases now.

## Why nobody noticed

The test file was invisible. Its integration `Context` was skipped with a name of `"???"` and a `# TODO: Fix later`, and its `-TestCases` were built in a `BeforeAll`, which runs *after* discovery, so the cases were empty. Pester 5 silently produced no tests from an empty case list; Pester 6 fails the file during discovery instead, which is how this surfaced.

With the cases built in `BeforeDiscovery` and the `Context` named and enabled, the file runs **45 tests instead of 1**.

## The lockout tests needed a second fix

Two tests lock a login out by failing five logons. After a failed logon SqlClient blocks the connection pool for a growing number of seconds and answers the following attempts itself, so most of those logons never reached the instance. `BadPasswordCount` plateaued at 3 no matter how many attempts were made, even with the `Start-Sleep -s 5` the test already had, so the login was never locked.

They use `-NonPooledConnection` now. The bad password count then climbs one per attempt and the login locks exactly at the threshold, which also makes the sleeps unnecessary and the file faster.

## Testing

Against a lab whose host has an account lockout threshold of 5: **45 passed, 0 failed, 0 skipped, 0 warnings.**

The precondition is only the threshold. SQL Server reads it from the local policy of the host running the instance, not from the domain policy - verified in a lab where the domain has no threshold and the local one has five, and the lockout follows the local value.

## After #10529

Once the runners have a threshold, this test passes on CI unchanged. No follow-up change to this file is needed.

### DIFF
--- a/public/Set-DbaLogin.ps1
+++ b/public/Set-DbaLogin.ps1
@@ -504,9 +504,10 @@ function Set-DbaLogin {
                         $l.ChangePassword($NewSecurePassword, $Unlock, $PasswordMustChange)
                         $passwordChanged = $true
 
-                        if (Test-Bound PasswordMustChange) {
-                            $l.Refresh()  # necessary so that the read only property PasswordMustChange is updated
-                        }
+                        # necessary so that the read only properties PasswordMustChange and IsLocked are
+                        # updated. Changing the password with unlock clears the lock on the instance, but
+                        # without this the login we return still reports itself as locked.
+                        $l.Refresh()
                     } catch {
                         $notes += "Couldn't change password"
                         $passwordChanged = $false
--- a/tests/Set-DbaLogin.Tests.ps1
+++ b/tests/Set-DbaLogin.Tests.ps1
@@ -38,21 +38,25 @@ Describe $CommandName -Tag UnitTests {
 }
 
 Describe $CommandName -Tag IntegrationTests {
-    # TODO: Fix later
-    Context -Skip "???" {
-        BeforeAll {
+    Context "Parameter validation" {
+        BeforeDiscovery {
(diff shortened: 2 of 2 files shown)


potatoqualitee:
Genuine failure, unfortunately (isn't it nice that we mostly get genuine errors nowadays?)

potatoqualitee:
oh no! undo undo

potatoqualitee:
Merged by accident, lemme see how to fix

andreasjordan:
Just skip the one test again.

potatoqualitee:
i reverted it, looks like it'll need a new PR, sorry

andreasjordan:
No problem, I still have the local branch. Will open a new pr after the environment change.

PR #10531, opened 2026-08-07
Tests - Run without warnings and fix test cases that never ran

andreasjordan:
Found while running the full suite against a live lab. Three separate problems, all of which made tests quieter than they should have been.

## Test cases that were never built

Two files built their `-TestCases` inside a `BeforeAll`. That runs *after* discovery, so the variable was still empty when Pester expanded the cases and the tests did not exist:

| File | Tests before | after |
| --- | --- | --- |
| `Invoke-DbaDiagnosticQuery.Tests.ps1` | 7 | 19 |
| `Remove-DbaClientAlias.Tests.ps1` | 4 | 10 |

Pester 5 silently produced no tests from an empty case list. Pester 6 fails the file during discovery instead (`Value can not be null or empty array ... Parameter name: ForEach`), which is how these surfaced - as files that reported `Failed` with a `FailedCount` of `0`.

The cases are built in `BeforeDiscovery` now, as `tests/CLAUDE.md` already requires. An AST scan over every test file in the tree found only one other file with the same mistake, `Set-DbaLogin.Tests.ps1`, which is in #10530 because fixing it also needed a change to the command.

## A test that depended on a leftover

`Remove-DbaDbCertificate.Tests.ps1` created its certificate inside the `It` and without a name, so `New-DbaDbCertificate` defaulted the name to the database name and created a certificate called `master`. A leftover from an earlier run then made the command warn, return nothing, and the test failed on the empty result rather than on anything about the command.

It now uses a unique name, sets up in `BeforeAll` and cleans up in `AfterAll`.

That `AfterAll` also removes `EnableException` from `$PSDefaultParameterValues` again, which matters beyond this one file: **every test file gets the same dictionary object from `$TestConfig.Defaults`**, so a file that leaves `EnableException` behind makes every later file in the same process turn its warnings into terminating errors. I introduced exactly that bug while writing this fix and watched it break the next file alphabetically, so it is worth knowing about.

## Warnings that reached the stream

The remaining files expected warnings but let them print. A test runs without `EnableException`, so a problem inside a command surfaces only as a warning - and Pester ignores warnings completely. That makes an unexpected warning the only signal that something went wrong, and it is worthless while expected warnings are printing too.

Expected warnings are now silenced with `-WarningAction SilentlyContinue` and asserted on `$WarnVar` (or the test's own warning variable) instead. Where a test asserts that a command must *not* warn, the call is deliberately left unsilenced.

## Testing

All twelve files pass against a live lab with **zero warnings**:

`Get-DbaDbFile` 21, `Get-DbaDbFileGroup` 12, `Get-DbaDbLogSpace` 14 (+1 pre-existing skip), `Get-DbaDbSpace` 14, `Get-DbaDbVirtualLogFile` 7, `Get-DbaUnusedLogin` 19, `Invoke-DbaDbDataGenerator` 14, `Invoke-DbaDiagnosticQuery` 19, `Remove-DbaClientAlias` 10, `Remove-DbaDbCertificate` 2, `Set-DbaNetworkCertificate` 8, `Start-DbaAzMigration` 20.

No command is changed by this PR.

### DIFF
--- a/tests/Get-DbaDbFile.Tests.ps1
+++ b/tests/Get-DbaDbFile.Tests.ps1
@@ -246,7 +246,9 @@ Describe $CommandName -Tag IntegrationTests {
         }
 
         It "Warns and returns nothing when FileGroup is used on a database that cannot be opened" {
-            $filteredResults = Get-DbaDbFile -SqlInstance $TestConfig.InstanceSingle -Database $restoringDbName -FileGroup PRIMARY -WarningVariable fileGroupWarning
+            # The warning is what this test is about, so it is silenced on the stream and asserted on
+            # its own warning variable instead. A test run must not print warnings.
+            $filteredResults = Get-DbaDbFile -SqlInstance $TestConfig.InstanceSingle -Database $restoringDbName -FileGroup PRIMARY -WarningVariable fileGroupWarning -WarningAction SilentlyContinue
             $filteredResults | Should -BeNullOrEmpty
             $fileGroupWarning | Should -Match "FileGroup cannot be honored"
         }
--- a/tests/Get-DbaDbFileGroup.Tests.ps1
+++ b/tests/Get-DbaDbFileGroup.Tests.ps1
@@ -73,7 +73,10 @@ Describe $CommandName -Tag IntegrationTests {
 
     Context "Returns values for Instance" {
         BeforeAll {
-            $results = Get-DbaDbFileGroup -SqlInstance $TestConfig.InstanceSingle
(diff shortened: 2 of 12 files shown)


potatoqualitee:
Nice! How did you find these?

andreasjordan:
I'm using my own testing framework (https://github.com/andreasjordan/testing-dbatools) in my lab. Currently running another complete run with environment checks to find test leftovers. Will take some hours and maybe there will be more PRs this evening...

PR #10533, opened 2026-08-07
Query Store - Warn instead of silently skipping named system databases

### LINKED ISSUE #10523: Set-DbaDbQueryStoreOption should not be a silent no-op on master, tempdb, and model
### Summarize Functionality

As far as I can tell, `Set-DbaDbQueryStoreOption -SqlInstance $MyBox -Database master, tempdb, model -Verbose` is a no-op. It does not enable Query Store, print a warning, or error. It should do at least one of these.

For comparison, the rules in T-SQL are:
* Trying to enable Query Store on either master or tempdb is an error
* Trying to enable Query Store on model works.

Yes

### Technical Details

I think we just need to add a special case. This should only be a guard clause or two.

andreasjordan:
Fixes #10523.

## Problem

`Set-DbaDbQueryStoreOption -SqlInstance $x -Database master, tempdb, model` is a silent no-op: no output, no warning, no error. @ReeceGoding reported it and was right that it should do at least one of them.

## Mechanism

All three read commands appended the names to `-ExcludeDatabase` in `begin`:

```powershell
$ExcludeDatabase += 'master', 'tempdb', "model"
```

`Get-DbaDatabase` lets `-ExcludeDatabase` win over `-Database`, so a database named on purpose was dropped before anything could be emitted.

`Test-DbaDbQueryStore` failed differently and worse: it read the server off `$dbDatabases[0].Parent` *after* filtering, so `-Database master` died with `Cannot index into a null array.`

## What the lab says

Reproduced and verified against SQL Server 2019 (v15), 2022 (v16) and 2025 (v17):

| | 2019 and earlier | 2022 and later |
|---|---|---|
| master / tempdb | `ALTER` errors with *"Query Store cannot be enabled on system database master"* | same error |
| model | `ALTER` succeeds, but `sys.database_query_store_options` returns **zero rows** and SMO reads back nothing | Query Store **on by default** (`READ_WRITE`), reports like any other database |

So the `model` exclusion added in #6777 ("SMO cannot tell if Query Store is enabled there") was correct for 2019 and is now wrong: from SQL Server 2022 on, model is exactly how newly created databases get their Query Store defaults, and dbatools could neither read nor change it.

## What changed

A shared private helper `Get-QueryStoreUnsupportedDatabase` works the exclusion list out per instance and warns about anything the caller named explicitly:

```
WARNING: [Set-DbaDbQueryStoreOption] Query Store cannot be enabled on system database master. Skipping master on [SQL03\SQL2019].
WARNING: [Set-DbaDbQueryStoreOption] Query Store cannot be read on model before SQL Server 2022. Skipping model on [SQL03\SQL2019].
```

- `Get-DbaDbQueryStoreOption`, `Set-DbaDbQueryStoreOption`, `Test-DbaDbQueryStore` now warn instead of dropping silently.
- An explicit `-Database model` is honoured from SQL Server 2022 on; before that it warns. Piping a database object counts as naming it explicitly.
- `Test-DbaDbQueryStore` connects before filtering, which removes the null-array crash.

`Get-` had to change with `Set-`, not optionally: `Set-` returns its result by calling `Get-`, so fixing `Set-` alone would still emit nothing.

### Also fixed

`Test-DbaDbQueryStore` passed `$dbDatabases.name` to `Get-DbaDbQueryStoreOption`. An empty list makes that `-Database $null`, which means *every* database — so an over-filtered request quietly evaluated the whole instance instead of nothing. The crash had been hiding it.

- **model stays out of the automatic sweep**, even on SQL Server 2022+. `-AllDatabases` means the user databases, and changing Query Store on model changes the default for every database created afterwards. It is honoured only when named.
- **`Copy-DbaDbQueryStoreOption` is untouched.** Its destination uses `-ExcludeSystem` and it already reports `No matching databases found. Check the spelling and try again.`, so it is not silent. Its source-side flaw — a null `$SourceQSConfig` splatted into `Set-` — is a separate defect.

## Tests

New regression coverage in all three test files, version-branching so it asserts the real behaviour on whatever the runner provisions rather than skipping:

- master/tempdb warn and return nothing (`Get-`, `Set-`, `Test-`)
- model is read/configured on v16+ and warns below that (`Get-`, `Set-`); the `Set-` test restores model's original value in a `finally`
- model stays out of a sweep that does not name it (`Get-`)
- `Test-` no longer crashes on `-Database master`, and does not fall back to evaluating every database

All 29 tests across the four Query Store files pass against a live 2019/2022/2025 lab, with the lab left clean.

### DIFF
--- a/private/functions/Get-QueryStoreUnsupportedDatabase.ps1
+++ b/private/functions/Get-QueryStoreUnsupportedDatabase.ps1
@@ -1,0 +1,75 @@
+function Get-QueryStoreUnsupportedDatabase {
+    <#
+    .SYNOPSIS
+        Internal function. Returns the databases the Query Store commands have to skip on a given instance.
+
+    .DESCRIPTION
+        Query Store is never available on master or tempdb. SQL Server rejects the ALTER there with
+        "Cannot perform action because Query Store cannot be enabled on system database master." and
+        sys.database_query_store_options has no row for those databases on any version.
+
+        model depends on the version. Before SQL Server 2022 the ALTER is accepted but
+        sys.database_query_store_options stays empty and SMO reports nothing back, so there is no state to read
+        or verify. From SQL Server 2022 (v16) on, Query Store is enabled on model by default, which is how newly
+        created databases get it, and model reports its configuration like any other database.
+
+        Even where model works it stays out of the automatic sweep, because -AllDatabases means the user
+        databases and changing Query Store on model changes the default for every database created afterwards.
+        It is honoured only when the caller names it explicitly.
+
+        The Query Store commands used to append these names to -ExcludeDatabase unconditionally, and
+        Get-DbaDatabase lets -ExcludeDatabase win over -Database, so an explicit -Database master produced no
+        output, no warning and no error. Callers pass their -Database list in here so that a database the user
+        named on purpose is reported rather than silently dropped.
+
+    .PARAMETER SqlInstance
+        The connected server object, used to decide whether model is supported.
+
+    .PARAMETER Database
+        The database names the caller was asked to work on. Any name in this list that has to be skipped gets a
+        warning explaining why.
+
+    .PARAMETER FunctionName
+        The name of the calling command, so the warning is attributed to it instead of to this helper.
+
+    .EXAMPLE
+        PS C:\> Get-QueryStoreUnsupportedDatabase -SqlInstance $server -Database master, model
+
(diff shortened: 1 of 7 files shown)


potatoqualitee:
Thank you!

ReeceGoding:
Much more code than I expected. Thanks!

Does it help if I mention that [I know a trick for checking if Query Store is turned on for `model` before SQL Server 2022](https://dbfiddle.uk/jUmUvy3k)?

PR #10535, opened 2026-08-08
Query Store - Silence the expected warnings of the system database tests

andreasjordan:
Follow-up to #10533, found by a full lab run of the merged branch.

The integration tests added by #10533 assert on their own `-WarningVariable` but never set
`-WarningAction`, so the warnings they expect still reach the warning stream. The run ended with
five warnings - three from `Get-DbaDbQueryStoreOption`, two from `Set-DbaDbQueryStoreOption` -
that all had to be traced back by hand before they could be dismissed as expected. That is the
noise the warning-free rule exists to prevent: the tests run without `EnableException`, so an
unexpected warning is the only signal that something went wrong inside a command, and that signal
is worth nothing while expected warnings print next to it.

Four calls now carry `-WarningAction SilentlyContinue`. The assertions are unchanged - the
warnings are still captured in the same `-WarningVariable` and still checked.

The two calls that are asserted *not* to warn deliberately stay unsilenced, so a regression still
shows up as an unexpected warning in the run.

Verified against SQL Server 2019 and 2025: `Get-DbaDbQueryStoreOption` 4/4 and
`Set-DbaDbQueryStoreOption` 8/8, with no warnings on the stream.

### DIFF
--- a/tests/Get-DbaDbQueryStoreOption.Tests.ps1
+++ b/tests/Get-DbaDbQueryStoreOption.Tests.ps1
@@ -34,14 +34,16 @@ Describe $CommandName -Tag IntegrationTests {
 
     Context "When a system database is named explicitly" {
         It "Warns about master and tempdb instead of silently returning nothing" {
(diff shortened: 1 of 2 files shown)


potatoqualitee:
thank you kindly

Issue #10536, opened 2026-08-08
Import-DbaParquet followed by New-DbaDacPackage crashes the PowerShell process with an access violation


andreasjordan:
Using `Import-DbaParquet` and then `New-DbaDacPackage` in the same Windows PowerShell session kills the
session. The process dies with exit code **-1073741819 (0xC0000005, access violation)** - no error record,
no terminating exception, the host is simply gone. Reproduced four times on 2026-08-08.

This is not only a test problem. Importing a parquet file and later building a dacpac in one session is an
ordinary thing to do, and there is no message to act on when it happens.

## Reproducing

Run the two test files in one Windows PowerShell 5.1 process, in this order:

```powershell
Invoke-Pester -Path tests\Import-DbaParquet.Tests.ps1
Invoke-Pester -Path tests\New-DbaDacPackage.Tests.ps1   # process dies during this file
```

The crash comes after `Import-DbaParquet.Tests.ps1` has completed - the parquet tests themselves pass.

## What has already been ruled out

Each of these was tested, not assumed:

- **`New-DbaDacPackage` on its own.** Passes 9/9 in a fresh process.
- **Any preceding test file.** With `Get-DbaExternalProcess.Tests.ps1` in the same preceding position both
  files pass and nothing crashes, so it is not "two files in one process".
- **The AppDomain `AssemblyResolve` handler.** `Import-DbaParquet` registers a PowerShell **scriptblock**
  as a `[System.ResolveEventHandler]` (`public/Import-DbaParquet.ps1`, around line 270) and never removes
  it, which looked like the obvious suspect, since a scriptblock delegate invoked on a thread with no
  runspace cannot run. But a single `Import-DbaParquet` call followed by a dacpac build survives **both**
  with the handler registered and with it explicitly removed first. The whole test file is needed to
  trigger the crash, so the handler alone is not the cause.
- **A broken Parquet.NET installation.** The lab's parquet assemblies were the wrong ones for the edition
  (separate problem, fixed elsewhere). After fixing that, the parquet tests go from 2/17 to 17/17 - and the
  pair **still** crashes. So the crash is not a side effect of failing assembly loads.

## The related symptom in a long run

In a full 742-file suite run in one process, the same pair did not crash. `New-DbaDacPackage` instead
failed one test with a caught error:

```
Model validation failed: There is no Runspace available to run scripts in this thread.
```

That message is the real inner reason unwrapped out of the DacFx `AggregateException` by #10527. It points
[...]

PR #10538, opened 2026-08-08
Install-DbaParquet - Keep the assemblies of the two PowerShell editions apart

andreasjordan:
`Import-DbaParquet` failed **every** import on Windows PowerShell in the lab - 2 of 17 tests passing - while
the same file passed 17/17 on PowerShell 7. The warning said the assemblies could not be loaded:

```
Could not load file or assembly 'System.Collections, Version=10.0.0.0' ... |
Could not load file or assembly 'System.Runtime, Version=10.0.0.0' ...
```

## The cause

`Install-DbaParquet` picks the assemblies that match the runtime it is running on: .NET Framework targets on
Windows PowerShell, `net<major>` targets on PowerShell 7. That part works. But it installed all of them into
**one folder that both editions share**, `Path.DbatoolsParquet`, default `<AppData>\PowerShell\dbatools\parquet`.

So whichever edition installs last decides what the other one finds. This installation had been made from
PowerShell 7, which left `Parquet.dll` and six of its dependencies as the `net10.0` builds. Those reference
`System.Runtime` and `System.Collections` 10.0.0.0, which .NET Framework cannot bind, so Windows PowerShell
could not load them at all.

Nothing was wrong with the package versions, and `Parquet.dll` was exactly the pinned 5.4.0 - which is why
this reads like a broken or missing installation rather than a layout problem.

## The fix

Each edition installs into its own folder below the configured path, `desktop` or `core`, chosen by the new
private function `Get-DbaParquetEditionFolder`. `Get-DbaParquetPath` looks in the folder of the running
edition first and falls back to the flat folder, so installations made before this change keep working.

Reporting an installation as already present now looks at the folder that **would be written to** rather than
at any folder `Get-DbaParquetPath` knows about. Without that, the other edition's assemblies counted as
installed and `Install-DbaParquet` skipped the install - so the command told you to run `Install-DbaParquet`,
and `Install-DbaParquet` then said it was already installed, with no way out except deleting the folder by
hand.

`Import-DbaParquet` also names the likely cause now when the assemblies are present but cannot be loaded, and
the help of `Install-DbaParquet` says which assemblies it installs and that both editions need their own run.

An explicit `-Path` is still used exactly as given.

## Testing

In a lab with SQL Server 2019/2022/2025:

| | before | after |
| --- | --- | --- |
| `Import-DbaParquet.Tests.ps1` on Windows PowerShell 5.1 | 2 passed / 15 failed | **17 passed / 0 failed** |
| `Import-DbaParquet.Tests.ps1` on PowerShell 7.4 | 17 passed | **17 passed** (through the fallback) |
| `Install-DbaParquet.Tests.ps1` | 2 passed | **3 passed** |

The PowerShell 7 run deliberately used the pre-existing flat installation to confirm the fallback keeps old
installations working.

`Install-DbaParquet.Tests.ps1` gains a test that installing without `-Path` lands in a subfolder named for the
running edition.

> [!NOTE]
> Unrelated to this PR, but found next to it: `Import-DbaParquet` followed by `New-DbaDacPackage` in one
> Windows PowerShell process crashes the process with an access violation. That survives this fix and is
> tracked in #10536.

### DIFF
--- a/private/functions/Get-DbaParquetEditionFolder.ps1
+++ b/private/functions/Get-DbaParquetEditionFolder.ps1
@@ -1,0 +1,33 @@
+function Get-DbaParquetEditionFolder {
+    <#
+    .SYNOPSIS
+        Gets the name of the Parquet.NET installation subfolder for the running PowerShell edition.
+
+    .DESCRIPTION
+        Install-DbaParquet picks the Parquet.NET assemblies that match the runtime it is running on:
+        .NET Framework targets on Windows PowerShell and net<major> targets on PowerShell 7. Both
+        editions share one dbatools data directory, so the assemblies have to live in a folder per
+        edition. Without that, whichever edition installs last leaves assemblies the other one cannot
+        load, and every import fails there with a type load error that looks like a broken install.
+
+    .NOTES
+        Tags: Parquet, Import
+        Author: the dbatools team + Claude
+
+        Website: https://dbatools.io
+        Copyright: (c) 2026 by dbatools, licensed under MIT
+        License: MIT https://opensource.org/licenses/MIT
+
+    .OUTPUTS
+        System.String. The subfolder name, either "core" or "desktop".
+    #>
+    [CmdletBinding()]
+    param ()
+
+    # PSEdition does not exist on PowerShell 3 and 4, and those are always Desktop.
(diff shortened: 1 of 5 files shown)


potatoqualitee:
thank you 🙏🏼

PR #10544, opened 2026-08-08
Set-DbaPrivilege - Write secedit's working database to temp, not cwd

potatoqualitee:
## Summary

A user reported that `Set-DbaPrivilege` leaves `secedit.sdb` and `secedit.jfm` behind in the current working directory after it runs.

Root cause: the command calls `secedit /configure /cfg $tempfile /db secedit.sdb /areas USER_RIGHTS /overwrite /quiet`. `/db` takes a bare, relative filename, which `secedit.exe` resolves against the process's current directory - not `$env:TEMP` - so the working database (and its `.jfm` journal file) land wherever the caller's shell happened to be.

- `public/Set-DbaPrivilege.ps1`: `secedit`'s three working files - the exported `.cfg`, the `/db` database `.sdb`, and its `.jfm` journal - all now live under an absolute `$temp` path, named with a single `Get-Random` token generated once per invocation and threaded through the export, configure, and cleanup `Invoke-Command2` calls via splatted `-ArgumentList`. Tokenizing all three (not just the `.sdb`/`.jfm` pair) matters: two `Set-DbaPrivilege` runs against the same computer at the same time would otherwise still collide on, or delete, each other's shared `.cfg` file and apply the wrong privileges. Cleanup removes all three tokenized files alongside the working database.
- `tests/Set-DbaPrivilege.Tests.ps1`: added a real (non-mocked) `IntegrationTests` `Describe` block. A single combined test grants `CreateGlobalObjects`, watches temp with a `FileSystemWatcher` (via a bounded `Wait-Event`) to positively confirm the tokenized `.sdb` database is actually created under temp during the run, tracks the exact path(s) observed and confirms they're gone afterward, and separately checks the working directory for both the tokenized pattern and the legacy hardcoded `secedit.sdb`/`secedit.jfm` names (so a regression back to the pre-fix filename would still be caught). `AfterAll` reverts the granted privilege unless the user already held it, only doing so once `BeforeAll` has actually confirmed the prior state, validating `secedit`'s exit code, and cleaning up its own scratch files in a `finally` block.

This same bug also exists in the compiled C# port (`dbatools.library`, `SetDbaPrivilegeCommand.cs`) - see the companion PR at dataplat/dbatools.library.

### Follow-up fixes from automated review (2 rounds)

- Fixed the pre-existing `Invoke-Command2` mock in the `Set-DbaPrivilege regressions` unit test, which discriminated "set privileges" vs. "cleanup" calls by `$ArgumentList.Count`, and later derives the same per-run token dbatools now uses so its fabricated `.cfg` path matches what the real scriptblock reads/writes.
- Fixed a SID-removal regex in the integration test's `AfterAll` that could match a SID as a string-prefix of a different, longer SID sharing the same prefix.
- Splatted every `Invoke-Command2`/`Register-ObjectEvent`/`Get-ChildItem` call that grew to 3+ parameters, with purpose-specific variable names for the per-run token in each scope (`seceditRunToken`, `ExportRunToken`, `ConfigureRunToken`, `CleanupRunToken`) instead of reusing one name across scopes.
- Guarded the `AfterAll` revert on a `$preTestStateCaptured` flag so a `BeforeAll` failure can't be misread as "user didn't already have the privilege" and trigger an unwanted revert.
- Replaced an immediate post-call `Get-Event` with a bounded `Wait-Event`, since `FileSystemWatcher` delivers `Created` events asynchronously and a same-tick poll can race the event queue.

## Verification limitation

My local shell isn't elevated, and both `secedit.exe` and dbatools' own `Test-ElevationRequirement` check require Administrator rights for a localhost target. I confirmed:
- AST parse of both changed files is clean.
- The `UnitTests`-tagged tests (parameter validation + the `Mock`-based regression test) pass locally.
- The new `IntegrationTests` test fails locally, but only for the expected/honest reason: `Set-DbaPrivilege` throws `Console not elevated, but elevation is required...` via its own `Test-ElevationRequirement` check, and the `AfterAll` revert fails with `secedit /export failed with exit code 740` (`ERROR_ELEVATION_REQUIRED`) for the same reason. Neither the `FileSystemWatcher` assertion nor the `/db` path fix has been exercised end-to-end on my machine - that needs a run on the elevated self-hosted CI runner to be conclusive.

## Test plan

- [ ] CI runs `Set-DbaPrivilege` `IntegrationTests` on the elevated self-hosted runner and the new test passes
- [x] `UnitTests` pass locally
- [x] AST parse clean

### DIFF
--- a/public/Set-DbaPrivilege.ps1
+++ b/public/Set-DbaPrivilege.ps1
@@ -112,12 +112,16 @@ function Convert-UserNameToSID ([string] `$Acc ) {
                         }
                         if ($SQLServiceAccounts.count -ge 1) {
                             Write-Message -Level Verbose -Message "Setting Privileges on $Computer"
-                            Invoke-Command2 -Raw -ComputerName $computer -Credential $Credential -Verbose -ArgumentList $ResolveAccountToSID, $SQLServiceAccounts, $SQLPerServiceSIDs, $Type -ScriptBlock {
+                            # A random token keeps this invocation's secedit database from colliding with
+                            # (or being deleted by) another concurrent Set-DbaPrivilege run against the same computer.
+                            $dbToken = Get-Random
+                            Invoke-Command2 -Raw -ComputerName $computer -Credential $Credential -Verbose -ArgumentList $ResolveAccountToSID, $SQLServiceAccounts, $SQLPerServiceSIDs, $Type, $dbToken -ScriptBlock {
                                 [CmdletBinding()]
                                 param ($ResolveAccountToSID,
                                     $SQLServiceAccounts,
                                     $SQLPerServiceSIDs,
-                                    $Type
+                                    $Type,
+                                    $DbToken
                                 )
                                 . ([ScriptBlock]::Create($ResolveAccountToSID))
                                 $temp = ([System.IO.Path]::GetTempPath()).TrimEnd("");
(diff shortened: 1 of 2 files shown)


potatoqualitee:
Seems like this is ready

PR #10545, opened 2026-08-09
Save-DbaCommunitySoftware - Accept an array of Software values, add All

potatoqualitee:
## Summary
- `-Software` now accepts an array (`[string[]]`), so multiple community tools can be downloaded in one call, with native tab-completion preserved via `ValidateSet`.
- Added a new `All` value that expands to every supported tool.
- `Url`, `LocalFile` and `LocalDirectory` only describe a single tool, so they now warn and no-op instead of downloading when combined with more than one `Software` value (including `All`).
- Fixed a retry-loop bug where a fully failed download only continued the inner retry loop instead of moving on to the next software item, letting it fall through into `Expand-Archive` against a missing zip file.
- Fixed a stale default: `dba-multitool`'s GitHub default branch was renamed from `master` to `main` upstream; the hardcoded default here was never updated, so `-Software DbaMultiTool` (or `All`) failed without an explicit `-Branch main`. Found via the new real-download `All` test.
- Fixed a blind spot in `.claude/hooks/validate-style.ps1`: it tracked here-string state but had no equivalent for `<# #>` block comments (only `#` line comments), so it mechanically rejected single quotes inside comment-based help text. Added tracking that mirrors the existing here-string state machine.

## Test plan
- [x] Parameter validation tests updated for the array type and `All` value
- [x] New real-boundary integration test downloads multiple named tools in one call
- [x] New real-boundary integration test downloads every tool via `-Software All` and verifies each one landed
- [x] Existing `LocalDirectory` warning tests extended to cover the `All` case
- [x] Full Pester suite for this command run locally: 9 passed, 0 failed

### DIFF
--- a/.claude/hooks/validate-style.ps1
+++ b/.claude/hooks/validate-style.ps1
@@ -26,6 +26,7 @@ $lines = $content -split "`n"
 # Track state for multi-line constructs
 $inHereStringSingle = $false
 $inHereStringDouble = $false
+$inBlockComment = $false
 $inHashtable = $false
 $hashtableLines = @()
 $hashtableStart = 0
@@ -33,6 +34,8 @@ $misalignedHashtables = @()
 
 # Patterns (using double quotes with escaping)
 $patternComment = "^\s*#"
+$patternBlockCommentStart = "<#"
+$patternBlockCommentEnd = "#>"
 $patternHereStringSingleStart = "@'"
 $patternHereStringDoubleStart = "@`""
 $patternHereStringSingleEnd = "^'@"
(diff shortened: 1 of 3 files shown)


SQLDataCloud:
Would it be possible to have something for new SQL builds: `Install-DbaCommunitySoftware` ?

potatoqualitee:
sure, lez do it @SQLDataCloud

SQLDataCloud:
My PowerShell isn't brilliant, but happy to help on this where I can.  Super curious and geeky to know how a dbatools command is created from the ground up. 🤘🤓

For `Install-DbaCommunitySoftware`  would I have to rise a new PR?

potatoqualitee:
oh, sorry @SQLDataCloud! I didn't see this until I created and merged 🫠  it will be available in the next release

SQLDataCloud:
Thanks @potatoqualitee much appreciated. Let me know if you need any help with it - granted my PS is okay, not at your level, 😄

PR #10552, opened 2026-08-11
Export-DbaInstance - Script us_english custom errors before their translations

### LINKED ISSUE #10550: Export-DbaInstance - Custom messages export order can generate error in script

No, I did not search

Hello team,

When custom messages are created with multiple languages, scripting order looks like responsible for popping an execution error with generated script. 

Actually, SQL Server `master.dbo.sp_addmessage` procedure considers that, for an identical `@msgnum`, `us_english` language argument has to be processed prior to any other non-english one (`Français` in my case). 
[Source](https://learn.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sp-addmessage-transact-sql?view=sql-server-ver17#remarks)

Greetings for the work done.
Jean-Yves

### Steps to Reproduce

```sql
EXEC master.dbo.sp_addmessage @msgnum=50100, @lang=N'Français', 
		@severity=11, 
		@msgtext=N'Test (fr)', 
		@with_log=false
GO

EXEC master.dbo.sp_addmessage @msgnum=50100, @lang=N'us_english', 
		@severity=11, 
		@msgtext=N'Test (us)', 
		@with_log=false
GO
```

2.8.2

### Other details or mentions

_No response_

Windows PowerShell (powershell.exe), PowerShell Core (pwsh.exe)

### PowerShell Host Version

7.5.4

SQL Server Standard Developer 2025 RTM

### .NET Framework Version

.NET 9.0.10

andreasjordan:
Fixes #10550. Reported and correctly diagnosed by @tivivi63.

## Problem

`Export-DbaInstance` writes `customererrors.sql` with the localized version of a message before its `us_english` original, so the generated script cannot be replayed:

```
You must add the us_english version of this message before you can add the 'Francais' version.
```

## Mechanism

`sp_addmessage` only accepts a localized message once the `us_english` version of that message id exists. `Get-DbaCustomError` enumerates `$server.UserDefinedMessages`, which SMO returns sorted by message id and then by language *name*, so `Francais` comes out before `us_english`. `Export-DbaInstance` piped that straight into `Export-DbaScript`, preserving the wrong order.

Reproduced on the current `development` source against SQL Server 2019 (15.0.4430.1); the ordering comes from the SMO collection, not from the server version.

## What changed

`public/Export-DbaInstance.ps1` orders the messages before scripting them, the same way `Copy-DbaCustomError` already does:

```powershell
$allCustomErrors = @(Get-DbaCustomError -SqlInstance $server -EnableException:$EnableException)
$orderedCustomErrors = @($allCustomErrors | Where-Object Language -eq "us_english")
$orderedCustomErrors += $allCustomErrors | Where-Object Language -ne "us_english"
```

I swept every command that touches user defined messages:

| Command | Verdict |
| --- | --- |
| `Copy-DbaCustomError` | Already puts `us_english` first, unchanged |
| `Remove-DbaCustomError` | Already handles the mirror rule on drop, unchanged |
| `Start-DbaMigration`, `Sync-DbaAvailabilityGroup` | Both go through `Copy-DbaCustomError`, unchanged |
| `Get-DbaCustomError` | Still returns the plain SMO order. It is the only caller-visible ordering in the module and no other `Get-Dba*` command imposes one, so the fix belongs in the export |

## Tests

`tests/Export-DbaInstance.Tests.ps1` now creates a French translation for one of the custom errors it already sets up, and the new test exports, drops every custom error on the instance and replays the exported script.

The `AfterAll` block drops whatever custom errors are still present instead of two fixed ids, so a failed replay does not leave the instance dirty.

Run against a SQL Server 2019 instance:

* with the fix: 29 passed, 0 failed, 1 skipped (the policies test skips on pwsh)
* with the fix reverted: the new test fails with exactly the error from the issue

### DIFF
--- a/public/Export-DbaInstance.ps1
+++ b/public/Export-DbaInstance.ps1
@@ -339,7 +339,13 @@ function Export-DbaInstance {
                     if ($Exclude -notcontains 'CustomErrors') {
                         Write-Message -Level Verbose -Message "Exporting custom errors (user defined messages)"
                         Write-ProgressHelper -StepNumber ($stepCounter++) -Message "Exporting custom errors (user defined messages)"
-                        $null = Get-DbaCustomError -SqlInstance $server -EnableException:$EnableException | Export-DbaScript -FilePath "$exportPath\customererrors.sql" -BatchSeparator $BatchSeparator -ScriptingOptionsObject $ScriptingOption -NoPrefix:$NoPrefix -EnableException:$EnableException
+                        # sp_addmessage rejects a localized message whose us_english version does not exist yet,
+                        # and SMO returns the messages sorted by language name. So we script all us_english
+                        # messages first, the same way Copy-DbaCustomError orders them.
+                        $allCustomErrors = @(Get-DbaCustomError -SqlInstance $server -EnableException:$EnableException)
+                        $orderedCustomErrors = @($allCustomErrors | Where-Object Language -eq "us_english")
+                        $orderedCustomErrors += $allCustomErrors | Where-Object Language -ne "us_english"
+                        $null = $orderedCustomErrors | Export-DbaScript -FilePath "$exportPath\customererrors.sql" -BatchSeparator $BatchSeparator -ScriptingOptionsObject $ScriptingOption -NoPrefix:$NoPrefix -EnableException:$EnableException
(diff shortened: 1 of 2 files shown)


potatoqualitee:
awesome change, thank you 🙏🏼

PR #10553, opened 2026-08-11
Issue skill - Read the issue body and its comments in one call

andreasjordan:
Step 1 of the issue skill told the agent to read the issue with:

```
gh issue view <number> --comments
```

That view is not additive. The two forms are disjoint:

| Command | Body | Comments |
| --- | --- | --- |
| `gh issue view 10549` | 70 lines | none |
| `gh issue view 10549 --comments` | none | 36 lines |
| `gh issue view 10550 --comments` | none | issue has no comments, so **nothing at all** |

A fresh bug report normally has no comments yet, which is exactly when this skill runs. `gh` then prints nothing and exits 0. That looks like a broken `gh` rather than an empty comment list, and the natural reaction is to retry with `--repo`, then in another shell, then start ruling out pagers and encodings, none of which can matter. It cost a few minutes on #10550 before the `--json` fallback happened to work.

The fix is the `--json` form, which returns the body and the comments together in one call, plus a short note explaining the trap so nobody switches the command back:

```
gh issue view <number> --json number,title,state,author,body,labels,createdAt,comments
```

Verified against dataplat/dbatools#10549 (has comments) and #10550 (has none). No command code is touched, so there is nothing for CI to test here.

### DIFF
--- a/.claude/skills/issue/SKILL.md
+++ b/.claude/skills/issue/SKILL.md
@@ -13,8 +13,10 @@ The two failure modes this skill exists to prevent: **fixing something that `dev
 ## 1. Read the whole issue
 
 ```
-gh issue view <number> --comments
+gh issue view <number> --json number,title,state,author,body,labels,createdAt,comments
 ```
+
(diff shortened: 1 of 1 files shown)


potatoqualitee:
nice, thanks

Issue #10554, opened 2026-08-11
Commands disconnect SQL Server connections they do not own


andreasjordan:
Eight sites in `public/` and `private/` close a connection the caller owns. `Invoke-DbaQuery` is the one that started this and is described in full first; the complete inventory follows below, together with the places that already get it right.

The damage is invisible on a pooled connection - SMO transparently reopens it from the pool. It is real on anything session-scoped: `-NonPooledConnection`, a dedicated admin connection, temp tables, `SET` options, session context, `sp_getapplock`. It also makes the database-context leak in #10555 look unreproducible, because the accidental reconnect resets the current database.

## Invoke-DbaQuery

When `Invoke-DbaQuery` decides it cannot reuse the passed-in SMO server object, it builds its own connection with `NonPooledConnection = $true` and closes it afterwards:

```powershell
# if the given connection started out open, don't close it.
if ($connDbaInstanceParams.NonPooledConnection -and -not $startedWithAnOpenConnection) {
    $null = $server | Disconnect-DbaInstance -Verbose:$false
}
```

[`public/Invoke-DbaQuery.ps1:544-548`](https://github.com/dataplat/dbatools/blob/development/public/Invoke-DbaQuery.ps1#L544-L548)

The assumption behind the comment does not hold: `Connect-DbaInstance` frequently hands back **the same object that was passed in** instead of a new connection, because it only copies the context when something actually differs ([`public/Connect-DbaInstance.ps1:699-745`](https://github.com/dataplat/dbatools/blob/development/public/Connect-DbaInstance.ps1#L699-L745)). In that case `Invoke-DbaQuery` disconnects the caller's connection.

Two ways into it, both verified on SQL Server 2022 with `Connect-DbaInstance -NonPooledConnection`:

1. **`-Database` naming the database the connection is already on.** `$startedWithAnOpenConnection` compares `ConnectionContext.DatabaseName` ([`public/Invoke-DbaQuery.ps1:503`](https://github.com/dataplat/dbatools/blob/development/public/Invoke-DbaQuery.ps1#L503)), which is empty for a connection built by `Connect-DbaInstance` without `-Database`, while `Connect-DbaInstance` compares `ConnectionContext.CurrentDatabase` ([`public/Connect-DbaInstance.ps1:712`](https://github.com/dataplat/dbatools/blob/development/public/Connect-DbaInstance.ps1#L712)), which is `master`. The two disagree, so nothing is copied and the caller's connection is closed.
[...]

Issue #6210, opened 2019-11-19
[Bug] Invoke-DbaQuery doesn't close connections which it creates


Jaykul:
#### PowerShell version:

```
Name                           Value
----                           -----
PSVersion                      5.1.18362.145
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.18362.145
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```
#### dbatools Module version:
```
Name    : dbatools
Path    : C:\Users\joel\Documents\PowerShell\Modules\dbatools\1.0.72\dbatools.psd1
Version : 1.0.72
```

#### SQL Server: 

```
Microsoft SQL Azure (RTM) - 12.0.2000.8
   Oct 12 2019 22:46:48
   Copyright (C) 2019 Microsoft Corporation
```

### Report

As stated above, Invoke-DbaQuery creates connections but does not close them. At least when working against Azure SQL

Run any simple query against an Azure SQL database in a loop, **over 100 times**.

For what it's worth, the little I know: 
- the default Maximum size of an ADO.NET connection pool is 100
- I'm using SQL Azure. This bug doesn't seem to affect us when it's not Azure
- You can reproduce this just running `Connect-DbaInstance` in a loop if you don't call `.ConnectionContext.Disconnect()` on the connections it outputs.
- Adding the database name has no effect on _this_ bug.

#### Host used

- [x] powershell.exe
- [ ] ISE
- [ ] VS Code
- [ ] Other (please specify)

#### Errors Received

```

writeErrorStream      : True
PSMessageDetails      :
Exception             : System.Exception: Timeout expired.  The timeout period elapsed prior to obtaining a connection from the pool.  This may have occurred                           because all pooled connections were in use and max pool size was reached. ---> System.Exception: Timeout expired.  The timeout period                           elapsed prior to obtaining a connection from the pool.  This may have occurred because all pooled connections were in use and max pool                          size was reached. ---> System.Management.Automation.MethodInvocationException: Exception calling "Connect" with "0" argument(s):                                "Failed to connect to server TCP:ausc1-frg9sqlwttqrx.database.windows.net,1433." --->
                        Microsoft.SqlServer.Management.Common.ConnectionFailureException: Failed to connect to server
[...]

PR #10559, opened 2026-08-11
Connect-DbaInstance - Report whether a new connection was opened

andreasjordan:
First step of #10554. Adds the ownership signal and uses it in `Invoke-DbaQuery`; the other seven sites follow in their own pull requests.

## What was wrong

A command that connects and cleans up afterwards had no way to tell whether `Connect-DbaInstance` opened a connection for it or handed back the object the caller passed in. `Invoke-DbaQuery` inferred it from its own parameters:

```powershell
if ($connDbaInstanceParams.NonPooledConnection -and -not $startedWithAnOpenConnection) {
}
```

`Connect-DbaInstance` only copies the connection context when something about it actually has to change, so it frequently returns the very object that was passed in - and then this closed the caller's connection. With `-NonPooledConnection` that takes the session with it:

```powershell
# Invalid object name '#t'.
```

## What this changes

`Connect-DbaInstance` gets `-IsNewConnectionReference`, which reports whether it opened a new connection, and `Invoke-DbaQuery` only disconnects when it did.

The value is already known internally as `$isNewConnection`: `$false` for Server and SqlConnection inputs, `$true` for string, connection string and RegisteredServer inputs, and `$true` when a connection context has to be copied. Only its exposure is new; no connection logic changed. Both places that emit a server set it, including the `-SqlConnectionOnly` path.

## Why a reference and not a variable name

The plan in the issue proposed `-IsNewConnectionVariable` written through `$PSCmdlet.SessionState.PSVariable.Set()`. That works for a caller outside of dbatools, but not for a caller inside it, which is exactly the case that matters here. Both commands share the module session state, so the value lands in the local scope of `Connect-DbaInstance` instead of the scope of the calling command:

```
external caller, name : set via SessionState
in-module caller, name : unset
in-module caller, ref  : set via ref
```

`Set-Variable -Scope 1` does not help either - from a module function the parent scope is the module scope, not the caller. A `[ref]` behaves the same way in both cases, so that is what the parameter takes.

The first test run caught this: with the variable-name version, `Invoke-DbaQuery` never learned that it owned its connections and the new session-count test failed with five leaked sessions. That test is in this pull request.

## Tests

- `Connect-DbaInstance`: the parameter list, plus four assertions that the reference is `$true` for a string, `$false` when a server object is passed back in, `$true` when the context has to be copied, and that the same object comes back when nothing has to change
- `Invoke-DbaQuery`: the caller's connection stays open and its temp table survives, and the connections the command opens itself are still closed, so #6210 does not come back

Both files pass against SQL Server 2025 and 2022: `Connect-DbaInstance` 32 passed, 1 skipped (Azure), `Invoke-DbaQuery` 29 passed.

---

### DIFF
--- a/public/Connect-DbaInstance.ps1
+++ b/public/Connect-DbaInstance.ps1
@@ -157,6 +157,10 @@ function Connect-DbaInstance {
         Creates a dedicated administrator connection (DAC) for emergency access to SQL Server.
         Use this when SQL Server is unresponsive to regular connections, allowing you to diagnose and resolve critical issues. Remember to manually disconnect the connection when finished.
 
+    .PARAMETER IsNewConnectionReference
+        Reports whether a new connection was opened, by writing $true or $false into the referenced variable: pass a variable that already exists as [ref]$variable.
+        Use this when your command has to clean up after itself: only close a connection when this is $true, because a connection that was passed in belongs to the caller and closing it takes their session, their temp tables and their database context with it. When several instances are connected in one call, the value reflects the last connection that was returned.
+
     .PARAMETER DisableException
         Changes exception handling from throwing errors to displaying warnings.
         Use this in interactive sessions where you want graceful error handling instead of script-stopping exceptions, which is the default behavior for this command.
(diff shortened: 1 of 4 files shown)


andreasjordan:
One unrelated test failed:

```
Running C:\github\dbatools\tests\Install-DbaFirstResponderKit.Tests.ps1 ...[gha] test started: Install-DbaFirstResponderKit.Tests.ps1
WARNING: [06:05:26][Install-DbaFirstResponderKit] Certificate signing failed for dbatoolsci_frk_819873629 on dbatoolsTDZLC4\sql2022. | Cannot find the object 'sp_BlitzFirst', because it does not exist or you do not have permission.
```

Will try to analyze later.

andreasjordan:
Superseded by #10564, which combines this with #10563. On its own this change breaks `Install-DbaFirstResponderKit`: it removes a disconnect that was accidentally repairing a database context leak, and #10563 is what makes the repair unnecessary. Details in #10564.

Issue #10555, opened 2026-08-11
Database-scoped SMO calls silently change the current database of the shared connection


andreasjordan:
## Summary

`$db.Query(...)`, `$db.Invoke(...)`, `$db.ExecuteNonQuery(...)` and `$db.ExecuteWithResults(...)` on an SMO `Database` object do not run on a private connection. SMO's `Database.ExecutionManager.ConnectionContext` **is** the parent server's connection context, so these calls issue a `USE [database]` on the connection the caller owns and never switch back.

Every command that uses this pattern therefore hands the connection back to the caller pointing at a different database. Anything the caller runs afterwards through `$server.ConnectionContext` (or through any dbatools command that reuses that open connection) executes in the wrong database.

Proof that the execution manager is not a private connection:

```powershell
$server = Connect-DbaInstance -SqlInstance $instance -NonPooledConnection
$db = $server.Databases["SomeDatabase"]
[object]::ReferenceEquals($db.ExecutionManager.ConnectionContext, $server.ConnectionContext)
# True
```

## Steps to Reproduce

```powershell
$server = Connect-DbaInstance -SqlInstance $instance -NonPooledConnection
$server.ConnectionContext.ExecuteScalar("SELECT DB_NAME()")     # master

$null = $server.Databases["SomeDatabase"].ExecuteNonQuery("SELECT 1")
$server.ConnectionContext.ExecuteScalar("SELECT DB_NAME()")     # SomeDatabase  <-- leaked

$null = $server.Databases["SomeDatabase"].Query("SELECT 1")
$server.ConnectionContext.ExecuteScalar("SELECT DB_NAME()")     # SomeDatabase  <-- leaked
```

Verified on SQL Server 2019 and 2022, dbatools 2.8.4 (PowerShell Gallery, Windows PowerShell 5.1) and the current `development` build (PowerShell 7.6), with Windows authentication and with a `-SqlCredential` holding a Windows account, on pooled and non-pooled connections. `$db.Refresh()` beforehand makes no difference.

`$server.Query($sql, $database)` is affected as well: the two-argument overload in [`xml/dbatools.Types.ps1xml`](https://github.com/dataplat/dbatools/blob/development/xml/dbatools.Types.ps1xml) routes through `$this.Databases[$Database].ExecuteWithResults($Query)`, so it leaks exactly like the `Database` methods.

## Why this is only sometimes visible

Two things mask the leak, which is why it has survived this long:

[...]

PR #10563, opened 2026-08-12
Invoke-DbaQuery - Test the database the connection is on, not the one it was opened with

andreasjordan:
**Draft: needs more testing, and waits for #10559.**

## Why this waits

The change sends the connection-reuse case through `Connect-DbaInstance` in situations where it did not go there before. On `development` alone that would be unsafe, because the old disconnect logic then closes the connection of the caller - which is what #10559 fixes. So this is based on the branch of #10559 and only the second commit belongs here. GitHub retargets the pull request to `development` once #10559 is merged.

## What is wrong

To decide whether the connection that was passed in can be reused, the command compares `ConnectionContext.DatabaseName` with `-Database`. That property only holds the database the connection was *opened* with. `Connect-DbaInstance` asks the same question with `ConnectionContext.CurrentDatabase`, the database the connection is on *right now*.

The two differ as soon as anything runs a `USE` on the connection - which is exactly what the database-scoped SMO calls in #10555 do. The connection is then reused although it sits on another database, and the query silently runs in the wrong one:

```powershell
$server = Connect-DbaInstance -SqlInstance $instance -Database tempdb -NonPooledConnection
Invoke-DbaQuery -SqlInstance $server -Database tempdb -Query "SELECT DB_NAME() AS dbname"
```

```
after a stray USE [master]     : DatabaseName=[tempdb] CurrentDatabase=[master]
Invoke-DbaQuery -Database tempdb runs in: master
```

## Measured effect

Using `@@SPID` to see whether a connection was reused, on SQL Server 2022:

| Connection of the caller | `-Database <the database it is on>` before | after |
|---|---|---|
| non-pooled, open | reused, but through a needless trip into `Connect-DbaInstance` | reused directly |
| pooled, closed | new connection | new connection, unchanged - `CurrentDatabase` is empty while the connection is closed |

## Tests

Two assertions: the connection is still reused while it is on the requested database, and a query runs in the requested database after the connection was moved away from it. Reverting the one-line change makes the second one fail with `Expected: 'tempdb' But was: 'master'`, so it does test the change. With the change, `Invoke-DbaQuery` passes 31 of 31.

## What still needs testing before this leaves draft

- a full suite run, because `Invoke-DbaQuery` carries a large part of the module and the reuse path is used everywhere
- Azure SQL Database, where the database of a connection cannot be changed the same way
- availability group listeners together with `-ReadOnly`
- inputs that are not a Server object: `SqlConnection`, connection strings and registered servers
- the case where `CurrentDatabase` is empty because the connection has never been opened, which is unchanged behaviour but deserves a deliberate check rather than an assumption

---

### DIFF
--- a/public/Invoke-DbaQuery.ps1
+++ b/public/Invoke-DbaQuery.ps1
@@ -496,11 +496,17 @@ function Invoke-DbaQuery {
                 # Again, here we are, but we cannot use the connection when an information is lost, which is that we are:
                 #   - Integrated Security=True
                 #   - using ConnectAsUser, ConnectAsUserName, ConnectAsUserPassword to "log on as a different user"
+                #
+                # The database is tested with ConnectionContext.CurrentDatabase, the database the connection is on
+                # right now, and not with ConnectionContext.DatabaseName, which is only the database the connection
+                # was opened with. The two differ as soon as anything runs a USE, and then reusing the connection
+                # runs the query in the wrong database. Connect-DbaInstance tests CurrentDatabase as well, so both
+                # commands now ask the same question.
 
                 $startedWithAnOpenConnection = # we want to bypass Connect-DbaInstance if
                 ($instance.InputObject.GetType().Name -eq "Server") -and # we have Server SMO object and
                 (-not $ReadOnly) -and # no readonly intent is requested and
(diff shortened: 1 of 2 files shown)


andreasjordan:
Superseded by #10564, which combines this with #10559. This turned out to be a prerequisite rather than a cleanup: without it, #10559 breaks `Install-DbaFirstResponderKit`. Details in #10564.

PR #10565, opened 2026-08-13
Install-DbaCommunitySoftware - add unified community tooling installer (do Install-DbaCommunitySoftware)

potatoqualitee:
## What this does

Adds `Install-DbaCommunitySoftware`, the install-side counterpart to `Save-DbaCommunitySoftware`. That command already unifies the *download* step behind one `-Software` parameter; this does the same for installing, so you don't need to remember six separate command names.

```powershell
Install-DbaCommunitySoftware -SqlInstance sql2017 -Software All
Install-DbaCommunitySoftware -SqlInstance sql2017 -Software FirstResponderKit, WhoIsActive -Database DBAtools
```

| Software | Installer |
| --- | --- |
| MaintenanceSolution | `Install-DbaMaintenanceSolution` |
| FirstResponderKit | `Install-DbaFirstResponderKit` |
| DarlingData | `Install-DbaDarlingData` |
| SQLWATCH | `Install-DbaSqlWatch` |
| WhoIsActive | `Install-DbaWhoIsActive` |
| DbaMultiTool | `Install-DbaMultiTool` |

The `-Software` names match `Save-DbaCommunitySoftware` so the download and install steps read the same way.

## Design notes

- **Output is pass-through and unshaped.** The installers don't agree on a shape — `MaintenanceSolution` returns `Results` with no `Database`/`Name`/`Status`, `SQLWATCH` returns `DashboardPath` and no `Name`, and `WhoIsActive` carries a `Version`. All three variants are documented in `.OUTPUTS` rather than normalized away.
- **Forwarding is derived, not hardcoded.** Each target's real parameter list decides what gets passed, so `-Branch` reaches only the three tools that have one and a warning names any selection it was dropped from.
- **`-Database` and `-Branch` forward only when explicitly bound**, so each installer keeps its own default otherwise — notably SqlWatch, which defaults to the `SQLWATCH` database and would be silently relocated by an unconditional `-Database master`.
- **`ShouldProcess` is declared but not called at this level**, so `-WhatIf`/`-Confirm` flow down and prompt once instead of twice.
- **AzSqlTips is excluded on purpose** — `Save-DbaCommunitySoftware` downloads it, but `Invoke-DbaDbAzSqlTip` consumes it as a query rather than installing procedures.

Two behaviors deliberately differ from calling an installer yourself, both documented in the help:

1. WhoIsActive is given `master` when `-Database` is omitted. Called directly with no database it opens an interactive `Show-DbaDbList` picker, which would stall an unattended run.
2. A failure against one tool warns and continues rather than ending the batch.

## Testing

18 tests, run against SQL Server 2022. **18 passed, 0 failed, 0 skipped.**

Integration coverage spans multi-tool dispatch, single-tool pass-through, continuation past an unreachable instance (a real unreachable host ordered ahead of a live one — no mocks), `-Branch` warning, and `-WhatIf` asserting the side effect did *not* happen.

Two edge cases are covered because they were live bugs during development:

- `-Database ""` reached `Install-DbaWhoIsActive`'s own `if (-not $Database)` test and opened the interactive picker. Now rejected at binding.
- `Select-Object -Unique` compares case-sensitively on **both** 5.1 and 7, while `ValidateSet` accepts any casing — so `-Software WhoIsActive, whoisactive` installed twice and falsely tripped the single-tool `-LocalFile` guard. Deduplication is now case-insensitive and order-preserving.

The WhoIsActive-defaults-to-master test never touches a `master.dbo.sp_WhoisActive` it did not create; it installs only when the procedure is absent and reports itself skipped otherwise, since an encrypted procedure has no recoverable definition and permissions do not survive a drop.

Verified under both PowerShell 5.1 and 7. `Invoke-DbatoolsFormatter` and PSScriptAnalyzer are clean.

### DIFF
--- a/dbatools.psd1
+++ b/dbatools.psd1
@@ -435,6 +435,7 @@
         'Import-DbaSpConfigure',
         'Import-DbatoolsConfig',
         'Import-DbaXESessionTemplate',
+        'Install-DbaCommunitySoftware',
         'Install-DbaDarlingData',
         'Install-DbaFirstResponderKit',
         'Install-DbaInstance',
--- a/dbatools.psm1
+++ b/dbatools.psm1
@@ -879,6 +879,7 @@ $script:xplat = @(
         'Export-DbaServerRole',
         'Get-DbaBuild',
         'Update-DbaBuildReference',
+        'Install-DbaCommunitySoftware',
         'Install-DbaFirstResponderKit',
         'Install-DbaSqlPackage',
         'Install-DbaWhoIsActive',
--- a/public/Install-DbaCommunitySoftware.ps1
+++ b/public/Install-DbaCommunitySoftware.ps1
@@ -20,7 +20,9 @@ function Install-DbaCommunitySoftware {
         - SQLWATCH is skipped with a warning on PowerShell Core, and the rest of the batch still runs. Install-DbaSqlWatch supports Windows PowerShell only, and left to itself it downloads its payload before reaching that check, so selecting All from PowerShell Core would fetch a file it can never use.
 
         - WhoIsActive is given master when you do not pass Database. Called directly with no database, Install-DbaWhoIsActive opens an interactive picker, which would stall an unattended run.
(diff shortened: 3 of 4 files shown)


potatoqualitee:
Mayyybe in 3.0, I'll make all the sub commands internal 🤔

PR #10567, opened 2026-08-13
Get-DbaDbExtentDiff - Leave the connection of the caller alone

andreasjordan:
Second of the sites listed in #10554, after #10564 brought the mechanism.

## The disconnect

The command opened its connection with `-NonPooledConnection` and closed it at the end, without asking whether it had opened it. When a caller passes their own server object, `Connect-DbaInstance` hands the same object back and the command then closed the connection of the caller, which takes the session with it. It now only closes what it opened itself, via `-IsNewConnectionReference`.

## The leak the disconnect was hiding

This is the part worth a second look. The extents query ran *in* the examined database:

```powershell
$DBCCPageResults = $server.Query($DBCCPageQueryDMV, $db.Name)
```

That two argument form routes through `$server.Databases[$Database].ExecuteWithResults()`, and SMO leaves the connection in that database - the defect in #10555. Until now the disconnect covered it up: the connection was closed right afterwards, and the reconnect landed back on the database from the connection string. Removing the disconnect without touching the query would have left the connection of the caller in whichever database was examined last.

So the query now names the database instead of running in it:

```sql
FROM [$dbNameEscaped].sys.dm_db_file_space_usage
```

Verified that this returns the same numbers as the context switch, per database:

```
master   three-part=71     with context switch=71
msdb     three-part=297    with context switch=297
tempdb   three-part=512    with context switch=512
```

The branch for SQL Server 2016 before SP2 needs no change: it queries `master.sys.master_files` and runs `DBCC PAGE` with the database as an argument, neither of which moves the connection.

## Tests

Two assertions on a connection the caller opened: the session survives the call, and the connection is still on the database it was on. Each one covers one of the two changes - reverting the disconnect guard fails the first, reverting the query change fails the second with `Expected: 'master' But was: 'dbatoolsci_test_1751040644'`.

`Get-DbaDbExtentDiff` passes 9 of 9 against SQL Server 2019.

---

### DIFF
--- a/public/Get-DbaDbExtentDiff.ps1
+++ b/public/Get-DbaDbExtentDiff.ps1
@@ -105,8 +105,17 @@ function Get-DbaDbExtentDiff {
     process {
 
         foreach ($instance in $SqlInstance) {
+            # Connect-DbaInstance tells us whether it opened a connection for us. We must only close what we opened
+            # ourselves, because closing a connection of the caller takes their session with it. See #10554.
+            $isNewConnection = $false
+            $splatConnect = @{
+                SqlInstance              = $instance
+                SqlCredential            = $SqlCredential
+                NonPooledConnection      = $true
+                IsNewConnectionReference = [ref]$isNewConnection
+            }
             try {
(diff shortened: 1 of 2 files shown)


potatoqualitee:
I digt it, thanks

PR #10568, opened 2026-08-13
Update-DbaMaintenanceSolution - Update the procedures in the database that was named and leave the connection of the caller alone

### LINKED ISSUE #10569: Update-DbaMaintenanceSolution - Updates the procedures in the current database of the connection, not in -Database
`Update-DbaMaintenanceSolution` updates the procedures of the maintenance solution in whatever database the connection happens to be on, and it is the right database today only because another command leaks its context.

## What happens

The command executes the update script without naming a database ([`Update-DbaMaintenanceSolution.ps1:171`](https://github.com/dataplat/dbatools/blob/development/public/Update-DbaMaintenanceSolution.ps1#L171)):

```powershell
$null = Invoke-DbaQuery -SqlInstance $server -File $file
```

The scripts of the maintenance solution carry no `USE` statement - `Install-DbaMaintenanceSolution` rewrites the one in `MaintenanceSolution.sql` for exactly that reason - so the target is the current database of the connection.

That happens to be `$Database`, because the line above calls `Get-DbaModule`, which runs `$server.Query($sql, $db.name)` ([`Get-DbaModule.ps1:199`](https://github.com/dataplat/dbatools/blob/development/public/Get-DbaModule.ps1#L199)) and SMO leaves the connection in that database - the defect in #10555.

## Steps to Reproduce

```powershell
$server.ConnectionContext.ExecuteScalar("SELECT DB_NAME()")

$null = Get-DbaModule -SqlInstance $server -Database $dbName
$server.ConnectionContext.ExecuteScalar("SELECT DB_NAME()")

# what line 171 does: no database is named
$null = Invoke-DbaQuery -SqlInstance $server -File $someScript
```

```
connected                     : DB_NAME()=master
after Get-DbaModule -Database : DB_NAME()=dbatoolsci_olactx
procedure landed in dbatoolsci_olactx     : 1
procedure landed in master                : 0
```

Verified on SQL Server 2019 with current `development`.

## Why this matters

Fixing `Get-DbaModule` - one of the sites of #10555 - turns this into a silent data problem: `Update-DbaMaintenanceSolution -Database DBA` would update the procedures in `master`, report `Updated` for every one of them, and leave the ones in `DBA` untouched. Nothing fails, nothing warns.

## Suggested fix

Name the database on the call that does the work:

```powershell
$null = Invoke-DbaQuery -SqlInstance $server -Database $Database -File $file
```

Since #10564, `Invoke-DbaQuery` sets the context of its own connection for that, so this is independent of whatever the shared connection is doing.

## Test

A regression test needs the leak out of the way to have any teeth, so this is best fixed together with `Get-DbaModule`, or with a test that connects to a different database on purpose and asserts where the procedure ended up.

---

andreasjordan:
Third of the sites listed in #10554, and it fixes #10569 as well - the two belong together, see below.

## The disconnect

The command closed its connection when it was done, without asking whether it had opened it. When a caller passes their own server object, `Connect-DbaInstance` hands the same object back, and the command then closed the connection of the caller, taking the session with it. It now only closes what it opened itself.

## The database the update runs in

The command executed the update script without naming a database:

```powershell
$null = Invoke-DbaQuery -SqlInstance $server -File $file
```

The scripts of the maintenance solution carry no `USE` statement, so the procedures were updated in whatever database the connection happened to be on. That was the right one only by accident: `Get-DbaModule` on the line above runs `$server.Query($sql, $db.name)`, and SMO leaves the connection in the database it queried - #10555.

```
connected                     : DB_NAME()=master
after Get-DbaModule -Database : DB_NAME()=dbatoolsci_olactx
procedure landed in dbatoolsci_olactx     : 1
procedure landed in master                : 0
```

Two things follow from that. The obvious one: the moment `Get-DbaModule` stops leaking, this command starts updating `master` and reports `Updated` while doing it. The less obvious one: as long as the leak is what makes it work, no test of the target database can fail, so the fix cannot be covered.

Both halves are explicit now. The installed procedures are read with a query that names the database - the command only ever used the four names - and the update names the database as well. `Get-DbaModule` is untouched: its query uses `DB_NAME()` and unqualified catalog views, so fixing it belongs to that command.

## Tests

A context with its own database, called through a connection that sits on `master`:

- the procedure is reported as updated
- its definition in the named database is Ola's, not the stub that was there before
- `master` has no `CommandExecute` afterwards that it did not have before
- the session of the caller survives
- the connection of the caller is still on `master`

Reverting the `-Database` on the update fails two of them:

```
updates the procedure in the database that was named
  Expected regular expression '@DatabaseContext' to match 'CREATE PROCEDURE dbo.CommandExecute AS SELECT 1', but it did not match.
does not create the procedure in the database the connection is on
  Expected $false, but got $true.
```

and reverting the disconnect guard fails the session test. `Update-DbaMaintenanceSolution` passes 7 of 7 against SQL Server 2019.

---

### DIFF
--- a/public/Update-DbaMaintenanceSolution.ps1
+++ b/public/Update-DbaMaintenanceSolution.ps1
@@ -128,8 +128,17 @@ function Update-DbaMaintenanceSolution {
         }
 
         foreach ($instance in $SqlInstance) {
+            # Connect-DbaInstance tells us whether it opened a connection for us. We must only close what we opened
+            # ourselves, because closing a connection of the caller takes their session with it. See #10554.
+            $isNewConnection = $false
+            $splatConnect = @{
+                SqlInstance              = $instance
+                SqlCredential            = $SqlCredential
+                NonPooledConnection      = $true
+                IsNewConnectionReference = [ref]$isNewConnection
+            }
             try {
-                $server = Connect-DbaInstance -SqlInstance $instance -SqlCredential $SqlCredential -NonPooledConnection
+                $server = Connect-DbaInstance @splatConnect
             } catch {
                 Stop-Function -Message "Failure" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
             }
@@ -181,8 +190,10 @@ function Update-DbaMaintenanceSolution {
                 }
             }
 
-            # Close non-pooled connection as this is not done automatically. If it is a reused Server SMO, connection will be opened again automatically on next request.
-            $null = $server | Disconnect-DbaInstance
+            if ($isNewConnection) {
+                # Close non-pooled connection as this is not done automatically.
+                $null = $server | Disconnect-DbaInstance
+            }
         }
     }
 }
--- a/public/Update-DbaMaintenanceSolution.ps1
+++ b/public/Update-DbaMaintenanceSolution.ps1
@@ -148,7 +148,15 @@ function Update-DbaMaintenanceSolution {
                 Stop-Function -Message "Database $Database not found on $instance. Skipping." -Target $instance -Continue
             }
 
-            $installedProcedures = Get-DbaModule -SqlInstance $server -Database $Database | Where-Object Name -in 'CommandExecute', 'DatabaseBackup', 'DatabaseIntegrityCheck', 'IndexOptimize'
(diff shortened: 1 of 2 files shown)


potatoqualitee:
thank you 🙏🏼

PR #10570, opened 2026-08-13
Install-DbaMaintenanceSolution - Leave the connection of the caller alone

andreasjordan:
Fourth of the sites listed in #10554.

## The disconnect

The command closed its connection when it was done, without asking whether it had opened it. When a caller passes their own server object, `Connect-DbaInstance` hands the same object back, and the command then closed the connection of the caller, taking the session with it. It now only closes what it opened itself.

## The leak the disconnect was hiding

With `-ReplaceExisting`, the existing objects are dropped through `$db.Invoke($cleanupQuery)`, which runs on the connection of the caller and leaves it in that database - #10555. The disconnect covered that up, because the connection was closed right afterwards.

The installation of the scripts a few lines below already names the database:

```powershell
$null = Invoke-DbaQuery -SqlInstance $server -Database $Database -Query $sql -EnableException
```

so the cleanup now does the same. Same statements, same database, and the connection of the caller is untouched.

## Tests

A new context installs with `-ReplaceExisting` through a connection the caller opened, and asserts that the session survives and that the connection is still on `master`. Each assertion covers one of the two changes, verified by putting each one back:

```
=== branch as committed (both fixes) ===
session survived: True
DB_NAME()       : master

=== with $db.Invoke() restored ===
session survived: True
DB_NAME()       : tempdb

=== with the disconnect guard removed ===
session survived: False
DB_NAME()       : master
```

Jobs are deliberately left out of that context (`-InstallJobs` not set), because the SQL Agent commands move the connection themselves - `New-DbaAgentJob` and `Remove-DbaAgentJob` are their own sites of #10555 and are not touched here.

`Install-DbaMaintenanceSolution` passes 52 of 52 against SQL Server 2025 and 2022, in 20 minutes.

---

### DIFF
--- a/public/Install-DbaMaintenanceSolution.ps1
+++ b/public/Install-DbaMaintenanceSolution.ps1
@@ -456,8 +456,17 @@ function Install-DbaMaintenanceSolution {
         }
 
         foreach ($instance in $SqlInstance) {
+            # Connect-DbaInstance tells us whether it opened a connection for us. We must only close what we opened
+            # ourselves, because closing a connection of the caller takes their session with it. See #10554.
+            $isNewConnection = $false
+            $splatConnect = @{
+                SqlInstance              = $instance
+                SqlCredential            = $SqlCredential
+                NonPooledConnection      = $true
+                IsNewConnectionReference = [ref]$isNewConnection
+            }
             try {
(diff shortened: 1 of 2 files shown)


potatoqualitee:
thank you 🙏🏼

PR #10577, opened 2026-08-14
Registered server commands - Leave the connection of the caller alone

### LINKED ISSUE #10572: Registered server commands disconnect the connection of the caller, including the Get- commands
Seven sites in the registered server commands disconnect a connection they did not open. They share one mechanism, so they need one fix rather than seven, and they were missed in the inventory in #10554.

## What happens

`Get-DbaRegServerStore` builds the store on a connection from `Connect-DbaInstance` ([`Get-DbaRegServerStore.ps1:76`](https://github.com/dataplat/dbatools/blob/development/public/Get-DbaRegServerStore.ps1#L76)), which hands back the server object of the caller when one was passed in. The commands then close that connection when they are done:

| Site | Statement |
|---|---|
| [`Get-DbaRegServerGroup.ps1:181`](https://github.com/dataplat/dbatools/blob/development/public/Get-DbaRegServerGroup.ps1#L181) | `$serverstore.ServerConnection.Disconnect()` |
| [`Get-DbaRegServer.ps1:222`](https://github.com/dataplat/dbatools/blob/development/public/Get-DbaRegServer.ps1#L222) | `$serverstore.ServerConnection.Disconnect()` |
| [`Add-DbaRegServerGroup.ps1:155`](https://github.com/dataplat/dbatools/blob/development/public/Add-DbaRegServerGroup.ps1#L155) | `$currentInstance.ConnectionContext.Disconnect()` |
| [`Move-DbaRegServerGroup.ps1:133`](https://github.com/dataplat/dbatools/blob/development/public/Move-DbaRegServerGroup.ps1#L133) | `$parentserver.ServerConnection.Disconnect()` |
| [`Move-DbaRegServer.ps1:129`](https://github.com/dataplat/dbatools/blob/development/public/Move-DbaRegServer.ps1#L129) | `$parentserver.ServerConnection.Disconnect()` |
| [`Remove-DbaRegServerGroup.ps1:118`](https://github.com/dataplat/dbatools/blob/development/public/Remove-DbaRegServerGroup.ps1#L118) | `$parentserver.ServerConnection.Disconnect()` |
| [`Remove-DbaRegServer.ps1:134`](https://github.com/dataplat/dbatools/blob/development/public/Remove-DbaRegServer.ps1#L134) | `Disconnect-RegServer -Server $regserver.Parent` |

The private helper `Disconnect-RegServer` walks up the parents until it finds a `ServerConnection` and disconnects that.

## Steps to Reproduce

```powershell
$null = $server.ConnectionContext.ExecuteNonQuery("CREATE TABLE #marker (id INT)")

$null = Get-DbaRegServerGroup -SqlInstance $server -Id 1

$server.ConnectionContext.ExecuteScalar("SELECT COUNT(*) FROM #marker")
```

Measured on SQL Server 2019 with current `development`, a temp table as the marker and `Get-DbaDatabase` as a control:

```
Get-DbaRegServerGroup                      session survived: False
Get-DbaRegServer                           session survived: False
Add-DbaRegServerGroup                      session survived: False
Remove-DbaRegServerGroup                   session survived: False
Get-DbaDatabase (control)                  session survived: True
```

Two `Get-` commands are in that list. Reading registered servers ends the session of the connection that was handed in - temp tables, `SET` options, session context, all gone, and the connection is silently reopened afterwards so nothing looks wrong.

## Why one fix and not seven

Removing the disconnect in `Add-DbaRegServerGroup` alone changes nothing, because it calls `Get-DbaRegServerGroup`, which disconnects as well. Anything short of a decision that covers the whole family leaves the symptom in place. That is what makes this different from the individual sites in #10554.

The decision is the same one as in #10564: only close what this module opened. `Get-DbaRegServerStore` is where the connection is obtained, so it is the natural place to know the answer - it can ask `Connect-DbaInstance` with `-IsNewConnectionReference` and pass that on, and `Disconnect-RegServer` becomes the single place that acts on it.

## Test

Per `tests/CLAUDE.md`, one regression test per fixed command in the shape used for #10554: connect with `-NonPooledConnection`, create a temp table, call the command with that server object, and assert the temp table is still there afterwards.

---

andreasjordan:
## Type of Change
 - [x] Bug fix (non-breaking change, fixes #10572)
 - [x] Pester test is included

### Purpose

Seven sites in the registered server commands closed a connection they did not open. **Two of them are `Get-` commands**, so merely reading registered servers ended the session of the connection that was handed in - temp tables, `SET` options, session context - and the connection was silently reopened afterwards, so nothing looked wrong.

Measured on SQL Server 2019 with a temp table as the marker and `Get-DbaDatabase` as a control:

```
Get-DbaRegServerGroup                      session survived: False
Get-DbaRegServer                           session survived: False
Add-DbaRegServerGroup                      session survived: False
Move-DbaRegServerGroup                     session survived: False
Move-DbaRegServer                          session survived: False
Remove-DbaRegServer                        session survived: False
Remove-DbaRegServerGroup                   session survived: False
Get-DbaRegServerStore (control)            session survived: True
Get-DbaDatabase (control)                  session survived: True
```

This is the family that the inventory in #10554 missed, because the grep behind it matched `Disconnect-DbaInstance` and `ConnectionContext.Disconnect()` but not the generic `.Disconnect()` these use.

### Approach

One fix, not seven. Removing the disconnect in `Add-DbaRegServerGroup` alone would change nothing, because it calls `Get-DbaRegServerGroup`, which disconnects as well - anything short of a decision covering the whole family leaves the symptom in place.

`Get-DbaRegServerStore` is where the connection is obtained, so it is where the answer is known. It asks `Connect-DbaInstance` through `IsNewConnectionReference` and records the result as `IsNewConnection` on the store.

A note property is safe **here**, unlike on a server object: the store is built fresh on every call and is never handed back to `Connect-DbaInstance`, so nothing can re-stamp it. That was the objection to a note property in #10554, and it does not apply to this object.

`Disconnect-Regserver` becomes the single place that acts on it. It already walked up the parents to find the connection; it now also accepts the store itself, and closes the connection only when the store says we opened it:

```powershell
$i = 0
while ($null -ne $Server -and $null -eq $Server.ServerConnection -and $i++ -le 20) {
    $Server = $Server.Parent
}

if ($Server.ServerConnection -and $Server.IsNewConnection) {
    $Server.ServerConnection.Disconnect()
}
```

The six sites that closed the connection by hand now go through it too, so there is one place left in the family that can close a connection.

### Commands to test

```powershell
$null = $server.ConnectionContext.ExecuteNonQuery("CREATE TABLE #marker (id INT)")

$null = Get-DbaRegServerGroup -SqlInstance $server -Id 1

$server.ConnectionContext.ExecuteScalar("SELECT COUNT(*) FROM #marker")   # 0, was: Invalid object name '#marker'
```

### Tests

One regression context per fixed command, all on `InstanceSingle`, in the shape used for #10554: connect with `-NonPooledConnection`, create a temp table, call the command with that server object, assert the command still did its work and the temp table is still there. A pooled connection passes either way, so the tests have to be non-pooled.

`Get-DbaRegServerStore.Tests.ps1` covers the mechanism itself: `IsNewConnection` is `$true` for an instance name and `$false` for a connection of the caller.

For the commands that take an `-InputObject`, the marker is created **after** the lookup on purpose. `Get-DbaRegServer` and `Get-DbaRegServerGroup` close the connection too, so a marker created before the lookup would measure them instead of the command under test.

53 tests across the eight files, all passing. Against `development` every one of the eight files fails on exactly its new assertion.

### DIFF
--- a/private/functions/Disconnect-Regserver.ps1
+++ b/private/functions/Disconnect-Regserver.ps1
@@ -1,8 +1,22 @@
 function Disconnect-Regserver ($Server) {
+    <#
+    .SYNOPSIS
+        Internal function. Closes the connection behind a registered server object, but only if we opened it.
+
+    .DESCRIPTION
+        Takes any object of the registered server tree - a store, a group or a registered server - and walks up
+        to the RegisteredServersStore, which is the object that holds the connection.
+
+        The connection is only closed when Get-DbaRegServerStore opened it itself, which it records as
+        IsNewConnection on the store. A connection that was handed in belongs to the caller, and closing it takes
+        their session, their temp tables and their database context with it. See #10572.
+    #>
     $i = 0
-    do { $server = $server.Parent }
-    until ($null -ne $server.ServerConnection -or $i++ -gt 20)
-    if ($server.ServerConnection) {
-        $server.ServerConnection.Disconnect()
+    while ($null -ne $Server -and $null -eq $Server.ServerConnection -and $i++ -le 20) {
+        $Server = $Server.Parent
+    }
+
+    if ($Server.ServerConnection -and $Server.IsNewConnection) {
+        $Server.ServerConnection.Disconnect()
     }
 }
--- a/public/Add-DbaRegServerGroup.ps1
+++ b/public/Add-DbaRegServerGroup.ps1
@@ -151,9 +151,7 @@ function Add-DbaRegServerGroup {
                     $newgroup.Alter()
 
                     Get-DbaRegServerGroup -SqlInstance $currentInstance -Group (Get-RegServerGroupReverseParse -object $newgroup)
-                    if ($currentInstance.ConnectionContext) {
-                        $currentInstance.ConnectionContext.Disconnect()
-                    }
+                    Disconnect-RegServer -Server $newgroup
                 } catch {
                     Stop-Function -Message "Failed to add $reggroup" -ErrorRecord $_ -Continue
                 }
--- a/public/Get-DbaRegServer.ps1
+++ b/public/Get-DbaRegServer.ps1
@@ -219,7 +219,7 @@ function Get-DbaRegServer {
                 }
             } else {
                 $servers += ($serverstore.DatabaseEngineServerGroup.GetDescendantRegisteredServers())
-                $serverstore.ServerConnection.Disconnect()
+                Disconnect-RegServer -Server $serverstore
             }
 
             # save the $serverstore for later usage
--- a/public/Get-DbaRegServerGroup.ps1
+++ b/public/Get-DbaRegServerGroup.ps1
@@ -177,9 +177,7 @@ function Get-DbaRegServerGroup {
                     $groups = $serverstore.DatabaseEngineServerGroup.GetDescendantRegisteredServers().Parent | Where-Object Id -In $Id
                 }
             }
-            if ($serverstore.ServerConnection) {
-                $serverstore.ServerConnection.Disconnect()
-            }
+            Disconnect-RegServer -Server $serverstore
 
             foreach ($groupobject in $groups) {
                 Add-Member -Force -InputObject $groupobject -MemberType NoteProperty -Name ComputerName -Value $serverstore.ComputerName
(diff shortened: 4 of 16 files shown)


potatoqualitee:
thank you 🙏🏼

Issue #10556, opened 2026-08-11
Invoke-DbaDbUpgrade - Leaves the connection pointing at the upgraded database


andreasjordan:
## Summary

`Invoke-DbaDbUpgrade` runs `DBCC CHECKDB`, `DBCC UPDATEUSAGE`, `sp_updatestats` and `sp_refreshview` through `$db.ExecuteNonQuery(...)`:

- [`public/Invoke-DbaDbUpgrade.ps1:236`](https://github.com/dataplat/dbatools/blob/development/public/Invoke-DbaDbUpgrade.ps1#L236) (`DBCC CHECKDB ... WITH DATA_PURITY`)
- [`public/Invoke-DbaDbUpgrade.ps1:252`](https://github.com/dataplat/dbatools/blob/development/public/Invoke-DbaDbUpgrade.ps1#L252) (`DBCC UPDATEUSAGE`)
- [`public/Invoke-DbaDbUpgrade.ps1:269`](https://github.com/dataplat/dbatools/blob/development/public/Invoke-DbaDbUpgrade.ps1#L269) (`sp_updatestats`)
- [`public/Invoke-DbaDbUpgrade.ps1:294`](https://github.com/dataplat/dbatools/blob/development/public/Invoke-DbaDbUpgrade.ps1#L294) (`sp_refreshview`)

Those calls run a `USE [database]` on the caller's connection and never switch back, so when the command returns, the connection that was passed in points at the upgraded database instead of the database it was connected to. See #10555 for the mechanism.

## Steps to Reproduce

The database must actually need upgrading. With compatibility level and target recovery time already matching the instance, the command skips the database entirely at [`public/Invoke-DbaDbUpgrade.ps1:191-197`](https://github.com/dataplat/dbatools/blob/development/public/Invoke-DbaDbUpgrade.ps1#L191-L197) and nothing leaks - which is why a freshly created test database shows nothing.

```powershell
$server = Connect-DbaInstance -SqlInstance $instance -NonPooledConnection
$null = New-DbaDatabase -SqlInstance $server -Name dbatoolsci_upgrade
$null = Invoke-DbaQuery -SqlInstance $instance -Database master -Query "ALTER DATABASE [dbatoolsci_upgrade] SET COMPATIBILITY_LEVEL = 100"
$null = $server.ConnectionContext.ExecuteNonQuery("USE [master]")

$server.ConnectionContext.ExecuteScalar("SELECT DB_NAME()")     # master
[...]

PR #10579, opened 2026-08-14
Database context - Put the database back in the Query and Invoke script methods

### LINKED ISSUE #10555: Database-scoped SMO calls silently change the current database of the shared connection

andreasjordan:
## Type of Change
 - [x] Bug fix (non-breaking change, fixes #10555 in part)
 - [x] Pester test is included

## Overview

*Added 2026-08-23, after two rounds of review added to what the original description below covers. Nothing was removed - the original text follows unchanged.*

### The problem

The `Query` and `Invoke` script methods on `Server` and `Database` move the caller's connection into another database and leave it there.

```powershell
$server = Connect-DbaInstance -SqlInstance $instance      # connection is in master
$server.Query("SELECT 1", "msdb")
$server.ConnectionContext.ExecuteScalar("SELECT DB_NAME()")   # msdb, not master
```

The execution manager of a database object is the connection context of the parent server, which belongs to the caller - not a private connection. SMO issues a `USE` to run the statement and never switches back, so everything that runs on that connection afterwards silently runs in the wrong database. Four methods do this: `Database.Query`, `Database.Invoke`, and the `-Database` overloads of `Server.Query` and `Server.Invoke`.

### What this pull request changes

Each of the four remembers the current database and puts it back afterwards. The rest of the change is about the ways that goes wrong.

1. **The comparison that decides whether to restore is case sensitive.** Database names use the collation of the instance, so on a case sensitive instance `AppDb` and `appdb` are two different databases, and a case insensitive `-ne` reported them as equal and skipped the restore - leaving the caller in the wrong database, which is the very thing this fixes. It cannot restore needlessly: both sides are read from the same property, so one database always spells itself the same way.

2. **The name is escaped**, so a database with a `]` in its name cannot break the statement.

3. **Restoring never becomes the outcome of the call.** A statement that makes the previous database unreachable - taking it offline, dropping it, renaming it, revoking access - succeeds, and then the `USE` that would put the context back cannot work. Throwing there would tell the caller "it did not run" about a statement that ran, and a caller that retries would run it twice. Where the statement itself failed, the restore error would replace its error. So the restore warns instead of throwing.

4. **The warning cannot terminate the call either.** `Write-Warning` obeys the caller's `$WarningPreference`, and `Stop` turns it into a terminating error - which would hand the outcome back to the housekeeping after all. `Inquire` does the same where nothing can answer the prompt. Both are neutralised for that one warning. Suppression is deliberately left alone, so a caller who asked for silence still gets silence.

### Evidence

- `InModule.TypeExtensions.Tests.ps1` against a case **insensitive** instance: 13 tests, 12 pass, 1 skips
- The same file against a case **sensitive** instance: **13 pass, 0 skip**
- Every new test verified to fail against the old code, with the specific error it guards
- 16 test files of commands that use these wrappers: 186 tests, 184 pass, 2 skip, 0 fail
- CI: 20 pass, 1 skipping, 0 failures

### Two things worth knowing, because they shaped the tests

- **A warning written inside a ps1xml script method goes straight to the host.** It can be suppressed by the caller's preference, but it cannot be captured - not with `-WarningVariable`, not with `3>&1`, which returns zero objects. So the tests assert on behaviour, never on warning text.
- **Two databases whose names differ only in case need distinct file names.** NTFS is case insensitive, so the derived `.mdf` and then `.ldf` collide. `New-DbaDatabase` cannot express that, so the fixture creates the second database under a name of its own and renames it.

### What is covered where

The case sensitive scenario cannot be built at all on a case insensitive instance, so it skips there - which is every CI instance. Whether the instance is case sensitive is asked of the instance as a behaviour, not matched against the collation name, because `_BIN` and `_BIN2` are case sensitive too and carry no `_CS_`. The lab has no binary collation instance, so that case is handled by construction rather than demonstrated.

---

### Purpose

Step one of #10555: the mechanism, on its own.

The `Query` and `Invoke` script methods of `Server` and `Database` in [`xml/dbatools.Types.ps1xml`](https://github.com/dataplat/dbatools/blob/development/xml/dbatools.Types.ps1xml) do not run on a private connection. The execution manager of an SMO database **is** the connection context of the parent server, which belongs to the caller, so they issue a `USE` and never switch back:

```
call                                       result    DB_NAME() afterwards
$db.Query()                                LEAKED    dbatoolsci_wrap
$db.Invoke()                               LEAKED    dbatoolsci_wrap
$server.Query(sql) - one argument          ok        master
$server.Query(sql, db) - two arguments     LEAKED    dbatoolsci_wrap
$server.Invoke(sql, db) - two arguments    LEAKED    dbatoolsci_wrap
```

These four methods are reached from roughly 68 database-scoped `.Query()` / `.Invoke()` call sites plus 45 two-argument `$server.Query($sql, $db)` calls across 28 files, so this one file is the cheapest place in the module to fix it.

### Approach

Each method remembers `ConnectionContext.CurrentDatabase` and puts it back in a `finally`, so a query that throws restores the context as well.

**The `Server` pair needs its own copy of that.** `Server.Query` and `Server.Invoke` call `$this.Databases[$Database].ExecuteWithResults(...)` directly and never go through the `Database` methods, so fixing `Database.Query` alone does not reach them. That is four edits, not two - the issue body assumed otherwise.

**Restoring, not copying.** `ConnectionContext.Copy().GetDatabaseConnection($name)` also works, and was the other candidate, but it is a different session:

```
=== copied connection context ===
    query ran in          : dbatoolsci_ctx  on SPID 64      <-- caller is on SPID 62
    copy sees temp table  : no (separate session)

=== remember and put back ===
    query ran in          : dbatoolsci_ctx  on SPID 62
    caller temp table     : still there
```

A copy cannot see the temp tables or `SET` options of the caller and opens a connection per call, which would be a silent behaviour change for any command that builds session state and then queries through the wrapper. Restoring keeps the session and costs one round trip, and only when the database actually moved.

**The caller's database is restored, not master.** A connection sitting in `msdb` is returned to `msdb`. Restoring to master would have passed every other test and still been wrong - and it matters, because the Agent commands move the context to `msdb` rather than `master`.

The database name is escaped for the `USE`, so a database containing `]` in its name is handled.

### Commands to test

```powershell
$server.ConnectionContext.ExecuteScalar("SELECT DB_NAME()")   # master, was: SomeDatabase
```

### Tests

`tests\InModule.TypeExtensions.Tests.ps1`, following the existing `InModule.*` naming for test files that are not the test of a single command. 10 tests on `InstanceSingle`:

- each of the four methods leaves the database context alone
- the query still runs in the database that was asked for, and `AllTables` still returns every table
- the session is kept, so a temporary object created before the call is still there afterwards
- a failing query still puts the database back
- a caller connected to `msdb` is returned to `msdb`, not to master

All 10 pass. Against `development` 7 of them fail; the 3 that pass are the correctness assertions, which are there to catch the fix breaking something rather than to prove the bug.

Because this reaches every command that uses the wrappers, 15 further test files of wrapper-using commands were run on top: `Find-DbaSimilarTable`, `Get-DbaCpuRingBuffer`, `Get-DbaDatabase`, `Get-DbaDbFeatureUsage`, `Get-DbaDbFile`, `Get-DbaDbSnapshot`, `Get-DbaDbVirtualLogFile`, `Get-DbaHelpIndex`, `Get-DbaInstanceInstallDate`, `Get-DbaModule`, `Get-DbaSchemaChangeHistory`, `Install-DbaWhoIsActive`, `Invoke-DbaDbClone`, `New-DbaLinkedServer`, `Set-DbaDbFileGrowth`. 104 tests, no failures.

### What this does not fix

Only the script methods. The other two sources in #10555 are untouched and still leak:

- the 35 direct `$db.ExecuteNonQuery(...)` / `$db.ExecuteWithResults(...)` call sites, which are SMO's own methods and cannot be shadowed
- SMO's own `Create()` and `Drop()` of server-level objects

`Invoke-DbaDbUpgrade` (#10556) is in the first of those groups. Verified against a database forced to compatibility level 100 so the upgrade really ran - it went to 150 and the connection was still left in the upgraded database.

### DIFF
--- a/tests/InModule.TypeExtensions.Tests.ps1
+++ b/tests/InModule.TypeExtensions.Tests.ps1
@@ -15,7 +15,14 @@ Describe $CommandName -Tag IntegrationTests {
         # caught. The collation decides it, not the version, so this is asked of the instance rather than
         # assumed. On a case insensitive instance the scenario cannot be built at all and the Context skips.
         $caseDiscoveryServer = Connect-DbaInstance -SqlInstance $TestConfig.InstanceSingle
-        $instanceIsCaseSensitive = $caseDiscoveryServer.Collation -match "_CS_"
+        # Asked as a behaviour rather than matched against the collation name. The _BIN and _BIN2
+        # collations are case sensitive too and carry no _CS_, so a name match would skip this whole
+        # scenario on a binary collation instance. Comparing a name in sys.databases asks the exact
+        # question instead, because that column carries the collation of the instance - the one that
+        # database names are compared with. DB_NAME(1) is master, so the comparison needs no string
+        # literal of its own.
+        $caseProbeQuery = "SELECT CASE WHEN EXISTS (SELECT 1 FROM sys.databases WHERE name = UPPER(DB_NAME(1))) THEN 0 ELSE 1 END"
+        $instanceIsCaseSensitive = [bool]$caseDiscoveryServer.ConnectionContext.ExecuteScalar($caseProbeQuery)
         $null = $caseDiscoveryServer | Disconnect-DbaInstance
     }
 
@@ -142,9 +149,17 @@ Describe $CommandName -Tag IntegrationTests {
         BeforeAll {
             $PSDefaultParameterValues["*-Dba*:EnableException"] = $true
 
+            # One database and one connection per test. The scenario is spent once it has run: the restore
+            # fails because the previous database went offline, so the connection is left in master, and a
+            # second call from the same connection would have master as its previous database and restore
+            # it without trouble.
             $offlineDbName = "dbatoolsci_typeext_offline_$(Get-Random)"
             $null = New-DbaDatabase -SqlInstance $TestConfig.InstanceSingle -Name $offlineDbName
             $offlineServer = Connect-DbaInstance -SqlInstance $TestConfig.InstanceSingle -Database $offlineDbName -NonPooledConnection
+
+            $offlineStopDbName = "dbatoolsci_typeext_offlinestop_$(Get-Random)"
+            $null = New-DbaDatabase -SqlInstance $TestConfig.InstanceSingle -Name $offlineStopDbName
+            $offlineStopServer = Connect-DbaInstance -SqlInstance $TestConfig.InstanceSingle -Database $offlineStopDbName -NonPooledConnection
 
             $PSDefaultParameterValues.Remove("*-Dba*:EnableException")
         }
@@ -152,9 +167,11 @@ Describe $CommandName -Tag IntegrationTests {
         AfterAll {
             $PSDefaultParameterValues["*-Dba*:EnableException"] = $true
 
-            $null = $offlineServer | Disconnect-DbaInstance
-            $null = Set-DbaDbState -SqlInstance $TestConfig.InstanceSingle -Database $offlineDbName -Online -Force -ErrorAction SilentlyContinue
-            $null = Remove-DbaDatabase -SqlInstance $TestConfig.InstanceSingle -Database $offlineDbName -ErrorAction SilentlyContinue
+            $null = $offlineServer, $offlineStopServer | Disconnect-DbaInstance
+            foreach ($offlineDatabase in $offlineDbName, $offlineStopDbName) {
+                $null = Set-DbaDbState -SqlInstance $TestConfig.InstanceSingle -Database $offlineDatabase -Online -Force -ErrorAction SilentlyContinue
+                $null = Remove-DbaDatabase -SqlInstance $TestConfig.InstanceSingle -Database $offlineDatabase -ErrorAction SilentlyContinue
+            }
 
             $PSDefaultParameterValues.Remove("*-Dba*:EnableException")
         }
@@ -171,6 +188,28 @@ Describe $CommandName -Tag IntegrationTests {
             { $offlineServer.Databases["master"].Invoke($offlineStatement) } | Should -Not -Throw
 
             (Get-DbaDbState -SqlInstance $TestConfig.InstanceSingle -Database $offlineDbName).Status | Should -Be "OFFLINE"
+        }
+
+        It "returns normally even when the caller turned warnings into terminating errors" {
+            # The same scenario with $WarningPreference = Stop, which is the one setting that can hand the
+            # outcome of the call back to the housekeeping after all: Write-Warning obeys the preference of
+            # the caller, and Stop makes it throw. The statement would have succeeded and the caller would
+            # still see an exception - and where the statement itself failed, that exception would replace
+            # its error. The wrapper neutralises the preference for its own warning, so this returns.
+            # The test above still proves that a caller asking for silence gets silence, which is why the
+            # preference is not simply forced to Continue.
+            $WarningPreference = "Stop"
+            $offlineStatement = "ALTER DATABASE [$offlineStopDbName] SET OFFLINE WITH ROLLBACK IMMEDIATE"
+
+            # Unlike the test above, the caller has not asked for silence here, so the warning is written -
+            # and it cannot be asserted on, because a script method writes to the host and neither
+            # -WarningVariable nor a redirection inside the method can reach it. It is merged into the
+            # output stream so that it does not land in the run, where the harness reads every warning as
+            # a defect. That does not soften the test: a preference of Stop throws before anything is
+            # written at all, which is how this fails against the unguarded code.
+            { $null = $offlineStopServer.Databases["master"].Invoke($offlineStatement) 3>&1 } | Should -Not -Throw
+
+            (Get-DbaDbState -SqlInstance $TestConfig.InstanceSingle -Database $offlineStopDbName).Status | Should -Be "OFFLINE"
         }
     }
 
--- a/tests/InModule.TypeExtensions.Tests.ps1
+++ b/tests/InModule.TypeExtensions.Tests.ps1
@@ -1,0 +1,122 @@
+#Requires -Module @{ ModuleName="Pester"; ModuleVersion="5.0" }
+param(
+    $ModuleName  = "dbatools",
+    $CommandName = "InModule.TypeExtensions",
+    $PSDefaultParameterValues = $TestConfig.Defaults
+)
+
+Describe $CommandName -Tag IntegrationTests {
+    # The Query and Invoke script methods of Server and Database in xml\dbatools.Types.ps1xml run through
+    # the execution manager of a database, which is the connection context of the parent server and belongs
+    # to the caller. SMO issues a USE and never switches back, so the methods put the database back. See #10555.
+    BeforeAll {
+        # We want to run all commands in the BeforeAll block with EnableException to ensure that the test fails if the setup fails.
+        $PSDefaultParameterValues["*-Dba*:EnableException"] = $true
+
+        $contextDbName = "dbatoolsci_typeext_$(Get-Random)"
+        $null = New-DbaDatabase -SqlInstance $TestConfig.InstanceSingle -Name $contextDbName
+
+        # We want to run all commands outside of the BeforeAll block without EnableException to be able to test for specific warnings.
+        $PSDefaultParameterValues.Remove("*-Dba*:EnableException")
+    }
+
+    AfterAll {
+        # We want to run all commands in the AfterAll block with EnableException to ensure that the test fails if the cleanup fails.
+        $PSDefaultParameterValues["*-Dba*:EnableException"] = $true
+
+        $null = Remove-DbaDatabase -SqlInstance $TestConfig.InstanceSingle -Database $contextDbName -ErrorAction SilentlyContinue
+
+        $PSDefaultParameterValues.Remove("*-Dba*:EnableException")
+    }
+
+    Context "The database context of the caller survives the script methods (#10555)" {
+        BeforeAll {
+            $PSDefaultParameterValues["*-Dba*:EnableException"] = $true
+
+            # Only a non-pooled connection can show this. SMO reopens a pooled connection at its default
+            # database, which hides the leak.
+            $callerServer = Connect-DbaInstance -SqlInstance $TestConfig.InstanceSingle -NonPooledConnection
+
+            $PSDefaultParameterValues.Remove("*-Dba*:EnableException")
+        }
+
+        AfterAll {
+            $PSDefaultParameterValues["*-Dba*:EnableException"] = $true
+
+            $null = $callerServer | Disconnect-DbaInstance
+
+            $PSDefaultParameterValues.Remove("*-Dba*:EnableException")
+        }
+
+        It "leaves the database context alone in Database.Query" {
+            $null = $callerServer.Databases[$contextDbName].Query("SELECT 1")
+            $callerServer.ConnectionContext.ExecuteScalar("SELECT DB_NAME()") | Should -Be "master"
+        }
+
+        It "leaves the database context alone in Database.Invoke" {
+            $null = $callerServer.Databases[$contextDbName].Invoke("SELECT 1")
+            $callerServer.ConnectionContext.ExecuteScalar("SELECT DB_NAME()") | Should -Be "master"
+        }
+
+        It "leaves the database context alone in Server.Query with a database" {
+            $null = $callerServer.Query("SELECT 1", $contextDbName)
+            $callerServer.ConnectionContext.ExecuteScalar("SELECT DB_NAME()") | Should -Be "master"
+        }
+
+        It "leaves the database context alone in Server.Invoke with a database" {
+            $null = $callerServer.Invoke("SELECT 1", $contextDbName)
+            $callerServer.ConnectionContext.ExecuteScalar("SELECT DB_NAME()") | Should -Be "master"
+        }
+
+        It "still runs the query in the database that was asked for" {
+            $callerServer.Databases[$contextDbName].Query("SELECT DB_NAME() AS dbname").dbname | Should -Be $contextDbName
+            $callerServer.Query("SELECT DB_NAME() AS dbname", $contextDbName).dbname | Should -Be $contextDbName
+        }
+
+        It "still returns every table when AllTables is used" {
+            $allTables = $callerServer.Databases[$contextDbName].Query("SELECT 1 AS a; SELECT 2 AS b", $true)
+            $allTables.Count | Should -Be 2
+        }
+
+        It "keeps the session, so temporary objects survive" {
+            $null = $callerServer.ConnectionContext.ExecuteNonQuery("CREATE TABLE #dbatoolsci_marker (id INT)")
+            $null = $callerServer.Databases[$contextDbName].Query("SELECT 1")
+            { $callerServer.ConnectionContext.ExecuteScalar("SELECT COUNT(*) FROM #dbatoolsci_marker") } | Should -Not -Throw
+        }
+
+        It "puts the database back even when the query fails" {
+            { $callerServer.Databases[$contextDbName].Query("SELECT * FROM dbatoolsci_does_not_exist") } | Should -Throw
+            $callerServer.ConnectionContext.ExecuteScalar("SELECT DB_NAME()") | Should -Be "master"
+        }
+    }
+
+    Context "The database the caller was on is restored, not master (#10555)" {
+        BeforeAll {
+            $PSDefaultParameterValues["*-Dba*:EnableException"] = $true
+
+            # A connection that starts out somewhere other than master. Restoring to master would pass the
+            # tests above and still be wrong here.
+            $msdbServer = Connect-DbaInstance -SqlInstance $TestConfig.InstanceSingle -Database msdb -NonPooledConnection
+
+            $PSDefaultParameterValues.Remove("*-Dba*:EnableException")
+        }
+
+        AfterAll {
+            $PSDefaultParameterValues["*-Dba*:EnableException"] = $true
+
+            $null = $msdbServer | Disconnect-DbaInstance
+
+            $PSDefaultParameterValues.Remove("*-Dba*:EnableException")
+        }
+
+        It "leaves the connection in msdb after Database.Query" {
+            $null = $msdbServer.Databases[$contextDbName].Query("SELECT 1")
+            $msdbServer.ConnectionContext.ExecuteScalar("SELECT DB_NAME()") | Should -Be "msdb"
+        }
+
+        It "leaves the connection in msdb after Server.Query with a database" {
+            $null = $msdbServer.Query("SELECT 1", $contextDbName)
+            $msdbServer.ConnectionContext.ExecuteScalar("SELECT DB_NAME()") | Should -Be "msdb"
+        }
+    }
+}
--- a/tests/InModule.TypeExtensions.Tests.ps1
+++ b/tests/InModule.TypeExtensions.Tests.ps1
@@ -9,6 +9,16 @@ Describe $CommandName -Tag IntegrationTests {
     # The Query and Invoke script methods of Server and Database in xml\dbatools.Types.ps1xml run through
     # the execution manager of a database, which is the connection context of the parent server and belongs
     # to the caller. SMO issues a USE and never switches back, so the methods put the database back. See #10555.
+    BeforeDiscovery {
+        # Two databases whose names differ only in case can only exist on an instance with a case sensitive
+        # collation, and that is the only place where a case insensitive comparison in the restore can be
+        # caught. The collation decides it, not the version, so this is asked of the instance rather than
+        # assumed. On a case insensitive instance the scenario cannot be built at all and the Context skips.
+        $caseDiscoveryServer = Connect-DbaInstance -SqlInstance $TestConfig.InstanceSingle
+        $instanceIsCaseSensitive = $caseDiscoveryServer.Collation -match "_CS_"
+        $null = $caseDiscoveryServer | Disconnect-DbaInstance
+    }
+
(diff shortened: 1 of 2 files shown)


potatoqualitee:
gonna put this through a round of ChatGPT Pro since it's such a sensitive change

potatoqualitee:
Glad I checked! 

## Findings

### P1: The restore check fails on case-sensitive SQL Server instances

**File:** `xml/dbatools.Types.ps1xml`
**Lines:** 25, 47, 80, and 122

All four methods use this comparison:

```powershell
$connectionContext.CurrentDatabase -ne $previousDatabase
```

PowerShell string comparison operators are case-insensitive unless the `-c*` form is used. SQL Server database names are instance-level identifiers and use the instance collation. On a case-sensitive instance, `AppDb` and `appdb` can be different databases. In that situation, SMO can move the connection from one to the other, but this comparison evaluates as equal and skips the restore. The original context leak therefore remains on a valid SQL Server configuration. ([[Microsoft Learn](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_comparison_operators?view=powershell-7.6)][1])

At minimum, use `-cne`:

```powershell
if ($previousDatabase -and $connectionContext.CurrentDatabase -cne $previousDatabase) {
```

An ordinal comparison is more explicit:

```powershell
$databaseChanged = -not [string]::Equals(
    $connectionContext.CurrentDatabase,
    $previousDatabase,
    [System.StringComparison]::Ordinal
)
```

A server-collation-aware comparison would be exact, but an ordinal comparison is safe here. At worst it performs an unnecessary restore; it does not miss a real database change. This also needs a regression test using database names that differ only by case on a case-sensitive instance.

---

### P2: A restoration failure can hide the real error, or make a successful command appear to have failed

**File:** `xml/dbatools.Types.ps1xml`
**Lines:** 24–28, 46–50, 79–83, and 121–125

The restoration command is unguarded inside each `finally`:

```powershell
$null = $connectionContext.ExecuteNonQuery("USE [$escapedDatabase]")
```

There are two problematic outcomes:

1. The query fails, then the restore also fails. The restore exception replaces the original query exception.
2. The command succeeds but makes the previous database unavailable, for example by dropping it, taking it offline, renaming it, or revoking access. The wrapper then throws during restoration even though the requested SQL already completed.

`USE` requires `CONNECT` permission and can legitimately fail. This is especially dangerous for `Invoke`, because a caller may interpret the exception as “the command did not execute” and retry an operation that already succeeded. ([[Microsoft Learn](https://learn.microsoft.com/ga-ie/sql/t-sql/language-elements/use-transact-sql?preserve-view=true&view=sql-server-ver17&viewFallbackFrom=azure-sqldw-latest)][2])

The wrapper should retain the original `ErrorRecord` as the primary failure. If the operation succeeded but restoration failed, it should throw a distinct error stating that the SQL completed but the connection context could not be restored. If both fail, preserve both errors without replacing the original.

---

### P2: The temporary-table test does not prove that the wrapper uses the same session

**File:** `tests/InModule.TypeExtensions.Tests.ps1`
**Lines:** 81–84

The test currently:

1. Creates the temporary table through the caller connection.
2. Executes `Database.Query("SELECT 1")`.
3. Checks for the temporary table through the caller connection again.

A copied-connection implementation would also pass this test. The temporary table remains on the original connection regardless of which session executed `SELECT 1`. The test therefore does not enforce the same-session behavior that motivated the implementation.

Query the temporary table through the wrapper:

```powershell
$null = $callerServer.ConnectionContext.ExecuteNonQuery(
    "CREATE TABLE #dbatoolsci_marker (id INT)"
)

$result = $callerServer.Databases[$contextDbName].Query(
    "SELECT OBJECT_ID('tempdb..#dbatoolsci_marker') AS object_id"
)

$result.object_id | Should -Not -BeNullOrEmpty
```

Or compare `@@SPID` directly:

```powershell
$callerSpid = $callerServer.ConnectionContext.ExecuteScalar("SELECT @@SPID")

$querySpid = $callerServer.Databases[$contextDbName].Query(
    "SELECT @@SPID AS spid"
).spid

$querySpid | Should -Be $callerSpid
```

## Verdict

**Request changes.** Restoring the original context on the existing session is the correct overall approach, and the database-name escaping is correct. However, the case-insensitive comparison leaves the original bug unfixed on case-sensitive instances. The cleanup error handling also introduces ambiguous and potentially dangerous failure reporting. The temporary-table test should be corrected so it actually locks in the same-session guarantee.

[1]: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_comparison_operators?view=powershell-7.6 "https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_comparison_operators?view=powershell-7.6"
[2]: https://learn.microsoft.com/ga-ie/sql/t-sql/language-elements/use-transact-sql?preserve-view=true&view=sql-server-ver17&viewFallbackFrom=azure-sqldw-latest "https://learn.microsoft.com/ga-ie/sql/t-sql/language-elements/use-transact-sql?preserve-view=true&view=sql-server-ver17&viewFallbackFrom=azure-sqldw-latest"

andreasjordan:
All three findings were right and all three are fixed. Two of them I could reproduce, and one of those needed a lab that did not exist yet.

## P1 — the case insensitive comparison

Correct, and it is the one that needed new hardware. Three case sensitive instances now exist in the lab (`SQL_Latin1_General_CP1_CS_AS`, otherwise identical builds and editions to their case insensitive counterparts), which is what made this testable at all.

All four methods compare with `-cne` now. It cannot restore needlessly, because both sides are read from the same property, so one database always spells itself the same way — the worst case you described cannot even arise. Against the old code the new test fails like this:

```
Expected: 'dbatoolsci_CaseCtx120377875'
But was:  'dbatoolsci_casectx120377875'
           -----------^
```

The caller was left in the *other* database, which is exactly the leak this PR removes, on a valid configuration.

**Where the regression test runs.** Its `Context` is guarded by a `BeforeDiscovery` probe of the instance collation, so it skips where the scenario cannot be built and runs where it can. On the current CI instance: 11 passed, 1 skipped. Against a case sensitive instance: 12 passed, 0 skipped. So this is not covered on CI until a case sensitive instance exists there — worth saying plainly rather than letting a green run imply otherwise.

## P2 — the restore could replace the outcome

Also correct, and it is not theoretical. Reproduced with a caller sitting in its own database and a statement that takes that database offline:

```
Invoke THREW: ... An exception occurred while executing a Transact-SQL statement or batch.
database state now: OFFLINE   <-- the statement the caller asked for had succeeded
```

The `USE` is now in its own `try`/`catch` and warns instead of throwing, so the result or the original error always reaches the caller untouched.

I did not take the suggestion of throwing a distinct error when the statement succeeded but the restore failed. Throwing on a statement that already ran is the dangerous half of the finding — the caller most likely to be hurt by it is one that retries — and a distinct exception type does not help a caller that simply propagates it. Restoring is housekeeping and should not be able to decide the outcome of the call.

Worth recording for anyone testing this: a `Write-Warning` inside a script method **can be suppressed** by the caller through `$WarningPreference` or `-WarningAction`, but it **cannot be captured** with `-WarningVariable` or `3>&1` — script methods do not run in the caller's stream context. That is why the test sets the preference and asserts on behaviour rather than on the warning text.

## P3 — the temporary table test proved nothing

Right, and for the reason given: the table never left the caller's session, so a copied connection would have passed too. It now queries the marker **through the wrapper** and compares `@@SPID`, so the design that was rejected fails both assertions instead of passing them.

## Tests

Verified to have teeth by reverting the source and keeping the tests — both new cases fail against the old code with the diagnostics above.

- `InModule.TypeExtensions`: 12 tests, 12 passed against a case sensitive instance, 11 passed and 1 skipped against a case insensitive one
- 16 test files of wrapper-using commands: 116 tests, 109 passed, 7 skipped, 0 failed, no leftovers in the lab

One small discovery from building the fixture: two databases whose names differ only in case need distinct **file** names, because NTFS is case insensitive and the derived `.mdf` and then `.ldf` collide. `New-DbaDatabase` cannot express that, so the test creates the second database under a name of its own and renames it with `ALTER DATABASE ... MODIFY NAME`.

---

potatoqualitee:
## Verdict

**Request changes.** The current head, `cfce992a`, correctly addresses all three findings from the first review, and the Azure CI matrix is green. One remaining correctness defect still violates the PR’s stated guarantee that restoration housekeeping must never determine the outcome of the requested SQL operation.

## Findings

### P2: `$WarningPreference = 'Stop'` makes a failed restore terminate the call

All four restore handlers catch the failed `USE` and call an unqualified `Write-Warning`:

```powershell
catch {
    Write-Warning "The database context could not be restored ..."
}
```

That is not guaranteed to remain nonterminating. PowerShell documents that when `$WarningPreference` is `Stop`, `Write-Warning` displays the warning and then stops execution. `WarningAction` can override that behavior for the individual command.     ([[Microsoft Learn](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_preference_variables?view=powershell-7.5)][1])

Consequently, in the PR’s existing offline-database scenario:

1. The requested `ALTER DATABASE ... SET OFFLINE` succeeds.
2. Restoring the former context fails.
3. `Write-Warning` becomes terminating because the caller uses `$WarningPreference = 'Stop'`.
4. The wrapper throws even though the requested SQL completed.

That recreates the exact retry ambiguity described in the source comments. If the original query also failed, the action-preference exception raised by `Write-Warning` can replace the original SQL exception. For `Server.Query` and `Server.Invoke`, it additionally enters the existing outer exception-rewriting block.

The current test explicitly sets `$WarningPreference = "SilentlyContinue"`, so it verifies suppression but cannot detect this path.

**Required change:** make the restore diagnostic incapable of becoming terminating, for example by forcing the internal warning emission to `Continue`, or otherwise absorbing action-preference termination. Then run the existing failed-restore fixture with:

```powershell
$WarningPreference = "Stop"

{
    $offlineServer.Databases["master"].Invoke($offlineStatement)
} | Should -Not -Throw
```

The database-state assertion should remain, proving the SQL really executed.

### P3: The case-sensitive test skips binary collations

The discovery probe considers an instance case-sensitive only when the collation name contains `_CS_`:

```powershell
$instanceIsCaseSensitive = $caseDiscoveryServer.Collation -match "_CS_"
```

SQL Server’s `_BIN` and `_BIN2` collations are also case-sensitive, but their names do not contain `_CS_`. Because database names are instance-level identifiers and therefore use the instance collation, a binary-collation instance is a valid place for the two-databases-differing-only-by-case regression test. The current probe incorrectly skips it.  ([[Microsoft Learn](https://learn.microsoft.com/pt-pt/sql/relational-databases/collations/collation-and-unicode-support)][2])

This is nonblocking test coverage. A behavioral probe is preferable, or the name check should at least include `_BIN` and `_BIN2`, such as:

```powershell
$instanceIsCaseSensitive =
    $caseDiscoveryServer.Collation -match '_(CS|BIN2?)(_|$)'
```

## What is now correct

The prior review findings were fixed properly:

* All four comparisons now use `-cne`, preserving distinct database names on case-sensitive instances.
* The restore `USE` is protected, so its SQL exception no longer directly replaces the requested operation’s result.
* Database names are correctly escaped.
* The same-session test now queries the temporary object through the wrapper and checks `@@SPID`.
* The implementation restores the caller’s actual previous database rather than always returning to `master`.
* The standard CI matrix passed, although the differing-only-by-case fixture is skipped on the current case-insensitive CI instance.

## Copy-ready review comment

> ### P2: `$WarningPreference = 'Stop'` still allows restoration housekeeping to become the call outcome
>
> All four restore catch blocks call `Write-Warning` without forcing it to remain nonterminating. PowerShell turns `Write-Warning` into a terminating error when the effective `$WarningPreference` is `Stop`.
>
> In the existing offline-database scenario, the requested SQL succeeds, the restore fails, and the method still throws. That recreates the retry ambiguity this change is intended to eliminate. When the requested SQL also failed, the warning-generated exception can replace the original SQL error.
>
> The current regression test uses `$WarningPreference = "SilentlyContinue"`, so it cannot detect this case. Please make the restore diagnostic incapable of terminating the call, then run the failed-restore fixture with `$WarningPreference = "Stop"` and assert both `Should -Not -Throw` and that the database was actually taken offline.
>
> Nonblocking: the case-sensitive discovery check only recognizes `_CS_`. SQL Server `_BIN` and `_BIN2` collations are also case-sensitive, so the differing-only-by-case test is incorrectly skipped on those instances. A behavioral check, or a regex that includes `BIN`/`BIN2`, would cover them.

**I would not approve this revision until the `$WarningPreference = 'Stop'` path is fixed.**

[1]: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_preference_variables?view=powershell-7.5 "https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_preference_variables?view=powershell-7.5"
[2]: https://learn.microsoft.com/pt-pt/sql/relational-databases/collations/collation-and-unicode-support "https://learn.microsoft.com/pt-pt/sql/relational-databases/collations/collation-and-unicode-support"

andreasjordan:
Both fixed. The blocker is real and reaches a little further than the report, and I did not take the suggested form of the fix, for a reason worth stating.

## P2 — a preference of Stop, and Inquire with it

Measured before changing anything, on a script method with a failing restore, on both editions:

| caller's `$WarningPreference` | before | now |
| --- | --- | --- |
| `SilentlyContinue` | silent, returns | silent, returns |
| `Continue` | warns, returns | warns, returns |
| `Stop` | **throws** | warns, returns |
| `Inquire` | **throws** | warns, returns |

`Inquire` is the same defect by another route - it stops just as dead where nothing can answer the prompt - so it is neutralised with `Stop`. Identical on Windows PowerShell 5.1 and PowerShell 7.6.

**I did not use `-WarningAction Continue`.** It fixes the terminating case, but it overrides the preference in *both* directions, and that is not what we want:

```
caller sets SilentlyContinue, method passes -WarningAction Continue -> warning is written anyway
```

That takes away the suppression the existing test asserts, and it is worse behaviour: a caller who asked for silence should get silence. So only the preferences that can terminate are neutralised, and only for that one call:

```powershell
if ($WarningPreference -notin "SilentlyContinue", "Continue") { $WarningPreference = "Continue" }
Write-Warning "The database context could not be restored to [$previousDatabase]: ..."
```

All four handlers carry it. The existing `SilentlyContinue` test stays exactly as it was and still passes, which is what proves the suppression survived.

The new test is the fixture you asked for, with `$WarningPreference = "Stop"`, asserting both that the call returns and that the database really went offline. It needs a database and a connection of its own: the scenario is spent once it has run, because the failed restore leaves the connection in `master`, so a second call from the same connection has `master` as its previous database and restores it without trouble.

Against the unguarded code it fails with exactly the predicted error:

```
Expected no exception to be thrown, but an exception "... The running command stopped because the
preference variable "WarningPreference" or common parameter is set to Stop: The database context
could not be restored to [dbatoolsci_typeext_offlinestop_...]" was thrown
```

**One thing worth recording**, because it decides what such a test can assert: the warning from a **ps1xml script method goes straight to the host**. `3>&1` captures nothing at all - I checked, zero objects - while the warning is still displayed. So the test can assert the absence of a throw and the database state, but never the warning text. Merging the stream does keep it out of the run, which the harness reads as a defect.

## P3 — took the behavioural probe rather than the regex

You offered `_(CS|BIN2?)(_|$)` as the lesser option and a behavioural probe as the better one. The probe asks the instance directly:

```sql
SELECT CASE WHEN EXISTS (SELECT 1 FROM sys.databases WHERE name = UPPER(DB_NAME(1))) THEN 0 ELSE 1 END
```

`sys.databases.name` carries the collation of the instance, which is the collation database names are compared with - so this asks the exact question the `-Skip:` needs, and it is right for `_CS_`, `_BIN`, `_BIN2` and anything added later without matching a name at all. `DB_NAME(1)` is `master`, so it needs no string literal of its own, which also keeps it inside the repository's quoting rule.

It agrees with the old check on all five lab instances:

```
SQL05\SQL2025    SQL_Latin1_General_CP1_CS_AS   probe=True   name match=True
SQL05\SQL2019    SQL_Latin1_General_CP1_CS_AS   probe=True   name match=True
SQL03\SQL2019    SQL_Latin1_General_CP1_CI_AS   probe=False  name match=False
SQL03\SQL2025    SQL_Latin1_General_CP1_CI_AS   probe=False  name match=False
SQL03\SQL2022    SQL_Latin1_General_CP1_CI_AS   probe=False  name match=False
```

**Being straight about what is not covered:** the lab has no `_BIN` or `_BIN2` instance, so the binary collation case is handled by construction rather than demonstrated. The probe no longer depends on the collation name, so there is nothing left for a binary collation to trip over, but I would rather say that than imply coverage that does not exist.

## Tests

`InModule.TypeExtensions`, run against both kinds of instance:

| instance | collation | result |
| --- | --- | --- |
| `SQL03\SQL2019` | `SQL_Latin1_General_CP1_CI_AS` | 13 tests, 12 passed, 1 skipped |
| `SQL05\SQL2025` | `SQL_Latin1_General_CP1_CS_AS` | 13 tests, **13 passed, 0 skipped** |

The skip on the case insensitive instance is the two-databases-differing-only-by-case context, which cannot be built there at all.

16 test files of wrapper using commands - `Copy-DbaDbAssembly`, `Copy-DbaSystemDbUserObject`, `Get-DbaDbFile`, `Get-DbaInstanceInstallDate`, `Get-DbaModule`, `Install-DbaFirstResponderKit`, `Install-DbaWhoIsActive`, `Invoke-DbaDbDecryptObject`, `Invoke-DbaDbShrink`, `Move-DbaDbFile`, `Read-DbaTransactionLog`, `Set-DbaDbCompression`, `Test-DbaDbCompression`, `Test-DbaMigrationConstraint`, `Test-DbaTempDbConfig`, `Write-DbaDbTableData` - 186 tests, 184 passed, 2 skipped, 0 failed, no warnings, no leftovers in the lab.

---

andreasjordan:
## Where this pull request stands, in one place

### The problem

The `Query` and `Invoke` script methods on `Server` and `Database` move the caller's connection into another database and leave it there.

```powershell
$server = Connect-DbaInstance -SqlInstance $instance      # connection is in master
$server.Query("SELECT 1", "msdb")
$server.ConnectionContext.ExecuteScalar("SELECT DB_NAME()")   # msdb, not master
```

The execution manager of a database object is the connection context of the parent server, which belongs to the caller - not a private connection. SMO issues a `USE` to run the statement and never switches back, so everything that runs on that connection afterwards silently runs in the wrong database. Four methods do this: `Database.Query`, `Database.Invoke`, and the `-Database` overloads of `Server.Query` and `Server.Invoke`.

### What this pull request changes

Each of the four remembers the current database and puts it back afterwards. The rest of the change is about the ways that goes wrong.

1. **The comparison that decides whether to restore is case sensitive.** Database names use the collation of the instance, so on a case sensitive instance `AppDb` and `appdb` are two different databases, and a case insensitive `-ne` reported them as equal and skipped the restore - leaving the caller in the wrong database, which is the very thing this fixes. It cannot restore needlessly: both sides are read from the same property, so one database always spells itself the same way.

2. **The name is escaped**, so a database with a `]` in its name cannot break the statement.

3. **Restoring never becomes the outcome of the call.** A statement that makes the previous database unreachable - taking it offline, dropping it, renaming it, revoking access - succeeds, and then the `USE` that would put the context back cannot work. Throwing there would tell the caller "it did not run" about a statement that ran, and a caller that retries would run it twice. Where the statement itself failed, the restore error would replace its error. So the restore warns instead of throwing.

4. **The warning cannot terminate the call either.** `Write-Warning` obeys the caller's `$WarningPreference`, and `Stop` turns it into a terminating error - which would hand the outcome back to the housekeeping after all. `Inquire` does the same where nothing can answer the prompt. Both are neutralised for that one warning. Suppression is deliberately left alone, so a caller who asked for silence still gets silence.

### Evidence

- `InModule.TypeExtensions.Tests.ps1` against a case **insensitive** instance: 13 tests, 12 pass, 1 skips
- The same file against a case **sensitive** instance: **13 pass, 0 skip**
- Every new test verified to fail against the old code, with the specific error it guards
- 16 test files of commands that use these wrappers: 186 tests, 184 pass, 2 skip, 0 fail
- CI: 20 pass, 1 skipping, 0 failures

### Two things worth knowing, because they shaped the tests

- **A warning written inside a ps1xml script method goes straight to the host.** It can be suppressed by the caller's preference, but it cannot be captured - not with `-WarningVariable`, not with `3>&1`, which returns zero objects. So the tests assert on behaviour, never on warning text.
- **Two databases whose names differ only in case need distinct file names.** NTFS is case insensitive, so the derived `.mdf` and then `.ldf` collide. `New-DbaDatabase` cannot express that, so the fixture creates the second database under a name of its own and renames it.

### What is covered where

The case sensitive scenario cannot be built at all on a case insensitive instance, so it skips there - which is every CI instance. Whether the instance is case sensitive is asked of the instance as a behaviour, not matched against the collation name, because `_BIN` and `_BIN2` are case sensitive too and carry no `_CS_`. The lab has no binary collation instance, so that case is handled by construction rather than demonstrated.

---

### What is worth looking hardest at

Point 4, neutralising the warning preference. It is the one place the code deliberately overrides something the caller asked for. The argument is that `Stop` cannot be honoured without breaking point 3, and that showing the diagnostic is better than hiding it - but preferring silence there would be a reasonable position too.

---

*This overview now also heads the pull request body, added today. Nothing was removed there - the original description follows underneath it unchanged.*

potatoqualitee:
thank you 💯

PR #10580, opened 2026-08-14
Database context - Run server level statements on the connection

### LINKED ISSUE #10555: Database-scoped SMO calls silently change the current database of the shared connection

andreasjordan:
## Type of Change
 - [x] Bug fix (non-breaking change, fixes #10555 in part)
 - [x] Pester test is included

### Purpose

Step two of #10555, and the half that needs no new mechanism, so it is independent of #10579.

Ten call sites went through a database object only because they needed *somewhere* to run a statement. The execution manager of an SMO database is the connection context of the parent server, so each of them issued a `USE` and left the connection of the caller in `master` or `msdb`. None of the statements needed a database context at all.

The clearest example is [`Export-DbaLogin.ps1:391-393`](https://github.com/dataplat/dbatools/blob/development/public/Export-DbaLogin.ps1#L391-L393), where the primary path was already right and only the fallback was not:

```powershell
$hashedPass = $server.ConnectionContext.ExecuteScalar($sql)          # fine
} catch {
    $hashedPassDt = $server.Databases['master'].ExecuteWithResults($sql)   # same query, leaks
```

### Approach

Each statement now runs on the connection itself, so there is no context to put back:

| Command | Statement | Why no database is needed |
|---|---|---|
| `Export-DbaLogin`, `New-DbaLogin`, `Get-LoginPasswordHash` | password hash from `sys.sql_logins` / `sys.server_principals` | server level catalog views; the primary path already used the connection |
| `Get-DbaDbDetachedFileInfo` | `fn_helpcollations()` | available in every database |
| `Get-OfflineSqlFileStructure` | `SERVERPROPERTY(...)` | does not depend on the current database |
| `Set-DbaTempDbConfig` | `ALTER DATABASE tempdb ...` | does not depend on the current database |
| `Remove-DbaAgentJob` | `sp_delete_job` | named in full as `msdb.dbo.sp_delete_job` |

Only the last one changes a statement. Verified that the three part name works and leaves the context alone:

```
context before : master
3-part sp_delete_job: worked
context after  : master
job still there: False
```

Also verified that `ConnectionContext.ExecuteNonQuery` takes the string array `Set-DbaTempDbConfig` hands it, which is the one signature difference between the two receivers.

The help of `Connect-DbaInstance` recommended the pattern this removes:

> To execute SQL commands directly: `$server.ConnectionContext.ExecuteReader($sql)` or `$server.Databases['master'].ExecuteNonQuery($sql)`

It now points at the connection context and says why.

### Tests

Regression contexts for the two sites that a test can actually reach, both asserting on a `-NonPooledConnection`, because SMO reopens a pooled one at its default database and hides the leak:

- `Remove-DbaAgentJob` still removes the job and leaves the connection where it was
- `Get-DbaDbDetachedFileInfo` still reads the file and resolves the collation, and leaves a connection that was on `msdb` on `msdb`

The other sites have no reachable test, and it is worth being explicit about why rather than adding a test that proves nothing:

- the three password hash sites are `catch` fallbacks for `ExecuteScalar`, which does not fail on a supported SQL Server, so the changed line does not run
- `Get-OfflineSqlFileStructure` has **no callers anywhere in the module**; the change is correct but the function is dead code
- `Set-DbaTempDbConfig` reaches its changed line only by really reconfiguring tempdb. The existing `Set-DbaTempDbConfig` tests do execute that path, so it is exercised, but a context assertion cannot pass yet - see below.

Test files run: `Export-DbaLogin` (20), `New-DbaLogin` (15), `Get-DbaDbDetachedFileInfo` (9), `Remove-DbaAgentJob` (12 + 1 skipped), `Sync-DbaLoginPassword` (9), all passing.

### Set-DbaTempDbConfig is only half fixed by this

It also reads tempdb through `$server.Databases['tempdb'].Query(...)` in three places, which is the script method fix in #10579. Measured on this branch, the connection still ends up in tempdb from those calls even when the command stops before the line changed here:

```
context before : msdb
command failed: Current tempdb ... is not suitable to be reconfigured ...
context after  : tempdb
```

So the command is free of the leak only once both changes are in, and a context test for it belongs with the second one.

### DIFF
--- a/private/functions/Get-LoginPasswordHash.ps1
+++ b/private/functions/Get-LoginPasswordHash.ps1
@@ -63,7 +63,9 @@ function Get-LoginPasswordHash {
             $hashedPass = $server.ConnectionContext.ExecuteScalar($sql)
         } catch {
             try {
-                $hashedPassDt = $server.Databases["master"].ExecuteWithResults($sql)
+                # Same query as above, so it runs on the connection as well. Going through the master
+                # database would leave the connection of the caller there. See #10555.
+                $hashedPassDt = $server.ConnectionContext.ExecuteWithResults($sql)
                 $hashedPass = $hashedPassDt.Tables[0].Rows[0].Item(0)
             } catch {
                 Stop-Function -Message "Failed to retrieve password hash for login $($Login.Name)" -ErrorRecord $_ -Target $Login -Continue
--- a/private/functions/Get-OfflineSqlFileStructure.ps1
+++ b/private/functions/Get-OfflineSqlFileStructure.ps1
@@ -27,7 +27,9 @@ Internal function. Returns dictionary object that contains file structures for S
 
     if ($filestream) {
         $sql = "SELECT COALESCE(SERVERPROPERTY('FilestreamConfiguredLevel'),0) AS fs"
-        $fscheck = $server.databases['master'].ExecuteWithResults($sql)
+        # SERVERPROPERTY does not depend on the current database, so this runs on the connection. Going
+        # through the master database would leave the connection of the caller there. See #10555.
+        $fscheck = $server.ConnectionContext.ExecuteWithResults($sql)
         if ($fscheck.tables.fs -eq 0) { return $false }
     }
 
--- a/public/Connect-DbaInstance.ps1
+++ b/public/Connect-DbaInstance.ps1
@@ -21,7 +21,7 @@ function Connect-DbaInstance {
 
         To execute SQL commands directly: $server.ConnectionContext.ExecuteReader($sql) or $server.ConnectionContext.ExecuteNonQuery($sql)
 
-        Run statements through the connection context rather than through a database object. A database object executes on the same connection and leaves it in that database, which changes the database of every later command that reuses the connection.
+        For statements that do not depend on a particular database, use the connection context rather than a database object. A database object executes on the same connection and leaves it in that database, which changes the database of every later command that reuses the connection. A statement that does need a database still has to say so, because the connection context runs it in whatever database the session currently happens to be in.
 
     .PARAMETER SqlInstance
         The target SQL Server instance or instances. This can be a collection and receive pipeline input to allow the function to be executed against multiple SQL Server instances.
--- a/public/Connect-DbaInstance.ps1
+++ b/public/Connect-DbaInstance.ps1
@@ -19,7 +19,9 @@ function Connect-DbaInstance {
         https://msdn.microsoft.com/en-us/library/system.data.sqlclient.sqlconnectionstringbuilder.aspx
         https://msdn.microsoft.com/en-us/library/system.data.sqlclient.sqlconnection.aspx
 
-        To execute SQL commands directly: $server.ConnectionContext.ExecuteReader($sql) or $server.Databases['master'].ExecuteNonQuery($sql)
+        To execute SQL commands directly: $server.ConnectionContext.ExecuteReader($sql) or $server.ConnectionContext.ExecuteNonQuery($sql)
+
+        Run statements through the connection context rather than through a database object. A database object executes on the same connection and leaves it in that database, which changes the database of every later command that reuses the connection.
 
     .PARAMETER SqlInstance
         The target SQL Server instance or instances. This can be a collection and receive pipeline input to allow the function to be executed against multiple SQL Server instances.
--- a/public/Export-DbaLogin.ps1
+++ b/public/Export-DbaLogin.ps1
@@ -390,7 +390,9 @@ function Export-DbaLogin {
                             try {
                                 $hashedPass = $server.ConnectionContext.ExecuteScalar($sql)
                             } catch {
-                                $hashedPassDt = $server.Databases['master'].ExecuteWithResults($sql)
+                                # Same query as above, so it runs on the connection as well. Going through the
+                                # master database would leave the connection of the caller there. See #10555.
+                                $hashedPassDt = $server.ConnectionContext.ExecuteWithResults($sql)
                                 $hashedPass = $hashedPassDt.Tables[0].Rows[0].Item(0)
                             }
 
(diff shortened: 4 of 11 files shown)


potatoqualitee (commented):
## Findings

### [P1] `Set-DbaTempDbConfig -Force` still leaves the caller in `tempdb`

**File:** `public/Set-DbaTempDbConfig.ps1:364`

The forced file-reduction path adds batches beginning with `USE [tempdb]` to `$sql` for every file being emptied and removed. The PR then executes that array directly through the caller’s `ConnectionContext`. This avoids SMO’s implicit switch to `master`, but it does nothing about the explicit `USE [tempdb]` already inside the SQL. After a successful reduction, or an error after the first such batch, a shared non-pooled connection remains in `tempdb`.

This also remains broken after PR #10579 because these statements bypass the `Query` and `Invoke` wrappers entirely. The existing real-instance `-Force` test verifies the removed files and preserved data but never checks `DB_NAME()` afterward.

The execution needs either:

* A `try/finally` that remembers and restores `ConnectionContext.CurrentDatabase`.
* A copied connection targeted to `tempdb`, since these operations do not depend on the caller’s session state.

The regression test should use a fresh `-NonPooledConnection` beginning in `master` or `msdb`, perform an actual forced reduction, and then assert the original database.

### [P2] The new collation test passes even when the changed query fails

**File:** `tests/Get-DbaDbDetachedFileInfo.Tests.ps1:94`

`Get-DbaDbDetachedFileInfo` catches every failure from the new `ConnectionContext.ExecuteWithResults()` call and falls back to the numeric collation ID. The test only asserts that `Collation` is non-null and nonempty, so it passes even if the changed call always throws and never resolves a collation name.

Capture the expected collation before detaching the database and compare it directly:

```powershell
$expectedCollation = $server.Databases[$dbname].Collation
# detach and execute command
$callerResult.Collation | Should -Be $expectedCollation
```

### [P3] The new help text overgeneralizes the safe pattern

**File:** `public/Connect-DbaInstance.ps1:24`

The advice now says to run statements through `ConnectionContext` rather than a database object. That is correct only for statements that genuinely do not require a database context. A database-scoped statement sent through `ConnectionContext` executes in whatever database the session currently occupies, which can be just as dangerous.

I would change it to something like:

> For server-scoped statements that do not depend on a particular database, execute through the connection context rather than a database object.

## Verdict

**Request changes.** The login/hash, detached-file, offline-file and Agent substitutions otherwise look mechanically correct. The primary Actions run is green, but it does not cover the remaining `Set-DbaTempDbConfig -Force` context leak.

andreasjordan:
All three findings were right and all three are fixed. The first one is the interesting one, because the conclusion was that this command cannot be made to avoid the database context at all — so it has to put it back instead.

## P1 — `Set-DbaTempDbConfig -Force` still left the caller in tempdb

Confirmed, and there were two sources rather than one:

- the batches built for `-Force`, which carry an explicit `USE [tempdb]`, because `DBCC SHRINKFILE` empties a file of the **current** database only
- reading the tempdb data files, because `FILEPROPERTY` reports on the **current** database and, outside it, silently returns NULL rather than failing

So unlike the other ten sites in this PR, this command has no server-level form to move to. It now remembers the database of the caller and puts it back at both places, the second in a `finally`, because a reconfiguration that fails half way moves the context just as much. Measured against a connection that starts in msdb:

```
without the fix:  caller starts in msdb -> ends in tempdb
with the fix:     caller starts in msdb -> ends in msdb
```

That holds on the failure path too, which is worth having: a real forced reduction in the lab hit `DBCC SHRINKFILE: Page 6:2521 could not be moved because it is a work table page`, and the caller still came back to msdb.

The two file path lookups **did** turn out to be server level and now read `sys.master_files` on the connection, so they no longer touch the context at all.

### One suggestion I looked at and rejected

Making the whole command server level by taking used space from `sys.dm_db_file_space_usage` instead of `FILEPROPERTY` works — it is reachable cross-database as `tempdb.sys.dm_db_file_space_usage` — but the two do not agree:

```
file      DMV allocated_extent_page_count      FILEPROPERTY SpaceUsed
temp2     0.50 MB                              0.56 MB
temp3     0.44 MB                              0.50 MB
```

They differ by an extent. That value feeds the check that refuses a reduction when tempdb is already fuller than the requested target, and a leak fix is no place to quietly change what a size check measures.

## P2 — the collation test passed even if the changed call always threw

Right. The command catches every failure of the lookup and falls back to the numeric collation id, which is not null or empty either. It now compares against the collation the database really had, remembered before it is detached, so the fallback cannot satisfy it.

## P3 — the help overgeneralised

Also right, and the wording you suggested is the one I took, with the reason added: a database-scoped statement sent through the connection context runs in whatever database the session happens to be in, so it still has to name one.

## Tests

The forced reduction that `Set-DbaTempDbConfig.Tests.ps1` already performs now runs through a connection that starts in msdb, so the context assertion comes for free rather than reconfiguring tempdb a second time. msdb rather than master, because restoring to master would pass and still be wrong.

- `Set-DbaTempDbConfig`: 6 passed against `SQL Server 2019` with a default tempdb
- `Export-DbaLogin` 20, `Get-DbaDbDetachedFileInfo` 9, `New-DbaLogin` 15, `Remove-DbaAgentJob` 12 and 1 skipped, `Sync-DbaLoginPassword` 9 — all passed

The context fix was verified to have teeth the deterministic way rather than through the reduction: with the fix reverted, the same call leaves the connection in tempdb.

## A warning about that reduction test for whoever reviews it next

`adds real files and safely force-removes the highest file ids` is **environment sensitive, and it can fail for reasons that have nothing to do with the code under review**:

```
Cannot move all contents of file "tempdev4" to other places to complete the emptyfile operation.
DBCC SHRINKFILE: Page 6:2521 could not be moved because it is a work table page.
```

Work table pages cannot be moved by `EMPTYFILE`. It failed deterministically on one lab instance with 8 MB tempdb files and passed on the same instance with 256 MB files and on another with default files. I confirmed it was not this change by re-running the **unmodified** test on the failing instance, where it fails identically.

When it does fail, its `AfterAll` restore can fail as well and the instance is left with surplus tempdb files and an oversized log — so it is worth pointing at a throwaway instance rather than a shared one while working on this command.

---

potatoqualitee:
thank you 🙏🏼

PR #9505, opened 2024-10-13
Connect-DbaInstance: Fix for issue when changing database context

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes # )
 - [ ] New feature (non-breaking change, adds functionality, fixes # )
 - [ ] Breaking change (affects multiple commands or functionality, fixes # )
 - [ ] Ran manual Pester test and has passed (`.\tests\manual.pester.ps1`)
 - [ ] Adding code coverage to existing functionality
 - [ ] Pester test is included
 - [ ] If new file reference added for test, has is been added to github.com/dataplat/appveyor-lab ?
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

The problem was discovered while working on the test for Remove-DbaDbTableData. Inside of Backup-DabaDatabase changing the database context to master was not possible and resulted in using a cached connection to an already dropped database.

In the first commit, I added a test inside of Connect-DbaInstance to write a warning:

```
if ($Database -and $server.ConnectionContext.CurrentDatabase -ne $Database) {
    Write-Message -Level Warning -Message "Changing connection context to database $Database was not successful. Current database is $($server.ConnectionContext.CurrentDatabase). Please open an issue on https://github.com/dataplat/dbatools/issues."
}
```

In the second commit I added a test to show the warning:

image

In the third commit, I implement the fix, changing `$connContext = $connContext.GetDatabaseConnection($Database)` to `$connContext = $connContext.GetDatabaseConnection($Database, $false)`.

PR #8025, opened 2021-12-14
Invoke-DbaQuery - Remove fixed default for QueryTimeout and use global default from Connect-DbaInstance

andreasjordan:
## Type of Change

 - [ ] Bug fix (non-breaking change, fixes # )
 - [x] New feature (non-breaking change, adds functionality, fixes # )
 - [ ] Breaking change (effects multiple commands or functionality, fixes # )
 - [ ] Ran manual Pester test and has passed (`.\tests\manual.pester.ps1`)
 - [ ] Adding code coverage to existing functionality
 - [ ] Pester test is included
 - [ ] If new file reference added for test, has is been added to github.com/sqlcollaborative/appveyor-lab ?
 - [ ] Unit test is included
 - [ ] Documentation
 - [ ] Build system

### Purpose
Invoke-DbaQuery uses a fixed default of 10 minutes for QueryTimeout which causes problems for long runing queries like index rebuilds.
Most other commands use the default of Connect-DbaInstance. The parameter is named StatementTimeout there and has a default of `Get-DbatoolsConfigValue -FullName 'sql.execution.timeout'` with a default of 0 = unlimited.

### Approach
Removing the fixed default and reading StatementTimeout from ConnectionContext.

### Commands to test
```
# This will fail as it failed before:
Invoke-DbaQuery -SqlInstance sql01 -Query "WAITFOR DELAY '00:00:20'" -QueryTimeout 10

# This now fails as well, so it uses the StatementTimeout from Connect-DbaInstance
$server = Connect-DbaInstance -SqlInstance sql01 -StatementTimeout 10
Invoke-DbaQuery -SqlInstance $server -Query "WAITFOR DELAY '00:00:20'"

# Fails as well, so even if the ConnectionContext is changed to the new database 
Invoke-DbaQuery -SqlInstance $server -Database tempdb -Query "WAITFOR DELAY '00:00:20'" 

# Fails as well, so even when the database SMO is used for pipeline input
$server.Databases['model'] | Invoke-DbaQuery -Query "WAITFOR DELAY '00:00:20'"
```

image

PR #7548, opened 2021-07-14
Connect-DbaInstance - Reorder changes when connection context is copied

andreasjordan:
## Type of Change

 - [x] Bug fix (non-breaking change, fixes #7546 )

```
$connContext = $connContext.GetDatabaseConnection($Database)
```
opens the new connection and thus has to be the last change and not the first made in the copied connection context.

PR #6904, opened 2020-10-03
Connect-DbaInstance - sql.connection.experimental to be able to safely test new code paths

andreasjordan:
In order to be able to test new functions in various environments, the switch "experimental" is introduced. This switch can be set with "Set-DbatoolsConfig -FullName sql.connection.experimental -Value $true" for the active session and within this function leads to the following code path being used.

Please have a look and give feedback. I will continue to work on this and publish code to test this as well in the next days.

PR #10584, opened 2026-08-16
Connect-DbaInstance - Give the cloned server ownership of its database connection

### LINKED ISSUE #7546: Connect-DbaInstance - Property NonPooledConnection cannot be changed or read after a connection string has been set.

Yes

```
WARNING: [22:52:39][Write-DbaDbTableData] Error occurred while establishing connection to localhost | Property NonPooledConnection cannot be changed or read after a connection string has been set.
```

### Steps to Reproduce

```
$server = Connect-DbaInstance -SqlInstance "localhost,14333"
$server | Watch-DbaDbLogin -SqlInstance localhost -Database tempdb
WARNING: [22:52:39][Write-DbaDbTableData] Error occurred while establishing connection to localhost | Property NonPooledConnection cannot be changed or read after a connection string has been set.
```

Yes

### Other details or mentions

_No response_

PowerShell Core (pwsh.exe)

### PowerShell Host Version

PS 7 on Windows

docker

andreasjordan:
## Overview

*Added 2026-08-23, after several rounds of review grew this beyond what the original description below covers. Nothing was removed - the original text follows unchanged.*

### The problem

Passing an existing server object back into `Connect-DbaInstance` together with a different `-Database` returned a server whose connection **nothing could ever close**.

```powershell
$clone  = Connect-DbaInstance -SqlInstance $server -Database model
$clone | Disconnect-DbaInstance     # closed nothing
```

The session stayed open for the life of the process, sitting in the target database and holding a shared lock on it. On `model` that is enough to make the next `CREATE DATABASE` on that instance fail:

```
Could not obtain exclusive lock on database 'model'. Retry the operation later.
```

which surfaced as an intermittent failure in whatever test file happened to run next, with nothing pointing back at the cause. In the full suite of 2026-08-15, 69 files failed with `Timeout expired ... prior to obtaining a connection from the pool`.

**Why it happened:** the code called `ConnectionContext.Copy().GetDatabaseConnection($Database, $false)`. `GetDatabaseConnection` opens the connection on the intermediate copy and returns a *different* context. The server handed back therefore never owned the connection that had been opened, and `Disconnect-DbaInstance` can only reach the context of the server it is given.

### What this pull request changes

All of it is in the one branch of `Connect-DbaInstance` that clones an existing server object.

1. **Ownership.** The database is set on the copy the returned server will own, instead of through `GetDatabaseConnection`. Disconnecting now closes what was opened. Sessions stop accumulating: repeated connect and disconnect cycles went from `4, 5, 4, 5, 5, 6, 6, 7` to flat.

2. **Servers built from a raw `SqlConnection`.** Such a server has a fixed connection string, and SMO refuses to assign `DatabaseName`, `NonPooledConnection` or `ServerInstance` on it, so `-Database`, `-NonPooledConnection` and `-DedicatedAdminConnection` used to fail outright. Those settings now go into the connection string instead, where they are `Initial Catalog`, `Pooling` and `Data Source`.

3. **Never hand back a server that cannot log in.** Building a new connection means logging in again, and only a connection string that still carries the whole login can do that. Once a `SqlConnection` has been opened with `Persist Security Info=False`, SqlClient hides the password - and a `SqlCredential` is not copied over at all. Where the login cannot be reproduced, the command switches the database on the connection that already exists, and where even that is impossible it refuses with a clear message rather than returning something that fails on first use.

4. **Azure SQL Database.** It is the only engine that refuses to switch the database of a connection it has already opened - it answers a `USE` with error 40508 and wants a new connection instead. It gets one, or an explicit error. Azure SQL Managed Instance is not grouped with it, because it takes the `USE`.

5. **`-ApplicationIntent`.** On a fixed connection string, assigning the property succeeded silently without reaching the string - and the string is what the login is made with. The context reported `ReadOnly` over a connection that had logged in as `ReadWrite`. It goes into the connection string now.

6. **Failure paths release their connection.** `Copy()` clones an open connection as an open connection. Any exit that did not build a server object - a refusal, a failed connect, a failed database switch - left one behind that nothing could reach.

7. **The tab expansion cache** was handed an open copy and keeps it for the life of the process. Whenever the context was already open, every call parked another session: `3, 4, 5, 6, 7, 8`. It gets a disconnected copy now.

### What it does not change

Connecting from a plain instance name behaves exactly as before - that path never had the problem. The reasons behind the four earlier fixes to this line (#6904, #7548, #8025, #9505) are each preserved or provably no longer needed. #9505's non pooled connection mattered because a pooled one could come from an already dropped database, and putting the database in as `Initial Catalog` makes it part of the pool key, which achieves the same thing.

### Evidence

- `Connect-DbaInstance.Tests.ps1`: 51 tests, 49 pass, 2 skip - the local dedicated admin connection in a lab of remote instances, and the SSPI provider
- Reverting the source and keeping the tests makes **8 of them fail**, each with the defect it guards
- Commands that use this path - `Invoke-DbaQuery`, `Remove-DbaDbData`, `New-DbaLogin`, `Get-DbaUserPermission` and eight more: 96 tests, no failures
- CI: 20 pass, 1 skipping, 0 failures

The Azure SQL Database cases run against a real Azure SQL Database in a lab configuration and skip where none is configured, which is every CI environment.

---

When an existing server object is passed in together with a different `-Database`, the connection context is copied and the database connection was then created with `GetDatabaseConnection`:

```powershell
$connContext = $connContext.GetDatabaseConnection($Database, $false)
```

`GetDatabaseConnection` opens the connection on the intermediate copy and returns a *different* `ConnectionContext`, so the server object that is handed back never owns that connection. `Disconnect-DbaInstance` can only reach the context of the server it is given, so nothing ever closed it.

The session therefore stayed open for the life of the process, sitting in the target database and holding a shared `DATABASE` lock on it. On `model` that is enough to make a later `CREATE DATABASE` on the same instance fail:

```
Could not obtain exclusive lock on database 'model'. Retry the operation later.
CREATE DATABASE failed.
```

which surfaced as an intermittent failure in whatever test file happened to run next, with no connection to the command that caused it.

## Where the connection goes

Running the steps of the old code one at a time and counting sessions on the instance after each:

```
0. server connection only                         sessions=1  [62:master]
1. after ConnectionContext.Copy()                 sessions=1
2. after setting NonPooledConnection              sessions=1
3. after GetDatabaseConnection(model)             sessions=2  [62:master, 64:master]
4. after New-Object Smo.Server                    sessions=2
5. after reading CurrentDatabase [model]          sessions=2
6. after Disconnect-DbaInstance                   sessions=2   <-- closes nothing
7. after disconnecting the intermediate copy      sessions=1   <-- this owned it
```

Step 3 opens the connection on the copy. Step 3 also reassigns `$connContext`, which drops the only reference to the object that owns it. Step 6 shows the consequence: the server we return has nothing to close. Step 7 shows who did own it - and that disconnecting the copy is not a usable fix either, because that is the working connection.

## The change

Setting `DatabaseName` on the copy keeps the connection with the context that the returned server owns, so `Disconnect-DbaInstance` closes it. This is also what the connection string paths of this command already do, so the server object path now behaves like the others.

## History of the line, and why each earlier fix still holds

The line has been touched four times, and none of the reasons are lost by this change:

| PR | What it did | Status now |
| --- | --- | --- |
| #6904 (2020) | Introduced it in the new code path, as `ConnectionContext.Copy().GetDatabaseConnection($Database)`, next to a `# TODO: Do we have to check if its the same database?` | Not a fix for a reported bug, it was the chosen implementation |
| #7548, fixes #7546 (2021) | Moved the call to be the **last** change to the copied context, because `GetDatabaseConnection` opens the connection and later property assignments would come too late | Dissolved. Setting `DatabaseName` does not open a connection - the trace above shows nothing opens until the first query. The assignment is kept in the same last position anyway |
| #8025 (2021) | Added a save and restore around it, because `GetDatabaseConnection` resets `StatementTimeout` | No longer needed and removed. `StatementTimeout` is set earlier on the copy and simply stays. `clones when using parameter StatementTimeout` covers it |
| #9505 (2024) | Added the second argument, `GetDatabaseConnection($Database, $false)`, to force a non-pooled connection. Without it, `Backup-DbaDatabase` got **a cached connection to an already dropped database** and the context change silently did not happen. Also added the warning when `CurrentDatabase` does not match | Verified not to regress, see below. The warning is kept |

#9505 is the one worth care, because `$false` forced a non-pooled connection and setting `DatabaseName` does not. It holds because `DatabaseName` puts `Initial Catalog` into the connection string, so the database is part of the pool key: a pooled connection then comes from that database's pool instead of being taken from the original database's pool and switched.

Replaying the exact #9505 scenario against this change - create a database, connect with it as context, drop it, then ask for a clone on `master`:

```
1. connected with -Database dbatoolsci_ctx_853144975 -> CurrentDatabase = dbatoolsci_ctx_853144975
2. dropped dbatoolsci_ctx_853144975
3. clone with -Database master -> CurrentDatabase = master
   warning raised          : none
   a query really runs in  : master
```

And with pooling left on everywhere, which is the case `$false` used to opt out of:

```
base NonPooledConnection = False
clone CurrentDatabase    = tempdb
clone NonPooledConnection= False
query runs in            = tempdb
original still in        = master
warning raised           : none
```

The test #9505 added, `clones when using Backup-DabInstace`, is still in the suite and passes.

## Measured

The clearest way to see it is to repeat the call. Eight cycles of connect to a database from an existing server object, each followed by `Disconnect-DbaInstance`, counting the module's sessions on the instance after every cycle:

**Before** - sessions accumulate without bound, and every new one sits in the target database:

```
cycle 1: 4 sessions [58:master, 60:model, 61:master, 64:model]
cycle 2: 5 sessions [58:master, 60:model, 61:master, 64:model, 73:model]
cycle 3: 4 sessions [...]
cycle 4: 5 sessions [...]
cycle 5: 5 sessions [...]
cycle 6: 6 sessions [58:master, 60:model, 61:master, 64:model, 73:model, 76:model]
cycle 7: 6 sessions [...]
cycle 8: 7 sessions [58:master, 60:model, 61:master, 64:model, 73:model, 76:model, 77:model]
```

**After** - flat, and the same spids are reused every cycle:

```
cycle 1: 3 sessions [58:master, 60:model, 61:master]
cycle 2: 3 sessions [58:master, 60:model, 61:master]
...
cycle 8: 3 sessions [58:master, 60:model, 61:master]
```

That is the difference between an orphaned non-pooled connection, which nothing can ever reuse or close, and a pooled one that goes back to the pool on disconnect. The sessions that remain after the fix are the pool's: they stay `sleeping`, they are reused by the next call, and their number does not grow. Connecting from a plain instance name behaves the same way in both versions, which is why that path never had the problem.

At suite scale the same difference: the full 744 file run of 2026-08-15 had 69 files failing with `Timeout expired ... prior to obtaining a connection from the pool`, and the same run with this fix had **zero**.

Commands that read a per database view through `Invoke-DbaQuery -Database` stop leaking without being touched. `Get-DbaDbQueryStoreOption -Database model` went from three leaked sessions, one of them in `model`, to one leaked session and none in `model`.

## Tests

`tests/Connect-DbaInstance.Tests.ps1` gets a regression test that connects to a database, disconnects again and asserts the session count is back where it started. It was verified to fail against the old implementation (`Expected 4, but got 5`) and to be the only test that fails there.

Run against SQL Server 2025, including the commands that use this path:

| File | Result |
| --- | --- |
| `Connect-DbaInstance` | 33 passed, 1 skipped |
| `Invoke-DbaQuery` | 31 passed |
| `Remove-DbaDbData` | 13 passed |
| `Get-DbaUserPermission` | 6 passed |
| `Get-DbaDbRecoveryModel` | 6 passed |
| `Get-DbaDbQueryStoreOption` | 4 passed |
| `Disconnect-DbaInstance`, `Remove-DbaDbAsymmetricKey`, `Remove-DbaDbCertificate`, `Remove-DbaDbEncryptionKey` | all passed |

106 tests, no failures.

---

### DIFF
--- a/private/testing/Get-TestConfig.ps1
+++ b/private/testing/Get-TestConfig.ps1
@@ -25,6 +25,15 @@ function Get-TestConfig {
         ClusterStorage   = $null
         # ClusterWitness needs a file share witness.
         ClusterWitness   = $null
+        # The Azure SQL Database that the tests of the Azure code paths need. Azure SQL Database is the
+        # only engine that refuses to switch the database of an existing connection - it answers a USE
+        # with error 40508 - so a real one is the only place that branch can be covered. No CI
+        # environment has one, so these stay empty there and those tests skip themselves. A local
+        # configuration that has one sets all three: the server, a database on it, and a credential of a
+        # SQL Server login that can reach that database.
+        AzureSqlDbServer = $null
+        AzureSqlDbName   = $null
+        AzureSqlDbCred   = $null
     }
 
     if (Test-Path $LocalConfigPath) {
--- a/public/Connect-DbaInstance.ps1
+++ b/public/Connect-DbaInstance.ps1
@@ -752,29 +752,52 @@ function Connect-DbaInstance {
                 } elseif ($copyContext) {
                     $isNewConnection = $true
                     $connContext = $inputObject.ConnectionContext.Copy()
+                    # A ServerConnection that was built from a SqlConnection has its connection string set
+                    # explicitly, and SMO then refuses to let some of its properties be assigned:
+                    #   Property <name> cannot be changed or read after a connection string has been set.
+                    # DatabaseName, NonPooledConnection and ServerInstance are all in that group, which is
+                    # why -Database, -NonPooledConnection and -DedicatedAdminConnection used to fail on a
+                    # server that was created from a SqlConnection. Those settings are collected here and
+                    # put into the connection string further down instead, where the same three are
+                    # Initial Catalog, Pooling and Data Source. No readable property tells the two kinds of
+                    # context apart - every one of them reads fine on both - so we find out by trying.
+                    $connectionStringKeyword = @{ }
                     if ($ApplicationIntent) {
                         $connContext.ApplicationIntent = $ApplicationIntent
                     }
                     if ($NonPooledConnection) {
-                        $connContext.NonPooledConnection = $true
+                        try {
+                            $connContext.NonPooledConnection = $true
+                        } catch {
+                            $connectionStringKeyword["Pooling"] = $false
+                        }
                     }
                     if (Test-Bound -Parameter StatementTimeout) {
                         $connContext.StatementTimeout = $StatementTimeout
                     }
-                    if ($DedicatedAdminConnection -and $inputObject.ConnectionContext.ServerInstance -notmatch '^ADMIN:') {
+                    if ($DedicatedAdminConnection -and $inputObject.ConnectionContext.ServerInstance -notmatch "^ADMIN:") {
                         if ($instance.IsLocalHost) {
                             # Use localhost to avoid multiple IP resolution on multi-homed servers (issue #10151)
-                            if ($instance.InstanceName -ne 'MSSQLSERVER') {
-                                $connContext.ServerInstance = "ADMIN:localhost\$($instance.InstanceName)"
+                            if ($instance.InstanceName -ne "MSSQLSERVER") {
+                                $dedicatedAdminServerInstance = "ADMIN:localhost\$($instance.InstanceName)"
                             } else {
-                                $connContext.ServerInstance = "ADMIN:localhost"
+                                $dedicatedAdminServerInstance = "ADMIN:localhost"
                             }
                             # Trust the server certificate because 'localhost' may not match the certificate CN (e.g., FQDN), issue #10254
                             $connContext.TrustServerCertificate = $true
                         } else {
-                            $connContext.ServerInstance = 'ADMIN:' + $connContext.ServerInstance
+                            $dedicatedAdminServerInstance = "ADMIN:$($connContext.ServerInstance)"
                         }
-                        $connContext.NonPooledConnection = $true
+                        try {
+                            $connContext.ServerInstance = $dedicatedAdminServerInstance
+                        } catch {
+                            $connectionStringKeyword["Data Source"] = $dedicatedAdminServerInstance
+                        }
+                        try {
+                            $connContext.NonPooledConnection = $true
+                        } catch {
+                            $connectionStringKeyword["Pooling"] = $false
+                        }
                     }
                     if ($Database) {
                         # We set DatabaseName instead of calling GetDatabaseConnection on purpose.
@@ -788,7 +811,32 @@ function Connect-DbaInstance {
                         # which is also what the connection string code paths of this command already do.
                         # It does not reset StatementTimeout either, so the save and restore that
                         # GetDatabaseConnection needed is gone with it.
-                        $connContext.DatabaseName = $Database
+                        try {
+                            $connContext.DatabaseName = $Database
+                        } catch {
+                            # Falling back to GetDatabaseConnection here would bring back the very leak this
+                            # change is about, so the database goes into the connection string instead. As
+                            # Initial Catalog it is part of the pool key, which is what setting DatabaseName
+                            # achieves as well, so the reason #9505 forced a non pooled connection still holds.
+                            $connectionStringKeyword["Initial Catalog"] = $Database
+                        }
+                    }
+                    if ($connectionStringKeyword.Count -gt 0) {
+                        $changedKeywords = $connectionStringKeyword.Keys -join ", "
+                        Write-Message -Level Debug -Message "Connection context does not accept property assignments, using the connection string for: $changedKeywords"
+                        # The copy has to be disconnected first: setting the connection string of an open
+                        # connection does not move it, and CurrentDatabase would silently stay where it was.
+                        # The copy is a session of its own - a different SPID than the server that was passed
+                        # in - so this does not touch the connection of the caller.
+                        $connContext.Disconnect()
+                        $connectionStringBuilder = New-Object -TypeName Microsoft.Data.SqlClient.SqlConnectionStringBuilder -ArgumentList $connContext.ConnectionString
+                        foreach ($keyword in $connectionStringKeyword.Keys) {
+                            # PowerShell routes property assignments on a connection string builder through its
+                            # dictionary indexer, so the keyword has to be used rather than the property name:
+                            # $connectionStringBuilder.InitialCatalog fails with "Keyword not supported".
+                            $connectionStringBuilder[$keyword] = $connectionStringKeyword[$keyword]
+                        }
+                        $connContext.ConnectionString = $connectionStringBuilder.ConnectionString
                     }
                     $server = New-Object -TypeName Microsoft.SqlServer.Management.Smo.Server -ArgumentList $connContext
                     if ($Database -and $server.ConnectionContext.CurrentDatabase -ne $Database) {
--- a/public/Connect-DbaInstance.ps1
+++ b/public/Connect-DbaInstance.ps1
@@ -777,10 +777,18 @@ function Connect-DbaInstance {
                         $connContext.NonPooledConnection = $true
                     }
                     if ($Database) {
-                        # Save StatementTimeout because it might be reset on GetDatabaseConnection
-                        $savedStatementTimeout = $connContext.StatementTimeout
-                        $connContext = $connContext.GetDatabaseConnection($Database, $false)
-                        $connContext.StatementTimeout = $savedStatementTimeout
+                        # We set DatabaseName instead of calling GetDatabaseConnection on purpose.
+                        # GetDatabaseConnection opens the connection on the copy we are holding and returns
+                        # a *different* ConnectionContext, so the server we build below never owns that
+                        # connection. Disconnect-DbaInstance can only reach the context of the server it is
+                        # given, so nothing ever closed it: the session stayed open for the life of the
+                        # process, holding a shared lock on the database. On model that is enough to make a
+                        # later CREATE DATABASE on the same instance fail on the exclusive lock.
+                        # Setting DatabaseName keeps the connection with the context we hand to the server,
+                        # which is also what the connection string code paths of this command already do.
+                        # It does not reset StatementTimeout either, so the save and restore that
+                        # GetDatabaseConnection needed is gone with it.
+                        $connContext.DatabaseName = $Database
                     }
                     $server = New-Object -TypeName Microsoft.SqlServer.Management.Smo.Server -ArgumentList $connContext
                     if ($Database -and $server.ConnectionContext.CurrentDatabase -ne $Database) {
--- a/public/Connect-DbaInstance.ps1
+++ b/public/Connect-DbaInstance.ps1
@@ -752,138 +752,198 @@ function Connect-DbaInstance {
                 } elseif ($copyContext) {
                     $isNewConnection = $true
                     $connContext = $inputObject.ConnectionContext.Copy()
-                    # A ServerConnection that was built from a SqlConnection has its connection string set
-                    # explicitly, and SMO then refuses to let some of its properties be assigned:
-                    #   Property <name> cannot be changed or read after a connection string has been set.
-                    # DatabaseName, NonPooledConnection and ServerInstance are all in that group, which is
-                    # why -Database, -NonPooledConnection and -DedicatedAdminConnection used to fail on a
-                    # server that was created from a SqlConnection. Those settings are collected here and
-                    # put into the connection string further down instead, where the same three are
-                    # Initial Catalog, Pooling and Data Source. No readable property tells the two kinds of
-                    # context apart - every one of them reads fine on both - so we find out by trying.
-                    $connectionStringKeyword = @{ }
-                    $databaseNeedsSwitch = $false
-                    if ($ApplicationIntent) {
-                        $connContext.ApplicationIntent = $ApplicationIntent
-                    }
-                    if ($NonPooledConnection) {
-                        try {
-                            $connContext.NonPooledConnection = $true
-                        } catch {
-                            $connectionStringKeyword["Pooling"] = $false
+                    # From here on the copy owns a connection of its own: Copy() clones the underlying
+                    # SqlConnection and, when the source is open, hands back a copy that is open as well.
+                    # Only the server object built at the end of this block gives anyone a way to close it
+                    # again, because Disconnect-DbaInstance can reach nothing but the context of the server
+                    # it is given. Every path that leaves without building that server - a refusal below, a
+                    # failed connect, a failed database switch - therefore has to close the copy itself, or
+                    # it becomes exactly the orphaned session this whole code path exists to remove.
+                    $connContextTransferred = $false
+                    try {
+                        # A ServerConnection that was built from a SqlConnection has its connection string set
+                        # explicitly, and SMO then refuses to let some of its properties be assigned:
+                        #   Property <name> cannot be changed or read after a connection string has been set.
+                        # DatabaseName, NonPooledConnection and ServerInstance are all in that group, which is
+                        # why -Database, -NonPooledConnection and -DedicatedAdminConnection used to fail on a
+                        # server that was created from a SqlConnection. Those settings are collected here and
+                        # put into the connection string further down instead, where the same three are
+                        # Initial Catalog, Pooling and Data Source. No readable property tells the two kinds of
+                        # context apart - every one of them reads fine on both - so we find out by trying.
+                        $databaseNeedsSwitch = $false
+                        $connectionStringKeyword = @{ }
+                        if ($ApplicationIntent) {
+                            $connContext.ApplicationIntent = $ApplicationIntent
                         }
-                    }
-                    if (Test-Bound -Parameter StatementTimeout) {
-                        $connContext.StatementTimeout = $StatementTimeout
-                    }
-                    if ($DedicatedAdminConnection -and $inputObject.ConnectionContext.ServerInstance -notmatch "^ADMIN:") {
-                        if ($instance.IsLocalHost) {
-                            # Use localhost to avoid multiple IP resolution on multi-homed servers (issue #10151)
-                            if ($instance.InstanceName -ne "MSSQLSERVER") {
-                                $dedicatedAdminServerInstance = "ADMIN:localhost\$($instance.InstanceName)"
-                            } else {
-                                $dedicatedAdminServerInstance = "ADMIN:localhost"
+                        if ($NonPooledConnection) {
+                            try {
+                                $connContext.NonPooledConnection = $true
+                            } catch {
+                                $connectionStringKeyword["Pooling"] = $false
                             }
-                            # Trust the server certificate because 'localhost' may not match the certificate CN (e.g., FQDN), issue #10254
-                            $connContext.TrustServerCertificate = $true
-                        } else {
-                            $dedicatedAdminServerInstance = "ADMIN:$($connContext.ServerInstance)"
                         }
-                        try {
-                            $connContext.ServerInstance = $dedicatedAdminServerInstance
-                        } catch {
-                            $connectionStringKeyword["Data Source"] = $dedicatedAdminServerInstance
+                        if (Test-Bound -Parameter StatementTimeout) {
+                            $connContext.StatementTimeout = $StatementTimeout
+                        }
+                        if ($DedicatedAdminConnection -and $inputObject.ConnectionContext.ServerInstance -notmatch "^ADMIN:") {
                             if ($instance.IsLocalHost) {
-                                # TrustServerCertificate was assigned above, and on a context with a fixed
-                                # connection string that assignment does not reach the string - the property
-                                # reads back as True while the string still says False. Since the string is
-                                # what the new connection is built from, the same has to be put in there, or
-                                # the localhost DAC fails on a certificate that does not match. See #10254.
-                                $connectionStringKeyword["Trust Server Certificate"] = $true
+                                # Use localhost to avoid multiple IP resolution on multi-homed servers (issue #10151)
+                                if ($instance.InstanceName -ne "MSSQLSERVER") {
+                                    $dedicatedAdminServerInstance = "ADMIN:localhost\$($instance.InstanceName)"
+                                } else {
+                                    $dedicatedAdminServerInstance = "ADMIN:localhost"
+                                }
+                                # Trust the server certificate because 'localhost' may not match the certificate CN (e.g., FQDN), issue #10254
+                                $connContext.TrustServerCertificate = $true
+                            } else {
+                                $dedicatedAdminServerInstance = "ADMIN:$($connContext.ServerInstance)"
+                            }
+                            try {
+                                $connContext.ServerInstance = $dedicatedAdminServerInstance
+                            } catch {
+                                $connectionStringKeyword["Data Source"] = $dedicatedAdminServerInstance
+                                if ($instance.IsLocalHost) {
+                                    # TrustServerCertificate was assigned above, and on a context with a fixed
+                                    # connection string that assignment does not reach the string - the property
+                                    # reads back as True while the string still says False. Since the string is
+                                    # what the new connection is built from, the same has to be put in there, or
+                                    # the localhost DAC fails on a certificate that does not match. See #10254.
+                                    $connectionStringKeyword["Trust Server Certificate"] = $true
+                                }
+                            }
+                            try {
+                                $connContext.NonPooledConnection = $true
+                            } catch {
+                                $connectionStringKeyword["Pooling"] = $false
                             }
                         }
-                        try {
-                            $connContext.NonPooledConnection = $true
-                        } catch {
-                            $connectionStringKeyword["Pooling"] = $false
+                        if ($Database) {
+                            # We set DatabaseName instead of calling GetDatabaseConnection on purpose.
+                            # GetDatabaseConnection opens the connection on the copy we are holding and returns
+                            # a *different* ConnectionContext, so the server we build below never owns that
+                            # connection. Disconnect-DbaInstance can only reach the context of the server it is
+                            # given, so nothing ever closed it: the session stayed open for the life of the
+                            # process, holding a shared lock on the database. On model that is enough to make a
+                            # later CREATE DATABASE on the same instance fail on the exclusive lock.
+                            # Setting DatabaseName keeps the connection with the context we hand to the server,
+                            # which is also what the connection string code paths of this command already do.
+                            # It does not reset StatementTimeout either, so the save and restore that
+                            # GetDatabaseConnection needed is gone with it.
+                            try {
+                                $connContext.DatabaseName = $Database
+                            } catch {
+                                # A fixed connection string. Falling back to GetDatabaseConnection here would
+                                # bring back the very leak this change is about, so the database is either put
+                                # into the connection string or switched on the connection further down.
+                                $databaseNeedsSwitch = $true
+                            }
                         }
-                    }
-                    if ($Database) {
-                        # We set DatabaseName instead of calling GetDatabaseConnection on purpose.
-                        # GetDatabaseConnection opens the connection on the copy we are holding and returns
-                        # a *different* ConnectionContext, so the server we build below never owns that
-                        # connection. Disconnect-DbaInstance can only reach the context of the server it is
-                        # given, so nothing ever closed it: the session stayed open for the life of the
-                        # process, holding a shared lock on the database. On model that is enough to make a
-                        # later CREATE DATABASE on the same instance fail on the exclusive lock.
-                        # Setting DatabaseName keeps the connection with the context we hand to the server,
-                        # which is also what the connection string code paths of this command already do.
-                        # It does not reset StatementTimeout either, so the save and restore that
-                        # GetDatabaseConnection needed is gone with it.
-                        try {
-                            $connContext.DatabaseName = $Database
-                        } catch {
-                            # Falling back to GetDatabaseConnection here would bring back the very leak this
-                            # change is about. The database is switched on the connection of the copy further
-                            # down instead, which needs no new connection and therefore cannot lose anything.
-                            $databaseNeedsSwitch = $true
-                        }
-                    }
-                    if ($connectionStringKeyword.Count -gt 0) {
-                        # Rebuilding the connection string means opening a new connection from it, and that is
-                        # only safe when the string still carries what it takes to log in. Once a SqlConnection
-                        # has been opened with Persist Security Info=False, SqlClient hides the password, so
-                        # the string we can read back no longer has one and the new connection would fail with
-                        # "Login failed". Rejecting that is better than handing back a server that cannot be
-                        # used - and the caller can always pass the instance name and a SqlCredential instead.
+
+                        # Everything above that could not be assigned needs a new connection built from a
+                        # changed connection string. Work out here whether that is possible at all, because
+                        # opening a new connection means logging in again, and only a connection string that
+                        # still carries the whole login can do that.
                         $connectionStringBuilder = New-Object -TypeName Microsoft.Data.SqlClient.SqlConnectionStringBuilder -ArgumentList $connContext.ConnectionString
-                        $usesSqlLogin = -not $connectionStringBuilder["Integrated Security"] -and $connectionStringBuilder["User ID"]
-                        if ($usesSqlLogin -and -not $connectionStringBuilder["Password"] -and -not $connContext.SqlConnectionObject.Credential) {
-                            Stop-Function -Message "Cannot apply the requested settings to [$instance]: they need a new connection, and the password of the SQL Server login is no longer readable from the connection that was passed in. Connect with the instance name and -SqlCredential instead, or pass a SqlConnection that uses Persist Security Info=True." -Target $instance -Continue
+                        if ($ApplicationIntent -and $connectionStringBuilder["ApplicationIntent"] -ne $ApplicationIntent) {
+                            # The property was assigned above without complaint, but on a fixed connection
+                            # string that assignment never reaches the string, and the string is what the login
+                            # is made with. SQL Server reads the intent at login time and routes on it then, so
+                            # a context that reports ReadOnly over a connection that logged in as ReadWrite is
+                            # simply wrong. Where the assignment did reach the string - every context that
+                            # accepts property assignments - the two match here and nothing is added, which is
+                            # what keeps this from forcing a new connection where none is needed.
+                            $connectionStringKeyword["ApplicationIntent"] = $ApplicationIntent
+                        }
+                        $authenticationMethod = [string]$connectionStringBuilder["Authentication"]
+                        if ($connectionStringBuilder["Integrated Security"] -or $connContext.SqlConnectionObject.AccessToken -or $connContext.SqlConnectionObject.AccessTokenCallback) {
+                            # Windows authentication needs no secret from the string, and an access token - or
+                            # the callback that fetches one - lives on the SqlConnection, where both survive
+                            # being given a new connection string.
+                            $connectionStringCanLogIn = $true
+                        } elseif ($authenticationMethod -in "ActiveDirectoryPassword", "ActiveDirectoryServicePrincipal") {
+                            # The only two Microsoft Entra methods that use a password of their own.
+                            $connectionStringCanLogIn = [bool]$connectionStringBuilder["Password"]
+                        } elseif ($authenticationMethod -ne "NotSpecified") {
+                            # Every other Entra method - managed identity, workload identity, default,
+                            # integrated, interactive, device code - is passwordless by design. Several of them
+                            # do put a client id in User ID, so User ID alone must not be read as a SQL Server
+                            # login, or these get rejected over a password they never had.
+                            $connectionStringCanLogIn = $true
+                        } else {
+                            # A SQL Server login, and both halves have to be in the string. Once a SqlConnection
+                            # has been opened with Persist Security Info=False, SqlClient hides the password,
+                            # and a SqlCredential is hidden the same way - ServerConnection.Copy() does not even
+                            # carry one over. So neither can be recovered from an open connection, and what is
+                            # left in the string would log in as nobody.
+                            $connectionStringCanLogIn = [bool]$connectionStringBuilder["User ID"] -and [bool]$connectionStringBuilder["Password"]
                         }
 
                         if ($databaseNeedsSwitch) {
-                            # A new connection is opened anyway, so the database belongs in the string. As
-                            # Initial Catalog it is part of the pool key, which is what setting DatabaseName
-                            # achieves as well, so the reason #9505 forced a non pooled connection still holds.
-                            $connectionStringKeyword["Initial Catalog"] = $Database
-                            $databaseNeedsSwitch = $false
+                            if ($connectionStringCanLogIn) {
+                                # A new connection can be built, so the database belongs in the string. As
+                                # Initial Catalog it is part of the pool key, which is what setting DatabaseName
+                                # achieves as well, so the reason #9505 forced a non pooled connection still
+                                # holds. This is also the only way that works on Azure SQL Database.
+                                $connectionStringKeyword["Initial Catalog"] = $Database
+                                $databaseNeedsSwitch = $false
+                            } elseif ($inputObject.ConnectionContext.DatabaseEngineEdition -in "SqlDatabase", "SqlDataWarehouse", "SqlOnDemand") {
+                                # Read the edition from the server that was passed in, never from the copy:
+                                # reading it there connects the copy, and every property of a connected context
+                                # is locked from then on, including the ones assigned above.
+                                # Azure SQL Database answers a USE with error 40508 and wants a new connection
+                                # for another database, so the switch below cannot work there. Refusing is the
+                                # honest answer - the alternative is a server object in the wrong database.
+                                Stop-Function -Message "Cannot change the database of [$instance] to [$Database]: it needs a new connection, because Azure SQL Database does not support switching the database of an existing one, and the authentication of the connection that was passed in cannot be reproduced from its connection string. Connect with the instance name and -SqlCredential, or with an access token, instead." -Target $instance -Continue
+                            }
                         }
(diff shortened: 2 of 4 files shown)


potatoqualitee:
seems straightforward but there must have been a reason we did it so weirdly. at the same time, it does cause issues. im going to ask gpt to investigate why we likely did it and what the consequences are of the change.

potatoqualitee:
GPT Pro said on my behalf: I dug through the history because this code was weird enough that I didn't believe we did it accidentally.

### The weird code had real reasons

The original `GetDatabaseConnection()` approach goes back to an Azure SQL problem where directly setting `DatabaseName` could fail with:

> Property DatabaseName cannot be changed or read after a connection string has been set.

Then more fixes accumulated around it:

* **#7548:** moved `GetDatabaseConnection()` to the end because it effectively opened/used the connection, after which properties like `NonPooledConnection` couldn't be changed.
* **#8025:** saved/restored `StatementTimeout` because `GetDatabaseConnection()` returned a new context and lost it.
* **#9505:** changed it to `GetDatabaseConnection($Database, $false)` because SMO could return a cached pooled connection for a database that had already been dropped.

So no, we didn't do this for funsies.

### But this PR found a real bug

The ownership problem described in #10584 checks out.

Basically:

**A → Copy to B → GetDatabaseConnection creates C → return C**

But SMO can open/use **B** while creating C. We then throw away our reference to B and return a Server containing C.

`Disconnect-DbaInstance` can only disconnect the returned server's `ConnectionContext`, which is C. It has no way to find B.

So B can sit there for the life of the process holding a database lock. That explains both the accumulating sessions and the bizarre later failures trying to get an exclusive lock on `model`.

For normal dbatools connections, changing this to:

```
$connContext.DatabaseName = $Database
```

is much cleaner. It also makes several old workarounds unnecessary: no premature connection, no lost `StatementTimeout`, and the target database becomes `Initial Catalog` in the connection string, giving it the correct ADO.NET pool.

### One thing I think this PR currently breaks

We also support creating a Server from a raw `Microsoft.Data.SqlClient.SqlConnection`.

Those SMO contexts have an explicit connection string. `Copy()` preserves it, and SMO specifically refuses to let you set `DatabaseName` afterward.

So please add a regression test for:

```
$params = @{
    SqlInstance      = $sqlConnection
    EnableException = $true
}
$server = Connect-DbaInstance @params

$params = @{
    SqlInstance      = $server
    Database         = "tempdb"
    EnableException  = $true
}
$clone = Connect-DbaInstance @params
```

I expect the second call to fail on the new `DatabaseName` assignment.

CI doesn't catch this because we test **SqlConnection → Server** and separately **Server → different Database**, but not **SqlConnection → Server → different Database**.

### What I'd change before merging

Keep the new `DatabaseName` approach for contexts where it's mutable. **Don't just fall back to the old `GetDatabaseConnection()` when it isn't**, because that recreates the leak.

For explicit-connection-string contexts, construct the final database-specific context directly with the target database already in `Initial Catalog`. Be careful to preserve authentication state that may live outside the connection string, such as access tokens/SSPI.

I'd also add coverage for `-Database` + `-NonPooledConnection`. The old `$false` explicitly forced the database connection to be nonpooled. The new implementation normally allows pooling. I think that's probably desirable, but it's still a behavioral change and we have commands that depend on a stable session for temp tables/session state.

Finally, I'd make the leak regression run several cycles. With pooling, a sleeping SQL session remaining is legitimate. What matters is that **the session count stabilizes instead of growing every time**, which your eight-cycle test already demonstrates really well.

### Bottom line

**The PR is fixing a real bug and I prefer the new design. I wouldn't revert to the old approach.** I just want the raw `SqlConnection` case handled before we merge, plus coverage for nonpooled connections and repeated session growth.

andreasjordan:
Thank you for putting this through GPT Pro — the `SqlConnection` finding was real, I reproduced it, and hunting it down turned up two more bugs of the same kind that are already on `development`.

## The report was right, and it was a regression

`SqlConnection` → `Server` → `-Database` works on `development` and threw on this branch:

```
development: clone OK, CurrentDatabase = tempdb
this branch: Property DatabaseName cannot be changed or read after a connection string has been set.
```

The explanation of why CI misses it is right too: we cover `SqlConnection` → `Server`, and separately `Server` → another `Database`, but never the two chained.

## Two more of the same, and they are not new

`DatabaseName` is not the only property SMO locks once a connection string has been set. `NonPooledConnection` and `ServerInstance` are locked as well, and both are already broken on `development`:

| base server built from a `SqlConnection` | development | this branch before | now |
| --- | --- | --- | --- |
| `-Database` | works, leaks the connection | **throws** | works, no leak |
| `-Database -NonPooledConnection` | **throws** | throws | works |
| `-DedicatedAdminConnection` | **throws** | throws | works |

`ApplicationIntent`, `StatementTimeout` and `TrustServerCertificate` stay settable, so the group is exactly those three. They are fixed together here rather than split off, because it is one defect in one command and the middle row is the very combination you asked for coverage of.

## What it does

As advised, it does **not** fall back to `GetDatabaseConnection` — that would bring back the leak this branch exists to remove. Whatever cannot be assigned is collected and put into the connection string, where the same three settings are `Initial Catalog`, `Pooling` and `Data Source`. `Initial Catalog` is part of the pool key, so the reason #9505 forced a non-pooled connection still holds.

There is no way to detect the two kinds of context up front — every property reads fine on both, and `DatabaseName` returns an empty string rather than throwing — so it finds out by trying.

Two things that cost me a while and are worth knowing:

- **The copy has to be disconnected before its connection string is set.** Setting it on an open connection does not move it, and `CurrentDatabase` silently stays where it was. My first attempt looked like it worked and did nothing. The copy is a session of its own — a different SPID from the server that was passed in — so disconnecting it never touches the caller.
- **`$builder.InitialCatalog = ...` fails with "Keyword not supported".** PowerShell routes property assignments on a `SqlConnectionStringBuilder` through its dictionary indexer, so the keyword has to be used: `$builder["Initial Catalog"]`. The developer notes in this command already write it that way.

On preserving authentication: the password survives. `ConnectionContext.ConnectionString` still carries `Password=` after the connection is open, checked with a real SQL login, so rebuilding from it does not lose SQL authentication.

## Tests

A new `Context` for cloning from a server that was created from a `SqlConnection` — the chained shape — covering all three parameter combinations. The DAC test asks the session which endpoint it is on rather than trusting the context name, because `ServerInstance` is one of the properties that cannot be assigned.

Verified to have teeth by reverting the source and keeping the tests: each fails with exactly the exception it guards against.

**The leak regression now runs five cycles**, as suggested. With pooling a single sleeping session that stays behind is legitimate, and only growth proves an orphaned connection, so one cycle could not tell the two apart. Against `development` it fails with `Expected 4, but got 7`.

`Connect-DbaInstance`, `Disconnect-DbaInstance` and `Invoke-DbaQuery`: 71 tests, 70 passed, 1 skipped, 0 failed, no leftovers in the lab.

One note on the diff: the DAC block was restructured, so a few pre-existing single-quoted strings in it became double-quoted. That is the repo style rule and the pre-commit style hook refuses to touch those lines otherwise.

---

potatoqualitee:
giving this a bit more review bc connect-dbainstance is so important -- needs more integration tests

## Verdict

**Request changes.** The leak diagnosis is correct, and replacing `GetDatabaseConnection()` is the right direction. I would not merge the current revision yet because the connection-string fallback has one likely authentication regression and one definite DAC configuration bug.

## Findings

### 1. High, blocker until disproved: an already-open SQL-authenticated `SqlConnection` can lose its password

The new fallback does this:

```powershell
$connContext.Disconnect()
$connectionStringBuilder = New-Object Microsoft.Data.SqlClient.SqlConnectionStringBuilder $connContext.ConnectionString
# modify builder
$connContext.ConnectionString = $connectionStringBuilder.ConnectionString
```

That works for the current test because the input `SqlConnection` uses integrated security and is initially closed. It is unsafe when the original raw `SqlConnection` was **already open using SQL authentication with `Persist Security Info=False`**.

The relevant sequence is:

1. Once the raw connection has opened, SqlClient hides its password and may hide its `SqlCredential` through the public properties.
2. SMO explicitly documents that it cannot recover the password from a `SqlConnection` in this situation.
3. `ServerConnection.Copy()` still creates a functioning clone because SqlClient’s internal clone retains the hidden authentication state.
4. The PR disconnects that functioning clone and rebuilds from the publicly visible, sanitized connection string.
5. On reconnect, SMO clears the underlying `SqlConnection.Credential` and applies the sanitized string, leaving no usable password.

That should produce a login failure when the cloned server is first used. I have not executed this against a SQL-authenticated test instance, but the source path is strong enough that I would require a passing regression test before merging.

Add a test shaped like this:

```powershell
$sqlConnectionString = @(
    "Data Source=$($TestConfig.InstanceMulti1)"
    "Initial Catalog=master"
    "User ID=$sqlLogin"
    "Password=$sqlPassword"
    "Persist Security Info=False"
    "Encrypt=False"
) -join ";"

$sqlConnection = [Microsoft.Data.SqlClient.SqlConnection]::new($sqlConnectionString)
$sqlConnection.Open()

$server = Connect-DbaInstance -SqlInstance $sqlConnection

$serverClone = Connect-DbaInstance `
    -SqlInstance $server `
    -Database tempdb `
    -EnableException

$serverClone.ConnectionContext.ExecuteScalar("select db_name()") |
    Should -Be "tempdb"
```

The implementation needs to preserve authentication state that exists only on the underlying `SqlConnection`, or explicitly reject this situation rather than silently constructing an unusable clone. Falling back to `GetDatabaseConnection()` would not be acceptable because it restores the original leak.

### 2. High: the local DAC fallback loses the forced `TrustServerCertificate=True`

For local DAC connections, the code correctly does this:

```powershell
$connContext.TrustServerCertificate = $true
```

But when `ServerInstance` cannot be changed, the fallback builder starts from the original explicit connection string and only changes `Data Source` and `Pooling`. On an explicit-string `ServerConnection`, setting the `TrustServerCertificate` property changes the SMO field but does **not** modify the explicit connection string returned by `ConnectionString`.

Consequently, an input such as:

```text
Encrypt=True;TrustServerCertificate=False
```

can become:

```text
Data Source=ADMIN:localhost;Encrypt=True;TrustServerCertificate=False
```

That defeats the earlier `localhost` certificate-name fix and can make the DAC connection fail.

The new test cannot detect this because it begins with:

```text
Encrypt=False;Trust Server Certificate=True
```

It would pass even if the local-DAC assignment were deleted entirely.

When the fallback builder is used for a local DAC, explicitly add:

```powershell
$connectionStringKeyword["Trust Server Certificate"] = $true
```

Then change the test input to `Trust Server Certificate=False` and assert that the clone’s resulting connection string contains `TrustServerCertificate=True`.

The same explicit-connection-string behavior affects `ApplicationIntent`: assigning the SMO property succeeds, but it does not alter the actual explicit connection string. That appears to be pre-existing, rather than introduced by this PR, but the new builder path should include `Application Intent` whenever that parameter is bound. Otherwise the context can report `ReadOnly` while the actual SqlClient connection still uses its original routing intent.

### 3. Low: the leak regression can pass despite session-count instability

The new test only compares the first and last values:

```powershell
$countPerCycle[-1] | Should -Be $countPerCycle[0]
```

A sequence such as this passes:

```text
4, 5, 4, 5, 4
```

That still indicates connection churn or intermittent leakage. Assert that every post-cycle count is identical:

```powershell
($countPerCycle | Select-Object -Unique).Count | Should -Be 1
```

If one warm-up cycle is legitimately needed, discard the first measurement and require all remaining values to be identical.

## What the PR gets right

The original ownership diagnosis checks out against SMO’s implementation. For a `ServerConnection` backed by a `SqlConnection`, `Copy()` clones the underlying connection and can open that clone. `GetDatabaseConnection()` then creates yet another `ServerConnection` from a database-specific connection string. Returning the latter leaves the intermediate copied connection outside the returned server’s ownership, so `Disconnect-DbaInstance` has no reference through which to close it.

For normal mutable dbatools contexts, setting `DatabaseName` is substantially cleaner:

* The returned server owns the connection it eventually opens.
* `Disconnect-DbaInstance` can reach it.
* `StatementTimeout` no longer needs save-and-restore handling.
* The database becomes part of the generated connection string and pool identity.
* It avoids SMO’s separate database-connection cache entirely.

The added raw-`SqlConnection`, nonpooled, DAC, and repeated-session tests are the right categories. They just need the two missing edge cases above.

**Recommended GitHub review state: Request changes.** Once the open SQL-authentication case is proven safe and the local-DAC trust setting is put into the actual connection string, I would approve without reverting to `GetDatabaseConnection()`.

andreasjordan:
All three findings reproduced in the lab, and the first one was right to be called a blocker — it is a regression this branch introduced. Pushed a revision that changes the approach rather than patching the fallback.

## 1. The password loss is real

Reproduced against a real SQL Server login, with an already open `SqlConnection` using `Persist Security Info=False`:

```
raw SqlConnection state           : Open
raw ConnectionString has Password : False      <-- SqlClient has hidden it
context ConnectionString has pwd  : False
clone FAILED                      : Login failed for user 'dbatoolsci_revprobe'.
```

Exactly the chain described. `Copy()` still produces a working clone because the authentication state survives on the `SqlConnection`; rebuilding the string from the sanitized public value throws it away.

**The fix is to stop rebuilding the string for `-Database` at all.** The copy already holds a working connection of its own — a different SPID from the server that was passed in — so the database is switched on it:

```powershell
if ($connContext.SqlConnectionObject.State -ne "Open") {
    $connContext.Connect()
}
$connContext.SqlConnectionObject.ChangeDatabase($Database)
```

Nothing is rebuilt, so nothing that lives on the `SqlConnection` rather than in its string can be lost, and the copy stays the context the returned server owns — which is the whole point of the branch. Measured on a SQL authenticated, already open connection: the clone lands in `tempdb`, runs as that login, and the caller stays in `master`.

`GetDatabaseConnection()` is not used, as you asked.

**The two settings that genuinely need a new connection** — `-NonPooledConnection` and `-DedicatedAdminConnection` — still have to rebuild the string, and there the password can genuinely be unrecoverable. Those now **refuse** rather than return an unusable server:

```
Cannot apply the requested settings to [SQL03\SQL2025]: they need a new connection, and the password
of the SQL Server login is no longer readable from the connection that was passed in. Connect with the
instance name and -SqlCredential instead, or pass a SqlConnection that uses Persist Security Info=True.
```

The regression test you asked for is in, shaped as suggested. Against the previous revision both of its cases fail with `Login failed for user`.

## 2. The DAC trust setting is confirmed, and it is worse than a DAC problem

Confirmed, and it is not specific to `TrustServerCertificate`:

```
before  : Data Source=...;Integrated Security=True;Encrypt=False;Trust Server Certificate=False
set TrustServerCertificate : OK, property reads True
set ApplicationIntent      : OK, property reads ReadOnly
after   : Data Source=...;Integrated Security=True;Encrypt=False;Trust Server Certificate=False
string carries TrustServerCertificate=True : False
string carries ApplicationIntent           : False
```

The property assignment succeeds and reads back, while the string is untouched. `Trust Server Certificate` is now put into the builder from the same branch that sets the property, and `ApplicationIntent` with it, for the reason you gave — otherwise the context reports `ReadOnly` while the connection routes as it always did.

**One honest gap:** the localhost DAC path is not covered by a test, because every instance in the lab used for this work is remote, so `IsLocalHost` is never true there. The non-local DAC path is tested and passes. If you want that covered it needs a runner with a local instance.

## 3. Session-count instability

Fair, `4, 5, 4, 5, 4` would have passed. Now every cycle has to report the same number:

```powershell
($countPerCycle | Select-Object -Unique).Count | Should -Be 1
```

Measured over six cycles on both paths, from an instance name and from a `SqlConnection`: `2, 2, 2, 2, 2, 2`, all identical.

## Where it stands

`Connect-DbaInstance`: 41 tests, 40 passed, 1 skipped. The matrix, on a base server built each way:

| | from an instance name | from a `SqlConnection` |
| --- | --- | --- |
| `-Database` | tempdb, caller untouched | tempdb, caller untouched |
| `-Database -NonPooledConnection` | works | works |
| `-DedicatedAdminConnection` | on the DAC endpoint | on the DAC endpoint |
| sessions over 6 cycles | all identical | all identical |

Also rebased onto current `development`, which now carries #10580.

---

andreasjordan:
The gap I flagged in the previous comment is closed: the CI instances are on the machine running the tests, so the local DAC path can be exercised there even though a lab of remote instances cannot reach it. Added a test that skips where the instance is not local and runs on the runners.

It starts from `Trust Server Certificate=False` on purpose, for the reason you gave about the existing one:

```powershell
$localDacConnectionString = "Data Source=$($TestConfig.InstanceMulti1);Integrated Security=True;Encrypt=False;Trust Server Certificate=False"
```

With `True` it would pass even if the command did nothing.

**One thing worth knowing for anyone asserting on connection strings**, which caught me while writing it: a `SqlConnectionStringBuilder` keeps the spelling it was given.

```
input : ...;Trust Server Certificate=False
output: Data Source=ADMIN:localhost\SQL2019;Integrated Security=True;Pooling=False;Encrypt=False;Trust Server Certificate=True

matches "TrustServerCertificate=True" : False
```

So the obvious assertion would have failed on CI for a reason that has nothing to do with the fix. The test reads the setting back through a builder instead, which normalises it:

```powershell
$cloneStringBuilder = New-Object -TypeName Microsoft.Data.SqlClient.SqlConnectionStringBuilder -ArgumentList $serverClone.ConnectionContext.ConnectionString
$cloneStringBuilder["Trust Server Certificate"] | Should -BeTrue
$cloneStringBuilder["Data Source"] | Should -Match "^ADMIN:localhost"
```

`Data Source` covers both shapes the command builds, `ADMIN:localhost` for a default instance and `ADMIN:localhost\<name>` for a named one.

`Connect-DbaInstance` is at 42 tests: 40 passed, 2 skipped here - this one and the SSPI provider one. On the runners the local DAC test should run rather than skip, and that will be the first real exercise of it, so it is worth a look at the CI result rather than trusting my say-so.

---

andreasjordan:
CI is green, and the local DAC gap from the previous comment is now actually closed rather than just intended.

**The MULTI lane failure was infrastructure, not a test.** The job ran exactly 10 minutes, steps 4 to 7 finished with no conclusion at all, and its log blob was never uploaded (`BlobNotFound`), while every other lane in that run took under a minute because they skipped as unaffected. Re-run of the same job: **success**, `Connect-DbaInstance.Tests.ps1` passed in 16.6s.

**The local DAC test did run there.** `Get-TestConfig` builds the CI instance as `"$env:COMPUTERNAME\$env:InstanceMulti1"`, so it is `<runner>\sql2022` - prefixed with the runner's own name, which makes `IsLocalHost` true:

```
localhost        IsLocalHost=True
127.0.0.1        IsLocalHost=True
ADMIN01\FOO      IsLocalHost=True     <-- the shape CI uses
SQL03\SQL2019    IsLocalHost=False    <-- the shape a lab of remote instances uses
```

So the `-Skip:` is false on the runners and true in a lab like mine, which is exactly the split that was wanted: the path that cannot be reached from a lab of remote instances is covered where it can be reached.

Full result for the branch: 22 checks, 21 pass, 1 skipping (the `windows-tests` lane), 0 failures.

One aside that came out of reading those logs and is worth recording somewhere less transient: the ci-azure MULTI lane is `sql2022, sql2017`, so **CI does have a SQL Server 2017 instance**. That was the open question on #10562 - whether the SQL Server 2017 branch of `Get-DbaDbQueryStoreOption` could be verified anywhere.

---

potatoqualitee:
This is a tough cookie, as expected. Changes requested:

## Verdict

**Request changes.** The PR is substantially better than its first revision. The main fix, retaining ownership of the copied `ConnectionContext` rather than replacing it with the result of `GetDatabaseConnection()`, is correct for ordinary SQL Server contexts. The added fixed-connection-string, SQL-authentication, DAC, certificate, and repeated-session tests are useful.

I still found three direct blockers and one authentication issue that needs to be addressed in the same pass.

## 1. Blocking: `ChangeDatabase()` breaks `-Database` on Azure SQL Database

When `DatabaseName` cannot be assigned and no other option requires rebuilding the connection string, the new fallback is:

```powershell
$connContext.Connect()
$connContext.SqlConnectionObject.ChangeDatabase($Database)
```

That works against SQL Server and Azure SQL Managed Instance, but not Azure SQL Database. `SqlConnection.ChangeDatabase()` implements the switch by issuing a `USE [database]` batch. Azure SQL Database does not support switching databases with `USE`; it returns error 40508 and requires a new connection whose `Initial Catalog` targets the destination database.   ([[Microsoft Learn](https://learn.microsoft.com/nb-no/sql/t-sql/language-elements/use-transact-sql?view=fabric)][1])

This means the new integration test proves:

```text
SqlConnection -> Server -> different Database
```

only for regular SQL Server. The same supported input pattern remains broken, and is now a regression from `development`, on Azure SQL Database.

**Required change:** when the context has a fixed connection string, rebuild it with `Initial Catalog` and reconnect whenever the authentication mechanism can be reproduced. If the password or other required secret is genuinely unavailable, then:

* `ChangeDatabase()` can remain a fallback for engines that support `USE`.
* Azure SQL Database must return an explicit error rather than attempting `ChangeDatabase()`.
* Azure SQL Managed Instance should not be grouped with Azure SQL Database here, because MI supports database switching.

An Azure SQL Database regression test would be ideal. At minimum, add a testable engine-type branch proving that the Azure SQL Database path never calls `ChangeDatabase()`.

## 2. Blocking: rejected and failed operations leak the copied connection

SMO’s `ServerConnection.Copy()` clones the underlying `SqlConnection`, and if the source connection is open, it immediately opens the clone as well.

The PR then has multiple exits before ownership is transferred to a returned `Server`:

```powershell
$connContext = $inputObject.ConnectionContext.Copy()

# ...

Stop-Function ... -Continue

# disconnect occurs later
$connContext.Disconnect()
```

The unreadable-password refusal happens before `Disconnect()`. A failed `Connect()`, `ChangeDatabase()`, connection-string assignment, or `New-Object Smo.Server` has the same problem. The open clone has no deterministic cleanup and is left for eventual garbage collection or pool reclamation. That recreates the exact class of orphaned-session problem this PR is intended to remove.

The new SQL-authentication test currently verifies the error text, but not the session count after the rejected operation.

**Required change:** wrap the copied context in ownership-transfer cleanup:

```powershell
$connContext = $inputObject.ConnectionContext.Copy()
$contextTransferred = $false

try {
    # configure the context

    $server = New-Object Microsoft.SqlServer.Management.Smo.Server $connContext
    $contextTransferred = $true
} finally {
    if (-not $contextTransferred -and $connContext) {
        $connContext.Disconnect()
    }
}
```

The exact structure can differ, but every failure or `Stop-Function` path before construction of the returned server must close the copy.

Add two repeated-call tests:

1. Repeated hidden-password refusals do not increase the session count.
2. Repeated invalid database switches do not increase the session count.

## 3. Blocking: `-ApplicationIntent` alone is silently ineffective for fixed connection strings

The code correctly recognizes in its comment that assigning:

```powershell
$connContext.ApplicationIntent = $ApplicationIntent
```

does not update a fixed connection string. However, it only adds `ApplicationIntent` to `$connectionStringKeyword` inside this condition:

```powershell
if ($connectionStringKeyword.Count -gt 0) {
    # ...
    if ($ApplicationIntent) {
        $connectionStringKeyword["ApplicationIntent"] = $ApplicationIntent
    }
}
```

When `-ApplicationIntent ReadOnly` is the only requested change, the hashtable remains empty, the block never executes, and the connection string remains `ReadWrite`. The returned context can report `ReadOnly`, while the actual login was made without read-only intent.

The same thing happens with `-Database X -ApplicationIntent ReadOnly` when the database takes the `ChangeDatabase()` fallback: the existing read-write connection merely issues `USE X`.

Read-only routing requires `ApplicationIntent=ReadOnly` as part of the connection attempt. SQL Server stores and processes the application intent during login; changing an SMO property after that does not reroute the connection.  ([[Microsoft Learn](https://learn.microsoft.com/en-us/sql/database-engine/availability-groups/windows/listeners-client-connectivity-application-failover?view=sql-server-ver17)][2])

**Required change:** determine whether the actual connection string already contains the requested intent, and add the keyword before checking the hashtable count:

```powershell
if ($ApplicationIntent -and
    $connectionStringBuilder.ApplicationIntent -ne $ApplicationIntent) {
    $connectionStringKeyword["ApplicationIntent"] = $ApplicationIntent
}
```

The builder may need to be created earlier to do this cleanly.

Add raw-`SqlConnection` tests for:

```powershell
Connect-DbaInstance -SqlInstance $serverFromSqlConnection `
    -ApplicationIntent ReadOnly

Connect-DbaInstance -SqlInstance $serverFromSqlConnection `
    -Database $agDatabase `
    -ApplicationIntent ReadOnly
```

At minimum, assert the resulting connection string. On an AG-enabled test target, also verify the actual replica or session intent.

## 4. Authentication detection mistakes passwordless Microsoft Entra authentication for SQL authentication

This heuristic is too broad:

```powershell
$usesSqlLogin =
    -not $connectionStringBuilder["Integrated Security"] `
    -and $connectionStringBuilder["User ID"]
```

A non-integrated connection with `User ID` and no password is not necessarily a SQL login. Valid examples include:

* User-assigned managed identity, where `User ID` is the identity’s client ID.
* Active Directory Workload Identity, where `User ID` is the client ID.
* Active Directory Default with a user-assigned managed identity.
* Active Directory Interactive with an optional user ID.
* Active Directory Integrated on supported platforms with an optional user ID.

These modes intentionally have no password, so the current branch rejects valid connections with the “password is no longer readable” error whenever a connection-string rebuild is needed. Microsoft’s SqlClient documentation explicitly describes these passwordless modes and their use of `User ID`.  ([[Microsoft Learn](https://learn.microsoft.com/th-th/sql/connect/ado-net/sql/azure-active-directory-authentication?view=sql-server-ver17)][3])

This becomes especially important when fixing the Azure SQL Database problem, because changing `Initial Catalog` there necessarily requires a new connection.

**Required change:** classify authentication using `SqlConnectionStringBuilder.Authentication`, together with `Credential`, `AccessToken`, and `AccessTokenCallback`. Only require a recoverable password or secret for authentication modes that actually use one. Do not infer authentication type from `User ID` alone.

The error also recommends:

```text
pass a SqlConnection that uses Persist Security Info=True
```

That should not be presented as a normal workaround. Microsoft recommends leaving `Persist Security Info=False`; setting it to true makes the user ID and password retrievable after opening the connection. The safer recommendation is the existing instance-name plus `-SqlCredential` path, or an appropriate token/credential provider. ([[Microsoft Learn](https://learn.microsoft.com/en-us/dotnet/framework/data/adonet/protecting-connection-information)][4])

## What is good in the current PR

The underlying diagnosis is correct: `GetDatabaseConnection()` returned a different context and left the intermediate copied context owning a connection that `Disconnect-DbaInstance` could not reach. The new `DatabaseName` approach is the right design for mutable `ServerConnection` contexts. The retained #9505 backup regression, normal repeated-session test, fixed connection-string tests, DAC coverage, and local certificate handling are all worthwhile.

**Merge recommendation:** do not merge the current head. Fix the Azure SQL Database branch, add deterministic cleanup around the copied context, propagate `ApplicationIntent` before deciding whether to rebuild, and replace the `User ID` authentication heuristic. After those changes, the normal SQL Server path looks suitable.

[1]: https://learn.microsoft.com/nb-no/sql/t-sql/language-elements/use-transact-sql?view=fabric "https://learn.microsoft.com/nb-no/sql/t-sql/language-elements/use-transact-sql?view=fabric"
[2]: https://learn.microsoft.com/en-us/sql/database-engine/availability-groups/windows/listeners-client-connectivity-application-failover?view=sql-server-ver17 "https://learn.microsoft.com/en-us/sql/database-engine/availability-groups/windows/listeners-client-connectivity-application-failover?view=sql-server-ver17"
[3]: https://learn.microsoft.com/th-th/sql/connect/ado-net/sql/azure-active-directory-authentication?view=sql-server-ver17 "https://learn.microsoft.com/th-th/sql/connect/ado-net/sql/azure-active-directory-authentication?view=sql-server-ver17"
[4]: https://learn.microsoft.com/en-us/dotnet/framework/data/adonet/protecting-connection-information "https://learn.microsoft.com/en-us/dotnet/framework/data/adonet/protecting-connection-information"

andreasjordan:
Ok, we will work on that...

potatoqualitee:
pls ask claude to create integration tests for all of this too! so we can prevent regressions in the future

andreasjordan:
All four findings reproduced and fixed, and the second one led to a leak of the same kind that was already on this branch. The lab now has an Azure SQL Database, so the first finding is covered by a test that really runs rather than by an engine-type branch.

## 1. `ChangeDatabase()` on Azure SQL Database

Confirmed, and the scope is narrower than it looked. Against a real Azure SQL Database, with a base server built each way:

| base built from | `development` | this branch before | now |
| --- | --- | --- | --- |
| a `SqlConnection`, password recoverable | works | **USE is not supported** | works |
| a `SqlConnection`, password hidden | Login failed | USE is not supported | refuses, with its own message |
| an instance name (mutable context) | works | works | works |

Only the first row is a regression this branch introduced. The middle row was already broken on `development`, just with a different error, and the third was never affected.

As asked, the connection string is rebuilt with `Initial Catalog` whenever the authentication can be reproduced - which is also what gives the clone the right pool, so the reason #9505 forced a non pooled connection still holds. `ChangeDatabase` survives only as the fallback for engines that take a `USE`, and Azure SQL Database refuses instead of reaching it. Azure SQL Managed Instance is deliberately not grouped with it.

The refusal reads:

```
Cannot change the database of [sqllab...] to [master]: it needs a new connection, because Azure SQL
Database does not support switching the database of an existing one, and the authentication of the
connection that was passed in cannot be reproduced from its connection string. Connect with the
instance name and -SqlCredential, or with an access token, instead.
```

## 2. The copied context, and a second leak the tests you asked for found

The ownership-transfer `try`/`finally` went in as suggested, and `finally` does run when `Stop-Function -Continue` unwinds - both when it throws and when it `continue`s, which I checked, because the warning path is the one no test covers.

**The repeated-cycle tests then found the success path leaking too**, and not through this block at all. At the end of the command the tab expansion cache is handed `$server.ConnectionContext.Copy()` and keeps it for the life of the process. `Copy()` clones an open connection as an open connection, and switching the database on the connection is exactly what leaves the context open, so every call parked a session that nothing could reuse or close:

```
before : 3, 4, 5, 6, 7, 8
after  : 3, 3, 3, 3, 3, 3
```

It gets a disconnected copy now - the updater connects when it runs, and it is off by default. This was already on the branch before this revision, and a single cycle could never have shown it.

**One thing worth knowing about measuring this**, which cost me the first version of the test: counting with `Get-DbaProcess -SqlInstance <name>` opens a new connection per call, and each of those registers itself with the tab expansion cache under the same instance key. That drops the reference to the leaked connection and lets it be collected, so the test passed against the broken code. Counting through a connection opened once shows the growth.

## 3. `-ApplicationIntent`

Confirmed and fixed. The keyword is now added before the count is checked, so `-ApplicationIntent` on its own rebuilds the string instead of doing nothing. The decision compares the requested intent against what the connection string actually carries, so a context that accepts the property assignment already matches and is not pushed into a connection it does not need. Covered alone and together with `-Database`, asserted on the string rather than the property - the property was always right.

I did not do the AG part: verifying the actual replica needs read-only routing configured on the HADR instances, which is more than this PR should carry.

## 4. Authentication detection

Replaced, as asked, with `SqlConnectionStringBuilder.Authentication` plus the access token on the connection. Two measured facts decided the shape of it:

- **`ServerConnection.Copy()` does not carry a `SqlCredential` over.** It reads back empty on the copy even when the source has one, and after `Open()` SqlClient hides it on the source as well, exactly like the password. So a credential can never be evidence that a rebuilt string will log in - the old `-and -not $connContext.SqlConnectionObject.Credential` was only ever giving the right answer by accident. It is gone.
- **`AccessToken` and `AccessTokenCallback` do survive** both `Copy()` and a later `ConnectionString` assignment, so token authentication can be rebuilt. Both are in the check.

A SQL Server login now requires *both* `User ID` and `Password` in the string, which is what correctly rejects the hidden-`SqlCredential` case that has neither. The message no longer suggests `Persist Security Info=True`.

## The Azure SQL Database

The lab has one now, so this is a real integration test rather than a branch assertion. `Get-TestConfig` gains `AzureSqlDbServer`, `AzureSqlDbName` and `AzureSqlDbCred`, empty by default like `ClusterStorage`, so the tests skip themselves in every CI environment and run for real against a lab that has one. `tests/CLAUDE.md` documents it next to the failover cluster exception, so nobody removes the skip later.

Two things keep that skip honest, and both are in the tests: the context opens by asserting the engine really is `SqlDatabase`, without which the rest would pass everywhere while proving nothing, and the refusal asserts the error is **not** the raw `USE statement is not supported`, which is what fails if the code ever reaches `ChangeDatabase` again.

## Tests

`Connect-DbaInstance`: 51 tests, 49 passed, 2 skipped - the local DAC one in a lab of remote instances, and the SSPI provider one.

Verified to have teeth by reverting the source and keeping the tests: **eight fail, each with the defect it guards** - both `ApplicationIntent` assertions on `ReadWrite`, the refusal on the old message, all three session counts on growth, and both Azure tests on `USE statement is not supported`.

Also run, since they use this path: `Invoke-DbaQuery` (31), `Remove-DbaDbData` (13), `New-DbaLogin` (15), `Get-DbaUserPermission` (6), `Get-DbaDbRecoveryModel` (6), `Add-` and `Remove-DbaInstanceList` (9), `Get-DbaDbQueryStoreOption` (4), `Remove-DbaDbAsymmetricKey` (5), `Remove-DbaDbEncryptionKey` (3), `Disconnect-DbaInstance` and `Remove-DbaDbCertificate` (4). 96 tests, no failures, no warnings, no leftovers in the lab.

---

andreasjordan:
## Where this pull request stands, in one place

### The problem

Passing an existing server object back into `Connect-DbaInstance` together with a different `-Database` returned a server whose connection **nothing could ever close**.

```powershell
$clone  = Connect-DbaInstance -SqlInstance $server -Database model
$clone | Disconnect-DbaInstance     # closed nothing
```

The session stayed open for the life of the process, sitting in the target database and holding a shared lock on it. On `model` that is enough to make the next `CREATE DATABASE` on that instance fail:

```
Could not obtain exclusive lock on database 'model'. Retry the operation later.
```

which surfaced as an intermittent failure in whatever test file happened to run next, with nothing pointing back at the cause. In the full suite of 2026-08-15, 69 files failed with `Timeout expired ... prior to obtaining a connection from the pool`.

**Why it happened:** the code called `ConnectionContext.Copy().GetDatabaseConnection($Database, $false)`. `GetDatabaseConnection` opens the connection on the intermediate copy and returns a *different* context. The server handed back therefore never owned the connection that had been opened, and `Disconnect-DbaInstance` can only reach the context of the server it is given.

### What this pull request changes

All of it is in the one branch of `Connect-DbaInstance` that clones an existing server object.

1. **Ownership.** The database is set on the copy the returned server will own, instead of through `GetDatabaseConnection`. Disconnecting now closes what was opened. Sessions stop accumulating: repeated connect and disconnect cycles went from `4, 5, 4, 5, 5, 6, 6, 7` to flat.

2. **Servers built from a raw `SqlConnection`.** Such a server has a fixed connection string, and SMO refuses to assign `DatabaseName`, `NonPooledConnection` or `ServerInstance` on it, so `-Database`, `-NonPooledConnection` and `-DedicatedAdminConnection` used to fail outright. Those settings now go into the connection string instead, where they are `Initial Catalog`, `Pooling` and `Data Source`.

3. **Never hand back a server that cannot log in.** Building a new connection means logging in again, and only a connection string that still carries the whole login can do that. Once a `SqlConnection` has been opened with `Persist Security Info=False`, SqlClient hides the password - and a `SqlCredential` is not copied over at all. Where the login cannot be reproduced, the command switches the database on the connection that already exists, and where even that is impossible it refuses with a clear message rather than returning something that fails on first use.

4. **Azure SQL Database.** It is the only engine that refuses to switch the database of a connection it has already opened - it answers a `USE` with error 40508 and wants a new connection instead. It gets one, or an explicit error. Azure SQL Managed Instance is not grouped with it, because it takes the `USE`.

5. **`-ApplicationIntent`.** On a fixed connection string, assigning the property succeeded silently without reaching the string - and the string is what the login is made with. The context reported `ReadOnly` over a connection that had logged in as `ReadWrite`. It goes into the connection string now.

6. **Failure paths release their connection.** `Copy()` clones an open connection as an open connection. Any exit that did not build a server object - a refusal, a failed connect, a failed database switch - left one behind that nothing could reach.

7. **The tab expansion cache** was handed an open copy and keeps it for the life of the process. Whenever the context was already open, every call parked another session: `3, 4, 5, 6, 7, 8`. It gets a disconnected copy now.

### What it does not change

Connecting from a plain instance name behaves exactly as before - that path never had the problem. The reasons behind the four earlier fixes to this line (#6904, #7548, #8025, #9505) are each preserved or provably no longer needed. #9505's non pooled connection mattered because a pooled one could come from an already dropped database, and putting the database in as `Initial Catalog` makes it part of the pool key, which achieves the same thing.

### Evidence

- `Connect-DbaInstance.Tests.ps1`: 51 tests, 49 pass, 2 skip - the local dedicated admin connection in a lab of remote instances, and the SSPI provider
- Reverting the source and keeping the tests makes **8 of them fail**, each with the defect it guards
- Commands that use this path - `Invoke-DbaQuery`, `Remove-DbaDbData`, `New-DbaLogin`, `Get-DbaUserPermission` and eight more: 96 tests, no failures
- CI: 20 pass, 1 skipping, 0 failures

The Azure SQL Database cases run against a real Azure SQL Database in a lab configuration and skip where none is configured, which is every CI environment.

---

### What is worth looking hardest at

The decision of whether a connection string can log in on its own. It classifies by the `Authentication` keyword together with any access token on the connection, and requires both `User ID` and `Password` for a plain SQL Server login. Wrong in the permissive direction, it returns a server that fails on first use; wrong in the strict direction, it refuses a connection that would have worked. The passwordless Microsoft Entra methods are the interesting cases, and the lab cannot exercise them.

---

*This overview now also heads the pull request body, added today. Nothing was removed there - the original description follows underneath it unchanged.*

potatoqualitee:
Thank you, let's see how this goes

Issue #10125, opened 2026-01-30
Refactoring Set-DbaNetworkCertificate


andreasjordan:
Implementing best practice in security is important, so we want to encourage the users to configure a certificate for their sql server instances. @claude, please have a look at the code of the command Set-DbaNetworkCertificate and give critical feedback on where the command needs refactoring to be very stable and to implement the current best practices.

andreasjordan:
Very good feedback from claude. I agree with most parts, but also disagree on some.

* About 1. We should let go PowerShell v3 long term, maybe starting next major release. But I would also follow claude here.
* About 2. We should refactor that like I did some time ago in other commands. Will do that and maybe provide cloude with a "best practice".
* About 3. I'm not sure how to deal with certificates without privat key. I will investigate this as we have code for that case.
* About 4. I have a FCI for testing, so I will have a look at that. Maybe in that case there are more tests needed for the certificate to be working.
* About 5. I disagree here as the property is created at installation time so it should be there in every case. But we could check it anyway.
* About 8. The string part uses backslaches anyway, so where is the point in using Join-Path for only one part? And we are on windows anyway - so what is the point here?
* About 9. Good point, I like that.
* About 10. I think the code is ok. I think claude does not have information about how Stop-Function works.
* About 13. I disagree here as we don't do that in other commands.
* About 15. I disagree here as we generally don't implement retry logic.
* About 16. I'm not sure about this. This is a general topic that should be the same in all commands using Invoke-Command2.

@potatoqualitee Do you have additional feedback? I will start refactoring this command in a feq days.

andreasjordan:
@claude please implement a new version Set-DbaNetworkCertificate which uses Test-DbaNetworkCertificate to very the certificate and follow all best practices.

rjleue:
@andreasjordan Previously, the Set-DbaNetworkCertificate function granted the SQL service account, which is being used, read permissions to the private key of the certificate. This is also mentioned in the function's description and is an official requirement documented by Microsoft. Unfortunately, this functionality has been removed with this PR. Could you please explain the reason behind this? Thank you

andreasjordan:
Short answer: It is not needed. The running process gets all the rights that come from the account that service is started as (the "startname") plus all the rights that are granted to the service(account) itself ("servicename" / "NT Service\MSSQLSERVER"). As the startname can be changed and the sericename can not, it is better to grant the rights to that object that can not be changed. That way wou don't have to change rights when you change the startname.

[...]

PR #10586, opened 2026-08-19
Clarify private key permissions in Set-DbaNetworkCertificate help

andreasjordan:
Documentation-only fix for the wording that caused the confusion in #10125.

## Problem

The comment-based help said:

> This command also grants read permissions for the service account on the certificate's private key.

Readers reasonably take "the service account" to mean the **startname** of the service. That is not what the code does, and @rjleue hit exactly that in #10125 — their SCOM monitoring flagged the instance because the expected ACE was not on the startname.

## Mechanism

The code grants `Read` on the private key file to the **per-service SID** of the instance — `NT SERVICE\MSSQLSERVER` for the default instance, `NT SERVICE\MSSQL$<InstanceName>` for named instances ([Set-DbaNetworkCertificate.ps1#L193-L206](https://github.com/dataplat/dbatools/blob/development/public/Set-DbaNetworkCertificate.ps1#L193-L206)):

```powershell
# Grant permissions to the Service SID
$sqlSSID = "NT SERVICE\MSSQLSERVER"
if ($instance.InstanceName -ne "MSSQLSERVER") {
    $sqlSSID = "NT SERVICE\MSSQL$" + $instance.InstanceName
}
```

The service SID is always part of the process token of the SQL Server service, so this grants the running instance exactly the access Microsoft's docs ask for, and it survives a change of the startup account. With the default virtual account the two are literally the same object; the difference only shows up with a domain account or a gMSA.

So the behaviour is correct — the help was not.

## What changed

Two blocks of comment-based help in `public/Set-DbaNetworkCertificate.ps1`:

1. `.DESCRIPTION` — now names the per-service SID explicitly and explains why it is used instead of the startname.
2. `.PARAMETER Thumbprint` — dropped the "the SQL Server service account will be granted read permissions to it" clause and points at the description.

- **No code.** The begin/process blocks are untouched; the diff is entirely inside the help block.
- **`.OUTPUTS` `ServiceAccount`** still reads "The service account running the SQL Server instance". That property comes from `$wmiService.ServiceAccount` and really is the startname, so the wording is accurate there.

## Sweep for the same wording elsewhere

`Set-DbaNetworkCertificate` is the only command in the repo that ACLs a certificate private key — `NT SERVICE\MSSQL` appears nowhere else in `public/` or `private/` outside of this command. I checked the sibling network-certificate commands and the computer-certificate family:

- `Get-DbaNetworkCertificate`, `Test-DbaNetworkCertificate`, `New-DbaComputerCertificate`, `Add-DbaComputerCertificate`, `Backup-DbaComputerCertificate` — none of them claim to grant private key permissions, so there is nothing to correct.
- `Remove-DbaNetworkCertificate` mentions `ServiceAccount` only as an output property, same accurate usage as above.

No other help text needed changing.

## Testing

Documentation-only, so no behavioural test applies. Verified that the file parses cleanly with `[System.Management.Automation.Language.Parser]::ParseFile` and that the help block still renders as intended via `Get-Help -Full`, including the literal `MSSQL$<InstanceName>` surviving PowerShell's parsing of the comment block.

Reported by @rjleue in #10125.

### DIFF
--- a/public/Set-DbaNetworkCertificate.ps1
+++ b/public/Set-DbaNetworkCertificate.ps1
@@ -19,7 +19,12 @@ function Set-DbaNetworkCertificate {
         If the given certificate is not returned as a suitable certificate, the command gets detailed information
         about why the given certificate is not suitable and fails with that information.
 
-        This command also grants read permissions for the service account on the certificate's private key.
+        This command also grants read permissions on the certificate's private key
+        to the per-service SID of the instance (NT SERVICE\MSSQLSERVER for the default
+        instance, NT SERVICE\MSSQL$<InstanceName> for named instances). The service SID
+        is always part of the process token of the SQL Server service, so the permission
+        is effective regardless of the account the service runs as, and it does not have
+        to be changed when that account changes.
 
         The currently configured certificate can be unset by using the parameter -UnsetCertificate.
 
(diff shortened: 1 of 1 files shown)


potatoqualitee:
thank you 🙏🏼

PR #10588, opened 2026-08-20
Get-DbaWsfc* - Fix the State property read from an undefined variable, and report the witness path

### LINKED ISSUE #10573: Retrieve the UNC path used as a cluster fileshare witness
### Summarize Functionality

I have tried the below windows cluster commands but I can’t seem to retrieve the UNC path that is used as a fileshare witness. Unless I am missing something, I am requesting that the fileshare path is retrieved for a fileshare witness

Get-DbaWsfcCluster
Get-DbaWsfcNode
Get-DbaWsfcResource

Yes

### Technical Details

_No response_

andreasjordan:
Five of the `Get-DbaWsfc*` commands built their `State` NoteProperty from `$resource.State`, but `$resource` is never assigned in any of them. Only `Get-DbaWsfcResource` and `Get-DbaWsfcResourceGroup` were correct, and only because their loop variable happens to be named `$resource`. Everywhere else `State` was empty on every cluster.

Found while investigating #10573, which the last commit then fixes as well.

## What was wrong

| Command | Defect |
|---|---|
| `Get-DbaWsfcRole` | `State` always empty, and the whole collection was piped into one `Add-Member`, so every role would have shared a single value even after the variable was fixed |
| `Get-DbaWsfcCluster` | `State` always empty - `MSCluster_Cluster` has no `State` property at all |
| `Get-DbaWsfcAvailableDisk` | Same, `MSCluster_AvailableDisk` has no `State` property |
| `Get-DbaWsfcSharedVolume` | Same, plus it queried WMI class `ClusterSharedVolume`, which does not exist in `root\MSCluster`. The command could never return a volume |

## What changed

`Get-DbaWsfcRole` queries `MSCluster_ResourceGroup`, which does carry `State`, so it now loops and maps each role's own value. Those are resource *group* codes, not resource codes, so it uses `Get-ResourceGroupState` - which moves out of `Get-DbaWsfcResourceGroup` into `private/functions` rather than being duplicated in two files.

For the three classes that have no `State` property, the NoteProperty is removed along with its `.OUTPUTS` documentation, and `Get-DbaWsfcCluster` loses the always-empty `State` column from its default view. `Get-DbaWsfcSharedVolume` now queries `MSCluster_ClusterSharedVolume`.

Verified with `Get-CimClass` against a live cluster: `MSCluster_Cluster`, `MSCluster_AvailableDisk` and `MSCluster_ClusterSharedVolume` expose no `State`, and `ClusterSharedVolume` is not a concrete class in the namespace.

## Fixes #10573: where is the witness

The reporter of #10573 could not find the UNC path of a file share witness, and looked in the obvious place - `Get-DbaWsfcCluster`. That instinct was right and the command simply did not carry it: `MSCluster_Cluster.QuorumPath` is filled for a *disk* witness and stays empty for a file share witness, whose path exists only in the private properties of the witness resource.

`Get-DbaWsfcCluster` now has a `WitnessPath` property that answers the question for whichever witness a cluster uses:

| Cluster | QuorumType | QuorumPath | WitnessPath |
|---|---|---|---|
| CLUSTER01 | Node and Disk Majority | `Q:\Cluster\` | `Q:\Cluster\` |
| CLUSTER02 | Node and File Share Majority | *(empty)* | `\\fs\ClusterWitness02` |

```powershell
(Get-DbaWsfcCluster -ComputerName cluster02).WitnessPath
```

The witness resource is queried only when `QuorumTypeValue` says there is a file share witness, so no other cluster pays for it. Measured against live clusters: a disk witness cluster stays at 68 ms per call, a file share witness cluster goes to 138 ms.

An `MSCluster_ClusterToQuorumResource` association would additionally name the quorum resource for every quorum type, but it costs 136 ms against a 46 ms base query, on a command that eight other `Get-DbaWsfc*` commands call internally. That was measured and rejected - not worth it for a name.

Cloud witness is not covered. Such a cluster keeps an empty `WitnessPath` rather than failing, and there was no cloud witness to test against.

`Get-DbaWsfcResource` and `Get-DbaWsfcResourceGroup` behave exactly as before. No CSV or available-disk state is synthesised from associated resources - `MSCluster_ClusterSharedVolumeToResource` would allow it, but that is new behaviour, not a fix.

## Tests

The reason these bugs survived since 2018 is that no test ever ran against a real cluster. This PR adds integration coverage for all five commands, verified against a two-node Windows Server 2025 cluster: **30 tests pass, and six of them fail without the fix**, each naming its own defect - including `Get-DbaWsfcSharedVolume` returning nothing at all with a cluster shared volume present. The witness test compares `WitnessPath` against the `SharePath` that `Get-DbaWsfcResource` reports, rather than against a hard-coded path, so it verifies the plumbing and not one particular lab.

No CI runner can host a Windows failover cluster, so the tests read the cluster name from `$TestConfig.ClusterStorage`, which `Get-TestConfig` defines as `$null` and only a local configuration sets. The integration `Context` skips when it is empty.

That is an exception to the rule in `tests/CLAUDE.md` that a missing boundary must fail rather than skip, so the guide now documents it as a narrow, named exception, along with the fixtures such a lab needs: a cluster shared volume, a shared disk kept out of the cluster, and a file share witness. Without those two disks the commands return nothing and cannot fail, which is exactly how the wrong class name went unnoticed.

---

text by Claude, reviewed by Andreas Jordan

### DIFF
--- a/private/functions/Get-ResourceGroupState.ps1
+++ b/private/functions/Get-ResourceGroupState.ps1
@@ -1,0 +1,9 @@
+function Get-ResourceGroupState ($state) {
+    switch ($state) {
+        -1 { "Unknown" }
+        0 { "Online" }
+        1 { "Offline" }
+        2 { "Failed" }
+        default { $state }
+    }
+}
--- a/private/testing/Get-TestConfig.ps1
+++ b/private/testing/Get-TestConfig.ps1
@@ -17,6 +17,14 @@ function Get-TestConfig {
         # When testing a remote SQL Server instance this must be a network share
         # where both the SQL Server instance and the test script can write to.
         Temp             = 'C:\Temp'
+        # The Windows failover clusters that the Get-DbaWsfc* tests need. No CI environment has a
+        # cluster, so these stay empty there and the cluster tests skip themselves. A local
+        # configuration whose lab has clusters sets them to the cluster names.
+        # ClusterStorage needs shared storage: one cluster shared volume, and one disk that every
+        # node can see but that is deliberately not part of the cluster.
+        ClusterStorage   = $null
+        # ClusterWitness needs a file share witness.
+        ClusterWitness   = $null
     }
 
     if (Test-Path $LocalConfigPath) {
--- a/public/Get-DbaWsfcAvailableDisk.ps1
+++ b/public/Get-DbaWsfcAvailableDisk.ps1
@@ -37,7 +37,6 @@ function Get-DbaWsfcAvailableDisk {
         All properties from the MSCluster_AvailableDisk WMI class are returned, including:
 
         Properties added via Add-Member:
-        - State: Current operational state of the disk
         - ClusterName: Name of the cluster
         - ClusterFqdn: Fully qualified domain name of the cluster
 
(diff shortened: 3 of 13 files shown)


potatoqualitee:
nice change 🙏🏼

PR #10589, opened 2026-08-20
Add Find-DbaDbQueryStoreRegression

deepeshd87:
## Type of Change
- [x] Unit test is included

### Purpose

dbatools has strong Query Store *configuration* coverage
(Get/Set/Copy-DbaDbQueryStoreOption, Test-DbaDbQueryStore) but nothing in-module
that *analyzes* Query Store runtime statistics to find performance regressions.
This command fills that gap.

The concept and the `Find-` verb were checked with @niphlod in the #dbatools
Slack channel before building.

### Approach

`Find-DbaDbQueryStoreRegression` reads sys.query_store_runtime_stats and flags
queries that have regressed, using an execution-weighted baseline: each query's
historical duration is weighted by execution count over a baseline window, then
compared against a current window, filtering out low-frequency and low-total-duration
noise.

Detection is at the **query level** (aggregating across plans), not the plan level.
This is deliberate — it catches plan-flip regressions, the common case where a query
degrades because the optimizer switched to a worse plan, which a plan-level comparison
would miss. A `PlanChanged` column flags those cases.

It is read-only (queries DMVs, returns objects, changes nothing) and requires
SQL Server 2016+.

### Commands to test

Find-DbaDbQueryStoreRegression -SqlInstance sql2019 -Database YourDb
Find-DbaDbQueryStoreRegression -SqlInstance sql2019 -Database YourDb -SlowdownThreshold 2.0 -MinExecutionCount 50

### DIFF
--- a/dbatools.psd1
+++ b/dbatools.psd1
@@ -170,6 +170,7 @@
         'Find-DbaDbDisabledIndex',
         'Find-DbaDbDuplicateIndex',
         'Find-DbaDbGrowthEvent',
+        'Find-DbaDbQueryStoreRegression',
         'Find-DbaDbUnusedIndex',
         'Find-DbaInstance',
         'Find-DbaLoginInGroup',
--- a/dbatools.psm1
+++ b/dbatools.psm1
@@ -539,6 +539,7 @@ $script:xplat = @(
         'Get-DbaQueryExecutionTime',
         'Get-DbaTempdbUsage',
         'Find-DbaDbGrowthEvent',
+        'Find-DbaDbQueryStoreRegression',
         'Test-DbaLinkedServerConnection',
         'Get-DbaDbFile',
         'Get-DbaDbFileGrowth',
--- a/public/Find-DbaDbQueryStoreRegression.ps1
+++ b/public/Find-DbaDbQueryStoreRegression.ps1
@@ -1,0 +1,211 @@
+function Find-DbaDbQueryStoreRegression {
+    <#
+    .SYNOPSIS
+        Detects query performance regressions from Query Store runtime statistics using an
+        execution-weighted baseline.
+
+    .DESCRIPTION
+        Reads sys.query_store_runtime_stats and flags queries whose recent performance has
+        regressed against a historical baseline.
+
+        Rather than a naive average-vs-average comparison, each plan's historical duration is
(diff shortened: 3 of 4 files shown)


deepeshd87:
Thanks for the review. The CI flagged my test file structure,  looks like it doesn't match the current test scaffold (the -ForEach discovery pattern, the mandatory header, and the retired TestConfig instance names). I've pushed a fix addressing those three, but I want to make sure I match your current conventions exactly. Is there a current test template I should base this on, or an existing recent command's .Tests.ps1 you'd point me to? Happy to align it however you prefer.

potatoqualitee:
Thank you, @deepeshd87 ! This is nice, I'll throw Claude at it to address the issues including needing to add it to the psm1 and psd1.

deepeshd87:
Thank you so much, that's very kind. Really appreciate you taking it the rest of the way. Happy to make any changes if anything comes up. Just let me know.

potatoqualitee:
Pushed 675748b to your branch with the fixes, @deepeshd87 - thanks again for this one.

- Registered the command in `dbatools.psd1` and `dbatools.psm1`.
- Test file moved onto the current scaffold (`tests/CLAUDE.md`): static `$CommandName` header, no `-ForEach` fed from `BeforeAll`, `$TestConfig.InstanceSingle`. The placeholder test is replaced by real ones against a Query Store-on and a Query Store-off database (SQL 2022 turns Query Store on for new databases, so "off" is set explicitly).
- Your `Get-DbaDatabase | Find-DbaDbQueryStoreRegression` example could not bind - `DbaInstanceParameter` does not accept Database objects - so the command now has the usual `-InputObject` for piped databases (instances still pipe into `-SqlInstance`), and each instance's databases are analyzed before the next instance connects.
- Added the `.OUTPUTS` help block and a couple of style nits (double quotes, hashtable alignment).

Locally: 7/7 command tests against SQL 2022 and the full `dbatools.Tests.ps1` compliance suite green. CI will confirm.

potatoqualitee:
thanks again, your command will be available in the next release 🙏🏼

deepeshd87:
Thank you so much, @potatoqualitee - this made my week. I really appreciate you taking the time to polish it up and walk it through. Learned a lot from your changes (the -InputObject piping especially). Looking forward to contributing more.

PR #10591, opened 2026-08-20
Invoke-DbaDbDecryptObject - reject ambiguous ciphertext chunks

potatoqualitee:
## Summary

Follow-up hardening after #10581.

The raw-page reader is intentionally defensive because stale or malformed physical records can otherwise look plausible. This adds a final invariant at the decryption boundary: one object may not contain the same `colid` more than once, and a chunk may not arrive without ciphertext.

A duplicate `(object id, colid)` is ambiguous. Concatenating both values can combine current ciphertext with a stale or duplicate physical row and still produce output that looks trustworthy, so the command now refuses that object instead of returning a definition.

## Changes

- Reject duplicate `colid` values in `ConvertFrom-EncryptedObjectChunk`.
- Reject null ciphertext at the same boundary rather than relying only on the caller to filter it.
- Add unit tests for duplicate chunks, null ciphertext, and the valid byte-exact path.

## Scope

This is deliberately narrow. The review after #10581 also identified broader defensive ideas around export-path containment, allocated/ghost page filtering, slot-pointer bounds, and the database-wide `dm_db_database_page_allocations(..., 'DETAILED')` fallback. Those touch legacy export behavior or undocumented physical-page assumptions and should be handled separately with targeted fixtures rather than folded into this small invariant fix.

## Type of Change

- [x] Bug fix / hardening
- [x] Unit tests
- [ ] Breaking change

Follow-up to #10581.

### DIFF
--- a/private/functions/ConvertFrom-EncryptedObjectChunk.ps1
+++ b/private/functions/ConvertFrom-EncryptedObjectChunk.ps1
@@ -15,13 +15,16 @@ function ConvertFrom-EncryptedObjectChunk {
         so a byte for byte comparison against a created object cannot reach this code with more than one
         chunk. The multi chunk behaviour of this function is covered by its unit tests and by nothing else.
 
-        Two mistakes are easy to make here and both produce text of exactly the right length, which is the
-        one failure mode that the rest of the suite cannot see:
+        Three mistakes are easy to make here and all can produce output that looks more trustworthy than it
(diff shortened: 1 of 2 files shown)


potatoqualitee:
@claude why didnt my azure stuff run? is it bc it's queued up? if not, why isnt it showing as a queue?

Issue #10560, opened 2026-08-11
Get-DbaDbQueryStoreOption - Returns cached values, and discards two it queried on SQL Server 2019 and later


andreasjordan:
## Summary

Two defects that reinforce each other:

1. The command never refreshes the SMO object it reads from, so every value that comes from `$db.QueryStoreOptions` is as old as the connection. Pass a server object that was used before and the command reports what Query Store looked like back then.
2. On SQL Server 2019 and later it queries `max_plans_per_query` and `wait_stats_capture_mode_desc` from `sys.database_query_store_options` and then throws both away. Only the four `CustomCapturePolicy*` values from that query reach the output, so `MaxPlansPerQuery` and `WaitStatsCaptureMode` fall back to the cached SMO object even though the correct values were already fetched. The SQL Server 2017 branch right above does add them.

The second one makes the first visible even to callers who never reuse a connection themselves, because `Set-DbaDbQueryStoreOption` returns the object that `Get-DbaDbQueryStoreOption` produces: it reports the old values for a change it has just made.

## Steps to Reproduce

```powershell
$server = Connect-DbaInstance -SqlInstance $instance
$null = Set-DbaDbQueryStoreOption -SqlInstance $server -Database $dbName -State ReadWrite

$before = Get-DbaDbQueryStoreOption -SqlInstance $server -Database $dbName

# Change the settings from a separate connection
$splatChange = @{
    SqlInstance = $instance
    Database    = "master"
    Query       = "ALTER DATABASE [$dbName] SET QUERY_STORE = ON (MAX_PLANS_PER_QUERY = 321, WAIT_STATS_CAPTURE_MODE = OFF)"
}
$null = Invoke-DbaQuery @splatChange

Get-DbaDbQueryStoreOption -SqlInstance $server -Database $dbName    # the connection from above
Get-DbaDbQueryStoreOption -SqlInstance $instance -Database $dbName  # a fresh connection

# And what the command hands back for a change it made itself
Set-DbaDbQueryStoreOption -SqlInstance $server -Database $dbName -MaxPlansPerQuery 444 -WaitStatsCaptureMode On
```

Output, identical on SQL Server 2019 (v15) and SQL Server 2022 (v16), dbatools 2.8.4 and current `development`:

```
before change, same server object : MaxPlansPerQuery=200 WaitStatsCaptureMode=On
instance really has now           : MaxPlansPerQuery=321 WaitStatsCaptureMode=OFF
[...]

PR #10592, opened 2026-08-21
Get-DbaDbQueryStoreOption - Read from SMO what SMO carries, and query only the rest

### LINKED ISSUE #10562: Get-DbaDbQueryStoreOption - Queries MaxPlansPerQuery and WaitStatsCaptureMode on SQL Server 2019 and later, then reports something else
## Summary

For SQL Server 2017 and later the command runs a query against `sys.database_query_store_options` because some values are not read back through SMO. On SQL Server 2019 and later that query fetches six values, but only the four `CustomCapturePolicy*` ones are added to the output object. `MaxPlansPerQuery` and `WaitStatsCaptureMode` are dropped, so the two properties are reported from the SMO object instead - out of a query that has already fetched the correct values and runs either way.

The SQL Server 2017 branch directly above adds all of them. So the same command reports live values on 2017 and whatever SMO holds on 2019 and later.

This is not about refreshing anything - see #10560, which was closed for going that way. It is about which of two values that are already in hand gets used.

## Steps to Reproduce

```powershell
$null = Set-DbaDbQueryStoreOption -SqlInstance $server -Database $dbName -State ReadWrite

# Change the settings from a separate connection
$splatChange = @{
    SqlInstance = $instance
    Database    = "master"
    Query       = "ALTER DATABASE [$dbName] SET QUERY_STORE = ON (MAX_PLANS_PER_QUERY = 321, WAIT_STATS_CAPTURE_MODE = OFF)"
}
$null = Invoke-DbaQuery @splatChange

# What the command itself queries internally
$splatRead = @{
    SqlInstance = $instance
    Database    = $dbName
    Query       = "SELECT max_plans_per_query AS MaxPlansPerQuery, wait_stats_capture_mode_desc AS WaitStatsCaptureMode FROM sys.database_query_store_options"
}
Invoke-DbaQuery @splatRead

# What it reports
Get-DbaDbQueryStoreOption -SqlInstance $server -Database $dbName | Select-Object MaxPlansPerQuery, WaitStatsCaptureMode
```

Output on SQL Server 2019 (v15) and SQL Server 2022 (v16), dbatools 2.8.4 and current `development`:

```
the query the command runs      : MaxPlansPerQuery=321 WaitStatsCaptureMode=OFF
what the command reports        : MaxPlansPerQuery=200 WaitStatsCaptureMode=On
```

A connection that was used before makes this visible, but the discarded values are discarded on every call.

## Root cause

[`Get-DbaDbQueryStoreOption.ps1:134`](https://github.com/dataplat/dbatools/blob/development/public/Get-DbaDbQueryStoreOption.ps1#L134) queries six values for SQL Server 2019 and later, and [`:148-153`](https://github.com/dataplat/dbatools/blob/development/public/Get-DbaDbQueryStoreOption.ps1#L148-L153) adds four of them. The SQL Server 2017 branch at [`:144-147`](https://github.com/dataplat/dbatools/blob/development/public/Get-DbaDbQueryStoreOption.ps1#L144-L147) adds `MaxPlansPerQuery` and `WaitStatsCaptureMode` from the same query. Both properties stay in `Select-DefaultView`, so they are displayed either way - just from a different source depending on the version.

## Two ways out

1. **Add the two `Add-Member` lines to the SQL Server 2019 branch.** Both branches then report these two properties from the query, at no extra cost, since the query runs anyway.
2. **Drop the two columns from the SQL Server 2019 query.** Then SMO is the single source on 2019 and later - but 2017 still reports them from its query, so the inconsistency between the branches remains unless that branch goes too.

Option 1 keeps the two versions in step and is one line each. Option 2 is only complete if SMO delivers these values on SQL Server 2017 as well, which is presumably why that branch exists in the first place. Settling that needs a SQL Server 2017 instance, which the lab used for this report does not have - it runs 2019, 2022 and 2025.

## Test

Per `tests/CLAUDE.md`, whichever way it goes: read the options, change `MAX_PLANS_PER_QUERY` from a separate connection, and assert what the command reports through a server object that was used before.

---

andreasjordan:
## Type of Change
 - [x] Bug fix (non-breaking change, fixes #10562)
 - [x] Pester test is included

### Purpose

`MaxPlansPerQuery` and `WaitStatsCaptureMode` are properties of the SMO `QueryStoreOptions` object, and the command queried them from `sys.database_query_store_options` as well:

- On **SQL Server 2019 and later** the queried values were then thrown away. Six columns were fetched, four were added to the output, and those two were left to the SMO object. The query paid for two columns nobody read.
- On **SQL Server 2017** they were added with `Add-Member`, which is worse than it looks.

### Why Add-Member is the wrong tool for these two

Adding a note property over a **real** property does not add anything, it replaces it on the object of the caller - and the note property keeps its value even after the object is refreshed:

```
member kind before Get   : Property
member kind after Get    : NoteProperty
value after SMO Refresh  : 200   (the instance has 999)
```

The four `CustomCapturePolicy*` values are a different case: they are not properties of `QueryStoreOptions` at all, so adding them hides nothing.

```
--- before Get, on development ---
MaxPlansPerQuery                         Property
WaitStatsCaptureMode                     Property
CustomCapturePolicyExecutionCount        <none>
CustomCapturePolicyStaleThresholdHours   <none>
```

### Approach

The catalog view is read only for what SMO does not carry, which is the four `CustomCapturePolicy*` values. SMO is the single source for the other two on every version, and the SQL Server 2017 query is gone entirely, saving a round trip there.

Refreshing inside a `Get-` command was rejected in #10560 and is not revisited here.

### What changes for users

**Nothing on SQL Server 2019 and later.** Those two already came from SMO; only the two unread columns are gone from the query.

**On SQL Server 2017** they now come from SMO like every other value in the output, so a connection that has read them before reports what it read then - exactly as it already does for `MaxStorageSizeInMB`, `StaleQueryThresholdInDays` and the rest. In exchange, the command no longer replaces a live property on the caller's object with a frozen copy.

### Tests

Assertions that `MaxPlansPerQuery` and `WaitStatsCaptureMode` stay properties of the SMO object, and that `CustomCapturePolicyExecutionCount` is a note property, which is the rule this PR follows.

**Being explicit about what that test proves:** it guards the rule, it is not a regression test. It cannot fail on SQL Server 2019 and later, because nothing observable changed there, and the change that *is* observable is on SQL Server 2017, which the lab used for this work does not have - the open question on #10562. If someone applies `Add-Member` to these two again on any version, it fails.

`Get-DbaDbQueryStoreOption`: 7 tests, all passing, no leftovers in the lab.

---

### DIFF
--- a/public/Get-DbaDbQueryStoreOption.ps1
+++ b/public/Get-DbaDbQueryStoreOption.ps1
@@ -128,10 +128,15 @@ function Get-DbaDbQueryStoreOption {
                 Write-Message -Level Verbose -Message "Processing $($db.Name) on $instance"
                 $qso = $db.QueryStoreOptions
 
-                if ($server.VersionMajor -eq 14) {
-                    $QueryStoreOptions = Invoke-DbaQuery -SqlInstance $server -Database $db.Name -Query "SELECT max_plans_per_query AS MaxPlansPerQuery, wait_stats_capture_mode_desc AS WaitStatsCaptureMode FROM sys.database_query_store_options;" -As PSObject
-                } elseif ($server.VersionMajor -ge 15) {
-                    $QueryStoreOptions = Invoke-DbaQuery -SqlInstance $server -Database $db.Name -Query "SELECT max_plans_per_query AS MaxPlansPerQuery, wait_stats_capture_mode_desc AS WaitStatsCaptureMode, capture_policy_execution_count AS CustomCapturePolicyExecutionCount, capture_policy_stale_threshold_hours AS CustomCapturePolicyStaleThresholdHours, capture_policy_total_compile_cpu_time_ms AS CustomCapturePolicyTotalCompileCPUTimeMS, capture_policy_total_execution_cpu_time_ms AS CustomCapturePolicyTotalExecutionCPUTimeMS FROM sys.database_query_store_options;" -As PSObject
+                # Only values that the SMO object does not carry are read from the catalog view. The four
+                # CustomCapturePolicy ones are not properties of QueryStoreOptions at all, so they have to
+                # be added. MaxPlansPerQuery and WaitStatsCaptureMode are, and used to be queried here as
+                # well - on SQL Server 2019 and later the result was then thrown away, and on SQL Server
+                # 2017 it was added over the real property, which turns it into a note property that no
+                # longer follows a Refresh of the object. SMO is the single source for those two now, on
+                # every version. See #10562.
+                if ($server.VersionMajor -ge 15) {
+                    $QueryStoreOptions = Invoke-DbaQuery -SqlInstance $server -Database $db.Name -Query "SELECT capture_policy_execution_count AS CustomCapturePolicyExecutionCount, capture_policy_stale_threshold_hours AS CustomCapturePolicyStaleThresholdHours, capture_policy_total_compile_cpu_time_ms AS CustomCapturePolicyTotalCompileCPUTimeMS, capture_policy_total_execution_cpu_time_ms AS CustomCapturePolicyTotalExecutionCPUTimeMS FROM sys.database_query_store_options;" -As PSObject
                 }
 
                 Add-Member -Force -InputObject $qso -MemberType NoteProperty -Name ComputerName -Value $server.ComputerName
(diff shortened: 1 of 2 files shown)


potatoqualitee:
thank you 🙏🏼

PR #10593, opened 2026-08-21
Set-DbaDbQueryStoreOption - Refresh the Query Store options after changing them with T-SQL

### LINKED ISSUE #10561: Set-DbaDbQueryStoreOption - Returns the old values for the settings it changes with T-SQL
## Summary

Most Query Store settings are changed through SMO, but `MaxPlansPerQuery`, `WaitStatsCaptureMode` and the four `CustomCapturePolicy*` options are changed with an `ALTER DATABASE` statement. After that statement the SMO object is not refreshed, so the object the command returns still carries the values from before the change.

This is not the SMO caching that a reused connection brings with it - it happens on every call, even when the command opens the connection itself, because the command reads `$db.QueryStoreOptions` before it runs the T-SQL and that populates the cache.

Follow-up to #10560, which went at this from the wrong end.

## Steps to Reproduce

Every connection here is opened by the command itself:

```powershell
$null = New-DbaDatabase -SqlInstance $instance -Name $dbName
$null = Set-DbaDbQueryStoreOption -SqlInstance $instance -Database $dbName -State ReadWrite

$result = Set-DbaDbQueryStoreOption -SqlInstance $instance -Database $dbName -MaxPlansPerQuery 555 -WaitStatsCaptureMode Off
$result | Select-Object MaxPlansPerQuery, WaitStatsCaptureMode

Invoke-DbaQuery -SqlInstance $instance -Database $dbName -Query "SELECT max_plans_per_query, wait_stats_capture_mode_desc FROM sys.database_query_store_options"
```

```
returned by Set : MaxPlansPerQuery=200 WaitStatsCaptureMode=On     <-- the values from before the change
instance really has: MaxPlansPerQuery=555 WaitStatsCaptureMode=OFF
```

Verified on SQL Server 2019 and 2022, with dbatools 2.8.4 and current `development`. SQL Server 2016 is not affected, because none of these options exist there.

`Copy-DbaDbQueryStoreOption` runs into the same thing, because it copies the settings by calling this command.

## Root cause

The command already does the right thing for the SMO path - [`Set-DbaDbQueryStoreOption.ps1:247-249`](https://github.com/dataplat/dbatools/blob/development/public/Set-DbaDbQueryStoreOption.ps1#L247-L249):

```powershell
$db.QueryStoreOptions.DesiredState = $State
$db.QueryStoreOptions.Alter()
$db.QueryStoreOptions.Refresh()
```

The T-SQL path further down never got the same treatment. `$db.Refresh()` at [`:344`](https://github.com/dataplat/dbatools/blob/development/public/Set-DbaDbQueryStoreOption.ps1#L344) runs *before* the statement is executed at [`:347`](https://github.com/dataplat/dbatools/blob/development/public/Set-DbaDbQueryStoreOption.ps1#L347), and nothing reads the values back afterwards. The trailing `Get-DbaDbQueryStoreOption` at [`:355`](https://github.com/dataplat/dbatools/blob/development/public/Set-DbaDbQueryStoreOption.ps1#L355) then reports what the cached object holds.

## Suggested fix

Refresh right after the statement was executed:

```powershell
if ($query -ne "") {
    $db.Query($query, $dbName)
    $db.QueryStoreOptions.Refresh()
}
```

Verified on SQL Server 2022:

```
returned by Set for its own change: MaxPlansPerQuery=444 WaitStatsCaptureMode=On     <-- was 200/On
instance really has now           : MaxPlansPerQuery=444 WaitStatsCaptureMode=ON
```

That is one round trip per database, and only for the calls that actually use one of these options. What a reused connection reports about changes made elsewhere is unaffected, as it should be.

Note for whoever picks this up: #10558 changes the line that executes the statement, so the refresh belongs after whatever that line looks like at the time.

## Test

Per `tests/CLAUDE.md`, a regression test that sets `-MaxPlansPerQuery` and `-WaitStatsCaptureMode` and expects the returned object to carry the values that were just set. No connection reuse needed - the old behaviour fails this with a fresh connection as well.

---

andreasjordan:
## Type of Change
 - [x] Bug fix (non-breaking change, fixes #10561)
 - [x] Pester test is included

### Purpose

`MaxPlansPerQuery`, `WaitStatsCaptureMode` and the four `CustomCapturePolicy*` options are changed with an `ALTER DATABASE` statement rather than through SMO. The SMO object knows nothing about that and still holds what it read before, so the command reported the values from **before its own change** and left the object of the caller saying the same:

```
returned by Set     : MaxPlansPerQuery=200
the instance has    : MaxPlansPerQuery=555
```

This is not the staleness that a reused connection brings with it. It happens on every call, including one where the command opens the connection itself, because the command reads `$db.QueryStoreOptions` before it runs the statement and that fills the cache.

`Copy-DbaDbQueryStoreOption` is fixed with it, because it copies the settings by calling this command.

### Approach

The SMO path a few lines above already does the right thing:

```powershell
$db.QueryStoreOptions.DesiredState = $State
$db.QueryStoreOptions.Alter()
$db.QueryStoreOptions.Refresh()
```

The T-SQL path now gets the same treatment, right after the statement:

```powershell
if ($query -ne "") {
    $null = $server.Query($query)
    $db.QueryStoreOptions.Refresh()
}
```

That is one round trip per database, and only for the calls that actually use one of these options. What a reused connection reports about changes made elsewhere is unaffected, as it should be.

### Commands to test

```powershell
$null = Set-DbaDbQueryStoreOption -SqlInstance $server -Database $dbName -State ReadWrite

Set-DbaDbQueryStoreOption -SqlInstance $server -Database $dbName -MaxPlansPerQuery 555 -WaitStatsCaptureMode Off |
    Select-Object MaxPlansPerQuery, WaitStatsCaptureMode      # 555 / Off, was 200 / On
```

### Tests

A `Context` that sets both options and expects:

- the returned object to carry them
- the SMO object of the caller to carry them, which is the part a `Get-` command cannot paper over
- the instance to really have them, which passes either way and says the T-SQL was never the problem

The first two fail against `development` with `Expected 555, but got 200`.

`Set-DbaDbQueryStoreOption`: 11 tests, all passing, no leftovers in the lab.

### Note on the test fixture

The new `Context` uses the database that the `Describe` already creates rather than one of its own. A `CREATE DATABASE` at that point fails intermittently with

```
Could not obtain exclusive lock on database 'model'. Retry the operation later.
```

because the tests above it read Query Store on `model` and leave a session parked in it. That is the leak #10584 fixes, which is not merged yet - worth knowing that it now reaches test files that have nothing to do with connection handling.

---

### DIFF
--- a/public/Set-DbaDbQueryStoreOption.ps1
+++ b/public/Set-DbaDbQueryStoreOption.ps1
@@ -345,6 +345,12 @@ function Set-DbaDbQueryStoreOption {
 
                         if ($query -ne "") {
                             $null = $server.Query($query)
+                            # These options are changed with T-SQL rather than through SMO, so the SMO
+                            # object knows nothing about it and still holds what it read before. Without
+                            # this the command reports the values from before its own change, and leaves
+                            # the object of the caller saying the same. It is the treatment the SMO path
+                            # above already gets after its Alter. See #10561.
+                            $db.QueryStoreOptions.Refresh()
                         }
                     } catch {
                         Stop-Function -Message "Could not modify configuration." -Category InvalidOperation -InnerErrorRecord $_ -Target $db -Continue
--- a/tests/Set-DbaDbQueryStoreOption.Tests.ps1
+++ b/tests/Set-DbaDbQueryStoreOption.Tests.ps1
@@ -37,6 +37,14 @@ Describe $CommandName -Tag UnitTests {
 }
 
 Describe $CommandName -Tag IntegrationTests {
+    BeforeDiscovery {
+        # MaxPlansPerQuery and WaitStatsCaptureMode arrived with SQL Server 2017. The value decides a
+        # Skip, which Pester needs while it discovers the tests, so it cannot be read in BeforeAll.
+        $discoveryServer = Connect-DbaInstance -SqlInstance $TestConfig.InstanceMulti1
+        $multi1VersionMajor = $discoveryServer.VersionMajor
+        $null = $discoveryServer | Disconnect-DbaInstance
+    }
+
     BeforeAll {
         Get-DbaDatabase -SqlInstance $TestConfig.InstanceMulti1, $TestConfig.InstanceMulti2 | Where-Object Name -Match "dbatoolsci" | Remove-DbaDatabase
         New-DbaDatabase -SqlInstance $TestConfig.InstanceMulti1, $TestConfig.InstanceMulti2 -Name dbatoolsciqs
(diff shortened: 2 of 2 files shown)


potatoqualitee:
looks good 👌🏼  ty

PR #10596, opened 2026-08-22
Copy-DbaAgentSchedule - Make -Force actually drop and recreate existing schedules

### LINKED ISSUE #10594: Copy-DbaAgentSchedule - -Force never drops and recreates existing schedules

No error. `$error[0]` is empty. `-Force` just silently does nothing.

The help for `-Force` (lines 43-45) says it "Forces the overwrite of existing schedules on the destination instances by dropping and recreating them". The drop never runs. And if the existing destination schedule has no jobs attached, nothing is emitted for it either, so the no-op doesn't show up in the output you would check afterwards.

### Steps to Reproduce

```powershell
# Schedule 'NightlyETL' exists on both instances. On sql02 it has no jobs attached.
# Expected per .PARAMETER Force: dropped and recreated on sql02.
# Actual: nothing emitted for it, destination schedule unchanged.
Copy-DbaAgentSchedule -Source sql01 -Destination sql02 -Force
```

To be straight about provenance: I found this by reading the source, not from a live migration, so the block above is the scenario the code path describes rather than a transcript of a run I made. The control-flow walk in "Other details" is verifiable from a clone at the refs I list there. Happy to run it against a live pair if that would help confirm.

```
Get-Module dbatools -ListAvailable | Select -ExpandProperty Version

Major  Minor  Build  Revision
-----  -----  -----  --------
2      1      5      -1
```

So no, my installed copy is not the latest. I checked the source at the published refs rather than relying on it.

The defect is present on the current release. public/Copy-DbaAgentSchedule.ps1 is byte-identical on v2.8.1, v2.8.2 and v2.8.4 (current, published 2026-07-31) and on development at 79e624ab7cb99f377766e81aa67cbf374f84758c. The line numbers I cite below are from that file.

It is also present in the 2.1.5 I have installed. That file is not identical to current (it predates the `$destServer.JobServer.SharedSchedules.Refresh()` call and some surrounding restructuring), but the `-Force` arm is the same shape: the `ShouldProcess` wrapper at line 151, the unconditional `continue` at 158 and the unreachable drop at 160-163 correspond to lines 168, 175 and 177-180 in the current file.

Happy to upgrade and re-confirm on the latest if that would help.

### Other details or mentions

The relevant part of public/Copy-DbaAgentSchedule.ps1:

```text
123  if ($Force) { $ConfirmPreference = 'none' }              # begin block
158  if ($destSchedules.Name -contains $scheduleName) {
159      if ($Force -ne $true) { ...skip and report... continue }
167      } else {                                             # -Force
168          if ($Pscmdlet.ShouldProcess($destinstance, "Schedule [$scheduleName] has associated jobs. Skipping.")) {
169              if (...JobSchedules.Name -contains $scheduleName) {    # only reports if jobs exist
175              continue                                     # always, jobs or not
176          } else {                                         # only when ShouldProcess said no
177              if ($Pscmdlet.ShouldProcess($destinstance, "Dropping schedule $scheduleName and recreating")) {
180                  ...Drop()
```

In a normal run `ShouldProcess` returns true, so line 175 continues and the drop block at 177-186 is never reached. Line 123 makes that certain: `-Force` sets `$ConfirmPreference = 'none'`, so the line 168 `ShouldProcess` cannot prompt and cannot return false. The only way I can find to reach `Drop()` is `-Force -Confirm`, answering No to "has associated jobs. Skipping." and then Yes to "Dropping schedule and recreating", which nobody is going to do by accident.

I think this is a regression rather than something that was never written. Before 9f67479581 (2018-07-18, "hid object output in whatif") the `-Force` arm was:

```powershell
else {                                                    # -Force
    if ($destServer.JobServer.Jobs.JobSchedules.Name -contains $scheduleName) {
        ...report Skipped...
        continue                                          # has jobs, skip
    }
    else {                                                # no jobs
        if ($Pscmdlet.ShouldProcess($destinstance, "Dropping schedule $scheduleName and recreating")) {
            $destServer.JobServer.SharedSchedules[$scheduleName].Drop()
        }
    }
}
```

That commit wrapped the associated-jobs test in a `ShouldProcess` call. The `else` that had belonged to the jobs test ended up attached to the new `ShouldProcess` instead, so "no associated jobs, drop it" became "user declined the prompt, drop it". Given the commit title was about WhatIf output, I doubt that was intended.

Two smaller things from the same structure. Schedules with no attached jobs produce no MigrationObject at all, because the reporting block at 170-173 is inside the jobs check while the `continue` at 175 is outside it. That `continue` moved to its current position in 2af6205cbf ("Standardize Whatif Output", #8846, 2023-04-09). And under `-WhatIf` both `ShouldProcess` messages print, so the output claims the schedule "has associated jobs" before anything has checked whether it does, and then claims it is "Dropping schedule ... and recreating" when it isn't.

For a fix, restoring the pre-9f67479581 shape would do it: check for associated jobs first with no `ShouldProcess` around the read, report Skipped and continue if there are any (the "cannot be overwritten, even with Force" case in help line 45), otherwise `ShouldProcess("Dropping schedule $scheduleName and recreating")`, drop, and fall through to the existing create block at line 192.

For transparency: this was surfaced by static analysis of the module, then verified by hand against the source and the git history before reporting.

Windows PowerShell (powershell.exe), Windows PowerShell ISE (powershell_ise.exe)

### PowerShell Host Version

```
PSVersion                      5.1.26100.9168
BuildVersion                   10.0.26100.9168
```

SQL Server 2022.

I have not pasted `SELECT @@VERSION` output because no live repro was run for this report (see Steps to Reproduce) — 2022 is simply the version I have available. The behaviour described is in the PowerShell control flow of the command and does not depend on the SQL Server version, edition or build. Happy to provide full `@@VERSION` output from a repro pair if you would like this confirmed live.

### .NET Framework Version

```
.NET Framework 4.8.9337.0
```

andreasjordan:
fixes #10594.

## Problem

`-Force` on `Copy-DbaAgentSchedule` never dropped anything. It reported nothing, wrote no warning, and left the destination schedule as it was. Reported by @greenmtnsun in #10594 with a complete and correct control-flow analysis.

## Mechanism

In the `-Force` arm, the test for associated jobs was made *inside* a `ShouldProcess` call, and the `continue` after it was unconditional:

```powershell
if ($Pscmdlet.ShouldProcess($destinstance, "Schedule [$scheduleName] has associated jobs. Skipping.")) {
    if ($destServer.JobServer.Jobs.JobSchedules.Name -contains $scheduleName) { ...report Skipped... }
    continue                       # always, jobs or not
} else {
    if ($Pscmdlet.ShouldProcess($destinstance, "Dropping schedule $scheduleName and recreating")) { ...Drop()... }
}
```

`ShouldProcess` returns true on any normal run, so the `continue` fires and the `Drop()` in the `else` is unreachable. `if ($Force) { $ConfirmPreference = 'none' }` in the `begin` block makes that certain: with `-Force` the prompt cannot appear and cannot return false. The only route to `Drop()` was `-Force -Confirm`, answering **No** to "has associated jobs. Skipping." and then **Yes** to "Dropping schedule and recreating".

A second consequence of the same shape: a destination schedule with no attached jobs emitted no `MigrationObject` at all, because the reporting block was inside the jobs test while the `continue` was outside it.

This is a regression from 9f67479581 (2018-07-18, "hid object output in whatif"). That commit wrapped the associated-jobs test in `ShouldProcess`, and the `else` that had belonged to the jobs test ended up attached to the new `ShouldProcess` instead — "no associated jobs, drop it" became "user declined the prompt, drop it". The commit title makes clear that was not the intent.

## What changed

The associated-jobs test is the outer condition again, matching the pre-2018 shape and matching how `Copy-DbaAgentJobCategory` writes the same `-Force` pattern today:

- attached to a job → report `Skipped` / `Schedule has associated jobs` and `continue` (the "cannot be overwritten, even with Force" case the help describes)
- otherwise → `ShouldProcess("Dropping schedule ... and recreating")`, drop, and fall through to the existing create block

Reporting stays *inside* `ShouldProcess` so `-WhatIf` still suppresses object output, which is what 9f67479581 and #8846 were after. The failed-drop status now also sets `Notes`, which it did not before.

- The non-`-Force` arm.
- The `-WhatIf` object-output suppression.
- The skip behaviour for schedules a job uses — those still cannot be overwritten, as `.PARAMETER Force` documents.
- No other command. I swept every `Copy-Dba*` and `Start-DbaMigration` for an `else` hanging off a `ShouldProcess` at the same nesting level. The three other hits are false positives — their `else` belongs to `if ($force -eq $false)` or `if ($UseLastBackup)`. `Copy-DbaAgentSchedule` was the only one.

## Tests

Two new integration contexts in `tests/Copy-DbaAgentSchedule.Tests.ps1`. Both create the same schedule name on source and destination with **different start times**, so a real drop and recreate is distinguishable from a silent no-op:

- destination schedule with no jobs, `-Force` → status `Successful`, destination start time becomes the source's, exactly one schedule left
- destination schedule attached to a job, `-Force` → status `Skipped` with `Schedule has associated jobs`, destination start time unchanged

Run against the lab (`InstanceCopy1` = SQL03\SQL2022, `InstanceCopy2` = SQL04\SQL2025): **11/11 passed**, no warnings, lab left clean.

Reverting only `public/Copy-DbaAgentSchedule.ps1` and re-running reproduces the report exactly:

```
FAILED  Reports the schedule as copied
        Expected 'Successful', but got $null.
FAILED  Replaces the destination schedule with the source definition
        Expected '01:00:00', but got 05:00:00.
```

---

### DIFF
--- a/public/Copy-DbaAgentSchedule.ps1
+++ b/public/Copy-DbaAgentSchedule.ps1
@@ -165,25 +165,28 @@ function Copy-DbaAgentSchedule {
                         }
                         continue
                     } else {
-                        if ($Pscmdlet.ShouldProcess($destinstance, "Schedule [$scheduleName] has associated jobs. Skipping.")) {
-                            if ($destServer.JobServer.Jobs.JobSchedules.Name -contains $scheduleName) {
+                        # A schedule that is attached to a job at the destination cannot be dropped, so it cannot be overwritten even with -Force.
+                        # This has to be tested before ShouldProcess, otherwise the drop below would be unreachable in a normal run.
+                        if ($destServer.JobServer.Jobs.JobSchedules.Name -contains $scheduleName) {
+                            if ($Pscmdlet.ShouldProcess($destinstance, "Schedule [$scheduleName] has associated jobs. Skipping.")) {
                                 $copySharedScheduleStatus.Status = "Skipped"
                                 $copySharedScheduleStatus.Notes = "Schedule has associated jobs"
                                 $copySharedScheduleStatus | Select-DefaultView -Property DateTime, SourceServer, DestinationServer, Name, Type, Status, Notes -TypeName MigrationObject
                                 Write-Message -Level Verbose -Message "Schedule [$scheduleName] has associated jobs. Skipping."
                             }
                             continue
-                        } else {
-                            if ($Pscmdlet.ShouldProcess($destinstance, "Dropping schedule $scheduleName and recreating")) {
-                                try {
-                                    Write-Message -Level Verbose -Message "Dropping schedule $scheduleName"
-                                    $destServer.JobServer.SharedSchedules[$scheduleName].Drop()
-                                } catch {
-                                    $copySharedScheduleStatus.Status = "Failed"
-                                    $copySharedScheduleStatus | Select-DefaultView -Property DateTime, SourceServer, DestinationServer, Name, Type, Status, Notes -TypeName MigrationObject
-                                    Write-Message -Level Verbose -Message "Issue dropping schedule $scheduleName on $destinstance | $PSItem"
-                                    continue
-                                }
+                        }
+
+                        if ($Pscmdlet.ShouldProcess($destinstance, "Dropping schedule $scheduleName and recreating")) {
+                            try {
+                                Write-Message -Level Verbose -Message "Dropping schedule $scheduleName"
+                                $destServer.JobServer.SharedSchedules[$scheduleName].Drop()
+                            } catch {
+                                $copySharedScheduleStatus.Status = "Failed"
+                                $copySharedScheduleStatus.Notes = "Issue dropping schedule on destination"
+                                $copySharedScheduleStatus | Select-DefaultView -Property DateTime, SourceServer, DestinationServer, Name, Type, Status, Notes -TypeName MigrationObject
+                                Write-Message -Level Verbose -Message "Issue dropping schedule $scheduleName on $destinstance | $PSItem"
+                                continue
                             }
                         }
                     }
--- a/tests/Copy-DbaAgentSchedule.Tests.ps1
+++ b/tests/Copy-DbaAgentSchedule.Tests.ps1
@@ -74,4 +74,131 @@ Describe $CommandName -Tag IntegrationTests {
             $schedule.ActiveStartTimeOfDay | Should -Be "01:00:00"
         }
     }
+
+    Context "When the destination schedule has no associated jobs and -Force is used" {
+        BeforeAll {
+            $PSDefaultParameterValues["*-Dba*:EnableException"] = $true
+
+            # The same schedule name on both instances, but with different start times, so we can
+            # tell a real drop and recreate from a silent no-op. The destination copy has no jobs.
+            $splatSourceForceSchedule = @{
+                SqlInstance       = $TestConfig.InstanceCopy1
+                Schedule          = "dbatoolsci_ForceSchedule"
+                FrequencyType     = "Daily"
+                FrequencyInterval = "Everyday"
+                StartTime         = "010000"
+                Force             = $true
+            }
+            $null = New-DbaAgentSchedule @splatSourceForceSchedule
+
+            $splatDestForceSchedule = @{
+                SqlInstance       = $TestConfig.InstanceCopy2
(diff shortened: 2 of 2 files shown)


potatoqualitee:
Thank you 👌🏼

Issue #10583, opened 2026-08-16
SMO enumeration fails on SQL Server 2005 even though the connection and raw T-SQL succeed


howarthcd:
## Summary

On a real SQL Server 2005 instance (9.00.5000.00, Express Edition with Advanced Services), the TDS connection succeeds and raw T-SQL through `Invoke-DbaQuery` returns results, but every command that walks SMO's object model fails. `$server.Databases` and `$server.Logins` throw `'CONNECTIONPROPERTY' is not a recognized built-in function name`, and `Get-DbaDatabase` throws `Invalid column name 'is_cdc_enabled'`. Both `CONNECTIONPROPERTY` and `is_cdc_enabled` arrived in SQL Server 2008 - SMO issues queries that depend on them regardless of which version it is talking to, so no command that enumerates a database or login collection can reach 2005. `VersionMajor`, `VersionString` and `Edition` all come back empty on the connection too, which matters for the suggested fix below.

This surfaced while working on #10581 (open, not yet merged into `development`), which adds a `-DataPages` switch to `Invoke-DbaDbDecryptObject`. The repro below does not depend on that PR - `Get-DbaDatabase`, `Get-DbaLogin` and plain `Connect-DbaInstance` already reproduce it on current `development`.

This is a different finding from #9821, not a duplicate. That issue was closed as an unsupported combination of PowerShell 7, TLS 1.0 and `Microsoft.Data.SqlClient`, with the connection itself failing during login. Here the connection succeeds (also over TLS 1.0, from PowerShell 7) and `Invoke-DbaQuery` returns results normally - the failure is one layer up, inside SMO's own queries, and would happen the same way for anybody reaching a 2005 instance regardless of client stack.

## Steps to Reproduce

```powershell
# Any command touching SMO's Databases or Logins collection reproduces this against a SQL Server 2005 instance.
Connect-DbaInstance -SqlInstance $instance -SqlCredential $cred | Select-Object VersionMajor, VersionString, Edition
Get-DbaDatabase -SqlInstance $instance -SqlCredential $cred -EnableException
Get-DbaLogin -SqlInstance $instance -SqlCredential $cred -EnableException

# Raw T-SQL through dbatools works fine on the same instance, for contrast.
Invoke-DbaQuery -SqlInstance $instance -SqlCredential $cred -Database master -Query "SELECT SERVERPROPERTY('ProductVersion') AS ProductVersion"
```

[...]

PR #10597, opened 2026-08-22
Connect-DbaInstance - Make MinimumVersion decide on a version it can read

andreasjordan:
**Confirmed on the instance it was written for.** @howarthcd ran the check on the real SQL Server 2005 instance (9.00.5000.00, Express Edition with Advanced Services) against `adcf6bd92`: `SERVERPROPERTY('ProductVersion')` answers there, `-MinimumVersion 10` now refuses with the message #10583 asked for, and `-MinimumVersion 9` still connects. Details in [this comment](https://github.com/dataplat/dbatools/pull/10597#issuecomment-5394777249).

That settles the question this PR was drafted around. The first fallback it originally carried - `ConnectionContext.ServerVersion` - turned out to be as empty as `VersionMajor` on that instance, which is why there is now a third source, and why the third is the one doing the work. What remains is @potatoqualitee's call on whether the code change is wanted at all, or whether #10583 should stay documentation-only as the reporter proposed.

## Problem

`Connect-DbaInstance -MinimumVersion` did nothing on the versions it exists to refuse. The guard read:

```powershell
if ($MinimumVersion -and $server.VersionMajor) {
    if ($server.VersionMajor -lt $MinimumVersion) {
```

The comparison only ran when `VersionMajor` was truthy, so an unreadable property skipped the check rather than failing closed. On the reporter's SQL Server 2005 instance `-MinimumVersion 9` and `-MinimumVersion 10` behaved identically: both connected.

## Mechanism

`VersionMajor` is served by SMO's server information enumerator, and that enumerator declares its own floor. In `dbatools.library` 2026.5.3 (SMO 18.100.1.19), `Information.xml`, `Server.xml`, `Database.xml` and `Login.xml` all open with `min_major='10'`, and 224 of the 298 definitions that declare a floor declare 10:

```powershell
$assembly = [Reflection.Assembly]::LoadFrom("$smoPath\Microsoft.SqlServer.SqlEnum.dll")
$stream = $assembly.GetManifestResourceStream("Information.xml")
(New-Object System.IO.StreamReader($stream)).ReadToEnd() -match "min_major='(\d+)'"
```

So below SQL Server 2008 `VersionMajor`, `VersionString`, `Edition` and `ProductLevel` are all empty, and `$server.Databases` / `$server.Logins` throw `'CONNECTIONPROPERTY' is not a recognized built-in function name`. The 2008 boundary is stated by SMO, not deduced from the symptoms.

## What changed

**`public/Connect-DbaInstance.ps1`** - the version is resolved before it is compared, through three sources in order:

1. `$server.VersionMajor`, as before.
2. `ConnectionContext.ServerVersion.Major`, which comes from the connection rather than from an enumerator. Measured empty on the 2005 instance, kept because it is free.
3. `SERVERPROPERTY('ProductVersion')`, answered since SQL Server 2000 and travelling the same plain T-SQL path that already carries the connection test at line 1170, so it can be read wherever the connection itself works. Measured at `9.00.5000.00` on the 2005 instance while every SMO-served property was empty, and it is the step that carries the check there.

The round trip in step 3 only happens when `-MinimumVersion` was passed *and* both properties came back empty, which is never on an instance SMO can describe. When no source knows the version the connection is allowed through exactly as before, so nothing that connects today starts being refused.

Ruled out along the way: the raw `SqlConnection.ServerVersion` reads only while the connection is open, and SMO closes the pooled connection right after the connection test - `[15.00.4430]` open, `[]` closed on SQL Server 2019 - so it is not available where the guard runs.

**`.github/prompts/sql-version-support.md`** - a new section, *The Floor SMO Enforces*, with the check above so the numbers can be re-verified against whatever library version is pinned, and a record of which of the three version sources a pre-2008 instance can actually answer. `MinimumVersion 9` is now marked as correct only for commands that stay in T-SQL; anything walking SMO collections needs `MinimumVersion 10`. This is the change the issue actually asked for.

Fourteen sites across ten commands (`Backup-DbaDatabase`, `Copy-DbaDatabase`, `Find-DbaOrphanedFile`, `Get-DbaCpuUsage`, `Get-DbaDatabase`, `Get-DbaDbccStatistic`, `Get-DbaDbState`, `New-DbaDatabase`, `Restore-DbaDatabase`, `Connect-DbaInstance`) branch on `VersionMajor -eq 8` or `-eq 9`, and every one of them misroutes for the same reason: both comparisons are false against an empty `VersionMajor`, so the legacy path is skipped and the modern one runs.

That is how `Get-DbaDatabase` produced `Invalid column name 'is_cdc_enabled'` in the report - the query is ours, at `public/Get-DbaDatabase.ps1:348`, and the `-eq 9` branch immediately above it that avoids the column simply did not fire. Worth recording, but not worth fixing: all ten commands walk SMO collections and cannot reach a pre-2008 instance regardless of which query they choose. Routing them correctly would not make them work.

Comparisons written as `-lt` fail safe by accident, since `$null -lt 9` is `$true`.

## Testing

`tests/Connect-DbaInstance.Tests.ps1` gains one Context with five assertions: the connection is returned at the instance's own version, refused below it, warns instead of throwing under `-DisableException`, and both fallback sources - `ConnectionContext.ServerVersion.Major` and the T-SQL `ProductVersion` - agree with `VersionMajor`.

Those last two are the point of the test file rather than an afterthought. No instance in the CI matrix is old enough to exercise a fallback, so what can be pinned is each fallback's source: if either ever stops agreeing with SMO, the chain silently stops refusing the versions it exists to refuse, and this test fails instead.

Run against the lab (`SQL03\SQL2025`): 38 tests, 37 passed, 1 pre-existing skip, no warnings, nothing left behind.

The chain itself cannot be exercised by any instance available to CI or to the lab, so it was covered twice by hand instead:

- **Simulated**, by shadowing both SMO-served properties on a real SQL Server 2019 connection while leaving T-SQL working. `-MinimumVersion 16` then refuses with `SQL Server version 16 required - [SQL03\SQL2019] not supported.` and `-MinimumVersion 10` still connects. Before the third source both of those connected.
- **Real**, by @howarthcd on the SQL Server 2005 instance. Both SMO-served properties empty, `ExecuteScalar` returning `9.00.5000.00`, `-MinimumVersion 10` refused, `-MinimumVersion 9` connected without a warning, and the debug output showing the chain walking all three steps with only the third landing.

Reported and diagnosed by @howarthcd in #10583, including the observation that `VersionMajor` is what `-MinimumVersion` cannot read, the measurement that ruled out the first fallback, and the confirmation of the one that replaced it.


howarthcd:
@andreasjordan Ran the experiment asked for in #10583, against the real SQL Server 2005 instance (9.00.5000.00, Express Edition with Advanced Services), on this PR's branch.

## Result

`ConnectionContext.ServerVersion.Major` is unreadable on this instance, same as `VersionMajor`. The fallback this PR adds has nothing to fall back to here. The code change is inert on the instance it was written for.

## Evidence

```
=== Connect-DbaInstance, PR branch ===
VersionMajor                          : []
VersionString                         : []
Edition                                : []
ConnectionContext.ServerVersion       : []
ConnectionContext.ServerVersion.Major : []

=== Connect-DbaInstance -MinimumVersion 10 ===
Connected anyway. Warning raised: []

=== server.Databases ===
threw: The following exception occurred while trying to enumerate the collection:
"'CONNECTIONPROPERTY' is not a recognized built-in function name.".

=== Invoke-DbaQuery (raw T-SQL, for contrast) ===
ProductVersion : 9.00.5000.00
Edition        : Express Edition with Advanced Services
```

`-MinimumVersion 10` connects without warning, exactly as it did before this PR. Raw T-SQL and the `CONNECTIONPROPERTY` failure both reproduce as in #10583 - nothing about the instance changed between the two reports.

## What this means

The mechanism section of this PR is correct about SMO's own floor, but that floor is exactly what makes the fallback unreachable: an instance old enough to leave `VersionMajor` empty is old enough to leave `ConnectionContext.ServerVersion` empty too. The guard falls through to "let the connection through," the same outcome as before the change.

Harmless, not a fix. `.github/prompts/sql-version-support.md` is the part of this PR that reflects what's actually true on hardware like this.

*This text was created by Claude and reviewed by Chris Howarth.*

andreasjordan:
@howarthcd Thank you - that is exactly the experiment that needed running, and it settles the fallback as written: `ConnectionContext.ServerVersion` is no more readable than `VersionMajor` on that instance, so the guard fell through and nothing changed. You are right that the code in the branch was inert.

But your first report already contains a source that *does* answer on that box:

```
=== Invoke-DbaQuery (raw T-SQL, for contrast) ===
ProductVersion : 9.00.5000.00
```

`SERVERPROPERTY('ProductVersion')` has been answered since SQL Server 2000 and is plain T-SQL, so it does not go near the enumerators that impose the 2008 floor. It also travels the same path that already carries `Connect-DbaInstance`'s own connection test - `ExecuteWithResults("SELECT 'dbatools is opening a new connection'")` at [Connect-DbaInstance.ps1:1170](https://github.com/dataplat/dbatools/blob/development/public/Connect-DbaInstance.ps1#L1170) - so a version read this way cannot be less available than the connection itself.

So I have pushed a third step to the chain rather than giving up on the code change. `VersionMajor` is tried first, then `ConnectionContext.ServerVersion`, then the T-SQL read; the extra round trip only happens when `-MinimumVersion` was passed *and* both properties came back empty, which is never on an instance SMO can describe.

## One path I checked and ruled out

The raw `SqlConnection.ServerVersion` that reported `09.00.5000` in your first test only reads while the connection is open, and SMO closes the pooled connection immediately after the connection test. Confirmed here on SQL Server 2019: `[15.00.4430]` while open, `[]` once closed. So it is not usable at the point where the guard runs, and T-SQL is the honest route.

## What I can verify here, and what I cannot

No instance in the lab is old enough to blank `VersionMajor`, so I simulated one - shadowing both SMO-served properties on a real 2019 connection while leaving T-SQL working:

```
shadowed VersionMajor   : []
shadowed ServerVersion  : []
--- MinimumVersion 10 (should connect, 15 >= 10) ---
returned: [True] warning: []
--- MinimumVersion 16 (should refuse, 15 < 16) ---
SQL Server version 16 required - [SQL03\SQL2019] not supported.
```

Before this commit both of those connected. The test file gains an assertion that the T-SQL major agrees with `VersionMajor`, alongside the existing one for `ConnectionContext.ServerVersion` - the same reasoning as before: the versions the guard exists to refuse are not in the matrix, but the fallback's *source* is readable everywhere, so that is what gets pinned. 38 tests, 37 passed, 1 pre-existing skip, no warnings, nothing left behind.

What I cannot verify is the only thing that matters: that it returns `9` on your instance. If you have the appetite for one more run, this is the whole of it:

```powershell
$server = Connect-DbaInstance -SqlInstance $instance -SqlCredential $cred
$server.ConnectionContext.ExecuteScalar("SELECT SERVERPROPERTY('ProductVersion')")
Connect-DbaInstance -SqlInstance $instance -SqlCredential $cred -MinimumVersion 10 -WarningVariable w -WarningAction SilentlyContinue
$w
```

If the first line returns `9.00.5000.00` and the second refuses with `SQL Server version 10 required - ... not supported.`, this is a real fix and your report gets the clear "your SQL Server is too old" message it asked for. If it somehow comes back empty as well, then documentation-only is the answer, we drop the code change, and your reading of it stands as written.

The doc section now records all three sources and which of them your instance can actually answer, so the finding survives either outcome. And the out-of-scope call on the fourteen `-eq 8` / `-eq 9` sites is unchanged: a readable version would let them route correctly, but SMO collections still cannot reach 2005, so it would not make them work.

howarthcd:
@andreasjordan Ran the check from your comment against the same SQL Server 2005 instance (9.00.5000.00, Express Edition with Advanced Services), on `adcf6bd92`.

## Result

`SERVERPROPERTY('ProductVersion')` answers: `9.00.5000.00`, major `9`. `-MinimumVersion 10` now refuses with the message #10583 asked for, and `-MinimumVersion 9` still connects. Your prediction held - this is a real fix, not an inert one.

## Evidence

```
=== every version source ===
VersionMajor                          : []
VersionString                         : []
Edition                               : []
ConnectionContext.ServerVersion       : []
ConnectionContext.ServerVersion.Major : []

=== the T-SQL read the third step uses ===
ExecuteScalar ProductVersion          : [9.00.5000.00]
derived major                         : [9]

=== Connect-DbaInstance -MinimumVersion 10 ===
refused with             : [SQL Server version 10 required - [<instance>] not supported.]
returned a server object : [False]

=== Connect-DbaInstance -MinimumVersion 9 ===
returned a server object : [True]
warning                  : []
```

The debug messages show the chain walking all three steps, and only the third one landing:

```
VersionMajor is not available, using ServerVersion.Major [] from the connection
ServerVersion is not available, using ProductVersion [9.00.5000.00] from T-SQL
```

## What this means

Both SMO-served properties are as empty as in my last run - nothing about the instance changed between the two reports, so the first two steps stay unreachable on hardware like this and the third is what carries the check. `-MinimumVersion 10` connected silently before this commit; it now stops. `-MinimumVersion 9` connecting without a warning is the other half of that: the boundary is at the right place, and nothing that works today started being refused.

The T-SQL route is the one that works on the instance the fix was written for. Documentation-only is no longer the answer.

*This text was created by Claude and reviewed by Chris Howarth.*

andreasjordan:
@howarthcd That settles it - thank you for running both rounds. The first one saved a fallback that could not work from going in, and the second one confirmed the replacement on the only hardware that can answer the question. The debug trace showing all three steps walked with only the third landing is exactly what I would have wanted to see and could not produce here.

PR description updated to record the result, and the PR is out of draft and ready for review. The remaining call is @potatoqualitee's: whether the code change is wanted at all, or whether #10583 stays documentation-only as you originally proposed. Your reading of the boundary holds either way.

niphlod:
I *think* this needs to stay in because of Connect-DbaInstance's "contract" ... 
but at the same time I'm thinking that functions built on top of this (read "for this version run this, for that version run that") should explicitely call something *other* than Connect-DbaInstance's `-Min/-Max` that can be the only "refactored" source of truth of ProductVersion --> feature available. 
We have too many places where a major/minor have been declared as "enums" and every time a new version comes up there are too many places to look for.
Another "worrysome thing" is that if SMO's new versions are going to increasingly "drop support" for some older versions, and dbatools wanna continue to support everything, it will be reaally hard to track where to intervene to support this or that, without a central place where to "hook" things into.

andreasjordan:
Agreed on both halves: the guard belongs to `Connect-DbaInstance`'s contract, and feature gating does not belong in the same place.

The scale is worse than it looks. There are 166 `VersionMajor -<op> <n>` comparisons across 78 files, and 149 `-MinimumVersion <n>` call sites, 88 of them passing 9. With the documentation change in this PR, most of those 88 are wrong by one - anything walking SMO's collections needs 10.

The drift is not theoretical either. `Copy-DbaSsisCatalog` connects to the source with `-MinimumVersion 11` and to the destination with `-MinimumVersion 1`, while the help for both parameters says SQL Server 2012 or higher. A dropped digit that nothing catches.

On tracking SMO's floor: that part is machine readable per library version. The check in this PR's doc section reads `min_major` straight out of the shipped `Microsoft.SqlServer.SqlEnum.dll`, so a test against whatever we pin in `.github/dbatools-library-version.json` could tell us when SMO drops a version, instead of us finding out from a bug report.

I would rather not grow this PR into that refactor. Opened #10600 for the central feature-to-floor table with these numbers in it - say the word if you want it here instead.

potatoqualitee:
awesome work everyone! 🙏🏼  i will release this today once i go thru the PRs

Issue #10600, opened 2026-08-24
Version gating has no single source of truth - 166 comparisons across 78 files


andreasjordan:
## Summary

Version gating is spread across the module with no single source of truth. Each site carries a bare major version number, so adding a SQL Server release, or reacting to SMO dropping an old one, means finding every one of them by hand. Raised by @niphlod in https://github.com/dataplat/dbatools/pull/10597#issuecomment-5396767235.

## Scale

Measured on `development`:

- 166 `VersionMajor -<op> <n>` comparisons across 78 files.
- 149 `Connect-DbaInstance -MinimumVersion <n>` call sites: 88 pass 9, 36 pass 10, 20 pass 11, 4 pass 13, one passes 8, one passes 1.
- 30 further `$server.Version` comparisons, plus `DatabaseEngineType` (27) and `DatabaseEngineEdition` (16) checks that gate on the same kind of thing.

Nothing ties a number to the feature it stands for, so the intent has to be re-derived from context at every site.

## Two things this already costs us

**The numbers drift silently.** `public/Copy-DbaSsisCatalog.ps1:287` connects to the source with `-MinimumVersion 11`; line 316 connects to the destination with `-MinimumVersion 1`. The help for both parameters says SQL Server 2012 or higher. A dropped digit, invisible to review and to the test suite.

**The guidance moved and the call sites did not.** #10597 establishes that SMO's own enumerators require SQL Server 2008, so any command walking `Databases`, `Logins` or similar needs `-MinimumVersion 10`, not 9. That makes most of the 88 sites passing 9 nominally wrong. Correcting them one by one is exactly the work a central table would make unnecessary next time.

## Shape of a fix

A table mapping feature to minimum version, and a private helper that commands ask by name rather than by number, so a new release or a changed floor is one edit:

```powershell
if (Test-DbaFeatureSupport -SqlInstance $server -Feature QueryStore) { ... }
```

`Connect-DbaInstance -MinimumVersion` stays as it is. It is the connection-level contract - "do not hand me a server older than this" - and #10597 makes it enforceable on versions where SMO cannot report one. What moves out is per-feature branching that currently reuses it.

## SMO's floor is machine readable

The concern that SMO will keep dropping old versions faster than we notice can be answered mechanically. Each enumerator definition in `Microsoft.SqlServer.SqlEnum.dll` declares its own floor:

```powershell
[...]

PR #10601, opened 2026-08-24
Copy-DbaSsisCatalog - Require SQL Server 2012 on the destination, as the help already promises

andreasjordan:
Found while counting version gates for #10600, and small enough to fix rather than leave in the tally.

## What was wrong

`public/Copy-DbaSsisCatalog.ps1:316` connected to the destination with `-MinimumVersion 1`:

```powershell
$destinationConnection = Connect-DbaInstance -SqlInstance $destinstance -SqlCredential $DestinationSqlCredential -MinimumVersion 1
```

Every instance satisfies 1, so the guard did nothing. Line 287 connects to the source with `-MinimumVersion 11`, and the help for both `-Source` and `-Destination` says "Requires sysadmin privileges and SQL Server 2012 or higher". A dropped digit, invisible to review because the line still reads like a version check.

The SSISDB catalog is a 2012 feature, so a pre-2012 destination failed later and less clearly - `New-Object ...IntegrationServices` against a server with no catalog support, rather than "SQL Server version 11 required".

## Two more in the same loop

Both within three lines of the first, so they belong in this PR rather than a follow-up:

- The catch for the destination integration services passed `-Target $destinationCon`, a variable that is never assigned anywhere in the command. The connection is `$destinationConnection`, so the error record carried a null target.
- Neither that catch nor the Integration Services check used `-Continue`, although both sit inside `foreach ($destinstance in $Destination)` and the connection catch three lines above does use it. Without `-Continue`, `Stop-Function` warned and execution fell straight through to the next statement, which then worked on a null `$destinationSSIS` - so one bad destination produced a confusing second error instead of moving on to the next instance.

## Testing

This command cannot be integration tested here, and I would rather say so than imply otherwise:

- It refuses to run on PowerShell Core (`begin` block, line 129), and the lab and the container workflow both run PowerShell 7.
- No CI scenario provisions an SSISDB catalog, so there is no boundary to test against.
- Exercising the corrected guard needs a destination older than SQL Server 2012, and the oldest instance in the lab is 2019.

So the existing unit test is unchanged - the parameter surface did not move - and the change was verified by reading the two call sites against each other and against the help. The file parses clean and passes the repository style check.

If a maintainer with an SSIS lab wants to confirm the destination guard before this merges, the check is `Copy-DbaSsisCatalog -Source <2012+> -Destination <pre-2012>`, which should now refuse with `SQL Server version 11 required - ... not supported.`

### DIFF
--- a/public/Copy-DbaSsisCatalog.ps1
+++ b/public/Copy-DbaSsisCatalog.ps1
@@ -313,7 +313,7 @@ function Copy-DbaSsisCatalog {
         }
         foreach ($destinstance in $Destination) {
             try {
-                $destinationConnection = Connect-DbaInstance -SqlInstance $destinstance -SqlCredential $DestinationSqlCredential -MinimumVersion 1
+                $destinationConnection = Connect-DbaInstance -SqlInstance $destinstance -SqlCredential $DestinationSqlCredential -MinimumVersion 11
             } catch {
                 Stop-Function -Message "Error occurred while establishing connection to $destinstance" -Category ConnectionError -ErrorRecord $_ -Target $destinstance -Continue
             }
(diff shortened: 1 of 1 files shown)


potatoqualitee:
👌🏼  thank you

PR #10605, opened 2026-08-26
New-DbaFirewallRule - Adapt the tests to a default instance on the default port

andreasjordan:
## Summary

The integration tests assumed InstanceSingle is a named instance. Run against a default instance on the default port they fail ten times, although the command behaves exactly as documented:

- The Engine rule of a default instance is named `SQL Server default instance`, not `SQL Server instance MSSQLSERVER`.
- A Browser rule is only created for a named instance or a static port other than 1433, because a default instance on 1433 is reachable without the Browser. So a default instance gets exactly one rule, and every count, index and Browser assertion missed.

## Changes (tests only, both contexts the same way)

- The expected Engine rule name and the expected minimum rule count are computed from the instance in `BeforeAll`.
- A `$browserExpected` value is computed at discovery time next to the existing `$isUsingDynamicPort` probe, mirroring the condition of the command, and the Browser tests carry `-Skip:(-not $browserExpected)`.
- A negative counterpart test with `-Skip:$browserExpected` asserts that a default instance on the default port gets no Browser rule, so this shape is now covered instead of skipped over.
- The counts use `@()` because a single result from `Select-DefaultView` does not answer `.Count`.

The named-instance assertions are unchanged in strictness and keep running on CI and the named-instance lab configurations.

Verified against both shapes: a clustered default instance on port 1433 (11 passed, Browser tests skipped, negative tests run) and a named instance (full Browser coverage), both warning-free.

### DIFF
--- a/tests/New-DbaFirewallRule.Tests.ps1
+++ b/tests/New-DbaFirewallRule.Tests.ps1
@@ -142,7 +142,13 @@ Describe $CommandName -Tag UnitTests {
 Describe $CommandName -Tag IntegrationTests {
     # The context "RuleType Port (traditional port-based rules)" does not work with dynamic ports.
     # So we test at discovery time if dynamic ports are used and skip the tests if so.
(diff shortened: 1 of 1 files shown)


potatoqualitee (approved):

PR #10606, opened 2026-08-26
Rename-DbaDatabase - Build expected file paths without validating drives of the test client

andreasjordan:
## Summary

The physical-name template test builds its expected paths with `Join-Path` from the file paths of the instance. Windows PowerShell validates the drive qualifier of `Join-Path` against the drives of the machine running the test, so against a remote instance whose files live on a drive the test client does not have - such as a failover cluster instance keeping its databases on a clustered disk - the test fails with "Cannot find drive. A drive with the name 'S' does not exist" although the command works.

## Changes (tests only)

The four expected paths are now built with `[System.IO.Path]::Combine`, which composes strings without touching the provider - and is also what Rename-DbaDatabase itself builds the new path with, so the expectation mirrors the command exactly.

Verified against a clustered instance with databases on a clustered drive (the previously failing shape) and against a regular instance, both warning-free with all tests passing.


potatoqualitee (approved):

Issue #10607, opened 2026-08-27
Found some bugs in dbatools — how do you want them?


greenmtnsun:
### Verified issue does not already exist?

I have searched and found no existing issue

### What error did you receive?

I'm greenmtnsun and/or kramsey from the SQL Server community Slack — been lurking and asking
questions there for a while.  I think we may have crossed paths at a SQL Saturday a
while back as well if you were in Portland. Longtime dbatools user, DBA by day, PowerShell tinkerer the rest of the
time.

Here's my situation. I've been building a Pester test generator on the side, and I use
the dbatools public commands as its main test corpus — about 700 real functions is the
best torture test I could ask for. While I was validating a new static-analysis pass in
it, the thing started finding actual defects in dbatools rather than in my tool.

Two I can demonstrate rather than wax on about it, both run against SQL Server 2025 with
dbatools 2.8.3:

- `Get-DbaDiskSpace` reports the same "1.82 TB" whether I pass `-Unit MB`, `-Unit KB`,
  or nothing at all. The parameter is declared, documented, and never read.
- `Get-DbaDbTable` returns an identical 172 tables with and without
  `-IncludeSystemDBs`, and master and msdb come back either way. So the switch does
  nothing, and system databases show up when they shouldn't.

There are more — a couple of dozen dead parameters in the same shape, two Remove
commands that ignore their `-AvailabilityGroup` filter (so scoping a removal to one
availability group removes from all of them), and a `-Force` path in
`Copy-DbaAgentSchedule` that can't be reached. I filed one through the issue form
already (10594) and was going to stage the rest out over a few days. Then it
occurred to me that a pile of same-shaped issues from one account is going to look like
a bot went through your repo, which is not the impression I want to make.

So I figured I'd just ask. I can file them one at a time, on whatever schedule suits
your triage. Or I send you the findings document and the detector and you do whatever
you like with it.

That second option might be the more interesting one, because we're clearly working the
same way. I saw the `.claude` directory in the repo — the style hooks, ralph, the whole
setup. Most of my analysis came out of Claude too. So rather than me hand-writing two
dozen issues and you hand-reading two dozen issues, I could send the structured findings
and let your tooling chew on them.

Better still: the dead-parameter check is the same shape as your `validate-style.ps1`
PreToolUse hook — static analysis over the .ps1 content, no execution. It's about forty
lines of AST work, and it knows about your `Test-Bound` idiom, which is why these
[...]

PR #10608, opened 2026-08-27
Remove-DbaAgDatabase - Honor the AvailabilityGroup filter

andreasjordan:
## Problem

`Remove-DbaAgDatabase` declares and documents `-AvailabilityGroup` ("Limits the operation to databases within specific availability groups") but never uses it. The internal `Get-DbaAgDatabase` call only received `-Database`, so a removal scoped to one availability group removed the database from whatever availability group actually contained it.

## What changed

The `-AvailabilityGroup` parameter is now passed through to `Get-DbaAgDatabase`, which already supported it. The call was converted to a splat per the style guide.

- The pipeline path (`-InputObject`) is untouched: piped objects are already the caller's explicit selection.
- The error text for the `-SqlInstance` path used to demand "one or more Availability Groups" although the check only requires `-Database`; a follow-up commit on this branch corrects the wording to match what is actually required.

## Tests

Added a regression test that creates a second availability group, scopes a removal to it, and asserts the database in the first availability group survives and nothing is returned. Verified against the lab HADR instance (SQL Server 2025): the new test fails on the old code with exactly the reported behavior (`Status=Removed` for the out-of-scope database) and the full file passes with the fix (4/4).

### DIFF
--- a/public/Remove-DbaAgDatabase.ps1
+++ b/public/Remove-DbaAgDatabase.ps1
@@ -98,7 +98,7 @@ function Remove-DbaAgDatabase {
 
         if ((Test-Bound -ParameterName SqlInstance)) {
             if ((Test-Bound -Not -ParameterName Database)) {
-                Stop-Function -Message "You must specify one or more databases and one or more Availability Groups when using the SqlInstance parameter."
+                Stop-Function -Message "You must specify one or more databases when using the SqlInstance parameter."
                 return
             }
         }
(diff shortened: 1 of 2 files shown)


potatoqualitee (approved):

potatoqualitee:
Thank you 🙏 I'm on mobile which makes commenting a bit harder so fewer will have it but please know it is all appreciated 🫶🏼

PR #10609, opened 2026-08-27
Remove-DbaAgListener - Honor the AvailabilityGroup filter

andreasjordan:
## Problem

`Remove-DbaAgListener` declares and documents `-AvailabilityGroup` ("Filters listener removal to only those within the specified availability groups") but never uses it. The internal `Get-DbaAgListener` call only received `-Listener`, so a removal scoped to one availability group removed the named listener regardless of which availability group it belonged to.

## What changed

The `-AvailabilityGroup` parameter is now passed through to `Get-DbaAgListener`, which already supported it. The call was converted to a splat per the style guide.

A follow-up commit also corrects the `-SqlInstance` validation message, which demanded "one or more Availability Groups" although the check only requires `-Listener`.

The pipeline path (`-InputObject`) is untouched: piped listener objects are already the caller's explicit selection.

## Tests

Added a regression test that creates a second availability group, scopes the removal of the first group's listener to that second group, and asserts nothing is returned and the listener survives. Verified against the lab HADR instance (SQL Server 2025): the new test fails on the old code with exactly the reported behavior (`Status=Removed` for the out-of-scope listener) and the full file passes with the fix (3/3).

Same defect shape as #10608. Found by @greenmtnsun via static analysis, reported in #10607.

### DIFF
--- a/public/Remove-DbaAgListener.ps1
+++ b/public/Remove-DbaAgListener.ps1
@@ -97,7 +97,13 @@ function Remove-DbaAgListener {
         }
 
         if ($SqlInstance) {
-            $InputObject += Get-DbaAgListener -SqlInstance $SqlInstance -SqlCredential $SqlCredential -Listener $Listener
+            $splatGetAgListener = @{
+                SqlInstance       = $SqlInstance
+                SqlCredential     = $SqlCredential
+                Listener          = $Listener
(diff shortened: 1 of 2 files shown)


potatoqualitee (approved):

PR #10610, opened 2026-08-27
Get-DbaDbTable - Make IncludeSystemDBs actually exclude system databases by default

andreasjordan:
## Problem

`Get-DbaDbTable` declares `-IncludeSystemDBs` and its help promises "By default system databases are excluded since they rarely contain user tables of interest" — but the switch was never referenced in the body. A default scan always returned the tables of master, model, msdb and tempdb (151 system tables on a bare lab instance), and the switch changed nothing.

## What changed

When scanning the instance (no `-Database` given), the internal `Get-DbaDatabase` call now passes `-ExcludeSystem` unless `-IncludeSystemDBs` is set. The call was converted to a splat per the style guide.

The semantics follow `Get-DbaDbEncryption`: a database requested explicitly with `-Database` is always honored, whether it is a system database or not. This keeps `Get-DbaDbTable -Database msdb -Table backupset` working without the switch, and keeps the in-repo caller in `Install-DbaMaintenanceSolution` (which passes `-Database` together with `-IncludeSystemDBs`) working unchanged. The help now states this explicitly.

- The pipeline path (`-InputObject`) is untouched: piped database objects are the caller's explicit selection.
- This is a behavior change for callers who relied on the buggy default returning system tables in a full scan — but it is the behavior the parameter and its documentation have promised since the parameter was introduced.

## Tests

Added a "System database handling" context: default scan contains no msdb tables, `-IncludeSystemDBs` brings them back, and an explicit `-Database msdb` works without the switch. Verified in the lab against SQL Server 2019: the default-scan test fails on the old code (msdb found) and the full file passes with the fix (9/9).

### DIFF
--- a/public/Get-DbaDbTable.ps1
+++ b/public/Get-DbaDbTable.ps1
@@ -27,6 +27,7 @@ function Get-DbaDbTable {
     .PARAMETER IncludeSystemDBs
         Includes system databases (master, model, msdb, tempdb) in the table scan.
         By default system databases are excluded since they rarely contain user tables of interest.
+        Databases requested explicitly with -Database are always processed, whether they are system databases or not.
 
     .PARAMETER Table
         Specifies specific tables to retrieve using one, two, or three-part naming (table, schema.table, or database.schema.table).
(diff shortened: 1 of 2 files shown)


potatoqualitee (approved):

Issue #5023, opened 2019-01-29
get-dbadiskspace -checksql needs some tlc


potatoqualitee:
get-dbadiskspace -checksql doenst work so well with clusters and is a bit rough around the edges

wsmelton:
Personally I'd rather remove that code all together from that command and just make something like `Test-DbaSqlDisk` or say `Test-DbaDisk`. There are to many variations that can say a given drive is part of SQL Server (e.g. backups could be written to it when not default backup path). So we could expand an individual command to really validate whether a given disk is used by the SQL Server instance.

potatoqualitee:
I agree with you, @wsmelton - good call

potatoqualitee:
prerelease cuz this is breaking change

PR #10611, opened 2026-08-27
Get-DbaDiskSpace - Make Unit work and clean up the two other dead parameters

andreasjordan:
## Problem

Three parameters of `Get-DbaDiskSpace` were declared, documented with examples, and never read:

- `-Unit` was deprecated in the 2017 rewrite (when the output became self-formatting `Size` objects), but the parameter, its `GB` default and its help examples all survived, so `-Unit MB` silently returned the same display as no parameter at all.
- `-SqlCredential` lost its only consumer when `-CheckForSql` was removed in 2019 (#5023).
- `-CheckFragmentation` lost its DefragAnalysis implementation in the same 2017 rewrite.

## What changed

- **`-Unit` now works.** The `Size` type has a writable per-object display style, so the command sets it on `Capacity` and `Free` when `-Unit` is bound: `-Unit MB` displays `1,907,348.63 MB` where the default displays `1.82 TB`. The underlying values are unchanged and all `SizeIn*`/`FreeIn*` properties remain available. `PB` is dropped from the ValidateSet because the `SizeStyle` enum does not offer it, and the never-applied `GB` default is dropped so the default display stays dynamic — both exactly as the command has actually behaved for nine years.
- **`-SqlCredential` is removed.** A deprecation warning was not an option here: the test guide and many environments set a wildcard `*:SqlCredential` default via `$PSDefaultParameterValues`, which binds the parameter on every call and would warn for everyone.
- **`-CheckFragmentation` stops with a message** (the `Get-DbaDbBackupHistory -AgCheck` precedent), since silently returning results without the requested fragmentation data misleads the caller.

Help and `.OUTPUTS` updated accordingly.

## Tests

Parameter validation updated; new tests assert `-Unit MB` changes the displayed unit without changing the values, and that `-CheckFragmentation` returns nothing and warns. Verified locally: all three fail against the old code (`-Unit MB` displaying `49.47 GB` is exactly the reported symptom) and the file passes 6/6 with the fix.

### DIFF
--- a/public/Get-DbaDiskSpace.ps1
+++ b/public/Get-DbaDiskSpace.ps1
@@ -4,7 +4,7 @@ function Get-DbaDiskSpace {
         Retrieves disk space and filesystem details from SQL Server host systems for capacity monitoring and performance analysis.
 
     .DESCRIPTION
-        Queries Windows disk volumes on SQL Server systems using WMI to gather critical storage information for database administration. Returns comprehensive disk details including capacity, free space, filesystem type, and optional fragmentation analysis.
+        Queries Windows disk volumes on SQL Server systems using WMI to gather critical storage information for database administration. Returns comprehensive disk details including capacity, free space and filesystem type.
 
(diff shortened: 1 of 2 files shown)


potatoqualitee (approved):

potatoqualitee:
🙏

PR #10615, opened 2026-08-27
Set-DbaTempDbConfig - Make forced file removal survive work tables and stop resetting the log growth

andreasjordan:
## Problem 1: forced reductions die on work table pages

`Set-DbaTempDbConfig -Force` empties and removes the highest-numbered files with `DBCC SHRINKFILE (..., EMPTYFILE)`. EMPTYFILE cannot move work table pages, and cached work tables survive the queries that created them — so on any instance whose cache is warm the reduction fails with:

> Cannot move all contents of file "tempdev4" ... Page 6:2521 could not be moved because it is a work table page.

This looked environment-sensitive for weeks (it failed deterministically on instances with small tempdb files and passed elsewhere) because it depends on where proportional fill placed the cached work tables — the freshly added, emptiest files are exactly where new work tables land, and exactly the files the reduction removes.

**Fix:** the removal travels as one T-SQL batch per file that releases the unused cache entries first (`DBCC FREESYSTEMCACHE (''ALL'')` — measured as the minimal sufficient statement; the plan cache''s active entries and the buffer pool are untouched), then empties and removes, and retries up to five times inside the batch, because a fresh work table can appear in the file at any moment. Measured on the lab: a single up-front release is not enough (a round trip later, the file was blocked again), and even release+empty in one batch failed intermittently — hence the retry. The retry lives in the emitted script, so `-OutputScriptOnly` users get the same robustness, and it avoids `THROW` so the script stays runnable on older versions.

## Problem 2: every call silently reset the log file growth

When `-LogPath` is not bound the command derives it from the running instance, and the decision whether to emit the log file statement read the variable (`if ($LogPath -or $LogFileSize)`) — which after the derivation is always true. So **every** call, even one asking for nothing log-related, rewrote the log file with `filegrowth=512` (the `-LogFileGrowth` default). Verified with a sentinel: set log growth to 64 MB, run a reduction without log parameters, growth is 512 MB afterwards. This contradicts the documented contract ("When omitted, the existing log file size remains unchanged") and is how lab instances ended up with oversized tempdb logs — one growth step is half a gigabyte.

**Fix:** the decision is made from the bound parameters: the log statement is emitted only when at least one of `-LogPath`, `-LogFileSize`, `-LogFileGrowth` or `-DisableGrowth` is specified. `$LogFileSize` is still defaulted from the current size so the output object keeps reporting the real log size.

## Tests

- The reduction test now seeds 40 cached work tables (distinct static cursors) before reducing, making defect 1 deterministic: with the seeding and the old command, the test fails with exactly the production error; with the fix it passes.
- A new assertion runs the reduction without log parameters and asserts the log growth is unchanged (red on the old code, where it flips to 512 MB).
- The `AfterAll` now restores the tempdb log size (the allocation table grows it through normal autogrowth and nothing put it back — the other half of the drift), and its size-restore and file-removal fallbacks release caches the same way the command does.

Verified in the lab: three consecutive clean passes (6/6, no leftovers, no warnings) against SQL05\SQL2025 (the 8 MB-file shape that failed deterministically) and one against the stand-alone SQL 2019 default instance. Lab tempdb left pristine after every run.

### DIFF
--- a/public/Set-DbaTempDbConfig.ps1
+++ b/public/Set-DbaTempDbConfig.ps1
@@ -198,17 +198,20 @@ function Set-DbaTempDbConfig {
             }
 
             #Set DataFileCount if not specified. If specified, check against best practices.
+            # Per-iteration local: assigning the fallback to the parameter variable would carry the
+            # first instance's core count into every later instance of the same call.
             if (-not $DataFileCount) {
-                $DataFileCount = $cores
-                Write-Message -Message "Data file count set to number of cores: $DataFileCount" -Level Verbose
+                $effectiveDataFileCount = $cores
+                Write-Message -Message "Data file count set to number of cores: $effectiveDataFileCount" -Level Verbose
             } else {
-                if ($DataFileCount -gt $cores) {
-                    Write-Message -Message "Data File Count of $DataFileCount exceeds the Logical Core Count of $cores. This is outside of best practices." -Level Warning
+                $effectiveDataFileCount = $DataFileCount
+                if ($effectiveDataFileCount -gt $cores) {
+                    Write-Message -Message "Data File Count of $effectiveDataFileCount exceeds the Logical Core Count of $cores. This is outside of best practices." -Level Warning
                 }
-                Write-Message -Message "Data file count set explicitly: $DataFileCount" -Level Verbose
+                Write-Message -Message "Data file count set explicitly: $effectiveDataFileCount" -Level Verbose
             }
 
-            $DataFilesizeSingle = $([Math]::Floor($DataFileSize / $DataFileCount))
+            $DataFilesizeSingle = $([Math]::Floor($DataFileSize / $effectiveDataFileCount))
             Write-Message -Message "Single data file size (MB): $DataFilesizeSingle." -Level Verbose
 
             if (Test-Bound -ParameterName DataPath) {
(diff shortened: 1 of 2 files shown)


potatoqualitee (changes requested):
The new log-size preservation uses the parameter variable as a per-instance fallback. Inside the `foreach ($instance in $SqlInstance)`, `$LogFileSize = $logfile.SizeMb` persists into later instances. With sql1 at 1024 MB and sql2 at 256 MB plus `-LogFileGrowth 64`, sql2 is unexpectedly changed to 1024 MB; reversed sizes can trigger a shrink failure. Please use a per-iteration effective/current log-size local for the SQL and output, and add a two-instance regression with differing log sizes.

andreasjordan:
Fixed in 01848bac3 - and the regression hunt surfaced two same-class siblings that are fixed in the same commit:

- **The reviewed defect:** the unbound `-LogFileSize` fallback is now a per-iteration local (`\`); with sql1 at one size and sql2 at another, each instance's statement and result object carry its own size.
- **Same defect, other parameter:** the unbound `-DataFileCount` fallback assigned `\` to the parameter variable too, carrying the first instance's core count into every later instance. Same per-iteration treatment (`\`). Invisible on this lab only because the instances share a host.
- **Pipeline pollution:** every executed batch leaked a raw `-1` (the `ExecuteNonQuery` return) into the pipeline alongside the result object - twelve objects came back where two were promised. Now discarded.

**The requested two-instance regression** runs one call over both MULTI instances with differing log sizes (grown deliberately, and measured against the running files - the same source the command's fallback reads, since configured and running drift apart between restarts on a lab instance). It needed a second instance, so the test file moves wholly into the MULTI lane per the one-lane rule (#10628/#10629). Red on the old code with all three symptoms - `Expected 2, but got 12` result objects, and the core one, instance2's log configured to `8` instead of its own `136` - green on the fix, 9/9 via the lab harness, both tempdbs restored, lab left clean.

potatoqualitee (approved):

PR #10616, opened 2026-08-27
Get-DbaPrivilege, Set-DbaPrivilege - Pass the credential to the connectivity test and the service discovery

andreasjordan:
## What this fixes

`Get-DbaPrivilege` and `Set-DbaPrivilege` accept `-Credential` and pass it to every `Invoke-Command2` call, but not to the `Test-PSRemoting` pre-flight check - and `Set-DbaPrivilege` also not to the `Get-DbaService` call that discovers the service accounts. Those calls authenticated with the implicit identity instead.

When the calling script itself runs in a remoting session (Ansible `win_powershell`, a nested `Invoke-Command`, a scheduled task with a network logon), the implicit identity is a network logon token that cannot authenticate to a second machine (the classic double hop). The pre-flight check then fails on the **client** with WSManFault 1312 ("A specified logon session does not exist"), although the supplied credential would have worked: an explicit credential is a fresh authentication and needs no delegation. `Get-DbaPrivilege` aborted; `Set-DbaPrivilege` skipped the computer with only a warning - even with `-EnableException` it looked like a success.

```
Get-DbaPrivilege -ComputerName server-005 -Credential $cred

<f:WSManFault Code="1312" Machine="client-005.ordix.local">
  <f:Message>A specified logon session does not exist. ...</f:Message></f:WSManFault>
```

Until now the only workaround was to wrap the calls in `Invoke-Command -Authentication Credssp` and run them locally on the target.

## The change

- Both commands pass `-Credential` to `Test-PSRemoting` when one is supplied, matching what `Get-DbaMsdtc` already does.
- `Set-DbaPrivilege` passes `-Credential` to its internal `Get-DbaService` call.
- For the one case that remains unavoidable - no credential supplied from a network logon session - the error (`Get-DbaPrivilege`) and the warning (`Set-DbaPrivilege`) now name the double hop and the ways out (pass `-Credential`, or use an authentication that supports delegation such as CredSSP) instead of only relaying the raw WSMan fault.

No behavior changes otherwise: from a normal session, with or without a credential, both commands work exactly as before.

## Verification

Reproduced and verified in a lab (ADMIN01 -> nested WinRM session -> SQL03, Windows PowerShell 5.1):

| Scenario | Before | After |
|---|---|---|
| Nested session, `-Credential` supplied, `Get-DbaPrivilege` | WSManFault 1312 | works (20 rows) |
| Nested session, `-Credential` supplied, `Set-DbaPrivilege` | silent no-op + "Failed to connect" warning | works, warning-free |
| Nested session, no credential | fails (true double hop) | still fails, but the message now explains the double hop and the ways out |
| Normal session, with and without credential | works | works |

The nested-session condition was created with a loopback `Invoke-Command -ComputerName localhost`, which yields the same network logon token as Ansible's `win_powershell`.

The underlying mechanism was confirmed with primitives from inside the nested session: `Test-WSMan -Authentication Default` with the implicit identity fails with 1312, the same call with the explicit credential succeeds, and `Invoke-Command -Credential` reaches the target - so the fix passes the credential to the only calls that were missing it.

## Tests

Two mock-based regression tests pin the credential pass-through (`Should -Invoke ... -ParameterFilter { $Credential -eq $testCredential }`), one per command, in the existing `InModuleScope` regression blocks. They are mock-based because the failing boundary - a nested WinRM session plus a second Windows machine reachable with alternate credentials - cannot be provisioned on any CI runner; the real-boundary verification is the lab run documented above. Both test files pass completely (8/8).

---

### DIFF
--- a/public/Get-DbaPrivilege.ps1
+++ b/public/Get-DbaPrivilege.ps1
@@ -91,18 +91,22 @@ function Convert-SIDToUserName ([string] `$SID ) {
     }
     process {
         foreach ($computer in $ComputerName) {
+            # Invoke-Command2 executes on the local computer under the process identity and ignores
+            # -Credential, so the connectivity test has to authenticate the same way: with the
+            # credential only for remote targets. Otherwise a credential that is valid on the remote
+            # computers of a mixed list but not locally would reject the local computer although the
+            # actual operation would succeed. Without a credential the implicit identity is used,
+            # which fails when it cannot authenticate to the target - for example when the caller
+            # itself runs in a remoting session with a network logon token (double hop).
+            $useCredentialForPreflight = $Credential -and -not ([DbaInstanceParameter]$computer).IsLocalHost
             try {
-                # Pass the credential so that the connectivity test authenticates the same way as the
-                # Invoke-Command2 calls below. Without it, the test uses the implicit identity, which
-                # fails when that identity cannot authenticate to the target - for example when the
-                # caller itself runs in a remoting session with a network logon token (double hop).
-                if ($Credential) {
+                if ($useCredentialForPreflight) {
                     $null = Test-PSRemoting -ComputerName $Computer -Credential $Credential -EnableException
                 } else {
                     $null = Test-PSRemoting -ComputerName $Computer -EnableException
                 }
             } catch {
-                if ($Credential) {
+                if ($useCredentialForPreflight) {
                     Stop-Function -Message "Failure on $computer" -ErrorRecord $_ -Continue
                 } else {
                     Stop-Function -Message "Failure on $computer. If this session itself runs in a remote session (for example via WinRM or Ansible), its network logon cannot authenticate to $computer (double hop). Pass -Credential or connect with an authentication that supports delegation, like CredSSP." -ErrorRecord $_ -Continue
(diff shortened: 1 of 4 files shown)


potatoqualitee (changes requested):
Credential forwarding is correct for remote targets, but this regresses mixed local/remote lists. `Invoke-Command2` executes localhost under the current process identity and ignores `-Credential`; the new preflight and service-discovery calls now authenticate localhost with the supplied remote credential. A credential valid on remote hosts but invalid/unauthorized locally can reject the local target even though the actual local operation would succeed. Please normalize the target and forward `Credential` only for non-local hosts (or otherwise align preflight/discovery with the actual execution identity), with separate local and remote tests. The new localhost mock currently enshrines the mismatch.

andreasjordan:
Fixed in 682f94807 - both commands normalize the loop target back to a `DbaInstanceParameter` (the loop had flattened it to a string) and compute `\ = \ -and -not IsLocalHost`, mirroring exactly the decision `Invoke-Command2` makes for the execution identity. The credential now reaches `Test-PSRemoting` and `Get-DbaService` only for remote targets.

The mock coverage is split as requested: the remote case keeps asserting the credential is forwarded (`Get-DbaPrivilege`'s existing test already targeted a fake remote and stays), and new local cases assert no credential reaches preflight or discovery for the local computer - the previous `Set-DbaPrivilege` mock that enshrined the mismatch is now the remote-targeted variant. Both local cases are red on the old code; 5/5 + 5/5 via the lab harness, warning-free.

potatoqualitee (approved):

PR #10618, opened 2026-08-27
Data generation - Stop on configuration errors and write the generated unique values

andreasjordan:
Fixes three defects in the data generation family, found while working PR #10517 and verified against the lab.

## Configuration errors did not stop the generation

`Invoke-DbaDbDataGenerator` called `Test-DbaDbDataGeneratorConfig -EnableException` without assigning the result. The findings of that command are output objects, not exceptions, so the surrounding `try/catch` could never fire: an invalid configuration was reported *into the output stream of the generator itself*, mixed with the per-table results, and the generation ran anyway. The command now collects the findings and stops with "Errors found testing the configuration file.", returning the finding objects - the same pattern `Invoke-DbaDbDataMasking` already uses.

## The two supported data type lists had drifted

Stopping on findings surfaced a second defect: the `$supportedDataTypes` list in `Test-DbaDbDataGeneratorConfig` was missing `float`, `guid`, `numeric`, `real`, `smallint` and `tinyint`, all of which `Invoke-DbaDbDataGenerator` supports. Harmless while the findings were ignored - but with the stop in place, a configuration fresh from `New-DbaDbDataGeneratorConfig` for a table with such a column would have been rejected. The lists now match, and both carry a comment pointing at the other.

## The unique index branch never delivered its values

The branch that pre-generates values for tables with a unique index has been broken since the command was created in 2019, in four ways:

- the insert read the values with `$uniqueValues[$rowNumber]`, and `$rowNumber` is never assigned anywhere - so every row read `$null`,
- `$uniqueValueColumns` was only filled inside the collision-retry copy of the code, so in the common no-collision case the main loop did not even look at the pre-generated values and inserted fresh independent random draws,
- a collision retry started from an empty object, wiping the values already generated for earlier indexes of the same row,
- the collision check matched objects with `-match`, i.e. against a stringified regex.

The branch is rewritten: one candidate per row, generated for all configured unique index columns, checked per index against the earlier rows, retried up to 100 times, and the insert reads it by the row loop index. The unique value branch also moved ahead of the nullable branch, because a second NULL would already violate the index.

## Tests

- The existing invalid-configuration tests now assert the up-front rejection and that the finding objects name the column; the in-command column checks remain as the only validation for configurations fetched over http, which skip `Test-DbaDbDataGeneratorConfig`.
- New regression: ten rows into a unique-indexed int column with a ten-value domain must form a permutation. On the old code this fails with "Cannot insert duplicate key"; verified red-on-old (7 tests fail), green-on-new (17/17).
- New regression in `Test-DbaDbDataGeneratorConfig.Tests.ps1`: a configuration generated for a table with the six previously missing numeric types produces no findings (5 findings on the old code).

`Invoke-DbaDbDataMasking` was checked for the same defects: its unique index handling uses a different mechanism (temp table keyed by a properly incremented row number) and is not affected.

### DIFF
--- a/public/Invoke-DbaDbDataGenerator.ps1
+++ b/public/Invoke-DbaDbDataGenerator.ps1
@@ -299,8 +299,27 @@ function Invoke-DbaDbDataGenerator {
 
                         # Only the index columns that are part of the configuration get a value generated
                         # here. The insert statement further down only names configured columns, so an
-                        # index column outside the configuration is left to SQL Server.
-                        $uniqueValueColumns = @($uniqueIndexes.IndexedColumns.Name | Select-Object -Unique | Where-Object { $PSItem -in $tableobject.Columns.Name })
+                        # index column outside the configuration is left to SQL Server. Identity columns
+                        # are excluded even when they carry the unique index - the typical primary key:
+                        # their values come from the identity branch of the insert, and random values
+                        # here would jump the identity seed toward the type limit.
+                        $identityColumnNames = @($tableobject.Columns | Where-Object Identity | Select-Object -ExpandProperty Name)
+                        $uniqueValueColumns = @($uniqueIndexes.IndexedColumns.Name | Select-Object -Unique | Where-Object { $PSItem -in $tableobject.Columns.Name -and $PSItem -notin $identityColumnNames })
+
+                        # The collision check must also see what is already persisted: without
+                        # TruncateTable, a candidate matching an existing indexed value would pass
+                        # the generation and then fail at insert time.
+                        $existingUniqueValues = @()
+                        if ($uniqueValueColumns.Count -gt 0 -and -not $tableobject.TruncateTable) {
+                            $existingColumnList = "[" + ($uniqueValueColumns -join "], [") + "]"
+                            $existingValuesQuery = "SELECT DISTINCT $existingColumnList FROM [$($tableobject.Schema)].[$($tableobject.Name)]"
+                            try {
+                                $existingUniqueValues = @(Invoke-DbaQuery -SqlInstance $server -Database $db.Name -Query $existingValuesQuery -EnableException)
+                            } catch {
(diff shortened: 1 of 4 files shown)


potatoqualitee (changes requested):
The configuration-validation fixes are sound, but the rewritten uniqueness path has two blocking data-generation defects:

1. `HasUniqueIndex` includes ordinary unique primary-key indexes, generated configs include identity columns, and the new unique-value branch now precedes the existing identity branch. A normal `int IDENTITY PRIMARY KEY` is therefore inserted as random explicit values instead of `IDENT_CURRENT + increment`, potentially jumping the identity seed toward the type limit. Preserve the identity path and add an identity-PK regression.
2. Collision checking seeds `$uniqueValues` only with rows generated in this invocation. With the default `TruncateTable = false`, candidates can duplicate existing persisted indexed values. For example, an existing unique value `1` plus a one-row `MinValue = MaxValue = 1` config deterministically fails at insert. Include existing indexed key combinations (or an equivalent persisted-value check) when appending, with regression coverage.

andreasjordan:
Fixed in b34d879ba - both points:

1. **Identity path preserved:** identity columns are excluded from `\` even when they carry the unique index, so the insert's identity branch (`IDENT_CURRENT + increment`) handles them again. Cleaner than reordering the branches: the unique machinery never generates for them at all, and index collisions on the identity column cannot happen by construction. Regression: an `int IDENTITY PRIMARY KEY` table must come out as `1,2,3,4,5` - red on the old code with `207,477,730,800,836`.
2. **Persisted values respected:** with `TruncateTable` disabled, the distinct existing values of the configured index columns are read up front into a separate collection that the collision check walks alongside the rows generated in this invocation (separate on purpose - the insert consumes `\` by row index, which seeding would misalign). Your deterministic example is the regression: existing value `1` plus a `MinValue = MaxValue = 1` one-row config now stops cleanly with "Could not generate a unique value" before any insert - red on the old code, which crashed on the duplicate key at insert time.

22/22 via the lab harness, warning-free (outside the asserted warning), lab left clean.

potatoqualitee (approved):

PR #10619, opened 2026-08-27
Testing New-DbaDbDataGeneratorConfig - Guard columntypes.json against the randomizer types

andreasjordan:
`New-DbaDbDataGeneratorConfig` assigns `MaskingType` and `SubType` from `bin\datamasking\columntypes.json` when a column name matches one of its synonyms. Nothing tied that file to the randomizer types in `bin\randomizer\en.randomizertypes.csv`, so a pair drifting apart produces configurations that `Test-DbaDbDataGeneratorConfig` rejects. PR #10517 added the equivalent guard for the randomizer types against the Bogus library (`Get-DbaRandomizedType.Tests.ps1`); this is the missing half for `columntypes.json`.

The new unit test resolves every one of the 22 pairs through `Get-DbaRandomizedType` and reports each unknown combination by its `TypeName`. Verified in the lab: green as-is, and deliberately breaking one pair fails the test with `Bic: Finance/DoesNotExist`.

### DIFF
--- a/tests/New-DbaDbDataGeneratorConfig.Tests.ps1
+++ b/tests/New-DbaDbDataGeneratorConfig.Tests.ps1
@@ -25,6 +25,30 @@ Describe $CommandName -Tag UnitTests {
             Compare-Object -ReferenceObject $expectedParameters -DifferenceObject $hasParameters | Should -BeNullOrEmpty
         }
     }
+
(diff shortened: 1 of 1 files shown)


potatoqualitee (approved):
Reviewed the complete current head, read-only data-source behavior, pair matching, failure diagnostics, and exact-head CI evidence. No material correctness or test-isolation issues found.

PR #10622, opened 2026-08-28
Get-DbaWaitingTask - Harden flaky test setup and fix always-null InfoUrl

andreasjordan:
## Summary

The full test suite run of 2026-08-28 (setC config, FCI01 as InstanceSingle) failed only one file: `Get-DbaWaitingTask.Tests.ps1`, with both assertions comparing against `$null`. Standalone the test passes, so this is a timing flake, and while fixing the test a small command defect turned up.

### Test hardening

The integration test starts a background job that imports dbatools and runs a `WAITFOR DELAY`, then gave it a hard 30 seconds to show up on the instance. On a machine under load (four hours into a full suite run) importing the module in a fresh job can exceed that, the spid stays `$null`, and the tests fail with unhelpful messages like `Expected like wildcard '*WAITFOR*' to match $null`.

- The setup now waits up to 60 seconds for the session and **throws with the job output** if it never appears, so a future failure says what actually went wrong.
- Instead of a single blind attempt one second after the session appears, it polls up to 30 seconds for the waiting task to be reported.
- The module path for the job is derived from the loaded module instead of the hard-coded `C:\Github\dbatools\dbatools.psm1`.
- Setup runs with `EnableException` per test policy.

### Command fix: InfoUrl was always null

The query aliases the SQLSkills documentation link as `[URL]`, but the output object reads `$row.InfoUrl` - so the `InfoUrl` property has been silently `$null` since the command was added in 2018 (the `.OUTPUTS` documentation describes it as a real property). The alias now matches, and a new test asserts the URL comes back for the WAITFOR wait type.

## Verification

All 4 tests (including the new InfoUrl regression test) pass against FCI01 (clustered default instance, SQL Server 2022) via the lab test harness.

### DIFF
--- a/public/Get-DbaWaitingTask.ps1
+++ b/public/Get-DbaWaitingTask.ps1
@@ -110,7 +110,7 @@ function Get-DbaWaitingTask {
                 [er].[database_id] AS [DbId],
                 [est].[text] AS [SqlText],
                 [eqp].[query_plan] AS [QueryPlan],
-                CAST ('https://www.sqlskills.com/help/waits/' + [owt].[wait_type] AS XML) AS [URL]
+                CAST ('https://www.sqlskills.com/help/waits/' + [owt].[wait_type] AS XML) AS [InfoUrl]
             FROM sys.dm_os_waiting_tasks [owt]
             INNER JOIN sys.dm_os_tasks [ot] ON
                 [owt].[waiting_task_address] = [ot].[task_address]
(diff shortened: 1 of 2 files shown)


potatoqualitee (approved):
Reviewed the complete current head, InfoUrl output contract, job/test timing hardening, cleanup, and available exact-head CI/lab evidence. No material correctness or compatibility issues found.

PR #10624, opened 2026-08-28
Get-DbaDbMailHistory - Make the Status filter work on case sensitive instances

andreasjordan:
## Summary

Found by the first suite run against a case sensitive instance (SQL05\SQL2025). The msdb view `sysmail_allitems` emits `sent_status` in lowercase (`unsent`, `sent`, `failed`, `retrying` - verified from the view definition), but the command''s documented `-Status` values are capitalized and were injected into the WHERE clause verbatim, so `-Status Sent` returned nothing on a case sensitive instance. The filter now lowercases both sides: `LOWER(sent_status) IN (''sent'')`.

The test fixture had a companion defect that masked this: it inserted `importance` and `sensitivity` as `''Normal''`, which violates msdb''s CHECK constraints on a case sensitive instance (`sysmail_OutMailImportanceMustBeValid` compares against uppercase `NORMAL`/`HIGH`/`LOW` - also verified against the live constraint definitions). The fixture now inserts the uppercase values, which is also what Database Mail itself writes.

## Verification

11/11 tests green on SQL05\SQL2025 (case sensitive) and on SQL03\SQL2019 (case insensitive), via the lab test harness. Sibling defect of the same shape in Get-DbaDbMailLog (`-Type`) gets its own PR.

### DIFF
--- a/public/Get-DbaDbMailHistory.ps1
+++ b/public/Get-DbaDbMailHistory.ps1
@@ -169,7 +169,9 @@ function Get-DbaDbMailHistory {
                 if ($Status) {
                     # The sysmail_allitems view emits sent_status in lowercase, so compare lowercased or the
                     # documented values (Sent, Unsent, Failed, Retrying) never match on a case sensitive instance.
-                    $Status = ($Status -join "', '").ToLower()
+                    # Invariantly: a culture-sensitive lowercase turns the I of FAILED or RETRYING into a
+                    # dotless i under Turkish rules, which never matches the ASCII tokens sysmail stores.
+                    $Status = ($Status -join "', '").ToLowerInvariant()
                     $wherearray += "LOWER(sent_status) IN ('$Status')"
                 }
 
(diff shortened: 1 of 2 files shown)


potatoqualitee (changes requested):
The new normalization uses current-culture `.ToLower()`. `ValidateSet` accepts case-insensitive inputs such as `FAILED` and `RETRYING`; under Turkish or Azerbaijani culture those become strings containing dotless `ı`, so the generated filter cannot match Database Mail's invariant ASCII lowercase tokens. Please use `.ToLowerInvariant()` (or explicit invariant culture) and add focused culture coverage if practical.

andreasjordan:
Fixed in fa136207c - `.ToLowerInvariant()` with a comment naming the Turkish-I trap.

The requested culture coverage found the failure mode is even harsher than the review describes: the lowercased value is assigned back to the `ValidateSet` parameter variable, so under tr-TR the old code did not just miss the match - it terminated with *"the value faıled is not a valid value for the Status variable"*. The new regression adds a failed-status fixture row and runs `-Status FAILED` under `tr-TR` (uppercase I is the letter that goes dotless): red on the old code with exactly that validation error, green on the fix, 12/12 via the lab harness, warning-free, lab left clean.

potatoqualitee (approved):

PR #10625, opened 2026-08-28
Get-DbaDbMailLog - Make the Type filter work on case sensitive instances

andreasjordan:
## Summary

Sibling of #10624, found by the same first suite run against a case sensitive instance. The msdb view `sysmail_event_log` emits `event_type` in lowercase (`success`, `information`, `warning`, `error` - verified from the view definition), and the command itself relies on that in its SELECT, where a CASE maps the lowercase values to display case. But the `-Type` filter injected the documented capitalized values verbatim into the WHERE clause, so `-Type Information` returned nothing on a case sensitive instance. The filter now compares `LOWER(event_type)` against a lowercased value list.

## Verification

10/10 tests green on SQL05\SQL2025 (case sensitive) and SQL03\SQL2019 (case insensitive), via the lab test harness. No test changes needed - the existing `-Type Information` test is the regression test, red on the old code on the case sensitive instance and green on the fix.

### DIFF
--- a/public/Get-DbaDbMailLog.ps1
+++ b/public/Get-DbaDbMailLog.ps1
@@ -131,8 +131,10 @@ function Get-DbaDbMailLog {
                 }
 
                 if ($Type) {
-                    $combinedtype = $Type -join "', '"
-                    $wherearray += "event_type IN ('$combinedtype')"
+                    # The sysmail_event_log view emits event_type in lowercase, so compare lowercased or the
+                    # documented values (Error, Warning, Information, Success, Internal) never match on a case sensitive instance.
+                    $combinedtype = ($Type -join "', '").ToLower()
+                    $wherearray += "LOWER(event_type) IN ('$combinedtype')"
                 }
 
                 $wherearray = $wherearray -join ' AND '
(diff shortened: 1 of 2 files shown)


potatoqualitee (changes requested):
The new normalization uses culture-sensitive `.ToLower()`. Under Turkish or Azerbaijani process culture, documented values such as `Information` and `Internal` become strings containing dotless `ı`, while `sysmail_event_log` emits invariant ASCII lowercase values; the filter still returns no rows. Because these are culture-independent SQL identifiers, please use `.ToLowerInvariant()` and add focused culture coverage if practical.

andreasjordan:
Fixed in c305c643a - `.ToLowerInvariant()` with a comment naming the Turkish-I trap.

On the culture coverage: the primitive is broken exactly as described - probed on this stack, `"INFORMATION".ToLower()` under `tr-TR` yields `ınformatıon` (U+0131). The end-to-end repro is subtler though: on this lane's Latin1 CI instance the dotless i in the varchar literal is best-fitted back to a plain `i` during parse, so the unfixed command still matches - the visible failure needs a Turkish-collation target database. The new `tr-TR` regression therefore pins the invariant (and its comment says plainly that it is green on the unfixed code here), unlike the sibling #10624, where the same bug terminates the command outright because the lowercased value is reassigned into the ValidateSet parameter. 11/11 via the lab harness, warning-free, lab left clean.

potatoqualitee (approved):

PR #10626, opened 2026-08-28
Install-DbaWhoIsActive - Detect an existing installation on case sensitive instances

andreasjordan:
## Summary

Found by the first suite run against a case sensitive instance. The command''s existence check queries `sys.procedures WHERE ... name LIKE ''%sp_WhoisActive%''` - lowercase `is` - but the procedure Adam Machanic ships is `sp_WhoIsActive`. Every case insensitive instance forgave that, and on a case sensitive instance the check counts zero, so reinstalling reports `Installed` instead of `Updated`. The check now compares `LOWER(name) LIKE ''%sp_whoisactive%''`.

The output object''s `Name` property (and its `.OUTPUTS` documentation) also said `sp_WhoisActive`; both now say `sp_WhoIsActive`, the name of the procedure that is actually installed. The test assertions were tightened to the exact casing.

## Verification

3/3 tests green on SQL05\SQL2025 (case sensitive) and SQL03\SQL2019 (case insensitive), via the lab test harness. The existing "Should update" test is the regression test - red on the old code on the case sensitive instance, green on the fix.

### DIFF
--- a/public/Install-DbaWhoIsActive.ps1
+++ b/public/Install-DbaWhoIsActive.ps1
@@ -198,7 +198,11 @@ function Install-DbaWhoIsActive {
             }
             if ($PSCmdlet.ShouldProcess($instance, "Installing sp_WhoisActive")) {
                 try {
-                    $ProcedureExists_Query = "SELECT COUNT(*) [proc_count] FROM sys.procedures WHERE is_ms_shipped = 0 AND LOWER(name) LIKE '%sp_whoisactive%'"
+                    # Every shipped script creates dbo.sp_WhoIsActive, so the exact schema-qualified
+                    # lookup is the reliable detection. The LOWER-LIKE probe it replaces depended on
+                    # the database collation twice: a case sensitive collation never matched the
+                    # lowercase pattern, and under Turkish rules LOWER() turns the I into a dotless i.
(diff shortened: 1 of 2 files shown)


potatoqualitee (changes requested):
`LOWER(name) LIKE '%sp_whoisactive%'` remains dependent on the target database collation. On a Turkish case-sensitive collation, the uppercase `I` in the canonical `sp_WhoIsActive` lowercases to dotless `ı`, so reinstall detection still misses it and reports `Installed`. Since every shipped script creates/alters `dbo.sp_WhoIsActive`, please detect that exact schema-qualified procedure instead (for example `OBJECT_ID(N'dbo.sp_WhoIsActive', N'P') IS NOT NULL`) and cover a locale-sensitive CS collation if practical.

andreasjordan:
Fixed in 66b61e337 - detection is now exactly your suggestion, `OBJECT_ID(N'dbo.sp_WhoIsActive', N'P') IS NOT NULL`, with a comment recording both collation traps the LOWER-LIKE probe had (case sensitive collations and the Turkish dotless i).

The requested locale coverage is in: a new context installs twice into a `Turkish_CS_AS` database - red on the old code with exactly the predicted symptom (`Expected 'Updated' but was 'Installed'`), green on the fix. The shipped script itself deploys cleanly under that collation. 4/4 via the lab harness, warning-free, lab left clean.

potatoqualitee (approved):

PR #10628, opened 2026-08-28
Move the two mixed-instance test files wholly into the MULTI lane

andreasjordan:
## Summary

`New-DbaDatabase.Tests.ps1` and `Install-DbaCommunitySoftware.Tests.ps1` were the only test files in the tree referencing `$TestConfig.InstanceSingle` *and* `$TestConfig.InstanceMulti*` at once. The scenario autodetection in `Get-TestsForScenario` excludes a file that also matches another scenario''s instance string - in both directions - so these two files ran in **no scenario-scoped run at all**: a `-Scenario SINGLE` or `-Scenario MULTI` lab run silently skipped them, and on CI `New-DbaDatabase.Tests.ps1` ran in **no lane whatsoever**, because its membership in the `2008R2SP2Express` group also removed it from the `default` leftover lane (a separate one-line fix for that leftover-lane computation follows in its own PR).

Both files genuinely need two instances, so per the lightest-instance-per-file rule they are MULTI files: their single-instance work now runs on `InstanceMulti1`, and both files autodetect cleanly into MULTI. The misleading `$InstanceSingle` variable in `New-DbaDatabase.Tests.ps1` - which actually held a connection to `InstanceMulti1` - is renamed `$serverMulti1`, with `$instance3` becoming `$serverMulti2`.

## Verification

Both files verified green against `SQL03\SQL2025` + `SQL03\SQL2022` via the lab harness: New-DbaDatabase 13/13, Install-DbaCommunitySoftware 21 passed / 4 skipped (the documented PowerShell-edition skips). Autodetection re-simulated after the change: both files match exactly one scenario (MULTI).

### DIFF
--- a/tests/Install-DbaCommunitySoftware.Tests.ps1
+++ b/tests/Install-DbaCommunitySoftware.Tests.ps1
@@ -134,10 +134,10 @@ Describe $CommandName -Tag IntegrationTests -Skip:([bool]$env:appveyor) {
             $PSDefaultParameterValues["*-Dba*:EnableException"] = $true
 
             $multiToolDb = "dbatoolsci_community_$(Get-Random)"
-            $null = New-DbaDatabase -SqlInstance $TestConfig.InstanceSingle -Name $multiToolDb
+            $null = New-DbaDatabase -SqlInstance $TestConfig.InstanceMulti1 -Name $multiToolDb
 
             $splatMultiTool = @{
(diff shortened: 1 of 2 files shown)


potatoqualitee (approved):
Reviewed the complete current head, MULTI selector routing, instance substitutions, preserved test topology, and exact-head matrix evidence. No material coverage or correctness issues found.

PR #10629, opened 2026-08-28
Keep version groups out of the leftover-lane exclusion

andreasjordan:
## Summary

The leftover lane - any `SCENARIO` value that is not a group key, which is how the CI matrix entry `default` works - is computed in `Get-TestsForBuildScenario` as all tests minus every valid group in `pester.groups.ps1`. That subtraction included the explicit `2008R2SP2Express` list, whose lane existed on AppVeyor but is not in the current CI matrix. Consequence: a file on that list that does not land in an autodetect scenario ran in **no lane at all**. Concretely: `Invoke-DbaDbLogShipping.Tests.ps1` (unit-only, no instance references - even its parameter validation never ran on CI), and `New-DbaDatabase.Tests.ps1` until #10628.

The autodetect path already anticipates exactly this class of problem: when computing tied functions it deliberately skips groups named `20*` so that version groups stay additive - "also run these on an old-version lane" - rather than subtractive. This PR applies the same one-line filter to the leftover-lane computation, with a comment explaining why. The `2008R2SP2Express` group itself stays untouched and ready for an old-version lane to return.

## Verification

Computed the `default` lane with the fix in a fresh process: 219 files, now including the orphaned `Invoke-DbaDbLogShipping.Tests.ps1`, while files belonging to autodetect scenarios (spot check: `Get-DbaDatabase.Tests.ps1`) remain excluded. Scenario-scoped lanes are unaffected - the change only touches the `else` branch for non-key scenario names.

### DIFF
--- a/tests/Install-DbaCommunitySoftware.Tests.ps1
+++ b/tests/Install-DbaCommunitySoftware.Tests.ps1
@@ -134,10 +134,10 @@ Describe $CommandName -Tag IntegrationTests -Skip:([bool]$env:appveyor) {
             $PSDefaultParameterValues["*-Dba*:EnableException"] = $true
 
             $multiToolDb = "dbatoolsci_community_$(Get-Random)"
-            $null = New-DbaDatabase -SqlInstance $TestConfig.InstanceSingle -Name $multiToolDb
+            $null = New-DbaDatabase -SqlInstance $TestConfig.InstanceMulti1 -Name $multiToolDb
 
             $splatMultiTool = @{
-                SqlInstance = $TestConfig.InstanceSingle
+                SqlInstance = $TestConfig.InstanceMulti1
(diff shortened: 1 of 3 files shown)


andreasjordan:
The `test (default)` failure on the first run was the dependency on #10628 proving itself: this branch released `New-DbaDatabase.Tests.ps1` into the leftover `default` lane, which provisions no instances, so all 10 of its integration tests failed at connect time. Merged the #10628 branch in so this PR is green standalone; the overlapping changes are identical, so merge order between the two does not matter.

potatoqualitee (approved):
Reviewed the complete current head, version-group/additive selector semantics, final MULTI routing, and exact-head matrix evidence. No material CI coverage or correctness issues found.

PR #10630, opened 2026-08-29
Testing Find-DbaCommand - Restore the index file the Rebuild test overwrites

andreasjordan:
## Summary

`Find-DbaCommand.Tests.ps1` runs `Find-DbaCommand -Pattern snapshot -Rebuild`, and the rebuild path writes the regenerated index straight over `bin\dbatools-index.json` in the module root. When the module runs from a git working copy, that overwrites a tracked release asset - the file is committed only by the release process (v2.8.0 through v2.8.4 in the history) - so every suite run left the tree with a permanent 13 MB modification.

The fix stays in the test: `BeforeAll` copies the index file to the test temp directory, `AfterAll` copies it back and removes the backup. The `-Rebuild` coverage is unchanged - the rebuild still executes against the real file and the assertion still reads its output - only the leftover goes away. Ignoring the file instead would not work (`.gitignore` has no effect on tracked files), and untracking it would change what the release ships.

## Verification

Ran the file via the lab harness: 7/7 green (1 unit, 6 integration), and `git status` shows a clean `bin\dbatools-index.json` after the run - before the fix it showed as modified after every run that included this file.

### DIFF
--- a/tests/Find-DbaCommand.Tests.ps1
+++ b/tests/Find-DbaCommand.Tests.ps1
@@ -25,6 +25,21 @@ Describe $CommandName -Tag UnitTests {
 }
 
 Describe $CommandName -Tag IntegrationTests {
+    BeforeAll {
+        # The -Rebuild test regenerates bin\dbatools-index.json in place. When the module runs from
+        # a git working copy, that overwrites a tracked release asset and leaves the tree modified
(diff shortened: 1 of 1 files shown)


potatoqualitee (approved):
Reviewed the complete current head, index backup/restore lifecycle, test isolation behavior, and exact-head CI evidence. The normal repository test path is corrected and no material blocker was found.

PR #10631, opened 2026-08-29
Restore-DbaDatabase - Honor EnableException in the backup information pipeline

### LINKED ISSUE #10621: Restore-DbaDatabase -EnableException restores nothing with only a warning when no full backup anchors the chain
### Summary

When `Restore-DbaDatabase` is given backup files that contain a differential or log backup but no full backup to anchor the chain, it warns and returns nothing - even with `-EnableException`, which promises "enables you to catch exceptions with your own try/catch". A caller that relies on the exception (for example a test fixture, or any script that restores and then works with the database) continues as if the restore had happened.

### Repro (measured on SQL03\SQL2019, dbatools development)

```powershell
$PSDefaultParameterValues["*-Dba*:EnableException"] = $true
# db restored from any full backup, then:
$full = Backup-DbaDatabase -SqlInstance $server -Database $db -Type Full -Path $dir -FilePath "full.bak"
$diff = Backup-DbaDatabase -SqlInstance $server -Database $db -Type Diff -Path $dir -FilePath "diff.bak"
$log  = Backup-DbaDatabase -SqlInstance $server -Database $db -Type Log  -Path $dir -FilePath "log.trn"

# no full in the set - warns, does not throw, returns nothing:
Restore-DbaDatabase -SqlInstance $server -Path $diff.BackupPath, $log.BackupPath -DatabaseName $db -WithReplace
# WARNING: [Select-DbaBackupInformation] Fullname property not found. This could mean that a full backup
# could not be found or the command must be re-run with the -Continue switch.
```

### Mechanism

`Select-DbaBackupInformation` correctly calls `Stop-Function` when it cannot find a full backup (public/Select-DbaBackupInformation.ps1, the two "Fullname property not found" sites). But the splat that `Restore-DbaDatabase` builds for the call (public/Restore-DbaDatabase.ps1, around line 857) does not pass `EnableException` on, so inside `Select-DbaBackupInformation` the stop degrades to a warning, control returns, and `Restore-DbaDatabase` proceeds with an empty `$FilteredBackupHistory` to a successful-looking no-op.

### Real-world impact

This is exactly how `Get-DbaDbRestoreHistory.Tests.ps1` was flaky for years: its fixture restored diff+log without the full, which only worked because the default backup file name has minute resolution and the full and diff usually share one file. On a minute boundary the restore silently did nothing and six tests failed downstream (fixed on the test side in #10620).

### Suggested direction

Pass `EnableException = $EnableException` in the `Select-DbaBackupInformation` splat, and check the other internal `*-DbaBackupInformation` calls in the same pipeline (`Get-`, `Format-`, `Test-`) for the same gap. Whether an empty selection after a *warning* should also stop the non-exception path is a behavior question worth deciding at the same time.

andreasjordan:
## Summary

Fixes #10621. When `Restore-DbaDatabase` is given backups without a full to anchor the chain, `Select-DbaBackupInformation` correctly calls `Stop-Function` - but the splat `Restore-DbaDatabase` builds for it never passed `EnableException` on, so under `-EnableException` the stop degraded to a warning, control returned, and the restore proceeded over an empty `$FilteredBackupHistory` to a successful-looking no-op. A caller relying on the promised exception (a test fixture, or any script that restores and then uses the database) continued as if the restore had happened - exactly how `Get-DbaDbRestoreHistory.Tests.ps1` was flaky for years (#10620).

## What changed

- `EnableException = $EnableException` added to the `Select-DbaBackupInformation` splat and the `Get-DbaBackupInformation` splat, and to the per-file `Get-DbaBackupInformation` call on the SQL Server 2000 path (now a splat, since it grew to three parameters).
- The "No backups passed through" bail-out is now a `Stop-Function` instead of a plain `Write-Message -Level Warning`, so it too honors `-EnableException`.

- **Without `-EnableException` nothing changes**: `Stop-Function` writes the identical warning and returns, so the default warn-and-return behavior is untouched. The open behavior question from #10621 - whether an empty selection should also stop the non-exception path - is left for a maintainer decision.
- The `Test-DbaBackupInformation` call already passed `EnableException = $true` inside try/catch and stays as it is.
- `Format-DbaBackupInformation` contains no `Stop-Function` at all, so there is nothing to propagate.
- `Get-DbaAgBackupHistory` has the same gap on its `Select-DbaBackupInformation` call and gets its own pull request, one command per PR.

## Tests

Three regression tests in a new context, each backup written to its own file so the minute-resolution default name cannot hide the missing full by merging it into the diff's file:

- diff+log without the full throws `*Fullname property not found*` with `-EnableException` (the exact repro from the issue),
- the same input still warns and returns nothing without it,
- an existing directory with no backups throws `*No backups passed through*` with `-EnableException`.

Full file verified via the lab harness on SQL03\SQL2019: 84 tests, 76 passed, 0 failed, 7 skipped (the documented StopAt and Azure skips), lab left clean.

### DIFF
--- a/public/Restore-DbaDatabase.ps1
+++ b/public/Restore-DbaDatabase.ps1
@@ -676,7 +676,12 @@ function Restore-DbaDatabase {
 
         if ($RestoreInstance.VersionMajor -eq 8 -and $true -ne $TrustDbBackupHistory) {
             foreach ($file in $Path) {
-                $bh = Get-DbaBackupInformation -SqlInstance $RestoreInstance -Path $file
+                $splatBackupInformation = @{
+                    SqlInstance     = $RestoreInstance
+                    Path            = $file
+                    EnableException = $EnableException
+                }
+                $bh = Get-DbaBackupInformation @splatBackupInformation
                 $bound = $PSBoundParameters
                 $bound['TrustDbBackupHistory'] = $true
                 $bound['Path'] = $bh
@@ -752,6 +757,7 @@ function Restore-DbaDatabase {
                     IgnoreLogBackup     = $IgnoreLogBackup
                     StorageCredential   = $StorageCredential
                     NoXpDirRecurse      = $NoXpDirRecurse
+                    EnableException     = $EnableException
                 }
                 $BackupHistory += Get-DbaBackupInformation @parms
             }
@@ -815,7 +821,7 @@ function Restore-DbaDatabase {
         }
         if ($PSCmdlet.ParameterSetName -like "Restore*") {
             if ($BackupHistory.Count -eq 0 -and $RestoreInstance.VersionMajor -ne 8) {
-                Write-Message -Level Warning -Message "No backups passed through. `n This could mean the SQL instance cannot see the referenced files, the file's headers could not be read or some other issue"
+                Stop-Function -Message "No backups passed through. `n This could mean the SQL instance cannot see the referenced files, the file's headers could not be read or some other issue"
                 return
             }
             Write-Message -message "Processing DatabaseName - $DatabaseName" -Level Verbose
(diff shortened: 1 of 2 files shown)


potatoqualitee (approved):
Reviewed the complete current head, all backup-information propagation paths, empty-history behavior, regression coverage, and exact-head SQL CI evidence. No material correctness or compatibility issues found.

potatoqualitee:
thank you, wanted some triple checks before merging

PR #10632, opened 2026-08-29
Get-DbaAgBackupHistory - Fix single replica discovery and honor EnableException

andreasjordan:
## Summary

Companion to #10631, sweeping issue #10621 into the one other caller of `Select-DbaBackupInformation` - and the new tests immediately exposed a second, harder defect in the same command.

## Defect 1: single replica availability groups fail outright

The single-server branch assigns `.AvailabilityReplicas.Name` to `$serverList` as-is. With exactly one replica, SMO returns a scalar string, so `$serverList[0]` indexes the *string* and yields its first character (`S`), which the `DbaInstanceParameter` binder turns into null: every query of a single replica AG without `-Database` failed with "Cannot bind argument to parameter 'SqlInstance' because it is null". Fixed by forcing the list to an array, with a comment stating the constraint.

## Defect 2: the -Last branch ignored -EnableException

`$AgResults | Select-DbaBackupInformation -ServerName $AvailabilityGroup` never passed `EnableException` on, so a backup chain without a full to anchor it degraded the stop to a warning and the command returned nothing even when the caller asked for exceptions. `EnableException` now flows into that call. Without the switch nothing changes - the identical warning is written and nothing is returned, as before.

## Tests

The command had no integration tests at all; the old comment ("no availability group running in AppVeyor") predates the HADR lane. The new tests build a single replica ClusterType NONE availability group with a full+log chain and verify:

- the healthy chain comes back warning-free with Full and Log tagged with the AG name (this test alone catches defect 1),
- with a `-Since` window that holds only a log backup, `-EnableException` throws `Fullname property not found` and the default path warns and returns nothing.

`-Since` is used instead of wiping msdb history because `Backup-DbaDatabase` refuses a log backup for a database whose msdb history holds no full backup - found out the hard way in the lab.

Verified via the lab harness on SQL04\SQL2025: 4 integration tests passed, lab left clean; the unfixed code failed the same tests with the null bind.

### DIFF
--- a/public/Get-DbaAgBackupHistory.ps1
+++ b/public/Get-DbaAgBackupHistory.ps1
@@ -260,7 +260,9 @@ function Get-DbaAgBackupHistory {
             $replicaNames = ($server.AvailabilityGroups | Where-Object { $_.Name -in $AvailabilityGroup } ).AvailabilityReplicas.Name
             Write-Message -Level Verbose -Message "We have found these replicas: $replicaNames"
 
-            $serverList = $replicaNames
+            # With a single replica, Name returns a scalar string and indexing it with [0] below
+            # would yield its first character instead of the replica name, so force an array.
+            $serverList = @($replicaNames)
         }
 
         Write-Message -Level Verbose -Message "We have more than one server, so query them all and aggregate"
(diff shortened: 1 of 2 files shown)


potatoqualitee (approved):
Reviewed the complete current head, single-replica scalar handling, exception propagation, integration coverage, and exact-head HADR CI evidence. No material correctness or compatibility issues found.

PR #10633, opened 2026-08-29
Add-DbaRegServer - Close the store connection its verification call reopens

andreasjordan:
## Summary

Every `Add-DbaRegServer` call leaked one connection to the CMS instance, held open until the process exits. This is the standout from the first `-CheckSleepingConnections` full lab run (2026-08-28), which flagged the `Export-DbaRegServer` test file at **+33 sleeping sessions** - that file runs `Add-DbaRegServer` three times per test across eleven tests.

## Mechanism

After `Create()`, the command verifies the registration with `Get-DbaRegServer -SqlInstance $reggroup.ParentServer`. That `ParentServer` rides the store connection `Get-DbaRegServerGroup` had already closed (correctly, per #10572), so SMO silently **reconnects** it for the verification - and because `Connect-DbaInstance` then rightly reports it did not open this connection, `Disconnect-RegServer` inside `Get-DbaRegServer` declines to close it. The reconnected session stays checked out of the pool, owned by a server object nothing references. `Import-DbaRegServer` inherits the leak because it registers through this command.

## Fix

Mirror what `Add-DbaRegServerGroup` already does after its own verification call: `Disconnect-RegServer -Server $newserver` after the `Get`, in both the plain and the `-ServerObject` branch. The helper walks up to the store and closes only a connection this module opened.

## Measurements (SQL03\SQL2019)

- Before: five `Add-DbaRegServer` calls grew the sleeping-session count from 1 to 6; the isolated `Export-DbaRegServer` test file left **25** sessions behind.
- After: five calls stay at **1** (the shared pooled connection); the `Export-DbaRegServer` file's runner measurement drops from 25 to **1**.

## Tests

New context counts sleeping dbatools sessions through a server object opened once (a per-call counting command would open connections of its own), takes a warm-up call so the shared pooled connection exists before the baseline, then asserts three further calls add nothing. Verified red on the unfixed command ("Expected 3, but got 6") and green on the fix (9 tests, 0 failed via the lab harness).

### DIFF
--- a/public/Add-DbaRegServer.ps1
+++ b/public/Add-DbaRegServer.ps1
@@ -245,14 +245,21 @@ function Add-DbaRegServer {
                             $newserver.CredentialPersistenceType = "PersistLoginNameAndPassword"
                             $newserver.Create()
 
-                            Get-DbaRegServer -SqlInstance $reggroup.ParentServer -Name $Name -ServerName $ServerName | Where-Object Source -ne 'Azure Data Studio'
-                            # The verification call above reconnects the store connection that Get-DbaRegServerGroup
-                            # already closed, and nothing closes it again - one session per call stayed open. Close it
-                            # like Add-DbaRegServerGroup does; only a connection this module opened is ever closed.
-                            Disconnect-RegServer -Server $newserver
+                            # Buffered rather than streamed: a downstream early stop (| Select-Object -First 1)
+                            # would end this command at a streamed emission, before the disconnect below.
+                            $verifiedServer = Get-DbaRegServer -SqlInstance $reggroup.ParentServer -Name $Name -ServerName $ServerName | Where-Object Source -ne 'Azure Data Studio'
                         } catch {
                             Stop-Function -Message "Failed to add $ServerName on $target" -ErrorRecord $_ -Continue
+                        } finally {
+                            # The verification call reconnects the store connection that Get-DbaRegServerGroup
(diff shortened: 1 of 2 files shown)


potatoqualitee (changes requested):
The verification call is the operation that reconnects the inherited store, but cleanup is only the next statement in both branches. A terminating verification error—or normal downstream early termination after the streamed verification object, such as `Add-DbaRegServer ... | Select-Object -First 1`—skips `Disconnect-RegServer`, leaving the module-owned store session open. Please buffer the verification result, perform the ownership-aware disconnect in `finally`, then emit the buffered result, in both the `-ServerObject` and plain branches. Add coverage for an early-stopping or error path; the current test proves only complete successful enumeration.

andreasjordan:
Fixed in 6fdef39a1 - exactly the requested shape, in both branches: the verification result is buffered, the ownership-aware `Disconnect-RegServer` sits in a `finally` (null-guarded for a failed `New-Object`), and the buffered object is emitted after the cleanup. The `catch`'s `-Continue` still skips the emission on the error path, so error-path output is unchanged.

The requested coverage is in: the counting context gains an early-termination round - three `Add-DbaRegServer ... | Select-Object -First 1` calls must add no sleeping session. Red on the old code (`Expected 1, but got 3`), green on the fix, 10/10 via the lab harness, warning-free, lab left clean.

potatoqualitee (approved):

PR #10635, opened 2026-08-29
Get-DbaXESession - Close the cloned connection behind the XEStore

andreasjordan:
## Summary

Third fix from the sleeping-connection inventory (after #10633 and #10634). Every `Get-DbaXESession` call leaked one connection: it clones the SqlConnection for its `XEStore` and nothing ever closed the clone. Because the whole Extended Events family funnels through this command - `Start-DbaXESession` even re-queries through it after every start - the leak multiplied across Start, Stop, Remove, ConvertTo, Watch and the Export commands. The first `-CheckSleepingConnections` full lab run flagged Start-DbaXESession at +15, ConvertTo at +7 and Remove at +6.

## Fix

Disconnect the store connection after the sessions are enumerated and emitted. The emitted objects keep their `Store` note property fully functional: SFC transparently reconnects from the pool when a downstream command uses the store again - confirmed in the lab for piped `Start-DbaXESession` and `Stop-DbaXESession`, and pinned by a test that uses the store of an emitted object after the connection was returned.

## Measurements (SQL03\SQL2019)

- Three Get calls used to add three sleeping sessions; with the fix they add none. The regression test is red on the unfixed command ("Expected 4, but got 7").
- Runner measurements of the family test files: Start-DbaXESession **15 -> 7**, Remove-DbaXESession **6 -> 2**, ConvertTo-DbaXESession **7 -> 3**.

## Deliberately left for a follow-up

The residue in Start-DbaXESession comes from a different, smaller mechanism: when a piped object's store reconnects for an operation, that reopened connection stays checked out once per pipeline. Fixing that means consumer-side disconnects across several commands and deserves its own change with its own measurements.

## Tests

Same counting pattern as #10633/#10634: sleeping dbatools sessions counted through a server object opened once, warm-up call before the baseline, three further calls must add nothing - plus the store-reconnect assertion described above. All green via the lab harness (5 tests, 0 failed).

### DIFF
--- a/public/Get-DbaXESession.ps1
+++ b/public/Get-DbaXESession.ps1
@@ -107,46 +107,52 @@ function Get-DbaXESession {
                 Stop-Function -Message "Failure" -Category ConnectionError -ErrorRecord $_ -Target $instance -Continue
             }
 
-            Write-Message -Level Verbose -Message "Getting XEvents Sessions on $instance."
-            $xesessions = $XEStore.sessions
+            # The finally makes sure the cloned store connection is also closed when a downstream
+            # command ends the pipeline early (... | Select-Object -First 1 stops this command in
+            # the middle of the emission loop) and when a terminating error unwinds.
+            try {
+                Write-Message -Level Verbose -Message "Getting XEvents Sessions on $instance."
+                $xesessions = $XEStore.sessions
 
-            if ($Session) {
-                $xesessions = $xesessions | Where-Object { $_.Name -in $Session }
-            }
-
-            foreach ($x in $xesessions) {
-                $status = switch ($x.IsRunning) { $true { "Running" } $false { "Stopped" } }
-                $files = $x.Targets.TargetFields | Where-Object Name -eq Filename | Select-Object -ExpandProperty Value
-
-                $filecollection = $remotefile = @()
-
-                if ($files) {
-                    foreach ($file in $files) {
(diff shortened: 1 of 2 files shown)


potatoqualitee (changes requested):
The cloned XEStore connection is disconnected only after normal, complete enumeration. Because sessions are streamed from the inner loop, ordinary downstream early termination such as `Get-DbaXESession ... | Select-Object -First 1` stops upstream execution before the trailing disconnect; a terminating error after the store opens has the same leak. That leaves the central ownership bug unfixed on common exit paths. Please wrap store use and streamed output in `try`/`finally`, disconnect in `finally`, and add early-pipeline-termination coverage. Also clean up the store that the new reconnect test deliberately reopens so the test itself does not leave shared connection state behind.

andreasjordan:
Fixed in a7e4d7cb9 - all three points of the review:

- **Early termination and error paths:** enumeration and the streamed emission loop now sit inside a `try` whose `finally` disconnects the store, so `... | Select-Object -First 1` and terminating errors close the cloned connection like the normal path does. (Unlike the WhatIf case on #10634, this leak is real and provable here: `$XEStore.sessions` opens the connection eagerly before the first object streams.)
- **Regression coverage:** three early-terminated calls after the baseline must add no sleeping session - red on the old code (`Expected 2, but got 4`), green on the fix (6/6 via the lab harness, warning-free, lab left clean).
- **The reconnect test cleans up after itself:** it now returns the store connection its assertion deliberately reopens.

potatoqualitee (approved):

PR #10637, opened 2026-08-29
Get-DbaDatabase - Stop eating the caller loop on invalid parameter combination

andreasjordan:
## Summary

First fix from the `Stop-Function -Continue` sweep that PR #10636 triggered. The begin-block validation for `-ExcludeUser` plus `-ExcludeSystem` called `Stop-Function -Continue`, but **no loop encloses that call**: without `-EnableException`, the `continue` escapes the command and consumes an iteration of whatever loop the caller is running in. Because the begin block runs once before anything else, a script looping over instances with that parameter mistake loses **every** iteration - the regression test demonstrates a `foreach` over three elements completing zero of them on the unfixed command ("Expected 3, but got 0").

## Fix

Drop `-Continue`, add `return` - the `Test-FunctionInterrupt` guard at the top of the process block already handles the rest. Same shape as the two end-block sites fixed in Restore-DbaDatabase (#10636), with a comment stating why `-Continue` is forbidden there.

## The sweep behind it

An AST pass over all 1591 `Stop-Function -Continue`/`-SilentlyContinue` call sites in `public/` and `private/functions/` found **91 without an enclosing loop or switch** in their function body (31 in begin blocks, 23 in process, 25 in end, 12 in nested helpers that need call-graph triage). Each affected command will get its own PR; this one goes first because `Get-DbaDatabase` is the most used command in the module and the site fires on a plain parameter mistake.

## Tests

Lab harness on SQL03\SQL2019: 17 tests, 0 failed. The new test loops three times over the invalid combination, asserts all three iterations complete, and asserts the warning text - red on the unfixed command, green on the fix.

### DIFF
--- a/public/Get-DbaDatabase.ps1
+++ b/public/Get-DbaDatabase.ps1
@@ -244,7 +244,10 @@ function Get-DbaDatabase {
     begin {
 
         if ($ExcludeUser -and $ExcludeSystem) {
-            Stop-Function -Message "You cannot specify both ExcludeUser and ExcludeSystem." -Continue -EnableException $EnableException
+            # No -Continue here: the begin block has no enclosing loop, so the continue would escape
+            # the command and eat an iteration of whatever loop the caller runs in.
+            Stop-Function -Message "You cannot specify both ExcludeUser and ExcludeSystem." -EnableException $EnableException
+            return
         }
 
     }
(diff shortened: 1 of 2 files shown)


potatoqualitee (approved):
Reviewed the complete current head, begin/process interruption behavior, regression coverage, and exact-head CI evidence. No material correctness or compatibility issues found.

Issue #10638, opened 2026-08-29
Stop-Function -Continue without an enclosing loop escapes the command and corrupts the caller


andreasjordan:
### Summary

`Stop-Function -Continue` executes PowerShell's `continue`. When no loop encloses the call site inside the command's own body, that flow control does not stop at the function boundary: it unwinds **into the caller** and consumes an iteration of whatever loop the caller happens to be running in. A user restoring or querying in a `foreach` over instances silently skips elements when a command hits its error path; under Pester it corrupts the test runner itself, which surfaced for years as the undiagnosable `Cannot bind argument to parameter 'ErrorRecord' because it is null` crash documented above the skipped StopAt tests.

Two demonstrations, both lab-measured:

- #10636 fixed two `end`-block sites in `Restore-DbaDatabase`; before the fix, a plain script calling the command lost its own control flow mid-`try` (a marker line after the call was never reached while the `finally` still ran).
- #10637 fixed the `begin`-block validation in `Get-DbaDatabase`: a `foreach` over three elements completed **zero** iterations against the unfixed command, because the escaped `continue` consumed one per call.

The escape bites only the non-`EnableException` path for `-Continue` (with `EnableException`, `Stop-Function` throws first). `-SilentlyContinue` executes `continue` in **both** modes, so it can escape even under `EnableException`.

### Inventory

An AST sweep over all 1,591 `-Continue`/`-SilentlyContinue` sites in `public/` and `private/functions/` (walking each call's parents for a loop or `switch`, which also catches `continue`) found **91 sites with no enclosing loop**: 31 in `begin` blocks, 23 in `process`, 25 in `end`, and 12 inside nested helper functions where the `continue` binds dynamically at the helper's caller and needs per-call-site triage.

[...]

PR #10639, opened 2026-08-29
Test-DbaLastBackup - Stop eating the caller loop on inaccessible directories

andreasjordan:
## Summary

Second command fix from the #10638 inventory. The `-Path` branch of `Test-DbaLastBackup` validates `-DataDirectory` and `-LogDirectory` with `Stop-Function -Continue`, but unlike its twin checks inside the per-instance loop further down the file, **no loop encloses these two calls**. Without `-EnableException`, the `continue` escaped the command and consumed an iteration of whatever loop the caller was running in - the regression test loops three times over an inaccessible `DataDirectory` and completed **zero** iterations on the unfixed command ("Expected 3, but got 0").

## Fix

Stop and `return`, mirroring #10636/#10637 - and mirroring the two correct patterns that already sit three lines above these calls in the same branch (the missing `-Destination` check and the connection catch). The twin checks in the per-instance loop keep their `-Continue`, which is correct there: it moves to the next instance.

## Tests

Lab harness on SQL03\SQL2019: 17 tests, 0 failed. The new test uses the loop-counter pattern from #10637, asserting all three iterations complete and the warning text appears.

References #10638

### DIFF
--- a/public/Test-DbaLastBackup.ps1
+++ b/public/Test-DbaLastBackup.ps1
@@ -369,7 +369,11 @@ function Test-DbaLastBackup {
             if ($DataDirectory) {
                 if (-not (Test-DbaPath -SqlInstance $destserver -Path $DataDirectory)) {
                     $serviceAccount = $destserver.ServiceAccount
-                    Stop-Function -Message "Can't access $DataDirectory Please check if $serviceAccount has permissions." -Continue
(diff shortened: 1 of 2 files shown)


potatoqualitee (approved):
Reviewed the complete current head, both path guards, the intentionally retained in-loop sites, regression coverage, and exact-head CI evidence. No material correctness or compatibility issues found.

PR #10640, opened 2026-08-29
Get-DbaRandomizedValue - Stop eating the caller loop on invalid parameters

andreasjordan:
## Summary

Third command fix from the #10638 inventory, and the one with the best evidence that the bug bites in practice: **the command's own test suite had grown a workaround for it**. All ten validation guards in the begin block called `Stop-Function -Continue` with no enclosing loop, so without `-EnableException` the `continue` escaped the command and consumed an iteration of the caller's loop. The every-randomizer-type test loop pre-writes its result entry precisely because the escaped `continue` used to eat its `foreach` iteration - its comment even said so. The data masking and generation commands call this command per row.

## Fix

All ten guards now stop and `return` (the process block's `Test-FunctionInterrupt` guard handles the rest), same shape as #10636-#10639. The test workaround stays as a harmless safety net, its comment rewritten to document the history instead of the mechanism.

## Tests

Lab harness: 10 tests, 0 failed. New loop-counter regression test (a call without any type parameter, three iterations, all must complete): red on the unfixed command ("Expected 3, but got 0"), green on the fix.

References #10638


potatoqualitee (approved):
Reviewed the complete current head, all ten validation guards, begin/process interruption behavior, regression coverage, and exact-head CI evidence. No material correctness or compatibility issues found.

PR #10641, opened 2026-08-29
Import-DbaSpConfigure - Stop eating the caller loop and skipping its own cleanup

andreasjordan:
## Summary

Fourth command fix from the #10638 inventory, with an extra twist: the escape also leaked the command's own connections. The four begin-block guards (three sysadmin checks and the missing-file check) called `Stop-Function -Continue` with no enclosing loop. Without `-EnableException`, the `continue` escaped the command, consumed an iteration of the caller's loop - and because the whole command aborted mid-flight, the end block that disconnects the connections the command had opened for itself (the #10554 ownership cleanup from #10575) never ran, leaking them on every error path.

## Fix

Stop and `return` at all four sites, mirroring #10636-#10640. The `Test-FunctionInterrupt` guard in the process block handles the rest, and the end-block cleanup now runs on the error paths too.

## Tests

Lab harness on SQL03\SQL2019: 15 tests, 0 failed, lab left clean. New loop-counter regression test over a nonexistent `-Path`: red on the unfixed command ("Expected 3, but got 0"), green on the fix.

References #10638

### DIFF
--- a/public/Import-DbaSpConfigure.ps1
+++ b/public/Import-DbaSpConfigure.ps1
@@ -301,9 +301,10 @@ function Import-DbaSpConfigure {
         }
     }
     end {
-        if (Test-FunctionInterrupt) { return }
-
-        # Only close the connections that were opened here. See #10554.
+        # Only close the connections that were opened here, and close them before the interrupt
+        # return below: a begin-block guard sets the interrupt after a connection was already
+        # opened (a missing -Path, a failed sysadmin check), and returning first would leak it
+        # on every guard path. See #10554.
         if ($isNewServerConnection) {
             $server.ConnectionContext.Disconnect()
         }
@@ -313,6 +314,9 @@ function Import-DbaSpConfigure {
         if ($isNewDestinationConnection) {
             $destserver.ConnectionContext.Disconnect()
         }
+
+        # Only the finished message stays suppressed when the command was interrupted.
+        if (Test-FunctionInterrupt) { return }
 
         If ($Pscmdlet.ShouldProcess("console", "Showing finished message")) {
             Write-Message -Level Output -Message "SQL Server configuration options migration finished."
--- a/public/Import-DbaSpConfigure.ps1
+++ b/public/Import-DbaSpConfigure.ps1
@@ -135,7 +135,11 @@ function Import-DbaSpConfigure {
             }
 
             if (-not (Test-SqlSa -SqlInstance $sourceserver -SqlCredential $SourceSqlCredential)) {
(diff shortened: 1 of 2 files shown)


potatoqualitee (changes requested):
The four guards no longer escape the caller loop, but the PR still leaks command-owned connections on those paths. `Stop-Function` sets the function interrupt flag; the following `return` exits `begin`, `process` returns on that flag, and `end` then returns at its own `Test-FunctionInterrupt` guard before the disconnects below it. A missing `-Path`, for example, opens `$server`, sets `$isNewServerConnection`, then skips cleanup on every caller-loop iteration. Please run the owned-connection disconnects before the end-block interrupt return (and preserve the no-finished-message behavior), and add regression coverage that verifies a command-owned connection is closed on an affected guard path.

andreasjordan:
Fixed in 643bd5684 - the ownership disconnects now run before the interrupt return in the end block, with a comment stating the ordering rule; only the finished message stays suppressed on the interrupt path. The control flow the review describes was verified exactly as stated: guard stop in begin, process returns on the flag, end returned at its own guard before the cleanup.

Building the requested regression turned into an instructive dead end worth recording. Three attempts to observe the leak on the lab (Windows PowerShell 5.1):

1. **Server-side session counting on the default path** cannot show it: the command's own connections ride the shared pooled connection, cached in the connection hash, so repeated guard calls reuse one entry and the pooled session lives on whether or not the disconnect ran.
2. **Client-side `ConnectionContext.IsOpen`** cannot show it either: probed on the unfixed code, the entry reports `IsOpen = False` - SMO's auto-disconnect (`DisconnectIfPooled`) returns the physical connection after every batch, before the end block is ever reached.
3. **A `Pooling=False` connection string** (the shape where a skipped disconnect would leak a real session for the process lifetime) also stays clean on the unfixed code, because SMO's pooled/non-pooled decision reads its own `NonPooledConnection` flag, not the connection string - it auto-disconnects after each batch, and with pooling off that closes the session outright.

So on this stack the skipped cleanup was not observable through any input shape the command accepts - the defect is latent, protected against the day a connection is held open eagerly (an SMO behavior change, or `Connect-DbaInstance` handing out an eagerly-held connection). The new guard-path test pins the invariant - a marked `Pooling=False` guard call must leave zero sessions - and its comment states plainly that it also passes on the unfixed code, like the other latent sites in this workstream (#10648, #10652, and the WhatIf half of #10634). Full file 17/17 via the lab harness, warning-free, lab left clean.

potatoqualitee (approved):

PR #10642, opened 2026-08-29
New-DbaAvailabilityGroup - Let the validation guards reach their own return

andreasjordan:
## Summary

Fifth command fix from the #10638 inventory, and the variant that shows the bug's camouflage best: four validation guards (inaccessible `-SharedPath`, missing `-SharedPath` for manual seeding, and the two Linux checks) called `Stop-Function -Continue` **and had a `return` on the next line** - but no loop encloses them, so without `-EnableException` the `continue` escaped the command before its own `return` ever ran. The return statements were dead code on that path, making the sites look correct in review. The neighboring guards (certificate, ClusterType version floor) already use the correct form.

## Fix

Drop `-Continue` at the four sites; the existing `return` statements finally run. The regression test loops three times over an inaccessible `SharedPath` and asserts all iterations complete - zero completed on the unfixed command.

## Tests

Lab harness on SQL04\SQL2025 (HADR lane): 4 tests, 0 failed. Red on the unfixed command ("Expected 3, but got 0"), green on the fix.

References #10638

### DIFF
--- a/public/New-DbaAvailabilityGroup.ps1
+++ b/public/New-DbaAvailabilityGroup.ps1
@@ -534,14 +534,16 @@ function New-DbaAvailabilityGroup {
         if (($SharedPath)) {
             if (-not (Test-DbaPath -SqlInstance $Primary -SqlCredential $PrimarySqlCredential -Path $SharedPath)) {
                 Write-Progress -Activity "Adding new availability group" -Completed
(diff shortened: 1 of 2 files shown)


potatoqualitee (approved):
Reviewed the complete current head, all four validation paths, preserved exception behavior, regression coverage, and exact-head CI evidence. No material correctness or compatibility issues found.

PR #10643, opened 2026-08-29
Invoke-DbaPfRelog - Stop eating the caller loop when no blg files match

andreasjordan:
## Summary

Sixth command fix from the #10638 inventory - and the first where triage mattered as much as the fix. The sweep flagged ten sites in this file; **only one is a real escape**. The no-files guard in the end block called `Stop-Function -Continue` with a `return` on the next line, but no loop encloses it: without `-EnableException` the `continue` escaped the command before its own dead `return` ever ran and consumed an iteration of the caller's loop.

## The nine sites deliberately left alone

They sit inside the relog script block. On the sequential path, `foreach ($file in $allpaths) { Invoke-Command -ScriptBlock $scriptBlock ... }` invokes it, and PowerShell's flow control binds a `continue` from inside an invoked script block to the nearest loop frame **on the call stack** - that per-file `foreach`. Skipping to the next file is exactly their intent. On the `-Multithread` path each runspace contains its own flow control. A comment at the fixed site records the distinction so nobody "fixes" them later.

## Tests

The command had no integration tests at all. The new loop-counter test (a `-Path` matching no `.blg` files, three iterations, all must complete) is the first: red on the unfixed command ("Expected 3, but got 0"), green on the fix (2 tests, 0 failed via the lab harness).

References #10638

### DIFF
--- a/public/Invoke-DbaPfRelog.ps1
+++ b/public/Invoke-DbaPfRelog.ps1
@@ -281,7 +281,11 @@ function Invoke-DbaPfRelog {
         $allpaths = $allpaths | Where-Object { $_ -match '.blg' } | Select-Object -Unique
 
         if (-not $allpaths) {
-            Stop-Function -Message "Could not find matching .blg files" -Target $file -Continue
+            # No -Continue here: the end block has no enclosing loop, so the continue escaped the
+            # command before the return below ever ran and ate an iteration of the caller's loop.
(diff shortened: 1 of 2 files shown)


potatoqualitee (approved):
Reviewed the complete current head, the changed end-block guard, the nine intentionally retained per-file flow-control sites, regression coverage, and exact-head CI evidence. No material correctness or compatibility issues found.

PR #10645, opened 2026-08-29
Get-DbaRandomizedDataset - Stop eating the caller loop on template validation

andreasjordan:
## Summary

Eighth fix from the #10638 inventory. Both validation guards in the process block (no template supplied, template not found) called `Stop-Function -Continue` with no enclosing loop - the escape ate an iteration of the caller's loop. Stop-and-`return` now, which for pipeline input also correctly proceeds to the next bound item.

## Tests

Loop-counter regression test over a call without any template: red on the unfixed command ("Expected 3, but got 0"), green on the fix, via the lab harness.

References #10638


potatoqualitee (approved):
Reviewed the complete current head, both validation paths, pipeline behavior, regression coverage, and exact-head CI evidence. No material correctness or compatibility issues found.

PR #10647, opened 2026-08-29
New-DbaDbDataGeneratorConfig - Stop eating the caller loop on failures

andreasjordan:
## Summary

Eleventh fix from the #10638 inventory. Both guards called `Stop-Function -Continue` without an enclosing loop: the columntypes.json import guard in the begin block, and the write-failure guard at the tail of the process block. The escape ate an iteration of the caller's loop; both now stop and `return`.

## Tests

The write-failure path is pinned by a regression test that generates the config once, sets the target file read-only, and loops three times over the then-failing write: red on the unfixed command ("Expected 3, but got 0"), green on the fix (lab harness, environment clean). The import guard carries no test because it fires only when the module's own shipped `bin\datamasking\columntypes.json` is missing - untriggerable without corrupting the installation, and the fix is the same two-line shape verified nine times in this series.

References #10638

### DIFF
--- a/public/New-DbaDbDataGeneratorConfig.ps1
+++ b/public/New-DbaDbDataGeneratorConfig.ps1
@@ -122,7 +122,10 @@ function New-DbaDbDataGeneratorConfig {
         try {
             $columnTypes = Get-Content -Path "$script:PSModuleRoot\bin\datamasking\columntypes.json" | ConvertFrom-Json
         } catch {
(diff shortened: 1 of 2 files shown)


potatoqualitee (approved):
Reviewed the complete current head, both failure guards, regression coverage, and exact-head CI evidence. No material correctness or compatibility issues found.

PR #10648, opened 2026-08-29
Get-DbaRandomizedType - Stop eating the caller loop when the type list is missing

andreasjordan:
## Summary

Twelfth fix from the #10638 inventory, the smallest possible shape: the csv import guard in the begin block called `Stop-Function -Continue` with no enclosing loop, escaping into the caller's loop on failure. Stop-and-`return` now; the process block already guards with `Test-FunctionInterrupt`.

## Tests

No new regression test: the guard fires only when the module's own shipped `bin\randomizer\en.randomizertypes.csv` is missing, which cannot be triggered without corrupting the installation. The existing tests (which exercise the healthy import on every run) stay green via the lab harness.

References #10638


potatoqualitee (approved):
Reviewed the complete current head, begin/process interrupt semantics, existing coverage, and exact-head CI evidence. No material correctness or compatibility issues found.

PR #10649, opened 2026-08-29
Invoke-DbaDbDataGenerator - Stop eating the caller loop when Bogus fails to load

andreasjordan:
## Summary

Thirteenth fix from the #10638 inventory, same minimal shape as #10648: the Bogus faker-creation guard in the begin block called `Stop-Function -Continue` with no enclosing loop, escaping into the caller's loop on failure. Stop-and-`return` now; the process block already guards with `Test-FunctionInterrupt`.

## Tests

No new regression test: the guard fires only when the Bogus library shipped by dbatools.library cannot load - untriggerable without corrupting the installation. The 13 existing integration tests stay green via the lab harness, lab left clean.

References #10638

### DIFF
--- a/public/Invoke-DbaDbDataGenerator.ps1
+++ b/public/Invoke-DbaDbDataGenerator.ps1
@@ -138,7 +138,10 @@ function Invoke-DbaDbDataGenerator {
         try {
             $script:faker = New-Object Bogus.Faker($Locale)
         } catch {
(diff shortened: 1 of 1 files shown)


potatoqualitee (approved):
Reviewed the complete current head, begin/process control flow, existing coverage, and exact-head CI evidence. No material correctness or compatibility issues found.

PR #10650, opened 2026-08-29
Convert-DbaMaskingValue - Stop eating the caller loop on missing input

andreasjordan:
## Summary

Fourteenth fix from the #10638 inventory. Both validation guards in this private function's begin block called `Stop-Function -Continue` with no enclosing loop. `Invoke-DbaDbDataMasking` - the only production caller - passes `EnableException` at all five call sites, so there the guards throw before the `continue`; the escape only bit non-EnableException callers. Stop-and-`return` now.

## Tests

Loop-counter regression test on the warning path: red on the unfixed function ("Expected 3, but got 0"), green on the fix via the lab harness.

References #10638

### DIFF
--- a/private/functions/Convert-DbaMaskingValue.ps1
+++ b/private/functions/Convert-DbaMaskingValue.ps1
@@ -70,11 +70,16 @@ function Convert-DbaMaskingValue {
 
     begin {
         if (-not $Nullable -and -not $Value) {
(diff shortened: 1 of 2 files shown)


potatoqualitee (approved):
Reviewed the complete current head, all production call sites, regression coverage, and exact-head CI evidence. No material correctness or compatibility issues found.

PR #10651, opened 2026-08-29
Add-DbaReplArticle - Stop eating the caller loop on parameter validation

andreasjordan:
## Summary

Fifteenth fix from the #10638 inventory. The two guards at the top of the process block (CreationScriptOptions type check, WHERE-in-filter check) called `Stop-Function -Continue` - but they sit **before** the instance loop, so the escape ate an iteration of the caller's loop. Stop-and-`return` now. The filter guard's legacy single quotes became double quotes per repo style, and its message drops the quoted WHERE.

## Tests

The guards fire before any connection is made, so the new loop-counter regression test contacts no instance at all and runs in about a second: red on the unfixed command ("Expected 3, but got 0"), green on the fix.

References #10638


potatoqualitee (approved):
Reviewed the complete current head, validation control flow, regression coverage, and CI evidence. No material correctness or compatibility issues found.

PR #10652, opened 2026-08-29
New-DbaReplSubscription - Remove the latent continue escape and guard the process block

andreasjordan:
## Summary

Sixteenth entry from the #10638 inventory, and the first **latent** one: the two connection catches in the begin block carried `Stop-Function -Continue` without an enclosing loop, but the sites turn out to be unreachable. Without `EnableException`, the inner `Get-DbaReplServer`/`Get-DbaReplPublication` warn-and-return instead of throwing, so the catches never fire; with it, `Stop-Function` throws before the `continue`. Converted to stop-and-`return` anyway, so the defect cannot activate if the inner calls ever change their error behavior.

## The real find

The process block had **no `Test-FunctionInterrupt` guard**: had any begin-block stop ever fired, the subscriber loop would have run against a null publisher and publication. The standard guard is now in place.

## Tests

First integration test of this file: the unreachable-publisher path warns and the caller's loop completes, with `sql.connection.timeout` lowered during the context so three failing connection attempts cost ~9 seconds instead of ~90. Stated plainly: this test also passes on the unfixed code, because the escape sites were latent - there is no red-on-old for a defect that cannot activate.

References #10638

### DIFF
--- a/public/New-DbaReplSubscription.ps1
+++ b/public/New-DbaReplSubscription.ps1
@@ -129,17 +129,23 @@ function New-DbaReplSubscription {
         try {
             $pubReplServer = Get-DbaReplServer -SqlInstance $SqlInstance -SqlCredential $SqlCredential -EnableException:$EnableException
         } catch {
-            Stop-Function -Message "Failure" -Category ConnectionError -ErrorRecord $_ -Target $SqlInstance -Continue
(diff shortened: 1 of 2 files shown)


potatoqualitee (approved):
Reviewed the complete current head, helper semantics, test limitations, and CI evidence. No material correctness or compatibility issues found.

PR #10653, opened 2026-08-29
Set-DbaAgentOperator - Stop eating the caller loop on connection failure

andreasjordan:
## Summary

Seventeenth fix from the #10638 inventory. The connection catch called `Stop-Function -Continue`, but it sits **before** the operator loop - the escape ate an iteration of the caller's loop. Stop-and-`return` now. The catch is reachable (the inner `Get-DbaAgentOperator` runs with `-EnableException` and throws on connection failure), unlike the latent sites documented in #10652.

## Tests

Loop-counter regression test over an unreachable instance, with `sql.connection.timeout` lowered during the context to keep the three failing attempts fast: red on the unfixed command ("Expected 3, but got 0"), green on the fix via the lab harness, lab left clean.

References #10638


potatoqualitee (approved):
Reviewed the complete current head, regression coverage, CI evidence, and the control-flow semantics. No material correctness or compatibility issues found.

Issue #10548, opened 2026-08-10
Request for Comment: Centralize Write-Progress handling so progress bars cannot leak


andreasjordan:
## Problem

When a command ends, it sometimes leaves a progress bar behind. The stale bar keeps being displayed while later, unrelated commands run, and the user has no way to get rid of it short of restarting the session.

The root cause is that cleanup is currently the responsibility of every individual exit path. `Backup-DbaDatabase` is the clearest example: it contains about 20 separate `Write-Progress -Completed` calls, one per way out of the command. Miss one path — or hit a `throw` — and the bar leaks.

We already have `Write-ProgressHelper`, but it is not used consistently, and it does not help with cleanup at all.

## Current state

| | count |
|---|---|
| raw `Write-Progress` calls | 334 across 26 files |
| `Write-ProgressHelper` calls | 182 across 30 files |

Branch work in progress already completes the bars that 22 commands left behind on their happy path. That is a worthwhile fix on its own, but it is symptomatic — the next command that gains an early `return` will leak again.

### Why the helper is not used much

This is a capability problem, not a discipline problem. `private/functions/Write-ProgressHelper.ps1` has no `-Id` or `-ParentId` support, so every command that needs nested or multiple concurrent bars *has* to bypass it. That is precisely the set of 26 files with raw calls.

It also has a few defects worth noting, since any rewrite should not carry them forward:

- The `switch` on the caller name contains two dead duplicate branches (`Invoke-DbaDbLogShipRecovery`, `Sync-DbaAvailabilityGroup`).
- The default activity for `Export-DbaInstance` interpolates `$instance`, which is not defined in the helper — it is picked up from the caller's scope by dynamic scoping.
- `TotalSteps` is derived by regex-counting occurrences of the string `Write-ProgressHelper` in the caller's entire `.Definition`, **on every call**. That is expensive on large functions, and it is simply wrong whenever the calls sit inside loops or conditional branches.

## The constraint that shapes any solution

Progress records are keyed by **(SourceId, ActivityId)**, where `SourceId` defaults to the top-level pipeline. The host removes a record only when a matching `-Completed` arrives carrying the *same* SourceId.

Two consequences follow:

1. A later command **cannot** clear an earlier command's stale bar. Different pipeline, different SourceId. A global janitor, or a "sweep leftovers when the next dbatools command starts" approach, cannot work across command boundaries.
[...]

PR #10660, opened 2026-08-30
Progress bars - Complete the bars that 22 commands left behind

andreasjordan:
Opened as a draft on 2026-08-30 as the tactical half of the progress-bar topic, parked while the structural RFC (#10548) waited for maintainer opinions. That discussion has not started, so this first step is put up for review on its own: it is self-contained and does not prejudge the RFC.

## What this fixes

An AST scan over the module found 22 commands that start a `Write-Progress` bar and **never complete it at all** - on any path, success included. Their stale bars stay on screen until some later command happens to reuse the same progress Id, or the pipeline ends.

This PR adds the missing `Write-Progress -Completed` calls: 25 completion calls across 22 commands, 30 insertions, no logic changes.

## What this deliberately does not fix

The second class found by the same scan: commands that complete their bar on the success path but leak it when they throw under `-EnableException` (confirmed example: `Add-DbaAgDatabase`, whose completion sits after the `foreach` and whose 28 `Stop-Function` calls all throw past it). That class needs a structural answer - completing bars from inside `Stop-Function`, as proposed in #10548 - rather than 28 hand-placed completions, and is left out here on purpose.

Scan notes for reviewers: a bar is identified by `-Id` (default 0), not by its Activity text, so Activity mismatches are not leaks; and `Invoke-Parallel.ps1` is a permanent false positive for grep-based scans (the splat carries the Id, completed later).

## Testing

All 22 affected test files were run individually against the lab at the time the branch was built (2026-08-10): 171 passed, 0 failed, 7 pre-existing skips. `Write-Progress` writes zero bytes to stdout/stderr when redirected, so these calls cannot affect captured test output; CI additionally suppresses progress rendering via `$ProgressPreference` in its runner scripts.

The branch was based on development as of 2026-08-10. Development was merged in on 2026-09-12: the only conflict was `Sync-DbaLoginPermission.ps1`, where #10603 wrapped the login loop in `try`/`finally` for the database-context restore. The completion call sits where it did, after the inner loop, now inside that `try` before the `finally`. The merged branch differs from development by exactly the 30 insertions and 4 deletions of the original commit; `Sync-DbaLoginPermission` re-run through the lab harness on both editions after the merge, green.

References #10548.

### DIFF
--- a/public/Disable-DbaDbEncryption.ps1
+++ b/public/Disable-DbaDbEncryption.ps1
@@ -126,6 +126,7 @@ function Disable-DbaDbEncryption {
                         Write-Message -Level Verbose -Message "Database state for $($db.Name) on $($server.Name): $($db.DatabaseEncryptionKey.EncryptionState)"
                     }
                     while ($db.DatabaseEncryptionKey.EncryptionState -notin "Unencrypted", "None")
+                    Write-ProgressHelper -Completed
 
                     if (-not $NoEncryptionKeyDrop) {
                         # https://www.sqlservercentral.com/steps/stairway-to-tde-removing-tde-from-a-database
--- a/public/Export-DbaUser.ps1
+++ b/public/Export-DbaUser.ps1
@@ -645,6 +645,7 @@ function Export-DbaUser {
                     $sql
                 }
             }
+            Write-ProgressHelper -Completed
         }
         # Just a single file, output path once here
         if (-Not $GenerateFilePerUser -and $FilePath) {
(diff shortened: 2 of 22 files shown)


potatoqualitee (approved):
The completion calls use the correct progress IDs across all changed commands, and removing the unsupported helper arguments is safe. The changed scripts parse successfully; unrelated MinIO setup failures do not affect this review.

potatoqualitee:
thank you 🙏🏼

PR #10664, opened 2026-08-31
Policy Based Management - Run on PowerShell 7 on Windows

### LINKED ISSUE #10662: PBM cmdlets (Get-DbaPbmStore, Get-DbaPbmPolicy, etc.) abort with "not supported on Linux or macOS" on Windows under PowerShell 7
**Environment:** dbatools 2.8.2, Windows 11 / Windows Server, PowerShell 7 (pwsh.exe)

**Description:**
`Get-DbaPbmStore` (and everything built on it - all `Get-DbaPbm*` commands) plus `Copy-DbaSsisCatalog`, `Export-DbaCredential`, `Export-DbaLinkedServer`, `Get-DbaSsisEnvironmentVariable`, `New-DbaSsisCatalog` abort with *"This command is not supported on Linux or macOS"* when run from PowerShell 7 **even though the host OS is Windows**. The check appears to key off `$PSEdition`/`$IsCoreCLR` rather than the actual OS, since the real constraint (SMO PolicyStore / .NET Framework dependency) is Windows-vs-non-Windows, not PS5.1-vs-PS7.

**Repro:**
- On a Windows machine, open PowerShell 7 (`pwsh`).
- Run `Get-DbaPbmStore -SqlInstance <instance>`.
- Result: immediate abort with "This command is not supported on Linux or macOS", despite running on Windows.
- Same script run from Windows PowerShell 5.1 on the identical machine/instance works fine.

**Impact:**
Combined with `-ErrorAction SilentlyContinue` in calling code, this becomes a **silent** failure - no exception surfaces, the caller just gets an empty result and can't distinguish "policy doesn't exist" from "cmdlet refused to run here." In our case this meant a PBM "On Change: Prevent" policy was never detected/disabled before a `CREATE LOGIN`, which then failed against the enforcing DDL trigger - identical script, identical instance, passed under PS 5.1 and failed under PS 7.

**Expected:**
Gate on actual OS (`$IsWindows`) rather than PS edition, so these commands work on Windows PowerShell 7 the same as on Windows PowerShell 5.1.

**Workaround used:**
Switched to plain T-SQL (`msdb.dbo.syspolicy_policies`, `sp_syspolicy_update_policy`) for anything that must work under both PS versions.

andreasjordan:
Fixes #10662

## Problem

`Get-DbaPbmStore`, `Get-DbaPbmCondition` - and with them every `Get-DbaPbm*` command - refuse to run on PowerShell 7 with "This command is not supported on Linux or macOS", even on Windows. `Export-DbaInstance` silently skips the policy export there. Combined with `-ErrorAction SilentlyContinue` in calling code this becomes a silent failure: an empty result that cannot be told apart from "no policy exists", which is how the reporter's "On Change: Prevent" policy went undetected before a `CREATE LOGIN`.

## Mechanism

The three gates tested the *edition*, `$PSVersionTable.PSEdition -eq "Core"`, while the message and the real constraint are about the *platform*. dbatools.library ships `Microsoft.SqlServer.Dmf.dll` and `Microsoft.SqlServer.Dmf.Common.dll` under `core/lib` as well as `desktop/lib`, and `Add-PbmLibrary` loads them from whichever edition folder is active. A `PolicyStore` built from the core assembly under pwsh 7.6 on Windows enumerates policies, conditions, categories and object sets exactly as on 5.1 (16 policies on the lab instance).

## What changed

- `Get-DbaPbmStore`, `Get-DbaPbmCondition`: the gate is now `$IsLinux -or $IsMacOS`, which is what `Export-DbaCredential` and `Export-DbaLinkedServer` already use. `Get-DbaPbmCondition` also gets the same message as the others instead of "not yet supported in PowerShell Core".
- `Export-DbaInstance`: the policy export runs unless on Linux or macOS; the verbose skip message says so.
- Tests: `Get-DbaPbmCategory`, `Get-DbaPbmCondition`, `Get-DbaPbmPolicy` and the "Export policies" test of `Export-DbaInstance` drop their `-Skip:($PSVersionTable.PSVersion.Major -gt 5)`. `Get-DbaPbmStore` gets its first integration test, deliberately without a pwsh skip - that is the regression test for this issue.

  A correction to the commit message, which says the pwsh lane in CI now covers this: CI runs the Pester suite under Windows PowerShell 5.1 only (`tests/appveyor.pester.ps1` starts `powershell.exe`, the cross-platform `windows-tests` job is `if: false`, and the only pwsh-on-Windows test is `Connect-DbaInstance.WindowsSspi.Tests.ps1`). So in CI these tests keep running on 5.1 as before; the pwsh coverage below comes from the lab. Removing the skips takes away the artificial exclusion so that the tests are ready the day a pwsh lane exists - the portable-pwsh step that runs the SSPI test would be the natural place, but that is a workflow decision, not part of this PR.

- Linux and macOS keep refusing exactly as before. Whether the Dmf assembly works there is untested and out of scope.
- The SSIS commands the issue also lists (`Copy-DbaSsisCatalog`, `New-DbaSsisCatalog`, `Get-DbaSsisEnvironmentVariable`) are a genuine limitation - the Integration Services object model is only shipped under `desktop/lib` - so their gate stays; only their message is wrong, and that gets its own small PR.
- `Export-DbaCredential` and `Export-DbaLinkedServer`, also listed in the issue, already gated on the platform in v2.8.2 and are untouched.

## Tests

Lab run through the testing-dbatools harness against SQL Server 2019 (`InstanceSingle`):

- `Get-DbaPbmStore`, `Get-DbaPbmCategory`, `Get-DbaPbmCondition`, `Get-DbaPbmPolicy` integration tests: green under pwsh 7.6.3 **and** Windows PowerShell 5.1.
- `Export-DbaInstance` integration tests under pwsh: 26 passed, 0 skipped - the policy export now runs there.
- Red on old: with the command change stashed, the new `Get-DbaPbmStore` test fails under pwsh with `Expected $null or empty, but got [Get-DbaPbmStore] This command is not supported on Linux or macOS` - the reporter's exact symptom.

Thanks to @JankeUwe for the report and the correct diagnosis of edition-versus-platform.

### DIFF
--- a/public/Export-DbaInstance.ps1
+++ b/public/Export-DbaInstance.ps1
@@ -446,10 +446,10 @@ function Export-DbaInstance {
                         Get-ChildItem -ErrorAction Ignore -Path "$exportPath\endpoints.sql"
                     }
 
-                    if ($Exclude -notcontains 'PolicyManagement' -and ($IsLinux -or $IsMacOS)) {
+                    if ($Exclude -notcontains "PolicyManagement" -and ($IsLinux -or $IsMacOS)) {
                         Write-Message -Level Verbose -Message "Skipping Policy Management -- not supported on Linux or macOS"
                     }
-                    if ($Exclude -notcontains 'PolicyManagement' -and -not ($IsLinux -or $IsMacOS)) {
+                    if ($Exclude -notcontains "PolicyManagement" -and -not ($IsLinux -or $IsMacOS)) {
                         Write-Message -Level Verbose -Message "Exporting Policy Management"
                         Write-ProgressHelper -StepNumber ($stepCounter++) -Message "Exporting Policy Management"
 
--- a/public/Export-DbaInstance.ps1
+++ b/public/Export-DbaInstance.ps1
@@ -446,10 +446,10 @@ function Export-DbaInstance {
                         Get-ChildItem -ErrorAction Ignore -Path "$exportPath\endpoints.sql"
                     }
 
-                    if ($Exclude -notcontains 'PolicyManagement' -and $PSVersionTable.PSEdition -eq "Core") {
-                        Write-Message -Level Verbose -Message "Skipping Policy Management -- not supported by PowerShell Core"
+                    if ($Exclude -notcontains 'PolicyManagement' -and ($IsLinux -or $IsMacOS)) {
+                        Write-Message -Level Verbose -Message "Skipping Policy Management -- not supported on Linux or macOS"
                     }
-                    if ($Exclude -notcontains 'PolicyManagement' -and $PSVersionTable.PSEdition -ne "Core") {
+                    if ($Exclude -notcontains 'PolicyManagement' -and -not ($IsLinux -or $IsMacOS)) {
                         Write-Message -Level Verbose -Message "Exporting Policy Management"
                         Write-ProgressHelper -StepNumber ($stepCounter++) -Message "Exporting Policy Management"
 
--- a/public/Get-DbaPbmCondition.ps1
+++ b/public/Get-DbaPbmCondition.ps1
@@ -101,8 +101,8 @@ function Get-DbaPbmCondition {
     }
     process {
         if (Test-FunctionInterrupt) { return }
-        if ($PSVersionTable.PSEdition -eq "Core") {
-            Stop-Function -Message "This command is not yet supported in PowerShell Core"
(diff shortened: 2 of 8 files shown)


potatoqualitee (approved):
No material defects found.

potatoqualitee:
thank you kindly 🙇🏼

PR #10665, opened 2026-08-31
SSIS commands - Say why they refuse to run on PowerShell 7

andreasjordan:
Refs #10662

## Problem

`Copy-DbaSsisCatalog`, `New-DbaSsisCatalog` and `Get-DbaSsisEnvironmentVariable` refuse to run on PowerShell 7 with "This command is not supported on Linux or macOS" - also on Windows, which sends the user looking for a platform problem that does not exist.

## Mechanism

Unlike the PBM commands (fixed separately), the refusal itself is correct: dbatools.library ships `Microsoft.SqlServer.Management.IntegrationServices.dll` only under `desktop/lib`, so the SSIS object model does not exist on pwsh at all. Only the message is wrong.

## What changed

The three messages now read "PowerShell Core is not supported because the SQL Server Integration Services object model is only shipped for Windows PowerShell, please use Windows PowerShell." - the wording `Install-DbaSqlWatch` and `Uninstall-DbaSqlWatch` already use for the same kind of limitation. The gate (`$PSVersionTable.PSEdition -eq "Core"`) is unchanged.

Each of the three test files gets a unit test that runs only on pwsh and asserts the warning names Windows PowerShell and does not mention Linux. It uses a placeholder instance name, because the gate fires before any connection is attempted.

Nothing is made to work on pwsh here; that would need the assembly in the core library first.

## Tests

- The three unit test files: green under pwsh 7.6.3 (new test runs) and Windows PowerShell 5.1 (new test skipped), through the testing-dbatools harness.
- Red on old: with the command change stashed, `New-DbaSsisCatalog.Tests.ps1` fails with `Expected regular expression 'Windows PowerShell' to match [New-DbaSsisCatalog] This command is not supported on Linux or macOS`.

Thanks to @JankeUwe for listing these commands in #10662.

### DIFF
--- a/public/Copy-DbaSsisCatalog.ps1
+++ b/public/Copy-DbaSsisCatalog.ps1
@@ -127,7 +127,7 @@ function Copy-DbaSsisCatalog {
     <# Developer note: The throw calls must stay in this command #>
     begin {
         if ($PSVersionTable.PSEdition -eq "Core") {
-            Stop-Function -Message "This command is not supported on Linux or macOS"
+            Stop-Function -Message "PowerShell Core is not supported because the SQL Server Integration Services object model is only shipped for Windows PowerShell, please use Windows PowerShell."
             return
         }
         $ISNamespace = "Microsoft.SqlServer.Management.IntegrationServices"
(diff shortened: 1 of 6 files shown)


potatoqualitee (approved):
No material defects found.

PR #10666, opened 2026-08-31
Copy-DbaDbTableData - Map query columns onto the writable destination columns

### LINKED ISSUE #10661: Copy-DbaDbTableData -Query silently miscounts computed columns without -ForceExplicitMapping, corrupting data
**Environment:** dbatools 2.7.1 and 2.8.2 (reproduced on both), SQL Server

**Description:**
When the destination table has a computed column (e.g. at ordinal position 3) and `Copy-DbaDbTableData` is used in `-Query` mode without `-ForceExplicitMapping`, the cmdlet falls back to implicit ordinal column mapping that appears to count the computed column as a real target slot. Every column after it silently shifts by one position. No error or warning is produced - values land in the wrong columns.

**Repro:**
- Destination table with 108 columns, one computed column at ordinal position 3.
- Source query returns matching non-computed columns in order.
- Run `Copy-DbaDbTableData -SqlInstance <src> -Database <db> -Query "SELECT ..." -DestinationSqlInstance <dst> -DestinationDatabase <db> -DestinationTable <table>` without `-ForceExplicitMapping`.
- After copy, two locale-ID columns (`GV_Source_ID`/`GV_Produkt` in our schema) land swapped/wrong (0 vs 1033) - confirmed by direct row inspection, not a formatting artifact.

**Expected:**
Either exclude computed columns from ordinal counting by default, or clearly document that `-ForceExplicitMapping` is required whenever the destination has any computed column - ideally with a loud warning when it's omitted and a computed column is detected on the target.

**Workaround used:**
Built a custom `SqlBulkCopy` wrapper with explicit name-based `ColumnMappings`, bypassing `-Query` mode's implicit mapping entirely.

andreasjordan:
Fixes #10661

## Problem

With `-Query` and without `-ForceExplicitMapping`, a destination table with a computed or a rowversion column between its writable columns makes the copy go wrong: either the server rejects the insert, or - worse - every column behind the unwritable one shifts by one position, the last column stays empty, no warning is written and the command reports success.

## Mechanism

In that mode the command added no column mappings at all and left `SqlBulkCopy` to map by position ([Copy-DbaDbTableData.ps1:710](https://github.com/dataplat/dbatools/blob/development/public/Copy-DbaDbTableData.ps1#L710)). `SqlBulkCopy`'s own list contains *every* destination column:

- a **computed** column stays in the list, so the server rejects the `INSERT BULK` with error 271 ("cannot be modified because it is either a computed column or is the result of a UNION operator") - a loud failure;
- a **rowversion** column is dropped from the list *together with the source column that lands on it*, and everything behind it moves one position to the left - the silent data corruption the issue describes.

Both reproduced in the lab (Microsoft.Data.SqlClient 6.1.5, SQL Server 2022 to 2025). The report names a computed column; on our stack that shape fails loudly and the silent shift needs a rowversion column. The reporter has since verified that their table has no rowversion column at all, so with their driver/server combination the computed column itself is dropped silently - the loud-vs-silent behaviour is version dependent. The fix covers both, since the command no longer relies on the implicit mapping in either shape.

## What changed

- The destination column list the command already kept (non-computed columns) now also excludes rowversion columns (`DataType.SqlDataType -eq "Timestamp"`), i.e. it is the list of columns that can actually be written.
- In `-Query` mode without `-ForceExplicitMapping` the command now builds the positional mapping itself: query column *n* is mapped to the *n*-th writable destination column via `ColumnMappings.Add([int], [string])`. This is what `SqlBulkCopy` did on its own, minus the unwritable columns.
- A query that returns more columns than the destination has writable columns is refused with a message naming both counts, inside the existing `try`/`catch` so it honours `-EnableException` and `-Continue`s to the next destination like every other failure there. Before, the surplus was either rejected by SqlBulkCopy with a generic error or silently lost.
- Help for `-Query` and `-ForceExplicitMapping` describes the new rule; the old text promised that ordinal mapping "will cause it to fail" on computed columns.

Identity placeholders keep working as documented: a mapping onto an identity column without `-KeepIdentity` is still ignored by the server, and with `-KeepIdentity` the value is written - verified with raw `SqlBulkCopy` before the change and by a test after it.

Table mode and `-ForceExplicitMapping` keep their name-based mapping. The only effect the new column list has there is that a source column named like a destination rowversion column is no longer mapped - `SqlBulkCopy` tolerated that mapping anyway, so the result is identical.

## Tests

New `Context` in `Copy-DbaDbTableData.Tests.ps1` (COPY lane), asserting the data, not the row count:

1. computed + rowversion between the writable columns: values land in the right columns, the computed column computes;
2. rowversion only: the silent variant, values behind it are not shifted;
3. identity placeholder without `-KeepIdentity`: identity generated, other columns right;
4. a query with one column too many is refused with the counts in the warning and writes nothing.

Lab through the testing-dbatools harness (SQL 2022 -> SQL 2025): 18/18 green with the fix. Red on old with the command change stashed: 1 and 4 fail with the server's computed-column error, 2 fails with `Expected @(22, 222) But was @(33, 333)` - the reported symptom; 3 passes on both, it pins the behaviour that must not change.

## Added while the PR was open

The two existing Query-mode tests ("relies on the default source database", "3 part query") selected `Id` from a table whose column is `id`. Fine on CI, which is case insensitive, but on the case sensitive SQL05 instances of the lab the server answers `Invalid column name 'Id'` and both tests fail with `Expected 1, but got $null`. Fixed in the third commit by using the column name as created; verified with the query on the CS instance and by re-running the whole file against the case sensitive COPY pair (`SQL05SQL2022` -> `SQL05SQL2025`): 18/18 green, where the same file failed both Query-mode tests in the morning run.

The review pointed out that the writable-column filter still counted GENERATED ALWAYS columns - the period columns of a temporal table and the ledger metadata columns, which SMO reports with `Computed = $false`. The fourth commit excludes columns whose `GeneratedAlwaysType` is not `None`, guarded with `$destServer.VersionMajor -ge 13` because SMO only supports the property on SQL Server 2016 and later, and adds a regression test with a system-versioned temporal destination whose period columns are interleaved between the writable ones. Green on both editions in the lab, red without the command change with exactly the server's 13536 error. Test file after the addition: 19/19 on both editions.

The second review round asked for Azure SQL Database, whose product version says 12 while it supports temporal tables. In the lab SMO's `VersionMajor` on Azure SQL Database is 18, because SMO derives it from `@@MICROSOFTVERSION` and Azure reports the real engine build there - so the version guard already passed and the copy worked, by accident of how the number is derived rather than by design. The fifth commit makes the guard say what SMO itself checks (`VersionMajor -ge 13 -or DatabaseEngineType -eq "SqlAzureDatabase"`), applies the same rule in the temporal test, and adds a context that copies the interleaved layout into a temporal table on a real Azure SQL Database, skipped where the configuration has none (every CI environment) and asserting the engine edition first. Green on both editions in the lab, Azure test included: 20 run, 0 failed.

Thanks to @JankeUwe for the report and the row-level inspection that made the shift undeniable.

### DIFF
--- a/public/Copy-DbaDbTableData.ps1
+++ b/public/Copy-DbaDbTableData.ps1
@@ -718,7 +718,7 @@ function Copy-DbaDbTableData {
                                 if ($destColumns -contains $sourceColumn) {
                                     $null = $bulkCopy.ColumnMappings.Add($sourceColumn, $sourceColumn)
                                 } else {
-                                    Write-Message -Level Verbose -Message "Skipping column '$sourceColumn' (not in destination or not writable)"
+                                    Write-Message -Level Verbose -Message "Skipping column $sourceColumn (not in destination or not writable)"
                                 }
                             }
                         } else {
--- a/public/Copy-DbaDbTableData.ps1
+++ b/public/Copy-DbaDbTableData.ps1
@@ -56,13 +56,14 @@ function Copy-DbaDbTableData {
         Custom SQL SELECT query to use as the data source instead of copying the entire table or view. Supports 3 or 4-part object names.
         Use this when you need to filter rows, join multiple tables, or transform data during the copy operation. Still requires specifying a Table or View parameter for metadata purposes.
 
-        Note: Columns are mapped by ordinal position. If the destination table has an identity column, include a placeholder value (e.g., 0) in your SELECT list at that position.
+        Note: Columns are mapped by position onto the writable columns of the destination table, so computed and rowversion columns of the destination do not count and must not have a counterpart in the SELECT list.
+        If the destination table has an identity column, include a placeholder value (e.g., 0) in your SELECT list at that position.
         The placeholder will be ignored and the identity value auto-generated unless -KeepIdentity is specified.
 
     .PARAMETER ForceExplicitMapping
         When used together with Query parameter, force the use of explicit column mapping (name-based) instead of switching over to ordinal position mapping. Use with care if query contains aliases.
         Default behaviour when using Query parameter is to use ordinal position mapping, due to the possibility of the query including aliases (SELECT x AS y) which could lead to column mismatching and data not copying.
-        The downside of it automatically switching over to ordinal mapping is that it also tries to copy over computed columns, which will cause it to fail.
+        Positional mapping skips the computed and rowversion columns of the destination table, so the SELECT list only has to match its writable columns.
 
     .PARAMETER AutoCreateTable
         Automatically creates the destination table if it doesn't exist, using the same structure as the source table.
@@ -672,11 +673,14 @@ function Copy-DbaDbTableData {
                         $bulkCopy.NotifyAfter = $NotifyAfter
                         $bulkCopy.BulkCopyTimeout = $BulkCopyTimeout
 
-                        # Get list of non-computed columns from destination table to avoid insert failures
+                        # Get list of writable columns from destination table to avoid insert failures. Computed and rowversion
+                        # columns cannot be written, and they are also what breaks the implicit positional mapping of SqlBulkCopy:
+                        # it counts a computed column (the server then rejects the insert) and silently drops the source column
+                        # that lands on a rowversion column, shifting every column behind it by one (see #10661).
                         # Refresh the columns collection to ensure it's populated
                         $desttable.Columns.Refresh()
-                        $destColumns = $desttable.Columns | Where-Object Computed -eq $false | Select-Object -ExpandProperty Name
-                        Write-Message -Level Verbose -Message "Destination table has $($destColumns.Count) non-computed columns"
+                        $destColumns = @($desttable.Columns | Where-Object { -not $PSItem.Computed -and $PSItem.DataType.SqlDataType -ne "Timestamp" } | Select-Object -ExpandProperty Name)
+                        Write-Message -Level Verbose -Message "Destination table has $($destColumns.Count) writable columns"
 
                         # The legacy bulk copy library uses a 4 byte integer to track the RowsCopied, so the only option is to use
                         # integer wrap so that copy operations of row counts greater than [int32]::MaxValue will report accurate numbers.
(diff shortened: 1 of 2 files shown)


potatoqualitee (changes requested):
The positional destination mapping still treats temporal and ledger GENERATED ALWAYS columns as writable. For example, with destination columns A, ValidFrom, B, ValidTo and a query returning A, B, source B maps to ValidFrom and SqlBulkCopy fails because GENERATED ALWAYS period columns cannot accept explicit values. This materially breaks copying query results into supported temporal destinations. Exclude columns whose version-supported GeneratedAlwaysType is not None, guard the metadata lookup for older SQL Server versions, and add an interleaved temporal-column regression test.

andreasjordan:
Ich can try to have a look at this tomorrow.

andreasjordan:
Confirmed and fixed in 863aaa76d. SMO reports the period columns of a temporal table with `Computed = $false`, so the writable-column filter did include them, the positional mapping put a query column onto `ValidFrom`, and the server rejected the insert with error 13536 ("Cannot insert an explicit value into a GENERATED ALWAYS column") - reproduced in the lab exactly as described, interleaved layout and all.

The filter now also excludes columns whose `GeneratedAlwaysType` is not `None`, which covers the temporal period columns and the ledger metadata columns in one condition. The lookup is guarded with `$destServer.VersionMajor -ge 13`, because SMO only supports the property on SQL Server 2016 and later - same pattern `New-DbaDbMaskingConfig` already uses. The name-based mapping of table mode profits as well, since it works from the same list.

New regression test: a system-versioned temporal destination with the period columns interleaved between the writable ones (`Id, A, ValidFrom, B, ValidTo, C`), asserting the values land in the right columns and the server fills the period. It skips below SQL Server 2016 via a discovery-time version check. Green on both editions through the testing-dbatools harness (SQL 2022 -> SQL 2025), and red without the command change, failing with exactly the 13536 error.

potatoqualitee (changes requested):
`public/Copy-DbaDbTableData.ps1:684` still treats `VersionMajor >= 13` as the capability check for `GeneratedAlwaysType`. Azure SQL Database reports major version 12 while supporting temporal tables, so a temporal destination there keeps `ValidFrom`/`ValidTo` in `destColumns`. A query such as `SELECT Id, A, B, C` is then positionally mapped onto `(Id, A, ValidFrom, B, ValidTo, C)`, causing `B` to target the system-generated `ValidFrom` column and `WriteToServer` to fail. With `-Truncate`, the destination is emptied before that deterministic failure.

Please include Azure SQL Database in the generated-always capability check (or probe the property safely) and use the same capability logic in the temporal regression test; the current `VersionMajor >= 13` skip masks this supported path.

andreasjordan:
Confirmed the Azure point and fixed in 5487c4576.

What the lab shows on an Azure SQL Database first, because it is not quite the picture in the review: SMO's `VersionMajor` is not the product version. SMO derives it from `@@MICROSOFTVERSION`, and Azure SQL Database reports the real engine build there - 18 today - while `Version`, `VersionString` and `SERVERPROPERTY('ProductVersion')` all say 12.0. So `VersionMajor -ge 13` was already true on Azure, and the copy into an interleaved temporal table worked with the previous commit. That is an accident of how the number is derived, though, not a capability check. SMO itself gates `GeneratedAlwaysType` on "SQL Server 2016 or later, or Azure SQL Database", and the guard now says the same: `$destServer.VersionMajor -ge 13 -or $destServer.DatabaseEngineType -eq "SqlAzureDatabase"`. The temporal regression test uses that rule for its discovery-time skip and its table creation.

New context in the test file that copies the interleaved layout into a system-versioned temporal table on a real Azure SQL Database, following the pattern of the Azure tests in `Connect-DbaInstance.Tests.ps1`: skipped where `$TestConfig.AzureSqlDbServer` is empty, which is every CI environment, and it asserts the engine edition and the 12.0 product version first so it cannot pass against anything that is not this case. Ran for real in the lab against the serverless Azure SQL Database on both editions: every integration test of the file green (20 run, 0 failed), values in the right columns and the period filled by the server.

andreasjordan:
The red `external-table-tests` check is not this PR: the `minio/minio` image is gone from Docker Hub and the `mc` download URL answers 410 Gone, so the MinIO setup step fails on every branch since 2026-09-11, development included. Fix in #10688; once that is in, an update of this branch will turn the check green.

potatoqualitee (approved):
The current head resolves both previously blocking paths: generated-always columns are excluded from writable mappings, including Azure SQL Database, and the regression coverage exercises interleaved temporal columns. The relevant COPY lane passes; the MinIO failures are unrelated infrastructure failures addressed by #10688.

potatoqualitee:
ty!

PR #10668, opened 2026-09-01
Get-DbaComputerCertificate - Keep the friendly name of certificates on remote computers

andreasjordan:
## Problem

For a remote computer, `Get-DbaComputerCertificate` returns certificates whose `FriendlyName` is empty, although the certificate in the remote store has one and the command's own `Name` property carries it. `New-DbaComputerCertificate` returns this object for remote computers, so its documented `FriendlyName` was empty there too, while local results had it.

## Mechanism

PowerShell remoting rebuilds the `X509Certificate2` on the calling side from its raw bytes. The friendly name is a property of the store entry, not of the certificate, so it does not survive the trip - and a remote note property of the same name does not survive either (probed), which is why the scriptblock already exposes it as `Name`.

## What changed

The command copies `Name` back into `FriendlyName` on the calling side when it arrives empty. Local and remote results now carry the same properties; nothing changes for localhost. The `.OUTPUTS` note on `Name` says so.

## Tests

- New test: pins that both `Name` and `FriendlyName` carry the friendly name. It runs locally in CI, where no second computer exists.
- Lab: `Get-DbaComputerCertificate` and `New-DbaComputerCertificate` against a second computer on PowerShell 7.6 and 5.1 - both now return the friendly name for the remote host, unchanged for localhost. `Get-DbaComputerCertificate.Tests.ps1` green through the testing-dbatools harness.

### DIFF
--- a/public/Get-DbaComputerCertificate.ps1
+++ b/public/Get-DbaComputerCertificate.ps1
@@ -64,7 +64,7 @@ function Get-DbaComputerCertificate {
         - ComputerName: The name of the computer where the certificate is stored
         - Store: The certificate store location (CurrentUser or LocalMachine)
         - Folder: The certificate folder/container name (My, Root, AddressBook, etc.)
-        - Name: The friendly name of the certificate (added via Add-Member)
+        - Name: The friendly name of the certificate (added via Add-Member; FriendlyName carries the same value, also for remote computers)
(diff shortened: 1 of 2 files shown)


potatoqualitee (approved):
No material defects found.

PR #10670, opened 2026-09-01
Tests - Case sensitive lookups and backup cleanup in seven test files

andreasjordan:
Test-only. Found by a full suite run against three case sensitive instances (`SQL_Latin1_General_CP1_CS_AS`) plus two leftovers the same run reported.

## Case sensitivity

- **Copy-DbaDbViewData.Tests.ps1**: two query-mode tests select `Id` from a column called `id` - "Invalid column name 'Id'" on a case sensitive instance (the same defect in `Copy-DbaDbTableData.Tests.ps1` is part of #10666).
- **New-DbaDatabase.Tests.ps1**: the tests create the log file with `-LogFileSuffix "_Log"` and then look it up as `LogFiles["<name>_log"]`. SMO collections follow the server collation, so on a case sensitive instance the lookup returns `$null` and the size and growth assertions fail. The lookups with the default suffix (`randomDb`) are untouched, they are `_log` for real.

- **Sync-DbaLoginPermission.Tests.ps1**: the setup batch says `USE Master`, which a case sensitive instance rejects, so the whole batch is dropped and the login never exists on the source - the sync then finds nothing and warns. The test also passed `-WarningVariable $warn` (an empty variable name), which hid exactly that warning. Now `master`, `warn`, and the setup runs with EnableException so a failing setup fails the file.

## Leftovers

- **Enable-DbaForceNetworkEncryption.Tests.ps1** enabled force encryption on `InstanceSingle` and never put it back; the registry value survives everything except an explicit Disable, and four lab instances had drifted to `True` that way. Both the Enable and the Disable test now read the state in `BeforeAll` and restore it in `AfterAll`.
- **Connect-DbaInstance.Tests.ps1** ("clones when using Backup-DabInstace") took two msdb backups with the same default name (same minute) and removed each with a silenced `Remove-Item`. One removal failed silently once and the file was reported as a leftover by the 18 following test files. The backups now get unique names, and the removal retries a few times and then asserts that the file is gone.
- **Measure-DbaBackupThroughput.Tests.ps1** removed its backup folder with a single `Remove-Item -Recurse` seconds after the backup, which once left the whole folder behind (55 files flagged it). Same treatment: retry, and let the last attempt report its error.

Neither leftover reproduced when the file ran alone, so both are timing races on the share rather than deterministic bugs; the retries make the cleanup robust either way.

## Tests

All seven files green through the testing-dbatools harness against the case sensitive lab (SQL Server 2019/2022/2025), with the Enable/Disable pair verified to leave `ForceEncryption` as it found it.

### DIFF
--- a/tests/Connect-DbaInstance.Tests.ps1
+++ b/tests/Connect-DbaInstance.Tests.ps1
@@ -929,16 +929,33 @@ Describe $CommandName -Tag IntegrationTests {
             # runs against a path that does not exist on the machine running the tests and silently does
             # nothing. The backup was then left behind on every remote instance, and because this file
             # runs early in the suite, every later test file reported it as a leftover of its own.
-            $results = Backup-DbaDatabase -SqlInstance $server -Database msdb -Path $TestConfig.Temp
+            # Each backup gets its own file name, because two backups within the same minute share the
+            # default name and the second one appends to the first.
+            $backupFiles = @()
+            $results = Backup-DbaDatabase -SqlInstance $server -Database msdb -Path $TestConfig.Temp -FilePath "msdb_$(Get-Random).bak"
             if ($results.FullName) {
-                Remove-Item -Path $results.FullName -ErrorAction SilentlyContinue
+                $backupFiles += $results.FullName
             }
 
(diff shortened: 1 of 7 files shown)


potatoqualitee (approved):
No material defects found.

PR #10671, opened 2026-09-01
Set-DbaNetworkCertificate - Say why a private key is unsuitable

andreasjordan:
## Problem

A certificate whose private key lives in a Key Storage Provider - what `New-SelfSignedCertificate` creates by default - is refused by `Set-DbaNetworkCertificate` with `Failed checks: PrivateKeyInvalid`. That sends people looking at key permissions, which are not the problem.

## Mechanism

SQL Server can only use a legacy CSP private key created with KeySpec `AT_KEYEXCHANGE`; CNG keys are not supported at all (Microsoft documentation). So the check was right - it just did not say why it failed. `Test-DbaNetworkCertificate` already knows the key type and key number, but the message did not use them.

## What changed

The failed check now names the requirement and describes the key it found: the type plus `KeyNumber` when the key is accessible, otherwise "not accessible or a CNG (Key Storage Provider) key". The check itself is unchanged. The `PrivateKeyValid` description of `Test-DbaNetworkCertificate` says the same.

## Tests

- New test on `InstanceRestart`: a default `New-SelfSignedCertificate` on the target host is refused with the new message and nothing is configured; the certificate is tracked and removed again in `AfterAll`.
- Lab: the command verified on PowerShell 7.6 and 5.1; `Set-DbaNetworkCertificate.Tests.ps1` green through the testing-dbatools harness against a case sensitive SQL Server 2022 as `InstanceRestart`. The first harness run caught the test itself handing the command an `Invoke-Command2` result without `-Raw` (an object with only a `Length`), fixed in the second commit.

### DIFF
--- a/public/Set-DbaNetworkCertificate.ps1
+++ b/public/Set-DbaNetworkCertificate.ps1
@@ -316,7 +316,20 @@ function Set-DbaNetworkCertificate {
                     if (-not $detailedCertTest.CertificateFound) { $failedChecks += "CertificateNotFound" }
                     if ($detailedCertTest.CertificateFound -and -not $detailedCertTest.KeyUsagesValid) { $failedChecks += "KeyUsagesInvalid" }
                     if ($detailedCertTest.CertificateFound -and -not $detailedCertTest.DnsNamesValid) { $failedChecks += "DnsNamesInvalid" }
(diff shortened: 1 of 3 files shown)


potatoqualitee (approved):
No material defects found.

potatoqualitee:
thank you 🙏🏼  btw fable 5.1 came out today. consensus so far is that it's more token hungry than fable 5 and so far, no bigger payoff, we'll see.

PR #10672, opened 2026-09-01
Connect-DbaInstance - Keep the tab completion cache of instance names an array

andreasjordan:
## Problem

The tab completion cache for `-SqlInstance` (`[Dataplat.Dbatools.TabExpansion.TabExpansionHost]::Cache["sqlinstance"]`) is a `System.String` instead of an array as soon as one instance is in it. From the second distinct instance on, every new name is concatenated onto that string, and because `-notcontains` then compares the whole string, every later connection to an already known instance appends its name again. In a long session the cache reads `sql05\sql2025sql05\sql2019sql05\sql2019sql05\sql2019...`, and the completer offers that blob instead of instance names. Both editions, reproduced with two `Connect-DbaInstance` calls: type `System.String`, value `sql05\sql2019sql05\sql2022`.

Surfaced in a full lab run: `Add-DbaInstanceList.Tests.ps1` failed with `Expected 'dbatoolsci_testinstance_...' to be found in collection sql05\sql2025sql05\sql2019sql05\sql2019...`. The file passes alone because a fresh process has an empty cache and a one-name string still satisfies `-contains`.

## Mechanism

The cache starts as `$null`. `$null += "x"` yields the string `"x"`, not a one-element array, and every writer used `+=`: `Connect-DbaInstance` after each connection, `Add-DbaInstanceList`, and `private/dynamicparams/sqlinstance.ps1` for the configured and environment instances. The `@()` initialisation in that script does not help, because the TEPP maintenance task that loads it runs later than the first connection in most sessions.

## What changed

- New private function `Add-DbaTeppInstanceName` (in `private/functions/tabcompletion`, so the TEPP maintenance runspace has it too): rebuilds the cache as an array every time, lower cases the names as the completer expects, skips empty entries and duplicates.
- The three writers call it instead of `+=`. No behaviour change beyond the cache being an array of unique names.

## Tests

- `Add-DbaInstanceList.Tests.ps1`: new test adds two instances and asserts the cache is an array containing both, with the first one exactly once. Red on old: `Expected $true, but got $false` on the array check (the cache is the string `dbatoolsci_testinstance2_...dbatoolsci_testinstance_...`).
- `Connect-DbaInstance.Tests.ps1`: new context connects to `InstanceMulti1` and `InstanceMulti2` and asserts both names are separate entries.
- Both files green through the testing-dbatools harness on PowerShell 7.6 and 5.1.

### DIFF
--- a/private/dynamicparams/sqlinstance.ps1
+++ b/private/dynamicparams/sqlinstance.ps1
@@ -5,18 +5,13 @@ if (-not [Dataplat.Dbatools.TabExpansion.TabExpansionHost]::Cache["sqlinstance"]
 
 # Load user-defined instances from config (set via Add-DbaInstanceList)
 foreach ($instance in (Get-DbatoolsConfigValue -FullName "TabExpansion.KnownInstances" -Fallback @())) {
-    if ($instance -and [Dataplat.Dbatools.TabExpansion.TabExpansionHost]::Cache["sqlinstance"] -notcontains $instance) {
-        [Dataplat.Dbatools.TabExpansion.TabExpansionHost]::Cache["sqlinstance"] += $instance
-    }
+    Add-DbaTeppInstanceName -Name $instance
 }
 
 # Load from environment variable (comma-separated list, e.g. set in PowerShell profile)
 if ($env:DBATOOLS_KNOWN_INSTANCES) {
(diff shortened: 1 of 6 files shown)


potatoqualitee (approved):
No material defects found.

potatoqualitee:
nice change, thank you 🙏🏼

PR #10673, opened 2026-09-01
Get-DbaWsfcResourceGroup - Name the PartialOnline and Pending states of a resource group

andreasjordan:
## Problem

`Get-DbaWsfcResourceGroup` and `Get-DbaWsfcRole` report `State = 3` for a resource group that has one resource offline, and `4` for a group in transition. The documentation promises "Online, Offline, Failed, or Unknown", and both test files assert that every state is translated - which is how this surfaced in the lab: after a cluster resource restart both SQL Server groups sat at `PartialOnline` for a while (their CEIP service resources had stayed offline) and both files failed with `Expected $null or empty, but got @('SQL Server (MSSQLSERVER)', 'SQL Server (SQL2022)')`.

## Mechanism

`private/functions/Get-ResourceGroupState.ps1` maps -1/0/1/2 and returns the raw number for everything else. `MSCluster_ResourceGroup.State` has two more documented values: 3 = PartialOnline, 4 = Pending.

## What changed

- `Get-ResourceGroupState` names 3 as `PartialOnline` and 4 as `Pending`; anything else still comes back as the number, as before.
- The `.OUTPUTS` / `.DESCRIPTION` of both commands list the two new names.
- Both test files accept them in "Translates the state of every resource group / role".
- New unit test in `Get-DbaWsfcResourceGroup.Tests.ps1` that pins all six translations, run in module scope. Red on old: `Expected 'PartialOnline', but got 3`.

## Tests

- Unit test green in both editions.
- Lab: both files green against the lab cluster with the groups fully online; and with one CEIP resource taken offline on purpose (group `PartialOnline`), the old code fails the translation assertion and the new code passes it. CEIP brought back online afterwards.

### DIFF
--- a/private/functions/Get-ResourceGroupState.ps1
+++ b/private/functions/Get-ResourceGroupState.ps1
@@ -4,6 +4,8 @@ function Get-ResourceGroupState ($state) {
         0 { "Online" }
         1 { "Offline" }
         2 { "Failed" }
+        3 { "PartialOnline" }
+        4 { "Pending" }
         default { $state }
     }
 }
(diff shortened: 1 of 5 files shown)


potatoqualitee (approved):
No material defects found.

PR #10677, opened 2026-09-03
Get-BulkRowsCopiedCount - Read the rows copied counter as Int64

### LINKED ISSUE #10675: Copy-DbaDbTableData verbose progress counter over-reports rows copied on final batch
**Environment:** dbatools 2.8.2 (also seen on 2.7.1), SQL Server

**Description:**
The live progress number `Copy-DbaDbTableData` writes to its verbose stream (e.g. "adjusted total rows copied = N") can over-report versus the destination table's real row count once the copy finishes. This looks like the command's own internal bookkeeping - built to work around `SqlBulkCopy`'s legacy 32-bit `RowsCopied` counter - has a double-counting edge case on the final batch.

**Repro:**
- Run `Copy-DbaDbTableData` with `-Verbose` on a table large enough to span multiple batches (`-BatchSize`).
- Watch the verbose "adjusted total rows copied" value as the final batch completes.
- Compare against `SELECT COUNT_BIG(*)` on the destination table after the run finishes.
- The verbose total can be higher than the real row count.

**Impact:**
Cosmetic only - the extra count is not reflected in what's actually written to the destination, confirmed by a real `COUNT_BIG(*)` afterwards. But it's misleading for anyone live-monitoring a large transfer and expecting the reported total to match the final row count, and it can make a clean transfer look like it over-copied when it didn't.

**Expected:**
The verbose "rows copied" total should match the destination's actual row count once the last batch completes, or the discrepancy should be documented as a known artifact of the 32-bit counter workaround.

**Workaround used:**
Don't trust the live verbose counter for the final total - do a direct `COUNT_BIG(*)` comparison against source/destination after the transfer instead.

andreasjordan:
Fixes #10675

## Problem

The report: the verbose rows-copied total of `Copy-DbaDbTableData` can over-report against the destination's real row count on the final batch, while the copied data itself is correct.

## What reproduced and what did not

On the current stack (dbatools.library with Microsoft.Data.SqlClient 6.1.5, SQL Server 2022 -> 2025) a multi-batch copy shows correct live counters and a correct final total, with the batch size aligned to the row count and not - the reported symptom does not reproduce below `[int32]::MaxValue` rows. @niphlod is right that the wrapping 4-byte counter itself is history: `SqlBulkCopy._rowsCopied` is an **Int64** in current Microsoft.Data.SqlClient, so the counter never wraps any more.

But the machinery the issue points at does over-report the final total, exactly once the copy exceeds `[int32]::MaxValue` rows:

1. `Get-BulkRowsCopiedCount` reads `_rowsCopied` by reflection and casts it to `[int]`. Above 2,147,483,647 rows that cast throws.
2. The `catch` returns **-1**, the helper's failure sentinel.
3. The three callers feed that sentinel into `Get-AdjustedTotalRowsCopied`, whose integer-wrap branch takes a negative value for a wrapped counter and adds ~1.3 billion phantom rows.

Demonstrated by setting the field via reflection: a 3,000,000,000-row copy reports **4,294,967,295** rows copied while the destination holds the correct 3,000,000,000. The data is right, the reported total is not - the shape the issue describes. Ironically the over-report comes from the workaround for the old 4-byte counter, on a library whose counter no longer wraps.

## What changed

- `Get-BulkRowsCopiedCount` returns the counter as `[long]` (`[OutputType([long])]`); the value passes through unharmed on both the Int64 field of current SqlClient and the Int32 field of the legacy library.
- The three callers with the final-adjustment pattern - `Copy-DbaDbTableData`, `Import-DbaCsv`, `Import-DbaParquet` - skip the final adjustment when the helper signals failure with -1, keeping the running total from the copy notifications instead of adding a bogus wrap correction.

- `Get-AdjustedTotalRowsCopied` keeps its wrap logic untouched: it is still correct for genuinely wrapped (negative) values that an old library delivers through the `SqlRowsCopied` event.
- `Write-DbaDbTableData` uses only the event-side adjustment and never calls `Get-BulkRowsCopiedCount`, so it has no sentinel path and stays as it is.

## Tests

New `tests/Get-BulkRowsCopiedCount.Tests.ps1` (unit, `InModuleScope`, no instance needed): the helper returns the counter below and above `[int32]::MaxValue`, and the callers' final-adjustment pattern stays correct above it. Red on `development` (`Expected 3000000000, but got -1.`), green with the fix.

Full runs of the four affected test files through the testing-dbatools harness (SQL 2022 -> SQL 2025), both editions: 79/79 green on pwsh and on 5.1.

Copying two billion real rows is not something CI or the lab can afford, which is why the regression test sets the counter field by reflection instead - the same access path the helper itself uses.

### DIFF
--- a/private/functions/Get-BulkRowsCopiedCount.ps1
+++ b/private/functions/Get-BulkRowsCopiedCount.ps1
@@ -14,6 +14,7 @@ function Get-BulkRowsCopiedCount {
             - Copy-DbaDbTableData
             - Copy-DbaDbViewData
             - Import-DbaCsv
+            - Import-DbaParquet
 
         .EXAMPLE
             Get-BulkRowsCopied $bulkObject
@@ -27,14 +28,17 @@ function Get-BulkRowsCopiedCount {
         Copyright: (c) 2020 by dbatools, licensed under MIT
         License: MIT https://opensource.org/licenses/MIT
     #>
-    [OutputType([int])]
+    [OutputType([long])]
     param (
         [Microsoft.Data.SqlClient.SqlBulkCopy] $BulkCopy
     )
     $BindingFlags = [Reflection.BindingFlags] "NonPublic,GetField,Instance"
     $rowsCopiedField = [Microsoft.Data.SqlClient.SqlBulkCopy].GetField("_rowsCopied", $BindingFlags)
     try {
-        return [int]$rowsCopiedField.GetValue($BulkCopy)
+        # The field is an Int64 in current Microsoft.Data.SqlClient (an Int32 in the legacy library),
+        # so it must not be narrowed to [int]: above [int32]::MaxValue rows that cast throws, and the
+        # -1 then taken for the row count inflates the reported total by billions of rows (see #10675).
+        return [long]$rowsCopiedField.GetValue($BulkCopy)
     } catch {
         return -1;
     }
--- a/public/Copy-DbaDbTableData.ps1
+++ b/public/Copy-DbaDbTableData.ps1
@@ -722,7 +722,13 @@ function Copy-DbaDbTableData {
                         $bulkCopy.WriteToServer($reader)
                         $finalRowCountReported = Get-BulkRowsCopiedCount $bulkCopy
 
-                        $script:totalRowsCopied += (Get-AdjustedTotalRowsCopied -ReportedRowsCopied $finalRowCountReported -PreviousRowsCopied $script:prevRowsCopied).NewRowCountAdded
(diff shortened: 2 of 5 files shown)


potatoqualitee (approved):
Reading the bulk-copy counter as Int64 fixes the reproducible overflow above Int32.MaxValue, and the callers correctly keep the -1 failure sentinel out of adjustment logic. CI and the focused large-counter tests pass; I found no material defects.

potatoqualitee:
thank you, test doesnt follow usual pattern but it's unit tests and internal command and you approved so we're good 👍🏼

PR #10678, opened 2026-09-03
Copy-DbaDbTableData - Name the Query requirement in the parameter guard message

### LINKED ISSUE #10676: Copy-DbaDbTableData -Query still requires -Table, with an unhelpful error otherwise
**Environment:** dbatools 2.8.2 (also seen on 2.7.1), SQL Server

**Description:**
`Copy-DbaDbTableData` in `-Query` mode still requires `-Table` to be specified, even though the query itself already determines what's being read from the source. Omitting `-Table` when using `-Query` produces a generic error that doesn't mention `-Query` at all:

```
You must pipe in a table or specify SqlInstance, Database and [View|Table]
```

**Repro:**
- Run `Copy-DbaDbTableData -SqlInstance <src> -Database <db> -Query "SELECT ..." -DestinationSqlInstance <dst> -DestinationDatabase <db> -DestinationTable <table>` without `-Table`.
- Result: the error above, which reads as if `-SqlInstance`/`-Database` were missing (they weren't) rather than pointing at the actual missing parameter.

**Where this bit us:**
Hit this while chunked-transferring a large partitioned table (date-column chunking via `-Query`), the most expensive case to get an unclear error on.

**Expected:**
Either drop the `-Table` requirement when `-Query` is supplied (since the query already defines the source), or make the error message explicitly say that `-Table` is required alongside `-Query` in this mode.

**Workaround used:**
Always pass `-Table` alongside `-Query`, even though the query is what actually determines the source data.

andreasjordan:
Fixes #10676

## Problem

`Copy-DbaDbTableData -Query` without `-Table` or `-View` answers with "You must pipe in a table or specify SqlInstance, Database and [View|Table]." - with `-SqlInstance`, `-Database` and `-Query` all supplied, that reads as if the wrong parameters were missing, on what is often the most expensive scenario to get an unclear error on (chunked transfers of large tables).

## Why the requirement itself stays

The resolved SMO table or view is not just metadata garnish in `-Query` mode: it supplies the source connection context (`$sqlObject.Parent`), the structure `AutoCreateTable` scripts, the default `-DestinationTable` name, the clustered-index and primary-key lookups, and the source columns of the output object. Making `-Table` optional with `-Query` means rebuilding all of that from the query alone - a feature-level change with real design questions (what does `AutoCreateTable` script? what do the Source* output columns show?), not a bug fix. The help for `-Query` already documents the requirement ("Still requires specifying a Table or View parameter for metadata purposes").

## What changed

The parameter guard now answers a `-Query` call with a message that names the actual gap:

> When using Query, you still have to specify SqlInstance, Database and [View|Table]. The query determines the data that is copied, but the command needs the table or view as the source object for its metadata.

Everything else about the guard is unchanged, and calls without `-Query` keep the original message. `Copy-DbaDbViewData` exposes `-Query` too and splats into this function, so it gets the same message for free.

## Tests

New regression test beside the existing `-Query` tests: calling with `-Query` and no `-Table` returns nothing and warns with the new message. Red on `development` (the generic message came back), green with the fix - through the testing-dbatools harness on both editions.

Thanks @JankeUwe for the report.

### DIFF
--- a/public/Copy-DbaDbTableData.ps1
+++ b/public/Copy-DbaDbTableData.ps1
@@ -326,7 +326,12 @@ function Copy-DbaDbTableData {
 
     process {
         if ((Test-Bound -Not -ParameterName InputObject) -and ((Test-Bound -Not -ParameterName SqlInstance, Database -And) -or (Test-Bound -Not -ParameterName Table, View))) {
-            Stop-Function -Message "You must pipe in a table or specify SqlInstance, Database and [View|Table]."
+            if (Test-Bound -ParameterName Query) {
+                # Without the hint at Query the message reads as if SqlInstance or Database were missing (see #10676).
+                Stop-Function -Message "When using Query, you still have to specify SqlInstance, Database and [View|Table]. The query determines the data that is copied, but the command needs the table or view as the source object for its metadata."
+            } else {
+                Stop-Function -Message "You must pipe in a table or specify SqlInstance, Database and [View|Table]."
+            }
             return
         }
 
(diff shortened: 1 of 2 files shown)


potatoqualitee (approved):
The guard now accurately explains the metadata requirement for query-based copies without changing non-query behavior. The focused regression test covers the reported path, and I found no material defects.

potatoqualitee:
reasonable, ty 🙏🏼

PR #10683, opened 2026-09-10
Get-DbaWindowsLog - Compile the remote script inside the pool runspace so local instances do not hang

andreasjordan:
## What

Get-DbaWindowsLog hands its remote-execution script to a runspace pool and, for a local instance, runs it in process with Invoke-Command. That script was passed as a scriptblock object created in the calling runspace. The command now passes it as text and compiles it inside the pool runspace, so the script belongs to the runspace that runs it.

## Why

The ci-azure SINGLE 5/5 job has been dying on this test file with the watchdog message "Test file Get-DbaWindowsLog.Tests.ps1 has run for 15 minutes" on the development runs of 2026-09-02 and 2026-09-09 (both runs that day). The CI instances are local to the runner, so the command takes the in-process path, which the lab never exercises because its instances are remote.

A scriptblock object keeps its affinity to the runspace that created it. When the pool runspace invokes it in process, Windows PowerShell executes it against the calling runspace's session state while that runspace is busy in the wait loop of the command, and Get-WinEvent intermittently never returns. The remoting path is unaffected because remoting serializes the scriptblock to text anyway.

## Evidence

Reproduced on this workstation without SQL Server by registering an event source and writing a fake event 17111 that points at a copied ERRORLOG, then running `Get-DbaWindowsLog -SqlInstance localhost\REPRO` with a fresh event source each time:

| Edition | Before | After |
|---|---|---|
| Windows PowerShell 5.1 | 3 hangs in 15 runs (a 4th earlier in 9 runs) | 0 hangs in 25 runs |
| PowerShell 7.6 | 0 hangs | 0 hangs |

Marker output of the hung runs always stopped right before the Get-WinEvent call inside the pool runspace. The parsing itself takes under three seconds for seven ERRORLOG files, so this is a wait, not slowness.

Lab verification through the test harness: Get-DbaWindowsLog unit and integration tests pass on 5.1 and 7.6 against the remote instance, and the command returns the same six rows against SQL03\SQL2019 on both editions.


potatoqualitee (approved):
Recreating the worker script block inside the pool runspace removes the problematic runspace affinity while preserving the remote and multi-instance paths. CI is green, including the targeted Windows test path, and I found no material defects.

potatoqualitee:
nice fix thank you 👍🏼

PR #10684, opened 2026-09-10
Invoke-DbaDbLogShipping - Give the helper failure test a UNC shared path on CI

andreasjordan:
## What

The test "Emits the failed status object instead of skipping it", added in #10644, passes the shared path in the form the command requires. On CI, where `$TestConfig.Temp` is a local folder on the runner that also hosts the instance, the path is converted to the administrative share of its drive. In the lab, where the temp folder already is a share, nothing changes. `IgnoreFileChecks` skips the reachability test of the share, the full backup is still taken to the local path, and the per-database subfolder is created by the test session when the instance cannot see the share.

## Why

The ci-azure HADR job has failed on every development run since #10644 merged on 2026-09-01 with

```
Invoke-DbaDbLogShipping > When a helper fails after the setup phases > Emits the failed status object instead of skipping it: Expected a value, but got $null or empty.
```

The command rejects any shared path that is not of the form `\server\share` with a warning, and the test silences warnings, so the command returned nothing before it ever reached the mocked helper. The first context of the same file already documents this trap, this one had missed it.

## Verification

Both contexts pass in the lab through the test harness on Windows PowerShell 5.1 and PowerShell 7.6. The lab uses a share for its temp folder, so the conversion branch cannot be exercised there. It is verified in isolation: `C:\Temp\dbatoolsci_lsfail_123` becomes `\ADMIN01\C$\Temp\dbatoolsci_lsfail_123`, which satisfies the command's UNC rule, and a share path passes through unchanged. The HADR job of this pull request is the real test.


potatoqualitee (approved):
The CI path conversion correctly supplies a UNC administrative-share path while preserving existing UNC paths, and the regression test meaningfully exercises the intended helper-failure path. I found no material defects.

potatoqualitee:
looks good 💯  thank you

PR #10686, opened 2026-09-10
Copy-DbaDbTableData - Close the source reader when the bulk copy fails

### LINKED ISSUE #10685: Copy-DbaDbTableData leaks source connection and SqlDataReader when bulk copy fails

The error surfaced to the caller is just an example of a transient network failure during `WriteToServer` - in our case an unstable link to Azure SQL:

```
An existing connection was forcibly closed by the remote host
(German original: Eine vorhandene Verbindung wurde vom Remotehost geschlossen)
```

That error itself is not the bug and any failure during `WriteToServer` will do. The actual problem is the side effect: after **any** exception, the source connection and the `SqlDataReader` are never closed.

On the source instance a session remains stuck indefinitely:

```
session_id  status     command     wait_type          wait_time  total_elapsed_time
141         suspended  SELECT      ASYNC_NETWORK_IO   1779       834926

query_text: SELECT [...] FROM SourceTable
```

The orphaned reader holds a schema stability lock, which then blocks any DDL on the source table:

```
session_id  command     wait_type   wait_time  blocking_session_id
51          DROP TABLE  LCK_M_X     444463     141
```

The only way to recover is `KILL 141` by hand.

### Steps to Reproduce

Run any `Copy-DbaDbTableData` call and interrupt the **destination** connection while `WriteToServer` is still running, for example by issuing `KILL <spid>` against the destination session or by dropping the network link.

Then check the **source** instance:

```sql
SELECT session_id, status, command, wait_type
FROM sys.dm_exec_requests
WHERE wait_type = 'ASYNC_NETWORK_IO';
```

The `SELECT` from the failed attempt is still there and stays until killed manually.

2.8.4

### Analysis

In `public/Copy-DbaDbTableData.ps1` the cleanup of the source side lives **inside** the `try` block, right after `WriteToServer`:

```powershell
$bulkCopy.WriteToServer($reader)
...
$server.ConnectionContext.SqlConnectionObject.Close()   # source
$bulkCopy.Close()
$bulkCopy.Dispose()
$reader.Close()                                          # reader
...
} catch {
    Stop-Function -Message "Something went wrong" -ErrorRecord $_ -Target $server -continue
} finally {
    if ($bulkCopyConnection) {
        $bulkCopyConnection.Close()      # only the DESTINATION is cleaned up
        $bulkCopyConnection.Dispose()
    }
}
```

If `WriteToServer` throws, execution jumps straight to `catch`, so those four lines never run. The `finally` block only disposes `$bulkCopyConnection` (destination). `$reader`, `$bulkCopy` and the source connection are leaked.

### Suggested fix

Move the source-side cleanup into the existing `finally` block, mirroring what is already done for the destination:

```powershell
} finally {
    if ($reader) {
        try { $reader.Close(); $reader.Dispose() } catch { }
    }
    if ($bulkCopy) {
        try { $bulkCopy.Close(); $bulkCopy.Dispose() } catch { }
    }
    if ($server.ConnectionContext.SqlConnectionObject.State -eq 'Open') {
        try { $server.ConnectionContext.SqlConnectionObject.Close() } catch { }
    }
    if ($bulkCopyConnection) {
        $bulkCopyConnection.Close()
        $bulkCopyConnection.Dispose()
    }
}
```

### Why this matters

The function is a natural fit for chunked/retry loops over unreliable links. In such a loop every failed attempt leaks one source session. In our case 106 chunks produced 2 transient failures; each left an `ASYNC_NETWORK_IO` session behind that eventually blocked `DROP TABLE` on a persisted staging table, created before the sync, requiring manual intervention.

### PowerShell Host Version

PSVersion                      5.1.17763.9020                                                                                
BuildVersion                   10.0.17763.9020                                                                               

Source: 
Microsoft SQL Server 2019 (RTM-GDR) (KB5102336) - 15.0.2180.2 (X64)   Jun 18 2026 18:09:53   Copyright (C) 2019 Microsoft Corporation  Standard Edition (64-bit) on Windows Server 2019 Standard 10.0 <X64> (Build 17763: ) (Hypervisor)  (on-premises)

Destination: 
Microsoft SQL Azure (RTM) - 12.0.2000.8   Aug 19 2026 09:20:28   Copyright (C) 2026 Microsoft Corporation 

### .NET Framework Version

.NET Framework 4.8.4795.0

andreasjordan:
Fixes #10685

## Problem

When `WriteToServer` throws (a dropped link to the destination, a killed destination session, a value the destination column cannot take), the source side is never cleaned up. The `SqlDataReader` stays open, the SELECT keeps running on the source with wait type `ASYNC_NETWORK_IO` once the network buffers are full, and it holds its schema stability lock on the source table until someone kills the session by hand. In a chunked retry loop every failed chunk leaks one more session.

## Mechanism

The four source-side cleanup lines (close the source connection, close and dispose the `SqlBulkCopy`, close the reader) sit inside the `try` after `WriteToServer`, so an exception jumps straight past them into `catch`. The `finally` only disposed the destination connection. @mangenendt-sudo's analysis in the issue is exactly right.

## What changed

- The `finally` now handles the leftovers: if the reader is still open, it cancels the command and closes the reader, then disposes the `SqlBulkCopy` and returns the source connection, the same way the success path already does.
- The command is cancelled *before* the reader is closed. Closing a reader with unread rows makes SqlClient drain the whole remaining result set from the source first, which for a large table can take as long as the copy itself. With the cancel the failing call returns in about a second in the lab, no matter how many rows were left.
- The three objects are reset to `$null` before the `try`, so a reader from a previous table or destination in the same call can never be mistaken for the current one.

- The success path keeps its existing cleanup inside the `try`, so a successful copy behaves exactly as before. The `finally` only acts on what the success path left behind (it checks `IsClosed`).
- The destination connection was already cleaned up in the `finally` and stays there.
- The progress bar is not completed on the failure path. That belongs to the progress-bar work in #10548 / #10660.

## Sibling check

The other `WriteToServer` sites (`Import-DbaCsv`, `Import-DbaParquet`) and `Export-DbaBinaryFile` already clean up their readers in a `finally`; `Write-DbaDbTableData` writes a DataTable and has no reader. `Export-DbaCsv` has the same shape (reader closed only inside the `try`), but its `finally` disconnects a connection the command opened itself, which closes the reader with it. It only leaks for a caller-owned server object when the CSV writer throws, so it is left for a separate PR if wanted.

## Tests

- New integration test in `Copy-DbaDbTableData.Tests.ps1`: a query with two cross joins over `sys.all_objects` returns a string for an `INT` destination column, so the bulk copy fails on the first row while millions of rows are still pending on the source. The test asserts that no request on the source is still running the query afterwards and that an `ALTER TABLE` on the source table goes through with a 5 second lock timeout. Its `AfterAll` kills a leaked request before dropping the table, so a regression fails the test instead of hanging the run.
- Red on old: against the current `development` version of the command the test fails with "Expected 0, but got 1" on the leaked request.
- Green on the fix: 15/15 on Windows PowerShell 5.1 and on PowerShell 7.6 against SQL Server 2022 (source) and 2025 (destination) in the lab, zero warnings, lab left clean.
- Reproduced the report first on unmodified code on both editions: the source SELECT stayed running with an `IS` lock on the table, and a `DROP TABLE` with a 5 second lock timeout failed.


potatoqualitee (approved):
The failure path now cancels and closes the source-side resources and disposes both bulk-copy and destination connection resources. The regression test demonstrates the source request and schema lock are released, and I found no material defects.

potatoqualitee:
thank you 💯

PR #10690, opened 2026-09-12
Workflows - Wait for the SQL Server containers before running the core tests

### LINKED ISSUE #10590: Gallery workflow: "sets up a mirror" fails intermittently, nothing waits for the second container to be ready
### Summarize the problem

The `sets up a mirror` test in the "Test version in Gallery" workflow fails intermittently, because nothing waits for the second SQL Server container to be ready before the test run starts.

Seen on PR #10588, whose changes cannot reach this test at all - it only touches `Get-DbaWsfc*`, which cannot even run on Linux. Re-running the identical commit passed, so no code changed between the failure and the success.

- failed: https://github.com/dataplat/dbatools/actions/runs/32394459115/attempts/1
- passed on re-run of the same commit: https://github.com/dataplat/dbatools/actions/runs/32394459115

### The failure

```
WARNING: [Invoke-DbaDbMirroring] Failure on primary | An exception occurred while executing a
Transact-SQL statement or batch. Communications to the remote server instance 'TCP://mssql2:5022'
failed before database mirroring was fully started. The ALTER DATABASE command failed. Retry the
command when the remote database is started.

[-] sets up a mirror 2.06s
  at Invoke-DbaDbMirroring @params | Select-Object -ExpandProperty Status | Should -Be "Success",
  .github/scripts/gh-actions.ps1:83
```

That message is SQL Server's own "the partner was not listening yet" error.

### Mechanism

`.github/workflows/gallery.yml` starts three containers with `docker run -d` and then moves straight on:

```yaml
docker run -p 1433:1433  ... --name mssql1 ... -d dbatools/sqlinstance
docker run -p 14333:1433 ... --name mssql2 ... -d dbatools/sqlinstance2
docker run -p 14334:1433 ... --name mssql3 ... -d dbatools/sqlinstance2
```

There is no health check and no wait loop. The steps that follow - cloning appveyor-lab, importing dbatools - happen to give the containers a few seconds, and that is usually enough. When it is not, the first test that needs the *second* instance fails.

Mirroring makes this more likely than the other tests, because it needs more than a reachable engine: `mssql2` must also have its mirroring endpoint listening on 5022, and the primary reaches it container to container over `localnet` as `TCP://mssql2:5022`, not through the published port. The engine can be accepting logins on 1433 while the endpoint is still starting.

### How often

Once in the last 41 runs of the workflow, so this is rare rather than routinely flaky. Opening it anyway because the failure is indistinguishable from a real regression at first glance, and it cost a review cycle on an unrelated PR to establish that it was not one.

### Suggested fix

Wait for readiness in the workflow before running the tests, rather than relying on the incidental delay of the intervening steps. Something that polls each instance until it accepts a connection, and additionally waits for the mirroring endpoint on `mssql2`:

```sql
SELECT state_desc FROM sys.database_mirroring_endpoints
```

until it reports `STARTED`.

A retry around the single test would also make it green, but it would only hide the same race for whichever test happens to run first against `mssql2`.

I have not touched this, since it is CI scripting rather than module code.

created by Claude

andreasjordan:
Fixes #10590

## Problem

`Run Cross Platform Tests` and `Test version in Gallery` start three SQL Server containers with `docker run -d` and move straight on. Nothing waits for them; the steps in between happen to give them a few seconds, and when that is not enough the first test that needs the second instance fails. The recorded case is the `sets up a mirror` test failing on an unrelated PR with SQL Server's own "Communications to the remote server instance 'TCP://mssql2:5022' failed before database mirroring was fully started", and passing on a rerun of the same commit.

Mirroring is the likeliest victim because it needs more than a reachable engine: `mssql2` must have its mirroring endpoint listening on 5022, and the engine can accept logins while the endpoint is still starting.

## What changed

Both workflows get one step after the containers are started, `Wait for the SQL Server containers to be ready`. For each of the three instances it polls up to three minutes until a login succeeds and every mirroring endpoint the instance has reports STARTED (`sys.database_mirroring_endpoints` with `state <> 0` is empty; an instance without an endpoint is ready as soon as it takes logins, since `Invoke-DbaDbMirroring` creates and starts endpoints itself). An instance that does not become ready fails the step with its name, so a broken container image shows up there instead of in whichever test runs first.

The step is the same in both files apart from how dbatools is imported: from the checkout in the cross-platform workflow, from the Gallery in the gallery workflow. Nothing else in the workflows changes.

The mirror test keeps no retry of its own: a retry there would only hide the same race for whichever test happens to run first against `mssql2`.

## Verification

Both workflows trigger on `push` for every branch, so the runs on this branch exercise the new step for real; their step log lists each instance with the version it answered with. The embedded script parses cleanly as PowerShell.


potatoqualitee (approved):
Both workflow variants now wait for all three SQL instances and their mirroring endpoints with bounded retries and clear instance-specific failure messages. The exact-head workflow runs exercised this step successfully and all checks pass.

potatoqualitee:
thank you 🙏🏼

PR #10691, opened 2026-09-12
Invoke-DbaQuery - Close the connection it opened when the query fails

### LINKED ISSUE #10659: Invoke-DbaQuery - Failed executions leak the non-pooled connection it opened
Noticed in passing in PR #10634's description (Export-DbaCsv got its disconnect placed in a `finally` explicitly because "the sibling pattern in `Invoke-DbaQuery` places the disconnect after the catch and would miss it") — verified in source and recorded here so it is not lost.

## Behavior

`Invoke-DbaQuery` opens a **non-pooled** connection per instance when it is not handed an open one (`NonPooledConnection = $true`, see #8491), and closes it after execution via the #10554 ownership pattern:

```powershell
} catch {
    Stop-Function -Message "[$instance] Failed during execution" -ErrorRecord $_ -Target $instance -Continue
}
# Only close the connection if Connect-DbaInstance opened a new one for us. ...
if ($isNewConnection) {
}
```

The disconnect sits **after** the try/catch. When the query fails:

- without `-EnableException`, the catch's `-Continue` moves to the next instance of the loop — the disconnect is skipped;
- with `-EnableException`, `Stop-Function` throws — the disconnect is skipped as well.

Either way the freshly opened non-pooled connection stays checked out until the process exits. Non-pooled connections are precisely the ones nothing else will ever clean up. A retry loop around failing queries accumulates one sleeping session per attempt.

## Fix shape

Move the ownership disconnect into a `finally` around the execution, exactly as `Export-DbaCsv` does since #10634 — the disconnect then runs on the success, failure and `-Continue` paths alike. A counting regression test per the #10633–#10635 pattern, with the failure path (a query that throws) asserted explicitly, proves it.

andreasjordan:
Fixes #10659

## Problem

`Invoke-DbaQuery` opens a non-pooled connection per instance when it is not handed an open one, and closed it after the execution block - after the `try`/`catch`, not inside a `finally`. When the query fails, that block is left through the `-Continue` of the catch, or through the exception it throws under `-EnableException`, and both paths skipped the disconnect. The non-pooled session stayed checked out until the process ended. Non-pooled connections are exactly the ones nothing else cleans up, so a retry loop around a failing query accumulated one sleeping session per attempt.

Reproduced in the lab: with development's command, three failing calls without `-EnableException` raised the count of sleeping dbatools sessions of the test process from 3 to 6, and three more with `-EnableException` to 9.

## What changed

The ownership disconnect moved into a `finally` around the execution, the same shape `Export-DbaCsv` got in #10634, so it runs on the success, warning and throw paths alike. As there, the disconnect passes `-WhatIf:$false -Confirm:$false`: it is bookkeeping for a connection the command opened itself, and `Disconnect-DbaInstance` supports ShouldProcess, so a propagated `$WhatIfPreference` would otherwise skip it.

Nothing else changes. Connections handed in by the caller are still left open, which the existing #10554 tests keep asserting.

## Sweep

The same ownership pattern lives in six other commands. Two of them share the defect - `Install-DbaMaintenanceSolution` and `Update-DbaMaintenanceSolution` both `-Continue` out of the loop between opening their non-pooled connection and closing it - and get the same fix in their own pull requests. `Get-DbaDbExtentDiff` has no exit between open and close, `Invoke-DbaAdvancedRestore` uses a pooled connection, and `Connect-DbaInstance`'s own sites are its early-exit cleanups, so those stay as they are.

## Tests

New context `Connections it opens are closed when the query fails (#10659)`: counts the sleeping dbatools sessions of the test process through a server object opened once, runs three failing queries without `-EnableException` and three with it, and asserts the count is back at the baseline after each batch, plus the warning text on the first batch. Through the testing-dbatools harness: green on both editions with the fix; red without the command change, failing exactly those two assertions with the counts above.


potatoqualitee (approved):
The finally block closes only command-owned connections across success, warning, and terminating-error paths while preserving caller-owned connections. The integration coverage exercises both failure modes and all exact-head checks pass.

potatoqualitee:
awesome ty

PR #10688, opened 2026-09-12
Workflows - Pull MinIO from quay.io and mc from GitHub releases

andreasjordan:
## Problem

`Run S3 Backup Integration Tests` and `Run External Table Integration Tests` fail on every branch since 2026-09-11 22:25 UTC, development included, in the step that starts MinIO:

```
docker: Error response from daemon: pull access denied for minio/minio, repository does not exist or may require 'docker login'
```

The `minio/minio` repository no longer exists on Docker Hub (the Hub API answers 404 for it). The next line of the same step would have failed as well: `https://dl.min.io/client/mc/release/linux-amd64/mc`, where both workflows fetch the `mc` client, now answers `410 Gone`. Nothing in dbatools changed; the three development pushes of that evening and PR #10666 all fail identically.

## Fix

Both workflows pull the server image from quay.io, where MinIO keeps publishing its images, and the `mc` client from the GitHub release of `minio/mc`, pinned by release tag in both places:

- `quay.io/minio/minio:RELEASE.2025-09-07T16-13-09Z` - the last community release; `latest` on quay.io points to the same manifest, but a pin cannot move under us
- `https://github.com/minio/mc/releases/download/RELEASE.2025-08-13T08-35-41Z/mc.linux-amd64.RELEASE.2025-08-13T08-35-41Z` - the current `mc` release, with a real `linux-amd64` asset

Both fetch anonymously (verified with a quay.io pull token and a HEAD on the release asset, 30 MB). Everything else in the steps - TLS certificates, `mc alias`, bucket creation, the seeded CSV - is unchanged.

## Verification

Both workflows trigger on changes to their own file, so the checks on this PR are the proof.

created by Claude and reviewed by Andreas Jordan

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10692, opened 2026-09-12
Restore-DbaDatabase - Say that nothing was restored when the selection is empty

### LINKED ISSUE #10657: Restore-DbaDatabase - Empty backup selection is a silent no-op without -EnableException
Deferred from #10621 and PR #10631, which fixed the `-EnableException` half of the problem and explicitly left this half "for a maintainer decision". Since #10621 closes when that PR merges, the open half gets its own issue so it is not lost.

## Behavior after #10631

When the backup information pipeline selects nothing restorable — diff/log backups without a full to anchor the chain, or a path holding no backups at all:

- with `-EnableException`, the command now throws (that was #10621, fixed);
- without it, `Select-DbaBackupInformation`'s stop degrades to a **warning** and `Restore-DbaDatabase` returns nothing — a successful-looking no-op. A script that restores and then uses the database continues against a database that was never restored; the years-flaky `Get-DbaDbRestoreHistory` tests (#10620) were exactly this.

## Decision to make

Should the non-exception path also treat an empty selection as a failure rather than a warn-and-return? Options discussed in #10621/#10631:

1. Leave as is — warning only. Default dbatools behavior for soft failures; scripts that care can use `-EnableException` or check the (empty) return value.
2. Make the bail-out louder without changing the contract — e.g. a clearly worded terminal warning stating that *nothing was restored and why*, instead of the current low-level "Fullname property not found".
3. Treat it as an error even without `-EnableException`. Most protective, but a behavior change for callers who deliberately point the command at possibly-empty paths.

Option 2 is cheap and helps every caller; whether to go further is the maintainers' call.

andreasjordan:
Fixes #10657

## Problem

When the backup information pipeline selects nothing restorable - diff and log backups without the full that anchors the chain, or nothing newer than the continue point - `Select-DbaBackupInformation` stops. With `-EnableException` that throws since #10631. Without it, the stop is a warning, and `Restore-DbaDatabase` handed the empty selection on to the restore, which returned nothing. The only trace was the low-level "Fullname property not found" warning; the empty result looked like a successful restore, and a script that restores and then uses the database went on against a database that was never restored. The years-flaky `Get-DbaDbRestoreHistory` tests (#10620) were exactly this.

## What changed

Option 2 from the issue: the command stops right after the selection when it is empty, with a warning that says so and names the database and instance:

```
Nothing to restore: the backup information selected no restorable backups for <db> on <instance>. A full backup has to anchor the chain (or -Continue has to point at a database restored with -NoRecovery), see the warning above for what was missing.
```

The warning from the selection still comes first and names what was missing. The check sits after `-StopAfterSelectBackupInformation`, so that switch keeps returning the (empty) selection as before.

The contract. Without `-EnableException` this is still a warning and an empty result, so callers who point the command at possibly empty paths on purpose are not turned into failures (option 3 in the issue). With `-EnableException` nothing changes: the selection already throws before this check is reached. The empty-path case ("No backups passed through") already stopped with its own message since #10631 and is untouched.

## Tests

The #10621 context in `Restore-DbaDatabase.Tests.ps1` now asserts, without `-EnableException`, that both warnings are present and that the new one names the database, and gets a fourth test for the empty path without `-EnableException` (warns "No backups passed through", returns nothing). Through the testing-dbatools harness: green on both editions with the fix; red without the command change on the "Nothing to restore" assertion.


potatoqualitee (approved):
The empty-selection guard reports that no restore occurred without changing the existing EnableException behavior or normal restore paths. The regression test covers the concrete missing-full-backup case; unrelated MinIO failures do not affect this patch.

andreasjordan:
The red `s3-backup-tests` check is not this PR: its MinIO setup step fails on the `minio/minio` image that is gone from Docker Hub, on every branch since 2026-09-11. Fix in #10688; an update of this branch after that merge turns the check green.

potatoqualitee:
fixed! thank you 🙏🏼

PR #10693, opened 2026-09-12
Install-DbaMaintenanceSolution - Close the connection it opened when it skips an instance

andreasjordan:
Same defect as #10659, found by the sweep in #10691.

## Problem

`Install-DbaMaintenanceSolution` opens a non-pooled connection per instance when it is not handed an open one, and closed it at the end of the loop body. The skips after the connection is opened - the database not found on the instance, an unsupported version - leave the body through `-Continue`, or through the exception thrown under `-EnableException`, and both paths skipped the disconnect. The non-pooled session stayed checked out until the process ended, one per skipped instance.

Reproduced in the lab with development's command: three calls naming a database that does not exist raised the count of sleeping dbatools sessions of the test process by one per call, with and without `-EnableException`.

## What changed

The loop body after the connect runs inside a `try` whose `finally` holds the ownership disconnect, the shape `Export-DbaCsv` (#10634) and `Invoke-DbaQuery` (#10691) use, with `-WhatIf:$false -Confirm:$false` so a propagated preference cannot skip bookkeeping for a connection the command opened itself. The body itself is only re-indented; `git diff -w` shows the real change, a handful of lines.

Connections handed in by the caller are still left open, which the existing #10554 context keeps asserting.

## Tests

New context `Connections it opens are closed when it skips an instance (#10659)`: counts the sleeping dbatools sessions of the test process through a server object opened once, calls the command three times with a database that does not exist without `-EnableException` and three times with it, and asserts the count is back at the baseline after each batch, plus the "not found" warning. Through the testing-dbatools harness: green on both editions with the fix; red without the command change on both count assertions.


potatoqualitee (approved):
The skip path now disposes the connection opened by the command without altering installation behavior for eligible instances. The added cleanup coverage and all applicable exact-head checks pass.

potatoqualitee:
Thank you 🙏🏼

PR #10695, opened 2026-09-12
Invoke-DbaAdvancedRestore - Stop applying logs once the stop mark is reached

### LINKED ISSUE #10656: Restore-DbaDatabase - A successful stop-at-mark restore still applies the tail files and throws under -EnableException
Deferred from PR #10636, whose description flagged it as "a design question worth its own issue" — recording it here so it does not get lost when that PR merges.

## Behavior

When `Restore-DbaDatabase` stops at a marked transaction (`-StopMark`, with `-StopBefore` or not) or at `-RestoreTime`, and the stop point is **not** in the last log file of the chain, the command still attempts to apply the remaining log files. SQL Server refuses them ("The log in this backup set begins at LSN ..., which is too recent to apply"), so:

- without `-EnableException`, a successful stop-at-mark restore ends with warnings about the tail files;
- with `-EnableException`, a **successful** stop-at-mark restore **throws** on the first tail file — the caller cannot tell success from failure without probing the database afterwards.

The resurrected StopAt tests in PR #10636 work around this by arranging the mark to be reachable without tail files where necessary; the underlying behavior is unchanged there on purpose.

## Design question

Should the restore loop end gracefully once the stop point has been reached — i.e. detect that the target (mark, or point in time) has been satisfied and skip the remaining files instead of letting SQL Server refuse them?

Arguments for: a restore that achieved exactly what was asked should not warn or throw. Arguments for caution: detecting "the stop point is reached" requires interpreting the restore result rather than just forwarding files, and a wrong detection would silently drop log files from a restore that should have continued. A middle ground could be to catch specifically the "too recent to apply" refusal after a satisfied stop and downgrade it to a verbose message.

andreasjordan:
Fixes #10656

## Problem

A restore that stops at a mark (`-StopMark`, with or without `-StopBefore`, and `-StopAtLsn`, which is a mark to SQL Server) reaches its stop point in one of the log backups. When that is not the last log of the chain, SQL Server refuses every log after it with error 4305 ("The log in this backup set begins at LSN ..., which is too recent to apply"). `Invoke-DbaAdvancedRestore` attempted them anyway, so a restore that did exactly what was asked ended with warnings, threw under `-EnableException`, and left the database in the restoring state, because the recovery sits on the last file and that file was refused. Reproduced in the lab with a full, a log holding the mark and a second log: with `-EnableException` the command threw the 4305 error, without it the third output object reported `RestoreComplete = False`, and the database stayed restoring either way.

## How the stop point is detected

SQL Server says whether a log reached the mark. When it did not, the restore succeeds with message 4329, "This log file contains records logged before the designated mark. The database is being left in the Restoring state so you can apply another log file." When it did, there is no such message. Verified in the lab through T-SQL and through SMO for `STOPATMARK`, `STOPBEFOREMARK` and an LSN mark, on the log holding the mark and on one before it. This is the middle ground the issue asked about: the command interprets one documented server message rather than guessing from LSNs, and a wrong guess cannot drop log files, because a log that does not reach the mark always carries the message.

Point-in-time restores (`-RestoreTime`) are not affected and not touched: SQL Server accepts the later logs of a chain after a `STOPAT` and applies nothing from them, so they never failed this way.

## What changed

In `Invoke-DbaAdvancedRestore`, only when a mark or LSN was requested and the action is a log restore that really executes:

- an `InfoMessage` handler on the connection records whether message 4329 arrived; it is added before the restore and removed in the `finally`. The connection is used rather than SMO's `Information` event so the paths that execute the script directly (`-ExecuteAs`, `-KeepCDC`, `-ErrorBrokerConversations`) are covered as well.
- after a successful log restore without 4329 the stop point is reached: the remaining backups of that database are skipped with a verbose message each, and no output object is produced for them.
- if backups were skipped and the last one would have recovered the database, the command recovers it right there, `WITH RECOVERY` or `WITH STANDBY` when `-StandbyDirectory` was given, so the database ends in the state it would have reached had the mark been in the last file. The output object of the file that reached the mark reports `NoRecovery = False` and carries the recovery statement in its `Script`. With `-NoRecovery` the database is left restoring, as requested, and the tail is still skipped.

`-OutputScriptOnly` and `-VerifyOnly` are unchanged: nothing executes there, so nothing can be detected, and the scripts list every file as before.

## Tests

In the StopMark context of `Restore-DbaDatabase.Tests.ps1`, whose fixture has the mark in the first of two logs:

- the two existing tests no longer suppress warnings and no longer call `-Recover`; they assert no warning, two output objects (the full and the log with the mark, the last one with `NoRecovery = False`), the database online, and the expected data;
- new: the same restore under `-EnableException` does not throw;
- new: with `-NoRecovery` the database is left restoring, the second log is still skipped, and a `-Recover` afterwards works as before.

The StopAfterDate test (mark in the last file) and the StopAtLsn test are unchanged. Through the testing-dbatools harness: green on both editions with the fix; red with development's `Invoke-DbaAdvancedRestore`, where the two existing tests fail on the warning and the database state and the new EnableException test throws.


potatoqualitee (changes requested):
`public/Invoke-DbaAdvancedRestore.ps1:579-589` recovers an early stop-mark restore with a standalone `RESTORE DATABASE ... WITH RECOVERY`, but does not carry the requested `KEEP_CDC` or `ERROR_BROKER_CONVERSATIONS` options into that final recovery. For the marked log, `$restore.NoRecovery` is true, so the existing option-appending branch at lines 507-522 is skipped; when later logs are then skipped, this new recovery statement becomes the only recovery operation. A `-StopMark -KeepCDC` restore can therefore lose CDC configuration despite the caller's explicit request, and `-ErrorBrokerConversations` is likewise ignored.

Please include the applicable recovery-only options in the generated recovery statement and add coverage for an early stop mark combined with these switches, verifying the resulting CDC/Broker state.

andreasjordan:
Confirmed and fixed in 95fc16b1e: the recovery statement issued after a reached stop mark is the one that recovers, and it carried none of the recovery-only options, because the option branch at 507-522 only decorates a restore whose `NoRecovery` is false and the marked log is restored with `NORECOVERY`.

**What changed:** `KEEP_CDC` and `ERROR_BROKER_CONVERSATIONS` are appended to that statement when the switches are set, for the `RECOVERY` and the `STANDBY` form alike.

**Tests:** the stop-mark fixture now enables change data capture on the database and the `steps` table before the chain is taken, so the chain carries CDC metadata; `AfterAll` disables CDC before dropping, so the capture and cleanup jobs do not linger. New test: the chain restored under another name with `-StopMark -KeepCDC -ErrorBrokerConversations` stops at the first mark (max step 2), keeps `is_cdc_enabled` and the capture instance in `cdc.change_tables`, and comes up with `is_broker_enabled` on. Restoring under another name is what makes the assertions meaningful: without the options SQL Server drops the CDC metadata and leaves the broker disabled because the source database with the same broker id still exists. Through the testing-dbatools harness: the whole Restore-DbaDatabase file is green on both editions (85 passed, 3 skipped); against the reviewed head, the new test fails at the CDC assertion.

potatoqualitee (changes requested):
`public/Invoke-DbaAdvancedRestore.ps1:584-586` constructs the new standalone recovery statement as `RESTORE DATABASE [$database]` without escaping closing brackets in the identifier. `-DatabaseName` accepts unrestricted names, so restoring an early-stop-mark chain as a valid name such as `db]copy` generates `RESTORE DATABASE [db]copy] WITH RECOVERY` instead of `RESTORE DATABASE [db]]copy] WITH RECOVERY`.

That recovery statement fails after the marked log was restored with `NORECOVERY`; the command warns or throws under `-EnableException`, and the database remains in `RESTORING`, which materially reintroduces the failure mode this PR is intended to fix. The STANDBY branch has the same identifier construction.

Please SQL-quote the database identifier (at minimum, double embedded `]` characters) for both RECOVERY and STANDBY statements, and add an early-stop-mark regression test using a destination database name containing `]`.

andreasjordan:
Confirmed and fixed in 4ada74635: the recovery statement after a reached stop mark took the name between brackets as it was, so a destination name like `db]copy` produced `RESTORE DATABASE [db]copy] WITH RECOVERY`, which SQL Server refuses with "Incorrect syntax near ']'", and the database stayed restoring behind the warning.

**What changed:** the closing brackets in the name are doubled, the way SMO quotes it for the scripted restores, for the `RECOVERY` and the `STANDBY` form alike.

**Tests:** a new test in the stop-mark context restores the chain under `<source>]copy` with `-StopMark` and asserts no warning, two results with the last one recovered, status `Normal` and max step 2. The cleanup covers the third database and skips `sp_cdc_disable_db` for a copy that came up without CDC. Through the testing-dbatools harness the whole Restore-DbaDatabase file is green on both editions (86 passed, 3 skipped). Against the reviewed head the same scenario fails with "Incorrect syntax near ']'" and leaves the copy restoring, probed with that head's function body because the file takes ten minutes per run.

potatoqualitee (approved):
No material defects found.

potatoqualitee:
ty

PR #10697, opened 2026-09-12
Get-DbaDbObjectTrigger - Stop eating the caller loop on an input object that is not a table or view

andreasjordan:
Part of #10638, the sweep for `Stop-Function -Continue` outside an enclosing loop.

## Problem

`Get-DbaDbObjectTrigger` checks piped objects inside a `ForEach-Object` scriptblock and rejects anything that is not a table or view with `Stop-Function -Continue` ([Get-DbaDbObjectTrigger.ps1:119](https://github.com/dataplat/dbatools/blob/development/public/Get-DbaDbObjectTrigger.ps1#L119)). A scriptblock is not a loop of the command, so without `-EnableException` the `continue` unwinds out of the command and consumes an iteration of whatever loop the caller runs in; the `return` after it never ran. Proven in the lab with the plain-script recipe: three calls piping a string in a `foreach` left the loop counter at 0 on development.

## What changed

The check is a `foreach` loop now, so the `-Continue` has a loop to continue and skips the bad object inside the command. A bad object that stays in the input has no triggers and produces no output, as before. Nothing else changes.

## Tests

New context `When something other than a table or view is piped in`: three calls in a loop, asserting the loop count and the warning. Through the testing-dbatools harness: green on both editions with the fix (29 passed); the plain-script proof above is the red-on-old.


potatoqualitee (approved):
Invalid pipeline objects now exit only the current command invocation rather than consuming the caller loop, while table and view handling remains unchanged. The relevant integration coverage and all applicable exact-head checks pass.

potatoqualitee:
straightforward 👍🏼

PR #10698, opened 2026-09-12
Stop-DbaExternalProcess - Stop eating the caller loop when the process cannot be stopped

andreasjordan:

## Problem

`Stop-DbaExternalProcess` reports a failed kill from the catch at the end of its process block with `Stop-Function -Continue` ([public/Stop-DbaExternalProcess.ps1:98](https://github.com/dataplat/dbatools/blob/development/public/Stop-DbaExternalProcess.ps1#L98)). No loop encloses that site inside the command, so without `-EnableException` the `continue` unwinds out of the command and consumes an iteration of whatever loop the caller runs in. Proven in the lab with the plain-script recipe: three calls in a `foreach` left the loop counter at 0 on development and at 3 with the fix.

## What changed

The `-Continue` is dropped; the catch is the last statement of the process block, so nothing else is needed.

## Tests

New context `When the process does not exist`: three calls with a process id that does not exist on localhost, asserting the loop count and the `Error killing` warning. Through the testing-dbatools harness: green on both editions with the fix; the plain-script proof above is the red-on-old.


potatoqualitee (approved):
The process-stop failure now returns from the command invocation instead of consuming the caller loop, with successful process termination unchanged. The relevant Windows PowerShell coverage and all applicable exact-head checks pass.

PR #10699, opened 2026-09-12
Get-DbaWaitResource - Stop eating the caller loop on connection failure

andreasjordan:

## Problem

`Get-DbaWaitResource` reports a failed connection from the catch in its process block with `Stop-Function -Continue` ([public/Get-DbaWaitResource.ps1:114](https://github.com/dataplat/dbatools/blob/development/public/Get-DbaWaitResource.ps1#L114)). No loop encloses that site inside the command, so without `-EnableException` the `continue` unwinds out of the command and consumes an iteration of whatever loop the caller runs in. Proven in the lab with the plain-script recipe: three calls in a `foreach` left the loop counter at 0 on development and at 3 with the fix.

## What changed

## Tests

New context `When the instance cannot be reached`: lowers `sql.connection.timeout` to 2 for its duration, three calls against a host that does not exist, asserting the loop count and the warning, the shape of #10653. Through the testing-dbatools harness: green on both editions with the fix; the plain-script proof above is the red-on-old.


potatoqualitee (approved):
The connection-failure path now exits the command invocation without consuming the caller loop, while successful wait-resource retrieval is unchanged. The targeted Windows PowerShell coverage and exact-head CI checks pass.

potatoqualitee:
looks good 👍🏼

PR #10700, opened 2026-09-12
New-DbaAzAccessToken - Stop eating the caller loop when the token request fails

andreasjordan:

## Problem

`New-DbaAzAccessToken` reports a failed token request from the catch at the end of its process block with `Stop-Function -Continue` ([public/New-DbaAzAccessToken.ps1:257](https://github.com/dataplat/dbatools/blob/development/public/New-DbaAzAccessToken.ps1#L257)). No loop encloses that site inside the command, so without `-EnableException` the `continue` unwinds out of the command and consumes an iteration of whatever loop the caller runs in. Proven in the lab on Windows PowerShell with the plain-script recipe: three `ServicePrincipal` calls with a tenant that does not exist left the loop counter at 0 on development. The catch is reachable there because the ADAL type the path uses is not present, which throws; on PowerShell Core the path refuses to run before any request is made, so the catch cannot be reached that way.

## What changed

The `-Continue` is dropped; the catch is the last statement of the process block, so nothing else is needed.

## Tests

The test file had no integration tests. New context `When the service principal cannot be authenticated`: three calls in a loop, asserting the loop count and the `Failure` warning, skipped on Core with a comment saying why. Through the testing-dbatools harness: green on Windows PowerShell with the fix; the plain-script proof above is the red-on-old.


potatoqualitee (approved):
The token-request failure now returns from the command invocation instead of consuming the caller loop, without changing successful token acquisition. The targeted Windows PowerShell coverage and all applicable exact-head checks pass.

potatoqualitee:
thank you 👍🏼

PR #10702, opened 2026-09-12
Export-DbaDacPackage - Stop eating the caller loop when only ExcludeDatabase is given

andreasjordan:

## Problem

`Export-DbaDacPackage` rejects a call that names neither `-Database` nor `-AllUserDatabases` (only `-ExcludeDatabase` passes the earlier check) with `Stop-Function -Continue` ([public/Export-DbaDacPackage.ps1:168](https://github.com/dataplat/dbatools/blob/development/public/Export-DbaDacPackage.ps1#L168)). No loop encloses that site inside the command, so without `-EnableException` the `continue` unwinds out of the command and consumes an iteration of whatever loop the caller runs in. Proven in the lab with the plain-script recipe: three calls in a `foreach` left the loop counter at 0 on development and at 3 with the fix.

## What changed

The `-Continue` is dropped and an explicit `return` ends the block for that input, as in #10637. Nothing else changes. The check still fires before any connection is made.

## Tests

New context `When only ExcludeDatabase is given`: three calls in a loop, asserting the loop count and the warning. Through the testing-dbatools harness: green on both editions with the fix; the plain-script proof above is the red-on-old.


potatoqualitee (approved):
The ExcludeDatabase-only validation path now exits the command invocation without terminating its caller loop, while normal export behavior is preserved. Targeted Windows PowerShell coverage and all applicable exact-head checks pass.

PR #10703, opened 2026-09-12
Expand-DbaDbLogFile - Stop eating the caller loop when the log file processing fails

andreasjordan:

## Problem

`Expand-DbaDbLogFile` reports a failed run from the catch that wraps its whole process block with `Stop-Function -Continue` ([public/Expand-DbaDbLogFile.ps1:689](https://github.com/dataplat/dbatools/blob/development/public/Expand-DbaDbLogFile.ps1#L689)). No loop encloses that site inside the command, so without `-EnableException` the `continue` unwinds out of the command and consumes an iteration of whatever loop the caller runs in. No lab trigger was found for that outer catch: every failure inside it is caught by the per-database loop, so the site is latent as far as the lab can tell, and the fix mirrors the pattern of #10637.

## What changed

The `-Continue` is dropped; the catch is the last statement of the process block. The message of that catch also read a variable that does not exist, `$_InvocationInfo`, so the line number was always empty; it is `$_.InvocationInfo` now.

## Tests

No new test: no trigger reaches the outer catch. The existing test file passes on both editions through the testing-dbatools harness.


andreasjordan:
The SINGLE 1/5 failure is the same kind of finding as on #10701: a test that had never really run.

**What failed.** The mocked unit test "Uses a smaller final growth when that keeps VLFs within TargetVlfCount" expects one result and got none. On `development` the catch at the end of the process block ran `Stop-Function -Continue` with no enclosing loop, so whenever the mocked run hit that catch, the `continue` unwound out of the command and out of the `It` block before its assertions, and Pester recorded a pass. With the escape removed, the assertions run and the test shows that the mocked run always ended in that catch.

**Why the mocked run failed, three gaps in a row** (found by asserting on the collected warnings first, so Pester prints them):

1. `$Databases.Count` was `$null` for the mock's single filtered database, a PSCustomObject rather than a collection, so the progress bar divided by zero. In production the SMO collection always has a Count. The three filter assignments now wrap their result in `@()`, which is harmless for SMO and makes the count reliable for a single database; that is the only change to the command in this commit.
2. `Mock Measure-DbaDbVirtualLogFile` keeps the real `[DbaInstanceParameter[]]` binding, which rejects the mocked server object. The test now defines a plain function instead, the way it already does for `Write-Message`, `Resolve-DbaComputerName` and `Select-DefaultView`.
3. The `Select-DefaultView` stub had no `-ExcludeProperty`, which the command passes. Added.

The lab runs behind this PR used the IntegrationTests tag only, which is why the unit test was not caught before CI; the full file now runs green on both editions here: 5/5 on PowerShell 7 and 5/5 on Windows PowerShell 5.1, unit and integration tags together.

potatoqualitee (approved):
The log-file failure path now returns from the helper scope instead of terminating the caller loop, preserving successful expansion behavior and multi-input processing. All applicable exact-head CI checks pass.

potatoqualitee:
tks 💯

PR #10704, opened 2026-09-12
Get-LoginPasswordHash - Stop eating the caller loop when the hash query fails

andreasjordan:

## Problem

`Get-LoginPasswordHash` reports a failed hash query from its inner catch, followed by a `return` that never ran with `Stop-Function -Continue` ([private/functions/Get-LoginPasswordHash.ps1:71](https://github.com/dataplat/dbatools/blob/development/private/functions/Get-LoginPasswordHash.ps1#L71)). No loop encloses that site inside the command, so without `-EnableException` the `continue` unwinds out of the command and consumes an iteration of whatever loop the caller runs in. The only caller, `Sync-DbaLoginPassword`, calls it inside its login loop, so the escape was caught there - but it bypassed the caller's own handling of a missing hash and the `return` after the call. Latent as far as a user is concerned.

## What changed

The `-Continue` is dropped; the `return` that follows now runs as written.

## Tests

No new test: no red-on-old exists for a latent escape. The caller's test file passes on both editions through the testing-dbatools harness.


potatoqualitee (approved):
The failure path now returns from the helper scope rather than terminating the caller loop, while successful hash retrieval remains unchanged. The control-flow reproduction completes all caller iterations and the current CI checks are green.

potatoqualitee:
looks good 💯

PR #10705, opened 2026-09-12
Convert-DbaIndexToTable - Stop eating the caller loop when no instance is given

andreasjordan:

## Problem

`Convert-DbaIndexToTable` rejects a missing `-SqlInstance` in its begin block with `Stop-Function -Continue` ([private/functions/Convert-DbaIndexToTable.ps1:18](https://github.com/dataplat/dbatools/blob/development/private/functions/Convert-DbaIndexToTable.ps1#L18)). No loop encloses that site inside the command, so without `-EnableException` the `continue` unwinds out of the command and consumes an iteration of whatever loop the caller runs in. The only caller, `Invoke-DbaDbDataMasking`, always passes `SqlInstance` and calls it inside its table loop, so the site cannot fire and the continue would be caught. Latent.

## What changed

## Tests

No new test: no red-on-old exists for a latent escape. The caller's test file passes on both editions through the testing-dbatools harness.


potatoqualitee (approved):
The missing-instance path now exits only the helper invocation instead of terminating the caller loop, with no change to successful conversions. The relevant command coverage passes; the failing COPY lane is unrelated Remove-DbaDatabaseSafely coverage.

PR #10706, opened 2026-09-12
Invoke-DbaDbDecryptObject - Stop eating the caller loop when the export folder cannot be created

andreasjordan:

## Problem

`Invoke-DbaDbDecryptObject` guards the creation of `-ExportDestination` in its begin block with `Stop-Function -Continue` ([public/Invoke-DbaDbDecryptObject.ps1:214](https://github.com/dataplat/dbatools/blob/development/public/Invoke-DbaDbDecryptObject.ps1#L214)). No loop encloses that site inside the command, so without `-EnableException` the `continue` unwinds out of the command and consumes an iteration of whatever loop the caller runs in. The guard itself turned out to be dead: a failing `New-Item` is not a terminating error, so the catch never ran, a folder that could not be created was silently ignored, and the export failed later. In the lab, three calls with a destination on a drive letter that does not exist left the loop counter at 3 on development - because the catch never fired - and wrote no warning. (A folder under a file is no trigger at all: `New-Item -Force` returns quietly for it on both editions, even with `-ErrorAction Stop`.)

## What changed

The `-Continue` is dropped and an explicit `return` ends the begin block, as in #10637; the process block is already guarded by `Test-FunctionInterrupt`. `New-Item` gets `-ErrorAction Stop`, so the guard fires as designed.

## Tests

New context `When the export destination cannot be created`: three calls with a destination on a drive letter that is free at test time, asserting the loop count and the `Couldn't create destination folder` warning. Through the testing-dbatools harness: green on both editions with the fix; on development the test fails on the missing warning, since the guard never fired there.


potatoqualitee (approved):
The export-folder failure now returns from the helper scope rather than terminating the caller loop. Successful decryption/export behavior is unchanged, and all applicable exact-head CI checks pass.

potatoqualitee:
nice and simple 👍🏼

Issue #10655, opened 2026-08-30
Command invariants - The structural rules behind #10638 and friends, sized by AST sweeps


andreasjordan:
The same structural defect classes keep resurfacing one command at a time — the `Stop-Function -Continue` escape (#10638) alone is 91 sites and ~20 pull requests so far. This issue collects the rules in one place, sizes each with a fresh AST sweep over all of `public/` and `private/functions/` (run 2026-08-30), and proposes making the enforceable ones a structural test so the classes cannot come back.

### 1. `Stop-Function -Continue` requires an enclosing loop

*91 sites found, 73 resolved, 18 remaining — tracked as #10638.*

Without a loop in the same function body, the `continue` unwinds out of the command and silently consumes an iteration of the **caller's** loop. `-SilentlyContinue` escapes even under EnableException. The fix wave is under way (#10636–#10653); the rule is proven AST-checkable.

### 2. `Stop-Function` in stop mode must be followed by `return`

*48 candidate sites in 26 files (upper bound, untriaged).*

Stop-Function's own help says the caller should follow with `return` — without EnableException the call returns normally and the code runs on with invalid state (seen live at `Invoke-DbaDbLogShipping:707`, fixed in #10644). The sweep counts sites where a stop-mode call sits mid-block with no `return`/`throw`/`continue`/`break` after it and is not the last statement of its block. Some will be intentional warn-and-fall-through, so enforcement needs a one-time triage into fixes plus an allowlist baseline.

### 3. A `begin` block that can stop needs the interrupt guard in `process`

*146 commands can stop in begin — 133 guarded (91 %), 13 missing.*

When a begin-block `Stop-Function` returns normally (the non-EnableException path), `process` still runs — against null or half-built state (found live in #10652). The established convention is `if (Test-FunctionInterrupt) { return }` as the **first statement** of process, and the codebase already follows it 133 times, always in first position. The holdouts:

- [ ] Export-DbaExecutionPlan
- [ ] Get-DbaDbStoredProcedure
- [ ] Import-DbaCsv
- [ ] Import-DbaParquet
- [ ] Invoke-DbaDbAzSqlTip
- [ ] Measure-DbaDiskSpaceRequirement
- [ ] Remove-DbaXESession
- [ ] Reset-DbaAdmin
- [ ] Start-DbaXESession
- [ ] Stop-DbaXESession
- [ ] Test-DbaDiskAlignment
- [ ] Get-BackupAncientHistory (private)

[...]

PR #10707, opened 2026-09-12
Command invariants - Add the missing Test-FunctionInterrupt guards and fix the Test-Bound typo in Export-DbaExecutionPlan

andreasjordan:
Part of #10655, rule 3: a `begin` block that can call `Stop-Function` in stop mode needs `if (Test-FunctionInterrupt) { return }` as the first statement of `process`, otherwise the non-EnableException path warns and then runs the process block against the state the begin block refused to build.

The issue lists twelve commands. A per-command triage with the AST shows that only four of them actually have a stop-mode `Stop-Function` directly in their begin block. The other eight have the call inside a helper function defined in `begin`, or use `-Continue`. In both cases the interrupt flag that `Test-FunctionInterrupt` reads is never set in the command's scope (Stop-Function sets it in *its caller's* scope, and the `-Continue` path returns before setting it), so a guard would be dead code there. This PR fixes the four and documents the eight, so the issue's checklist can be closed in one go.

## What was wrong, per command

| Command | What the begin block stops on | What the process block did anyway |
| --- | --- | --- |
| Get-DbaDbStoredProcedure | `-Name` that does not parse | returned every procedure of the database, as if no `-Name` had been given |
| Invoke-DbaDbAzSqlTip | no `-Database`, `-ExcludeDatabase` or `-AllUserDatabases` | connected to every instance and warned once more that it is not an Azure SQL Database |
| Import-DbaParquet | Parquet.NET missing or unloadable | connected to the instance and warned "Failed to open Parquet file" once per file |
| Get-BackupAncientHistory (private) | connection failure | called methods on the missing server object |

Import-DbaParquet has a second way out of `begin`: when `Get-DbaParquetPath` finds nothing it warns (or throws) in its own scope and the command returns from `begin` without any flag set. A `$parquetLoaded` variable now records whether the assembly is available, and `process` checks both that and `Test-FunctionInterrupt`.

## Not changed, and why

These eight from the issue's checklist keep their process block as it is. The only `Stop-Function` calls in their begin blocks are:

- Export-DbaExecutionPlan, Remove-DbaXESession, Start-DbaXESession, Stop-DbaXESession, Import-DbaCsv, Measure-DbaDiskSpaceRequirement: `-Continue` calls inside helper functions (`Export-Plan`, `Remove-XESessions`, `New-SqlTable`, `Get-MountPointFromPath`, ...).
- Reset-DbaAdmin: inside the boolean helper `Invoke-ResetSqlCmd`, which has its own `finally` and returns `$false`.
- Test-DbaDiskAlignment: inside the helper `Get-DiskAlignment`, each followed by its own `return`.

A guard in their process blocks would never fire. The future structural test for rule 3 should therefore skip `Stop-Function` calls nested in a `FunctionDefinitionAst` or a `ScriptBlockExpressionAst`, and calls with `-Continue`/`-SilentlyContinue`; with that, the codebase is at 100 % for rule 3 after this PR.

## Found along the way

The issue already notes it: Export-DbaExecutionPlan's process block read `Test-Bound -ParamterName Path`. `Test-Bound` is a plain function, so the misspelled parameter went into `$args` and the check was always false. Consequences: a file passed as `-Path` was never rejected, and the default export directory was never created when missing. Fixed to `-ParameterName`, with a test.

## Tests

Each fix has a regression test that is red on `development` and green with the fix:

- Get-DbaDbStoredProcedure: `-Name "a.b.c.d"` returns nothing and warns; throws with `-EnableException`.
- Invoke-DbaDbAzSqlTip: no database selection gives exactly one warning and no connection; throws with `-EnableException`. This is the first test of that file that runs at all: the existing Describe is skipped unless `$env:azuredbpasswd` equals a fixed string.
- Import-DbaParquet: Parquet.NET stays loaded once any test has used it, so the missing-assembly path is exercised in a child process of the same PowerShell edition, with both Parquet search paths pointed at an empty folder. It must produce exactly one warning, the one that names `Install-DbaParquet`.
- Get-BackupAncientHistory: new test file; an unreachable instance warns and returns nothing instead of throwing. The command name does not match the `*-Dba*` default for the warning variable, so the test passes it explicitly.
- Export-DbaExecutionPlan: a file as `-Path` warns "must be a directory" and exports nothing.

Lab results, both Windows PowerShell 5.1 and PowerShell 7: all five files green (Export-DbaExecutionPlan 2/2, Get-BackupAncientHistory 2/2, Get-DbaDbStoredProcedure 15/15, Import-DbaParquet 18/18, Invoke-DbaDbAzSqlTip 3 passed, 3 skipped). With the five source files swapped back to `development` and the new tests kept, exactly one test fails per file, the new one.


potatoqualitee (changes requested):
`public/Export-DbaExecutionPlan.ps1:190` still gates the path bootstrap on `Test-Bound -ParameterName Path`. A parameter's declared default is not present in `$PSBoundParameters`, so omitting `-Path` makes this condition false even when the configured `Path.DbatoolsExport` directory does not exist. On a fresh profile, the subsequent XML `Save()` calls therefore target a missing directory and export nothing; the new test always binds `Path`, so it does not cover the default invocation.

Please validate/create the resolved `$Path` regardless of whether the caller explicitly bound it, while continuing to reject a caller-supplied file path, and add a regression test that omits `-Path` with the configured default directory absent.

andreasjordan:
Confirmed and fixed in the commits up to 7b7fc492e: `Test-Bound` does not see a parameter default, so with `-Path` omitted the bootstrap never ran and, on a profile whose configured export directory does not exist yet, the `Save()` calls wrote into a missing directory.

**What changed:** the resolved `$Path` is validated whether or not the caller bound it. The configured `Path.DbatoolsExport` directory is created when missing, as before for the bound case; anything else still has to be an existing directory, so a caller-supplied file path is still rejected with the same message. This also covers the piped parameter set, where `$Path` carries its default too.

Writing the test surfaced two more defects on the same path, fixed in the same PR:

- **Named instances never got a file.** The plan file name starts with `$object.SqlInstance`, backslash included, so for `SQL03\SQL2019` every `Save()` tried `<Path>\SQL03\SQL2019-tempdb-…sqlplan` and failed with "Could not find a part of the path"; the command still reported the objects and warned "Skipped … because it is null" for each. The characters a file name cannot carry are now replaced by a hyphen.
- **Setting `Path.DbatoolsExport` moved the dbatools message log.** Its config handler in `private/configurations/settings/paths.ps1` was a copy of the `Path.DbatoolsLogPath` handler and set `LogHost.LoggingPath`. The test noticed because its export directory could not be removed: the log file was open in it. The handler is gone; the log path setting keeps its own.

**Tests:** new context that points `Path.DbatoolsExport` at a directory that does not exist, caches an ad hoc plan in tempdb, and exports with `-Path` omitted: the directory exists afterwards, the results carry `OutputFile` under it and `.sqlplan` files are there. The config value is restored in `AfterAll`. Through the testing-dbatools harness: the file is green on both editions (3/3, plan files present); against the reviewed head, the new test fails because the directory is never created, and no export directory is left behind any more.

potatoqualitee (approved):
The current head resolves my previous requested change. No material defects remain. Reviewed at head 7b7fc492ec55ea55ceb82907948e323662c04a94.

potatoqualitee:
thank you 🙏🏼

PR #10708, opened 2026-09-12
Dead parameters - Wire up seven parameters that were declared and documented but never read

andreasjordan:
Second wave for #10607, the dead-parameter report by @greenmtnsun. A fresh AST scan of `public/` (parameters with no variable or string reference outside the param block, `EnableException` excluded, commands that forward `$PSBoundParameters` wholesale checked by hand) confirms the ~21 remaining candidates. This PR wires up the seven whose intent is clear and whose effect can be tested; the rest are removal or design decisions and are listed on the issue.

Grouped into one PR at the maintainers' request; each change is a few lines in one command, with its own regression test.

## What each parameter now does

| Command | Parameter | Before | Now |
| --- | --- | --- | --- |
| Copy-DbaAgentJobCategory | `-AlertCategory` (alias `AgentCategory`) | declared as `-AgentCategory`, but the dispatch filtered on `$AlertCategory`, so the name filter never applied and every alert category was copied | renamed to match the code and the help, old name kept as an alias |
| Export-DbaReplServerSetting | `-NoClobber` | never checked, an existing file was appended to | stops with a warning before anything is scripted; `-Passthru` is unaffected |
| Get-DbaDbLogShipError | `-ExcludeDatabase` | never applied | filters the results like `-Database` does |
| Import-DbaPfDataCollectorSetTemplate | `-Subdirectory` | never written into the template | set on the XML document. The templates spread the empty `Subdirectory` element over two lines, so the existing line-based text replacement could not have filled it. Found next to it: `Set-Variable -Name RootName` where `RootPath` was meant, so the documented default root path was never applied either; fixed |
| Invoke-DbaXEReplay | `-Database` | never handed to sqlcmd, every replay ran in the login's default database | passed as `-d`; with several databases the file is replayed once per database. The four sqlcmd calls are now one argument list |
| New-DbaAgentAlert | `-Disabled` | never applied, every alert came up enabled | sets `IsEnabled = $false` before `Create()` |
| Remove-DbaDatabaseSafely | `-ReuseSourceFolderStructure` | never forwarded | forwarded to the `Restore-DbaDatabase -OutputScriptOnly` call that builds the agent job step |

## Tests

One regression test per command, red on `development` and green here, except where noted:

- Get-DbaDbLogShipError: an instance without log shipping has nothing to filter, so the test only proves the parameter is accepted and harmless. A red-on-old test would need a log shipping setup with errors.
- Remove-DbaDatabaseSafely: no new test. Proving the forwarded switch needs a destination whose folder layout matches the source, and the existing tests already run the full backup, job and restore cycle on both lab instances; they stay green.
- Invoke-DbaXEReplay: first integration test of that file, skipped where `sqlcmd` is not on the path.

Lab results, both PowerShell 7 and Windows PowerShell 5.1: all seven files green (Copy-DbaAgentJobCategory 5/5, Export-DbaReplServerSetting 2/2, Get-DbaDbLogShipError 3/3, Import-DbaPfDataCollectorSetTemplate 4/4, Invoke-DbaXEReplay 2/2, New-DbaAgentAlert 4/4, Remove-DbaDatabaseSafely 3/3). With the seven source files swapped back to `development` and the new tests kept, the five new regression tests fail plus the parameter-validation test of Copy-DbaAgentJobCategory, and nothing else.


potatoqualitee (changes requested):
`public/Import-DbaPfDataCollectorSetTemplate.ps1:295` assigns `$xml.DataCollectorSet.Subdirectory` directly, but that property exists only when the template already contains a `<Subdirectory>` element. Four of the five bundled templates (all PAL templates) omit it, so using the newly wired `-Subdirectory` parameter with those templates throws `The property 'Subdirectory' cannot be found on this object` and skips the import. The regression test uses `Long Running Queries`, the only bundled template that already has the node.

Please create and insert the XML element when it is absent before setting its value, and add coverage with one of the bundled PAL templates.

andreasjordan:
Confirmed and fixed in d050e4e0d: `Long Running Queries` is the only bundled template with a `Subdirectory` element; the four PAL templates have none, so the assignment threw for them and the import was skipped whenever `-Subdirectory` was used.

**What changed:** the element is looked up on the document and created when missing, inserted before `SubdirectoryFormat` where the collector-set schema keeps it (appended when that element is missing too), then set and saved as before.

**Tests:** new context that imports `PAL - SQL Server 2014 and Up` with `-Subdirectory` and checks the imported set carries it, with its own cleanup of that collector set. Through the testing-dbatools harness: the file is green on both editions (5/5); against the reviewed head, the new test fails with exactly the message from the review, "The property Subdirectory cannot be found on this object".

potatoqualitee (approved):
The current head resolves my previous requested change, including regression coverage for templates without a Subdirectory element. No material defects remain. Reviewed at head d050e4e0dd96fad59365194a54a99b229bace79e.

potatoqualitee:
great changes~

PR #10709, opened 2026-09-12
Enable-DbaAgHadr, Disable-DbaAgHadr - Report a refused service restart instead of ignoring it

andreasjordan:
Found through the COPY lane failure on #10705, which touches nothing near restores: `Remove-DbaDatabaseSafely` failed three times with "SQL Server Agent is not running" on sql2022, while the green development run of the same day ran the identical file list.

## Mechanism

The prep phase of the job explains it. `tests/appveyor.sql2022.ps1` runs `Enable-DbaAgHadr -Force`, which does `Stop-DbaService -Type Agent, Engine` and then `Start-DbaService`. This time the Agent went to StopPending and came back (Update-ServiceStatus: "The service failed to stop"), so the engine stop was refused ("other services that are running are dependent on it"), the Agent finished stopping on its own afterwards, and nothing started it again. Update-ServiceStatus reports all of that only as `Status = "Failed"` on the returned objects plus warnings; `-EnableException` does not turn it into an exception. Enable-DbaAgHadr discarded the objects, so it returned normally, the setup stage passed, and the failure surfaced six minutes later in an unrelated test. `tests/appveyor.sql2019.ps1` has the same line, so the HADR lane is exposed the same way.

## Changes

**Enable-DbaAgHadr and Disable-DbaAgHadr** (same block in both): the stop and start results are kept and any object whose Status is not Successful turns the restart into a failure, reported through the existing `Stop-Function "Issue restarting ..."` with the service messages as the error record. Without `-EnableException` that is a warning, with it an exception. Nothing changes when the restart works.

**tests/appveyor.sql2019.ps1 and tests/appveyor.sql2022.ps1**: the HADR step is wrapped so that a failed restart is handled instead of ignored: wait for a StopPending service to settle, start whatever is not running, and run `Enable-DbaAgHadr -Force` once more. The two `Start-Service` lines already in the scripts run before the HADR step and could not help.

## Tests

The lab cannot provoke a refused service stop on demand, so the restart-failure path has no red-on-old test; the happy path is covered by the existing tests, which restart the HADR instance: Enable-DbaAgHadr 2/2 and Disable-DbaAgHadr 2/2 on both PowerShell 7 and Windows PowerShell 5.1. The setup scripts only run on the self-hosted runners; this PR's own ci-azure run is their test.


potatoqualitee (dismissed):
The refusal path now surfaces the failed service restart instead of continuing as though HADR had been enabled or disabled. The surrounding multi-input flow and cleanup behavior remain intact, and the relevant HADR/restart CI lanes pass.

potatoqualitee:
one thing is.. if agent is already stopped, this starts it (old bug)

potatoqualitee (changes requested):
pls detect if agent is started before restarting

andreasjordan:
Confirmed and fixed in f9626ffdf: the forced restart stopped and started Agent and Engine unconditionally, so an Agent that had been stopped on purpose came back running.

**What changed (both commands):** the Agent state is read with `Get-DbaService` before the restart. The Agent still stops together with the engine it depends on (stopping an already stopped service reports "no action required", so that is harmless), but it is started again only if it had been running before. The `ShouldProcess` message says so.

**Tests:** both files gain "Leaves a stopped Agent stopped when -Force restarts the engine": stop the Agent, run the command with `-Force`, assert HADR changed and the Agent is still `Stopped`, then start the Agent again; `AfterAll` starts it in any case. On the reviewed head the old code started both services, so the new assertion fails there by construction. Lab through the testing-dbatools harness on SQL04\SQL2025: Enable and Disable green on pwsh and 5.1 (3/3 each), instance left with HADR on and both services running.

potatoqualitee (dismissed):
The current implementation preserves a stopped Agent, but the tests never assert the complementary branch that a running Agent remains running after -Force. Both existing success tests only assert IsHadrEnabled; if Agent state detection regresses false, the commands can leave a previously running Agent stopped while every current assertion still passes. This is changed behavior and the operational failure behind the PR, so the repository test policy requires focused real-boundary coverage. Arrange and assert a running Agent before the happy-path call, then assert it is still running afterward in both Enable-DbaAgHadr.Tests.ps1 and Disable-DbaAgHadr.Tests.ps1.

andreasjordan:
Agreed, the happy path only asserted `IsHadrEnabled`, so a detection regression that never starts the Agent again would have passed every test. Fixed in 5233c107d.

**What changed (both test files):** the happy-path test is now "Successfully enables/disables HADR and keeps a running Agent running". It starts the Agent if it is not running (collected first, because `Start-DbaService` with nothing piped in falls back to the local machine), asserts the state is `Running`, runs the command with `-Force`, asserts the HADR change and then asserts the Agent is `Running` again. Together with the existing "Leaves a stopped Agent stopped" test both branches of the state detection are covered.

**Red on the regression:** with `$agentWasRunning` forced to `$false` in both commands (the regression you describe), exactly the new assertion fails in both files: "Expected: 'Running' But was: 'Stopped'", the other tests still pass and `AfterAll` brings the Agent back. Command files restored afterwards, only the tests changed.

Through the testing-dbatools harness against `SQL04\SQL2025`: both files green on PowerShell 5.1 and pwsh 7.6 (3/3 each), instance left with HADR on and engine and Agent running.

Created by Claude and reviewed by Andreas Jordan

Issue #10658, opened 2026-08-30
Extended Events commands leave one reconnected store connection checked out per pipeline


andreasjordan:
Deferred from PR #10635, whose description names it "deliberately left for a follow-up" — recording it so the follow-up has a tracker.

## Behavior

PR #10635 closes the cloned connection behind the `XEStore` that `Get-DbaXESession` creates, which removed the bulk of the Extended Events connection leak. A smaller, separate mechanism remains: the emitted session objects keep a functional `Store` note property, and when a downstream command uses that store (piped `Start-DbaXESession`, `Stop-DbaXESession`, ...), SFC transparently **reconnects** it from the pool — and that reopened connection stays checked out, once per pipeline, until the process exits.

Measured in the lab (SQL03\SQL2019, isolated runner measurements of the family test files, after #10635): `Start-DbaXESession` still leaves 7 sleeping sessions (down from 15), `Remove-DbaXESession` 2 (from 6), `ConvertTo-DbaXESession` 3 (from 7). The residue is this reconnect mechanism.

## Fix shape

Consumer-side disconnects: each command that operates through a piped object's store closes the store connection when it reopened it — several commands (`Start-`, `Stop-`, `Remove-`, `ConvertTo-DbaXESession`, the `Watch`/`Export` family), each with its own before/after measurement, following the counting-test pattern established in #10633–#10635. Needs care that a store used repeatedly within one command is not churned open/closed per operation.

Related: the sleeping-connection inventory that found the family (first `-CheckSleepingConnections` full lab run, 2026-08-28) and the ownership pattern from #10554.

This issue was created by Claude and reviewed by Andreas Jordan.

andreasjordan:
All three commands are fixed on `development`:

- #10710 Start-DbaXESession (merged 2026-09-16)
- #10711 Stop-DbaXESession (merged 2026-09-14)
- #10712 Remove-DbaXESession (merged 2026-09-14)

Each returns the reconnected store connection after its work, so a pipeline no longer leaves one connection checked out per session. Closing.

---
*This text was created by Claude and reviewed by Andreas Jordan.*

PR #10710, opened 2026-09-13
Start-DbaXESession - Return the store connection after starting a session

andreasjordan:
Follow-up of #10635 and one of the three consumer-side fixes for #10658; the sibling PRs fix the other two commands the same way.

## Problem

`Get-DbaXESession` returns the cloned connection behind its `XEStore` after enumerating (#10635), but the emitted objects keep the store. When `Start-DbaXESession` calls `Start()` on such an object, SFC transparently reconnects the store from the pool, and nothing returned that connection again: one session stayed checked out on the instance per started session until the process ended. Both forms are affected, the piped one and `-SqlInstance`, because the latter goes through `Get-DbaXESession` and the same helper.

Measured on development (SQL03\SQL2019, one NonPooled counting connection, sessions of the test process by `host_process_id`, four rounds of each form): the three commands together leave 12 sessions behind in both editions, Start 3, Stop 4, Remove 5, on top of a one-time pool growth of one from `Get-DbaXESessionTarget` that is not a leak. `Get-DbaXESession` alone and both `ConvertTo-DbaXESession` forms add nothing.

## What changed

The helper disconnects the store in a `finally` after the `Start()` call, the same disconnect `Get-DbaXESession` does after enumerating. SFC reconnects from the pool when the next session on the same store needs it, so several sessions from one `Get-DbaXESession` call still work through one pipeline. The store is reached through the `Store` note property: `Parent` is overwritten with the SMO server by `Get-DbaXESession`. The `finally` also covers the failure path, where the reconnect has already happened when the method throws.

`ConvertTo-DbaXESession` was checked too: it creates the session through the T-SQL script methods and reads it back through `Get-DbaXESession`, and the measurement shows it adds nothing in either form, so it needs no change.

## Measurements

With the three fixes applied, the same script leaves the total at the one-time pool growth in both editions: every Start, Stop and Remove pipeline adds zero over four rounds.

## Tests

New context `When starting sessions repeatedly`, the #10633-#10635 counting pattern with two changes: every session of the test process is counted, not only sleeping ones, because a pooled session flips between sleeping and dormant around a reset and a sleeping-only count moved by one in a process that already held two dozen pooled connections; and both forms are warmed up before the baseline. Three piped rounds and three `-SqlInstance` rounds must then add nothing; the fixture session is flipped through the counting connection between the rounds so only the command under test touches the store.

Through the testing-dbatools harness: green on both editions, alone (9/9) and inside the whole XE family (23 files); red on development's command in both editions, both assertions (piped: expected 10, got 11; -SqlInstance: expected 10, got 13). Sibling PRs cover Stop and Remove.

ðŸ¤– Generated with [Claude Code](https://claude.com/claude-code)


potatoqualitee (approved):
No material defects found. The failing SINGLE 1/5 check is the unrelated reporting/timing failure tracked in #10714 and reproduced on development. Reviewed at head aa9bae6c41fa897ae6f93194c96366547ebb0d17.

potatoqualitee:
unfortunately this caused an error in the related command's test

andreasjordan:
The red job is not this change: the SINGLE 1/5 stage died with "An item with the same key has already been added" right after the file completed, which is the harness crash from #10714 (a failed test whose output holds XE Session objects kills the failure summary before the failure is reported), so the log never says which test failed. The only timing-dependent tests in the file are the `-StartAt` and `-StopAt` ones: schedule an Agent job ten seconds ahead, sleep eleven, assert. On a freshly started instance the Agent can be late, and the change in this PR (returning the store connection after `Start()`) does not touch that path; the `-StartAt` test does not even call `Start()`.

Fixed the tests in 84c5aea88 anyway, so the file stops flaking on a fresh CI instance: schedule twenty seconds ahead and wait up to sixty seconds past that time for the job to have run and deleted itself, polling its row in msdb (enumerating the SMO job collection while the Agent deletes the job throws an invalid cast, which the first version of this poll hit). `AfterAll` now removes the jobs too, since a job that never ran stays behind, and the `-StartAt` test checks the START job at the end instead of the STOP job that the previous test had already removed.

Lab, through the testing-dbatools harness: 9/9 on SQL 2019 and SQL 2022, pwsh and 5.1, warm and immediately after restarting the SQL 2022 engine and Agent, where the job fired about eight seconds after its scheduled time. The harness fix itself stays with #10714.

PR #10711, opened 2026-09-13
Stop-DbaXESession - Return the store connection after stopping a session

andreasjordan:
Follow-up of #10635 and one of the three consumer-side fixes for #10658; the sibling PRs fix the other two commands the same way.

## Problem

`Get-DbaXESession` returns the cloned connection behind its `XEStore` after enumerating (#10635), but the emitted objects keep the store. When `Stop-DbaXESession` calls `Stop()` on such an object, SFC transparently reconnects the store from the pool, and nothing returned that connection again: one session stayed checked out on the instance per stopped session until the process ended. Both forms are affected, the piped one and `-SqlInstance`, because the latter goes through `Get-DbaXESession` and the same helper.

Measured on development (SQL03\SQL2019, one NonPooled counting connection, sessions of the test process by `host_process_id`, four rounds of each form): the three commands together leave 12 sessions behind in both editions, Start 3, Stop 4, Remove 5, on top of a one-time pool growth of one from `Get-DbaXESessionTarget` that is not a leak. `Get-DbaXESession` alone and both `ConvertTo-DbaXESession` forms add nothing.

## What changed

The helper disconnects the store in a `finally` after the `Stop()` call, the same disconnect `Get-DbaXESession` does after enumerating. SFC reconnects from the pool when the next session on the same store needs it, so several sessions from one `Get-DbaXESession` call still work through one pipeline. The store is reached through the `Store` note property: `Parent` is overwritten with the SMO server by `Get-DbaXESession`. The `finally` also covers the failure path, where the reconnect has already happened when the method throws.

`ConvertTo-DbaXESession` was checked too: it creates the session through the T-SQL script methods and reads it back through `Get-DbaXESession`, and the measurement shows it adds nothing in either form, so it needs no change.

## Measurements

With the three fixes applied, the same script leaves the total at the one-time pool growth in both editions: every Start, Stop and Remove pipeline adds zero over four rounds.

## Tests

New context `When stopping sessions repeatedly`, the #10633-#10635 counting pattern with two changes: every session of the test process is counted, not only sleeping ones, because a pooled session flips between sleeping and dormant around a reset and a sleeping-only count moved by one in a process that already held two dozen pooled connections; and both forms are warmed up before the baseline. Three piped rounds and three `-SqlInstance` rounds must then add nothing; the fixture session is flipped through the counting connection between the rounds so only the command under test touches the store.

Through the testing-dbatools harness: green on both editions, alone (6/6) and inside the whole XE family (23 files); red on development's command in both editions, both assertions (piped: expected 18, got 19; -SqlInstance: expected 18, got 21). Sibling PRs cover Start and Remove.

ðŸ¤– Generated with [Claude Code](https://claude.com/claude-code)


potatoqualitee (approved):
No material defects found. Reviewed the complete change and CI at head 0bb71eec7b748338e4d3b4436a58102dab7e18ee.

potatoqualitee:
thank you 💯

PR #10717, opened 2026-09-16
CI - Scope the Azure log shipping test databases by run id so concurrent runs do not share a blob name

### LINKED ISSUE #10667: CI: the Azure log shipping test collides across concurrent runs because its backup name is only second-unique

andreasjordan:
Fixes #10667. Carries the `.github/scripts/gh-actions.ps1` part of #10679 by @simonyang08, as proposed in the review there; the command change and the source-reading test file of that PR are left out.

## The flake

On 2026-08-31 three PRs pushed within about 30 seconds ran "sets up log shipping to Azure blob storage using SAS token" at the same second. The test used the fixed database name `dbatoolsci_logship_azure`, and `Invoke-DbaDbLogShipping` names the blob `<database>_FullBackup_PreLogShipping_<yyyyMMddHHmmss>.bak`, so all three runs wrote to the same blob in the shared container. The run for #10665 lost:

```
The file https://dbatools.blob.core.windows.net/dbatools/dbatoolsci_logship_azure_FullBackup_PreLogShipping_20260831075712.bak
exists on the remote endpoint, and WITH FORMAT was not specified. Backup cannot proceed.
```

## The change

Both Azure log shipping test database names (`dbatoolsci_logship_azure` and `dbatoolsci_logship_addsecondary`) now carry `GITHUB_RUN_ID` and `GITHUB_RUN_ATTEMPT`, for example `dbatoolsci_logship_azure_31789177009_1`. Every run and every rerun then works on its own database and its own blobs, whatever the timestamp says. The blob cleanup at the end of each test lists by the database name as prefix, so it keeps finding its own blobs; the derived names of the second test (`_second`, `_missing`) build on the scoped name and follow along. Nothing else in the repository uses the old literal names.

No command change: the collision is a property of the shared CI container and the fixed test name, not of the command, and a finer timestamp would also rename the on-premises backup file for every user.

## What this does not do

The first test's cleanup sits after its assertions, without a `finally`. A run that fails before it now leaves its blobs behind under its own run id, where the next run's broad prefix used to sweep them. Sweeping the old broad prefix again would delete another concurrent run's backup while it restores, which is the same race from the other side, so the leftovers are the price of the fix and can be removed by hand from the container when it matters.

## Verification

The script parses, and the name resolves to `dbatoolsci_logship_azure_<run id>_<attempt>` with the GitHub environment set. `Run Cross Platform Tests` triggers on the push of this branch with the real SAS token, so the Azure tests run on this PR's CI.


potatoqualitee (approved):
Reviewed the complete CI-script change, linked collision reports, name propagation, cleanup behavior, workflow topology, and exact-head logs. Run ID plus attempt isolates both Azure log-shipping tests without exceeding identifier limits, and both affected live Azure paths passed at this head. I found no material defect.

potatoqualitee:
very nice thank you

PR #10722, opened 2026-09-18
New-DbaComputerCertificate, New-DbaComputerCertificateSigningRequest - Add -Provider to create the private key in a Key Storage Provider

andreasjordan:
Implements the `-Provider` proposal from #10713 (second and third comment there). **Stacked on #10721**: this branch contains its two commits, because both change `New-DbaComputerCertificateSigningRequest` and its test file. Merge #10721 first; the diff here then shrinks to the last commit. Until #10721 is in, a Key Storage Provider certificate from this parameter would still be refused by `Test-DbaNetworkCertificate` and could not be configured by `Set-DbaNetworkCertificate`.

## What changes

`New-DbaComputerCertificate` and `New-DbaComputerCertificateSigningRequest` get

```powershell
[ValidateSet("Microsoft RSA SChannel Cryptographic Provider", "Microsoft Software Key Storage Provider")]
[string]$Provider = "Microsoft RSA SChannel Cryptographic Provider"
```

The legacy CSP stays the default, so existing callers see no change: it is what every supported SQL Server version accepts and what the Microsoft certificate requirements name. Whether the default should move to the Key Storage Provider at some point is a separate decision.

With the Key Storage Provider the certreq configuration differs in three lines: `ProviderName = "Microsoft Software Key Storage Provider"`, `KeyAlgorithm = RSA`, and no `KeySpec` / `ProviderType`, which belong to the CSP world. Everything else (self-signed request, PKCS10 request to a CA, `certreq -accept`, PFX export and import on a remote computer) is unchanged, because certreq generates the key locally in every path and the PFX carries the provider.

## Tests

- `New-DbaComputerCertificate.Tests.ps1`: two unit tests on the captured request configuration (KSP lines present and CSP lines absent, and the reverse for the default); an integration context that creates a self-signed certificate with each provider and checks the provider, the key usage of the CSP key (`AllUsages`, which is how CNG reports `AT_KEYEXCHANGE`) and the key file directory, and one that creates a Key Storage Provider certificate for the `InstanceSingle` host and reads the provider back on that host, which is the PFX transfer path.
- `New-DbaComputerCertificateSigningRequest.Tests.ps1`: a request with the Key Storage Provider whose pending entry in `LocalMachine\REQUEST` holds the key in that provider; the parameter list.

## Verification

Lab (Windows Server 2025 hosts, `InstanceSingle` on SQL03, tests through the harness):

- Windows PowerShell 5.1, both test files on the branch: 20/20 green (`New-DbaComputerCertificate` 17, `New-DbaComputerCertificateSigningRequest` 3). The remote context created the Key Storage Provider certificate on ADMIN01, shipped it as PFX to SQL03 and read the provider back there.
- PowerShell 7.6.3, both files: 20/20 green.
- Red on old (the #10721 code with the new test files, 5.1): the two parameter lists, the Key Storage Provider unit test, the whole Key Storage Provider integration context (the `BeforeAll` fails on the unknown parameter) and the Key Storage Provider request test fail; the default-provider unit test passes on the old code, as it should.
- The certreq configuration for the Key Storage Provider was proven earlier for the self-signed path, against the lab CA (`WebServer` template, issued at once) and for the PFX transfer to a remote host; that evidence is in the comments on #10713.
- Both machines left clean: no test certificates in the stores, pending requests removed.


potatoqualitee (changes requested):
Blocking test gap: the new real provider assertions in tests/New-DbaComputerCertificate.Tests.ps1 are inside Describe -Skip:([bool]$env:appveyor), while .github/workflows/ci-azure.yml sets APPVEYOR to True. As a result, Azure's green run skips the local KSP key, legacy CSP default, and remote PFX/provider-preservation tests; the running unit test only verifies generated request text. This is materially misleading for new cryptographic-provider behavior and does not meet this repository's required real-boundary CI coverage. Please move the new provider context outside the skipped Describe (or narrow/remove the skip) so these assertions run on the Azure Windows lane, and verify that lane passes.

andreasjordan:
Fixed in afee47071.

- The Key Storage Provider context now sits in its own `Describe` without the skip, so the local KSP key, the legacy CSP default and the provider on the target computer are asserted on the Azure lane. The pre-existing skipped `Describe` with its three old contexts is untouched.
- On CI the instance is local, so the PFX export/import path is only exercised against a lab whose instance runs on another computer; the test says so in a comment. The provider assertion on the target still runs on CI.
- The branch took `development` first (1c6c1f708), because #10721 was squash-merged and this branch was stacked on it.

Verified: the SINGLE 1/5 lane ran `New-DbaComputerCertificate.Tests.ps1` in 13.5 s (7.9 s before, when the context was skipped) and is green. Lab 17/17 in Windows PowerShell 5.1 and pwsh 7 with the remote certificate created on SQL03.

andreasjordan:
Follow-up in ada0e8714: the skip on the integration `Describe` is gone altogether, since you offered that option and the runners create certificates without trouble. All four contexts (default settings, custom settings, document encryption, Key Storage Provider) now run on the Azure lane; the file has no `-Skip:` left. The provider context is back in the one `Describe`, so the diff against development is the new context plus the removed skip.

Verified: SINGLE 1/5 ran the file green in 11 s with no retry; the lab gives 17/17 in Windows PowerShell 5.1 and pwsh 7.

potatoqualitee (approved):
Reviewed the complete provider implementation and tests at this exact head, including legacy CSP/KSP behavior, downstream paths, prior discussion, and exact-head Azure logs. The previously skipped real provider coverage now runs and passes in CI. No material defects found.

potatoqualitee:
thank you 🙏🏼

PR #10723, opened 2026-09-18
Get-DbaBackupInformation - Keep the readable files when one file of a Path array cannot be read

andreasjordan:
Follow-up to #10701, where this was noticed and left open (see the reply on the review there): the pipeline form `$bad, $good | Get-DbaBackupInformation` returned the good backup after that fix, but the array form `-Path $bad, $good` still returned nothing.

## Why

All files of one process invocation go into a single `Read-DbaBackupHeader` call with `-EnableException`. `Read-DbaBackupHeader` handles an unreadable file with `Stop-Function -Continue` inside its own loop, which under `-EnableException` throws at the first one, so the batch fails as a whole and the readable files are lost with it. The catch in `Get-DbaBackupInformation` then warned once for the instance and returned.

## What changes

When the batch read fails, the files are read one by one: the readable ones are used exactly as before, and each unreadable one gets its own warning that names the file (`Failure on <instance> reading <file>`). The batch read stays the fast path, so nothing changes for the normal case. Under `-EnableException` the batch failure still throws, as it did.

## Tests

`tests/Get-DbaBackupInformation.Tests.ps1`, context "When a file is not a backup": one new test that passes the text file and a real backup in one `-Path` array and asserts the backup history of the real one comes back and the warning names the text file. The three existing tests of that context (warning once, pipeline form, throw under `-EnableException`) are unchanged and still pass.

## Verification

Lab (`InstanceSingle` = SQL03\SQL2019, tests through the harness):

- Windows PowerShell 5.1: `Get-DbaBackupInformation.Tests.ps1` 23/23 green, run again on the final commit after the last style change.
- PowerShell 7.6.3: 23/23 green.
- Red on old (v2.9.0 code with the new test file, 5.1): exactly the new test fails, with the array form returning nothing; the other 22 pass.


potatoqualitee (approved):
Reviewed the complete change at this head, surrounding backup-header/error handling, originating discussion, tests, and CI. The per-file fallback preserves readable members of a mixed Path array, emits filename-specific warnings for failures, and retains terminating behavior with EnableException. The focused real-SQL regression executed successfully in the Azure SINGLE lane. I found no material defect.

potatoqualitee:
nice change, thank you

PR #10728, opened 2026-09-20
Connect-DbaInstance - Return the connection of a server object after reading its current database

andreasjordan:
## Summary

`Connect-DbaInstance -SqlInstance <server object> -Database X` has to know which database the connection of that object is on, and a server object that was connected without `-Database` has an empty `CurrentDatabase` until its connection is opened once. The command opened it with an explicit `ConnectionContext.Connect()` and left it open.

That connection never went back to the pool: SMO returns a pooled connection after each batch only when it opened the connection itself, an explicit `Connect()` keeps it checked out until `Disconnect()`. So every server object that went through this path kept one session for the life of the process, which only a `Disconnect-DbaInstance` on that very object could release. `ClearAllPools` does not touch a checked-out connection either.

The visible consequence, found during the parallel full runs of 2026-09-19/20: a loop that hands fresh server objects to `Invoke-DbaQuery -SqlInstance $server -Database X` exhausts the 100-connection pool after 100 iterations and every later connect fails with "Timeout expired ... obtaining a connection from the pool". The test harness hit that after ~100 test files because its environment check does exactly this once per file.

The fix reads the value and returns the connection right away, unless the caller had opened it before, in which case it stays theirs. SMO reconnects on the next batch as it always did. The value is kept in a variable, because `Disconnect()` clears `CurrentDatabase` again.

The clone path itself (Copy, DatabaseName, the connection string variants) was checked step by step on the lab and is clean; the only session that stayed behind was the one of the object passed in.

## Measurement

Sessions of the calling process on the instance, counted after `ClearAllPools` so that only checked-out connections and the counting session itself remain (pwsh, SQL Server 2019):

| Step | Sessions |
| --- | --- |
| baseline | 1 |
| `$server = Connect-DbaInstance` (no -Database) | 1 |
| `Invoke-DbaQuery -SqlInstance $server -Database master -Query "SELECT 1"` | 2, `$server.ConnectionContext.SqlConnectionObject.State` = Open |
| `$server.ConnectionContext.Disconnect()` | 1 |
| server connected with `-Database master`, then the same query | 1 |

Windows PowerShell 5.1 behaves the same.

## Test plan

- [x] New Context in `Connect-DbaInstance.Tests.ps1`: three server objects with distinct client names go through `Connect-DbaInstance -Database master` (same database, no clone) and through `Invoke-DbaQuery -Database tempdb` (clone), the session count after `Clear-DbaConnectionPool` must not grow.
- [x] Red on old: both new tests fail on `development` ("Expected 2, but got 5" and "Expected 2, but got 8"), the rest of the file passes.
- [x] Whole file green on the lab with the fix, PowerShell 7 and Windows PowerShell 5.1 alike: 59 tests, 57 passed, 2 skipped (Azure SQL Database).


potatoqualitee (approved):
Reviewed the complete connection-lifecycle patch at this exact head, including SMO ownership semantics, pooled/multi-input/error cleanup paths, regression coverage, discussions, and the passing exact-head MULTI CI lane. No material defects found.

potatoqualitee:
going to get chatgpt pro on this, brb

potatoqualitee:
awesome, pro approves 👏🏼

PR #10729, opened 2026-09-20
Stop-DbaProcess - Skip system sessions instead of trying to kill them

andreasjordan:
## Summary

`Stop-DbaProcess` called `KillProcess` for every session `Get-DbaProcess` returned, system sessions included. SQL Server refuses those with "Only user processes can be killed", so each of them ended in a `Couldn't kill spid N` warning, and that warning reaches the caller's `-WarningVariable` even through `-WarningAction SilentlyContinue`.

Until SQL Server 2022 this was mostly invisible, because system sessions sat below spid 50 and in `master`, where nobody aims a `Stop-DbaProcess -Database` at them. SQL Server 2025 parks system workers such as the "DB MIRROR" tasks above spid 50 and in user databases. On the lab's clustered SQL Server 2025 instance (FCI01) a "DB MIRROR" worker born by a mirroring test stayed in whatever user database was current, and from then on every restore of an existing database (`Invoke-DbaAdvancedRestore` kills the sessions of the database first, with `-WarningAction SilentlyContinue`) and every test that clears the sessions of its database before dropping it got the warning: `Restore-DbaDatabase.Tests.ps1` failed three `$WarnVar | Should -BeNullOrEmpty` assertions and `Set-DbaDbCompression.Tests.ps1` failed its `AfterAll` under `EnableException`.

`Get-DbaProcess` already reports `IsSystem`, which matches `is_user_process = 0` (checked on FCI01: the 29 system sessions above spid 50 are exactly the sessions with `IsSystem = True`). `Stop-DbaProcess` now skips those with a verbose message, the way it already skips its own spid. Nothing changes for user sessions, and the callers in `Invoke-DbaAdvancedRestore` need no change: the `ALTER DATABASE ... WITH ROLLBACK IMMEDIATE` that follows the kill handles whatever is left.

## Test plan

- [x] New Context in `Stop-DbaProcess.Tests.ps1`: a system session of the instance piped in and selected by `-Spid` produces no output and no warning. Every instance has system sessions, so this runs on every lane.
- [x] Red on old: both new tests fail on `development` with the `Couldn't kill spid` warning in `$WarnVar`.
- [x] Lab, with the DB MIRROR worker still parked on FCI01: `Restore-DbaDatabase.Tests.ps1` and `Set-DbaDbCompression.Tests.ps1` under the FCI configuration on PowerShell 7 (the two files that reproduced the failure alone).
- [x] `Stop-DbaProcess.Tests.ps1` green on PowerShell 7 and Windows PowerShell 5.1.


potatoqualitee (approved):
Reviewed the complete command and test changes at this exact head, including system-session classification, pipeline and explicit-SPID paths, callers, discussions, and passing integration CI. No material defects found.

potatoqualitee:
merci 🙏🏼

PR #10731, opened 2026-09-20
Update-ServiceStatus, Set-DbaPrivilege - Keep test variables out of the module scope and filter on bound parameters

andreasjordan:
## Summary

`Set-DbaPrivilege.Tests.ps1` failed its "does not pass the credential ... for the local computer" test in the reversed full run of 2026-09-19, and only there. Two test files together cause it:

- `Update-ServiceStatus.Tests.ps1` set up `$script:credential` (and a few other fixtures) in a `BeforeAll` inside `InModuleScope dbatools`. Pester runs such a `BeforeAll` in the script scope of the dbatools module itself, so every variable it assigns lands in the module, with or without the `$script:` prefix (checked: dropping the prefix alone changed nothing). The module stays loaded for the whole process, so the variable outlives the file.
- A Pester mock parameter filter defines only the parameters the mocked command was called with (`Get-ContextToDefine`); every other name resolves through the scope chain, up to the module's script scope. `{ $null -eq $Credential }` in `Set-DbaPrivilege.Tests.ps1` therefore read the leaked credential whenever `Update-ServiceStatus.Tests.ps1` had run earlier in the same process, and matched 0 of 1 calls.

Repro in a fresh process: Update-ServiceStatus then Set-DbaPrivilege = "Expected Test-PSRemoting in module dbatools to be called 1 times exactly, but was called 0 times"; Set-DbaPrivilege alone passes.

Both sides are fixed:

- `Update-ServiceStatus.Tests.ps1` sets everything up in `BeforeEach`, which runs in a scope of its own that ends with the test, and uses no `$script:`. The two call recorders that the fake `New-CimSession`/`Remove-CimSession` write to are entries of a hashtable owned by the test scope, because an assignment inside those functions would create a local variable. After the file has run, the module scope holds none of its variables (checked with `Get-Variable -Scope Script` inside the module).
- `Set-DbaPrivilege.Tests.ps1` asks the question it means: `-not $PesterBoundParameters.ContainsKey("Credential")`, which no variable from anywhere else can satisfy.

Left alone: `Set-DbaPrivilege.Tests.ps1` itself writes `$script:mockServiceUser`, and `Sync-DbaAvailabilityGroup.Tests.ps1` and `Start-DbaMigration.Tests.ps1` use `$script:` call recorders inside `InModuleScope`; those names collide with nothing.

## Test plan

- [x] Red on old: fresh process, `Update-ServiceStatus.Tests.ps1` then `Set-DbaPrivilege.Tests.ps1` (unit tests) = the failure above, and `$script:credential` is set inside the module afterwards.
- [x] Same order with the fix = 0 failed, and no `credential`/`mockCimSession`/`service`/`recorded`/`password` variable in the module scope after either file.
- [x] Both files green through the lab harness on PowerShell 7 and Windows PowerShell 5.1 (Set-DbaPrivilege 5 of 5, Update-ServiceStatus 1 of 1).


potatoqualitee (approved):
Reviewed the complete test-only patch at this exact head, including module-scope isolation, bound-parameter assertions, surrounding implementations, discussions, and passing CI logs for both affected files. No material defects found.

potatoqualitee:
👌🏼

PR #10732, opened 2026-09-20
Update-DbaInstance - Use the temp folder of the test configuration in the tests

andreasjordan:
## Summary

`Update-DbaInstance.Tests.ps1` builds its fake installer folder at a fixed `C:\Temp\dbatools_Update-DbaInstance` (unit tests and the skipped integration Describe). Every other test file writes below `$TestConfig.Temp`, which is `C:\Temp` on CI and a folder per test set on the lab.

With two runs on one machine the fixed path is shared: during the parallel full runs of 2026-09-19 one run created and removed the folder while the other run's environment check saw it and flagged two unrelated test files (`Add-DbaDbFile`, `Add-DbaDbMirrorMonitor`) as leaving a legacy temp folder behind.

The folder now lives in `$TestConfig.Temp`, like everything else. On CI that is the same `C:\Temp` as before.

One assertion had to follow: the `-Download` test expects `Save-DbaKbUpdate` to be called with the local temp folder, which is what the command does for a local `-Path`. For a network share it downloads into the share itself (`Test-NetworkPath` at the top of the download loop). On a lab where `$TestConfig.Temp` is a share, the expected folder is therefore the share, and the test now derives the expectation from the same rule. On CI nothing changes.

## Test plan

- [x] `Update-DbaInstance.Tests.ps1` green on PowerShell 7 and Windows PowerShell 5.1 on the lab, where `$TestConfig.Temp` is a UNC path: 69 tests, 68 passed, 1 skipped.
- [x] Before the assertion change, the same run failed exactly that one download test, with `Save-DbaKbUpdate` called with the share as `-Path`.
- [x] No `dbatools_Update-DbaInstance` folder left behind, neither below `$TestConfig.Temp` nor in `C:\Temp`.


potatoqualitee (approved):
Reviewed the complete patch at this exact head, including fixture paths, download assertions, production path handling, cleanup, discussions, and CI logs exercising the affected test. No material defects found.

PR #10733, opened 2026-09-20
Connect-DbaInstance - Skip the Windows SSPI test when the instance is not local

andreasjordan:
## Summary

`Connect-DbaInstance.WindowsSspi.Tests.ps1` creates a local Windows user on the machine running the tests and a login for it on `$TestConfig.InstanceSingle`. That only works when the instance runs on the same machine: a remote instance cannot resolve a local account of another computer, and `CREATE LOGIN [ADMIN01\dbasspi...] FROM WINDOWS` fails with error 15401, which fails both tests of the file in the `BeforeAll`. On the lab this was deterministic in every PowerShell 7 run (2026-09-01, 2026-09-19/20).

The Context now skips when the instance is not local, next to the existing platform skip. `Connect-DbaInstance.Tests.ps1` and five other files already use the same `IsLocalHost` skip for behaviour that only a local instance can show. The CI runners host their instances locally, so nothing changes there.

Worth knowing: CI runs Pester on Windows PowerShell 5.1 only, and the Context already skips outside PowerShell 7, so this test currently only runs on a lab with PowerShell 7 and a local instance. The skip added here does not change that.

## Test plan

- [x] Red on old: the file fails 0 of 2 on the lab (PowerShell 7, remote Single) with the 15401 error in the `BeforeAll`.
- [x] With the skip: 2 skipped on PowerShell 7 and on Windows PowerShell 5.1, no local user left behind.


potatoqualitee (approved):
Reviewed the complete patch at this exact head, including SSPI test setup/cleanup, locality behavior, surrounding implementation, discussions, and the dedicated passing CI execution. No material defects found.

potatoqualitee:
nice looks good thanks

